PR #1017, opened 2016-09-17
Migrate helper function from workflow.R to utils

mdietze:
@robkooper, you created most of these functions. Any reason they can't be moved into utils?


robkooper:
Only reason they were in the workflow.R and not utils package was that these functions are very specific to workflow.R and it did not seem like they had a lot of additional use for them.

Issue #835, opened 2016-05-13
Updates to mstmip vars


dlebauer:
Here is a markdown version of our current mstmip* tables: https://gist.github.com/dlebauer/e5fb9fb5f09735ccf791d118edd2984b

1. especially in mstmip_local, 
  * many variables are missing standard_name and description fields
  * many variables have incorrect 'type', e.g. carbon pools and energy fluxes listed as carbon fluxes (fluxes should have units of [mass or energy] per area / per time 
2. both GitHub preview and R's read.csv complain about the quoting / encoding of these files

I know that these tables are called by the `mstmipvars()` function that simplifies conversions and writing to netcdf files, but we should discuss (or at least all take a pause to consider once again) how to reduce redundancy with the trait.dicionary and betydb.variables tables.

Issue #729, opened 2016-02-22
Fix hack in trait.lookup function


tonygardella:
Currently within the utils package we have a trait.lookup function that looks up trait names and units from a table within the package itself. Taken straight from the function it says:
```
#HACK: shameless hack
#Ultimately we'll want this to be read once at the start of run time
#This could also be represented in the database, 
#but because it is used to determine which parameters to feed to the model,
#it could be argued that it's conceptually model specific
```

I'm opening this issue to turn this into a function that uses the database to return trait names and units rather than use the hard coded table within the package.

dlebauer:
I think the trait.dictionary is a hack that would be fixed by either adding
columns to the variables table or a separate table of variable attributes
that could store equivalent names from different vocabularies.

aditi25srivastava:
@dlebauer Heyy!!!!! I would love to work on this issue. Could you please assign it to me ?

dlebauer:
@aditi25srivastava thank you for your interest helping with this! since this is an old issue could you please start by checking if this is still relevant? There has been a lot of work to decouple PEcAn from BETYdb - it was originally designed with BETYdb as a requirement - so the solution should rely on a dataset inside PEcAn like trait.dictionary. And I'm not sure what the status is - though trait lookup is a very useful function.

If you want to query the entire betydb variables table you can 

```
install('traits')
library(traits)
betydb_search('variables')

```

Issue #622, opened 2015-09-11
Document mstmip_vars and mstmip_local data sets. 


dlebauer:
The mstmip data embedded in the utils package contains the following fields, but I can not find the documentation. What do the fields mean, and how are they used?

``` r
data(mstmip_local, package = "PEcAn.utils")
colnames(mstmip_local)
 [1] "Num"           "Group"         "order"         "Saveit"        "Variable.Name" "Units"         "Long.name"    
 [8] "Priority"      "Category"      "X3.hourly"     "Monthly"       "var_type"      "ndim"          "dim1"         
[15] "dim2"          "dim3"          "dim4"          "Description"  
```

These datasets can be documented using oxygen as described in Hadley's chapter on data in R packages. http://r-pkgs.had.co.nz/data.html

mdietze:
http://nacp.ornl.gov/MsTMIP_variables.shtml

dlebauer:
Here are the mstmip_vars (from MsTMIP) and mstmip_local (ad-hoc definitions) datasets. Some fields arent clearly described (e.g. 'priority', 'group', 'order', 'Saveit'). I assume 'Num' is the unique row identifier (?).

I'd suggest the following for `mstmip_local` variables.  

1. remove duplicates
2. Add human-readable versions of variable names to the Long.Name field
3. Update Category field (currently, all PEcAn additions use "Carbon Flux" as a category even though some variables belong to "Carbon Pool", "Energy Flux", or "Physical Variables"
4. add variable_id field to reference BETYdb variables table.

### MsTMIP 

|Variable.Name |Long.name                                |Units                              |Category           |
|:-------------|:----------------------------------------|:----------------------------------|:------------------|
|lon           |Longitude                                |degrees_east                       |Grid               |
|lat           |Latitude                                 |degrees_north                      |Grid               |
|lon_bnds      |Longitude west-east bounds               |degrees_east                       |Grid               |
|lat_bnds      |Latitude south-north bounds              |degrees_north                      |Grid               |
|time          |Time middle averaging period             |days since 1700-01-01 00:00:00 UTC |Time               |
|time_bnds     |Time beginning-end bounds                |days since 1700-01-01 00:00:00 UTC |Time               |
|dec_date      |Decimal date middle averaging period     |yr                                 |Time               |
|dec_date_bnds |Decimal date beginning-end bounds        |yr                                 |Time               |
[...]

PR #1019, opened 2016-09-18
Maat.model

serbinsh:
More updates to MAAT wrapper code to support additional functionality.  Also contains a few bug fixes.


models/maat/R/utils.R
```diff
@@ -1,34 +0,0 @@
-## Helper functions for the MAAT model
```
serbinsh: Removed this function....now unnecessary

models/maat/R/write.config.MAAT.R
```diff
@@ -35,6 +35,11 @@ convert.samples.MAAT <- function(trait.samples){
     trait.names[trait.names == "leaf_respiration_rate_m2"] <- "atref.rd"
     trait.names[trait.names == "Vcmax"] <- "atref.vcmax"
     trait.names[trait.names == "Jmax"] <- "atref.jmax"
+    trait.names[trait.names == "Ev_Arrhenius"] <- "Ha.vcmax" # Arrhenius activation energy
+    trait.names[trait.names == "Ej_Arrhenius"] <- "Ha.jmax" # Arrhenius activation energy
+    trait.names[trait.names == "stomatal_slope"] <- "g1_leuning"
+    trait.names[trait.names == "stomatal_slope.g1"] <- "g1_medlyn"
+    trait.names[trait.names == "stomatal_slope.BB"] <- "g1_ball"
```
serbinsh: g1_ball requires RH in met file.  Updates to MAAT model code needed

models/maat/R/write.config.MAAT.R
```diff
@@ -112,11 +125,10 @@ write.config.MAAT <- function(defaults=NULL, trait.values, settings, run.id){
   } else if (!is.null(settings$run$inputs$met)){
     met.dir <- dirname(settings$run$inputs$met$path)
     met.file <- basename(settings$run$inputs$met$path)
-    logger.info("-- Copy leaf_user_met.xml to rundirs --")
+    #logger.info("-- Copy leaf_user_met.xml to rundirs --")
```
serbinsh: Printing on every line.  Not nescessary

models/maat/R/write.config.MAAT.R
```diff
@@ -112,11 +125,10 @@ write.config.MAAT <- function(defaults=NULL, trait.values, settings, run.id){
...
     file.copy(file.path(met.dir,list.files(met.dir,'*.xml')), rundir, overwrite = TRUE, recursive = FALSE,
               copy.mode = TRUE, copy.date = TRUE)
-    logger.info("-- Create job.sh scripts and place in rundirs --")
-    # need to update code below to include met data dir
+    #logger.info("-- Create job.sh scripts and place in rundirs --")
```
serbinsh: Printing on every line.  Not nescessary

models/maat/R/model2netcdf.MAAT.R
```diff
@@ -89,6 +89,7 @@ model2netcdf.MAAT <- function(outdir, sitelat=-999, sitelon=-999, start_date=NUL
     output[[1]] <- out.year                       # Simulation year
     output[[2]] <- sub.maat.doy+day.steps         # Fractional day - NEED TO IMPLEMENT  
     output[[3]] <- (sub.maat.output$A)            # assimilation in umolsC/m2/s
+    output[[4]] <- (sub.maat.output$gs)           # stomatal conductance in ???
```
serbinsh: Need to confirm units then output gs in CF convention at the bottom of the script

models/maat/R/model2netcdf.MAAT.R
```diff
@@ -118,17 +119,20 @@ model2netcdf.MAAT <- function(outdir, sitelat=-999, sitelon=-999, start_date=NUL
     Mc <- 12.017 #molar mass of C, g/mol
     umol2kg_C <- Mc * ud.convert(1, "umol", "mol") * ud.convert(1, "g", "kg")
     
-    ### Convert outputs
+    ### Find missing and convert outputs
     #output <- conversion( 2, umol2kg_C)  ## convert GPP in umolC/m2 s-1 to kgC/m2 s-1 (MsTMIP)
     #output[[2]][output[[2]] != -999] <- output[[2]][output[[2]] != -999] * umol2kg_C
-    output[[3]] <- ifelse(output[[2]]==-999,-999,output[[3]]*umol2kg_C)  # convert A/GPP to kgC/m2/s
+    output[[3]] <- ifelse(output[[3]]==-999,-999,output[[3]]*umol2kg_C)  # convert A/GPP to kgC/m2/s
+    output[[4]] <- ifelse(output[[4]]=="Inf",-999,output[[4]]) # here is where we will convert gs to CF units
```
serbinsh: When g1, gs, or gc are not solved for they get written out of MAAT as Inf.  netCDF will crash writing out Inf so they needs to be converted to missing -999

serbinsh:
@tonygardella Any chance you would be interested, have time to review this within the next week?

models/maat/R/model2netcdf.MAAT.R
```diff
@@ -118,17 +119,20 @@ model2netcdf.MAAT <- function(outdir, sitelat=-999, sitelon=-999, start_date=NUL
...
+    
     
     ### Put output into netCDF format
     mstmipvar <- PEcAn.utils::mstmipvar
     var <- list()
     var[[1]]  <- mstmipvar("Year", lat, lon, t, NA)
     var[[2]]  <- mstmipvar("FracJulianDay", lat, lon, t, NA)
     var[[3]]  <- mstmipvar("GPP", lat, lon, t, NA)
+    var[[4]]  <- mstmipvar("gs", lat, lon, t, NA)
```
serbinsh: @tonygardella Yes right now PEcAn throws an error but the analysis still works fine. I agree we need to move away from buried files to the gitbook for units/names/conventions.  That would be much easier!  I need to look at what the CF would be for a water flux...

dlebauer: @tonygardella instead of having a function query the documentation, could we use the standard_name and standard_units fields in the BETYdb variables table to store the PEcAn-CF / MsTMIP variables, and then generate the table in gitbook from the database (compiling Rmd --> md), rather than querying the documentation from the code, which seems like an odd route?

tonygardella: @dlebauer Would it be a good idea to change the ```mstmipvar`` function to query the database for standard names and units? It would ensure model2netcdf naming conventions are consistent.

For updating gitbook I think your idea makes more sense. I'm not sure if code can be executed from gitbook itself when it's compiled , but manually doing it or making a chron job makes sense.

dlebauer: @tonygardella 

1. yes, have mstmipvar query BETYdb standard name / standard units fields 
2. we can have a build script or makefile for gitbook ... it wouldn't be called by the compiler on GitBook.com, but if we pull, make, push it should work to update the .md files and GitBook.com won't care / won't be required.

mdietze: This isn't trivial because we're not yet consistent in how we use the std name/units column, so it's not clear to me how to do this forward and backward. For example, not everything (or even anything at this point) that has a std name/unit is a PEcAn model variable, so how would we use that name to generate the documentation table?  On the other side, the mstmipvar function isn't passed a database connection, so it can't query the database.

dlebauer:
Luckily we haven't done much with these fields so we can use them here, and
edit any we need.

We could pass db connection info or this would be a simple place to get
info from the API (eg generate tables from betydb.org/variables.json)

mdietze:
You'd have to do the API because the conversions will run on machines that are remote. I'm worried that would be quite a bit of overhead if you're pinging that table constantly. Because variables are updated infrequently, it might be far simpler to write the table into the package at build time and then leave it as a static table internally.

dlebauer:
That makes sense. Would be good to keep the db to csv function separate to allow the user/developer update the table without rebuilding.

tonygardella:
@dlebauer @mdietze Are we not moving away from using the "mstmipvar" function, and instead going to update the csv from the db and update gitbook from the csv?

dlebauer:
@tonygardella I was assuming we would keep the mstmipvars function, which helps by handling the creation of objects suited for insertion into a netcdf file (with dimensions, units, etc). It currently depends on two files, `mstmip_vars.csv` and `mstmip_local.csv` (I think). The change is to generate these files when building the packages or on demand so that there is one definitive reference.

dlebauer:
PS the idea of using betydb.variables to combine information in trait.dictoinary and mstmip_vars is brought up in #622, #835 and #729. I am going to update #835 to cover merging these.

tonygardella:
@dlebauer I remember that our discussion involved phasing ```mstmipvar``` out because we didn't want to rely on a csv table stuck in the package itself. 

If we're going to regularly update the mstmip csv then I'll change my request for changes to just involve @serbinsh updating the ```mstmip_vars.csv``` and ```mstmip_local.csv``` to include all of his outputs, and I'll reflect those changes in the gitbook.

serbinsh:
@tonygardella @dlebauer I am trying to update the CF into in the mstmip_local.csv but cant seem to find a CF stomata, stomatal conductance, etc CF standard name.....Ideas for how to include that into the mstmip vars?

serbinsh:
@dlebauer @tonygardella 

Also it seems in our variables table water fluxes are in 

TVeg 	kg m-2 s-1 	Transpiration

So should I be outputting gs (stomatal conductance) in kg m-2 s-1?

dlebauer:
@serbinsh CF doesn't have a variable but here are some proposed CF style names https://docs.google.com/spreadsheets/d/1oEiDasdTslsm0VXFPWAUKS82BYbzXLyv2CYiicXMEys/edit#gid=0

serbinsh:
@dlebauer then it would seem an appropriate analog could be cuticular cond

```
38	33	cuticular_cond	conductance_of_fully_closed_stomata					cuticular conductance	umol H2O m-2 s-1	umol m-2 s-1	umol m-2 s-1		leaf conductance when stomata fully closed	Intercept for Ball-Berry model. Note units are umol: divide value by 10^3 to convert to mmol; divide by 10^6 to convert to mol

```

serbinsh:
So would we then say for variable names:
Gs: stomatal conductance
Gc: canopy conductance
Gm: mesophyll conductance

??  Given these sorts of vars? Also, would I think consider Gs an "Energy Flux"?  It seems that is how evap is labeled.  Is there a Water Flux category?  @dlebauer @mdietze @tonygardella I havent yet found water flux yet

mdietze:
@serbinsh a few responses and points of clarification:
1) I'd avoid two letter variable names -- that's really not what we're shooting for here!  Better would be stomatal_conductance, canopy_conductance, and mesophyll_conductance
2) I'd keep units consistent with other flux units in the PEcAn Output standard (augmented MsTMIP)
3) I wouldn't forget that CF, BETY, and MsTMIP are all different things -- virtually none of the variables in MsTMIP are the same as their CF standard.
4) That said, we're trying to slowly reconcile these the best we can. In particular, we want to gradually/eventually make sure BETY and PEcAn Output standards are identical so we can deprecate the need to translate between them. If we can get these two to match CF all the better!
5) Therefore, if a PEcAn Output variable doesn't yet exist, and it doesn't exist in CF either, you probably want to go with the BETY name. If not you want to update the BETY name (and check for any code that's dealing with variables by name).
6) Likewise if PEcAn Output and BETY units don't match, I'd advocate updating the BETY units (and the trait data and any write.configs.[MODEL] that's being passed traits in those units)

serbinsh:
Thanks @mdietze and apologies for conflating multiple things together and my general tomfoolery 

```
stomatal_conductance, canopy_conductance, and mesophyll_conductance
```
Great!  Happy to use the more descriptive variable names

```
I'd keep units consistent with other flux units in the PEcAn Output standard (augmented MsTMIP)

```
So that would advocate for the kg m2 s to match the C fluxes


Looking at BETY, there is a GS variable for stomatal conductance in mol H2O m2 s. However there is already a stomatal_conductance variable with strange units (mm s-1).

serbinsh:
So something like

```
Mw <- 18.01528  #g/mol
mol2kg_H2O <- Mw * ud.convert(1, "g", "kg")
output[[4]] <- ifelse(output[[4]]=="Inf",-999,output[[4]]*mol2kg_H2O) # stomatal_conductance in kg H2O m2 s1
```

So that stomatal_conductance for MAAT standard outputs gets converted from mol H2O m2 s to kg m2 s

mdietze:
@serbinsh yes, that looks good. 

For the existing stomatal conductance, recall that for water 1 mm = 1kg/m2 so the stomatal_conductance units are kg m-2 s-1. Go ahead and change that in BETY since they're equivalent and changing the units is more clear.

GS appears to be a duplicate with stomatal_conductance. Check BETY's list of duplicate variables and, if necessary, weight in that stomatal_conducatance is the one we want to keep. If you can't find it in the duplicate list create a new BETY issue

serbinsh:
@mdietze @tonygardella I have made the requested updates to the code and tested with MAAT. This PR should now be good to go.  I only edited the mstmip_local.csv. Should I have also added to mstmip_vars.csv?

PR #1022, opened 2016-09-22
Multisite workflow

rykelly:
Minor updates and a lot of new documentation for `papply`


mdietze:
Will pull as soon as conflicts are resolved and Travis CI is happy

rykelly:
I don't understand Travis' error... It seems like it's annoyed with my example, which previously used `T` shorthand for `TRUE` in an argument to a (custom) function. But I changed that in 0365da6... any thoughts?

PR #1023, opened 2016-09-23
SDA with SIPNET + LINKAGES with more generalization

araiho:
@mdietze and I talked about the base design of read.restart and write.restart today. I implemented a lot of the changes we talked about and got sipnet working again. 

I have a goal of making smaller pull requests. Let me know if this is too small? Then I can work on some more things and do another one later.


models/sipnet/R/read.restart.SIPNET.R
```diff
@@ -4,70 +4,65 @@
...
-  }
+  prior.sla <- params[[which(names(ensemble.samples)!='soil')[1]]]$SLA
   
   forecast <- list()
   
   #Read ensemble output
   ens <- read.output(runid = runid,outdir = file.path(outdir, runid),
-                     start.year = time, end.year=time,
-                     variables=variables)
+                     start.year = strftime(stop.time, '%Y'),
```
istfer: Does the `stop.time` argument passed to both `read.restart.LINKAGES` and `read.restart.SIPNET` have the same format or being used in the same format in `read.output`? because in `read.restart.LINKAGES` you directly use `stop.time`, whereas here you exract `year` by `strftime(stop.time, '%Y')`. Just wanted to note

models/linkages/R/write.restart.LINKAGES.R
```diff
@@ -77,10 +70,18 @@ write.restart.LINKAGES <- function(out.dir, runid, time, settings, analysis.vec,
   }
   
   spp.params <- spp.params.default[spp.params.save,]
+  
+  biomass_function<-function(dbh){ #kg/tree
+    .1193 * dbh^2.393 + ((slta+sltb*dbh)/2)^2 * 3.14 * fwt * frt * .001
+  }
+  merit<-function(dbh){
+    (b_obs - biomass_function(dbh))^2
```
istfer: there are non-local variables in both of these functions : `slta`,`sltb`,`fwt`,`frt`,`b_obs`

if you pass them as an argument explicitly, it would be less error-prone

models/linkages/R/write.restart.LINKAGES.R
```diff
@@ -17,10 +17,10 @@
 ##' @return NONE
 ##' @export
 ##' 
-write.restart.LINKAGES <- function(out.dir, runid, time, settings, analysis.vec,
-                                   RENAME = TRUE, variables,
+write.restart.LINKAGES <- function(outdir, runid, time, settings, analysis.vec,
```
istfer: it seems like arguments of `write.restart.LINKAGES` doesn't agree with how it's being called by `sda.enkf`, would that be a problem?

models/sipnet/R/read.restart.SIPNET.R
```diff
@@ -4,70 +4,65 @@
...
+read.restart.SIPNET <- function(outdir,runid,stop.time,settings,var.names,params){
 
-  if(sample_parameters == TRUE){
-    load(file.path(settings$outdir, "samples.Rdata"))
-    prior.sla <- mean(ensemble.samples[[which(names(ensemble.samples)!='soil')[1]]]$SLA) #HACK
-  }else{
-    load(file.path(settings$outdir, paste0("ensemble.samples.",settings$state.data.assimilation$prior,".Rdata")))
-    prior.sla <- ens.samples[[which(names(ensemble.samples)!='soil')[1]]]$SLA
-  }
+  prior.sla <- params[[which(names(ensemble.samples)!='soil')[1]]]$SLA
```
istfer: `ensemble.samples` is not passed to this function, does it get it from global environment? or should it be `params` as well?

mdietze: Good catch @istfer -- we previously were loading ensemble.samples from file within the function but now we're passing them from params, so that should be names(params) now.

models/sipnet/R/read.restart.SIPNET.R
mdietze: Yes, the date is standard and includes day and month. @araiho you'll want to double check LINKAGES. Also, in other places we've used the lubridate year function rather strftime.

models/sipnet/R/read.restart.SIPNET.R
```diff
@@ -4,70 +4,65 @@
...
+                     start.year = strftime(stop.time, '%Y'),
+                     end.year = strftime(stop.time,'%Y'),
+                     variables = var.names)
     
     last = length(ens$NPP)
     
     forecast<-list()
     
-    unit.conv <- (10000/1)*(1/10000)*(365.25*24*60*60)
+    unit.conv <- (10000/1)*(1/1000)*(365.25*24*60*60)
```
mdietze: please document what this unit conversion actually is. Use udunits2 if possible.

models/sipnet/R/split.inputs.SIPNET.R
```diff
@@ -0,0 +1,46 @@
...
+  
+  start.day<-strftime(start.time, format = "%j")
+  start.year<-strftime(start.time,"%Y")
+  
+  end.day<-strftime(stop.time, format = "%j")
+  end.year<-strftime(stop.time,"%Y")
+  
+  full.met <- c(settings$run$inputs$met$path)
+  new.met  <- file.path(settings$rundir,basename(full.met))
+  file.copy(full.met,new.met)
```
mdietze: Not sure why you're copying the met before splitting it rather than just opening the original file and writing out the subset.

models/sipnet/R/write.restart.SIPNET.R
```diff
@@ -18,85 +18,82 @@
...
 ##' 
-write.restart.SIPNET<- function(out.dir, runid, time, settings, analysis.vec,
-                                RENAME = TRUE, variables,
-                                sample_parameters = FALSE,
-                                trait.values = NA, met){
+write.restart.SIPNET<- function(outdir, runid, start.time, stop.time,
+                                settings, new.state,
+                                RENAME = TRUE,
+                                new.params = FALSE,
+                                inputs){
```
mdietze: Inconsistent variable names being passed between write.restart.SIPNET and write.restart.LINKAGES

models/sipnet/R/write.restart.SIPNET.R
```diff
@@ -18,85 +18,82 @@
...
   
   rundir<-settings$host$rundir
+  variables <- colnames(new.state)
  
   if(RENAME == TRUE) {
     file.rename(file.path(outdir,runid,"sipnet.out"),
-                file.path(outdir,runid,paste0("sipnet.out",time)))
+                file.path(outdir,runid,paste0("sipnet.out",strftime(as.Date(start.time)-1,"%Y"))))
```
mdietze: Why is there are -1 in start.time?  Also, we should probably put the full date in the filename so that we're not overwriting if the time step is subannual

models/sipnet/man/remove.config.SIPNET.Rd
```diff
@@ -1,14 +1,14 @@
+% Generated by roxygen2 (4.0.2): do not edit by hand
```
mdietze: These tags shouldn't be in individual Rd files anymore -- these are showing up because you're using an out-of-date version of roxygen -- you'll want to update and re-roxygenize

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -307,19 +337,19 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL){
           Xbar = laply(FORECAST[t1:t],function(x){return(mean(x[,i],na.rm=TRUE))})
           Xci  = laply(FORECAST[t1:t],function(x){return(quantile(x[,i],c(0.025,0.975)))})
 
-          plot(total.time[t1:t],Xbar,ylim=range(c(Ybar,Xci),na.rm=TRUE),
+          plot(strftime(obs.times[t1:t],'%Y'),Xbar,ylim=range(c(Ybar,Xci),na.rm=TRUE),
```
mdietze: With sub-annual data this change will cause a mess. If you make sure you pass obs.times through as.Date, R should be able to plot dates on the x-axis just fine.

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -481,51 +515,59 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL){
...
+                          settings = settings,
+                          new.state = new.state[i,],
+                          new.params = new.params[[i]],
+                          inputs = inputs,RENAME=TRUE))
       }
       ###-------------------------------------------------------------------###
       ### Run model                                                         ###
       ###-------------------------------------------------------------------### 
-      print(paste("Running Model for Year",total.time[t]+1))
+      print(paste("Running Model for Year",as.Date(obs.times[t])+1))
```
mdietze: what's the +1 in the Date?

models/sipnet/R/split.inputs.SIPNET.R
araiho: I just copied it from the old code

modules/assim.sequential/R/sda.enkf.R
araiho: I did that because it's running the model for the next time not the current loop time. Should I come up with a different idea?

models/linkages/R/read.restart.LINKAGES.R
```diff
@@ -4,26 +4,26 @@
...
+##' @param multi.settings    PEcAn settings object
+##' @param var.names   var.names to be extracted
 ##' 
 ##' @description Read Restart for LINKAGES
 ##' 
 ##' @return X.vec      vector of forecasts
 ##' @export
 ##' 
-read.restart.LINKAGES <- function(outdir,runid,time,settings,variables,sample_parameters=NULL){
+read.restart.LINKAGES <- function(outdir,runid,stop.time,multi.settings,var.names,params=NULL){
```
araiho: I was going to fix SIPNET in one pull request and LINKAGES in another. Or should I just fix them both now?

models/linkages/R/write.restart.LINKAGES.R
araiho: Should the functions be inside this function or somewhere else?

models/sipnet/R/read.restart.SIPNET.R
araiho: does it matter if we use lubridate versus strftime? strftime was being used in sda.enkf before

models/sipnet/R/read.restart.SIPNET.R
araiho: woo! I've never been so happy to learn about a new R package.

modules/assim.sequential/R/sda.enkf.R
mdietze: For just an informational message, maybe explicitly list the start/end dates of the next step. Beyond that, just be really careful with date math and document better your logic. For example, as.Date +1 adds one day but as.POSIX +1 adds one second.

models/sipnet/R/split.inputs.SIPNET.R
mdietze: Yeah, can't remember why we did that -- probably fear of corrupting the original me. But I think it's fine to drop that.

models/linkages/R/read.restart.LINKAGES.R
mdietze: I'd get the function arguments for read.restart and write.restart squared away for both models and the SDA workflow. But I'm fine with splitting the updating of the internals across 2 PR. This particular case (settings vs multi.settings) just looked like a simple bug, which is why I mentioned it.

models/linkages/R/write.restart.LINKAGES.R
mdietze: How about outside the function but inside the package and start the function name with a . (e.g. .merit) which @rykelly claims will keep it from being exported.

models/sipnet/R/read.restart.SIPNET.R
mdietze: Lubridate vs strftime isn't a huge deal. Sticking with one just has the advantage of developers sticking to one thing rather than having to learn/maintain multiple different approaches. Much more egregious examples are when we reinvent the same date processing functions in multiple different PEcAn packages that redundant with lubridate

models/linkages/R/write.restart.LINKAGES.R
rykelly: FWIW, I don't think the  `.` at the start forces the function to be internal, it's just a clue to developers that it is. I can't even find the reference I got that from, but I think it's probably related to discussions [here] (http://stackoverflow.com/questions/7526467/what-does-the-dot-mean-in-r-personal-preference-naming-convention-or-more), which note that (1) *variables* named with a leading `.` aren't shown by `ls()` by default, and (2) normal people wouldn't normally start typing a function name with a `.`. 

I think you should still stick to using roxygen tags to decide whether a function gets exported or not, but I think it's a fine convention (whether or not I pulled it out of my ass) to give non-exported functions names that start with `.` so it's readily apparent to other developers.

modules/assim.sequential/R/sda.enkf.R
araiho: @mdietze So, for write.restart I'm doing start.time = format(as.Date(obs.times[t])+1,"%Y/%m/%d") but that's assuming that we always want +1 day... So I'll have to think of something else. If obs.times is when you want the data to be read, the start.time should be the time step immediately following it right? I guess I could just always pass start.time = obs.times[t] then in write.restart the model can choose what part of the date it wants?

mdietze: I didn't mean to say you had to think of something else, I just meant that you need to better document what you're doing here. If obs.times is of class Date you should be fine. Also, within write.restart you need to be clear (& document) that for models that can start subdaily that you're starting at 00:00 and ending at 23:59 so that we don't accidentally loose days everytime we stop.

dlebauer: back to the strftime vs lubridate discussion ...  

One advantage of lubridate and other modern packages is the fact that code is easier to read and thus essentially self documenting, as this statement would be something like `ymd_hms(obs.times[t]) + days(1)`

strftime is that is part of base R; an advantage because it requires fewer dependencies but a disadvantage because syntax is archaic.

PR #1024, opened 2016-09-24
Miscellaneous updates / bugs fixed

rykelly:
A few fixes (mostly related to ED) independent of recent met.process stuff I've been working on.


models/ed/R/write.configs.ed.R
```diff
@@ -385,11 +385,21 @@ write.config.xml.ED2 <- function(settings, trait.values, defaults=settings$const
...
+      # used to look up default param values. 
+      #
+      # e.g. if you're trying to run temperate.Hydric PFT, which which doesn't exist in ED,
+      # you would need to assign it either an existing but currently unused ED2 PFT number 
+      # (e.g. #2 = "early tropical"), or an unused one (e.g. #18). Either way, that number
+      # isn't necessarily what you'd want to look up default parameters for the PFT. 
+      #
+      # What really should happen is for there to be two settings for each PFT: the "num"
+      # to use to represent the PFT to ED, and the "defaults.PFT" (name or number) to use
+      # for pulling default parameter values.
```
mdietze: I think @robkooper's idea from early on was that if the ED param XML is 100% complete and reproducible (which it isn't quite yet, see https://github.com/EDmodel/ED2/issues/173) then the number is the number we pull default parameter values from and then we can just pass ED the PFTs sequentially (1:n). The disadvantage of that is that PFT numbers need to match the numbers used in the input files (specifically the css file), so I think you are right about needing both. This also points to a slightly sticky provenance problem -- how do we know the css files are referring to the 'right' PFTs if all that's logged in them is a PFT number, and that's internal (no meta-data). Even if we did add metadata that mapped to explicit PEcAn PFTs that doesn't solve the problem because some PEcAn PFTs are equivalent (e.g. variants with the same species list but different parameters included or different priors)

modules/data.land/R/fia2ED.R
```diff
@@ -30,6 +30,7 @@ fia.to.psscss <- function(settings, gridres=0.075) {
 
 	## connect to database
 	con <-  db.open(settings$database$bety)
+	on.exit(db.close(con))
```
mdietze: I'd never seen on.exit before -- that's really nice to know about!

robkooper: that is cool (just as an FYI this is on exit of function, I had to look that up quickly).

rykelly: Yeah I only learned about it recently. I thought from some PEcAn code, honestly, but perhaps not. Anyway, doing all db.open calls like this would save PEcAn developers that obnoxious max.connections.reached error during bugfixing, if nothing else.

Also note the need for `add=TRUE` argument if doing more than one on.exit() call in a function...

PR #1029, opened 2016-10-04
Changes to read and align SoilResp

istfer:
Some changes to the PDA code and benchmark functions to load and align soil respiration data

This week @araiho is up for review, although @bcow also needs to look for benchmark changes


modules/assim.batch/R/pda.load.data.R
```diff
@@ -81,6 +89,9 @@ load.pda.data <- function(settings, con) {
       
       inputs[[i]]$obs <- AMFo
       inputs[[i]]$par <- c(AMF.params$intercept, AMF.params$slopeP, AMF.params$slopeN)
+    }else{
+      inputs[[i]]$obs <- inputs[[i]]$data[colnames(inputs[[i]]$data) %in% inputs[[i]]$variable.name]
+      inputs[[i]]$par <- sd(unlist(inputs[[i]]$obs), na.rm = TRUE) # testing
```
istfer: @mdietze here I pass the sd in the data and pass it along for calculating the likelihood

1) should I leave it this way or update it as part of the MCMC?
2) should I include a multiplicative bias term in the likelihood (see pda.define.llik.fn :point_up: )
3) should I use another likelihood all together for soil respiration?

modelanddata
modelvsdata
resid

modules/assim.batch/R/pda.load.data.R
```diff
@@ -38,13 +40,19 @@ load.pda.data <- function(settings, con) {
     
     vars.used.index <- which(format$vars$bety_name %in% c(inputs[[i]]$variable.name))
     
+    if(format$file_name == "AmeriFlux.level2.h.nc"){
+      time.row <- NULL
+    }else{
+      time.row <- which(format$vars$bety_name %in% c("year", "month", "week", "day", "hour", "minute", "second"))
+    }
```
mdietze: This shouldn't be necessary. First, the query to format should give you the time row automatically. Second, we definitely want to avoid format-specific hacks. Third, the else case will probably fail a lot because it will miss other possible time columns -- it doesn't make sense to rely on catching any possible column name when @bcow already has a way to catch time columns based on time formatting.

modules/assim.batch/R/pda.load.data.R
mdietze: 1) seems like an OK interim solution, but we'll want to generalize
2) unfortunately yes. The data itself is not consistent between the chambers and the tower, but I believe the soil resp data provides important info on the temperature (and possibly moisture) sensitivity of Rsoil
3) So the fit right now is too biased to assess the residuals, but if you imagine the multiplicative bias term added then Gaussian might not  be that bad. Need to reassess once that's in there with standard residual and QQ plots

modules/assim.batch/R/pda.load.data.R
istfer: Oh, OK, I see. You're referring to these lines right?

```
    st <- format$vars$storage_type
    time.row <- which(nchar(st)>1 & substr(st, 1,1) == "%")
```

 Makes sense. It was in calc.benchmak, my workflow doesn't go through calc.benchmark but I'll add the lines in `pda.load.data`

mdietze: Rather than adding it to pda.load.data, can we move it into the retrieval of the format itself and remove it from calc.benchmark? That way we're consistent and don't have to replicate this in other places that use load.data

istfer: It seems possible, @bcow is currently using `time.row` while calling `load.data` only in her workflow, `format` is also being passed to `load.data`, so` time.row` can be extracted from `format` within `load.data` function

@bcow would you have any objection moving it?

mdietze: I meant put it in query.format, not read.data

modules/benchmark/R/calc.benchmark.R
```diff
@@ -48,25 +48,21 @@ calc.benchmark <- function(bm.ensemble, con){
...
+    dat_vars <- format$vars$pecan_name # IF : is this line redundant?
     obvs_full = obvs
     
     # ---- LOAD MODEL DATA ---- #
 
-    model_vars <- format$vars$pecan_name[-time.row]
+    model_vars <- format$vars$pecan_name[-time.row] # IF : what will happen when time.row is NULL? 
+    # For example "AmeriFlux.level2.h.nc" format (38) has time vars year-day-hour listed, but storage type column is empty and it should be
+    # because in load.netcdf we extract the time from netcdf files using the time dimension
+    # we can remove time variables from this format's related variables list or can hardcode "time.row=NULL" in load.x_netcdf function
```
mdietze: First, I don't think this will fail (though it needs to be tested) because here we're removing the time row, so if time.row is null we're not removing anything, but that's fine since time wasn't in the variable list anyway.

Second, since load.data returns all variables in BETY standard names, could we not rely on the column always being named 'time' (or could we not modify read.data to make this true)? If that's the case we could update the time.row variable after load.data by matching (e.g. ```if(is.null(time.row)) time.row = which(colnames(obs) == 'time' ```

Another important thing to consider -- time.row seems ambiguous in what it means. Is it the row in the DATA that contains the time or the row in the RETURNED OBJECT that contains time. I'd assumed the former, which would cause this command to not work anyway. For example, if the format only knew about columns 1, 2 and 38 and col 38 is time, then time.row=38 but format$vars$pecan_name is only length 3.  But looking at the earlier code, it looks like time.row isn't EITHER of these but is the row in the FORMAT that contains the time variable.

Also, shouldn't the model_vars that we load be based on the vars used, not all the vars in the format?

I think the PR is ok, but @bcow should look at these issues

db/R/query.format.vars.R
```diff
@@ -76,14 +76,24 @@ query.format.vars <- function(input.id,con,format.id){
   
     header <- as.numeric(f$header)
     skip <- ifelse(is.na(as.numeric(f$skip)),0,as.numeric(f$skip))
-  
+    
+    
+    # Right now I'm making the inappropriate assumption that storage type will be 
+    # empty unless it's a time variable. 
+    # This is because I haven't come up for a good way to test that a character string is a date format
```
istfer: transferred @bcow's comment from calc.benchmark to query.format.vars

modules/benchmark/R/calc.benchmark.R
istfer: yes, time.row is the row in the `format` list

```
format$vars[,1:4]

  variable_id input_name    input_units storage_type
1    8.60e+01       airT      degrees C             
2    3.79e+02      soilT      degrees C             
3    3.82e+02       year           year           %Y
4    3.83e+02        day            DOY           %j
5    3.84e+02       hour          hours           %H
6    4.79e+02   SoilResp umol C m-2 s-1             
7    1.00e+09     minute        minutes           %M
```
so here `time.row `would be `3,4,5,7`

mdietze: Interesting point. I think all properly formatted netCDF *should* have a time dimension, but I know from experience that it won't always be named 'time' (I think that's true in Ameriflux). And as you note you can flag the time columns explicitly, so that will work in many cases. The edge case we don't have covered is when time is neither named time nor is a variable (i.e. is only a dimension, but named something else), since I don't think we have a way of identifying that in the format (but I could be wrong, for example if ```nc_get_var``` will let you get the time dimension)

mdietze: Good point. @bcow had always been aware that this case (time spread across multiple columns) was coming, but I don't think the existing code handles it yet (unless one of you added that without me knowing). Conceptually it should grab ALL the time columns and combine them into one composite POSIX time variable.

istfer: It is being handled later, `format` stores the info where the time variables are, `load.data` (and the functions it calls, currently `load.csv`, `load.x_netcdf`) reads them,  and then combines time variables into posix [here](https://github.com/PecanProject/pecan/blob/master/modules/benchmark/R/load.data.R#L62) in the `load.data`

PR #1030, opened 2016-10-05
Fix ED remote submission

ashiklom:
Switch from the old, non-functional `system(paste("ssh..."))` approach to @robkooper 's better `remote.execute.cmd`. This was breaking ED in the `remove.configs.ED2` function.

There are a few other places where this was broken -- most importantly `get.model.output.SIPNET` -- but I haven't fixed them yet because they're more complicated and I'm not entirely sure how they work.


models/ed/R/write.configs.ed.R
```diff
@@ -305,29 +305,23 @@ remove.config.ED2 <- function(main.outdir = settings$outdir, settings) {
...
-                    system(paste("ssh ", settings$host$name, " 'ls ", 
-                                 settings$host$outdir,
-                                 "'", sep = ''), intern = TRUE),sep="/")
+    remote_ls <- function(path, pattern) {
+        remote.execute.cmd(host = settings$host,
+                           cmd = "ls",
+                           args = file.path(path, pattern))
+    }
+    config <- remote_ls(settings$host$rundir, "c.*")
+    ed2in <- remote_ls(settings$host$rundir, "ls")
```
robkooper: I think this should not be ls but ED2INc.*

ashiklom: Good catch! Addressed in latest commit.

Issue #1032, opened 2016-10-05
How to perform analyses on derived variables


istfer:
How to do SA, PDA and SDA on derived variables? 

E.g.  curently we write TotalSoilC and LitterC as an output in SIPNET, if we want an SA on SoilC only (which = TotalSoilC - LitterC) how can we do that without writing it explicitly as a variable to the output file?

First ideas : to pass it through pecan.xml as string and parse it out through the workflow
- How to parse it?

@ashiklom's  suggestions:

`"*"(a,b)`

```
 eval(parse(text="5+8"))
[1] 13
```

```
x <- 5
y <- 8
expr <- "sqrt(x^2 + y^2)"
eval(parse(text=expr))
[1] 9.433981
```

What changes need to happen in the settings and the code?

I'll give it a first pass coming week

PR #1036, opened 2016-10-08
SA and PDA on derived variables

istfer:
Per issue #1032 I implemented some changes in the SA and PDA.

For SA : in pecan.xml the derivation can go in `<variable>` tag, so this will also work from web interface:
```
<variable>TotSoilCarb-Litter</variable>
```

For PDA : It was a little more complicated for PDA because now variable name in data also comes into play and it doesn't have to match anything in the derivation (e.g. model output variables could be TotSoilCarb and Litter,  variable in the data file could be SoilCarb and derivation would be TotSoilCarb-Litter)

I thought simplest implementation would be separating variable names in the pecan.xml explicitly:

example-1
```
    ....
     <variable.name>
         <data.var>SoilCarb</data.var>
        <model.var>TotSoilCarb-Litter</model.var>
      </variable.name>
     ...
```
example-2 (of course 2*Qle-NEE is not corresponding LE in the data, just for demonstration here)
```
    ....
     <variable.name>
        <data.var>
          <data.var>LE</data.var>                 <-- main data variable to assimilate against, name should match the variable name in data file                 
          <data.var>UST</data.var>                <-- not used in assimilation directly but necessary for calculating uncertainty, name should match the variable name in data file 
        </data.var>
        <model.var>2*Qle-NEE</model.var>          <-- derived variable to compare with data, names here should match model output variables
      </variable.name>
     ...
```
Thought it's better to discuss on a PR.

@jam2767 is next on review


db/R/query.format.vars.R
```diff
@@ -82,7 +82,7 @@ query.format.vars <- function(input.id,con,format.id){
     # empty unless it's a time variable. 
     # This is because I haven't come up for a good way to test that a character string is a date format
     
-    st <- fv$storage_type
+    st <- vars_full$storage_type
```
istfer: a fix related to the previous PR

utils/R/sensitivity.R
```diff
@@ -43,9 +45,25 @@ read.sa.output <- function(traits, quantiles, pecandir, outdir, pft.name='',
   for(trait in traits){
     for(quantile in quantiles){
       run.id <- sa.run.ids[[pft.name]][quantile, trait]
-      out <- read.output(run.id, file.path(outdir, run.id),
-                         start.year, end.year, variable)
-      sa.output[quantile, trait] <- sapply(out, mean, na.rm=TRUE)
+      
+      # if analysis on a derived variable is requested these lines will split them to variable names that can be read by read.input
+      # e.g. 2*Total_C+Litter.Carbon  :  Currently allows underscore and period separated names
```
istfer: might easily be missing something in the chunk below..

istfer:
in SA `<variable>TotSoilCarb-Litter</variable>`, derived variable name goes all the way to the file names, and I didn't thought about `/` operator which causes a problem (cannot be in the file name), currently I made changes to separate variables that are being used in derivation with `_`, but that's something to be discussed

e.g. for a derived variable like `TotSoilCarb/Litter` an SA file name would be something like "sensitivity.results.1000004591.TotSoilCarb_Litter.2005.2006.Rdata" but that would also be the same for  `TotSoilCarb-Litter`. I can think of a way around it, just making the point for now

istfer:
I'll implement the solution to the file naming issue that's suggested today and resolve the conflicts together after some of the other open PRs merged

istfer:
with the new implementation this is how to pass variable names from pecan.xml:

For SA:
```
 <variable>SoilCarb=TotSoilCarb-Litter</variable>
```

Now the derived variable name on the LHS of the equation will be used in the file name for SA (e.g. `sensitivity.results.ID.SoilCarb.YEAR.Rdata`) if a derivation is not requested the usual way of things apply

For PDA:
```
...
    <variable.name>
      <variable.name>FC</variable.name>
      <variable.name>UST</variable.name>
    </variable.name>
...
 <variable.name>SoilCarb=TotSoilCarb-Litter</variable.name>
```
The derived variable name on the LHS of the equation is also the name of the variable in the data.

In order to keep things simpler in the user end things get a little complicated in PDA, I tested it but there can still be bugs.

I'll resolve the conflicts all together after PRs #1081 #1069 #1063 are merged. There will be conflicts with them.

mdietze:
@jam2767 I need a review
@istfer you have a conflict on assim.batch/R/pda.utils.R

Issue #1041, opened 2016-10-09
Unreliable reference at end of `arrhenius.scaling.traits`


bpbond:
In the `arrhenius.scaling.traits` function in `db/R/query.trait.data.R`, `data` is undefined if we haven't gone through the `if` block above. Presumably there should be an `else` clause setting it to NULL?

...same problem appears in `filter.sunleaf.traits`.

Issue #1040, opened 2016-10-09
Unreliable test at end of `query.site.R`


bpbond:
In `db/R/query.site.R` (relevant code below), there are two problems:
* execution can reach the end of the function with no specified return value (i.e. there's no `else` clause)
* the `if` clause tests `site$lat` twice; I assume one should be `site$lon`?

```R
query.site <- function(site.id, con) {
 ...

  if (!(is.na(site$lat)) && !(is.na(site$lat))) {
    return(site)
  }
} # query.site
```

PR #1043, opened 2016-10-09
Fix end-of-function problems (#1040 and #1041)

bpbond:
Proposed fixes for issues #1040 and #1041. Re 1041, I’m assuming that`arrhenius.scaling.traits` and `filter.sunleaf.traits` should return NULL if there are no covariates.


mdietze:
@bpbond this one needs an 'Update branch' to bring it into sync with all your other changes

PR #1045, opened 2016-10-09
Style cleanup for `models/biocro/R/*.R` scripts

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces; cracked down on over-length lines; changed require calls to library.


bpbond:
Does anyone have insight into why this is failing? I see the proximate cause, that `rgdal` can't be installed on the Travis system, so `install.dependencies` errors out...but `biocro` isn't any different from the other model packages that I can see in this regard. In fact, it doesn't even mention `rgdal`. ???

dlebauer:
Looks like a typo ... I'll take a look at this later. I generally work on the biocro-module2 branch and intermittently merge changes to master. I am not sure how the master branch got borked but I can fix and merge updates.

bpbond:
Thanks. I'm probably missing something obvious.

dlebauer:
@bpbond the PR https://github.com/bpbond/pecan/pull/2 should fix the broken build. Here is the passed test: https://travis-ci.org/PecanProject/pecan/builds/167547082

I reran tidy_dir but am not sure if the changes in my PR are the ones you intended, or if accidentally reverted your changes. Because there are a lot of differences so maybe you did editing by hand?

bpbond:
Thanks @dlebauer ! Some things did get reverted but that's fine, I re-fixed them.

dlebauer:
I think this branch is 'out of sync' because the commits are redundant with https://github.com/PecanProject/pecan/pull/1082 if the merge is non trivial I'll take another look

mdietze:
Actually, Github's revised commit system is enforcing syncs with master even when a PR has no conflicts with other changes. e.g. there's zero overlap with #1082, but the mere fact that I merged it puts an "Update branch" block on this PR

Issue #1049, opened 2016-10-09
Comparison used in place of assignment in model2model.GDAY.R:143


bpbond:
Line 143 in `models/gday/R/model2model.GDAY.R` is clearly wrong - should be an assignment, not a comparison.

```R
 if(year %% 4 == 0) {
        ndays <= 366
      } else {
        ndays <- 365
      }
```

robkooper:
Would this be better to use:
```
yearDays(as.POSIXct(paste0(year, "-01-01"))+1
```
(not sure why but it returns number of days - 1, so 365 for leapyear)

dlebauer:
see also #801

bpbond:
Yeah, FWIW I second the suggestion in #801 - take advantage of the existing `lubridate`function!

PR #1050, opened 2016-10-09
Style cleanup for `models/gday/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces; took advantage of paste0 and file.path; changed require calls to library. Tried to make sure formatR didn’t make things too jaggy. Moved ellipses to end of met2model definition.

Fixes issue #1049.


mdietze:
@bpbond GDAY just needs an "Update branch"

PR #1053, opened 2016-10-09
Style cleanup of `models/linkages/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy.

Note these files have many code oddities…I’ve tried to call out some in comments.


models/linkages/R/met2model.LINKAGES.R
```diff
@@ -63,48 +61,52 @@ met2model.LINKAGES <- function(in.path, in.prefix, outfolder, start_date, end_da
   start_year <- year(start_date)
   end_year <- year(end_date)
   
-  year = sprintf("%04d",seq(start_year,end_year,1))
-  month = sprintf("%02d",seq(1,12,1))
+  year <- sprintf("%04d", seq(start_year, end_year, 1))
+  month <- sprintf("%02d", seq(1, 12, 1))
   
-  nyear = length(year) #number of years to simulate
+  nyear <- length(year)  # number of years to simulate
```
araiho: Should this be seq_len?

bpbond: No...that will produce a vector of values.

mdietze:
@bpbond mentioned on Gitter, but you may have missed it, that this is the PR I'm ready to commit next after it's Updated

PR #1056, opened 2016-10-09
Style cleanup for `models/maespa/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Moved ellipses to end of `met2model.MAESPA.R` declaration. Substituted `seq_len` and `seq_along`.


mdietze:
@bpbond MAESPA got out-of-date when I merged PRELES, so need an Update here.

mdietze:
Maespa out of sync again. Needs an Update gen I can merge

PR #1057, opened 2016-10-09
Style cleanup for `models/preles/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Substituted `seq_len` and `seq_along`.


mdietze:
Ready to go on Preles after Update

PR #1058, opened 2016-10-09
Style cleanup for `models/sipnet/R/*.R` scripts

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Moved ellipses to end of `met2model.MAESPA.R` declaration. Substituted `seq_len`, `seq_along`, and `paste0`.


mdietze:
Ready to go on SIPNET after Update

PR #1059, opened 2016-10-09
Style cleanup for `models/template/R/*.R` scripts 

bpbond:
Just to be complete; not much to do here.


mdietze:
Ready to go in template after Merge

PR #1061, opened 2016-10-10
Style cleanup models jules (again)

bpbond:
Changed stop() to logger.severe()  - per @robkooper suggestion


mdietze:
Needs an Update

PR #1066, opened 2016-10-11
Fia2ed updates

rykelly:
Updates to fia.to.psscss. Main thing is changing how the function chooses which FIA census cycle to draw from. Now you explicitly set the timeframe you're willing to use (defaults to +/- 5 yr). Then you identify all cycles (for a given state) that meet your criteria, and select the one that yields the most plots. 

Also added an overwrite option, cleaned up how the records are stored in BETY's inputs table.


mdietze:
@tonygardella who's assigned to review @rykelly this week?

tonygardella:
@rykelly needs an update, but otherwise I think it looks fine.

rykelly:
Hang on though, think I just fixed a bug.

rykelly:
OK I'm good if you're good @tonygardella , @mdietze

rykelly:
Threw in a bonus commit since this is still lingering...

modules/data.land/R/fia2ED.R
```diff
@@ -144,12 +144,13 @@ fia.to.psscss <- function(settings,
...
-                 'WHERE p.lon >= ', lonmin, ' AND p.lon < ', lonmax,
-                 ' AND p.lat >= ', latmin, ' AND p.lat < ', latmax, 
-                 ' AND p.measyear >= ', min.year, ' AND p.measyear < ', max.year, 
+                 'WHERE p.lon >= ', lonmin, ' AND p.lon <= ', lonmax,
+                 ' AND p.lat >= ', latmin, ' AND p.lat <= ', latmax, 
+                 ' AND p.measyear >= ', min.year, ' AND p.measyear <= ', max.year, 
                  ' GROUP BY p.cn')
 
   pss <- db.query(query, con=fia.con)
+  if(nrow(pss) == 0) stop(logger.error("No pss data found."))
```
mdietze: @rykelly if you want an error to cause the code to stop then you want to use logger.severe, not stop(logger.error)

rykelly: Ah, OK. That's helpful and will work here. It might be helpful to have logger.warn throw a real warning message too. If nothing else the fact that it doesn't prevented me from testing for a warning using testthat::expect_warning(). I had edited logger.message to silently return the message so developers could use it however they want while still relying on logger.*. Just some thoughts. I'll change it here.

PR #1068, opened 2016-10-11
Style cleanup for `modules/assim.sequential/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Substituted `seq_len`, `seq_along`, and `paste0`, and `file.path`.

Note these strange lines in `sda.enkf.R`:

```R
for (i in 2) {
  t1 <- 1
  Xbar <- laply(FORECAST[t1:t], function(x) { mean(x[, i], na.rm = TRUE) })
  Xci <- laply(FORECAST[t1:t], function(x) { quantile(x[, i], c(0.025, 0.975)) })
```

`laply`? (This also occurs lower down in file.) And a `for` loop that only executes once?


modules/assim.sequential/R/sda.enkf.R
```diff
@@ -274,454 +284,516 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL){
...
-    if(interactive() & t>1){
-      t1=1
-      names.y <- unique(unlist(lapply(obs.mean[t1:t],function(x){return(names(x))})))
-      Ybar = t(sapply(obs.mean[t1:t],function(x){
-        tmp <- rep(NA,length(names.y))
-        names(tmp) <- names.y 
-        mch = match(names(x),names.y)
-        tmp[mch] = x[mch]
-        return(tmp)
+    if (interactive() & t > 1) {
```
dlebauer: Does `if(interactive(` actually belong in functions, or is this just used for debugging (in lieu of proper debugging utilities like `debugonce` etc)? Occurs here and on line 351. There do not appear to be any other PEcAn functions that use this (though many scripts do).

mdietze: No, it's not for debugging and yes, it belongs there.  If someone is running sda.enkf from an interactive terminal then the code updates the progress of the enkf time series plot every iteration. This provides a real time update of the code's progress and performance. However, we don't want to generate the same plots if someone submits runs the workflow in batch mode or submits it to a queue.

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -179,72 +187,74 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL){
...
-  obs.times.POSIX = ymd_hms(obs.times)
+  obs.times <- names(obs.mean)
+  obs.times.POSIX <- ymd_hms(obs.times)
   
-  for(i in 1:length(obs.times)){
-    if(is.na(obs.times.POSIX[i])){
-      if(is.na(ymd(obs.times[i]))){
-        print('Error: no dates associated with observations')
-      }else{
+  for (i in seq_along(obs.times)) {
```
araiho: Is there a benefit to doing this over just having obs.times stand alone?

mdietze: If length(obs.time) is zero your way will loop from 1:0 while Ben's will skip the loop.   The latter is the desired behavior

mdietze:
Good to go after an Update

modules/assim.sequential/R/sda.enkf.R
bpbond: Plus, on related note, seq_len will throw an error if given a negative value. Defensive programming!

PR #1069, opened 2016-10-11
Style cleanup for `modules/benchmark/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Substituted `seq_len`, `seq_along`, and `paste0`, and `file.path`.

Note `metric.scatter.plot.R` doesn't seem to do anything - it constructs a ggplot that doesn't get printed or saved (?).


ashiklom:
I'm going to issue a (weak) protest against a 2 space tab width and cite the [Linux Kernel Style Guide](https://www.kernel.org/doc/Documentation/CodingStyle) (which actually requires a whopping _8 spaces_, though I'd be happier with 4). I know it's 2 spaces in the Google R style guide, but I think the Linux style guide convincingly argues that larger indentation is _(a)_ more readable and _(b)_ encourages better coding practices ("if you need more than three indentations in your code, you're screwed anyway").

[Apparently](http://programmers.stackexchange.com/questions/157277/indentation-is-there-any-evidence-to-show-whether-2-or-4-spaces-is-more-readabl) there was a study at Rice University that found that 4 space indentations are better, but nobody seems to be able to find it.

bpbond:
Yes, 4 is the default.

>if you need more than three indentations in your code, you're screwed anyway

Funny.

modules/benchmark/R/metric.residual.plot.R
```diff
@@ -4,35 +4,31 @@
...
   dat$diff <- abs(dat$model - dat$obvs)
   dat$zeros <- rep(0, length(time))
   
-  p <- ggplot(data = dat, aes(x=time)) + 
-    geom_path(aes(y=zeros), colour = "#666666", size=2, linetype = 2, lineend = "round") +
-    geom_point(aes(y=diff), size=4,  colour = "#619CFF") + labs(title=var, x= "years", y="abs(model - observation)")
+  p <- ggplot(data = dat, aes(x = time)) 
+  p <- p + geom_path(aes(y = zeros), colour = "#666666", size = 2, linetype = 2, lineend = "round") 
+  p <- p + geom_point(aes(y = diff), size = 4, colour = "#619CFF") 
+  p <- p + labs(title = var, x = "years", y = "abs(model - observation)")
```
bcow: Why is this format better than what was there before? For the purpose of readability? 
I don't usually see ggplot2 code formatted like this.

modules/benchmark/R/metric.scatter.R
```diff
@@ -2,17 +2,15 @@
...
-metric.scatter.plot <- function(dat, var){
+metric.scatter.plot <- function(dat, var) {
   
   library(ggplot2)
   
-  ggplot(data = dat) + 
-    geom_point(aes(x=model,y=obvs), size=4) + 
-    geom_abline(slope=1,intercept=0, colour = "#666666", size=2, linetype = 2, lineend = "round")  
+  ggplot(data = dat) + geom_point(aes(x = model, y = obvs), size = 4) + geom_abline(slope = 1, intercept = 0, 
+    colour = "#666666", size = 2, linetype = 2, lineend = "round")
```
bcow: I don't mean be difficult but this seems to be making the ggplot command harder to read. As opposed to easier.

modules/benchmark/R/metric.residual.plot.R
bpbond: Personal preference creeps in. It's not my code obviously; can change it back if you'd prefer!

modules/benchmark/R/metric.scatter.R
bpbond: Yeah sorry this is format::tidy going a bit crazy.

bcow:
I'm a bit confused about the reformatting just because in some cases comments that were more than one line were combined or commands that contained a carriage return to make the code fit within a certain width were also recombined - this makes things a little harder to read. 

But at the end of the day it doesn't change how the functions work and cleaning things up is great! I'll work harder to stick with the style guide in the future. 

Also, the plot functions still need work - I'm not surprised I left out something from the scatter plot - thanks for catching it!

bpbond:
formatR is _extremely_ aggressive about reflowing comments. Sorry. I tried to catch and reformat the most egregious cases, but feel free to call out things that still need fixing.

mdietze:
@bcow still need an "Approve" to be able to pull

tonygardella:
@bpbond branch needs an update.

mdietze:
@bpbond ready to go on this one next after Update. Ping me on Gitter when you're ready

PR #1071, opened 2016-10-11
Set negligible_nplant to greater than zero in ED history.csv

ashiklom:
For some PFTs, the default was 0, which breaks ED (due to a `log(0)`).


ashiklom:
The entire file shows up as a diff because I changed the spacing to make it readable in a plain text editor.

PR #1072, opened 2016-10-11
ED verbose path matching

ashiklom:
`write.configs.ed` checks that the paths for the ED css, pss, and site files match, and throws an error if they don't.

To make it easier to diagnose errors related to this, I added `print` statements so that the paths are visible on exit.


ashiklom:
Actually, don't pull this yet. Found some problems in the regular expressions for determining the prefix.

ashiklom:
@mdietze If you say so. I see references to both throughout the PEcAn code and am not sure exactly when we use `print` and when we stick to `logger` functions. But I replaced with `logger.info` in latest commit.

I also made the `.lat.*` regular expression more specific so it actually catches instances of `.lat.` (as in latitude) but ignores, for instance, `late`.

PR #1075, opened 2016-10-11
Style cleanup for `modules/meta.analysis/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Substituted `seq_len`, `seq_along`, and `paste0`, and `file.path` as necessary.


modules/meta.analysis/R/jagify.R
```diff
@@ -16,40 +16,36 @@
...
-                'for', names(result)[i], 'result from citation',
-                citationswithbadstats,'\n',
-                'SE <=0 set to NA \n')
-    r$stat[r$stat <= 0.0] <- NA
-  } 
-  r <- rename.jags.columns(r)
-  return(r)
+  if (length(r$stat[!is.na(r$stat) & r$stat <= 0]) > 0) {
+    citationswithbadstats <- unique(r$citation_id[which(r$stat <= 0)])
+    logger.warn("there are implausible values of SE: SE <= 0 \n", "for", names(result)[i], "result from citation",
```
ashiklom: Why isn't this wrapped?

modules/meta.analysis/R/meta.analysis.summary.R
```diff
@@ -22,74 +22,69 @@
...
     acfplot(mcmc.object[[trait]])
     dev.off()
- 
-    ## G-R diagnostics to ensure convergence    
-    gd <- gelman.diag(mcmc.object[[trait]])
-    mpsrf <- round(gd$mpsrf, digits = 3)
+    
+    ## G-R diagnostics to ensure convergence
+    gd            <- gelman.diag(mcmc.object[[trait]])
+    mpsrf         <- round(gd$mpsrf, digits = 3)
```
ashiklom: Seems like there are still a few places floating around the PEcAn code where this is coded up independently we should try to standardize them.

modules/meta.analysis/R/meta.analysis.write.model.R
```diff
@@ -26,33 +26,42 @@
...
-  model.text <- gsub("TAUB", format(signif(tauB,2), scientific=FALSE), model.text)
-  if(ghs.n == 1)  model.text <- gsub("\\#GGG", '\\#', model.text)
-  if(site.n == 1) model.text <- gsub("\\#SSS", '\\#', model.text)
-  if(trt.n == 1)  model.text <- gsub("\\#TTT", '\\#', model.text)
-  if(ghs.n > 1)   model.text <- gsub("\\#GGG", '', model.text)
-  if(site.n > 1)  model.text <- gsub("\\#SSS", '', model.text)
-  if(trt.n > 1)   model.text <- gsub("\\#TTT", '', model.text)
-  if(pr.dist == 'beta' & pr.param.b < 1) {
+  model.text <- gsub("TAUA", format(signif(tauA, 2), scientific = FALSE), model.text)
+  model.text <- gsub("TAUB", format(signif(tauB, 2), scientific = FALSE), model.text)
```
ashiklom: @mdietze @dlebauer Why write these in the model rather than passing them in as data?

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -6,129 +6,136 @@
...
+    
+    if (p.data <= 0.9995 & p.data >= 5e-04) {
       if (p.data <= 0.975 & p.data >= 0.025) {
         logger.info("OK! ", trait, " data and prior are consistent:")
       } else {
         logger.warn("CHECK THIS: ", trait, " data and prior are inconsistent:")
       }
-    } else if (p.data > 0.9995 | p.data < 0.0005) {
-      logger.debug("NOT OK! ", trait," data and prior are probably not the same:")
+    } else if (p.data > 0.9995 | p.data < 5e-04) {
```
ashiklom: All of these are magic numbers -- they should be coded as constants near the beginning of the function, or, better yet, set as default arguments.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -6,129 +6,136 @@
...
-  row.names(prior.variances) = row.names(prior.distns)
-  prior.variances[names(trait.average), ] = 0.001 * trait.average^2 
-  prior.variances["seedling_mortality", 1] = 1.0
-  taupriors <- list(tauA = 0.01,
-                    tauB = apply(prior.variances, 1, function(x) min(0.01, x)))
+  prior.variances <- as.data.frame(rep(1, nrow(prior.distns)))
+  row.names(prior.variances) <- row.names(prior.distns)
+  prior.variances[names(trait.average), ] <- 0.001 * trait.average ^ 2
+  prior.variances["seedling_mortality", 1] <- 1
+  taupriors <- list(tauA = 0.01, tauB = apply(prior.variances, 1, function(x) min(0.01, x)))
```
ashiklom: More magic numbers. See previous comment.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -6,129 +6,136 @@
...
-    if(p.ma.post <= 0.9995 & p.ma.post >= 0.0005){
+    if (p.ma.post <= 0.9995 & p.ma.post >= 5e-04) {
       if (p.ma.post <= 0.975 & p.ma.post >= 0.025) {
         logger.info("OK! ", trait, " posterior and prior are consistent:")
       } else {
         logger.warn("CHECK THIS: ", trait, " posterior and prior are inconsistent:")
       }
-    } else if (p.ma.post > 0.9995 | p.ma.post < 0.0005) {
-      logger.severe("NOT OK! ", trait," posterior and prior are probably not the same:")
+    } else if (p.ma.post > 0.9995 | p.ma.post < 5e-04) {
```
ashiklom: Given that we use these values in multiple places, we should _definitely_ define constant variables once and then use those instead.

ashiklom: More generally, this block is almost identical to the once above. It would be better to wrap it in its own function and call it twice. DRY.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -6,129 +6,136 @@
...
-  if(file.exists(file.path(pft$outdir, 'post.distns.Rdata'))) {
-    file.remove(file.path(pft$outdir, 'post.distns.Rdata'))
+  # Symlink to post.distns.Rdata (no 'MA' identifier)
+  if (file.exists(file.path(pft$outdir, "post.distns.Rdata"))) {
+    file.remove(file.path(pft$outdir, "post.distns.Rdata"))
   }
-  file.symlink(file.path(pft$outdir, 'post.distns.MA.Rdata'), file.path(pft$outdir, 'post.distns.Rdata'))
-
+  file.symlink(file.path(pft$outdir, "post.distns.MA.Rdata"), file.path(pft$outdir, "post.distns.Rdata"))
+
```
ashiklom: The two paths in here are used multiple times. It would be safer to define them as variables and use those here.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -206,16 +212,10 @@ runModule.run.meta.analysis <- function(settings) {
...
-  return(p.point)
-}
-#==================================================================================================#
-
-
-####################################################################################################
-### EOF.  End of R script file.      				
-####################################################################################################
+  do.call(paste0("p", prior$distn), 
+          list(point, prior$parama, prior$paramb))
```
ashiklom: Explicit returns generally make for more readable code, even if not required by R.

modules/meta.analysis/R/single.MA.R
```diff
@@ -18,106 +19,107 @@
...
-        m <- model.parms[[x]]
-        for (i in 1:m) {
-          if(i == 1 && x == 'site') {
-            vars <- c(vars, 'beta.site[1]')
-          }
-          if (i > 1) {
-            vars <- c(vars, paste('beta.', x, '[', i, ']', sep=''))
-          }
+        if (i > 1) {
+          vars <- c(vars, paste0("beta.", x, "[", i, "]"))
```
ashiklom: `sprintf` version would probably be much more readable.

modules/meta.analysis/R/single.MA.R
```diff
@@ -18,106 +19,107 @@
...
+                           thin = max(c(2, j.iter / (5000 * 2))))
+  
+  ## I would have done a while loop, but it could take forever 
+  ## So just give one chance to try again
+  if (gelman.diag(jags.out)$mpsrf > 1.2) {
+    logger.warn("model did not converge; re-running with j.iter * 10")
+    jags.out <- coda.samples(model = j.model, 
+                             variable.names = vars,
+                             n.iter = j.iter * 10, 
+                             thin = max(c(2, j.iter / (500 * 2))))
```
ashiklom: Some of the more recent Bayesian stuff I've read advises against thinning in favor of calculating an effective sample size. Should we change this accordingly...?

modules/meta.analysis/R/single.MA.R
rykelly: Dunno I agree with that. `paste` is so very R.

ashiklom:
@bpbond I just opened a PR that contains my revisions, in which I address basically all of my comments. Once you approve and merge that, I will approve this PR.

ashiklom:
@tonygardella @mdietze @robkooper I think this is good to go!

modules/meta.analysis/R/meta.analysis.write.model.R
mdietze: Some of these are prior parameters that could be passed as data, but others turn whole blocks of code on and off -- those need to stay as is.

modules/meta.analysis/R/meta.analysis.summary.R
dlebauer: Yep I think I brought this up last time @istfer and you were working on automated convergence diagnostics

modules/meta.analysis/R/meta.analysis.write.model.R
dlebauer: @ashiklom It's been at least five years since I wrote that ... if it ain't broke ... but if I were going to substantially refactor this code I'd take a hard look at other packages that have been released since then and use more conventional modeling syntax.

modules/meta.analysis/R/run.meta.analysis.R
dlebauer: @ashiklom these aren't any more magic than the log messages that they control - since there is no effect on the results I wouldn't worry too much; also, I don't think having a global P value for triggering debug messages is necessary (or even useful - since it will depend a lot on context).

modules/meta.analysis/R/run.meta.analysis.R
dlebauer: @ashiklom this is not only magical, it is a crude hack - if the means are very small the priors are too strong. Not sure what the 'best' approach would be, but rescaling the data (and unscaling the results) is one option.

PR #1079, opened 2016-10-12
make sure there are no spaces at beginning/end of params

robkooper:
make sure there are no spaces around the params which can trigger an error in insert workflow.

https://gitter.im/PecanProject/pecan?at=57fce8724fde72031429acfa


ashiklom:
Absolutely unacceptable...

PR #1081, opened 2016-10-13
Style cleanup for `modules/emulator/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure all if blocks have braces. Tried to make sure formatR didn’t make things too jaggy. Substituted `seq_len`, `seq_along`, and `paste0`, and `file.path` as necessary.


modules/emulator/R/calcSpatialCov.list.R
```diff
@@ -11,25 +11,18 @@
...
-  #    H[i,j] <- tau*exp(tmp)
-  #    H[j,i] <- H[i,j]
-  #  }
-  #}
-  for(k in 1:m) H <- H - psi[k]*d[[k]]
-  H <- tau*exp(H)
-  return(H)
-}
-
+  # for(i in 1:nl){ for(j in i:nl){ tmp <- 0 for(k in 1:m){tmp <- tmp -
```
istfer: some of the commented-out code lines are clumped together. Although this function is (and some others in this module are) not currently being used, it might not be bad to keep them tidy, @mdietze what do you think? should I point them out?

modules/emulator/R/GaussProcess.R
```diff
@@ -165,119 +185,131 @@ function(x,y,isotropic=TRUE,nugget=TRUE,method="bayes",ngibbs=5000,burnin=1000,t
...
+  }
+  if (haveTime) {
+    progressBar(1.1, prevTime)
   }
-  if(haveTime) progressBar(1.1,prevTime);
   
-  return(list(method=method,tauwjump=tauwjump,tauw = tauwgibbs,psijump=psijump,psi=psigibbs,mu=mugibbs,tauv=tauvgibbs,W=Wgibbs,nugget=nugget,isotropic=isotropic,d=d,samp=samp,x.id=x.id,x.compact=x.compact,y=y,zeroMean=zeroMean))
-}
-
+  list(method = method, tauwjump = tauwjump, tauw = tauwgibbs,
```
ashiklom: :+1: to leaving explicit returns. I also think they are much clearer.

modules/emulator/R/calcSpatialCov.list.R
ashiklom: :+1: to leaving commented code blocks formatted as they were originally.

I think a good workflow with `formatR` is to run it and then stage code interactively (`git add --patch`) before committing. That way, reasonable changes can be committed and weird artifacts like this can be easily undone after the commit with `git checkout --` or `git reset --hard`.

bpbond: Yeah, this is formatR and it's irritating. I try to revert these but missed this one.

modules/emulator/R/GaussProcess.R
bpbond: @istfer Reason - just that I feel it's more "R style". But see my other comment; happy to leave them in if that's desired.

modules/emulator/R/calcSpatialCov.list.R
istfer: Yeah, I also like to clean those up usually. But there are cases when leaving commented-out code might be useful. For example, I'm not the original author in many of these functions. I probably wouldn't think of checking out the version control if I don't know another implementation or an in-between step existed at all. It's sometimes useful for catching up with the thought process of the others

rykelly: I always think of that too, @istfer, especially in collaborative code. It's a tradeoff I suppose...

modules/emulator/R/GaussProcess.R
mdietze: Not going to ask @bpbond  to undo any previous changes to any modules, but moving forward I side with my team's clear preference for explicit returns. @tonygardella could you put that in our style guide. At the same time, could you put a lot of the fixes @bpbond has been implementing so previous practices don't creep in again (library, seq_along, seq_len, file.path, paste0, lubridate, UTC, etc)

bpbond: Sounds good. When all these PRs are merged, I may go back through and take care of explicit returns.

PR #1085, opened 2016-10-15
Style cleanup for `modules/rtm/R/*.R` scripts 

bpbond:


bpbond:
Not sure what is going on here. The travis log shows the `PEcAn.utils` installation failing (lines 4300-) because ggplot2 is required but not available. I don't think any of my `rtm` changes could have borked this.

mdietze:
@bpbond could you try a merge and conflict resolve on this PR?  I'd like to see if the changes you and @ashiklom have made to Travis, make, and package dependencies might have resolved the quirky errors this build was getting

bpbond:
@mdietze will do. Some gnarly conflicts, but should figure things out in a bit.

bpbond:
Checks passed!  🎉

mdietze:
quickly merged before something else caused another conflict...

Issue #759, opened 2016-03-21
Prior Elicitation tool


mdietze:
It would be nice to have a web-based tool in PEcAn that would help users through the process of eliciting their own priors (or the priors of other experts). This tool should follow normal social science procedures to avoid known human cognitive biases (e.g. eliciting the min and max first, and then more extreme quantiles, and then working inward to the mean/median, to avoid anchoring and adjustment). Other elicitiation software exists and should be checked out to see if it can be leveraged.  Graphical feedback should be given, preferably in an interactive way. Once points on the curve have been elicited, we need to also guide users through selecting the named distribution & parameters that best matches their prior (i.e. do the fits in the background, display the options).

Once the basic tool is functional, it would be good to wrap this in code that would also set up the database record. Key items that need to be discerned is whether a prior is for a general trait or a model-specific parameter, and whether the prior is being constructed for a specific PFT or for use across PFTs. These choices would help us determine whether there is any data we can query in the database to help check the elicitation against, or whether this would be double dipping.

mdietze:
Just got the following email from a list-serve that seems very relevant to this issue:

For anyone interested in expert elicitation of (computer model input) probability distributions, version 3.0 of the SHeffield ELicitation Framework (SHELF) is announced - see 

http://tonyohagan.co.uk/shelf

This is a major upgrade to version 2.0 (released 6 years ago!), featuring:
Additional extensive advice for organising and conducting SHELF elicitation workshops, in the form of several new documents.
Updated SHELF 1 and SHELF 2 templates, with the six versions of SHELF 2 now consolidated into a single template.
Two new SHELF 3 templates for eliciting distributions for a group of uncertain quantities.  This is the first time that SHELF has included templates for multivariate elicitation.
Several PowerPoint slide sets to assist with explaining the different kinds of judgements that experts are required to make, and to help them to make those judgements accurately.
A new R package (https://cran.r-project.org/package=SHELF) for fitting distributions to expert judgements.
Examples of an evidence dossier and completed SHELF templates.

PR #1086, opened 2016-10-15
Style cleanup for `modules/priors/R/*.R` scripts 

bpbond:
Ran through formatR::tidy_dir with arrow=TRUE and indent=2; made sure
all if blocks have braces. Tried to make sure formatR didn’t make
things too jaggy. Substituted `seq_len`, `seq_along`, and `paste0`, and
`file.path` as necessary. Made sure all functions end with an explicit `return`.

**Note**: I gave the `if` blocks in `priors.R` failure cases that`stop`.


modules/priors/R/plots.R
```diff
@@ -126,14 +119,8 @@ priorfig <- function(priordata = NA, priordensity = NA,
...
-plot.trait <- function(trait,
-                       prior = NULL,
-                       posterior.sample = NULL,
-                       trait.df = NULL,
-                       fontsize = list(title = 18, axis = 14),
-                       x.lim = NULL,
-                       y.lim = NULL,
-                       logx = FALSE) {
+plot.trait <- function(trait, prior = NULL, posterior.sample = NULL, trait.df = NULL, fontsize = list(title = 18, 
+                                                                                                      axis = 14), x.lim = NULL, y.lim = NULL, logx = FALSE) {
```
mdietze: I think the original formatting was cleaner here

modules/priors/R/priors.R
```diff
@@ -77,72 +86,80 @@ fit.dist <- function(trait.data, trait = colnames(trait.data),
 ##'                 x=c(2, 6, 3.3), 
 ##'                 alpha = 0.05, 
 ##'                 distn = 'lnorm')$optim$bestmem
-#--------------------------------------------------------------------------------------------------#
 prior.fn <- function(parms, x, alpha, distn, central.tendency = NULL, trait = NULL) {
-  if(!distn %in% c('lnorm', 'gamma', 'weibull', 'beta')){
+  if (!distn %in% c("lnorm", "gamma", "weibull", "beta")) {
```
dlebauer: This function could probably be replaced by or rewritten to use another package like fitdistplus or rrRiskDistributions that are way more comprehensive and general. I use those packages now when fitting priors so ... maybe this can be deprecated

modules/priors/R/priors.R
```diff
@@ -77,72 +86,80 @@ fit.dist <- function(trait.data, trait = colnames(trait.data),
...
       ct <- qbeta(0.5, a, b)  ## median
-    } else if (central.tendency == 'mode') {
-      ct <- ifelse(a>1 & b>1,(a-1)/(a+b-2) , 0) ## mode
-    } 
-  } 
+    } else if (central.tendency == "mode") {
+      ct <- ifelse(a > 1 & b > 1, (a - 1) / (a + b - 2), 0)  ## mode
+    }
+  }
   return(sum(abs(c(lcl, ucl, ct) - x)))
```
dlebauer: Although it is beautiful in its own way

modules/priors/R/priors.R
```diff
@@ -17,43 +16,53 @@
 ##' @return best fit distribution
 ##' @export
 ##' @author David LeBauer
-#--------------------------------------------------------------------------------------------------#
 fit.dist <- function(trait.data, trait = colnames(trait.data), 
-                     dists = c('weibull', 'lognormal', 'gamma'), n = NULL) {
-  if(class(trait.data) == 'data.frame') trait.data <- trait.data[,1]
+                     dists = c("weibull", "lognormal", "gamma"), n = NULL) {
```
dlebauer: Could be deprecated in favor of newer packages

modules/priors/R/priors.R
dlebauer: Can revisit when implementing #759

PR #1087, opened 2016-10-15
Style cleanup for `modules/photosynthesis/R/*.R` scripts 

bpbond:


modules/photosynthesis/R/fitA.R
```diff
@@ -4,77 +4,83 @@
...
-  Xa = t(t(Xa)-acenter)
-}
-
-
-## Define JAGS model
-
-my.model = "  
+  ## TO-DO: Random effects using design matrix Model selection output variable selection:
+  ## Pred Loss, WAIC?  function to do: multiple response curves specify priors in model
+  ## object integrate with meta-analysis
```
mdietze: To-do flattened by tidy

modules/photosynthesis/R/fitA.R
```diff
@@ -128,7 +134,7 @@ To <- 35    ## Representative value, would benifit from spp calibration!
 #RLEAF.A   Aleaf[i]~dnorm(0,tau.Aleaf)
 #RLEAF.A  }
 
-  for(i in 1:n){
+  for(i in seq_len(n)) {
```
mdietze: This is in JAGS, needs to revert to 1:n

modules/photosynthesis/R/fitA.R
```diff
@@ -155,95 +161,104 @@ To <- 35    ## Representative value, would benifit from spp calibration!
...
+              pi = dat$Ci[sel], 
+              q = dat$PARi[sel],
+              T = dat$Tleaf, 
+              n = length(sel), Kc = 46, 
+              Ko = 22000, 
+              po = 21000, 
+              rep = curve.id, 
+              nrep = n.curves)
+# Kc<-46 ## Michaelis constant CO2 (Pa) Ko<-33000 ## Michaelis constant O2 (Pa)
+# po<-21000 ## partial pressure of O2 (Pa)
```
mdietze: Flattened

PR #1090, opened 2016-10-17
Style cleanup for `utils/R/*.R` scripts 

bpbond:

**Non-style notes:**
* `convert.input`: changed GMT to UTC; added if-block failure case
* `sendmail`: note new line 32 (`sendmail <- Sys.which("sendmail")`) -
OK?
* `bugs.rdist`: added if-block failure
* added if-block failure to `misc.convert`

### DIFF
--- a/utils/R/convert.input.R
+++ b/utils/R/convert.input.R
@@ -5,228 +5,245 @@
 ##' @title convert.input
 ##' @export
 ##' @author Betsy Cowdery, Michael Dietze, Ankur Desai
-convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, start_date, end_date, 
-                          pkg, fcn, con=con, host, browndog, write=TRUE,  
-                          format.vars, overwrite=FALSE, ...) {
+convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, start_date, 
+                          end_date, pkg, fcn, con = con, host, browndog, write = TRUE, 
+                          format.vars, overwrite = FALSE, ...) {
   input.args <- list(...)
-
-  logger.debug(paste(
-    "Convert.Inputs", fcn, input.id, host$name, outfolder, formatname, mimetype, 
-    site.id, start_date, end_date))
-
+  
+  logger.debug(paste("Convert.Inputs", fcn, input.id, host$name, outfolder, formatname, 
+                     mimetype, site.id, start_date, end_date))
+  
   n <- nchar(outfolder)
-  if(substr(outfolder,n,n) != "/") {
-    outfolder = paste0(outfolder,"/")
+  if (substr(outfolder, n, n) != "/") {
+    outfolder <- paste0(outfolder, "/")
   }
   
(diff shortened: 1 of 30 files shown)


utils/R/convert.input.R
```diff
@@ -5,228 +5,245 @@
...
+          "ed.met_driver_header files format") {
+        outputtype <- "ed.zip"
+      } else if (formatname == "Sipnet.climna") {
+        outputtype <- "clim"
+      } else if (formatname == "DALEC meteorology") {
+        outputtype <- "dalec.dat"
+      } else if (formatname == "LINKAGES met") {
+        outputtype <- "linkages.dat"
+      } else {
+        logger.severe(paste("Unknown formatname", formatname))
```
mdietze: Just a reminder to @tonygardella that we've got to slowly work these model-specific hacks out of the code.

utils/R/convert.input.R
```diff
@@ -5,228 +5,245 @@
...
+    curloptions <- c(curloptions, followlocation = TRUE)
     
-    # check if we can do conversion 
-    out.html <- getURL(paste0(
-      "http://dap-dev.ncsa.illinois.edu:8184/inputs/", browndog$inputtype), .opts = curloptions)
-    if(outputtype %in% unlist(strsplit(out.html, '\n'))){
-      logger.info(paste("Conversion from", browndog$inputtype, "to", outputtype, "through Brown Dog"))
+    # check if we can do conversion
+    out.html <- getURL(paste0("http://dap-dev.ncsa.illinois.edu:8184/inputs/", 
+                              browndog$inputtype), .opts = curloptions)
```
mdietze: @robkooper should the URL be hard coded or a variable?  Seems like we shouldn't be hard coded to hit the DAP dev machine.

utils/R/convert.input.R
```diff
@@ -5,228 +5,245 @@
...
     # check for and create output folder
-    if(!file.exists(outfolder)){
-      dir.create(outfolder, showWarnings=FALSE, recursive=TRUE)
+    if (!file.exists(outfolder)) {
+      dir.create(outfolder, showWarnings = FALSE, recursive = TRUE)
     }
     
     # post zipped file to Brown Dog
-    html <- postForm(url,"fileData" = fileUpload(zipfile), .opts = curloptions)
+    html <- postForm(url, fileData = fileUpload(zipfile), .opts = curloptions)
```
mdietze: @robkooper was there a reason fileData was quoted here that's BrownDog specific or is Ben's fix correct?

utils/R/remote.R
```diff
@@ -8,10 +8,9 @@
 #-------------------------------------------------------------------------------
 
 #-------------------------------------------------------------------------------
-# Remote utilities. Allows for the following functionality
-# - execute system call on remote machine
-# - execute R call on remote machine, returns result
-# - copy files to/from remote machines
+# Remote utilities. Allows for the following functionality - execute system call
+# on remote machine - execute R call on remote machine, returns result - copy
+# files to/from remote machines
```
mdietze: tidy flattened

Issue #1088, opened 2016-10-16
`Imports` versus `Depends` in DESCRIPTION files


bpbond:
The PEcAn packages all seem to use `Depends:` exclusively in their DESCRIPTION files, and none uses `Imports:`; furthermore, there are many `library` calls throughout the packages' code. Finally, none of the `Depends:` specifies any version number restrictions. All this runs against CRAN suggested practice, and contributes to vulnerability w.r.t. namespace clashes.

From http://stackoverflow.com/questions/8637993/better-explanation-of-when-to-use-imports-depends :

>"Imports" is safer than "Depends" (and also makes a package using it a 'better citizen' with respect to other packages that do use "Depends").

From https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-Dependencies :

>Packages whose namespace only is needed to load the package using library(pkgname) should be listed in the ‘Imports’ field and not in the ‘Depends’ field. Packages listed in imports or importFrom directives in the NAMESPACE file should almost always be in ‘Imports’ and not ‘Depends’.

and 

>Field ‘Depends’ should nowadays be used rarely, only for packages which are intended to be put on the search path to make their facilities available to the end user (and not to the package itself): for example it makes sense that a user of package latticeExtra would want the functions of package lattice made available.

and

>R code in the package should call library or require only exceptionally. Such calls are never needed for packages listed in ‘Depends’ as they will already be on the search path. It used to be common practice to use require calls for packages listed in ‘Suggests’ in functions which used their functionality, but nowadays it is better to access such functionality via :: calls.

bpbond:
This has been superseded by #1116 - closing.

PR #1094, opened 2016-10-17
Test: have `install.packages` handle a specific package dependency installation (abind)

bpbond:
## Description
Test: have `install.packages` handle a specific package dependency installation (abind)

 A test. Removed `abind` from package list in `scripts/install.dependencies`; made sure it was listed in package DESCRIPTION files; added `abind::` namespace to `abind` calls; modified
`build.sh`.

## Motivation and Context
There shouldn't be any need for `install.dependencies` to do most of what it does.

## Types of changes

## Checklist:

### DIFF
--- a/models/ed/DESCRIPTION
+++ b/models/ed/DESCRIPTION
@@ -14,10 +14,11 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
     link the Ecosystem Demography Model, version 2, to PEcAn.
 Depends:
     PEcAn.utils,
-    abind,
     coda,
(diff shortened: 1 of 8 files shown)


bpbond:
Nice! OK, as noted above, this test PR removes the responsibility of installing `abind` (a dependency of `ed`, `data.atmosphere`, and `utils`) from `install.dependencies`--instead, we let R's `install.packages` handle it. The `abind` dependency is now listed in the packages' `Imports:` field, with a specific minimum version number; this will also reduce potential namespace collision problems (see #1088 ).

Issue #1083, opened 2016-10-14
Remove / deprecate use of now() from SQL insert statements


dlebauer:
From Scott in https://github.com/terraref/computing-pipeline/issues/174#issuecomment-253676605

>  Please don't include the created_at and/or updated_at columns in the insert and update statements! A trigger function handles these automatically so that created_at and updated_at are set to the current time when a row is inserted, and updated_at is reset to the current time when a row is updated. Setting them explicitly overrides this behavior, and using now() sets the column to machine time, not UTC time. (There is a function I defined, utc_now(), which will return UTC time, but you don't need to use it in this case.)

Technically this is a bug, since `now()` provides local instead of UTC datetime values. However, I am not of any errors this will cause.

This is low priority and a comprehensive update can wait until we transition to dplyr (#395).

PR #1095, opened 2016-10-17
The draft for read.settings compatibility with benchmarking

bcow:
## Major changes:
- Added benchmarking stage to `workflow.R`
- Reworking the process of reading settings:
   - `read.settings` reads the `pecan.xml` file
   - add additional settings such a benchmarking settings
   - update settings and write `pecan.CHECKED.xml`
   - This new configuration works with normal `pecan.xml` files - You can test by doing runs with [my pecan](https://test-pecan.bu.edu/pecan_ecowdery/)
- New improved `benchmarking.workflow.R` that can be used when not using `workflow.R`

## Example XML files 
Can be used to test benchmarking (both with `workflow.R` and `benchmark.workflow.R`
(I'll check them off when they are working)

### 1 variable, 1 metric, 1 site, 1 model

 - settings.file: `modules/benchmark/inst/scripts/bm.1var.1metric.1site.1model.xml`
 - [x] New Run 
 - [x] Existing Run 

### 2 variables, 2 metric, 1 site, 1 model

 - settings.file: `modules/benchmark/inst/scripts/bm.2var.2metric.1site.1model.xml`
 - [x] New Run 
 - [x] Existing Run 

### 2 variables, 2 metric, 1 site, 1 model

 - settings.file: `modules/benchmark/inst/scripts/bm.2var.2metric.2site.1model.xml`
 - settings will now be a multisite object
 - [ ] New Run 
 - [ ] Existing Run


.gitignore
```diff
@@ -69,5 +69,6 @@ tests/BC*
 compile_on_geo.sh
 documentation/tutorials/*/*.html
 pecan.Rproj
+shiny/BenchmarkReport/*
```
mdietze: shiny/BenchmarkReport exists in the repo already, so a commit that adds it to the ignore seems like a bad idea. Could you remove that. Better yet, if you've made changes to BenchmarkReport could you commit those changes too?

modules/benchmark/R/create.BRR.R
```diff
@@ -9,48 +9,49 @@
 ##' 
 ##' @author Betsy Cowdery 
 
-create.BRR <- function(ensemble.id, workflow, con) {
+create.BRR <- function(ens_wf, con){
```
mdietze: changed argument to something hard to understand, but didn't update the documentation (param still listed as ensemble.id)

modules/benchmark/R/create.BRR.R
```diff
@@ -9,48 +9,49 @@
...
   
-  cnd1 <- workflow$hostname == fqdn()
-  cnd2 <- workflow$hostname == "test-pecan.bu.edu" & fqdn() == "pecan2.bu.edu"
-  cnd3 <- workflow$hostname == "pecan2.bu.edu" & fqdn() == "test-pecan.bu.edu"
-
-  if (cnd1 | cnd2 | cnd3) {
-    # If the ensemble run was done on localhost, turn into a BRR
+  cnd1 <- ens_wf$hostname == fqdn() 
+  cnd2 <- ens_wf$hostname == 'test-pecan.bu.edu' & fqdn() == 'pecan2.bu.edu'
+  cnd3 <- ens_wf$hostname == 'pecan2.bu.edu' & fqdn() == 'test-pecan.bu.edu'
```
mdietze: would be nice to figure out how to avoid having hacks like this hardcoded (Fix not required in this PR)

modules/benchmark/R/create.BRR.R
```diff
@@ -9,48 +9,49 @@
...
-    logger.error(sprintf("Cannot create a benchmark reference run for a run on hostname: %s", 
-      workflow$hostname))
-  }
-} # create.BRR
-
-
-# Case in which we would need a remote connection to get the pecan.xml file - not functional
+    settings_xml <- toString(listToXml(clean, "pecan"))
+    
+    ref_run <- db.query(paste0(" SELECT * from reference_runs where settings = '", settings_xml,"'"),con)
```
mdietze: This is OK for now, but in the long run it's a pretty weak test -- an settings that is LOGICALLY the same but not character-for-character identical (e.g. whitespace, order of tags) wouldn't come up.

modules/benchmark/R/create.BRR.R
```diff
@@ -9,48 +9,49 @@
...
+    
+    if(length(ref_run) == 0){ # Make new reference run entry
+      ref_run <- db.query(paste0("INSERT INTO reference_runs",
+                                 "(model_id, settings, user_id, created_at, updated_at)",
+                                 "VALUES(",ens_wf$model_id,", '",settings_xml,"' , ",user_id,
+                                 ", NOW() , NOW()) RETURNING *;"),con)
```
mdietze: Per #1083 we're supposed to remove created_at and updated_at NOW(),NOW() from SQL. The database inserts these automatically and @dlebauer noted it is a bug since most machines insert NOW() in local time, not UTC.  @tonygardella make sure this goes into the style guide.

modules/benchmark/R/create.BRR.R
```diff
@@ -9,48 +9,49 @@
...
+    }else if(dim(ref_run)[1] > 1){# There shouldn't be more than one reference run with the same settings
+      logger.error("There is more than one reference run in the database with these settings. Review for duplicates. ")
+    }
+    
+    bm.ensemble <- db.query(paste0("INSERT INTO benchmarks_ensembles",
+                                   "(reference_run_id, ensemble_id, model_id, ",
+                                   "user_id, created_at, updated_at, citation_id)",
+                                   "VALUES(",ref_run$id,",",ens_wf$ensemble_id,",", ref_run$model_id,", ",user_id,
+                                   ", NOW() , NOW(), 1000000001 ) RETURNING *;"),con)
```
mdietze: 1) would be good to also check for duplicates here
2) as above, drop NOW() 
3) why are you inserting citation_id 1000000001 as a hard-coded value rather than leaving this blank?  Citation_id is an optional value, and I suspect that it will often be added when the results are published, not when they're created. Maybe make it an optional argument to the function, and one that defaults to blank or NULL?

modules/benchmark/R/load.data.R
```diff
@@ -14,6 +14,8 @@ load.data <- function(data.path, format, start_year = NA, end_year = NA, site =
   
   library(PEcAn.utils)
   library(PEcAn.benchmark)
+  library(lubridate)
+  library(udunits2)
   library(dplyr)
```
mdietze: Per changes to style guide, package references should be explicit in the code (e.g. pkg::fcn), which means we don't need these library calls.  Specifically I believe we've successfully removed these three CRAN packages from everywhere else in PEcAn -- would be good to go ahead and clean up this code too.

modules/benchmark/R/read.settings.RR.R
```diff
@@ -0,0 +1,52 @@
...
+##' @title Read settings from database using reference run id
+##' @param settings settings or multisettings object
+##' @export 
+##' 
+##' @author Betsy Cowdery 
+
+
+read.settings.RR <- function(settings){
+  library(PEcAn.DB)
+  library(dplyr)
```
mdietze: as above, use namespaces in functions rather than library

modules/benchmark/R/read.settings.RR.R
```diff
@@ -0,0 +1,52 @@
...
+  
+  bety <- src_postgres(dbname   = settings$database$bety$dbname,
+                       host     = settings$database$bety$host,
+                       user     = settings$database$bety$user,
+                       password = settings$database$bety$password)
+  
+  settings <- tbl(bety,"reference_runs") %>% 
+    filter(id == settings$benchmark$reference_run_id) %>% 
+      dplyr::select(settings) %>% collect() %>% unlist() %>%
+      xmlToList(.,"pecan") %>% append(settings,.) %>% Settings()
```
mdietze: Impressive that you can do the query and insert all in one line. Probably would have taken me a page of code.

settings/R/check.all.settings.R
```diff
@@ -0,0 +1,866 @@
...
+##-------------------------------------------------------------------------------
+##' check to see if inputs are specified - this should be part of the model code
+##' @title Check Inputs
+##' @param settings settings file
+##' @export check.inputs
+check.inputs <- function(settings) {
+  library(XML)
+  library(lubridate)
+  library(PEcAn.DB)
+  library(PEcAn.utils)
```
mdietze: packages

settings/R/fix.deprecated.settings.R
```diff
@@ -0,0 +1,53 @@
...
+##' @title Fix Deprecated Settings
+##' @param settings settings list
+##' @return updated settings list
+##' @author Ryan Kelly
+##' @export fix.deprecated.settings
+fix.deprecated.settings <- function(settings) {
+  library(XML)
+  library(lubridate)
+  library(PEcAn.DB)
+  library(PEcAn.utils)
```
mdietze: packages

settings/R/write.settings.R
```diff
@@ -0,0 +1,34 @@
...
+##' @param settings settings list
+##' @return updated settings list
+##' @author Ryan Kelly
+##' @export write.settings
+
+
+write.settings <- function(settings, outputfile = "pecan.CHECKED.xml"){
+  library(XML)
+  library(PEcAn.DB)
+  library(PEcAn.utils)
```
mdietze: 1) Can you think of a more appropriate name for this function? write.settings sounds like a function that reads in a settings object and writes it to file. This does a lot more than that!  Unfortunately clean.settings and update.settings are already in use. 
2) packages

web/workflow.R
```diff
@@ -39,6 +39,15 @@ if (is.na(args[1])){
...
 
+# Check for additional modules that will require adding settings
+if("benchmark" %in% names(settings)){
+  library(PEcAn.benchmark)
+  settings <- papply(settings, read.settings.RR)
+}
+
+# Write pecan.CHECKED.xml
+settings <- write.settings(settings, outputfile = "pecan.CHECKED.xml")
+
```
mdietze: I don't see the call to calc.benchmarks at the end of workflow.R

modules/benchmark/R/create.BRR.R
bcow: You commented on this in the last PR but I still don't have any ideas on what to do with it.
It's just a special case that involves our pecan servers. Could simplify it to:
```
ens_wf$hostname == fqdn()  | length(grep("pecan.bu.edu", c(fqdn(), "test-pecan.bu.edu")))==2
```

modules/benchmark/R/create.BRR.R
bcow: ✔️ Removed all created_at, updated_at, NOW(),NOW() in benchmarking

settings/R/write.settings.R
bcow: settings2pecan.CHECKED?
Or should we not call the other function update.settings?

web/workflow.R
bcow: ✔️ Added it!

modules/benchmark/R/create.BRR.R
bcow: 1) ✔️ Added check for duplicates - and actually moved that part to `create.benchmark`
2) ✔️ Done
3) I have no idea why I was doing that. For now I'm just removing it - but I will make a note to add it to arguments in the future.

modules/benchmark/R/create.BRR.R
robkooper: Would it be possible to add some text as a comment on the reason why we can not run a benchmark reference run on any of those other machines, what conditions do we need to be added to the list of exceptions.

modules/benchmark/R/create.BRR.R
bcow: I take it back - citation can't be NULL 
```
null value in column "citation_id" violates not-null constraint
```
What should I do then?

modules/benchmark/R/create.BRR.R
bcow: Here I'm just trying to find a way to say that test-pecan, and pecan2 are actually the "same" machine in that the files are accessible from either. 
But this is skirting around the issue that if one wants to turn an existing ensemble into a benchmark reference run and that ensemble was not run on localhost, getting all the information to populate the settings section of the reference run table id going to be difficult / may not be possible. This is because we may not be able to read the pecan.xml file on the remote host and the ensemble settings aren't saved anywhere else. This is something that @mdietze and I talked about earlier this week and maybe Mike can explain it better than I can.

robkooper: What if you look at the outdir and see if that exists on the local machine, if so you can (I think) assume that all the files are locally.

settings/R/write.settings.R
mdietze: OK for today, but renaming the update.settings function so you can name this function update is definitely an option.

PR #1096, opened 2016-10-17
Make build

ashiklom:
Make-based installation for PEcAn. The first draft very likely will not work, but I thought I'd push it to see exactly how and why it fails.

Also, the Makefile can _definitely_ be designed more efficiently with variables and patterns, but I figured I could get this hard-coded version working first before trying the other stuff.


modules/data.hydrology/DESCRIPTION
```diff
@@ -11,8 +11,7 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
     model parameterization, execution, and analysis. The goal of PECAn is to
     streamline the interaction between data and models, and to improve the
     efficacy of scientific investigation.
-Depends: RMySQL
-Suggests: testthat
+Suggests: testthat, RMySQL
```
robkooper: i think we should just kill RMySQL and say we only depend on postgresql.

Makefile
```diff
@@ -0,0 +1,211 @@
...
+
+all: ${BASE_PACKAGES} ${MODELS} ${MODULES}
+
+#### DEPENDENCY MAPPING ####
+
+${MODELS}: utils data.atmosphere
+
+#### BASE PACKAGES ####
+
+utils: .install.utils
```
robkooper: curious why do you have a utils that depends on .install.utils, etc.
would it make sense to add a dependency here that is all the R packages needed for that module and make sure they are installed. We should be able to pull them from the Dependencies file.

modules/data.hydrology/DESCRIPTION
robkooper: i think we should just kill RMySQL and say we only depend on postgresql.

Makefile
ashiklom: Because an R package installation has no explicit target. This is a `make` hack whereby, after a successful install of an R package, I make a small text file (`.install.*`) that indicates when the package was installed.

Dependencies are installed automatically by `devtools::install()`, so we shouldn't need to list them explicitly anywhere except each package's `DESCRIPTION`. At least that's the theory -- I'm testing it out now.

bpbond: Re this last point by @ashiklom , see #1088 and #1094

.travis.yml
```diff
@@ -44,34 +38,25 @@ addons:
...
       - secure: "gL81TwDOcK/32Hxxl2BcY7pioyrtyV4y1f+D/vGEpQz8mYL+M+55tUkhHJF53779XSUdQdML/gsr8JZVtApJdLIBysFU67GVYXm1s7x/b8J61CkMfDgsmposEWK4NFYHfeIRj32ioeajrQ+RKi1I6chjzYT7gLyl70gtPelRJ4s="
     on_success: always
     on_failure: always
     on_start: always
-  email:
-    on_success: always
-    on_failure: always
+  #email:
+    #on_success: always
+    #on_failure: always
```
mdietze: is there a reason you turned email notification off?

.travis.yml
```diff
@@ -44,34 +38,25 @@ addons:
...
 
-# cache local R libraries directory:
-cache:
-  - directories:
-    - $HOME/R/Library
-  - packages
```
mdietze: Won't the build take a lot longer if you don't cache R packages?

Makefile
```diff
@@ -0,0 +1,229 @@
...
+
+
+MODULES := allometry assim.batch assim.sequential benchmark \
+	       data.atmosphere data.hydrology data.land \
+	       data.mining data.remote emulator meta.analysis \
+	       photosynthesis priors rtm uncertainty
+
+.PHONY: all ${BASE_PACKAGES} ${MODELS} ${MODULES}
+
+all: .install PEcAn.all ${BASE_PACKAGES} ${MODELS} ${MODULES}
```
mdietze: not required for this PR, but it would be nice to have a 'make clean' option

.travis.yml
ashiklom: Packages are cached. I just moved that block to near the top of the file.

Makefile
ashiklom: What would that entail? Uninstalling the packages? Clearing out the `.install` directory?

mdietze: If the .install is the hack that allows make to know when packages are updated, then that's probably all than would need to be removed. I don't think you have to actually uninstall the packages.

Makefile
```diff
@@ -0,0 +1,232 @@
...
+#### DEPENDENCY MAPPING ####
+
+PEcAn.all: ${BASE_PACKAGES} ${MODULES} ${MODELS}
+
+${MODELS}: ${BASE_PACKAGES} ${MODULES}
+
+${BASE_PACKAGES} ${MODELS} ${MODULES}: .install/devtools
+
+.install/devtools:
+	Rscript -e "req <- require('devtools'); if(!req) install.packages('devtools')"
```
robkooper: this failed in a headless environment. I needed to add repos='http://cran.rstudio.com/'

Makefile
```diff
@@ -0,0 +1,232 @@
...
+
+
+MODULES := allometry assim.batch assim.sequential benchmark \
+	       data.atmosphere data.hydrology data.land \
+	       data.mining data.remote emulator meta.analysis \
+	       photosynthesis priors rtm uncertainty
+
+.PHONY: all ${BASE_PACKAGES} ${MODELS} ${MODULES}
+
+all: .install PEcAn.all ${BASE_PACKAGES} ${MODELS} ${MODULES}
```
robkooper: An option to do a make check which would run the devtools::check for all, same for documentation

modules/data.hydrology/DESCRIPTION
```diff
@@ -11,9 +11,9 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
     model parameterization, execution, and analysis. The goal of PECAn is to
     streamline the interaction between data and models, and to improve the
     efficacy of scientific investigation.
-Depends: RMySQL
-Suggests:
-    testthat (>= 1.0.2)
+Suggests: 
+    testthat (>= 1.0.2), 
+    RMySQL
```
robkooper: get rid of the RMySQL not sure it even works.

scripts/build.sh
```diff
@@ -16,7 +16,6 @@ export _R_CHECK_FORCE_SUGGESTS_="FALSE"
 
 # these variables are set using the command line arguments below
```
robkooper: just delete the build.sh and dependencies.R files.

Makefile
robkooper: Another option that can be a target is tests which would do the integration tests we have. You should be able to copy the whole block of code from build.sh. This way we just end up with a single file we need to maintain, and as a bonus we might be able to run multiple tests in parallel. Just need to make sure each test writes to its own folder.

robkooper:
looks good. tested it on a vm to see if I can use it to build the vm with this and it was easy.

Makefile
mdietze: @robkooper I think that's @ashiklom's ultimate plan, but I told him to start with putting the packages in make and call it from build.sh, and then to transition more and more into make until build.sh is deprecated. I wouldn't hold up this PR on this specific point.

Makefile
```diff
@@ -0,0 +1,245 @@
...
+
+clean:
+	rm -rf .install
+
+check: 
+	Rscript -e "library(devtools); \
```
robkooper: Maybe use the following so we have one place in the makefile with all modules

```
	for x in ${BASE_PACKAGES}; do \
		Rscript -e "devtools::check('$$x')"; \
	done
	for x in ${MODELS}; do \
		Rscript -e "devtools::check('models/$$x')"; \
	done
	for x in ${MODULES}; do \
		Rscript -e "devtools::check('modules/$$x')"; \
	done

bpbond:
Nice work @ashiklom

PR #1100, opened 2016-10-17
Move responsibility for handling `coda` dependency

bpbond:
## Description
Removed coda from package list in scripts/install.dependencies; made sure it was listed in package DESCRIPTION files; removed `library` calls; added `coda::` namespace to function calls. See #1088.

## Motivation and Context

## Types of changes

## Checklist:


utils/DESCRIPTION
```diff
@@ -21,7 +21,9 @@ Depends:
     RCurl
 Imports:
     abind (>= 1.4.5),
+    coda (>= 0.18),
     plyr (>= 1.8.4)
+>>>>>>> PecanProject/master
```
mdietze: conflict

PR #1102, opened 2016-10-17
Move responsibility for handling `XML` dependency

bpbond:
Removed responsibility for handling `XML` dependency  from `install.dependencies`.

## Description
Removed XML from package list in scripts/install.dependencies; made sure it was listed in package DESCRIPTION files; removed `library` calls; added `XML::` namespace to function calls. See #1088.

## Motivation and Context

## Types of changes

## Checklist:


mdietze:
This one seems to have ended up with a couple small conflicts. Sorry about that -- too many balls in the air across all these PR

PR #1105, opened 2016-10-18
Move responsibility for handling `lubridate` dependency

bpbond:
Removed responsibility for handling lubridate dependency from install.dependencies.

## Description
Removed lubridate from package list in scripts/install.dependencies; made sure it was listed in package DESCRIPTION files; removed `library` calls; added `lubridate::` namespace to function calls. See #1088.

## Motivation and Context

## Types of changes

## Checklist:


mdietze:
@bpbond lubridate has conflicts

PR #1110, opened 2016-10-18
Fates progress and misc bugfixes

mdietze:
In addition to slowly moving the FATES coupling forward, fixed a couple bugs that came up in the cleaning and in the modularization of met.process.  The problem found in the download module that apply to CRUNCEP likely apply to other met products as well and should be checked.


models/fates/R/write.configs.FATES.R
```diff
@@ -20,133 +20,156 @@
...
+   domain.nc <- nc_open(domain.file,write=TRUE)
+   ncvar_put(nc=domain.nc, varid='xc', vals=lon)
+   ncvar_put(nc=domain.nc, varid='yc', vals=lat)
+   ncvar_put(nc=domain.nc, varid='xv', vals=lon+c(-1,1,1,-1)*gridres)
+   ncvar_put(nc=domain.nc, varid='yv', vals=lat+c(-1,-1,1,1)*gridres)
+   ncvar_put(nc=domain.nc, varid='area', vals=(2*gridres*pi/180)^2)   
+   nc_close(domain.nc)
+   
+   ## SURF
+   surf.default <- "/home/carya/FATESinput/lnd/clm2/surfdata_map/surfdata_1x1_brazil_16pfts_simyr2000_c160127.nc"
```
serbinsh: @mdietze why point to a specific file on a specific machine? Can this be another system.file() call and the nc file placed in the R lib? Or is it too large for that?

models/fates/R/write.configs.FATES.R
```diff
@@ -20,133 +20,156 @@
...
+   surf.file    <- file.path(local.rundir,paste0("surfdata_",site_name,"simyr2000.nc"))
+   file.copy(surf.default,surf.file)
+   Sys.chmod(surf.file)
+   surf.nc <- nc_open(surf.file,write=TRUE)
```
serbinsh: @mdietze do we need to be copying and renaming these "defaults" or could we just point to the original location/name and read in the data to parse the variables?  That is do we need to move the files or just open them up and read in the defaults?

models/fates/R/write.configs.FATES.R
```diff
@@ -20,133 +20,156 @@
...
+   ncvar_put(nc=domain.nc, varid='xv', vals=lon+c(-1,1,1,-1)*gridres)
+   ncvar_put(nc=domain.nc, varid='yv', vals=lat+c(-1,-1,1,1)*gridres)
+   ncvar_put(nc=domain.nc, varid='area', vals=(2*gridres*pi/180)^2)   
+   nc_close(domain.nc)
+   
+   ## SURF
+   surf.default <- "/home/carya/FATESinput/lnd/clm2/surfdata_map/surfdata_1x1_brazil_16pfts_simyr2000_c160127.nc"
+   surf.file    <- file.path(local.rundir,paste0("surfdata_",site_name,"_simyr2000.nc"))
+   file.copy(surf.default,surf.file)
+   Sys.chmod(surf.file)
```
serbinsh: @mdietze Did you also mean to set a specific mode for this file? e.g. 755 or 775?

serbinsh: Also is there a need to copy (to the run dir) and rename these default inputs or could we just point to the original location and extract the vars needed?

models/fates/R/write.configs.FATES.R
mdietze: @serbinsh all this is still in development and this was the quickest hack to get started.

That said, system.file is an option, but I think it'd be better to grab this from the reference case so that CLM-FATES users can update that reference case that are more site- or system-specific (this file includes things like soils, etc).

models/fates/R/write.configs.FATES.R
mdietze: The default (0777) was fine. Just needed to be able to write to a file that is by default read only

models/fates/R/write.configs.FATES.R
serbinsh: OK.

PR #1112, opened 2016-10-18
lubridate call fixes

mdietze:
bunch of lubridate functions that now need a lubridate:: prefix.

In general, as we switch packages over we need to (A) be careful not to miss functions and (B) be vigilant that we will get this sorts of errors and will need to test the next release extensively


bpbond:
@mdietze were these generating errors?

mdietze:
Yes, the workflow was stopping when it hit these. Roughly a 1:1 mapping between commits and attempted runs

bpbond:
Huh. Ok.

modules/data.atmosphere/R/met2CF.ALMA.R
```diff
@@ -102,7 +102,7 @@ met2CF.PalEON <- function(in.path, in.prefix, outfolder, start_date, end_date, l
           met[[v]] <- ncvar_get(nc = nc1, varid = v)
         } else {
           tmp      <- ncvar_get(nc = nc1, varid = v)
-          met[[v]] <- abind(met[[v]], tmp)
+          met[[v]] <- abind::abind(met[[v]], tmp)
```
serbinsh: @mdietze just checking, just importing abind from abind function to use here correct?  That is 

"Combine multi-dimensional arrays. This is a generalization of cbind and rbind. Takes a sequence
of vectors, matrices, or arrays and produces a single array of the same or higher dimension."

mdietze: yep. variables have time, lat, lon dimensions so can't just rbind

db/R/dbfiles.R
```diff
@@ -59,10 +59,10 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
   inputid <- NULL
   if (nrow(existing.input) > 0) {
     # Convert dates to Date objects and strip all time zones (DB values are timezone-free)
-    startdate <- force_tz(as_date(startdate), 'GMT')
-    enddate <- force_tz(as_date(enddate), 'GMT')
-    existing.input$start_date <- force_tz(as_date(existing.input$start_date), 'GMT')
-    existing.input$end_date <- force_tz(as_date(existing.input$end_date), 'GMT')
+    startdate <- lubridate::force_tz(lubridate::as_date(startdate), 'GMT')
```
serbinsh: @mdietze Should these be forced to UTC instead of GMT?

ashiklom:
@bpbond `Imports` loads the packages, but doesn't add them to the search path, which is where R looks for packages. It's very confusing. More details [here](http://r-pkgs.had.co.nz/namespace.html#search-path).

PR #1117, opened 2016-10-19
try2 only fix lat/lon now

robkooper:
same as previous pull request however only fixes lat/lon this time.


robkooper:
yeah, one thing at a time to make it easier to accept requests.

Issue #1130, opened 2016-10-20
Bug Report: Example not working in `allometry::AllomAve`


SimonGoring:
## Description

Running the example in the help file for `AllomAve`:

```R
if(TRUE){                                # Note, I changed the flag so it would run(!)
  pfts = list(FAGR = data.frame(spcd=531,acronym='FAGR'))
  allom.stats = AllomAve(pfts,ngibbs=500)

  ## example of a PFT with multiple species (late hardwood)
  ## note that if you're just using Jenkins the acronym column is optional
  pfts = list(LH = data.frame(spcd = c(531,318),acronym=c('FAGR','ACSA3')))
}
```

Returns the following:

```
[1] "writing output to"                                
[2] "C:/Users/Simon Goring/Documents/GitHub/PalEON-FIA"
$spcd
[1] 531

$acronym
[1] FAGR
Levels: FAGR

[[3]]
[1] 6

[1] "Dropping allom rows: "
Error in (s1 <- s2) <- 0.1 : object 's1' not found
In addition: Warning message:
In nu(allom$parm$Component.ID) : NAs introduced by coercion
```

I propose that this be fixed.

## Context
My `sessionInfo()`:

```R
R version 3.3.1 (2016-06-21)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] tools     stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] PEcAn.allometry_1.4.8 mvtnorm_1.0-5         MCMCpack_1.3-7       
[4] MASS_7.3-45           coda_0.18-1           dplyr_0.5.0          
[7] knitr_1.14           

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.7        lattice_0.20-34    digest_0.6.10      withr_1.0.2       
 [5] assertthat_0.1     grid_3.3.1         R6_2.2.0           MatrixModels_0.4-1
 [9] DBI_0.5-1          magrittr_1.5       lazyeval_0.2.0     SparseM_1.72      
[13] Matrix_1.2-7.1     devtools_1.12.0    mcmc_0.9-4         memoise_1.0.0     
[...]

PR #1131, opened 2016-10-20
Multiple assignment bugfix

bpbond:
formatR helpfully changed  `s1     <- s2 = 0.1` to  `(s1    <- s2) <- 0.1` in `modules/allometry/R/AllomAve.R`. Which is incorrect. 
  
## Description
Thanks to @SimonGoring for spotting this. I don't see any other line like this in the PEcAn codebase.

## Motivation and Context
Code is incorrect.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)  #1130 

## Checklist:


modules/allometry/R/allom.BayesFit.R
```diff
@@ -126,7 +126,7 @@ allom.BayesFit <- function(allom, nrep = 10000, form = "power", dmin = 0.1, dmax
   }
   
   ## define priors
-  (s1    <- s2) <- 0.1  # IG prior on the within-study variance
+  s1     <- s2 <- 0.1  # IG prior on the within-study variance
```
robkooper: would it make sense to have this in the future be just two lines
s1 <- 0.1
s2 <- 0.1
just to make it easier to read.

SimonGoring: I agree.  I think it just makes for cleaner code.

PR #1132, opened 2016-10-20
Maat.model

serbinsh:
Bug squashing.  Fixing issues related to function use without calls to library first, or first defining the library for the function

## Description

Added the proper :: notation

## Motivation and Context

Bug fixes

## Types of changes

## Checklist:


models/ed/R/met2model.ED2.R
```diff
@@ -33,9 +33,11 @@
 met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, lst = 0, lat = NA, 
                           lon = NA, overwrite = FALSE, verbose = FALSE, ...) {
   overwrite <- as.logical(overwrite)
-  
+
+  # deprecated?  
   library(rhdf5)
```
mdietze: Once all references to functions from these packages are replaced with their pkg::func equivalent, yes the library calls can be dropped. But I don't think anyone's gone through ED2 wrapper in that much detail yet, so leave them for now.

models/maat/R/model2netcdf.MAAT.R
```diff
@@ -25,9 +25,6 @@ model2netcdf.MAAT <- function(outdir, sitelat = -999, sitelon = -999, start_date
   ## TODO is it OK to give site lat/long -999 if not running at a 'site'? 
   ## TODO !!UPDATE SO IT WILL WORK WITH NO MET AND WITH MET DRIVER!!
   
-  ### Load required libraries
-  library(PEcAn.utils)
```
mdietze: here you've dropped PEcAn.utils but haven't added any pkg::fcn refs -- was utils just not used?

models/maat/R/model2netcdf.MAAT.R
```diff
@@ -91,7 +88,7 @@ model2netcdf.MAAT <- function(outdir, sitelat = -999, sitelon = -999, start_date
     ############ Variable Conversions 
     ### Conversion factor for umol C -> kg C
     Mc <- 12.017  # molar mass of C, g/mol
-    umol2kg_C <- Mc * udunits2::ud.convert(1, "umol", "mol") * ud.convert(1, "g", "kg")
+    umol2kg_C <- Mc * udunits2::ud.convert(1, "umol", "mol") * udunits2::ud.convert(1, "g", "kg")
```
mdietze: can you not do this conversion in one call to ud.convert? "umol g" to "mol kg".  Also, those seem like odd units. Finally, check out the ```misc.convert``` function in utils that @istfer wrote -- it's a wrapper around ud.convert that tries to sort out other conversions, such at the molar mass conversion you've hacked into this.  Similar code is in model2netcdf.MAAT

models/maat/R/model2netcdf.MAAT.R
serbinsh: Some of the functions were already added properly (i.e. pkg::fcn) and I updated a few more then removed the need for library(PEcAn.utils)

utils/R/utils.R
```diff
@@ -609,18 +613,26 @@ misc.convert <- function(x, u1, u2) {
...
+#  }
+#}
 misc.are.convertible <- function(u1, u2) {
-  if (match(u1, c("umol C m-2 s-1", "kg C m-2 s-1")) ==
-      match(u2, c("kg C m-2 s-1", "umol C m-2 s-1"))) {
-    return(TRUE)
-  } else {
-    return(FALSE)
-  }
+  ifelse(match(u1, c("umol C m-2 s-1", "kg C m-2 s-1")) ==
```
serbinsh: @istfer if you think using ifelse() here is OK (revised version) I will delete the commented out code. I left that there for comparison.  

I will edit this before we merge these changes

istfer: @serbinsh it looks OK to me!

istfer: Oh wait I take it back.

istfer: how about something like this:

```
misc.are.convertible <- function(u1, u2) {
  
  # make sure the order of vectors match
  units.from <- c("umol C m-2 s-1", "kg C m-2 s-1", "mol H2O m-2 s-1", "kg H2O m-2 s-1")
  units.to <- c("kg C m-2 s-1", "umol C m-2 s-1", "kg H2O m-2 s-1", "mol H2O m-2 s-1")
  
  if(u1 %in% units.from & u2 %in% units.to){
    if (which(units.from == u1) == which(units.to == u2)) {
      return(TRUE)
    } else{
      return(FALSE)
    }
  } else {
    return(FALSE)
  }
}
```

serbinsh: @istfer I like it!  Do you want me to update in this PR?  I will be pushing some additional edits as well

istfer: yeah, sure!

serbinsh:
@mdietze OK to merge?  It looks like you already approved before I did my last push

PR #1142, opened 2016-10-24
Bugfixes get.traits depends and imports

rykelly:
Pretty straightforward... was just trying to test `run.meta.analysis`, but hit some bugs in the process. Mostly related to recent edits I think.

@mdietze, @dlebauer, please look at changes to `get.trait.data`. I know that we usually use the string 'AUTO' for `settings$meta.analysis$update`, but `get.trait.data`'s argument is called `forceupdate`. Sounds to me like an update should be forced iff `forceupdate==TRUE`; if it's `FALSE` or some string, then we should check first and only update if necessary. That's not what was happening before, but that's what I've done here.


utils/R/utils.R
```diff
@@ -643,17 +643,18 @@ misc.are.convertible <- function(u1, u2) {
...
+
   # split equation to LHS and RHS
   deri.var <- gsub("=.*$", "", expression) # name of the derived variable
   deri.eqn <- gsub(".*=", "", expression) # derivation eqn
     
   non.match <- gregexpr('[^a-zA-Z_.]', deri.eqn) # match characters that are not "a-zA-Z_."
   split.chars <- unlist(regmatches(deri.eqn, non.match)) # where to split at
   # split the expression to retrieve variable names to be used in read.output
-  variables <- unlist(stri_split_charclass(deri.eqn, paste0("[",noquote(paste0(split.chars, collapse="")),"]"), omit_empty = TRUE))
+  variables <- unlist(stringi::stri_split_charclass(deri.eqn, paste0("[",noquote(paste0(split.chars, collapse="")),"]"), omit_empty = TRUE))
```
rykelly: If you're saying that the regular `strsplit` function will return a vector with some empty string elements, can't you just do something like:

``` r
tmp <- strsplit(...)
result <- tmp[tmp != ""]
```

?

db/R/get.trait.data.R
```diff
@@ -82,8 +82,13 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon,
...
   traits <- names(trait.data.check)
 
+  # Set forceupdate FALSE if it's a string (backwards compatible with 'AUTO' flag used in the past)
+  if(!is.logical(forceupdate)) {
+    forceupdate <- FALSE
+  }
+  
   # check to see if we need to update
-  if (forceupdate || !as.logical(forceupdate)) {
+  if (!forceupdate) {
```
mdietze: @rykelly question here. If the logic was previously that we run this when update = TRUE or AUTO, and we've deprecated the update = FALSE option (now you just drop the meta-analysis block), wouldn't that imply that this check should be whether the meta.analysis tag exists? Is it not run in both cases?

db/R/get.trait.data.R
```diff
@@ -97,15 +102,15 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon,
           if (!file.exists(file.path(files$file_path[[id]], files$file_name[[id]]))) {
             foundallfiles <- FALSE
             logger.error("can not find posterior file: ", file.path(files$file_path[[id]], files$file_name[[id]]))
-          } else if (forceupdate && (files$file_name[[id]] == "species.csv")) {
+          } else if (files$file_name[[id]] == "species.csv") {
```
mdietze: I'm not following the logic of either the original or the revised. If we're doing AUTO, don't we want to do this check? If we're doing a force update, wouldn't we delete the file regardless of the check?  Maybe I'm just missing something. Whomever wrote this originally should take a close look. @dlebauer ?

db/R/get.trait.data.R
```diff
@@ -97,15 +102,15 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon,
...
             logger.debug("Checking if species have changed")
             testme <- read.csv(file.path(files$file_path[[id]], files$file_name[[id]]))
             if (!check.lists(species, testme)) {
               foundallfiles <- FALSE
               logger.error("species have changed: ", file.path(files$file_path[[id]], files$file_name[[id]]))
             }
             remove(testme)
-          } else if (forceupdate && (files$file_name[[id]] == "prior.distns.Rdata")) {
+          } else if (files$file_name[[id]] == "prior.distns.Rdata") {
```
mdietze: same

db/R/get.trait.data.R
```diff
@@ -116,7 +121,7 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon,
               logger.error("priors have changed: ", file.path(files$file_path[[id]], files$file_name[[id]]))
             }
             remove(testme)
-          } else if (forceupdate && (files$file_name[[id]] == "trait.data.Rdata")) {
+          } else if (files$file_name[[id]] == "trait.data.Rdata") {
```
mdietze: same

utils/R/utils.R
istfer: I guess, I could :smiley:  not sure what was my logic back then. could you make the change in your PR? I'll keep an eye on this function for a while anyway

```
variables <- unlist(strsplit(deri.eqn, paste0("[",noquote(paste0(split.chars, collapse="")),"]")))
variables[variables != ""]
```

istfer: I guess, I could :smiley: not sure what was my logic back then. could you make the change in your PR? I'll keep an eye on this function for a while anyway
```
variables <- unlist(strsplit(deri.eqn, paste0("[",noquote(paste0(split.chars, collapse="")),"]")))
variables[variables != ""]
```

db/R/get.trait.data.R
rykelly: Well, the standard workflow now uses the wrapper function `runModule.get.trait.data`, which sets forceupdate to FALSE if `is.null(settings$meta.analysis$update)` (which is the case if the meta.analysis block, or the update option within it, is missing). 

So in normal use cases, we should expect the argument to be either TRUE or FALSE (and if it's 'AUTO' the preceding expression will have set it to FALSE). Then we only enter this block if `forceupdate==TRUE`. Within this block, if all the files are found, then we'll return early and the rest of the function doesn't get run.

db/R/get.trait.data.R
rykelly: I don't know what the original was trying to do. With my changes above, we'll only be in here if `!forceupdate`, i.e., we're allowing an automatic update, so there's no need to think about the value of the flag within the block.

utils/R/utils.R
rykelly: I can try :)

rykelly:
@istfer I don't know how to find line comments on lines that have since been removed anymore (I thought they used to still be available?). Anyway, I made the change you proposed and committed it to this PR. Not tested, obviously...

istfer:
@rykelly oops, yes I updated my comment... just need the last line to be re-assigned to the `variables`
```
variables <- variables[variables != ""] # Remove empty entries
```
It should work, but I'll test it..

istfer:
@rykelly OK sorry for the mess, but I just did a quick test and it will break the workflow as it is. it does split the derivation expressions correctly but if no derivation eqn is provided (which is the usual case, just passing "NEE" for example) it throws an error, because `split.chars` will be empty. another check is needed before `strsplit`

```
convert.expr <- function(expression) {
  # split equation to LHS and RHS
  deri.var <- gsub("=.*$", "", expression) # name of the derived variable
  deri.eqn <- gsub(".*=", "", expression) # derivation eqn
  
  non.match <- gregexpr('[^a-zA-Z_.]', deri.eqn) # match characters that are not "a-zA-Z_."
  split.chars <- unlist(regmatches(deri.eqn, non.match)) # where to split at
  # split the expression to retrieve variable names to be used in read.output
  if(length(split.chars)!=0){
    variables <- unlist(strsplit(deri.eqn, paste0("[",noquote(paste0(split.chars, collapse="")),"]")))
    variables <- variables[variables != ""] # Remove empty entries
  } else {
    variables <- deri.eqn
  }

  return(list(variable.drv = deri.var, variable.eqn = list(variables = variables, expression = deri.eqn)))
}
```

Sorry for taking your time, if you want you can just revert back and I can deal with this later?

rykelly:
No problem, @istfer. Updated per your suggestion.

@mdietze, FYI, I just tried get.trait.data and run.meta.analysis (both via their 'runModule' counterparts), and they worked as expected for settings$meta.analysis$update=TRUE, FALSE, and AUTO, or if either that element or settings$meta.analysis itself is missing. 

I also noticed that I wrote runModule.get.trait.data to simply skip getting traits if there's no settings$meta.analysis block. Is there any other reason (in the workflow context) you'd want to get.trait.data?

mdietze:
I can't think of any reason you'd want to get the traits and priors if you're not running the MA.

rykelly:
OK, then the runModule code is correct as is. We may have discussed this at the time I wrote it...

istfer:
@rykelly have you pushed them? I don't see the changes in the PR

rykelly:
Pushed them now, @istfer

rykelly:
P.S. Sorry for the "confused" emoji. I couldn't find the "sheepishly" one I was looking for :)

Issue #1135, opened 2016-10-20
Optimize and improve make-based build system 


ashiklom:
## Description
The makefile I designed was a coarse first pass. The following are some of the outstanding issues that are worth addressing, if somebody has the time. I have this on my to-do list, but pretty far down -- I figured I'd open this to make my ideas public so others could try tackling them.

* [x] `make check` should use the main list of packages, not duplicate it -- a `foreach` statement or something is the way to go.
* [x] Add ability to build documentation for individual packages
* [x] Add ability to test and check individual packages
* [x] Generalize the build rules using patterns. Every package is installed in basically the same way, so this should be defined once and the only necessary mapping should be indicating package dependencies, as opposed to the current makefile, which has each package installation defined explicitly.

ashiklom:
All of these are addressed in #1143.

PR #1143, opened 2016-10-25
Update Makefile

ashiklom:
Per issue pecanproject/pecan#1135 , this does the following:

* Condenses commands down to general rules, so it's easier to add new packages. All you have to do now is add your package to either `BASE`, `MODULES`, or `MODELS` variable list, and, if necessary, add a line indicating dependencies for installation.
* Adds ability to document (`make document`), check (`make check`), and test (via `devtools::test` -- `make test`) packages. By default, running `make` will do `make document install`.

Also, corresponding updates to `scripts/build.sh`.

Also, during documentation process, I enocuntered a namespace conflict in `modules/emulator/R/zzz` that was resolved. `methods` package required for S4 methods is loaded by default for interactive sessions, but NOT when R scripts are run, which is confusing.

Also, updated some outdated documentation throughout.


modules/uncertainty/NAMESPACE
```diff
@@ -1,15 +1,15 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(plot,flux.uncertainty)
+S3method(plot,sensitivities)
+S3method(plot,sensitivity)
```
ashiklom: This is a naming issue that we should try to resolve. Every R style guide advises against using dots in names, and especially in exported functions, for this exact reason -- Roxygen interprets common ones as S3 methods.

mdietze:
Agreed. @ashiklom and @tonygardella can you make sure that the gitbook is updated for changes in make?

@ashiklom branch is out of date -- need you to merge from master before I can pull (for whatever reason it's not showing the "Update branch" button)

PR #1149, opened 2016-10-28
Various minor updates

rykelly:
## Description
* get.trait.data:
  * forceupdate argument defaults to FALSE
  * pft ouput dir is created if it doesn't exist (even though typically check.settings will have done that)
* Everywhere db.open is used, I immediately schedule closing it via on.exit. This ensures the connection is closed even if the function errors out.
  * Where db.connection was wrapped in `try()`, check now for a result of class 'try-error' rather than 'character'
* Added Western_* pfts to pftmapping.csv
* `get.parameter.samples` moved to utils package, which is the main place it's used
* Inadvertent but presumably correct changes to modules/uncertainty by re-roxygenizing
* Cleanup, add overwrite args to `do.conversions`
* In some **but not all** places, remove reliance on `NOW()` for finding just-added BETY entries. 
  * For one thing, per #1083 we shouldn't be setting the updated/created_at timestamps in BETY manually. 
  * For another, if multiple workflows are running it's entirely possible to get duplicate timestamps (at least down to the second, which is what we were using. 
  * Using the psql `RETURNING` keyword is cleaner and safer.
  * I'd doing this everywhere in the code we're currently relying on the set/query "unique" timestamp design, but again, have currently only done it for the ones that were causing me problems. 
* Updated get.analysis.filenames to put EA/SA data files into their own directories. It gets messy otherwise when doing a MultiSettings run with hundreds of sites...

## Motivation and Context
Fixing / cleaning things as they come up while setting up my multisite runs

## Types of changes

## Checklist:


mdietze:
Could you resolve conflicts here too so we can review

rykelly:
This branch got too screwed up. Git hates the fact that we have both Settings.Rd and settings.Rd, at least on my machine, and I can't fix it. I'll open a new PR.

PR #1151, opened 2016-10-28
Various minor updates (duplicate of broken #1149)

rykelly:
## Description
- get.trait.data:
  - forceupdate argument defaults to FALSE
  - pft ouput dir is created if it doesn't exist (even though typically check.settings will have done that)
- Everywhere db.open is used, I immediately schedule closing it via on.exit. This ensures the connection is closed even if the function errors out.
  - Where db.connection was wrapped in `try()`, check now for a result of class 'try-error' rather than 'character'
- Added Western_\* pfts to pftmapping.csv
- `get.parameter.samples` moved to utils package, which is the main place it's used
- Inadvertent but presumably correct changes to modules/uncertainty by re-roxygenizing
- Cleanup, add overwrite args to `do.conversions`
- In some **but not all** places, remove reliance on `NOW()` for finding just-added BETY entries. 
  - For one thing, per #1083 we shouldn't be setting the updated/created_at timestamps in BETY manually. 
  - For another, if multiple workflows are running it's entirely possible to get duplicate timestamps (at least down to the second, which is what we were using. 
  - Using the psql `RETURNING` keyword is cleaner and safer.
  - I'd doing this everywhere in the code we're currently relying on the set/query "unique" timestamp design, but again, have currently only done it for the ones that were causing me problems. 
- Updated get.analysis.filenames to put EA/SA data files into their own directories. It gets messy otherwise when doing a MultiSettings run with hundreds of sites...
## Motivation and Context

Fixing / cleaning things as they come up while setting up my multisite runs
## Types of changes
## Checklist:


utils/R/do.conversions.R
```diff
@@ -21,29 +22,30 @@ do.conversions <- function(settings) {
...
-                                                     start_date = settings$run$start.date, 
-                                                     end_date = settings$run$end.date, 
-                                                     model = settings$model$type, 
-                                                     host = settings$host, 
-                                                     dbparms = settings$database$bety,
-                                                     dir = settings$database$dbfiles, 
-                                                     browndog = settings$browndog)
-        settings$run$inputs[[i]][["path"]] <- result
-        status.end()
+      if (is.null(input$path)) {
```
mdietze: Not clear why status.check was removed from this if statement

utils/R/get.analysis.filenames.r
```diff
@@ -23,7 +23,7 @@ ensemble.filename <- function(settings, prefix = "ensemble.samples", suffix = "R
     ensemble.id <- "NOENSEMBLEID"
   }
   
-  ensemble.dir <- settings$outdir
+  ensemble.dir <- file.path(settings$outdir, 'ensemble.samples')
```
mdietze: This appears to put the ensemble samples in their own directory rather than in the base outdir. This will make them unaccessible via the web interface. Not sure the reason for this change

utils/R/get.analysis.filenames.r
```diff
@@ -76,14 +76,14 @@ sensitivity.filename <- function(settings,
...
-    sensitivity.dir <- settings$outdir
+    sensitivity.dir <- file.path(settings$outdir, 'sensitivity.samples')
   } else {
     ind <- which(sapply(settings$pfts, function(x) x$name) == pft)
     if (length(ind) == 0) {
       ## no match
       logger.warn("sensitivity.filename: unmatched PFT = ", pft, " not among ", 
                   sapply(settings$pfts, function(x) x$name))
-      sensitivity.dir <- file.path(settings$outdir, "pfts", pft)
+      sensitivity.dir <- file.path(settings$outdir, "pfts", pft, "sensitivity.samples")
```
mdietze: same issue as above with ensemble.dir

utils/R/get.analysis.filenames.r
```diff
@@ -76,14 +76,14 @@ sensitivity.filename <- function(settings,
   
   if (is.null(pft)) {
     # Goes in main output directory.
-    sensitivity.dir <- settings$outdir
+    sensitivity.dir <- file.path(settings$outdir, 'sensitivity.samples')
```
mdietze: same issue as above with ensemble.dir

utils/R/get.analysis.filenames.r
```diff
@@ -95,7 +95,7 @@ sensitivity.filename <- function(settings,
         ## no outdir
         settings$pfts[[ind]]$outdir <- file.path(settings$outdir, "pfts", pft)
       }
-      sensitivity.dir <- settings$pfts[[ind]]$outdir
+      sensitivity.dir <- file.path(settings$pfts[[ind]]$outdir, "sensitivity.samples")
```
mdietze: same issue as above with ensemble.dir

utils/R/do.conversions.R
rykelly: Accident. Reverted.

Issue #1156, opened 2016-11-01
Proposed updates to read.settings


rykelly:
## Description
I'd like to suggest a few changes to read.settings that I've half-assed implemented for my own purposes. If people like them, I'd work on the other half ass and then push them. Two things:

#### Add a block to settings that keeps track of whether `check.settings`, `update.settings`, and `fix.deprecated.settings` have been run before. 

Those functions add a flag to the block when they've been run, and by default won't do anything next time if they see the flag. This is useful if you're testing or just coming back to a workflow that's been run before, so you don't have to wait for those functions to run again unnecessarily. You might be thinking "who cares?", to which I reply "the poor sucker whose `settings` object is actually a list of 600 settings." To which you might retort "well rewrite the functions so they work more efficiently with `MultiSettings`," and I'd then have to admit "yes, that would be a great idea too but I don't have time now."

#### Allow `read.settings` to read .rda

It's *much* faster than reading the XML (again, really noticeable for a big MultiSettings). My hacky version currently saves both the XML and a .rda file after running the first time. On subsequent calls it looks to see whether the .rda has been saved already, and if so reads that by default instead of XML. 

For a cleaner version, I'd suggest that `read.settings`:

* Adds argument `format` that can be 'xml', 'rda', or 'auto'. The latter is default and decides based on file extension. 
* Adds argument `use.rda.if.available`. If `TRUE` (default), the function will look for an existing `<inputfileWithExtensionStripped>.rda`, and read that in if it's found. 
* If `inputfile` has no extension (and `format='auto'`), the function looks for `<inputfile>.rda` and `<inputfile>.xml` in that order. 

In addition, I'd add:
* A corresponding `save.settings` function that would save to .rda by default but allow saving to .xml manually if desired. Or, if folks want .xml to continue to be default that's fine with me.  Might be a good place to have a global option (similar to the `.db.utils` variable used by `PEcAn.DB`, though I think there are preferred methods) so that users can set this behavior once if desired and forget about it. 
* A function `convert.settings` to convert between formats. 

## Context
As noted, basically the work of `read.settings` and associated functions becomes nontrivial for large MultiSettings files, and these changes would make things a lot quicker. 

## Possible Implementation
See above

mdietze:
[...]

PR #1157, opened 2016-11-01
Implement #1156

rykelly:
## Description
Implemented the first idea discussed in #1156, i.e. that check/update/fix.settings functions only do their job once unless specifically asked to again. There are also a handful of cleanup / bugfixes included. 

## Motivation and Context
In most cases, there's no need to rerun the initial settings checks. 

## Types of changes

## Checklist:


web/workflow.R
```diff
@@ -39,29 +39,36 @@ if (is.na(args[1])){
...
+
 # Check for additional modules that will require adding settings
 if("benchmark" %in% names(settings)){
   library(PEcAn.benchmark)
   settings <- papply(settings, read.settings.RR)
 }
 
 # Write pecan.CHECKED.xml
-settings <- write.settings(settings, outputfile = "pecan.CHECKED.xml")
+write.settings(settings, outputfile = "pecan.CHECKED.xml")
```
mdietze: @rykelly not sure I follow what's going on here. My understanding was that @bcow deliberately split read.settings into two parts because she needed to put read.settings.RR in BEFORE settings were cleaned up so that she could leverage the checking/updating of settings and the filling in of machine-specific paths. You've now moved that ahead, which will break Benchmarking. 

@bcow also put all those misc helper functions inside of write.settings function so that it was easier to maintain. As per discussions on previous PR #1095, write.settings was a bad choice of name for that function because it is much more intuitive that write.settings would ONLY write out the settings (which I think is how you're using it), but I'd prefer it if there was a simple named single function here [something].settings that retains the functionality of the previous write.settings @bcow wrote

settings/R/write.settings.R
```diff
@@ -1,35 +1,18 @@
...
-    pecanfile <- file.path(settings$outdir, outputfile)
-    if (file.exists(pecanfile)) {
-      logger.warn(paste("File already exists [", pecanfile, "] file will be overwritten"))
-    }
-    saveXML(listToXml(settings, "pecan"), file=pecanfile)
+write.settings <- function(settings, outputfile, outputdir=settings$outdir){
+  pecanfile <- file.path(outputdir, outputfile)
+  if (file.exists(pecanfile)) {
+    logger.warn(paste("File already exists [", pecanfile, "] file will be overwritten"))
   }
```
mdietze: per discussion below on workflow.R, would prefer it if you restored @bcow's previous function, rename it to something more intuitive, and then recreated the write.settings that you have now.

web/workflow.R
```diff
@@ -39,29 +39,36 @@ if (is.na(args[1])){
...
-if (length(which(commandArgs() == "--continue")) == 0) {
-  file.remove(file.path(settings$outdir, "STATUS"))
+statusFile <- file.path(settings$outdir, "STATUS")
+if (length(which(commandArgs() == "--continue")) == 0 && file.exists(statusFile)) {
+  file.remove(statusFile)
 }
   
 # Do conversions
-settings <- do.conversions(settings)
+settings <- do.conversions(settings, overwrite.met=FALSE, overwrite.fia=FALSE)
```
mdietze: These are the default values (which themselves are hack) so I'm not sure we need to expose them at this time. Probably better to wait until we generalize this syntax.

web/workflow.R
bcow: @rykelly We had discussed changing the name of `update.settings` to something different and then renaming `write.settings` as `update.settings` or I thought maybe `settings2pecan.xml`. I agree that  the name`write.settings` is misleading.

rykelly: > @bcow ... needed to put read.settings.RR in BEFORE settings were cleaned up so that she could leverage the checking/updating of settings

OK, will change order. 

> As per discussions on previous PR #1095,

As usual, apologies for not being up to speed. OK, so it sounds like you both agree with me that `write.settings` is a function we need and it should simply write settings to file. So I'm going to leave this function for use wherever we just want to save the settings. (It will be easy to extend later to other formats too.)

The other thing that happens is this update/fix/addsecrets/check thing. It sounds like the consensus is for all of these to be wrapped in a single function (`clean.settings`?), so that's what I'll do. 

But for future discussion, honestly that doesn't sound right to me. Obviously `fix.deprecated` and `addSecrets`, based on their names, should only need to be run once near the beginning of the workflow. Without looking now, I'm pretty sure that much of check.settings also just sets defaults and check things that can't reasonably be expected to change over the course of a workflow (like the DB)—again, actions that should only happen once, and it doesn't really matter in what order as long as it's early in the workflow. So it seems like for the most part the functionality of this `clean.settings` function can be thought of as a one-time sanitation operation for wayward settings.

If indeed there are also some parts of these functions that need to be used at specific times for specific purposes, I would suggest that they should be broken out into their own functions and used as needed. 

I'd also suggest that in the long run default-setting behavior be moved closer to the functions that rely on it, but that can be another discussion.

mdietze: clean.settings is already in use.

I agree these functions are just a one-time sanitation process. Recall that benchmark_reference_runs stores a subset of a pecan xml settings object in the database, and those bits will definitely need cleaning/updating and localization to the specific machine they're running on.

I guess I like the idea of keeping all the parts of settings sanitation in one function because the average high-level user doesn't need to know/care about the details. In terms of how/whether those parts of sanitation are broken out into 1, 4, or n functions internally, I'm pretty agnostic

bcow: So to build on what @mdietze said: the XML files I'm using (you can see them [here](https://github.com/PecanProject/pecan/tree/master/modules/benchmark/inst/scripts) don't have any of the run information in them. The `read.settings.RR` inserts in everything using the ensemble.id. So the update/fix/addsecrets/check functions need to happen after that section of the workflow is run. Maybe we could combine those into `prepare.settings` and and keep `write.settings` separate. How does that sound?

settings/R/write.settings.R
bcow: `prepare.settings`?

rykelly:
Thought I responded somewhere but perhaps not. Anyway 8fb0bec should cover all our discussion above...

PR #1158, opened 2016-11-01
Sda minor tweaks

mdietze:


tonygardella:
There isn't anything clearly  sticking out to me that's breaking the build and for some reason the travis log isn't showing up for me like usual.

mdietze:
@tonygardella yeah, I was noticing that too, which is why I restarted the build. I wonder if there's some other Travis issue at play here or if it's somehow related to the make changes that @ashiklom pushed (but those passed the check, which they shouldn't have if they broke the build)

ashiklom:
The build was wrongly lenient before, where changes to package root files (e.g. DESCRIPTION, NAMESPACE) were ignored (I messed up my recursive wildcards). That may be why the build broke, though that kind of thing should show up in Travis.

mdietze:
@ashiklom yes, but Travis isn't even showing a log 2/3 of the time now, so it's pretty hard to figure out where it's choking.

ashiklom:
Yeah, that's weird. My personal Travis that started once I turned it on for my pecan fork (which uses an identical Travis.yml) always runs promptly after I push anything.

Maybe there are too many open pull requests or something? I think there may be a limit on the number of concurrent builds for a free account. Just a shot in the dark...

mdietze:
@ankurdesai build finally fixed

PR #1160, opened 2016-11-02
Benchmarking works for multi-site

bcow:
## Description

1. The benchmarking workflow works with multi-sites settings files. 
This means that `calc.benchmarks` works with `papply`  and `multisettings` objects. 
This did not actually require significant changes to benchmarking - I discovered an error I made in the initial settings xml file. 

2. Fixed a bug where `query.format.vars` returns a table with all the variables in the input file - which means that `load.data` loads all of them instead of just the specified variables in settings. So I added the optional argument `var.ids` to `query.format.vars` - default is `NA` ie load all the variables. 

3. Similarly, I fixed a typo where I left out a filter by metric ids specified in settings and so too many metrics were being calculated. 

4. Switching to explicit package references. 

5. Added `PEcAn.benchmark` to `PEcAn.all`

6. Small changes to documentation. 

### 1 variable, 1 metric, 1 site, 1 model
- settings.file: `modules/benchmark/inst/scripts/bm.1var.1metric.1site.1model.xml`
- [x] New Run 
- [x] Existing Run 
### 2 variables, 2 metric, 1 site, 1 model
- settings.file: `modules/benchmark/inst/scripts/bm.2var.2metric.1site.1model.xml`
- [x] New Run 
- [x] Existing Run 
### 2 variables, 2 metric, 1 site, 1 model
- settings.file: `modules/benchmark/inst/scripts/bm.2var.2metric.2site.1model.xml`
- settings will now be a multisite object
- [ ] New Run   <- I'll check this asap
- [x] Existing Run

## Types of changes

## Checklist:

- [ ] I have added tests to cover my changes. N

But I have added example XML files that can be used to test the benchmarking package.


db/R/query.format.vars.R
```diff
@@ -16,7 +18,7 @@ query.format.vars <- function(input.id,con,format.id){
   site.lat <- NULL
   site.lon <- NULL
   
-  if (missing(format.id)) {
+  if (is.na(format.id)) {
```
serbinsh: @bcow just to check the check for missing is on the format.id being NA and the ID, Lat/Long are set to NULL as placeholders until you fill in their values here correct?  Any cases where it wont be NA but you wont already have lat/long?

modules/benchmark/R/calc.benchmark.R
```diff
@@ -11,21 +11,29 @@
...
 calc.benchmark <- function(settings, bety) {
   
-  library(RPostgreSQL)
-  library(data.table)
+  # dplyr functions
+  tbl     <- dplyr::tbl
+  filter  <- dplyr::filter
+  rename  <- dplyr::rename
+  collect <- dplyr::collect
+  select  <- dplyr::select
```
serbinsh: if you end up using lubridate::year a lot you could put that at the top as well (i.e. year <- lubridate::year) and just use year...but that is just a preference thing of course

modules/benchmark/R/metric.R2.R
```diff
@@ -1,10 +1,12 @@
 ##' @name metric.R2
-##' @title Coefficient of Determination
+##' @title Coefficient of Determination (R2)
 ##' @export
 ##' @param dat dataframe
 ##' 
 ##' @author Betsy Cowdery
+
 metric.R2 <- function(dat, ...) {
+  logger.info("Metric: Coefficient of Determination (R2)")
```
serbinsh: @bcow just a nit picky thing.  Since we can't use superscript, what if it was instead (R_squared) to be descriptive and avoid any possible confusion....not a necessary change just thought I would voice my opinion

modules/benchmark/inst/scripts/benchmark.workflow.R
```diff
@@ -22,23 +18,38 @@ bety <- betyConnect("web/config.php")
 
 # ----- Pick a settings file -----#
 # WORKING: 1 variable, 1 metric, 1 site, 1 model 
-#settings.file <- "modules/benchmark/inst/scripts/bm.1var.1metric.1site.1model.xml"
+settings.file <- "modules/benchmark/inst/scripts/bm.1var.1metric.1site.1model.xml"
```
serbinsh: With these you could also set them as system.file()

modules/benchmark/man/create.BRR.Rd
```diff
@@ -4,7 +4,7 @@
 \alias{create.BRR}
 \title{Create benchmark reference run and ensemble}
 \usage{
-create.BRR(ens_wf, con)
+create.BRR(ens_wf, con, user_id = "")
```
serbinsh: Same here...blank or NULL?

serbinsh:
@bcow looks like you also need to update your branch now

modules/benchmark/R/calc.benchmark.R
robkooper: Not sure I like this, especially with things like year which is to easy a variable to use in the code as well and get strange compile/runtime errors. This is one problem I have in general with non strongly typed programming languages. Not sure if something like lintr would catch these type of errors (redeclaring a variable to a different type).

modules/benchmark/R/metric.R2.R
mdietze: @serbinsh since we can't use superscript you want to use subscript?  I think R2 is fine, or alternatively R^2

modules/benchmark/R/calc.benchmark.R
mdietze: I agree except for the places where a particular function is used a large number of times

serbinsh:
Its not a subscript its an underscore


On Nov 3, 2016, at 11:22 AM, Michael Dietze <notifications@github.com<mailto:notifications@github.com>> wrote:


@mdietze commented on this pull request.

________________________________

In modules/benchmark/R/metric.R2.R<https://github.com/PecanProject/pecan/pull/1160>:

>  metric.R2 <- function(dat, ...) {
+  logger.info("Metric: Coefficient of Determination (R2)")


@serbinsh<https://github.com/serbinsh> since we can't use superscript you want to use subscript? I think R2 is fine, or alternatively R^2

modules/benchmark/R/metric.R2.R
serbinsh: My point was "R_squared" as a simple plain text description.  It was an underscore not a subscript  since it was already wrapped in " "

PR #1163, opened 2016-11-03
Get rid of (empty) package-level man page for PEcAn.settings

rykelly:
## Description
Removed the mostly empty file that was responsible for roxygen creating a package-level man page for PEcAn.settings

## Motivation and Context
Mac OS (for one) doesn't allow two files whose names differ only in capitalization. But roxygen was auto-generating both settings.Rd (for the package) and Settings.Rd (for the class). The former doesn't contain anything useful, so I deleted it. If we want the package to have a man page in the future, we can simply name it PEcAn.settings anyway.


rykelly:
This is infuriating! Both git and os x are being very persnickety about case sensitivity! I think this will do it though.

PR #1168, opened 2016-11-03
Met.bug

tonygardella:
## Summary of Changes
Register files in model ```inst``` files are read in to tell which models need met with exact dates right before met2model.module. 

If they do, then ONE db input record is selected based on (site, format, and hostname - not date), then met2model is run to generate model specific met file, then that file record is associated with that existing input record. That input record will have it's start/end dates fixed to span all years files associated with it.

## Motivation and Context
In all the steps prior to met2model, dates are not handled at the file level because a "prefix" is used to name the file, but the full file name ```prefix.year.nc``` is generated in conversion functions. This functionality was throwing off models like SIPNET because it would find multiple valid input records for the format/site/dates and associate the dbfile record with one of them. This would make it impossible to do multiple runs at a site for different years a multiple input records would be found that were not associated with the correct dbfile needed for the dates of that run. 

## Notes
Just realized formatting changes to convert.input made it look like I changed the entire file. Also, I only tested this on a clean VM with no db records/files existing for a site and model, it could possibly break on machines that have a bunch of existing input records for a given site/format.


utils/R/convert.input.R
```diff
@@ -1,333 +1,388 @@
...
+    parent.id <- ifelse(is.null(input), NA, input$id)
+    
+    if ("newsite" %in% names(input.args) && !is.null(input.args[["newsite"]])) {
+      site.id <- input.args$newsite
+    }
+    
+    if (exact.dates && exists("inputid")){
+      
+      # New file creates but inew Input unecessary. Associate file with that Input record.
+      dbfileid <- dbfile.insert(in.path, in.prefix, 'Input', inputid, con, reuse=TRUE, hostname = machine$hostname)
```
mdietze: I think this case can be dropped

utils/R/convert.input.R
```diff
@@ -1,333 +1,388 @@
...
+                                    hostname = machine$hostname)
+    }
+    
+    successful <- TRUE
+
+    if (exact.dates){
+	newinput <- list()
+	newinput$input.id <- inputid
+	newinput$dbfile.id <- dbfileid
+	}
```
mdietze: Also seems unnessisary

mdietze:
@tonygardella it's been 10 days since I requested changes. Anything coming soon?

tonygardella:
@mdietze Yes!The two main things I'm doing-

 changing the date checking to dbfile.input.check  making it check for exact and/or containing dates depending on arguments passed to it. I think that should help make the logic easier to follow. 

Also, I'm separating out the flow of if statements for exact.dates vs. not exact.dates to make the logic easier to follow.

tonygardella:
Still testing right now. There are a couple more changes I need to make.

tonygardella:
@mdietze Seems to work fine with Sipnet-CRUNCEP. I will test Sipnet-Ameriflux next.

utils/R/convert.input.R
```diff
@@ -1,333 +1,388 @@
...
+                                        con = con, 
+                                        hostname = host$name, 
+                                        ignore.dates = TRUE)
+  
+  print(existing.dbfile, digits = 10)
+  print("end CHECK")
+  
+  if (nrow(existing.dbfile) > 0) {
+    
+    existing.input <- db.query(paste0("SELECT * FROM inputs WHERE id=", existing.dbfile[["container_id"]]),con)
```
tonygardella: existing.dbfile can never be >1 because the function that creates it, dbfile.input.check, selects the last one that was made and cuts out the other possible ones.

tonygardella:
@mdietze Tested on a site with no input records, it works. When I switched to working with ones with many input records the fix I commented on seems to do the trick. I still think this needs to be tested rigorously though.

mdietze:
@tonygardella do you mean that you're going to keep testing it before you want me to pull it, or that it's ready for me to look at?

utils/R/convert.input.R
```diff
@@ -22,73 +23,193 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
...
+      } else{
+      # No existing record found. Should be good to go with regular conversion.
+     }
+    
+    ##-------------------------end of exact.dates chunk------------------------------------#
+  
+    }else{
+    
+      ## Either Running met2model for models that do not need exact dates
+      ## Or running non met2model conversion scripts (download.raw, met2cf, etc.)
```
mdietze: This comment is misleading, as convert.inputs can (and hopefully will) be used outside of met.process.  You are simply running conversions that do not need exact date matches

utils/R/convert.input.R
```diff
@@ -22,73 +23,193 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
...
+                        successful <- FALSE
+                        on.exit(if (exists("successful") && successful) {
+            
+                        logger.info("Conversion successful, with overwrite=TRUE. Deleting old files.")
+                        remote.execute.R( file.deletion.commands$delete.tmp, host, user = NA, verbose = TRUE,  R = "R" )
+            
+                      } else {
+                          logger.info("Conversion failed. Replacing old files.")
+                          remote.execute.R(file.deletion.commands$replace.from.tmp, host, user = NA, verbose = TRUE, R = "R" )
+          })
```
mdietze: 1) is that parenthesis supposed to be there?
2) if this is the end of the else 3 lines above please move it over to be inline

mdietze: Also, I think I follow the flow of your logic as to why this needed to be here, but this block looks identical to the overwrite block above and it would be nice to avoid having the same code twice. Leave for now, but think about it.

Issue #1167, opened 2016-11-03
Met and input ensembles: adjustments to do.conversions, met.process, write.configs, and sda.enkf


mdietze:
PEcAn needs the ability to handle ensembles of inputs as a way of capturing uncertainties in drivers and initial conditions.

The following comes from a discussion with @araiho and @tonygardella at the PalEON SDA Hackathon

## Meteorology

Focusing first on the meteorology as the first case we want to deal with.

Key parts of the proposed design:
1. ensemble members will be named prefix.ens[ID] (e.g a CF met ensemble would be prefix.ens[ID].[year].nc). Therefore, most of the existing functions that will have to deal with ensembles will simply need to be called in a loop and internally they will have no idea that they're dealing with an ensemble member, they'll just treat the prefix.ens[ID] as the prefix.
2. Each ensemble member will be inserted into the database individually, not as one entry for the whole ensemble. This seems to make implementation easier and improves provenance (we'll know exactly which ensemble member was used for each run).

The above proposal would NOT be compatible with a previous discussion about the potential capacity to not store ensemble members but to be able to generate them on the fly having saved the seed. This comes at a cost of increased disk storage but much simpler and less error prone provenance & repeatability.

Within met.process we envision two major use cases:

1) the input meteorology is itself an ensemble

   within met.process we're basically just looping over each ensemble member

2) met.process itself generates the ensemble

For example, we would download one met, CF one met, generate N ensemble members within gapfill/downscale (so we'd end up with a whole list of results that would result in a whole vector of input.id's), then we'd loop over every ensemble member when calling met2model (so met2model would have no idea it's processing an ensemble member).

## write.configs

We'll need to update the ensemble option (code and settings) to let you choose WHAT you want an ensemble of (just params, just select inputs, both parameters and inputs). This would pass a specific list of inputs to each write.config.[model], so the model code doesn't need to know anything about ensembles

## SDA

* prior to the first call of split.inputs, need to detect the maximum number of ensembles (which input type has the largest ensemble) and create a vector of samples from that. To not waste ensembles, we'll take the first n ensembles in order then sample with replacement if we need a larger ensemble

### split.inputs.[MODEL]

split.inputs will take a new argument, ensemble number, which will default to 1

[...]

PR #1170, opened 2016-11-04
Sda met ensemble

araiho:
This code allows SDA to be run with input ensembles. 

## Description
I added a function to pecan/utils called get.ensemble.inputs to get specific ensemble inputs for a model run.

I also added to the SDA code to work with the above new utils function and model specific split input functions. Now split.inputs.MODEL will get one set of inputs from get.ensemble.inputs and split that into model specific inputs. This is trying to help with issue #1167


models/linkages/R/split.inputs.LINKAGES.R
```diff
@@ -17,10 +17,27 @@
...
-  new.met <- paste0(settings$rundir, "/climate.Rdata")  # doesn't do anything but write stuff to README
-  settings$run$inputs$met$path <- new.met  #HACK
+  #### Load all temp and precip met for site
+  met <- inputs$met$path
+  load(met)
   
-  return(settings$run$inputs)
+  #### Add years for indexing
+  temp.mat.years <- cbind(temp.mat,seq(850,2010,1))
+  precip.mat.years <- cbind(precip.mat,seq(850,2010,1))
```
mdietze: years should probably should go in met2model. This would break for any met other than PalEON 1

models/linkages/R/split.inputs.LINKAGES.R
```diff
@@ -17,10 +17,27 @@
...
+  
+  #### Grab needed years
+  temp.mat <- temp.mat.years[temp.mat.years[,13]==year(start.time):year(stop.time),1:12]
+  precip.mat <- precip.mat.years[precip.mat.years[,13]==year(start.time):year(stop.time),1:12]
+  
+  #### Save to climate file
+  save(temp.mat, precip.mat, file = paste0(settings$rundir, "/climate.Rdata"))
+  
+  new.met <- paste0('',settings$rundir, "/climate.Rdata")
+  settings$run$inputs$met$path <- new.met #Do I need to change the ID too? Idk what I would change it to.
```
mdietze: probably don't need to split the met at all -- would probably be much simpler to teach LINKAGES to read years

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -292,7 +307,7 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
   ###-------------------------------------------------------------------###
   ### loop over time                                                    ###
   ###-------------------------------------------------------------------###  
-  for (t in seq_len(nt)) {
+  for (t in 6:15) {
```
mdietze: don't hardcode!

PR #1174, opened 2016-11-07
Sda hackathn

ashiklom:
## Description
First pass at `read.` and `write.restart.ED2` functions for state data assimilation. 
Currently, they can only handle "AGB", but should be readily extensible to other variables (though each variable will likely have ED-specific quirks).

I also added some convenience functions for better editing of the ED2IN file.

This also fixes a hidden but important bug in `met2model.ED2` where the loop over years was coded incorrectly.

## Motivation and Context
* This is an important prerequisite for state data assimilation with ED

The following people will probably want to take a look at this: 
@mdietze @araiho @istfer @tonygardella

## Types of changes

## Checklist:


models/ed/tests/test.read.restart.R
```diff
@@ -0,0 +1,26 @@
+library(PEcAn.ED2)
+
```
mccabete: Is this a test that's meant to be included in this PR?

models/ed/tests/test.write.restart.ED2.R
```diff
@@ -0,0 +1,33 @@
+library(PEcAn.ED2)
```
mccabete: Again, is this a test that is meant to be included in this PR? It looks like a hackathon spesific script

models/ed/R/SDA.helpers.ED2.R
```diff
@@ -0,0 +1,63 @@
...
+                               file_year, 
+                               file_month,
+                               file_day)
+    histfile_string <- paste0("history-S-",
+                              datetime_string,
+                              ".*\\.h5$")
+
+    histfile <- list.files(histfile_path, 
+                           histfile_string, 
+                           full.names = TRUE)
```
ashiklom: Including the paths. By default, `list.files` (which is a base R function), only lists the file names, regardless of where you point it. `full.names` lists the full directory path, which is unambiguous.

models/ed/R/read.restart.ED2.R
```diff
@@ -0,0 +1,100 @@
+# TODO: Change inheritParams to template (also need to create template)
+
+#' @title State data assimilation read-restart for ED2
+#'
+#' @author Alexey Shiklomanov
+#' @inheritParams PEcAn.SIPNET::read.restart.SIPNET
+#' @examples
+#' \dontrun{
```
mdietze: @mccabete this is an example, not a test. Not sure what the objection is.

@ashiklom not 100% sure we want to @inheritParams PEcAn.SIPNET::read.restart.SIPNET -- that seems a bit dangerous (SIPNET could change and ED not) and to create a build order dependency that's probably not what we want.  If you think we should be defining the parameters ONCE and then referencing them, then this should be referencing the template package not SIPNET. Also, there's a trade-off here to balance in the goal of not duplicating code vs human readability -- on the one hand the package itself may be harder to maintain if you have to update the params for every model-specific variant, but on the other it's also harder to maintain if you have to go somewhere else to look up what the parameters mean when you're working in the code (i.e. does this run counter to the idea that the code should be internally self-documenting)

models/ed/R/write.configs.ed.R
```diff
@@ -163,10 +165,28 @@ write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings
...
     ed2in.text <- gsub("@PHENOL_END@", "", ed2in.text)
   }
-  
+
+  ## -------------
+  # Special parameters for SDA
+  # 
+  # TODO: This needs to be whatver line in pecan.xml that indicates that SDA 
+  # will be run.
+  if (!is.null(settings$sda)) {
```
mdietze: state.data.assimilation

models/ed/R/write.configs.ed.R
```diff
@@ -163,10 +165,28 @@ write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings
...
+    # Default values
+    sda_tags <- list(isoutput = 3,    # Save history state file
+                     unitstate = 1,   # History state frequency is days
+                     frqstate = 1)    # Write history file every 1 day
```
mdietze: Why would you be writing history every 1 day? Shouldn't you be using the assimilation frequency setting?

models/ed/R/write.restart.ED2.R
```diff
@@ -0,0 +1,167 @@
...
+                       "IDATEA" = strftime(start.time, "%d"),
+                       "IYEARA" = strftime(start.time, "%Y"),
+                       "ITIMEA" = strftime(start.time, "%H%M"),
+                       "IMONTHZ" = strftime(stop.time, "%m"),
+                       "IDATEZ" = strftime(stop.time, "%d"),
+                       "IYEARZ" = strftime(stop.time, "%Y"),
+                       "ITIMEZ" = strftime(stop.time, "%H%M"))
+
+  modstr <- 'Modified by write.restart.ED2'
+  ed2in_new <- ed2in_set_value_list(tag_val_list, ed2in_orig, modstr)
```
mdietze: 👍 to this syntax for modifying the ED2IN.

models/ed/tests/test.write.restart.ED2.R
mdietze: I'm going to side with @mccabete on this one -- up to now things in the 'tests' folder have been limited to formal tests called by testthat. Development code / examples have been kept in inst.

models/ed/R/read.restart.ED2.R
ashiklom: @mdietze I agree that this should be referencing a template package and not SIPNET -- this was a temporary solution because we don't have template `read.` and `write.` restart functions yet. I will draft blank versions as part of this PR -- that won't take long.

I will push back on the rest though. `inheritParams` is actually pretty smart -- what it does is match only existing function parameters that have not yet been documented, so that if there are any SIPNET-specific parameters that are not arguments to `write.restart.ED2`, they will be ignored ([more information](http://r-pkgs.had.co.nz/man.html#dry2). Given that, I think it's actually bad practice _not_ to use `inheritParams` for any standard `read.` and `write.restart` function parameters (e.g. `outdir`, `runid`, `settings`), since we will want these to mean the same thing in every context. For that matter, as we're revising PEcAn code, I think we should replace a lot of our parameter descriptions with `inheritParams` to enforce consistency and encourage parameters to be well documented once rather than haphazardly or inconsistently documented (or not documented at all) in many different places.

As far as human readability, if you actually look at the generated `.Rd` files, you'll see that the full parameter description is transferred from the SIPNET version -- there are no links to follow. The only place where this arguably reduces human readability is in the source code itself, but if people are looking at the sources anyway, it's easy enough for them to trace this back to the right package.

mccabete: @mdietze My mistake. I didn't read the "@example" comment.

models/ed/R/write.configs.ed.R
ashiklom: Yes, I should. Do I have to figure that out on my own using `difftime` or whatever, or is it a parameter I can grab from the settings file? I used days by default because I figured that was a setting guaranteed to generate files that are required, even though it's a lot slower.

mdietze: forecast.time.step tag

PR #1175, opened 2016-11-07
Substantial revisions to PEcAnRTM

ashiklom:
## Description
* Shortened the default burnin moving window width to 10% of chain width (was 20%), but changed it to use the Start iteration of the first window after the last non-converged window. This is more optimistic than the original, but successfully and precisely identified the burnin period in my tests.
* In PEcAnRTM, scale the log-likelihood by the ratio between true and effective sample size, to account for autocorrelation in the observations. This slightly increases the computational time (determining effective sample size requires fitting an AR1 model at each iteration) and substantially reduces the precision of posterior estimates, but they are now probably more accurate and consistently encompass the true value. Plus, posteriors are now much more sensitive to the prior, which is good for inversions.
* PEcAnRTM inversion functions now return more information, including the deviance (-2 log Likelihood) and effective sample size at each iteration as well as burnin and number of iterations (for `invert.auto`).

## Motivation and Context
Addressing autocorrelation in spectral observations has been on the RTM project to-do list for a while (it was highlighted by reviewers in revisions to Shiklomanov et al. 2016 RSE), and has become more important now that we are trying to do EDR inversions. 

@serbinsh @Viskari Note that this pull request changes the names and classes of some function returns,  ultimately making things more convenient, but meaning you will have to revise your scripts involving PEcAnRTM accordingly. 

## Types of changes

## Checklist:


modules/assim.batch/R/autoburnin.R
```diff
@@ -1,123 +1,58 @@
...
-    stop("Fewer than 50 iterations. Burnin will not be calculated")
-  }
-  if (method == "rmw"){
-    burnin <- getBurnin.rmw(jags_out, 
-                            width = width,
-                            njump = njump,
-                            threshold = threshold,
-                            use.confidence = use.confidence,
-                            ...)
-  } else if (method == "gelman.plot") {
```
istfer: @ashiklom have you got rid of the other implementation?

ashiklom: Temporarily, but I can add it back in if you prefer it.

istfer: my previous tests were with that version.. I may or may not use it in my own production, but at least for comparison I'd prefer having it around. So, yes, if it won't be too much trouble for you.

ashiklom: :+1: Added this back in https://github.com/PecanProject/pecan/pull/1175/commits/943a486540eb4237f6861dc1a540d7e408205431

modules/assim.batch/NAMESPACE
```diff
@@ -3,6 +3,7 @@
 export(assim.batch)
 export(autoburnin)
 export(correlationPlot)
+export(gelman_diag_mw)
```
istfer: need to roxygenise again? `gelman_diag_gelmanPlot ` is not exported

modules/assim.batch/R/autoburnin.R
```diff
@@ -1,123 +1,60 @@
-#' @name getBurnin
 #' @title Calculate burnin value
+#' 
 #' @description Automatically detect burnin based on one of several methods.
 #' @param jags_out List of MCMC sample matrices or `mcmc.list` object
-#' @param method Character string indicating method ("rmw" or "gelman.plot")
 #' @param threshold Maximum value of Gelman diagnostic
-#' @param use.confidence Logical. If TRUE (default), use 95% confidence interval for Gelman Diagnostic.
+#' @param method Character string indicating method (default="moving.window". 
+#' No other options are currently implemented.)
```
istfer: now that gelman.plot is back, would you like to change this?

modules/rtm/R/invert.auto.R
```diff
@@ -41,13 +57,25 @@ invert.auto <- function(observed, invert.options,
...
+  ngibbs.min <- invert.options$ngibbs.min
+  if (is.null(ngibbs.min)) {
+    ngibbs.min <- 5000
+  }
   ngibbs.step <- invert.options$ngibbs.step
+  if (is.null(ngibbs.step)) {
+    ngibbs.step <- 1000
+  }
   nchains <- invert.options$nchains
   inits.function <- invert.options$inits.function
```
istfer: unlike others you check above, do these (`invert.options$nchains` and `invert.options$inits.function`) always exist?

modules/rtm/R/invert.auto.R
```diff
@@ -73,14 +99,16 @@ invert.auto <- function(observed, invert.options,
       if (!is.numeric(parallel.cores) | parallel.cores %% 1 != 0) {
         stop("Invalid argument to 'parallel.cores'. Must be integer or NULL")
```
istfer: just curious, why not give a warning, assign a valid value and continue?

modules/rtm/R/invert.auto.R
```diff
@@ -73,14 +99,16 @@ invert.auto <- function(observed, invert.options,
...
       }
     }
     cl <- parallel::makeCluster(parallel.cores, "FORK", outfile = parallel.output)
     on.exit(parallel::stopCluster(cl))
-    print(sprintf("Running %d chains in parallel. Progress bar unavailable", nchains))
+    message(sprintf("Running %d chains in parallel. ", nchains), 
+            "Progress bar unavailable")
   }
   
   # Create inversion function to be passed to parLapply
```
istfer: just making a note, non-local arguments are used within the function

modules/rtm/R/invert.auto.R
```diff
@@ -126,79 +155,49 @@ invert.auto <- function(observed, invert.options,
...
-      out <- list(results = NA, 
-                  samples = PEcAn.assim.batch::makeMCMCList(samps.list))
+    inits <- lapply(out$samples, getLastRow)
+    inputs <- list()
+    for (i in seq_len(nchains)) {
+      inputs[[i]] <- list(runID = runID_list[i],
+                          inits = inits[[i]],
+                          resume = resume[[i]])
     }
+    invert.options$ngibbs <- ngibbs.step
```
istfer: does anything in the loop changes either of these values? what does this line do?

modules/rtm/R/invert.auto.R
```diff
@@ -218,17 +217,68 @@ combineChains <- function(samps1, samps2) {
     sampsfinal[[i]] <- rbind(samps1[[i]], samps2[[i]])
   }
   stopifnot(length(sampsfinal) == length(samps1))
```
istfer: maybe more informative error messages can be given within this function

modules/rtm/R/invert.custom.R
```diff
@@ -142,6 +151,11 @@ invert.custom <- function(observed, invert.options,
                       lower = param.mins, upper = param.maxs)
     inits <- fit$par
   }
+  if (!all(diag(init.Jump) > 0)) {
+    warning("Passed init.Jump matrix with zero values on diagonals. ",
```
istfer: you're not exactly checking only non-zeros here, although below you have another check for negative values as well.. maybe combine these into one check?

modules/rtm/R/invert.custom.R
```diff
@@ -242,32 +262,41 @@ invert.custom <- function(observed, invert.options,
     # Metropolis sampling step if all conditions have been met
```
istfer: how are the checks on [L.237](https://github.com/ashiklom/pecan/blob/40f0fe130286dd91898bf5cc012f668b1a321feb/modules/rtm/R/invert.custom.R#L237) and [L.257](https://github.com/ashiklom/pecan/blob/40f0fe130286dd91898bf5cc012f668b1a321feb/modules/rtm/R/invert.custom.R#L257) are different? I mean, why would a value within `param.mins` and `param.mins` fall outside the prior? 

Also, I was just wondering, instead of skipping to the next iteration right away, how it would affect to re-propose until all your parameters are valid and only then you calculate the metropolis ratio?

modules/rtm/R/invert.custom.R
```diff
@@ -176,7 +194,6 @@ invert.custom <- function(observed, invert.options,
   }
 
   # Precalculate quantities for first inversion step
-  rp1 <- tau_0 + nspec * nwl/2
   rsd <- 0.5
```
istfer: why is `rsd = 0.5`? you can update it with the mcmc maybe?

mccabete:
This isn't a comment on this PR, more a comment on my own ignorance, but when we're implementing a statistical test, where in the documentation do we reference theory? Is there a citation to a paper somewhere, or do we figure people know what they're doing and let google pick up the slack? Should people be assuming that the defaults are roughly representing the "standard practice" as represented in the literature?

istfer:
@mccabete I think that's a good point. As long as we implement something that is citable, putting a reference section to the header might be helpful, e.g.:

`#' @references Metropolis, Nicholas, et al. "Equation of state calculations by fast computing machines." The journal of chemical physics 21.6 (1953): 1087-1092.`

modules/rtm/R/invert.custom.R
ashiklom: 0.5 is the for the residual standard deviation. `rsd` is updated by the MCMC while sampling, but by a Gibbs sampling step rather than Metropolis proposal-rejection, so it always has a 100% acceptance rate.

modules/rtm/R/invert.custom.R
ashiklom: I implemented these differently to allow the use of less formal priors. For instance, `dnorm` is technically never not finite, but it might be convenient to assign a tight `dnorm` prior on a value.

Regarding your second question, I reject the parameters to force the Jump distribution to shrink. That way, if the sampler keeps proposing illegal values, the acceptance rate is low, and the jump distribution will narrow until it gets back to drawing mostly legal values.

modules/rtm/tests/invert_prospect.R
```diff
@@ -0,0 +1,77 @@
+library(devtools)
+#install('~/Projects/pecan/pecan/modules/rtm')
```
ashiklom: `devtools::install` is convenient because it reinstalls the package and attempts to reload it within the same R session, which is nice for developing because it lets me run tests and devlop code in the same session. `devtools::load_all` is an even more development-specific function that loads everything inside of a package, regardless of whether or not it is exported by the NAMESPACE. It's sort of like `source` ing every script in the R folder, plus their external package dependencies. I'm pretty sure you should never use `devtools::load_all` outside of testing.

modules/rtm/R/invert.auto.R
ashiklom: Good catch. This line is necessary because the initial number of samples `ngibbs.min` can be different from the number of samples between each convergence check after that (`ngibbs.step`), but that doesn't change as the run progresses. In an upcoming commit, I'll move that to right before the `while` loop begins.

modules/rtm/R/invert.auto.R
ashiklom: I dunno, I think between the call stack and the variable names, these are pretty self-explanatory. I'd have a hard time describing these errors in a way that this doesn't do already.

modules/rtm/R/invert.custom.R
ashiklom: I kept these separate to isolate problems between Jump distributions that are passed as inputs, which can be illegal (they just get dropped altogether), with the default initial Jump distributions that are set by the code, which _has_ to be legal and is therefore an error.

modules/rtm/R/invert.auto.R
ashiklom: Hah, good point, I forgot about that. I guess my thinking was that passing too many cores means the user gets the idea but doesn't know how many cores the computer has, while passing a non-integer number of cores means they don't understand how cores work.

modules/assim.batch/R/gelman_diag.R
```diff
@@ -0,0 +1,83 @@
+#' @title Calculate Gelman diagnostic on moving window
+#'
+#' @author Alexey Shiklomanov
+#' @param x MCMC samples, of class \code{mcmc} or \code{mcmc.list}
+#' @param width_fraction Fractional width of moving window. Default=0.1.
+#' @param width Width of moving window. Default is niter(x)*width_fraction
+#' @param njump Number of windows to calculate over
```
ashiklom: Actually ended up dropping that comment altogether, now that there are multiple methods.

ashiklom:
@mccabete The references thing is definitely a good idea. I added a reference here to the adaptive Metropolis algorithm, which is what this package uses, but a much more accessible, entry-level description is in Mike's forecasting book, I think in one of the chapters on Bayesian methods. [Wikipedia](https://en.wikipedia.org/wiki/Metropolis%E2%80%93Hastings_algorithm) also has nice descriptions.

Issue #1009, opened 2016-09-10
met.process refactor: met.module & user workflows


mdietze:
met.process has gotten too complicated, and as we discover things the original design couldn't handle, it's becoming a mess of hacks that is hard to maintain and extend. This is in contrast to the original idea, which was that it would be easy for contributors to add new modules or swap out modules (e.g. choose between alternative gap filling or down scaling routines).  What I propose below is intended to reduce code redundancy, reduce special cases, and increase flexibility & clarity.

**First idea**: A lot of what happens at each step in met process is the same. If the preprocessing says your supposed to run a step, you: (1) query the database and file system to perform checks [do files exist? do I need to extend date ranges? is this a force update? etc.] , (2) get variables organized, (3) pass the request off to download/convert.inputs/BrownDog, (4) do a bit of clean up, (5) pass the required ID's off to the next step.

I propose that these steps be refactored into a single function, met.module, that would take the module name, site info, and an arbitrary number (0 to n) of blocks of stuff returned by previous met.module calls.  Furthermore, lets store what's returned in a general list, rather than having a bunch of stage-specific id's floating around (raw.id, cf.id, ready.id, model.id, etc.)

So, for example, CRUNCEP might look like:
```
met.step[[1]] = met.module("download.CRUNCEP", site.info)
met.step[[2]] = met.module(paste0("met2model",model),site.info,met.step[[1]])
```
while Ameriflux might look like:
```
met.step[[1]] = met.module("download.Ameriflux", site.info)
met.step[[2]] = met.module("met2CF.Ameriflux", site.info,met.step[[1]])
met.step[[3]] = met.module("metgapfill",site.info,met.step[[2]])
met.step[[4]] = met.module(paste0("met2model",model),site.info,met.step[[3]])
```
 
**Second idea**: The above approach would be easy to put in a loop that could be generalized to allow more flexible workflow specifications.  

For example, the above Ameriflux example might be stored as:
```
1 = download.Ameriflux
2 = met2CF.Ameriflux(1)
3 = metgapfill(2)
4 = met2model.[MODEL](3)
```
or some XML equivalent and parsed at runtime as the loop iterates. A 'default' workflow for each met type could be stored in the met registry, but users could also specify workflows explicitly in the XML.  The approach would also allow for more flexible extensions, such as using one product to gapfill another:
```
1 = download.CRUNCEP
2 = download.Ameriflux
3 = met2CF.Ameriflux(2)
4 = gapFillFusion(3,1)
[...]

PR #1176, opened 2016-11-09
Downscale, Upscale, and Debias met code

jsimkins2:
Temporally downscale meteorological model output data based on observational data statistics.

Is this function name alright? temporal.downscaling.R is already taken, wasn't sure what to name it.


mdietze:
@tonygardella no one was ever assigned to review this

@jsimkins2 you might want to think about generalizing the name a bit -- I'm envisioning that we'll need a whole set of functions under the general heading of downscaling met and this is only one implementation of a specific part of temporal downscaling.  Oh, and go ahead and swap the period for an underscore in the name (per recent discussion to avoid periods)

The order and names of arguments here doesn't look like it will work with convert.input. I'd recommend thinking closely about how to get this function to work in that context and the discussion on issue #1009 or else it will be very hard to generalize met.process.

Next, you'll want to generalize the variable names as well as they're a bit too specific. What you're calling 'modelfile' isn't what PEcAn thinks of as being a model (i.e. output from an ecosystem model) and will often be reanalysis rather than pure GCM forward simulation. In keeping with convert.input and met.process, why not stick with in.path and in.prefix?  

Similarly, the second data set doesn't have to be flux data -- it's just the __training__ data -- it really could be anything. I'd recommend a general find/replace from flux to training or train in the arguments and throughout the code.

jsimkins2:
@mdietze got it. Thanks for the comments, just have a few questions.

1) Struggling to come up with a short name for this function. It's temporally downscaling a dataset using a normal distribution from the residual of a training dataset, having a tough time condensing that. Maybe we could poll the group in the telecon tomorrow?

2) Changed modelfile to in.prefix. Do I need to specify an in.path? 

3) Some of the arguments presented are important to the function such as the desired resolution a user wishes to downscale to or the window length they wish to use for the normal distribution. Will I need to remove these to work with convert.input?

4) For the changes you requested, I've also been changing debias.met (now debias_met) to keep consistency between these two functions. Is it ok to add that file to this pull request?

mdietze:
@jsimkins2 

1) the name doesn't have to be super short. Maybe something like met_temporal_downscale.Gaussian? Or for the second part maybe something else that really summarizes the specific type of met_temporal_downscale'ng you're doing -- this name would establish a convention that you and others could follow if they made alternative temporal downscale functions. Might want to put 'ensemble' in there too (e.g. Gaussian_ensemble)

2) Yeah, convert.inputs always sends the in.path (directory) separate from the in.prefix. Most functions then spit something out to the outfolder that includes (and often appends on) the in.prefix as the start of the file name

3) For now, I think it would be fine to start with sensible defaults [e.g. downscale to the resolution of the training data] and then later think about how to pass that info in explicitly via generalize met.process workflow pseudocode and/or the settings file. Either approach requires that you pass additional args to convert.input that then get pass on (by name) to the function.

4) sounds good.

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,420 @@
...
+      
+      # Precipitation_flux
+      # this takes the daily total of precipitation and uses that as a total possible amount of precip. 
+      # It randomly distributes the values of precipitation
+      rand_vect_cont <- function(N, M, sd = 1) {
+        vec <- rnorm(N, M/N, sd)
+        vec / sum(vec) * M
+      }
+      precip <- vector()
+      for (x in seq_along(model$precipitation_flux)){
```
araiho: Seems like the next four for loops are doing pretty much the same thing for different variables. Would it be a good place for another function?

jsimkins2: @araiho thanks for the comments and pointing me to formatR, this helps a lot. Will make changes and hopefully push to master before telecon tomorrow

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+    # within the window, and step through that window sequentially so that we are
+    # only using the standard devaition of values recorded at the same time as the
+    # value we are generating.
+    temp <- c("air_temperature", "air_temperature_max", "air_temperature_min")
+    df <- data.frame()
+    for (u in temp) {
+      woo <- vector()
+      a <- as.numeric(train[[u]])
+      sour <- as.numeric(source[[u]])
+      if (all(is.na(sour)) == TRUE) {
```
mdietze: the == TRUE is redundant

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+    # mean of the source value and the standard deviation of the training value
+    # within the window, and step through that window sequentially so that we are
+    # only using the standard devaition of values recorded at the same time as the
+    # value we are generating.
+    temp <- c("air_temperature", "air_temperature_max", "air_temperature_min")
+    df <- data.frame()
+    for (u in temp) {
+      woo <- vector()
+      a <- as.numeric(train[[u]])
+      sour <- as.numeric(source[[u]])
```
mdietze: you might want to think about more meaningful variable names -- I see that sour = source, but just 'a' for training data? No clue what woo is going to be.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+      if (all(is.na(sour)) == TRUE) {
+        woo <- rep(NA, reso_len)
+      }
+      if (all(is.na(sour)) == FALSE) {
+        for (x in seq_along(sour)) {
+          four <- vector()
+          for (n in seq_along(1:div)) {
```
mdietze: seq_len(div)

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+            if (lowday < 0) {
+              lowday <- x + n
+              highday <- x + n + wd * sd_step + n
+            }
+            if (highday > reso_len) {
+              highday <- reso_len
+              lowday <- x * div - wd * sd_step + n
+            }
+            four[n] <- rnorm(1, mean = sour[x], sd = sd(a[seq(from = lowday, 
+                                                              to = highday, by = sd_step)]))
```
mdietze: Trying to figure out what you're doing here. So I'm imagining, for example, downscaling a specific half hour within a 6 hour chunk of time from the source. First you're setting the mean temperature to be the temperature of the source, which will result in the downscaled met behaving like a step function since all 12 half hours in a 6hr block will have the same mean. That doesn't seem right to me -- it seems like we should be doing some sort of interpolation or some sort of correction that accounts for the anomalies in the training data (e.g. the 3rd half hour is always x% warmer than average across the training window).  The other thing that doesn't seem right is that the downscaling is only looking within a single source time step, so even if you fix the first problem you would still end up with discontinuities between time steps.  Next, for the standard deviation it looks like you're subsetting the training data that's within a window +/- some number of days around the current time and calculating it's standard deviation. If I understand correct, you are at least making sure you select the correct time of day. However, what you're doing fundamentally confuses uncertainty and variability. The SD you are calculating is a measure of how weather changes from day-to-day, not the uncertainty in the downscaling algorithm, which needs to be calculated as some sort of residual between the downscaled met and the observed met. Furthermore, the draws for any two time steps are independent, which is going to result in a driver that is very very noisy -- not only is that not how the world works, but it's going to blow up the land surface models. Finally, there seems to be a major time indexing problem. If, for example, my training data is 2 years long and my source data is 100 years long, when x is in years 3:100  then highday > reso_len, which implies every single day in every single year thereafter is downscaled  based on Dec 31 of year 2, and lowday will end up > reso_len, which will throw an error.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+    for (x in seq_along(source$precipitation_flux)) {
+      lowday <- (x - wd) * div
+      highday <- (x + wd) * div
+      if (lowday < 0) {
+        lowday <- 0
+      }
+      if (highday > reso_len) {
+        highday <- reso_len
+      }
+      four <- vector()
```
mdietze: this line is redundant

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+      four <- rand_vect_cont(div, source$precipitation_flux[x], sd = sd(train$precipitation_flux[lowday:highday]))
```
mdietze: Very similar issues as the downscaling of T. Furthermore, for precip you definitely don't want to use an untransformed Gaussian as it will result in massive amounts of negative precip (e.g. most days will be mean 0, which means half the simulated precip will be negative). I see you truncate the negative values below, but by doing that truncation without accounting for it in your mean and SD you're introducing a significant bias. For example, for a very dry season the observed mean has to be positive because it's a sum of a lot of zeros and a few small values. But to reproduce a timeseries that's mostly 0's and a few small values you need the (latent) mean of the Gaussian you're sampling from to be negative.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+          highday <- (x + wd) * div
+          if (lowday < 0) {
+            lowday <- 0
+          }
+          if (highday > reso_len) {
+            highday <- reso_len
+          }
+          four <- vector()
+          for (n in seq_along(1:div)) {
+            four[n] <- rnorm(1, mean = sour[x], sd = sd(a[lowday:highday]))
```
mdietze: same issues as Temp. Furthermore, since this is largely the same code as Temp, why not do it all in one block. Finally, these draws are done independent of all the other met drivers, which means you've got no correlations between your different met drivers.  Seems like it would make much more sense to draw all the met JOINTLY or to draw each new variable CONDITIONAL on the other met variables. Otherwise you'll end up with rainy days with low humidity and bright skies (which, besides being wrong, will blow up the land surface models)

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+        if (precip[i] == 0) {
+          p <- 2
+        }
+        hdry[i] <- A * p * (1 - exp(-1 * bmlist[i] * ((temp_max[i] - temp_min[i])^C)))
+      }
+      hdry[hdry < 0] <- 0
+      swflux <- hdry * I
+      swflux[swflux < 0] <- 0
+    }
+
```
mdietze: I couldn't find any uncertainties being dealt with in the radiation downscaling. So does every ensemble member get the same SW down?

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+    ncdf4::nc_close(loc)
+    
+    results$file <- loc.file
+    results$host <- PEcAn.utils::fqdn()
+    results$startdate <- paste0(year, "-01-01 00:00:00", tz = "UTC")
+    results$enddate <- paste0(year, "-12-31 23:59:59", tz = "UTC")
+    results$mimetype <- "application/x-netcdf"
+    results$formatname <- "CF Meteorology"
+    
+    invisible(results)
```
mdietze: First, by placing the return here you will exit the downscaling after the first ensemble member -- you need to return the results outside the loop!
Second, as is done here https://github.com/PecanProject/pecan/blob/master/modules/data.atmosphere/R/download.PalEON_ENS.R#L45 you'll want to put all these results in a list of data frames (one per ensemble member) rather than a single data frame.

modules/data.atmosphere/R/upscale_met.R
```diff
@@ -0,0 +1,107 @@
+##' upscale_met upscales the temporal resolution of a dataset
+##' @name upscale_met
+##' @title upscale_met
+##' @export
+##' @param outfolder
+##' @param input_met - met dataset
+##' @param site.id
+##' @param reso - numeric value in HOURS that is the desired resolution you wish the dataset to be upscaled to
```
mdietze: At some point (doesn't have to be this PR) you might want to generalize this as it doesn't allow upscaling to monthly or annual (since those are not fixed length)

modules/data.atmosphere/R/upscale_met.R
```diff
@@ -0,0 +1,107 @@
...
+  }
+  reso_len <- sp * 24/reso
+  
+  step <- length(data$air_temperature)/reso_len
+  upscale_data <- data.frame()
+  for (n in 1:length(var$CF.name)) {
+    for (x in 1:reso_len) {
+      upscale_data[x, n] <- mean(data[(x * step - step + 1):(x * step), n])
+    }
+  }
```
mdietze: Rather than two loops, it would be much more efficient to do this with apply (over variables) and tapply (over an indicator variable). e.g. ```apply(data,2,function(x){tapply(x,indicator,mean)})```

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+##' @param overwrite 
+##' @param verbose
+##' @author James Simkins
+met_temporal_downscale.Gaussian_ensemble <- function(in.path, in.prefix, outfolder, 
+                                                     input_met, train_met, site_id, overwrite = FALSE, verbose = FALSE, ...) {
+  # arguments
+  swdn_method <- "sine"
+  ensemble_members <- 10
+  wd <- 20
+  utc_diff <- -6
```
jsimkins2: Got it.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
jsimkins2: Got it, any ideas on how to do it differently to account for it in the mean and SD? Maybe having a moving average of some sort?

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
mdietze: I strongly disagree that ANN is the best option.

If you can only downscale for the years you already have data for, (1) why downscale at all when you can just us the training data? (2) how do you plan to apply this to the CMIP output, which is what I thought you were trying to downscale?

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+                 day_length = as.numeric(sunset$time - sunrise$time))
+    }
+    
+    swsource <- source$surface_downwelling_shortwave_flux_in_air
+    swdn <- vector()
+    
+    # The sine swdn_method produces an hourly sine wave of
+    if (swdn_method == "sine") {
+      
+      eph <- ephemeris(lat_train, lon_train, date = paste0(eph_year, "-01-01 00:00:00"),
```
jsimkins2: Yeah, really hadn't considered that. I'll change eph_year to be the year - 1 if it's a leap year

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+    
+    swsource <- source$surface_downwelling_shortwave_flux_in_air
+    swdn <- vector()
+    
+    # The sine swdn_method produces an hourly sine wave of
+    if (swdn_method == "sine") {
+      
+      eph <- ephemeris(lat_train, lon_train, date = paste0(eph_year, "-01-01 00:00:00"), 
+                       span = sp, tz = "UTC")
+      day_len <- eph$day_length
```
jsimkins2: The HMS portion of the ephemeris function is irrelevant to calculations, it's only needed for the function. It just calculates day length for each day of the year regardless of HMS time

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -0,0 +1,398 @@
...
+      
+      swflux <- vector()
+      swflux <- zoo::na.spline(tem.met)
+      swflux[swflux < 0] <- 0
+    }
+    
+    # The Waichler swdn_method doesn't need averaged SW train values, it sources SW
+    # downwelling flux based on Tmax-Tmin and Precipitation Reference is Waichler and
+    # Wigtosa 2003. Our no-precip coefficient is 2 instead of 1 (fits the data
+    # better)
```
jsimkins2: The real SW data. So when I ran this method with real data, using no-precip coefficient of 2 was much more realistic

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
jsimkins2: Not with the Waichler method, which uses ensemble specific temp_max, temp_min, and precip

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
jsimkins2: We only use the SD from the training data, but the means from the source data. So if source data is GFDL.2070, we use the SD from WCr.2005-WCr.2015. Which yes, the weather variability is going to be way different, but I literally couldn't find papers that had an answer. My best idea was increase the SD slightly per decade to try and better capture the increase of extremes. I could only find papers that said what not to use for temporal downscaling and even though this paper (http://people.bu.edu/dietze/manuscripts/Kumar2012-ICCS.pdf) claims ANN isn't the best method, it was one of the most complete papers I could find and had realistic correlations. My argument here is that yes, the method presented in this procedure is probably incorrect and unrealistic, but it's the best I could do in the time frame I had with no examples or templates to follow. My goal here was to try and kickstart the pecan temporal downscaling code development. Do you see any changes we could make to this before the next release to make it better?

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
mdietze: The real SW data for what site? I guess my question is whether this is a site-specific parameter that needs to be passed and/or fit to the training data, or whether 2 really will do better everywhere?

jsimkins2: Only tested for Cheas sites, I'll have to test for others. The reason I just went with it was because using 1 for no precip factor didn't even account for half of the actual radiation, but I see your concern, perhaps it would be regionally dependent

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+    }
+    colnames(upscale_flux) <- var$CF.name
+    for (x in 1:reso_len){
+      upscale_flux$air_temperature_max[x] <- max(flux$air_temperature[(x*step-step+1):(x*step)])
+      upscale_flux$air_temperature_min[x] <- min(flux$air_temperature[(x*step-step+1):(x*step)])}
+    
+    # Now we start a for loop for the ensemble members and begin downscaling. If method = "normal", a random normal distribution
+    # is used to downscale as so; (mean = value of modeled data) (sd = +/- window_days of upscale_flux data at the same time intervals)
+    # Future work is to implement more methods for downscaling
+    for (e in seq_along(1:ensemble_members)){
```
bpbond: I think you mean `seq_len(ensemble_members)`

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+      }
+      
+      # Air Temperature Max and Min
+      temp_max <- vector()
+      if (all(is.na(model$air_temperature_max)) == TRUE){
+        temp_max <- rep(NA,reso_len)}
+      if (all(is.na(model$air_temperature_max)) == FALSE){
+        for (x in seq_along(model$air_temperature_max)){
+          four <- vector()
+          for (n in seq_along(1:div)){
```
bpbond: see note above

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+              lowday = x*div-wd*sd_step+n}
+            four[n] <- rnorm(1, mean = model$air_temperature_max[x],
+                             sd = sd(upscale_flux$air_temperature_max[seq(from=lowday,to=highday,by=sd_step)]))}
+        temp_max <- append(temp_max,four)}
+      }
+      
+      temp_min <- vector()
+      if (all(is.na(model$air_temperature_min)) == TRUE){
+        temp_min <- rep(NA,reso_len)}
+      if (all(is.na(model$air_temperature_min)) == FALSE){
```
bpbond: Here and throughout, why not simply `if(!...)` (to test false)?

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+        for (x in seq_along(model$air_temperature_min)){
+          four <- vector()
+          for (n in seq_along(1:div)){
+            lowday <- x*div-wd*sd_step+n
+            highday <- x*div+wd*sd_step+n
+            if (lowday < 0){lowday = x+n
+            highday = x+n+wd*sd_step+n}
+            if (highday > reso_len){
+              highday = reso_len
```
bpbond: Use `<-` not `=`

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+  
+      north <- vector()
+      for (x in seq_along(model$northward_wind)){
+        lowday <- (x-wd)*div
+        highday <- (x+wd)*div
+        if (lowday < 0){lowday = 0}
+        if (highday > reso_len){highday = reso_len}
+        four <- vector()
+        for (n in seq_along(1:div)){
+          four[n] <- rnorm(1, mean = model$northward_wind[x],
```
bpbond: Since `rnorm` will generate as many values as you tell it, seems very inefficient to be doing this in a loop, generating one by one?

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+        for (n in seq_along(wav)){
+          l[n+hr] = wav[n]}
+        for(n in seq_along(1:(24-(length(wav)+hr)))){
+          l[n+hr+length(wav)] = 0}
+        
+        swdn <- append(swdn,l)}
+        
+        swflux <- vector()
+        sw_step <- length(swdn)/reso_len
+        for (x in 1:reso_len){
```
bpbond: `seq_len(reso_len)`

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+      if (swdn_method == "Waichler"){
+        inter <- paste0(reso," hour")
+        days <- seq(as.POSIXct(paste0(eph_year,"-01-01 00:00:00", tz = "US/Central")), 
+                    as.POSIXct(paste0(eph_year,"-12-31 18:00:00", tz = "US/Central")), by=inter)
+        
+        Z <- Ratmosphere::SZA(days,lat_flux, lon_flux)
+        I <- 1000*cos_d(Z)
+        m <- vector()
+        for (i in seq_along(1:12)){
+          m[i] <- Hmisc::monthDays(as.Date(paste0(year,'-',i,'-01')))}
```
bpbond: Similar to `rnorm` comment above - `Hmisc::monthdays` is parallelized, so why do this in a loop?

modules/data.atmosphere/R/downscale.met.R
```diff
@@ -0,0 +1,425 @@
...
+  ncdf4::nc_close(loc)
+  
+  results$file <- loc.file
+  results$host <- PEcAn.utils::fqdn()
+  results$startdate <- paste0(year,"-01-01 00:00:00", tz = "UTC")
+  results$enddate <- paste0(year,"-12-31 23:59:59", tz = "UTC")
+  results$mimetype <- 'application/x-netcdf'
+  results$formatname <- 'CF Meteorology'
+  
+  invisible(results)
```
bpbond: Add explicit `return`

modules/data.atmosphere/R/downscale.met.R
bpbond: and there's no need to do `== TRUE` ever

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
mdietze: * Ok, so as said in the overall review summary, let's focus on the smaller bugs in the code now and then move forward with discussions about how to do something better for the following release.
* Everything Christy's been doing on met ensembles for PalEON is a direct response to the frustrations the whole team has had with the Kumar et al 2012 product and code and the lessons learned. It's because I was a co-author on that paper/product that I feel so strongly that you should not go down that path again. This is also why I've been so strongly suggesting that you work more closely with Christy -- I'd really like to see her code as an alternative met_temporal_downscale module that users could choose.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
mdietze: Ok, so lets separate the issue of which variance to use from how you correct the statistical moments for truncation and focus on the second for now. You are working with a truncated Normal distribution https://en.wikipedia.org/wiki/Truncated_normal_distribution with one sided truncation. That means that the parameters you want to feed into your rnorm are not the sample mean and standard deviations, but rather bias corrected parameters. Look at the wikipedia link and other sources for the exact math on that, but the idea is that to get the correct observed mean and SD you need to feed in different parameters (e.g. in general a lower, and often negative, mu and a higher sigma) so that when you sample from that distribution and then truncate the zeros you end up matching the observed mean and sd

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
mdietze: In the conditional version, you have choices on what you condition on. For example, hydrologists often start by conditioning on the downscaled precip. Christy started with air T and then conditioned other variables on that.

One simple way to do that is to just filter the values in your window -- e.g. if you start with precip, once you draw the precip only use the values in your window that have a similar precip to choose from.  This is similar to how the MDS gap-filling code works. Indeed, you might want to look at the gapfilling lit as well because downscaling is very similar to gap-filling, it's just that the gaps are very regular and persistent.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
jsimkins2: Ok, sounds good. Yeah I've been trying to keep up with Christy's code the best I can but communicating has been rather difficult (normally takes her at least a week to respond to an email) so I kinda had to push forth alone. But this feedback is good, thank you for the explanations. Just to make sure I'm understanding this correctly, let's say we want to downscale a MACA product (daily res) at Harvard Forest. We have ~ 10 years of overlapping data, so we use linear regression to debias the MACA data and that's the 'training' phase. So we use the same regression in the 'application' phase to debias MACA data in all years without overlapping data. At what point do we downscale from daily to half hourly?

ankurdesai:
Some small points. First, sorry to not be on top of this over the last few days. Swamped with multiple commitments. I agree on the following points:
  - The downscaling should have numerous user-selected options that generate high frequency met from low frequency met, based on training data, include butt-simple but analytically simple to complex but hard to diagnose how met generated approaches, that downscales all time of the low freq met. I always chose butt-simple on first development
  - All methods should preserve the means (“bias”) of the training data in the output data
  - There may be differing need for monthly->daily and daily->sub-daily. Christina’s code, if I remember correctly, does not go sub-daily and hence most of that code doesn’t have direct application to what’s happening here (except the theory and general approach)
  - It should be agnostic of the type of training data or input data, as long as both are CF, UTC, and time dimension is identifiable. This is another limitation of Christina’s code which makes it not directly translatable to Pecan (it is specific to Paleo-CMIP5 I believe). I agree the code should be modified to be as flexible as possible on what is being input as long as it follows Pecan standards.
  - The methods should be statistically defensible, and preferably linked to published methods where available. I defer to Mike’s expertise here 
  - Methods should downscale all key variables used in ecosystem models (based on pecan standard), and for ones with no information, just “dumb-scale” it to the same value for each hour
  - Multi-variable approaches is ideal, but I encouraged James to just get univariate working so we can generate something (and this is the norm in the climate science community, even if wrong). Christina’s method does do some conditioning, but not a lot.
  - Methods should aim to generate ensembles, but there may be some methods where that is not the norm
  - I agree we don’t want to do anything with the ANN produced earlier

My thoughts for now.  -ankur

mdietze:
I agree with almost everything @ankurdesai, and in particular that it's important to start with something simple, but I did have a few comments:

1) Christy = Christine not Christina
2) Christy **DOES** have a daily -> subdaily downscaling code
3) The fact that Christy's code is currently Paleon / CMIP specific is exactly why I've been so eager for James and Christy to generalize it. I don't think any of the LOGIC in her code is specific to those data, just some of the variable names, timesteps, etc.
4) methods should preserve the means ON AVERAGE over ensemble members. Including uncertainty means that any particular ensemble member may differ from the original data. That difference may be persistent over time since many uncertainties don't just average out.
5) I agree that the simplest methods on the list, as well as a number of published methods from the literature, will not produce ensembles. That's fine provided those uncertainties get added in when we develop more advanced methods. I think we may be hard pressed to find published ensemble methods -- we may end up being the ones who write those papers

jsimkins2:
@ankurdesai @mdietze Yes, Christy does have a daily -> subdaily temporal downscaling script but she was developing it during October and didn't finish it until early November and communication was very slow with her. The logic shouldn't be too different with the procedure, but a lot of her syntax was paleon and NLDAS specific and I have a lot of unanswered questions about it, especially with the dataset. Further,  generalizing hers for ANY met product and with the PEcAn workflow is another challenge especially when not all of PEcAns met products have the same variables. Since emailing with Christy was difficult, I hope to be able to meet with her for a day or two when she starts up her new job in Chicago as I think that will be much more productive

PR #1177, opened 2016-11-10
Fix query.format.vars calls in PDA

istfer:
Minor bugfixes in the PDA code after recent benchmark changes

## Description
Changed` query.format.vars` calls to use `dplyr` connection.
Also addressed some of the changes in align.data and fixed one minor bug. 

## Types of changes

## Checklist:

### DIFF
--- a/modules/benchmark/R/match.timestep.R
+++ b/modules/benchmark/R/match.timestep.R
@@ -10,5 +10,5 @@ match.timestep <- function(date.coarse, date.fine, data.fine) {
   
   midpoints <- c(-Inf, head(as.numeric(date.fine), -1)) + c(0, diff(as.numeric(date.fine)) / 2)
   
-  return(data.fine[unique(findInterval(date.coarse, midpoints))])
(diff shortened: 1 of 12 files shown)


modules/benchmark/R/match.timestep.R
```diff
@@ -6,9 +6,9 @@
...
 ##' 
 ##' @author Istem Fer
-match.timestep <- function(date_coarse, date_fine, data_fine) {
+match.timestep <- function(date.coarse, date.fine, data.fine) {
   
-  midpoints <- c(-Inf, head(date_fine, -1)) + c(0, diff(as.numeric(date_fine)) / 2)
+  midpoints <- c(-Inf, head(as.numeric(date.fine), -1)) + c(0, diff(as.numeric(date.fine)) / 2)
   
-  return(data_fine[findInterval(date_coarse, midpoints)])
+  return(data.fine[unique(findInterval(date.coarse, midpoints))])
```
mdietze: Not sure I see why you need a unique here. That seems to risk changing the length and order of the data

PR #1178, opened 2016-11-14
PalEON SDA Met

mdietze:
Two weeks ago at the PalEON SDA hackathon I got started on trying to generalize met.process to work with an ensemble of meteorology. But I didn't get very far -- just getting to the step of having a download function that inserts the ensemble met and returns a list instead of a dataframe -- and then was offline all last week. I'm not going to get to this for a while (if I stay on my pre-AGU schedule, not till after Thanksgiving unless @araiho or @jsimkins2 need this for AGU) so I wanted to go ahead and PR this before I forgot and someone else reinvented it.

Relevant to issue #1167 

Non-breaking change -- should have no impact on any other code.  Folks who are working with met.process might want to look to see how I intend to return an ensemble.


modules/data.atmosphere/R/download.PalEON_ENS.R
```diff
@@ -0,0 +1,62 @@
...
+##' @title download.PalEON
+##' @export
+##' @param outfolder
+##' @param start_year
+##' @param end_year
+##' 
+##' @author Betsy Cowdery, Mike Dietze
+download.PalEON_ENS <- function(sitename, outfolder, start_date, end_date, overwrite = FALSE, ...) {
+  
+  library(PEcAn.utils)
```
ashiklom: This is bad practice. If you _need_ to load the entire package namespace, the way to do that is to add `@import PEcAn.utils` as a roxygen tag. Preferably, if you only need certain functions from that package, the recommended way to do that is to use `@importFrom PEcAn.utils logger.info logger.warn logger.severe`. I still think the best thing to do for code readability is to explicitly do `PEcAn.utils::logger.warn`, etc. at every instance.

modules/data.atmosphere/R/download.PalEON_ENS.R
```diff
@@ -0,0 +1,62 @@
...
+  ###### NEED TO AUTOMATE !!!!
+  
+  
+  ## extract the data and loop over ensemble members
+  dlpath <- "/home/carya/test_ensemble/"  ## download path
+  setwd(dlpath)
+  ens_zip <- dir(dlpath,pattern="tar.bz2",)
+  results <- list()
+  for(i in seq_along(ens_zip)){
+    system(paste("tar -xvjf",ens_zip[i]))  ## unzip file
```
ashiklom: `system2` is supposedly a more modern and reliable interface that we should adapt. The syntax for that is the name of the command followed by a character vector of arguments.

```
system2("tar", c("-xvjf", ens_zip[i]))
```

modules/data.atmosphere/R/download.PalEON_ENS.R
```diff
@@ -0,0 +1,62 @@
+##' Download PalEON files
+##'
+##' @name download.PalEON
+##' @title download.PalEON
```
ashiklom: I would drop the `@name` tag. I think the only time that should be used is if you want the documentation name to be explicitly different from the function it's describing, and in this case, I don't think that's what you want.

The `@title` tag is the title of the documentation page. I think what you _actually_ want is to drop this line as well and stick the `@title` tag before `Download PalEON files`. Incidentally, the way the documentation is formatted now, the title is implied. By default, roxygen interprets blocks intelligently (though I think it's best to be explicit anyway), so a block like the following

```
#' Page title <Assumed to be the @title tag>
#' <blank>
#' Text <Assumed to be the @description tag>
#' @param, @author, etc...
```

modules/data.atmosphere/R/download.PalEON_ENS.R
```diff
@@ -0,0 +1,60 @@
...
+  ###### NEED TO AUTOMATE !!!!
+  
+  
+  ## extract the data and loop over ensemble members
+  dlpath <- "/home/carya/test_ensemble/"  ## download path
+  setwd(dlpath)
+  ens_zip <- dir(dlpath,pattern="tar.bz2",)
+  results <- list()
+  for(i in seq_along(ens_zip)){
+    system2("tar",paste("-xvjf",ens_zip[i]))  ## unzip file
```
ashiklom: You actually don't even need `paste` here -- arguments are space-separated automatically. Just an FYI.

Issue #1114, opened 2016-10-19
met.process get error


yan130:
## Description

when I run met.process I get the following error which I did not enter a month ago:

[1] "start CHECK Convert.Inputs download.NARR NA localhost /home/polyglot/cache/PEcAn/NARR/ NARR application/x-netcdf 1135 2001-01-01 00:00:00 2001-12-31 23:59:59"
data frame with 0 columns and 0 rows
[1] "end CHECK"
Error in .parseArg(l[[i]]) : could not find function "is.POSIXt"
Calls: met.process ... .download.raw.met.module -> convert.input -> listToArgString -> .parseArg
Execution halted
## Possible Implementation

To me, listToArgString.R need to include lubridate package.  Or is it because I run something wrong?
@robkooper if you want to reproduce, just run: Rscript PEcAn.R example.xml US-Dk3.pecan.nc

mdietze:
I think the fix to this is in my open pull request #1112

yan130:
@mdietze I just saw it thanks.

yan130:
additional bug I get when I just pull the newest branch:

Error in convert.input(input.id = NA, outfolder = outfolder, formatname = register$format$name,  : 
  object 'username' not found
Calls: met.process -> .download.raw.met.module -> convert.input
Execution halted

I checked there is no definition for "username" in modules/data.atmosphere/R/download.raw.met.module.R

thanks

mdietze:
Probably an issue with the db connection argument -- that does have a username

mdietze:
That sounds right, as str_ns was also missing from the download module. Would be happy to quickly process a Pull Request with that fix.

yan130:
one more bug I enter after git pull.

line 97 in mat.process.R
 `   format.vars <- query.format.vars(con = con, format.id = register$format$id)  # query variable info from format id`

while this is the definition of query.format.vars:
```
query.format.vars <- function(input.id,bety,format.id=NA,var.ids=NA){
  con <- bety$con
  ......
}
```


so I get error:

Error in query.format.vars(con = con, format.id = register$format$id) : 
  unused argument (con = con)
Calls: met.process -> query.format.vars
Execution halted

mdietze:
@bcow this is your bug, please submit a pull request ASAP
https://github.com/PecanProject/pecan/commit/6b151343f9c4bc1af90289370ad6898366f541a7

If you change the arguments to any function, you need to make sure to grep for all uses of that function! 

[...]

PR #1179, opened 2016-11-15
Incorporating dplyr database functions

bcow:
## Description

- Moving `dplyr` database functions over to the `PEcAn.DB` package.
- Updating bety arguments  in `query.trait.vars` in `met.process` --> Issue #1114
- Cleaning up documentation 

## Types of changes

   - This may temporarily mess up dependencies in benchmarking but I'm working on fixing it and it won't affect other people's work. 

## Checklist:


modules/data.atmosphere/R/met.process.R
```diff
@@ -59,7 +59,12 @@ met.process <- function(site, input_met, start_date, end_date, model,
   }
   
   # set up connection and host information
-  con <- db.open(dbparms)
+  bety <- src_postgres(dbname   = dbparms$dbname,
```
tonygardella: Will this need to be  ```dplyr::src_postgres()```? You'll also probably need to add dplyr to the DESCRIPTION file in data.atmosphere under Imports.

db/R/query.base.R
```diff
@@ -70,7 +70,7 @@ query.base.con <- function(settings,...){
 ##' @param con database connection object
 ##' @return nothing, as a side effect closes all open connections
 ##' @author Rob Kooper
-##' @export
+##' @export query.close
```
ashiklom: Why did you do this? You do not need to specify which function you are exporting -- roxygen knows it's the one the documentation is associated with. This introduces another place where the function name needs to be maintained.

db/R/query.format.vars.R
```diff
@@ -2,12 +2,14 @@
...
 ##' @param con : database connection
-##' @export 
+##' @export query.format.vars
 ##' 
 ##' @author Betsy Cowdery , Ankur Desai
 ##' 
-query.format.vars <- function(input.id,bety,format.id=NA,var.ids=NA){
+query.format.vars <- function(bety,input.id=NA,format.id=NA,var.ids=NA){
 
+  if(is.na(input.id) & is.na(format.id)){logger.error("Must specify input id or format id")}
```
ashiklom: Should be `PEcAn.utils::logger.error`.

modules/data.atmosphere/R/met.process.R
bcow: Done. Good catch.

db/R/query.base.R
bcow: This is the only way that I have been able to work around functions with dots in their names getting exported as S3-methods instead of functions.
I'm not sure if we should also be using the `@usage` tag as well. 
http://stackoverflow.com/questions/24594507/exporting-non-s3-methods-with-dots-in-the-name-using-roxygen2-v4

db/R/query.format.vars.R
bcow: Done

db/R/query.base.R
mdietze: probably fine for now, but the sooner we can migrate names the less of a problem this will be.

PR #1184, opened 2016-11-17
Allometry: location info

mdietze:
Just adding the location information table that goes with the Jenkins allometry tables, per request from PalEON


ashiklom:
Looks good.

PR #1186, opened 2016-11-18
Cleanup, Reorganizing & Renaming in Benchmarking

bcow:
## Description
1. I reorganized functions so that things finally fit the definition / evaluation workflow that we planned. I propose a new function that replaces "create.benchmark" called "define_benchmark" which contains all the definition functions. 

2. I renamed all the functions in the benchmarking package to replace `.` by `_`. This resolved all the problems I was having with documentation. @istfer: I made sure to replace new names in the necessary pda functions but I didn't run any tests to verify that it all still works. 

3. More code cleanup and explicit package refs.  

## Types of changes

## Checklist:


bcow:
@istfer these are the PDA related changes:
https://github.com/PecanProject/pecan/pull/1186/files#diff-a6b3d204b6953798265c0458ebbab2b5
https://github.com/PecanProject/pecan/pull/1186/files#diff-c30742e852727636a80d3cc3c311f8a1

istfer:
thanks @bcow changes look good, I have some other small fixes for PDA regarding your previous changes (e.g. `query.format.vars` argument order change etc.). after this is pulled I can test some more and submit changes all together (if I find any regarding this PR)

modules/benchmark/R/calc_benchmark.R
```diff
@@ -18,24 +18,50 @@ calc.benchmark <- function(settings, bety) {
...
+  # Update benchmarks_ensembles and benchmarks_ensembles_scores tables
+  
+  ensemble <- tbl(bety,'ensembles') %>% filter(workflow_id == settings$workflow$id) %>% collect()
+  
+  # Retrieve/create benchmark ensemble database record
+  bm.ensemble <- tbl(bety,'benchmarks_ensembles') %>% 
+    filter(reference_run_id == settings$benchmarking$reference_run_id) %>% 
+    filter(ensemble_id == ensemble$id) %>% 
+    filter(model_id == settings$model$id) %>%
+    collect()
```
mdietze: Question I've been wondering about this syntax... Does this literally grab the entirety of the database table into R and then do the filtering in R, or is it using these function calls to build the SQL query that it then executes once?   If the former, this seems really inefficient. If the latter, do the rest of you really find this syntax easier to follow than SQL? I find it much harder to understand.  Not requesting a change (yet), just trying to understand memory impacts and get a feel for how this works.

modules/benchmark/R/calc_benchmark.R
```diff
@@ -18,24 +18,50 @@ calc.benchmark <- function(settings, bety) {
...
   # How are we dealing with ensemble runs? Right now I've hardcoded to select the first run.
   
-  var.ids <- as.numeric(unname(unlist(settings$benchmark$variables)))
-  metric.ids <- as.numeric(unname(unlist(settings$benchmark$metrics)))
-  
   # All benchmarking records for the given benchmarking ensemble id
   bms <- tbl(bety,'benchmarks') %>% rename(benchmark_id = id) %>%  
     left_join(.,tbl(bety, "benchmarks_benchmarks_reference_runs"), by="benchmark_id") %>% 
-    filter(reference_run_id == settings$benchmark$reference_run_id) %>% 
+    filter(reference_run_id == settings$benchmarking$reference_run_id) %>%
```
mdietze: If you've changed the XML tags, make sure to update the Gitbook!

web/workflow.R
```diff
@@ -145,8 +145,8 @@ if ('state.data.assimilation' %in% names(settings)) {
 # Run benchmarking
 if("benchmark" %in% names(settings)){
   status.start("BENCHMARKING")
-  settings <- papply(settings, function(x) create.benchmark(x, bety))
-  results <- papply(settings, function(x) calc.benchmark(x, bety))
+  settings <- papply(settings, function(x) create_benchmark(x, bety))
```
mdietze: 1) There is no create_benchmark function, so this will throw an error
2) I don't think you want to be calling define_benchmark here because the benchmarks should already be defined before workflow.R is run

modules/benchmark/R/calc_benchmark.R
bpbond: It actually builds an SQL query behind the scenes.

bpbond: I'd rewrite this as
```R
+    filter(reference_run_id == settings$benchmarking$reference_run_id,
                ensemble_id == ensemble$id,
                model_id == settings$model$id) %>%
+    collect()
```

bcow:
Everything is updated except the gitbook - working on that now!

PR #1187, opened 2016-11-18
FATES parameters

mdietze:
Flushing out list of parameters for write.configs.FATES. I've now covered everything in the NGEE-Tropics FATES variable list https://docs.google.com/spreadsheets/d/1LdMODbZOKtYZR5JYMdNLC90X9qDlKY9a56Zj5ojM44w/edit#gid=556525632 except the LITTERCLASS indexed variables. This includes PFT indexed variables and scalars (which are mostly soil BGC or SPITFIRE, which should be added as two additional pseudo-PFTS).

In parallel, a large number of variables were added to BETY.


models/fates/R/write.configs.FATES.R
```diff
@@ -203,10 +242,404 @@ write.config.FATES <- function(defaults, trait.values, settings, run.id){
...
+       }
+       if(var == "CWD_frac3"){
+         ncvar_put(nc=param.nc, varid='CWD_frac', start = 3, count = 1,
+                   vals=pft[v])
+       }
+       if(var == "CWD_frac4"){
+         ncvar_put(nc=param.nc, varid='CWD_frac', start = 4, count = 1,
+                   vals=pft[v])
+       }
+
```
ashiklom: As a temporary hack, I guess this is OK, but this violates the hell out of the "Don't Repeat Yourself" (DRY) principle and is super exposed to bugs. It would be much better to define a `list` or `data.frame` of trait-specific arguments and then have a single function that does `do.call` on `ncvar_put`.

ashiklom: This doesn't necessarily have to hold up the PR, but if we leave it as is, you should open an issue for revising it.

ashiklom: Setting parameters in general. As far as non-general unit conversions, I can envision a system where you store the unit conversion function as part of the list/data.frame object.

Like I said, I don't want to hold up this pull request over this, so we can move this discussion over to an issue and I'll mark the PR as good to go.

PR #1188, opened 2016-11-19
Bug fix to met.convert, update to tropical PFT

mdietze:
* found a small bug in met.convert
* updated the code I used to create the FATES tropical PFT to add more species


db/inst/import-ForestGEO/import_species_list.R
```diff
@@ -37,7 +37,7 @@ spp <- read.csv(sppFile,stringsAsFactors=FALSE,strip.white = TRUE)
 spp <- spp[-duplicated(spp$Species),]
 
 ## open database connection & grab current species list
-dbparms.psql <- list(driver="PostgreSQL" , user = "bety", dbname = "bety", password='bety')
+dbparms.psql <- list(driver="PostgreSQL" , user = "bety", dbname = "bety", password='bety',host='128.197.230.32')
```
serbinsh: @mdietze do you see this working for multiple dbs? Or would end users just all point to this one host for the forestGEO info?

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
     ncatt_get <- ncdf4::ncatt_get
     ncvar_add <- ncdf4::ncvar_add
-    misc.convert <- PEcAn.utils::misc.convert # unit conversions
+#    misc.convert <- PEcAn.utils::misc.convert # unit conversions
     logger.info <- PEcAn.utils::logger.info
     logger.severe <- PEcAn.utils::logger.severe
     logger.warn <- PEcAn.utils::logger.warn
+    
+    #        var_update("AR","AutoResp","kgC m-2 s-1")
+    var_update <- function(ncout,oldname,newname,newunits=NULL){
```
serbinsh: Looks good

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
     ## Get files and years
     files <- dir(outdir, "*clm2.h0.*.nc", full.names = TRUE)
     file.dates <- as.Date(sub(".nc", "", sub(".*clm2.h0.", "", files)))
-    file.years <- lubridate::year(file.dates)
-    num.years <- length(unique(file.years))
-    simulation.days <- as.Date(start_date):as.Date(end_date)
-    simulation.years <- as.numeric(unique(strftime(as.Date(simulation.days, origin = "1970-01-01"), "%Y")))
-    
-    # nescessary?  A check that output years and selected run years match
-    if (!all(file.years==simulation.years)) {
```
serbinsh: This was a check just in case the output file names didn't match the xml run years.....but I guess this would never happen

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
-        ysel <- which(file.years == year)  ## subselect files for selected year
-#        ysel <- sort(file.dates[ysel])  ## double check dates are in order
+    for (year in unique(years)) {
+        ysel <- which(years == year)  ## subselect files for selected year
         if (length(ysel) > 1) {
             logger.warn("PEcAn.FATES::model2netcdf.FATES does not currently support multiple files per year")
         }
         
-        # !!useful or do we not want this behavior? !!  Or some sort of clobber flag?
-        if (file.exists(file.path(outdir, paste(year, "nc", sep = ".")))) {
```
serbinsh: In case we wanted to skip for any output already converted, but year I wasn't sure if this is what we wanted the behavior to be.  In this case we always re-create the output.

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
         oname <- file.path(dirname(fname), paste0(year, ".nc"))
         logger.info(paste("model2netcdf.FATES - Converting:",  fname, "to", oname))
-        nc <- ncdf4::nc_open(fname, write = TRUE)
+        ncin <- ncdf4::nc_open(fname, write = TRUE)
         
         ## FATES time is in multiple columns, create 'time'
-        day  <- ncdf4::ncvar_get(nc, "mdcur")   #current day (from base day)
-        sec  <- ncdf4::ncvar_get(nc, "mscur")   #current seconds of current day
+        day  <- ncdf4::ncvar_get(ncin, "mdcur")   #current day (from base day)
+        sec  <- ncdf4::ncvar_get(ncin, "mscur")   #current seconds of current day
```
serbinsh: good idea to change nc to ncin, more descriptive

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
+        sitelon <- ncvar_get(ncin,"lon")
         ## time variable based on internal calc, nc$dim$time is the FATES output time
         t <- ncdim_def(name = "time", units = paste0("days since ", year, "-01-01 00:00:00"),
-                       vals = time, calendar = "standard", 
+                       vals = as.vector(time), calendar = "standard", 
                        unlim = TRUE)  # a direct analog of internal FATES output dim "time"
         lat <- ncdim_def("lat", "degrees_north", vals = as.numeric(sitelat), longname = "coordinate_latitude")
-        ##!! default is degrees east except we always input as degrees west, need to make sure this is sorted for all
-        ## pecan functions!!
         lon <- ncdim_def("lon", "degrees_east", vals = as.numeric(sitelon), longname = "coordinate_longitude")
```
serbinsh: @mdietze not sure if this comment was clear but I noticed that we often read in the long in -XXX.XXXX i.e. degrees west but we always save as degrees east and I am not sure we are converting properly in all of the model2netcdf codes

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
+        xyt <- list(lon, lat, t)
```
serbinsh: OK so here you define the xyT dims.  But what about Z (i.e. depth)?

db/inst/import-ForestGEO/import_species_list.R
mdietze: 1) No one should ever have to run this again, the species are now in the database. This is just a record of how they were added.
2) If you did want to run this again, you would just need to set the host to your own database

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
+#    misc.convert <- PEcAn.utils::misc.convert # unit conversions
     logger.info <- PEcAn.utils::logger.info
     logger.severe <- PEcAn.utils::logger.severe
     logger.warn <- PEcAn.utils::logger.warn
+    
+    #        var_update("AR","AutoResp","kgC m-2 s-1")
+    var_update <- function(ncout,oldname,newname,newunits=NULL){
+      oldunits <- ncdf4::ncatt_get(ncin,oldname,"units")$value
+      if(is.null(newunits)) newunits = oldunits
+      newvar <- ncdf4::ncvar_def(name = newname, units = newunits, dim = xyt)
```
mdietze: Haven't encountered any z dimensions yet, but since this is just a conversion of existing data it would be straightforward to (A) add more dimensions during the definition stage and (B) autodetect the dimensions of the variable being converted.

models/fates/R/model2netcdf.FATES.R
```diff
@@ -11,291 +11,105 @@
...
+      dat.new <- misc.convert(dat,oldunits,newunits)
+      ncdf4::ncvar_put(ncout,newname,array(dat.new,c(1,1,nt)))
+      return(ncout)
     }
-
+    
+    
     ## Get files and years
     files <- dir(outdir, "*clm2.h0.*.nc", full.names = TRUE)
     file.dates <- as.Date(sub(".nc", "", sub(".*clm2.h0.", "", files)))
```
mdietze: Right now they always have this name.

models/fates/R/model2netcdf.FATES.R
mdietze: Yeah, I dropped dates and lat/lon from the args since they really aren't necessary -- I'm assuming that if there's output in the folder from the run, we want to convert it.

models/fates/R/model2netcdf.FATES.R
mdietze: as above, would be easy to add when needed.

models/fates/R/model2netcdf.FATES.R
serbinsh: OK.  may be something we have to look at later if this changes.  But we would know so it would be an easy update

models/fates/R/model2netcdf.FATES.R
serbinsh: Got it.  I followed the syntax for the other model2netcdf's but I agree it isn't really necessary.  What it did do was enable the checks I had but again as I had commented, they weren't necessary nor perhaps the behavior we want in the code.

models/fates/R/model2netcdf.FATES.R
serbinsh: Yup. I like what you have done

PR #1189, opened 2016-11-20
Changes to build emulator on error statistics

istfer:
## Description
Changed the PDA code to calculate likelihood from error statistics and build emulator on error statistics. Now there is one emulator per input variable.

However transition is not complete yet, things left to resolve are:

- How to save things in the DB: Should I now save error statistics, but how? Should I use the likelihood table? or save likelihood for now, as before?
- Where should I save the output files (e.g. mcmc diagnostics plot) of the non-model parameters? Should I open another folder under PFT folder (e.g. `RUNID/pft/parameters`), or put them together with one of the PFT folders or to the main run folder where `pecan.xml` etc are? 
- I haven't updated the code for the other GP option (`kernlab` package) yet, but I was thinking of removing it altogether as I never use it, @mdietze do you want me to keep it? 
- I only changed the bruteforce algorithms such that they don't break with the new changes

## Types of changes

## Checklist:


mdietze:
> How to save things in the DB: Should I now save error statistics, but how? Should I use the likelihood table? or save likelihood for now, as before?

You should use the sufficient statistic in the likelihood table. The key part is that we need to be able to rebuild the emulator exclusively from the likelihood table

> I haven't updated the code for the other GP option (kernlab package) yet, but I was thinking of removing it altogether as I never use it, @mdietze do you want me to keep it?

I can't recall at the moment why we kept kernlab since we've been long convinced it's not doing what we need. I'm fine with breaking the backward compatibility at this point

> I only changed the bruteforce algorithms such that they don't break with the new changes

That should be fine -- since they don't use the emulator they shouldn't require much changing.

istfer:
@mdietze the SS values calculated for `SoilResp` are very small, such as `3.822417e-15`, `4.652113e-15`, it becomes `0` when I insert into DB. The (pecan) units are in `kg C m-2 s-1`, should I change to `umol C m-2 s-1` as units for SoilResp? Or what else can I do?

mdietze:
log(SS)?

istfer:
Oh of course, I was confused because I already logged the other ss fir NEE and LE but not SoilResp, thanks

modules/assim.batch/R/pda.mcmc.bs.R
```diff
@@ -188,9 +191,25 @@ pda.mcmc.bs <- function(settings, params.id = NULL, param.names = NULL, prior.id
...
+          isbias <- which(unlist(any.mgauss) == "multipGauss")
+          # testing now
+          nbias <- 1
+          bias.list <- return.bias(isbias, list(model.out), inputs, prior.list, nbias)
+          bias.terms <- bias.list$bias.params
+        } else {
+          bias.terms <- matrix(1, nrow = 1, ncol = 1) # just 1 for Gaussian
+        }
         
+        ## calculate error statistics
```
istfer: here I just work with optimum bias, should I have another bias specific propose/reject loop here (like in emulator where I can calculate SS for >1 bias per parameter set)?
```
chain
    iteration
        propose.parameters
        run.model
        get.model.output
            propose.bias (for a predefined number of times?)
            pda.calc.error
            pda.calc.llik.par
            pda.calc.llik
            accept/reject bias (store 1 bias)
        accept/reject model parameters

```

mdietze: Ok, so within the brute-force MCMC that's not required, but it is an option if you want to accelerate convergence. Pseudo-code is more like:
```
chain
  iteration
     propose.parameters | bias, variances
        run.model
        get.model.output
            pda.calc.error
            pda.calc.llik.par
            pda.calc.llik
    accept/reject model parameters

    propose bias | parameters, variances
      pda.calc.*
    accept/reject bias

    gibbs sample variances | parameters, bias
```
where you are definitely free to iterate over the bias and variance sampling multiple times (or equivalently, to skip model parameter proposals), you'd just need to decide whether you store all the bias and variance samples or just the 'final' iteration (this is just a question of thinning -- you technically are allowed to store all the samples, but you may choose not to)

modules/assim.batch/R/pda.define.llik.R
```diff
@@ -81,12 +83,33 @@ pda.calc.error <-function(settings, con, model_out, run.id, inputs, bias.terms){
...
+  if (!is.null(con)) {
+
+    # BETY requires sufficient statistics to be associated with inputs, so only proceed
+    # for inputs with valid input ID (i.e., not the -1 dummy id).
+    # Note that analyses requiring sufficient statistics to be stored therefore require
+    # inputs to be registered in BETY first.
+    db.input.ind <- which(sapply(inputs, function(x) x$input.id) != -1)
+    for (k in db.input.ind) {
+      
+      db.query(
```
mdietze: Correct

modules/assim.batch/R/pda.postprocess.R
```diff
@@ -0,0 +1,192 @@
...
+    pft.id <- db.query(paste0("SELECT pfts.id FROM pfts, modeltypes WHERE pfts.name='",
+                              settings$pfts[[i]]$name, 
+                              "' and pfts.modeltype_id=modeltypes.id and modeltypes.name='", 
+                              settings$model$type, "'"), 
+                       con)[["id"]]
+    
+    db.query(paste0("INSERT INTO posteriors (pft_id, created_at, updated_at) VALUES (",
+                    pft.id, ", '", now, "', '", now, "')"), con)
+    
+    posteriorid <- db.query(paste0("SELECT id FROM posteriors WHERE pft_id=", pft.id, " AND created_at='",
```
bcow: More instances where you can leave `created_at` & `updated_at` blank.

modules/assim.batch/R/pda.postprocess.R
```diff
@@ -0,0 +1,192 @@
+##' Postprocessing for PDA Results
+##'
+##' @title Postprocessing for PDA Results
+##' @param all params are the identically named variables in pda.mcmc / pda.emulator
+##'
+##' @return PEcAn settings list, updated with <params.id> pointing to the new params file.
+##'
+##' @author Ryan Kelly, Istem Fer
+##' @export
+pda.postprocess <- function(settings, con, mcmc.param.list, pname, prior, prior.ind) {
```
bcow: Ultimately, it would be good to assume that functions should take in `bety` as an argument instead of `con`. Then below in the code you can use the `dplyr` functions to do queries. (Though I do see that there's also insert which would use `bety$con`)

modules/assim.batch/R/pda.emulator.R
```diff
@@ -55,10 +64,10 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
     con <- NULL
   }
 
-  bety <- PEcAn.visualization::betyConnect("~/pecan/web/config.php")
+  bety <- betyConnect("~/pecan/web/config.php")
```
mdietze: Definitely!  I suspect @istfer was just copying the design pattern from elsewhere, but this could easily fail if someone installs pecan somewhere other than HOME

istfer: yeah, I'm still not used to the new connection.

@bcow so I just do
```
bety <- src_postgres(dbname = settings$database$bety$dbname, 
               host = settings$database$bety$host, 
               user = settings$database$bety$user, 
               password = settings$database$bety$password)
```
and don't actually use `betyConnect` function? and when I pass this `bety` to the `query.format.vars` it will work?

mdietze: Asking everyone to use this syntax everywhere is clunky and inefficient.  I'd recommend that we modify our existing db connection functions to work with the new dplyr connection. For example, betyConnect could  be overloaded such that it does it's current behavior is passed a string (path) or it does the above if passed a list (settings$database$bety). Or we could overload our existing db.open to accept a path to a config.php. Either way, we need to get down to one connection function and no manual database opens

modules/assim.batch/R/pda.bayesian.tools.R
```diff
@@ -128,8 +131,26 @@ pda.bayesian.tools <- function(settings, params.id = NULL, param.names = NULL, p
     ## Read model outputs
     model.out <- pda.get.model.output(settings, run.id, bety, inputs)
     
-    ## calculate and return likelihood
-    pda.calc.llik(settings, con, model.out, run.id, inputs, llik.fn)
+    # handle bias parameters if multiplicative Gaussian is listed in the likelihoods
+    if(any(unlist(any.mgauss) == "multipGauss")) {
+      isbias <- which(unlist(any.mgauss) == "multipGauss")
```
mdietze: Down the road, make sure to generalize this so that you can handle likelihoods with extra parameters automatically, rather than having to hack special cases in everywhere.

modules/assim.batch/R/pda.define.llik.R
```diff
@@ -5,98 +5,169 @@
...
+  neff <- n.vec * weights
   
   return(LL.total)
 } # pda.calc.llik
+
+
+
+##' @title pda.calc.llik.par
+##' @author Istem Fer
+##' @export
```
mdietze: don't forget to document functions

modules/assim.batch/R/pda.emulator.R
```diff
@@ -325,10 +351,14 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
...
+  dcores <- parallel::detectCores() - 1
+  ncores <- min(dcores, settings$assim.batch$chain)
+  cl <- parallel::makeCluster(ncores, type="FORK")
+  
   ## Sample posterior from emulator
-  mcmc.out <- lapply(1:settings$assim.batch$chain, function(chain) {
-    mcmc.GP(gp          = gp, ## Emulator
-            pckg        = pckg, ## flag to determine which predict method to use
+  mcmc.out <- parallel::parLapply(cl, 1:settings$assim.batch$chain, function(chain) {
+    mcmc.GP(gp          = gp, ## Emulator(s)
```
mdietze: 👍  I like the use of parallelization

modules/assim.batch/inst/llik.params.csv
```diff
@@ -0,0 +1,7 @@
+"likelihood","parameter","distn","parama","paramb"
+"Gaussian","sd","gamma",0.001,0.001
+"multipGauss","sd","gamma",0.001,0.001
+"multipGauss","bias","normal",0.001,0.001
+"Laplace","alpha1","norm",0.2,0.01
+"Laplace","beta1","norm",-0.2,0.01
```
mdietze: Are these just the default priors for different likelihoods?  Is this how the user changes priors or can you do that via the settings?

modules/assim.batch/R/pda.postprocess.R
istfer: @mdietze @bcow in this bit of code I'm inserting a posterior to the DB and need the `id` of the inserted posterior, if I don't insert it with `created_at`, how can I retrieve it back here?

currently I was doing this:
`     posteriorid <- db.query(paste0("SELECT id FROM posteriors WHERE pft_id=", pft.id, " AND created_at='",  now, "'"), con)[["id"]]`

I don't see any other column in the posterior table. I can filter for the latest created posterior but would that be prone to errors?

modules/emulator/NAMESPACE
```diff
@@ -33,4 +31,6 @@ export(minimize.GP)
 export(mvjump)
 export(nderiv)
 export(p)
+export(plot.jump)
+export(plot.mvjump)
```
mdietze: ok, so for most of our functions we've been moving from using a dot to using an underscore in order to avoid confusion with S3 methods. But in this case I think jump and mvjump actually are objects and thus we DO want these to be S3

modules/assim.batch/inst/llik.params.csv
istfer: this was my first thought to handle likelihood parameters, but now I don't use this file at all. heteroskedastic Laplacian parameters (`alpha`s `beta`s) are not being sampled, `bias` prior is constructed on the fly, `tau` priors are hardcoded

`tau <- rgamma(1, 0.001 + n[k]/2, 0.001 + error.stats[k]/2)`

modules/assim.batch/R/pda.emulator.R
istfer: what's the final verdict on this?

I can do
```
bety <- src_postgres(dbname = settings$database$bety$dbname, 
               host = settings$database$bety$host, 
               user = settings$database$bety$user, 
               password = settings$database$bety$password)
```
for now and open an issue, so that @bcow can look later?

bcow: Yes for now don't use the betyConnect function - I'll go back and look at it after AGU.  You should just be able to set bety to what I had above and have it work. If not, let me know and I'll fight with it.

istfer:
updates other algorithms as well

definitely there are things clean up and improve, but all PDA algorithms should work with the new changes now

modules/assim.batch/R/pda.postprocess.R
istfer: this is really cool!

Issue #1197, opened 2016-11-25
BUG from running SIPNET-Bartlett 2005-2006 Ameriflux


tonygardella:
End of Log:
```
2016-11-25 09:13:57 INFO   [convert.input] : 
   file host mimetype formatname startdate enddate dbfile.name 
          id site_id notes          created_at          updated_at start_date
1 1000000385     796       2015-05-06 03:53:25 2015-05-06 03:53:25 2001-01-01
2 1000000394     796       2015-05-06 05:37:12 2015-05-06 05:37:12 2001-01-01
3 1000000709     796       2016-02-04 08:22:59 2016-02-04 08:22:59 2004-01-01
4 1000000384     796       2015-05-06 02:22:14 2015-05-06 02:22:14 2001-01-01
5 1000000924     796       2016-08-25 11:49:22 2016-08-25 11:49:22 2006-01-01
    end_date                 name parent_id user_id access_level raw format_id
1 2006-12-31 Ameriflux_site_0-796        NA      NA            4  NA        38
2 2009-12-31 Ameriflux_site_0-796        NA      NA            4  NA        38
3 2004-12-31 Ameriflux_site_0-796        NA      NA            4  NA        38
4 2008-12-31 Ameriflux_site_0-796        NA      NA            4  NA        38
5 2006-12-31 Ameriflux_site_0-796        NA      NA            4  NA        38
2016-11-25 09:13:57 ERROR  [dbfile.input.insert] : 
   Duplicate inputs (in terms of site_id, name, and format_id) with 
   differing start/end dates are not allowed. The existing input record 
   printed above would conflict with the one to be inserted, which has 
   requested start/end dates of 2005-01-01/2006-12-31Please resolve this 
   conflict or set allow.conflicting.dates=TRUE if you want to allow 
   multiple input records with different dates.
```

## Description
```allow.conflicting.dates``` should be set to ```TRUE``` to resolve this bug.

## Context
An existing Input record exists, but the dates do match in this case. 

## Possible Implementation
Since this is occurring in the raw met download stage and there is an existing input record at that site, format, and machine that should have been updated to include the run dates, ```allow.conflicting.dates``` will be set to ```TRUE``` conditional on the chunk of code in ```convert.input``` that updates an existing input record.

tonygardella:
[...]

PR #1199, opened 2016-11-25
Add allow.conflicting.dates settings in convert.inputs

tonygardella:
Allow there to be conflicting dates in dbfile.input.insert if exact.dates are not needed. 

## Description
There was an error where valid inputs with different dates were found but triggered an error because they did not have the correct file record associated. This will allow convert.input to create the correct input record regardless of date with the correct site, format, and machine and associate the correct dbfile record with it. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) #1197 

## Checklist:


mdietze:
The issue Istem encountered of dates of existing met on other machines, and the fix of allowing conflicting dates, both need more thought.  Specifically, the whole point of the Inputs-DBfiles separation is to allow the same files to exist across multiple machines. This failure suggests that at least part of the problem comes from met.process not fully accounting for this. A quick fix that leads to redundant Inputs isn't the solution we're looking for.

tonygardella:
That makes sense. I took a look at it and I think this can be handled in convert.inputs. Instead of saying  there isn't a valid input when there is not a local file associated with that input, I'll have it get that input id and then dbfile.insert the new file and associate it with that input.

tonygardella:
latest change allows convert.input to find the correct input.record even if associated files are on a different machine. Next I need to sort out the download part, where I'll download all dates within that input record, and insert the path into the existing input record.

utils/R/convert.input.R
```diff
@@ -401,7 +415,7 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
       site.id <- input.args$newsite
     }
     
-    
+    if(exact.dates){allow.conflicting.dates = FALSE}
```
jsimkins2: if (exact.dates) {
     allow.conflicting.dates = FALSE
   }

utils/R/convert.input.R
```diff
@@ -173,10 +187,10 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
                    (end_date <= existing.input$end_date)) {
           
         
-            # There's an existing input that spans desired start/end dates. Use that one.
+         if (insert.new.file <- FALSE){ # There's an existing input that spans desired start/end dates. Use that one.
```
mdietze: Pretty sure you meant to use == here

utils/R/convert.input.R
```diff
@@ -8,7 +8,7 @@
 
 convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, start_date, 
                           end_date, pkg, fcn, con = con, host, browndog, write = TRUE, 
-                          format.vars, overwrite = FALSE, exact.dates = FALSE, ...) {
+                          format.vars, overwrite = FALSE, exact.dates = FALSE, allow.conflicting.dates = TRUE, ...) {
```
mdietze: I'm missing what allow.conflicting.dates actually means. By name, it seems to just be the opposite of exact.dates, but neither is documented

utils/R/convert.input.R
```diff
@@ -402,18 +428,33 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
...
+                                      in.prefix = result$dbfile.name[1], 
+                                      siteid = site.id, 
+                                      startdate = start_date,
+                                      nddate = end_date, 
+                                      mimetype, 
+                                      formatname, 
+                                      parentid = parent.id,
+                                      con = con, 
+                                      hostname = machine$hostname,
+                                      allow.conflicting.dates = allow.conflicting.dates)
```
serbinsh: @tonygardella I tried out the logic for conflicting dates as you have now (i was getting errors) and it seems to be working!

tonygardella: Cool!  but there's some cases where it doesn't still so I need to keep fixing and testing.

mdietze: If the current PR fixes some cases and doesn't break current cases I'd say we move ahead and then do another PR with the next fix. Important to keep this moving forward with AGU so close

db/man/dbfile.check.Rd
```diff
@@ -2,10 +2,13 @@
 % Please edit documentation in R/dbfiles.R
 \name{dbfile.check}
 \alias{dbfile.check}
-\alias{dbfile.input.check}
-\alias{dbfile.posterior.check}
-\title{Check for a file in the input/dbfiles tables}
+\title{Check for a file in the dbfiles tables}
 \usage{
+<<<<<<< HEAD
```
robkooper: remove conflict

modules/data.atmosphere/R/met.process.R
```diff
@@ -137,7 +137,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
   if (stage$download.raw) {
     raw.data.site.id <- ifelse(is.null(register$siteid), new.site$id, register$siteid)
     
-    raw.id <- PEcAn.data.atmosphere:::.download.raw.met.module(dir = dir,
+    raw.id <- .download.raw.met.module(dir = dir,
```
serbinsh: @tonygardella did you mean to delete "PEcAn.data.atmosphere::" here?

tonygardella: Yep I had added that accidentally as I was testing individual scripts. Originally it was  just ```.download.raw.met.module```.

db/R/dbfiles.R
```diff
@@ -253,7 +254,7 @@ dbfile.posterior.insert <- function(filename, pft, mimetype, formatname, con, ho
 ##'
 ##' This will check the dbfiles, inputs, machines and formats tables to see if the
 ##' file exists
-##' @name dbfile.check
+##' @name dbfile.posterior.check
```
mdietze: name conflict still exists.

modules/data.atmosphere/R/met.process.R
serbinsh: Yeah I can see that now, thanks @mdietze.  Makes sense that it is in the lib so don't need to link to the lib

db/R/dbfiles.R
tonygardella: The @name on Line 257 Goes with function on line 270 which is dbfile.posterior.check. There was a mismatch earlier that stuck ```dbfile.posterior.check``` and ```dbfile.input.insert``` along with ```dbfile.check``` in the same .Rd file. This changes it so that each function has their own .Rd file. If this is how it's supposed to be I can change it back.

mdietze: Sorry, you're OK. Didn't realize the amount of code Github was hiding from me between this line and the next chunk

utils/R/convert.input.R
```diff
@@ -130,6 +134,9 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
         
         existing.input <- db.query(paste0("SELECT * FROM inputs WHERE id=", existing.dbfile[["container_id"]]),con)
         
+
```
robkooper: should probably make sure we don't have excessive new lines here.

PR #1201, opened 2016-11-26
Simple script to estimate AGB timeseries at BCI.

mdietze:
Resulting csv file was added to Inputs and linked to BCI benchmarks (see other PR). Starting point for extending stats and metrics for ForestGEO plots. Contains some hacks, but needs follow up to refine and also to figure out why second and 3rd census give somewhat erratic values


modules/data.land/inst/BCI_AGB.R
```diff
@@ -0,0 +1,54 @@
+## Barro Colorado Island AGB timeseries
+library(dplyr)
+indir <- "~/Dropbox/Desktop/Projects/BCI/"
```
bcow: Should this be an argument to a function as opposed to hardcoded?

mdietze: Vonceivably, but right now this is just documenting how this Input was generated, and would take more work to generalize to something that could be applied to multiple sites. Eventual goal, but for now just want this archived.

PR #1202, opened 2016-11-27
Calling BDFiddle from load_data to perform file conversion through Brown Dog

jam2767:
Making changes to load_data to convert files from mimetypes pecan doesn't recognize, like .xlsx to something usable, starting with .csv. Calling BD API through BDFiddle. Converted file is not currently being returned. Ongoing work, creating PR for @mdietze and others to review and advise. Relevant commits on Nov. 26.


db/R/dbfiles.R
```diff
@@ -39,31 +39,31 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
   
   # find mimetype, if it does not exist, it will create one
   mimetypeid <- get.id("mimetypes", "type_string", mimetype, con, create=TRUE)
-  
+ 
   # find appropriate format, create if it does not exist
```
robkooper: this scares me since this will be the exact opposite logic of what we had before.

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +27,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
     fcn <- match.fun(fcn1)
   } else if (exists(fcn2)) {
     fcn <- match.fun(fcn2)
+  } else if (!exists(fcn1) & !exists(fcn2)) { 
+    #To Do: call to DAP to see if conversion to csv is possible
+    #Brown Dog API call through BDFiddle, requires username and password
+    #will be made into R package
+    install_github("yan130/bd.r")
```
robkooper: i think this should become a dependencies (or at least suggest). not have it install it in the code itself. At that point if the browndog library does not exist we can skip this if statement.

utils/R/utils.R
```diff
@@ -665,6 +665,32 @@ convert.expr <- function(expression) {
   return(list(variable.drv = deri.var, variable.eqn = list(variables = variables, expression = deri.eqn)))
 }
 
+## Convert a file using Brown Dog Conversion service
```
robkooper: since this function is commented out, i think it would be better not event to check it in.

modules/data.land/inst/meta2format.EML.R
```diff
@@ -0,0 +1,30 @@
+library(devtools)
```
robkooper: not sure if this is needed as part of this pull request.

db/R/dbfiles.R
mdietze: I'm not sure what @robkooper is referring to, as the only difference in this function is whitespace. That said, some of the changes lower down look like ones @tonygardella made, which suggests you might be out of sync with the mainline. I'd recommend a pull upstream master.

modules/benchmark/R/load_data.R
mdietze: Agree 100%. Also, you don't want to re-install the package every time you call the function.

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +27,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+    
+    key   = get_key("https://bd-api.ncsa.illinois.edu",username,password) #generic
```
mdietze: Where, exactly, are username and password coming from here?  This is a non-trivial problem since we don't want those in the settings file, where they would get archived with the run and potentially even inserted into the database via Benchmarking.  @robkooper is there an easy way to spawn a prompt to ask for the username and password interactively? Or should there be a PEcAn default (which also seems dangerous).

modules/benchmark/R/load_data.R
```diff
@@ -17,6 +17,7 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
   library(lubridate)
   library(udunits2)
   library(dplyr)
+  library(devtools)
```
mdietze: we want to be working library calls out of the system, not adding new ones

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +27,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+    #To Do: call to DAP to see if conversion to csv is possible
+    #Brown Dog API call through BDFiddle, requires username and password
+    #will be made into R package
+    install_github("yan130/bd.r")
+    
+    key   = get_key("https://bd-api.ncsa.illinois.edu",username,password) #generic
+    token = get_token("https://bd-api.ncsa.illinois.edu",key)
+    #output_path = "/fs/data3/jam2767/veg_test" where are we putting converted file?
+    convert_file(url = "https://bd-api.ncsa.illinois.edu", 
+                 input_filename = paste0(data.path, format$file_name),
```
mdietze: I think the input_filename is just data.path.  The format knows nothing about the specific file name.

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +27,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+    #Brown Dog API call through BDFiddle, requires username and password
+    #will be made into R package
+    install_github("yan130/bd.r")
+    
+    key   = get_key("https://bd-api.ncsa.illinois.edu",username,password) #generic
+    token = get_token("https://bd-api.ncsa.illinois.edu",key)
+    #output_path = "/fs/data3/jam2767/veg_test" where are we putting converted file?
+    convert_file(url = "https://bd-api.ncsa.illinois.edu", 
+                 input_filename = paste0(data.path, format$file_name), 
+                 output = "csv", output_path, token)
```
mdietze: where is output_path coming from?

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +27,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+    #not doing anything about mimetypes not convertible by BD right now
+    fcn <- match.fun(fcn2)
```
mdietze: In addition to updating the fcn to fcn2, you also need to update the data.path to the new file returned from BD. Otherwise, when you call fcn a few lines later you'll be passing it the original, unconverted, file.

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +27,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+  } else if (!exists(fcn1) & !exists(fcn2)) { 
+    #To Do: call to DAP to see if conversion to csv is possible
+    #Brown Dog API call through BDFiddle, requires username and password
+    #will be made into R package
+    install_github("yan130/bd.r")
+    
+    key   = get_key("https://bd-api.ncsa.illinois.edu",username,password) #generic
+    token = get_token("https://bd-api.ncsa.illinois.edu",key)
+    #output_path = "/fs/data3/jam2767/veg_test" where are we putting converted file?
+    convert_file(url = "https://bd-api.ncsa.illinois.edu",
```
mdietze: you aren't assigning the output of convert_file to anything. You need to do that to get the new filename. Also, you need to add a check after this make sure that convert_file returned something. If it didn't you need to return an informative error message (make user to use logger not print).

modules/data.land/inst/meta2format.EML.R
mdietze: since it's just in inst, might as well include it so it doesn't get lost

utils/R/utils.R
mdietze: 👍 yeah, this looks like it was replaced by the new BrownDog package

db/R/dbfiles.R
jam2767: Weird. I did a pull upstream master and it said my branch was already up to date.

utils/R/utils.R
jam2767: done

modules/benchmark/R/load_data.R
jam2767: If I put bd.r in Imports while it doesn't exist as a package will the pecan.benchmark package still load, or should I put it in Suggests for now?

modules/benchmark/R/load_data.R
jam2767: removed

modules/benchmark/R/load_data.R
jam2767: Great question. I have no idea where to get them from. For testing I used my username and password from signing up for BDFiddle.

modules/benchmark/R/load_data.R
jam2767: I assigned convert_file to converted.data.path and added an error if convert_file returns "NA".

modules/benchmark/R/load_data.R
jam2767: I think it's a required input for convert_file, but I didn't know where to actually assign the output_path

modules/benchmark/R/load_data.R
jam2767: added line

db/R/dbfiles.R
```diff
@@ -147,18 +147,17 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
 ##' @param con database connection object
 ##' @param hostname the name of the host where the file is stored, this will default to the name of the current machine
 ##' @param params database connection information
-##' @param exact.dates setting to include start and end date in input query
```
robkooper: please document ignore.dates

db/R/dbfiles.R
```diff
@@ -147,18 +147,17 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
...
-##' @author Rob Kooper, Tony Gardella
+##' @author Rob Kooper
 ##' @examples
 ##' \dontrun{
 ##'   dbfile.input.check(siteid, startdate, enddate, 'application/x-RData', 'traits', dbcon)
 ##' }
 dbfile.input.check <- function(siteid, startdate=NULL, enddate=NULL, mimetype, formatname, parentid=NA, 
-                               con, hostname=fqdn(), exact.dates=FALSE) {
+                               con, hostname=fqdn(), ignore.dates=FALSE) {
   if (hostname == "localhost") hostname <- fqdn();
```
robkooper: this basically flips the logic, and goes from not needing exact dates, to needing exact dates.

jam2767: This is a mistake and I must be out of sync. I still don't understand how though because my pull from pecan master doesn't update anything.

modules/benchmark/R/load_data.R
```diff
@@ -26,13 +26,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
   } else if (exists(fcn2)) {
     fcn <- match.fun(fcn2)
+  } else if (!exists(fcn1) & !exists(fcn2)) { 
+    #To Do: call to DAP to see if conversion to csv is possible
+    #Brown Dog API call through BDFiddle, requires username and password
+    key   <- get_key("https://bd-api.ncsa.illinois.edu",username,password)
+    token <- get_token("https://bd-api.ncsa.illinois.edu",key)
+    #output_path = "/fs/data3/jam2767/veg_test" where are we putting converted file?
+    converted.data.path <- convert_file(url = "https://bd-api.ncsa.illinois.edu", input_filename = data.path, 
+                 output = "csv", output_path, token)
```
mdietze: note from email with Yan, I had output and output path flipped in order in my code, which I sent you. You should probably refer to all args by name to be safe.

Also, all the BrownDog R function calls are missing their namespace, which you'll want to add.

db/R/dbfiles.R
mdietze: ignore.dates was part of @tonygardella's PR. This again looks like something that's reverting to before his updates.  @jam2767 given all the problems you're having and the relatively tiny bits of code that you're adding, you might just want to start with a fresh clone of PEcAn, create a branch specific for this work, re-add your bits, and then push it to your github.

db/R/query.format.vars.R
```diff
@@ -40,7 +40,7 @@ query.format.vars <- function(bety,input.id=NA,format.id=NA,var.ids=NA){
   # get variable names and units of input data
   fv <- db.query(paste("SELECT variable_id,name,unit,storage_type,column_number from formats_variables where format_id = ", f$id),con)
   
-  if(all(!is.na(var.ids))){
+  if(!is.na(var.ids)){
```
mdietze: I'm assuming this change was stray as well?

db/R/dbfiles.R
jam2767: OK, fixed it so that there are only the 2 files I _actually_ meant to include in this PR.

mdietze:
fails Travis:
```
Error in parse(outFile) : 
  /home/travis/build/PecanProject/pecan/modules/benchmark/R/load_data.R:53:3: unexpected '}'
52:     PEcAn.utils::logger.warn(paste0("Brown Dog is currently unable to perform conversion from ",mimetype," to a PEcAn usable format")
53:   }
```

modules/benchmark/R/load_data.R
```diff
@@ -34,13 +34,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
     fcn <- match.fun(fcn1)
   } else if (exists(fcn2)) {
     fcn <- match.fun(fcn2)
+  } else if (!exists(fcn1) & !exists(fcn2)) {
```
robkooper: Maybe add a `& require(bd)` here. This will return FALSE if the library is not found and it will skip this section. This allows us to make the browndog library be a suggestion and it will work wether or not somebody has it installed.

modules/benchmark/R/load_data.R
```diff
@@ -34,13 +34,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+                                        output = "csv", output_path, token)
+    if (is.na(converted.data.path)){
+      PEcAn.utils::logger.error("Converted file was not returned from Brown Dog")
+    }
+    #not doing anything about mimetypes not convertible by BD right now
+    fcn <- match.fun(fcn2)
+    data.path <- converted.data.path
   } else {
-    PEcAn.utils::logger.warn(paste0("no load_data for current mimetype ",mimetype," - converting using browndog"))
+    PEcAn.utils::logger.warn(paste0("Brown Dog is currently unable to perform conversion from ",mimetype," to a PEcAn usable format")
```
robkooper: missing parenthesis at the end. Also as a FYI, the logger functions do not require a paste0, but do this implicitly. So you can simply do a : `logger.debug("hello", "world")` and it will print "hello world".

modules/benchmark/R/load_data.R
```diff
@@ -34,13 +34,27 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
...
+    key   <- get_key("https://bd-api.ncsa.illinois.edu",username,password)
+    token <- get_token("https://bd-api.ncsa.illinois.edu",key)
+    #output_path = where are we putting converted file?
+    converted.data.path <- convert_file(url = "https://bd-api.ncsa.illinois.edu", input_filename = data.path, 
+                                        output = "csv", output_path = output_path, token = token)
+    if (is.na(converted.data.path)){
+      PEcAn.utils::logger.error("Converted file was not returned from Brown Dog")
+    }
+    #not doing anything about mimetypes not convertible by BD right now
+    fcn <- match.fun(fcn2)
```
mdietze: Sorry for not catching this earlier, but just realized that after the conversion you set fcn to fcn2, which is the ORIGINAL mime-type function. But if we got here we know that we can't load that mime-type and so instead we converted to a different mime-type. So we need to load using that different mime-type. Specifically, this code should probably be
```
fcn <- match.fun("load_csv")
```

PR #1203, opened 2016-11-28
Data tunnel

mdietze:
Generalizes tunnel system for cases where the cluster has separate nodes for login vs data transfers. This is the case at LBNL.

Also, found I had to make some changes to remote.execute.R because the command wasn't yet using tunnels and was treating host as a string not a list.

Changes should be backward compatible

To use with manually opened tunnels, add data server's tunnel file to settings$host$data_tunnel

To use with web, add data server's hostname under the hostlist array variable 'data_hostname'

In addition to my assigned reviewer @bcow, would like @robkooper to look as well

Code is not 100% tested at the moment because I got myself kicked off the LBNL server in the process of testing (probably for getting my password wrong too many times during debugging)


utils/R/kill.tunnel.R
```diff
@@ -3,12 +3,19 @@
 ##' @title kill.tunnel
 ##' @description kill tunnel to remote machine
 ##' @author Rob Kooper
-kill.tunnel <- function() {
-    if (exists("settings") && !is.null(settings$host$tunnel)) {
+kill.tunnel <- function(settings,exe=TRUE,data=TRUE) {
```
robkooper: being picky but we should make sure the whole file is either 2 or 4 spaces for a tab, not a mixture.

utils/R/remote.R
```diff
@@ -85,11 +85,12 @@ remote.copy.from <- function(host, src, dst, delete = FALSE, stderr = FALSE) {
   if (is.localhost(host)) {
     args <- c(args, src, dst)
   } else {
-    if (!is.null(host$tunnel)) {
-      if (!file.exists(host$tunnel)) {
-        logger.severe("Could not find tunnel", host$tunnel)
+    tunnel = ifelse(is.null(host$data_tunnel),host$tunnel,host$data_tunnel)
```
robkooper: should use <-

PR #1205, opened 2016-11-29
Data tunnel part 2

mdietze:
A few small additional tweaks to data tunnel work:

ability to specify alternative location for temporary files on remote in remote.execute.R. Change is in response to /tmp not being global on LBNL cluster

requested ifelse changes from @ashiklom

changes away from R --vanilla in remote.execute.R. Also suggested by @ashiklom and required for R to be able to find R_LIB_USER on some remote systems

documenting the test script I've been using to get things working on the LBNL cluster

convert.input switch from outdir to outfolder

small tweaks to php


mdietze:
@bcow thanks for noting that config.example.php needs updating -- actually I'm pretty sure all the changes are restricted to the hostlist part, since I've only changed things related to remote, and I've got the start of that in progress for the Gitbook.

bcow:
Right - I was confusing this with the other discussion about `$dbfiles_folder=$output_folder . "/dbfiles";` and if it should be inputs instead.

mdietze:
If you mean Gitter today, it's the same discussion. In both cases we were discussing settings$host$folder vs settings$host$dbfiles, which is distinct from the local dbfiles and output_folder

PR #1208, opened 2016-11-30
Use `importFrom` in function headers instead of defining namespace shortcuts

bpbond:
Use `importFrom` in function headers instead of defining namespace shortcuts.

## Description
PEcAn code has many places where we need frequent access to a function in a loaded package, and don't want to repeatedly use a namespace qualifier (e.g. `pkg:fun`). Rather than define a shortcut in-code as I originally wrote (`fun <- pkg::fun`) it's cleaner (and faster) to use the roxygen2 `importFrom` directive for this. More info: http://r-pkgs.had.co.nz/namespace.html

## Motivation and Context
 Improves readability, elegance, speed

## Types of changes

## Checklist:


mdietze:
Thanks @bpbond!  The previous way always seems a bit hacky to us, so I'm glad to know there's a 'right' way to do that.  So presumably we need to rerun Roxygen for this to actually go into effect? Would you mind doing that quick (should just be ```make document```)

bpbond:
@mdietze I'm having trouble installing `rhdf5` which `make document` needs to run. A lot of ugly C++ errors flowing by. Can I be rude and pass the doc rebuild off onto someone else?

mdietze:
@bpbond no problem. Once I have a free minute (after the next 2hr of telecons) I'll pull your branch, build docs, and PR it back to your branch.

PR #1209, opened 2016-11-30
minor fixes after VM tests

istfer:
- Forgot the case for parallelization where detect cores can return 1.

- Also added `GPfit` package to the `Depends`, package was not installed to the VM. Do I need to put the package dependency somewhere else so that it will be installed while VM is created?

- And[ this file ](https://psql-pecan.bu.edu/bety/inputs/1000002080) needs to be in the VM I think, if we want to use it in the PDA @robkooper @mdietze?

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.batch/DESCRIPTION
+++ b/modules/assim.batch/DESCRIPTION
@@ -16,6 +16,7 @@ Depends:
     ellipse,
     dplyr,
     parallel,
+    GPfit,
     IDPmisc
 Imports:
     coda (>= 0.18),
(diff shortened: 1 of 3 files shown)


mdietze:
The soil resp data we're using is not public (at least it wasn't when we got it, it may be by now) and thus shouldn't go on the public VM. If you work up a demo, you'll want to include data other than that one for the time being. Remind me to contact coauthors about data release when we get closer to publication.

PR #1213, opened 2016-12-01
FATES BCI PDA debugging

mdietze:
FATES:
* catch non-numeric lat/lon
* try to catch missing pft names

Emulator:
* wasn't writing pft names to models

pda.utils:
* added a case to catch when posteriorid not specified in settings explicitly but available in current workflow/PFT directories via meta-analysis output. Needs a more general solution.

start.model.runs
* skip if no list of runs (runs.txt)

workflow.R
* skip ensemble and sensitivity analysis if not in settings

misc namespace and Roxygen


settings/R/check.all.settings.R
```diff
@@ -20,10 +20,10 @@ check.inputs <- function(settings) {
   }
   
   # get list of inputs associated with model type
-  dbcon <- db.open(settings$database$bety)
-  on.exit(db.close(dbcon))
+  dbcon <- PEcAn.DB::db.open(settings$database$bety)
```
bcow: For now this works (especially since I haven't fixed `betyConnect` like I said I would) but we really should be switching over to `dplyr` functions. Essentially any time we use `db.query` with `SELECT` we can replace that with the `dplyr` functions `tbl` and `filter` and `select`.  I can tackle the full transition after AGU/Christmas but the sooner people get used to the `dplyr` functions, the better.

modules/assim.batch/R/pda.utils.R
```diff
@@ -256,12 +256,23 @@ pda.load.priors <- function(settings, con, path.flag = TRUE) {
...
       
       files <- dbfile.check("Posterior", settings$pfts[[i]]$posteriorid, con, settings$host$name)
+      
       pid <- grep("post.distns.*Rdata", files$file_name)  ## is there a posterior file?
       if (length(pid) == 0) {
         pid <- grep("prior.distns.Rdata", files$file_name)  ## is there a prior file?
       }
       if (length(pid) > 0) {
         prior.paths[[i]] <- file.path(files$file_path[pid], files$file_name[pid])
+      } else {
```
istfer: as I said in the gitter, if there are posterior objects in your workflow folders but the code can't find them, that indicates a problem that needs to be solved differently. PDA should be able to run as long as valid posterior/prior objects exist in a place where the code can load them, it shouldn't matter if they're produced in the same workflow. 

This looks like a hack to me. Now the `pid` obtained above and the object that's being loaded here doesn't have to be the same. PDA would run this way, but I'm not sure if this is what we want

istfer:
```
files <- dbfile.check("Posterior", settings$pfts[[i]]$posteriorid, con, settings$host$name)
```
this line collects the file information for `settings$host$name`. Because of this, when I do a PDA run in test-pecan, I can't load posterior/prior objects that are created in pecan2, even though they're under /fs/data2 or /fs/data3 which are accessible from both test-pecan and pecan2. 

I think you're having a similar problem, your posterior/prior id inserted to the DB by another machine, and although they're probably accessible to your current workflow, code can't load it?

Unless you can think of a solution right away, I can look at how we try to load posterior/prior objects here

mdietze: I respectfully disagree. I can see how 'files' would miss posteriors if they were on a different settings$host$name, but what I said on Gitter is that settings$pfts[[i]]$posteriorid is NULL in my case, and therefore files will be NULL.  More important, the code above had no check in place to deal with the case where files == NULL. It would cause pid to be NULL and the length(pid)==0 fix would stay NULL because files == NULL. The length(pid) > 0 check would be false, and thus prior.paths[[i]] would be NULL and load(prior.paths[[i]]) would attempt to load NULL.  My fix may be a hack, but the current code does need to catch this case and deal with it.  Also, my hack is the exact same logic that's used in get.parameter.samples to sample from the posteriors for ensemble runs (hence my argument that we need a shared set of functions for resolving posterior.ids throughout the code rather than different hacks in different places).

modules/assim.batch/R/pda.utils.R
```diff
@@ -256,12 +256,23 @@ pda.load.priors <- function(settings, con, path.flag = TRUE) {
     for (i in seq_along(settings$pfts)) {
       
       files <- dbfile.check("Posterior", settings$pfts[[i]]$posteriorid, con, settings$host$name)
```
istfer: Oh wait, in addition to the problem I mention below, I think there is another bug here

this line should be 
```
files <- dbfile.check("Posterior", settings$assim.batch$prior$prior.id[[i]] con, settings$host$name)
```

PR #1216, opened 2016-12-01
Revised PDA prior handling

istfer:
Fixed the bug and cleaned the prior handling some more. I tried to test it as much as I can (this can get a little complicated, because if you want to run emulator "longer" you need to load the priors used for the previous PDA whereas if you want to refine the emulator by proposing new "round" of knots you need both the priors used for the previous PDA and the posteriors came out of the previous PDA, and these extensions can come after one another, and be run more than twice etc.)

For now I tested these cases:
- normal run
- normal run followed by a "longer" extension run
- normal run followed by a "longer" extension run twice
- normal run followed by a "round" extension run
- normal run followed by a "round" extension run and then a "longer" extension

I also added back ability to handle failed runs, but we need to decide how many runs we can allow to fail.

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.batch/R/pda.emulator.R
+++ b/modules/assim.batch/R/pda.emulator.R
@@ -20,18 +20,6 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
     n.knot <- adapt <- adj.min <- ar.target <- jvar <- NULL
   }
   
-  ## -------------------------------------- Setup ------------------------------------- 
-  ## Handle settings
-  settings <- pda.settings(
-    settings=settings, params.id=params.id, param.names=param.names, 
-    prior.id=prior.id, chain=chain, iter=iter, adapt=adapt, 
-    adj.min=adj.min, ar.target=ar.target, jvar=jvar, n.knot=n.knot)
-  
-  
-  ## will be used to check if multiplicative Gaussian is requested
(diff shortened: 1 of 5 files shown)


modules/assim.batch/R/pda.emulator.R
```diff
@@ -281,6 +272,8 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
           SS.list[[inputi]] <- cbind(X, error.statistics[[inputi]])
       } # if-block
         
+      # remove failed runs
+      SS.list[[inputi]] <- SS.list[[inputi]][!rowSums(is.na(SS.list[[inputi]])), ]
     } # for-loop
```
istfer: for now I'm completely removing all the parameter sets of the failed runs because I don't know why the model failed, it may not only be due to unlikeliness of parameter combinations, so I'm not assuming anything and removing them.

But for how many failed runs should we allow this @mdietze?

mdietze: I'd do the check as a fraction and make it user specified (settings). I think a reasonable default if it's not specified might be 50% -- if more than half the runs fail something's probably really wrong.  But I could envision cases where someone might be very strict, and other cases where someone might be proposing very broad parameter that result in many impossible combinations that would require them to be fairly lenient.

PR #1219, opened 2016-12-02
Some final(!) PDA tweaks

istfer:
Some more tweaks to the PDA code as @mdietze suggests

-Returning aligned data as well.

-PEcAn.MA namespace fix

-Attempt to write out a pda.log file for the emulator, not clean though

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.batch/R/pda.emulator.R
+++ b/modules/assim.batch/R/pda.emulator.R
@@ -408,12 +408,12 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
   dcores <- parallel::detectCores() - 1
   ncores <- min(max(dcores, 1), settings$assim.batch$chain)
   
-  logger.setOutputFile(paste0(settings$outdir, "/pda.log"))
+  logger.setOutputFile(file.path(settings$outdir, "pda.log"))
   
   current.step <- "pre-MCMC"
   save(list = ls(all.names = TRUE),envir=environment(),file=pda.restart.file)
   
-  cl <- parallel::makeCluster(ncores, type="FORK", outfile = "pda.log")
+  cl <- parallel::makeCluster(ncores, type="FORK", outfile = file.path(settings$outdir, "pda.log"))
   
   ## Sample posterior from emulator
   mcmc.out <- parallel::parLapply(cl, 1:settings$assim.batch$chain, function(chain) {
--- a/modules/assim.batch/R/pda.emulator.R
+++ b/modules/assim.batch/R/pda.emulator.R
@@ -211,8 +211,11 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
       for (i in seq_len(settings$assim.batch$n.knot)) {
         align.return <- pda.get.model.output(settings, run.ids[i], bety, inputs)
         model.out[[i]] <- align.return$model.out
+        if(!is.na(model.out[[i]])){
+          inputs <- align.return$inputs
+        } 
       }
-      inputs <- align.return$inputs
+      
       
       current.step <- "pda.get.model.output"
       save(list = ls(all.names = TRUE),envir=environment(),file=pda.restart.file)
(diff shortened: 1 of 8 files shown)


modules/assim.batch/R/pda.emulator.R
```diff
@@ -404,10 +407,14 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
   # prepare for parallelization
   dcores <- parallel::detectCores() - 1
   ncores <- min(max(dcores, 1), settings$assim.batch$chain)
-  cl <- parallel::makeCluster(ncores, type="FORK")
+  
+  logger.setOutputFile(paste0(settings$outdir, "/pda.log"))
```
ashiklom: Probably better to do this with `file.path`.

modules/assim.batch/R/pda.emulator.R
```diff
@@ -404,10 +407,14 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
...
+
```
istfer: I tried something like this @ashiklom @mdietze @robkooper 

It writes out to an outfile but leaving it empty didn't make it print-out to the console. It's too messy to be printed out to the console anyway, writes out both the jump variances and the number of iterations at each 100th step (I set it to 200 now in the code). But I think both master and the workers are writing to the file somehow because messages are being written twice per worker or I missed something. 

[pda.txt](https://github.com/PecanProject/pecan/files/627854/pda.txt)

modules/assim.batch/R/pda.emulator.R
```diff
@@ -404,10 +407,14 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
...
   dcores <- parallel::detectCores() - 1
   ncores <- min(max(dcores, 1), settings$assim.batch$chain)
-  cl <- parallel::makeCluster(ncores, type="FORK")
+  
+  logger.setOutputFile(paste0(settings$outdir, "/pda.log"))
+  
   current.step <- "pre-MCMC"
   save(list = ls(all.names = TRUE),envir=environment(),file=pda.restart.file)
   
+  cl <- parallel::makeCluster(ncores, type="FORK", outfile = "pda.log")
```
ashiklom: You're sure `pda.log` will be in the right directory? Might be better to give an absolute path.

istfer:
did the file.path changes. I guess this is as far as I'm willing to go for now for printing out emulator progress :) I can remove it altogether or leave it like this

do you know why build fails? it says `ERROR: dependency ‘PEcAn.MA’ is not available for package ‘PEcAn.assim.batch’` I put it in the `Depends`

ashiklom:
You also have to add it to the appropriate spot in the Makefile, probably. That will tell Travis to build PEcAn.MA first.

Other than that, good to go!

istfer:
something like this?
```
$(call depends,modules/assim.batch): .install/meta.analysis 
```

ashiklom:
Yup, exactly! There should already be a `$(call depends...)` for `assim.batch` in there that you can just append to.

modules/assim.batch/R/pda.emulator.R
```diff
@@ -209,8 +209,11 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
     
       ## read model outputs    
       for (i in seq_len(settings$assim.batch$n.knot)) {
-        model.out[[i]] <- pda.get.model.output(settings, run.ids[i], bety, inputs)
+        align.return <- pda.get.model.output(settings, run.ids[i], bety, inputs)
+        model.out[[i]] <- align.return$model.out
       }
+      inputs <- align.return$inputs
```
mdietze: Doesn't the update to inputs need to be inside the loop and indexed by [[i]] as well?

istfer: I thought this only needs to be done once, but I didn't think about failed runs case

mdietze: Ahh, I think I was misunderstanding and thought we were looping over data not knots. But now that you mention it it would fail if the final run had failed.

PR #1230, opened 2016-12-07
Met.bug - Making queries for existing inputs more explicit/exhaustive and cleaning up some logic

tonygardella:
Machine check needed to be better implemented and added to. It also needed to happen all the time and upon failure grab the last updated record as a tiebreaker if there are multiple valid inputs and only then return empty.

Also if a input record is being updated with new dates. We needed to return the existing input/file record pair after the download.

## Types of changes

## Checklist:

### DIFF
--- a/db/R/dbfiles.R
+++ b/db/R/dbfiles.R
@@ -156,7 +156,8 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate, 
 ##'   dbfile.input.check(siteid, startdate, enddate, 'application/x-RData', 'traits', dbcon)
 ##' }
 dbfile.input.check <- function(siteid, startdate=NULL, enddate=NULL, mimetype, formatname, parentid=NA, 
-                               con, hostname=fqdn(), exact.dates=FALSE, machine.check = FALSE) {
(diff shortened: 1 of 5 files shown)


mdietze:
Also, in terms of the other bug we discussed (met.process not handling starting from a non-terminal input.id) here's a few bits that seem relevant:

Should be able to start met.process from id not source https://github.com/PecanProject/pecan/blob/master/modules/data.atmosphere/R/met.process.R#L74-L76 (needs updating for the case where source == NULL but input.id != NULL). Should also probably use logger.severe for the case where neither is present, otherwise the workflow will continue and then fail spectacularly later on.  

This code should be catching an existing input.id but didn't for my case. May be related to (and solved by) the bug fixes you've already implemented, but that needs testing. https://github.com/PecanProject/pecan/blob/master/modules/data.atmosphere/R/met.process.R#L100-L112

tonygardella:
@mdietze Got it. I'll take a look after I fix some errors I introduced by changing dbfile.check.

tonygardella:
@mdietze Can we revert this PR until I fix dbfile.check? This has broken PEcAn all over the place.

PR #1232, opened 2016-12-07
Met.bug - Same as last one except changed dplyr usage to db.query

tonygardella:
Machine check needed to be better implemented and added to. It also needed to happen all the time and upon failure grab the last updated record as a tiebreaker if there are multiple valid inputs and only then return empty.

Also if a input record is being updated with new dates. We needed to return the existing input/file record pair after the download.
## Types of changes

## Checklist:


tonygardella:
@mdietze I'm testing the fix to read in an input.id and determine stage now, but wanted to fix the changes I made that broke everything first.

PR #1244, opened 2017-01-18
Minor.rtm.update

serbinsh:
## Types of changes

## Checklist:


serbinsh:
@ashiklom Where I put this copy command, will that run every time or just at first runtime?

If always, where could I put it so it only copies when initially creating the output dirs?  Or should I just look for the file each time and only copy if it isn't there?

ashiklom:
As it is, you will do the copy and overwrite operation every time `EDR` is called, which of course is thousands of times for an inversion.

I would suggest something like:

```r
soil_path <- paths$soil
if (is.null(soil_path)) {
    soil_path <- system.file(...)
}
if (!is.na(soil_path)) {
    file.copy(soil_path, output_dir, ...)
}
# And the same for the other path.
```

This allows you to force the copy operation to be off by passing a `paths` list where `soil = NA`, to pass a custom soil path via `soil = "/your/path",` or to use the default by leaving `soil` unset or explicitly set to `NULL`.

serbinsh:
I will make the change before we commit

ashiklom:
I implemented the changes I suggested. Haven't had a chance to test this yet, but I think it's good to go.

Issue #1191, opened 2016-11-21
Re-organize PEcAn packages to have more precise purposes


ashiklom:
I think that we currently have too few packages that each do too many things. So I wanted to start a discussion about specific functionality from certain packages that can and should be isolated. Here are some initial ideas:

* Put all `logger.*` functions to a `pecan.logger` package
* Create a small set of date-related functions, and put them in a `pecan.dates` package that can be used by `met.process` and other date-related packages.
* Pull functions related to MCMC analysis (e.g. convergence checking, burnin calculation, distributions) out of the `rtm`, `assim.batch`, and `utils` packages (and wherever else they are hiding) into a `pecan.mcmctools` package.

From [Hadley's package recommendations](http://r-pkgs.had.co.nz/namespace.html#exports)...

> I believe that packages that have a wide audience should strive to do one thing and do it well. All functions in a package should be related to a single problem (or a set of closely related problems).

From the [UNIX philosophy](https://en.wikipedia.org/wiki/Unix_philosophy):

> Write programs that do one thing and do it well.

mdietze:
Yeah, I agree that utils in particular has gotten broad and unwieldy -- I'm open to someone going through that bit in particular post-AGU.  I think most of the others aren't bad in terms of focus.

mdietze:
A related thing I've been thinking about is a slight re-organization of the packages in the root directory (utils, db, vis, etc) -- I think there should be a folder that's explicitly for 'core' tools so that all packages are in one of core, modules, and models. I think it might make it easier for new users to follow our conceptual organization

ashiklom:
Closing this because I think we've actually done a decent job of this.

PR #1247, opened 2017-01-23
Species code matching function

ashiklom:
## Description
This is a draft of the species-matching function intended as part of the vegetation data initial conditions workflow. I'm opening the pull request early to start a discussion -- this isn't necessarily ready to be pulled yet (though the core functionality -- matching USDA and FIA codes to BETY species IDs is functional).

## Types of changes

## Checklist:

@mdietze @serbinsh @istfer @bcow @tonygardella @mccabete


modules/data.land/R/helpers.R
```diff
@@ -0,0 +1,7 @@
+stop <- function(...) {
```
dlebauer: shouldn't this be in the utils package?

ashiklom:
@dlebauer I was using the `Symbol` field to refer to USDA codes, but if `AcceptedSymbol` is better, that's an easy fix.

modules/data.land/R/helpers.R
ashiklom: My thinking behind these functions is they're a way to remove the dependency on `utils`. I always thought it was a bad idea to have packages depend on all of `utils` (which is huge, and has a lot of its own dependencies) just for the `logger` functions. So my approach is to use it only if it's installed (`requireNamespace`), and if not, to fall back on the standard R messaging functions. But that's just me -- happy to change this if folks think I should.

dlebauer:
@ashiklom I forgot there was a field called 'Symbol' .. I think 'AcceptedSymbol' would be the canonical one that cross references USDA.

modules/data.land/R/helpers.R
dlebauer: I see your point, though this seems clunky and not [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself). Might be a good argument for moving the logger functions into a separate logging package. I don't want to be overly pedantic, just throwing out the idea.

ashiklom: I totally agree! I suggested creating a separate logger package, and more generally, making our packages smaller and more precisely focused, in #1191. We should discuss at the call on Thursday -- it's easy to implement, but would require revising a lot of existing PEcAn code and packages.

modules/data.land/R/helpers.R
```diff
@@ -0,0 +1,7 @@
...
+    if (requireNamespace(PEcAn.utils)) {
+        PEcAn.utils::logger.error(...)
+    } else {
+        stop(...)
```
mdietze: doesn't this have to be base::stop so that you don't just get an infinite recursion? Also, are you sure you want to map stop to logger.error rather than logger.severe, since logger.error doesn't call base::stop?

modules/data.land/R/match_species_id.R
```diff
@@ -0,0 +1,65 @@
+#' Match BETY species ID.
+#'
+#' Parses species codes in input data and matches them with the BETY species ID.
+#' 
+#' @param input_codes Character vector of species codes
+#' @param format_name Species code format name (character, length 1)
+#' @param betydb \code{dplyr} \code{src} object containing BETY connection 
+#' @param custom_translation_path File path to translation
```
mdietze: I thought we were going to overload this arg to be either the path or the table itself. In which case you might to rename to translation_table

modules/data.land/R/match_species_id.R
```diff
@@ -0,0 +1,65 @@
+#' Match BETY species ID.
+#'
+#' Parses species codes in input data and matches them with the BETY species ID.
+#' 
+#' @param input_codes Character vector of species codes
+#' @param format_name Species code format name (character, length 1)
```
mdietze: To me (character, length 1) implies that this code has to be a single character, which isn't true. I'd either list the valid options right here and/or document the valid options under Details

modules/data.land/R/match_species_id.R
```diff
@@ -0,0 +1,65 @@
...
+#' @param betydb \code{dplyr} \code{src} object containing BETY connection
```
mdietze: almost everywhere else we call this 'con'.  Changing variable names for the same thing is confusing and makes debugging a pain.

modules/data.land/R/match_species_id.R
```diff
@@ -0,0 +1,65 @@
+#' Match BETY species ID.
+#'
+#' Parses species codes in input data and matches them with the BETY species ID.
+#' 
+#' @param input_codes Character vector of species codes
+#' @param format_name Species code format name (character, length 1)
+#' @param betydb \code{dplyr} \code{src} object containing BETY connection 
+#' @param custom_translation_path File path to translation
+#' @return \code{data.frame} containing the following columns:
```
mdietze: When this is a path, documentation is ambiguous about what format this table has to be in (RData, RDS, csv, etc)

modules/data.land/R/match_species_id.R
```diff
@@ -0,0 +1,65 @@
...
+#'                        password = 'bety',
+#'                        host = 'localhost')
+#' input_codes <- c('ACRU', 'PIMA', 'TSCA')
+#' format_name <- 'usda'
+#' match_species_id(input_codes = input_codes,
+#'                  format_name = format_name,
+#'                  betydb = betydb)
+#'                  
+#' @export
+match_species_id <- function(input_codes, format_name, betydb = NULL, custom_translation_path = NULL, ...) {
```
mdietze: format_name should default to "custom"

modules/data.land/R/helpers.R
ashiklom: Yeah, I think you're right. The possibility of recursion didn't occur to me.

modules/data.land/R/match_species_id.R
ashiklom: They are not the same thing, which is why this is not called `con`. A `dplyr` `src` object (as documented here) is a list containing a `con` as well as some information about what kind of connection it is. Building an `src` from a `con` object is complicated, but extracting a `con` from an `src` is straightforward (just `src$con`). I thought we had discussed moving towards using `dplyr` for database queries, which is why I designed it this way.

modules/data.land/R/match_species_id.R
ashiklom: I don't like overloading. If we're going to have it work this way, I'd rather have two separate arguments.

modules/data.land/R/match_species_id.R
ashiklom: I thought the point was to pass this off to `load_data`, which takes a format as well?

modules/data.land/R/match_species_id.R
mdietze: I'm not saying to build the dplyr from con, or to stop using dplyr, but just to make sure that we're 100% consistent in what we call that connection variable.

modules/data.land/R/match_species_id.R
mdietze: Ok, then you'll want to add an additional arg and all the checks to make sure that a user doesn't specify both.

modules/data.land/R/match_species_id.R
mdietze: That's an option, but in that case format should be an explicit argument, since it's required but not documented. Personally, I don't like that option because it means you need to make a Format entry for both the data file and the species translation file, or you have to make a Format list object, which is also complicated. I think this is a case where it's fine to require a standard file format, since (unlike the veg data itself) no one is going to have this table just sitting around already (since the key column is the bety species id's).  All of this is making me thing that passing the data frame itself, rather than the path, will probably be the more common usage.

Once this function is in place, it would be very useful to think about the utilities we'll need to provide to make it easy for people to make these tables -- this seems like a great use case for the ROpenSci traits package for pulling those id's out of BETYdb (plus taxize, etc).

modules/data.land/R/match_species_id.R
ashiklom: OK. The `query.dplyr.R` scripts in PEcAn.DB use a `bety` argument, so should I change it to that?

mdietze: 👍

PR #1248, opened 2017-02-01
Add debugging warning to get.trait.data.R

mccabete:
Contains a debugging warning with instructions of what to do when get.trait.data loads an empty file.
## Motivation and Context

By circumventing the if statement for one run, a file gets created and allows the runs to work in the future.  In other words, theses instructions would probably only have to be followed once per PFT

## Types of changes

## Checklist:


mccabete:
@tonygardella  Was this what you had in mind?

mccabete:
@tonygardella  Was this what you had in mind?

tonygardella:
👍 That seems fine to me.

mdietze:
Rather than always printing that message, which may be suppressed depending on what level of warnings the user has set, you should check file.exists BEFORE hand and then give this as an error only when the file is missing

mccabete:
Yeah, the check really needs to see if the file is greater than zero. The file existed, it was just empty.

mccabete:
Actually I take back my previous comment. The file might not have existed. I can't quite confirm is this check is fool-proof though because I can't find another PFT that is producing the same error.

mdietze:
Circling in on a solution. The thing that got lost in the latest fix is that you replaced an error message that had very specific suggestions for how to SOLVE the problem with one that's technically correct but less helpful.

PR #1252, opened 2017-02-06
NEON CF site meteorology Initial version

### LINKED ISSUE #1140: Implement NEON's JSON-based data API
## Description

NEON has a draft API based on JSON interface to query and download data products. We should work on implementing an interface to this in PEcAn, both for the met drivers and flux output, and more generally for all the bio products. Given the size/scope of data, may need to start on smaller examples first.
## Context

PEcAn is supposed to work for NEON - it's in the proposal. Now that Batelle Ecology CI team has a draft API and some data rolling in (50% of sites are up and running), would be cool to define all NEON sites and be able to run there!
## Possible Implementation

download.AmerifluxLBL and download.Fluxnet2015 are examples of how JSON queries work in R. It takes some trial and error to get it right but http://data.neonscience.org/data-api# provides a range of examples and test scripts.

ankurdesai:
NEON Meteorlogy download function!

## Description
New function to download met from NEON sites in PECAN CF format

## Motivation and Context
Closes #1140 
## Types of changes

## Checklist:
- [X ] My change requires a change to the documentation.
Need to update/test web workflow to download met from NEON sites, and met.process.R


modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+
+  if (!file.exists(outfolder)) {
+    dir.create(outfolder)
+  }
+  
+  #Check if site is a NEON site
+  site <- sub(".* \\((.*)\\)", "\\1", sitename)
+  siteinfo <- nneo::nneo_site(site)
+  if (!exists("siteinfo")) {
+    logger.error("Could not get information about", sitename, ".", "Is this a NEON site?")
```
ashiklom: Should be `PEcAn.utils::logger.error`.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+  end_date <- as.POSIXlt(end_date, tz = "UTC")
+  start_ymd <- lubridate::ymd(start_date)
+  end_ymd <- lubridate::ymd(end_date)
+
+  #Subset to max/min available dates if start or end date exceed those bounds
+  allDates = unlist(availDates)
+  minDate <- min(allDates)
+  maxDate <- max(allDates)
+  start_ym = substr(start_ymd,1,7)
+  end_ym = substr(end_ymd,1,7)
```
ashiklom: Style note -- use arrows, not equals for assignment.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+  results$formatname <- "CF"
+  results$startdate  <- paste0(all_years, "-01-01 00:00:00")
+  results$enddate    <- paste0(all_years, "-12-31 23:59:59")
+  
+  for (current_year in all_years) {
+    
+    #Figure out start and end time for this year
+      y_idx <- current_year - start_year + 1
+      if (current_year==start_year) { start_m <- substr(start_ym,6,7) } else { start_m <- "01" }
+      if (current_year==end_year) { end_m <- substr(end_ym,6,7) } else { end_m <- "12" }
```
ashiklom: Based on our style guide, these should probably be split across multiple lines.
```r
if (...) {
  <case 1>
} else {
  <case 2>
}
```

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+  
+  for (current_year in all_years) {
+    
+    #Figure out start and end time for this year
+      y_idx <- current_year - start_year + 1
+      if (current_year==start_year) { start_m <- substr(start_ym,6,7) } else { start_m <- "01" }
+      if (current_year==end_year) { end_m <- substr(end_ym,6,7) } else { end_m <- "12" } 
+      days_in_last_month <- as.character(lubridate::days_in_month(lubridate::ymd(paste0(current_year,end_m,"-01"))))
+      start_ymd = (paste0(current_year,"-",start_m,"-01"))
+      end_ymd = (paste0(current_year,"-",end_m,"-",days_in_last_month))
```
ashiklom: Assignment with equals. Should be arrows.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+      start_date <- as.POSIXlt(paste0(start_ymd," 00:00:00 UTC"), tz = "UTC")
+      end_date <- as.POSIXlt(paste0(end_ymd," 23:30:00 UTC"), tz = "UTC")
+
+  #Warn if no data is available for any months in given year
+      monthsNeeded = substr(seq(as.Date(start_ymd),as.Date(end_ymd),by='month'),1,7)
+      if (length(intersect(unlist(availDates),monthsNeeded))==0) {
+        logger.warn("No data available in year ",current_year)
```
ashiklom: Needs `PEcAn.utils::`.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+        next()
+      }
+      startMon <- min(monthsNeeded)
+      endMon <- max(monthsNeeded)
+  
+  #Set up netcdf file, dimensions, and sequence of dates
+    new.file <- all_files[y_idx]
+    if (file.exists(new.file) && !overwrite) {
+      logger.debug("File '", new.file, "' already exists, skipping.")
```
ashiklom: `PEcAn.utils::`

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+    y <- ncdf4::ncdim_def("latitude", "degrees_north", lat)
+    xytdim <- list(x, y, t)
+    FillValue = NA
+  
+  #STEPS: Download all months in startdate to enddate for given variable
+  # Read CSV, copy to array, add to NetCDF file
+
+  #TEMPERATURE NEON.DP1.00002 Air Temp profile or NEON.DP1.00003 Triple-aspirated T (preferred)
+    airTempLoc = grep("DP1.00002",availProducts)
+    airTemp3Loc = grep("DP1.00003",availProducts)
```
ashiklom: Assignment operator.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+    if (length(airTemp3Loc)!=0) { 
+      airTemp3Dates = unlist(availDates[airTemp3Loc[1]])
+      airTemp3GoodDates = which((airTemp3Dates >= startMon) & (airTemp3Dates <= endMon))
+      nairTemp3 = length(airTemp3GoodDates)
+      } else { nairTemp3 = 0 }
+    if ((length(airTempLoc)==0) && (length(airTemp3Loc)==0)) {
+      logger.error("Air temperature DP1.00002 or DP1.00003 not available") }
+    if ((nairTemp==0) && (nairTemp3==0)) {
+      logger.error("Air temperature DP1.00002 or DP1.00003 not available in date range ",startMon," ",endMon)
+    }
```
ashiklom: Assignment operators. `PEcAn.utils` for logger functions.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+  #define NetCDF variable and create NetCDF file
+    airT.var <- ncvar_def(name = "air_temperature", units = "K", dim = xytdim)
+    nc <- ncdf4::nc_create(new.file, vars = airT.var)  #create netCDF file
+    ncdata <- rep(FillValue,length(datetime))
+    if (nairTemp3>nairTemp) {
+      for (mon in airTemp3Dates[airTemp3GoodDates]) {
+        neonData <- nneo::nneo_data(product_code = availProducts[airTemp3Loc[1]], site_code = site, year_month = mon)
+        if (length(neonData$data$urls)>0) {
+          #Extract and read 30 minute data from the highest vertical level among files returned
+          url30 <- tail(sort(neonData$data$urls[grep("*.00000.000.*_30min",neonData$data$urls)]),1)
```
ashiklom: Note that the `.` is interpreted as a regular expression wildcard in `grep/gsub` -- as in, it matches any character. If you want a literal dot, you'll want to escape it with `\\.`

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+          csvVar[which(csvQF!=1)]=NA  #triple-asp temp always has QF of 1
+          ncdata[arrLoc] <- udunits2::ud.convert(csvVar,"celsius", "K")
+        }
+      }
+    } else {
+      for (mon in airTempDates[airTempGoodDates]) {
+        neonData <- nneo::nneo_data(product_code = availProducts[airTempLoc[1]], site_code = site, year_month = mon)
+        if (length(neonData$data$urls)>0) {
+          #Extract and read 30 minute data from the highest vertical level among files returned
+          url30 <- tail(sort(neonData$data$urls[grep("*.00000.000.*_30min",neonData$data$urls)]),1)
```
ashiklom: See regex comment above. Also, if you reuse the same regular expression, it may be a good idea to store it as a string so you only have to write it once.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+          csvData <- read.csv(url30,stringsAsFactors=FALSE,header=TRUE)
+          #Retreive time dimension and figure out where in array to put it
+          csvDateTime <- as.POSIXct(gsub("T"," ",csvData$startDateTime),tz="UTC")
+          arrLoc <- floor(as.numeric(difftime(csvDateTime,datetime[1],tz="UTC",units="hours"))*2)+1
+          csvVar <- csvData$tempSingleMean
+          csvQF <- csvData$finalQF
+          csvVar[which(csvQF!=0)]=NA
```
ashiklom: Assignment operator.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+    }
+    ncdf4::ncvar_put(nc, varid = airT.var, vals = ncdata)
+
+    # NEON.DP1.00004 Pressure
+    pSurfLoc = grep("DP1.00004",availProducts)
+    if (length(pSurfLoc)!=0) { 
+      pSurfDates = unlist(availDates[pSurfLoc[1]])
+      pSurfGoodDates = which((pSurfDates >= startMon) & (pSurfDates <= endMon))
+      npSurf = length(pSurfGoodDates)
+    } else {npSurf=0}
```
ashiklom: Assignment operators here. Also, not the dot in the `grep` regex, as above. Also, `else` block should be on its own line.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+    if (npSurf>0) {
+      Psurf.var <- ncdf4::ncvar_def(name = "air_pressure", units = "Pa", dim = xytdim)
+      nc <- ncdf4::ncvar_add(nc = nc, v = Psurf.var, verbose = verbose)
+      ncdata <- rep(FillValue,length(datetime))
+      for (mon in pSurfDates[pSurfGoodDates]) {
+          neonData <- nneo::nneo_data(product_code = availProducts[pSurfLoc[1]], site_code = site, year_month = mon)
+          if (length(neonData$data$urls)>0) {
+            #Extract and read 30 minute data from the highest vertical level among files returned
+            url30 <- tail(sort(neonData$data$urls[grep("*.00000.000.*_30min",neonData$data$urls)]),1)
```
ashiklom: See regex comments above.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+      ncdf4::ncvar_put(nc, varid = Psurf.var, vals = ncdata)
+    }
+      
+    # NEON.DP1.00024 PAR
+    PARLoc = grep("DP1.00024",availProducts)
+    if (length(PARLoc)!=0) { 
+      PARDates = unlist(availDates[PARLoc[1]])
+      PARGoodDates = which((PARDates >= startMon) & (PARDates <= endMon))
+      nPAR = length(PARGoodDates)
+    } else {nPAR=0}
```
ashiklom: Assignment operators. Regex. If-else spacing style.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+      PAR.var <- ncdf4::ncvar_def(name = "surface_downwelling_photosynthetic_photon_flux_in_air", 
+                                  units = "mol m-2 s-1", 
+                                  dim = xytdim)
+      nc <- ncdf4::ncvar_add(nc = nc, v = PAR.var, verbose = verbose)
+      ncdata <- rep(FillValue,length(datetime))
+      for (mon in PARDates[PARGoodDates]) {
+        neonData <- nneo::nneo_data(product_code = availProducts[PARLoc[1]], site_code = site, year_month = mon)
+        if (length(neonData$data$urls)>0) {
+          #Extract and read 30 minute data from the highest vertical level among files returned
+          url30 <- tail(sort(neonData$data$urls[grep("*.00000.000.*_30min",neonData$data$urls)]),1)
```
ashiklom: Regex comment above.

ashiklom: Also, it seems like there is a lot of repetition in here. I would suggest trying to write a generic function that gets called multiple times rather than repeating code with minor changes.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+    if (length(SWLoc)!=0) { 
+      SWDates = unlist(availDates[SWLoc[1]])
+      SWGoodDates = which((SWDates >= startMon) & (SWDates <= endMon))
+      nSW = length(SWGoodDates)
+    } else { nSW = 0 }
+    if (length(SWLWLoc)!=0) { 
+      SWLWDates = unlist(availDates[SWLWLoc[1]])
+      SWLWGoodDates = which((SWLWDates >= startMon) & (SWDates <= endMon))
+      nSWLW = length(SWLWGoodDates)
+    } else { nSWLW = 0 }
```
ashiklom: Assignment, if-else spacing, regular expressions.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+          csvDateTime <- as.POSIXct(gsub("T"," ",csvData$startDateTime),tz="UTC")
+          arrLoc <- floor(as.numeric(difftime(csvDateTime,datetime[1],tz="UTC",units="hours"))*2)+1
+          csvVar <- csvData$inSWMean
+          csvQF <- csvData$inSWFinalQF
+          csvVar[which(csvQF!=0)]=NA  
+          ncdata[arrLoc] <- udunits2::ud.convert(csvVar,"W m-2", "W m-2")
+          csvVar <- csvData$inLWMean
+          csvQF <- csvData$inLWFinalQF
+          csvVar[which(csvQF!=0)]=NA  
+          ncdata2[arrLoc] <- udunits2::ud.convert(csvVar,"W m-2", "W m-2")
```
ashiklom: These are unnecessary unit conversions. I'm guessing they're here because you copy-pasted a code block and revised small pieces of it, which is an excellent indication that a function may be a good idea here.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
...
+  #TERMS is a five-digit designation used in data product numbering to identify a sub-product or discrete vector of metadata. Since each download file typically contains several sub-products, this field is set to 00000 in the file name to maintain consistency with the data product numbering scheme.
+  #HOR is the Spatial Index #1 and refers to measurement locations within one horizontal plane. For example, if five surface measurements were taken, one at each of the five soil array plots, the number in the HOR field would range from 001-005.
+  #VER is the Spatial Index #2 and refers to measurement locations within one vertical plane. For example, if eight temperature measurements are collected, one at each tower level, the number in the VER field would range from 010-080.
+  #TMI is the Temporal Index; it is a three-digit designation and refers to the temporal representation, averaging period, or coverage of the data product (e.g., minute, hour, month, year, sub-hourly, day, lunar month, single instance, seasonal, annual, multi-annual)
+  #DESC is an abbreviated description of the data product 
+  
+  #VER - highest number = highest level
+  #Default take HOR 001 VER (max)
+
```
ashiklom: All of these comments should probably go in the `@details` documentation block so that people can access them while viewing documentation. More info [by Hadley](http://r-pkgs.had.co.nz/man.html#roxygen-comments).

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,486 @@
+##' Download NEON Site Met CSV files
+##'
+##' @name download.NEONmet
+##' @title download.NEONmet
```
ashiklom: The `@name` and `@title` tags are bad practice that we have, unfortunately, adapted somehow. I would strongly recommend removing them. The `@title` tag is redundant with what you already have as the function title ("Download NEON Site Met CSV files"), and the `@name` tag is read automatically from the function name. The only time you actually want to set the `@name` tag is when you specifically want the displayed function name to be different from the function being described, which is rarely the case.

The recommended structure for documentation (described [here](http://r-pkgs.had.co.nz/man.html#roxygen-comments)) is to not use tags for the introduction at all. Instead, the structure of the documentation should be:

```r
#' This is the function title
#'
#' This is a paragraph. This paragraph contains the function description.
#' It is technically the same as the @description tag, but the recommendation is to avoid
#' using that explicit tag.
#' 
#' This paragraph, and any paragraphs following, are the same as the @details tag. This is
#' a good place to put the NEON file documentation you have at the bottom.
#' 
#' This is another paragraph in the @details section. Etc, etc.
#'
#' @export
#' @param ...
```

mdietze: I think this practice is a legacy of an earlier version of ROxygen yelling at us when these tags were not explicit. If this is done automagically now I'm more than happy to see this updated.

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: I thought the @importsfrom line in the top made it so that those functions did not need library declaration, no?

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: I really wish RStudio would just automatically change one into the other when typed!

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: Thanks. Note that the grep statement does subtly change across products, but you're right, it's common enough to most to be abstracted into a string.

modules/data.atmosphere/R/download.NEONmet.R
ashiklom: Yes! You're right. I missed that. Disregard all the logger related comments below.

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: See comment below. Virtually no single variable gets manipulated exactly the same way. Sometimes I need the top URL, sometimes the bottom, sometimes grep for _30min, sometimes for _30_minutes. Sometimes there are multiple variables with similar data and I need to have a switch statement to only choose the one with most data (or one I know is better). Will think about it, but not sure much is gained by abstracting here.

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: I agree, but I think the overhead on the call is very small and it explicitly calls out the units here for readability, so that if there is ever a change, it's a simple fix. This was deliberate. I can abstract out lines 336-337, which are the same, but now we have a function call on two lines of code, that calls another function, so interpreter stack gets deeper. Further line 336 has a few variants (QF==0 QF!=0, multiple vars)

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: Got it, those were my cut and paste notes as I was developing frustrated by lack of easy to find NEON documentation on the API anywhere except buried in technical PDF. I could just remove out of the function.

ankurdesai:
Alexey, Thanks for the style guide comments, those are easy to fix. I'm at the AGU Editor's meeting for now, so won't get to it for a bit (next week I would guess). Yes, abstracting out reusable code was my initial goal and then it quickly fell apart. The redundancy is trickier here for a few reasons: 1) NEON API is a moving target, 2) variable names differ per product and can only be figured out by downloading and testing each file (same is true for units unless I want to pore over 1000s of pages of technical documents), 3) some variables need manipulation (like wind direction) and multiple NCDF writes, others just one, 4) the quality flag convention or variable name of this flag is not the same for each variable (sometimes 0 = good, sometimes 1 = good, I suspect this will change), 5) some variables have redundant sensors  and so selecting the variable name, horizontal height, vertical level varies, and hence the grep statement, screening, and storing does vary. I'll try to abstract out a few bits, put a few repeated string literals into variables. See met2CF.csv.R for a similar type of complication (the script that this was based on).

ankurdesai:
@ashiklom I have made the requested changes. Let me know if anything else should be changed.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,398 @@
...
+                        stringsAsFactors = FALSE)
+  
+  all_years <- start_year:end_year
+  all_files <- file.path(outfolder, paste0("NEONmet.", site, ".", as.character(all_years), ".nc"))
+  results$file <- all_files
+  results$host <- fqdn()
+  results$mimetype   <- "application/x-netcdf"
+  results$formatname <- "CF"
+  results$startdate  <- paste0(all_years, "-01-01 00:00:00")
+  results$enddate    <- paste0(all_years, "-12-31 23:59:59")
```
ashiklom: Assuming all of this has been tested, there's no need to change this. But for the future, it may be good to use functions specifically designed to process dates, either in base R, or functions from `lubridate`.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,398 @@
...
+        start_m <- "01" 
+      }
+      if (current_year==end_year) { 
+        end_m <- substr(end_ym,6,7) 
+      } else { 
+        end_m <- "12" 
+      } 
+      days_in_last_month <- as.character(lubridate::days_in_month(lubridate::ymd(paste0(current_year,end_m,"-01"))))
+      start_ymd <- (paste0(current_year,"-",start_m,"-01"))
+      end_ymd <- (paste0(current_year,"-",end_m,"-",days_in_last_month))
```
ashiklom: Related to above comment, I think the better way to do this is to use `strftime` and similar functions, rather than ad-hoc pasting like this. Those may fail for illegal dates, which makes them easier to debug.

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,398 @@
...
+  if (length(Loc)>0) { 
+    Dates <- unlist(availDates[Loc[1]])
+    GoodDates <- which((Dates >= startMon) & (Dates <= endMon))
+    if (length(GoodDates)>0) {
+      return(Dates[GoodDates])
+    } else {
+      return(NULL)
+    }
+  } else { 
+    return(NULL)
```
ashiklom: Should these `NULL` returns also come with a warning or an info message? E.g. `Warning: No dates found for parameters X and Y...`

modules/data.atmosphere/R/download.NEONmet.R
```diff
@@ -0,0 +1,398 @@
...
+      #If belowground, then take top most level (lowest value)
+      if (belowground==TRUE) {
+        url30 <- head(sort(urls[grep(urlstring,urls)]),1)        
+      } else {
+        url30 <- tail(sort(urls[grep(urlstring,urls)]),1)        
+      }
+      if (length(url30)!=0) {
+        csvData <- read.csv(url30,stringsAsFactors=FALSE,header=TRUE)
+        #Retreive time dimension and figure out where in array to put it
+        csvDateTime <- as.POSIXct(gsub("T"," ",csvData$startDateTime),tz="UTC")
```
ashiklom: No change required, but the `stringr` package from the Hadleyverse provides some nice substitutes for string matching functions, in case you have to do more complex string operations in the future.

ashiklom:
Looks good to me! My only real complaint is the absence of any unit tests or usage examples in the documentation. See the `@examples` section in [this discussion](http://r-pkgs.had.co.nz/man.html#man-functions).

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: The warnings are provided afterwards. If it seems helpful during runs to provide more messages, I suppose we can provide logger.warnings here.

mdietze:
While @ashiklom continues to have good suggestions, I'm going to pull this and leave it to @ankurdesai to decide how many of these he wants to follow up on in future PR

modules/data.atmosphere/R/download.NEONmet.R
ankurdesai: True. This is the standard in all the "results" objects used in all the download functions that met.process relies on for inserting into BETY (which knows little of R dates), so if we change, this has to work with met.process and should be changed everywhere.

PR #1253, opened 2017-02-07
first pass at matching species codes to PFTs for veg IC workflow.

mdietze:


modules/data.land/R/match_pft.R
```diff
@@ -0,0 +1,53 @@
...
+                      "pfts_species AS bps ON bps.pft_id = bp.id INNER JOIN species AS bs ON bs.id = bps.specie_id WHERE ", 
+                      "bp.name = '", pft$name, "'")
+    } else {
+      query <- paste0(query, " OR bp.name = '", pft$name, "'")
+    }
+  }
+  translation <- db.query(query, con = con)
+  
+  ## Check for duplicate bety_species_ids in PFTs
+  bad <- translation[duplicated(pfts$bety_species_id),]
```
istfer: I think this was supposed to be `translation[duplicated(translation$bety_species_id),]`

modules/data.land/R/match_pft.R
```diff
@@ -0,0 +1,53 @@
...
+##' 
+##' Matches BETYdb species IDs to model-specific PFTs
+##' 
+##' @param bety_species_id  vector of BETYdb species IDs
+##' @param pfts             settings$pfts.  List of pfts with database matching based on name
+##' @param con              database connection
+##' 
+##' @return table of BETYdb PFT IDs matched to species IDs
+##' 
+match_pft <- function(bety_species_id, pfts, con){
```
istfer: is this `pfts` going to be exactly the same with the settings? what about non-pft pfts like soil, should we handle them before or in the function? (e.g. soil case would fail for duplicated species check etc.)

mdietze: I don't think the soil pfts will cause a duplication or unmatched failure because they don't have any species associated with them

modules/data.land/R/match_pft.R
mdietze: yep, fixing

modules/data.land/R/match_pft.R
istfer: currently both ED and SIPNET `soil` PFT are associated with "unknown" as species, should we correct it on the database then? would it [a pft not being associated with any species] cause any problems elsewhere in the workflow? alternatively we can rename them as soil.ed, soil.sipnet

mdietze: I'd try correcting the db. Don't know if that will cause an error but it's worth a try. I don't think renaming will solve the problem

PR #1256, opened 2017-02-09
PDA updates - n_eff + ED soil

istfer:
## Description
- Added autocorrelation correction to PDA workflow
- Added ED decomposition parameters to be read from config.xml

## Types of changes


models/ed/R/write.configs.ed.R
```diff
@@ -377,30 +383,44 @@ write.config.xml.ED2 <- function(settings, trait.values, defaults = settings$con
       # happen is for there to be two settings for each PFT: the 'num' to use to represent the PFT to
       # ED, and the 'defaults.PFT' (name or number) to use for pulling default parameter values.
       pft.number <- pftmapping$ED[which(pftmapping == pft)]
-      if (length(pft.number) == 0) {
+      
+      if(pft=="soil"){
+        data(soil)
```
ashiklom: If this is coming from inside the package, I would specify it explicitly. `data(soil, package = "packagename")`.

modules/assim.batch/R/pda.emulator.R
```diff
@@ -211,15 +230,17 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
       for (i in seq_len(settings$assim.batch$n.knot)) {
         align.return <- pda.get.model.output(settings, run.ids[i], bety, inputs)
         model.out[[i]] <- align.return$model.out
-        if(!is.na(model.out[[i]])){
+        suppressWarnings(if(!is.na(model.out[[i]])){
           inputs <- align.return$inputs
-        } 
+        })
```
ashiklom: Why are you suppressing warnings here? If it's because the conditional doesn't work, a handy function I've discovered recently is `isTRUE()`, which returns `TRUE` if and only if the inside is TRUE. 

```r
isTRUE(TRUE)
# TRUE
isTRUE(1 > NA)
# FALSE
isTRUE(1 > NULL)
# FALSE
isTRUE(NA)
# FALSE
isTRUE(NULL)
# FALSE
```

modules/assim.batch/R/pda.emulator.R
```diff
@@ -450,19 +487,27 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
   
   for (c in seq_len(settings$assim.batch$chain)) {
     
-    m <- mcmc.out[[c]]$mcmc.samp
+    m <- matrix(NA, nrow =  nrow(mcmc.out[[c]]$mcmc.samp), ncol = length(prior.ind.all.ns))
```
ashiklom: For `mcmc` objects from the `coda` package, there are handy `coda::niter` and `coda::nchain` functions that may be somewhat more robust. Just a thought, no need to change it here...

modules/assim.batch/R/pda.emulator.R
```diff
@@ -450,19 +487,27 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
...
       prior.all <- do.call("rbind", prior.list)
       prior.fn.all <- pda.define.prior.fn(prior.all)
       
       ## Convert probabilities back to parameter values
-      for (i in seq_len(sum(n.param))) {
-        m[, i] <- eval(prior.fn.all$qprior[prior.ind.all][[i]], 
-                       list(p = mcmc.out[[c]]$mcmc.samp[, i]))
+      cc  <- 1
+      for (i in seq_along(prior.ind.all.ns)) {
+        # be careful that agrep(l) is working for now but it won't work for very similar parameter names
```
ashiklom: Why are we using `agrep` here? I am extremely uncomfortable with approximate pattern matching in this context. If you're looking for partial matches, basic `grep` will already do that. If you know that the results will follow a pattern, it's much safer to construct an explicit regular expression pattern and use that.

modules/assim.batch/R/pda.emulator.R
```diff
@@ -472,16 +517,20 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
...
+     format      = "lin"
+     mix         = mix
+     jmp0        = jmp.list[[chain]]
+     ar.target   = settings$assim.batch$jump$ar.target
+     priors      = prior.fn.all$dprior[prior.ind.all]
+     settings    = settings
+     run.block   = (run.normal | run.round)  
+     n.of.obs    = n.of.obs
+     llik.fn     = llik.fn
+     resume.list = resume.list[[chain]]
```
ashiklom: What is the purpose of this block? I'm assuming it's for interactive testing (which is smart! I'll have to start using that), and if so, it might be good to add a comment indicating that.

Also, I've taken a liking to using `interactive()` (no arguments), which returns `TRUE` if you're running interactively and `FALSE` if you're running via script. It's probably not appropriate here, but just thought I'd call your attention to it.

modules/assim.batch/R/pda.neff.R
```diff
@@ -0,0 +1,163 @@
+##' Autocorelation correction and efficient sample size calculation on latent process
+##'
+##' @title Calculate N_eff 
+##' @param inputs list
+##'
+##' @return inputs list, updated inputs with n_eff
+##'
+##' @author Istem Fer
+##' @export
+pda.neff.calc <- function(inputs){
```
ashiklom: In general, we should try to move away from using dots in function names. But no real need to change it here.

modules/assim.batch/R/pda.neff.R
```diff
@@ -0,0 +1,163 @@
...
+  
+  
+  for(i in seq_along(inputs)){
+    
+    n   <- inputs[[i]]$n
+    # for now we're doing autocorrelation correction on flux data only
+    # NEE, LE, FC
+    if(inputs[[i]]$variable.id %in% c(297, 298, 1000000042)){
```
ashiklom: Personally, I would rather see these be a named vector or separated, informative named variables. That reduces the likelihood that they will get lost at some point. Also, it might be a good idea to make these a function argument -- that will make this script more versatile later.

modules/assim.batch/R/pda.neff.R
```diff
@@ -0,0 +1,163 @@
...
+    }
+    
+    j.model   <- jags.model(file     = textConnection(HLModel),
+                            data     = data,
+                            inits    = init,
+                            n.chains = 3)
+
+    
+    
+  }else{
```
ashiklom: In cases like this, I think it's more robust to have explicit `else if`s and an informative warning/error message if the user tries to pass a model that is not supported.

modules/assim.batch/R/pda.neff.R
```diff
@@ -0,0 +1,163 @@
...
+##' @param input list that contains time-series data vector and parameters for heteroskedastic.laplacian
+##' @param model data model type, for flux data heteroskedastic laplacian, normal is an example 
+##'
+##' @return rho AR(1)
+##'
+##' @author Istem Fer
+##' @export
+pda.autocorr.calc <- function(input, model = "heteroskedastic.laplacian"){
+  
+  library(rjags)
```
ashiklom: `library` calls inside of functions are not a good idea. Since you only need like 2 functions from this package, I would just add `rjags::<name>` to them.

Also, you may be able to simplify some of the running syntax with the `runjags` package, which also has the added bonus of being easier to install _and_ can automatically run until convergence, which is better than arbitrarily setting `niter` as you do now.

modules/assim.batch/R/pda.neff.R
```diff
@@ -0,0 +1,163 @@
...
+    for(i in 2:n){
+      x[i] ~ dnorm(x[i-1], tau_add)
+    }
+
+    #### Priors
+    x[1] ~ dnorm(x_ic,tau_ic)
+    tau_obs ~ dgamma(a_obs,r_obs)
+    tau_add ~ dgamma(a_add,r_add)
+    }
+    "
```
ashiklom: This feels like it should have an analytical solution. If so, this could have the advantage of being considerably more computationally efficient, which doesn't matter much here because you're only running it once, but would be extremely useful for my algorithm which does this many times over the course of the MCMC.

modules/assim.batch/R/pda.neff.R
```diff
@@ -0,0 +1,163 @@
...
+                              variable.names = c("x"),
+                              n.iter         = 5000,
+                              thin           = 100)
+  
+  
+  out <- as.matrix(jags.out)
+  median.out <- apply(out, 2, median)
+  
+  ar  <- acf(median.out)
+  rho <- as.numeric(ar$acf[2])
```
ashiklom: We should chat about this -- I'm not exactly sure what this code is doing, and I'd like to know. I'm sure it's correct -- I'm just genuinely curious about the thought process.

modules/assim.batch/R/pda.utils.R
```diff
@@ -608,12 +608,23 @@ pda.adjust.jumps.bs <- function(settings, jcov, accept.count, params.recent) {
...
   probs <- matrix(0.5, nrow = n.knot, ncol = n.param.all)
   
-  # Fill in parameters to be sampled with probabilities sampled in a LHC design
-  probs[, prior.ind] <- PEcAn.emulator::lhc(t(matrix(0:1, ncol = length(prior.ind), nrow = 2)), n.knot)
+  # if the params are going to be scaled leave them out, if sf is NULL nothing happens
+  inds <- prior.ind[!prior.ind %in% which(pname %in% sf)]
+  
+  if(length(inds) !=0){
+    # Fill in parameters to be sampled with probabilities sampled in a LHC design
+    probs[, inds] <- PEcAn.emulator::lhc(t(matrix(0:1, ncol = length(inds), nrow = 2)), n.knot)
```
ashiklom: Instead of transforming the matrix, you can use the argument `byrow = TRUE`.

modules/assim.batch/R/pda.utils.R
```diff
@@ -608,12 +608,23 @@ pda.adjust.jumps.bs <- function(settings, jcov, accept.count, params.recent) {
...
+  }
+  
+  inds <- prior.ind[prior.ind %in% which(pname %in% sf)]
+  
+  if(!is.null(sf) & length(inds) !=0){
```
ashiklom: Single `&` is a vectorized conditional. Double `&&` is an evaluation left to right. Not sure it technically matters here, but for cases like this, you probably want `&&`.

modules/assim.batch/R/pda.utils.R
```diff
@@ -626,6 +637,27 @@ pda.generate.knots <- function(n.knot, n.param.all, prior.ind, prior.fn, pname)
...
+  n.sf <- length(sf)
+  # prior for scaling factor
+  prior.sf <- data.frame(distn = rep("unif", n.sf), 
+                         parama = rep(0, n.sf), 
+                         paramb = rep(1, n.sf), 
+                         n = rep(NA, n.sf))
+  rownames(prior.sf) <- paste0(sf,"_SF")
+  prior.list[[length(prior.list)+1]] <- prior.sf
+  
+  probs.sf <- PEcAn.emulator::lhc(t(matrix(0:1, ncol = n.sf, nrow = 2)), n.knot)
```
ashiklom: Same comment about `byrow` as above...

modules/assim.batch/R/pda.emulator.R
istfer: here model.outputs[[i]] itself is a list, and all that sublists are data frames of aligned model outputs

I want to check if the model run failed or not, so just looking at the first index is enough, the check gives me "using the first element" sorts of warnings, I was just being lazy :) I'll try to specify the check more precisely

modules/assim.batch/R/pda.emulator.R
istfer: I'm pasting the example we just discussed as a note to myself: `grep("growth.rate\\d*", "growth.rate.sf")`

modules/assim.batch/R/pda.emulator.R
istfer: fancy!

yes, these are the arguments to the mcmc.GP function I had to debug it so many times now that this comes very handy

modules/assim.batch/R/pda.neff.R
istfer: I know, all pda functions are dot separated :/ I need to sit and change them all at some point

modules/assim.batch/R/pda.neff.R
istfer: yep that's a remnant of developing, good catch

istfer:
thanks @ashiklom these are all very helpful, I'll work on them while re-testing this PR changes

istfer:
@ashiklom my tests seem to be working, I think I don't have anything else to add at this point

models/ed/R/write.configs.ed.R
```diff
@@ -377,30 +383,44 @@ write.config.xml.ED2 <- function(settings, trait.values, defaults = settings$con
       # happen is for there to be two settings for each PFT: the 'num' to use to represent the PFT to
       # ED, and the 'defaults.PFT' (name or number) to use for pulling default parameter values.
       pft.number <- pftmapping$ED[which(pftmapping == pft)]
-      if (length(pft.number) == 0) {
+      
+      if(pft=="soil"){
```
mdietze: This might be OK as a first pass, but it prohibits us from ever having additional soil PFTs -- maybe grepping against a agreed upon naming pattern might work better?

modules/assim.batch/R/pda.define.llik.R
```diff
@@ -88,17 +91,16 @@ pda.calc.error <-function(settings, con, model_out, run.id, inputs, bias.terms){
...
+      bc <- bc + 1
       pda.errors[[k]] <- SS 
       SSdb[[k]]       <- log(SS)
       
     } else { # Gaussian
       
       SS <- sum((model_out[[k]] - inputs[[k]]$obs)^2, na.rm = TRUE)
       
-      pda.errors[[k]] <- SS 
-      SSdb[[k]]       <- log(SS)
```
mdietze: In the current context, I'm not 100% following what these two variables are, so I'm curious why they were deleted here but kept for the other cases

modules/assim.batch/R/pda.define.llik.R
```diff
@@ -67,18 +67,21 @@ pda.calc.error <-function(settings, con, model_out, run.id, inputs, bias.terms){
   for (k in seq_len(n.input)) {
     
     if (settings$assim.batch$inputs[[k]]$likelihood == "Laplace") {
```
mdietze: More general design question (not something that needs to be done immediately): if we defined the llik functions more generically as actual functions with a name and I/O pattern (as we do in other places), could we avoid having all these special cases? I worry this function is going to get too long and hard to maintain as the number of likelihoods increases. Plus, using functions would make it easier for users to add new functions, including doing so on the fly (e.g. in local scripts rather than recompiling this package)

mdietze: Might also be a way to make them a special case of Betsy's more general Metric function catalog?

modules/assim.batch/R/pda.emulator.R
```diff
@@ -211,15 +246,17 @@ pda.emulator <- function(settings, params.id = NULL, param.names = NULL, prior.i
...
-        } 
+        }
       }
-      
+
       
       current.step <- "pda.get.model.output"
       save(list = ls(all.names = TRUE),envir=environment(),file=pda.restart.file)
       
+      inputs <- pda.neff.calc(inputs)
```
mdietze: This would benifit from a comment or two explaining what happens at this step

modules/assim.batch/R/pda.neff.R
mdietze: Perhaps @ashiklom's comment is an indication that this needs a better explanation. What we're trying to do is to calculate the autocorrelation of the latent state, after attempting to "remove" the observation error. The first step is thus to estimate the latent state using a simple 'process free' state-space model (e.g. random walk). In terms of available analytical solutions, that gets into the realm of Kalman Filters and the like, which would also give an approximate (though unidirectional) estimate that would be pointwise analytically tractable and must faster (though not analytically tractable across the whole time series). At this point, since it's a one-time calc, I prefer the full state-space, though I'd be open to approximations if we need to deal with considerably larger data sets (e.g. remotely-sensed space-by-time-by-spectral band data cubes)

modules/assim.batch/R/pda.define.llik.R
istfer: this is a typo, I tested with this code quite a bit, and at some point I took out the assignments out of the if-else block then put it back later, I was using multiplicative Gaussian in my tests so I missed this thanks!

modules/assim.batch/R/pda.neff.R
istfer: yeah, we chatted with @ashiklom afterwards, I'll include a brief version your comment in the code

modules/assim.batch/R/pda.define.llik.R
istfer: hmm OK I'll look int this later

PR #1263, opened 2017-02-27
IC workflow - first pass

istfer:
## Description

I haven't covered all the bases yet but I wanted to give you all some time to look at it before Thursday.

Some notes:
- This version works for FIA and my GapMacro inventory data so far (and only for ED). I'll try to include more test cases soon. 
- It's not modularized yet, but I think once we agree on the overall design, adding another level and modularizing shouldn't be that difficult. I can also take a stab at it.
- Currently, it's not handling the copying of the files to remote, we'll need to copy them by hand. 
- Before, `fia2ED` was processing all site/pss/css together, but now I process them in separate iterations. Therefore `css` needs to come after the `pss` in the xml. I think it should be a simple fix in the web later.

I'd especially appreciate feedback on database related issues: I'm confused about what to insert to the database (are we going to save it after each stage?) and how to insert them? (as input files? what about `parentids`? just as dbfiles -is that possible-?)

The other thing is that I included a function called `format_wide2long` (`wide2long` was taken). When `load_data` sees there are more than one columns of the same variable in the data it decides it's wide format. Then `format_wide2long` function decides how to transform the data according to the info in the `storage_type`. That was the solution I could come up with. I tried to put down many comments in that script but let me know if anything is not clear.

The general flow is like this:
``` 
workflow.R  
do.conversions 
ic_process 
   if FIA, 
        download.FIA 
   if GapMacro, 
        load_data <--> wide2long
   if css ,
        match_species_id 
   if css, 
        match_pft 
   veg2model.ED2
   return(ic.file.path)
```


models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+#' @param host 
+#' @param con bety$con
+#' @param overwrite
+#' @return ic.id initial conditions file input id
+#' @export
+#' @author Istem Fer
+#'  We need the in.path, the out folder, the start and endates and the lat/long. All of these are in setting$run. 
+veg2model.ED2 <- function(obs = NULL, inputinfo, runinfo, outfolder, host, con, overwrite = FALSE){
+  
+  start_year <- lubridate::year(runinfo$start.date)
```
tonygardella: I believe we're trying to move away from the  'package::function' method of calling external functions to using the '@importFrom package function1 function2' method in the roxygen heading so that it can include these imports in the NAMESPACE.

mdietze:
Haven't looked at the code yet, but I'm not sure I like the design decision to handle IC's in multiple passes. That seems like an approach that will be hard to generalize and ultimately make it more challenging to build IC files for different models (e.g. having to update an IC you just built with new information being added in a subsequent pass, which means every model needs to be able to both read and write it's IC, rather than just write). Forcing data fusion to be layer-by-layer could get complicated.

istfer:
Yes, that was something I wasn't sure about, whether to pass and return settings to/from IC workflow or handle one file at a time. I was trying to mimic met.process. This requires writing out mid-stages as we go, read them back and modify/update if necessary. Might result in redundant intermediate-files, which are not quite counterparts of `met.CF` files. We can discuss more

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+  end_year   <- lubridate::year(runinfo$end.date)
+  startdate   <- lubridate::as_date(paste0(start_year, "-01-01"))
```
istfer: I was copying the line from [fia2ED](https://github.com/PecanProject/pecan/blob/master/modules/data.land/R/fia2ED.R#L30_L31) to be honest, I also thought it was a weird thing to do but guessed it's an FIA quirk

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+  enddate     <- lubridate::as_date(paste0(end_year, "-12-31"))
+  lat <- as.numeric(runinfo$site$lat)
+  lon <- as.numeric(runinfo$site$lon)
+  
+  # handle file details
+  if(inputinfo$output == "site"){
+    
+    formatname <- "ED2.site"
+    dbfilename <- "site.file"
+    fileprefix <- paste("siteid", runinfo$site$id, inputinfo$source, start_year, end_year,
```
istfer: there is this small `site.style = FALSE / TRUE` thing we talked about, it's same for css/pss but slightly more rounded lat/lon for the site file, if that naming difference is unnecessary I can mode it out of the if chunk

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+      obs <- read.table(inputinfo$path, header = TRUE)
+    }else if(is.null(obs)){
+      logger.severe("No css data found.")
+    }
+    
+    if(inputinfo$source == "FIA"){ # we still need one more check from pss
+      
+      pss <- read.table(runinfo$inputs$pss$path, header = TRUE)
+      
+      # Remove rows that don't map to any patch
```
istfer: 1. I guess we don't
2. I can probably get rid of this

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+    }
+    
+    if(is.null(obs$n)){
+      # is this how n is calculated?
+      obs$n <- ifelse(!is.null(inputinfo$area), nrow(obs)/inputinfo$area, 0.001)
+    }
+    
+    if(inputinfo$source == "GapMacro"){
+      inv.years <- as.numeric(unique(obs$year))
+      # suitable years
```
istfer: FIA year check was in its query already, I might have some questions about this later, need to check what was happening to FIA data again

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+      av.years <- inv.years[inv.years < start_year]
+      if(length(av.years) == 0){
+        logger.severe("No available years found in the data.")
+      }
+      obs$time <- max(av.years)
+      # filter out other years
+      obs <- obs[obs$year == obs$time, ]
+      colnames(obs)[colnames(obs) == "DBH"] <- "dbh"
```
istfer: load_data converts to `DBH` which is the pecan name. 

On ED2 wiki, this column is named `dbh` (and it's what `fia2ED` was writing as well), I assumed it needs to be lower case `dbh` for ED to read properly, but now I remember something like these column headers didn't matter, just the order was important, is that correct?

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+    }
+  
+    
+    # --- Continue work formatting css 
+    n.cohort                      <- nrow(obs)
+    obs$time[is.na(obs$time)]     <- 1
+    obs$cohort[is.na(obs$cohort)] <- 1:sum(is.na(obs$cohort))
+    obs$dbh[is.na(obs$dbh)]       <- 1  # assign nominal small dbh to missing
+    density.median                <- median(obs$n[which(obs$n > 0)])
+    obs$n[is.na(obs$n) | obs$n == 0]    <- density.median
```
istfer: again, I was copying these lines from [fia2ED](https://github.com/PecanProject/pecan/blob/master/modules/data.land/R/fia2ED.R#L274_L281)  

I clearly need to go through these again

models/ed/R/veg2model.ED2.R
```diff
@@ -0,0 +1,234 @@
...
+    
+    obs <- obs[, c("time", "patch", "cohort", "dbh", "hite", "pft", "n", "bdead", "balive", "lai")]
+    
+    # Locally write files
+    write.table(obs, localfile, quote = FALSE, col.names = TRUE, row.names = FALSE)
+  }
+
+
+  # Insert into DB
+  ic.id <- dbfile.input.insert(
```
istfer: OK I'll do that, DB insertions were just for testing and was confusing me anyway, I'll get rid of those

modules/benchmark/R/format_wide2long.R
```diff
@@ -0,0 +1,103 @@
+##' @name format_wide2long
+##' @title format_wide2long
+##' 
+##' @param out wide format data
```
istfer: because it's the `out` coming from [load_data](https://github.com/PecanProject/pecan/blob/master/modules/benchmark/R/load_data.R#L55), I'll try to improve all my headers

modules/benchmark/R/format_wide2long.R
```diff
@@ -0,0 +1,103 @@
...
+##' @param format as returned by query.format.vars
+##' @param vars_used 
+##' @param time.row
```
istfer: it is, and then it is extracted in `load_data` before things are passed to this function, I could update `format` within this function and re-extract `time.row` in `load_data` once the output of this function is returned, but I decided to handle those here to keep `load_data` a bit cleaner

modules/benchmark/R/format_wide2long.R
```diff
@@ -0,0 +1,103 @@
...
+##' @param vars.used.index
```
istfer: these variables are coming from `load_data` with those names, I update them within this function according to the transformation, and they are being used in the downstream of the `load_data` later. I thought we were avoiding periods in the function names. Do you want me to change things in `load_data` too?

modules/benchmark/R/format_wide2long.R
```diff
@@ -0,0 +1,103 @@
...
+  ####################
+  # variable  value  #  
+  #  diam12     12   #
+  #  diam14     10   #
+  #  diam16     23   #
+  ####################
+  
+  # diam12/14/16 are the column headers in the input file
+  # you probably want to rename "value" column as "DBH" here
+  colnames(mout)[colnames(mout) == "value"] <- unique(vars_used$bety_name[duplicated(vars_used$bety_name)])
```
istfer: I thought about this too but one wide-variable was my only test case so far and this was a first pass :)  I'll see if the same or a similar approach will hold up with multiple wide variables in a single dataset

modules/benchmark/R/load_data.R
```diff
@@ -58,6 +58,16 @@ load_data <- function(data.path, format, start_year = NA, end_year = NA, site =
   
   vars_used <- format$vars[vars.used.index, ]
   
+  # check wide format and transform to long
+  if(any(duplicated(vars_used$bety_name))){
```
istfer: Hmm good point.. Can we trust that there will be a consistent naming pattern? I'll revise the wide2long transformation

utils/R/do.conversions.R
```diff
@@ -24,10 +24,20 @@ do.conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     
     input.tag <- names(settings$run$input)[i]
     
-    # fia database
+    # IC conversion : for now for ED only, hence the css/pss/site check
     if ((input.tag %in% c("css", "pss", "site")) &&
-        is.null(input$path) && !is.null(input$source) && (input$source == "FIA")) {
-      settings <- PEcAn.data.land::fia.to.psscss(settings, overwrite=overwrite.fia)
```
mdietze: 1) I wonder if we should retain a way to call fia.to.psscss until the new workflow is fully functioning? 
2) When fia.to.psscss is out of the workflow it needs to start the formal deprecation process (see Backward compatibility under http://r-pkgs.had.co.nz/release.html)

modules/data.land/R/match_species_id.R
```diff
@@ -68,18 +68,49 @@ match_species_id <- function(input_codes, format_name = 'custom', bety = NULL, t
...
+        # can call traits::betydb_query one at a time?
+        # reduce the number of calls
+        translation <- data.frame(input_code = unique(input_codes),
+                           bety_species_id  = rep(NA, length(unique(input_codes))),
+                           genus            = rep(NA, length(unique(input_codes))),
+                           species          = rep(NA, length(unique(input_codes))),
+                           stringsAsFactors = FALSE)
+                    
+        for(i in 1:nrow(unique.tmp)){
+          foo <- eval(parse(text =paste0("traits::betydb_query(",
```
istfer: @ashiklom I want to use the `traits` package in this function but the development version (`install_github("ropensci/traits")`) how to import it then?

PR #1251, opened 2017-02-03
Initial moving of documentation files to book_source

tonygardella:
Moving markdown files from gitbook to Rmarkdown files here in ```book_source```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1271, opened 2017-03-14
CRUNCEP minor revisions

ashiklom:
Add error message if outside CRUNCEP bounds (yes, ideally, it's better to put it earlier in the workflow, but can't have too many failsafes!).

### DIFF
--- a/modules/data.atmosphere/R/download.CRUNCEP_Global.R
+++ b/modules/data.atmosphere/R/download.CRUNCEP_Global.R
@@ -16,6 +16,16 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
   end_date <- as.POSIXlt(end_date, tz = "UTC")
   start_year <- lubridate::year(start_date)
   end_year <- lubridate::year(end_date)
+
+  # Check that the start and end date are within bounds
+  CRUNCEP_start <- 1901
+  CRUNCEP_end <- 2010
+  if (start_year < CRUNCEP_start | end_year > CRUNCEP_end) {
+    PEcAn.utils::logger.severe(sprintf('Input year range (%d:%d) exceeds the CRUNCEP range (%d:%d)',
+                                       start_year, end_year,
+                                       CRUNCEP_start, CRUNCEP_end))
+  }
+
   site_id <- as.numeric(site_id)
 #  outfolder <- paste0(outfolder, "_site_", paste0(site_id%/%1e+09, "-", site_id %% 1e+09))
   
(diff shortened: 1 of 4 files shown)


modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -1,30 +1,39 @@
-##' Download and conver to CF CRUNCEP single grid point from MSTIMIP server using OPENDAP interface
```
mdietze: convert

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -1,30 +1,39 @@
...
+
   site_id <- as.numeric(site_id)
 #  outfolder <- paste0(outfolder, "_site_", paste0(site_id%/%1e+09, "-", site_id %% 1e+09))
   
   lat.in <- as.numeric(lat.in)
   lon.in <- as.numeric(lon.in)
-  lat_trunc <- floor(2 * (90 - as.numeric(lat.in))) + 1
-  lon_trunc <- floor(2 * (as.numeric(lon.in) + 180)) + 1
+  # Convert lat-lon to grid row and column
+  lat_grid <- floor(2 * (90 - lat.in)) + 1
```
mdietze: Is there a reason for dropping the as.numeric? I believe @jsimkins2 had it there since sometimes values are being passed in as character strings

ashiklom: Because the same thing happens two lines above it, if you look at the whole file. These `as.numerics` are completely redundant.

modules/data.atmosphere/R/download.CRUNCEP_Global.R
ashiklom: This is the deleted line, which I fixed below. I abbreviated the formal title of each function to just the download part, and maintained the subsequent line (the description field) as the full, original description.

ashiklom:
I also added the same quickfixes to the GLDAS and NLDAS download functions, while this is fresh on my mind. I gotta run for the day, but I can skim through the remaining download functions tomorrow morning.

mdietze:
I don't think we want the same sort of hard-coded end-date restrictions on NLDAS, GLDAS, NARR, etc since they are continually updated on a rolling basis.

modules/data.atmosphere/R/download.CRUNCEP_Global.R
ManosGEM: @mdietze  I agree with @ashiklom , the variables are already converted to numeric , so there's no need for this.

ashiklom:
I didn't put end date restrictions on NLDAS or NARR, only start-date restrictions. Our documentation says GLDAS ends at 2010, but I checked the [product website][prod] and it does in fact go through the present, so that documentation will need to be fixed. @tonygardella We should get #1251 merged so we can start updating documentation in these sorts of pull requests.

[prod]: https://disc.gsfc.nasa.gov/uui/datasets/GLDAS_NOAH10SUBP_3H_V001/summary?keywords=GLDAS_NOAH10SUBP_3H_V001

mdietze:
@ashiklom I'll merge #1251 once you give it a thumbs up that it's ready

PR #1277, opened 2017-03-16
More documentation revisions

ashiklom:
These changes add most of the basic user's guide to the documentation (it was currently missing). They also re-organize that part of the documentation to render better, and to match what happens when a user actually goes through the web interface.


tonygardella:
@ashiklom The pages were not being used because the outline for documentation on the basic user's guide that was discussed was : 
Downloading and setting up your VM
Point to the demo #1
Add Video demonstration

Those pages were left out because a lot of that information is in the Demo Instructions already and to avoid doubling up work updating the demo pages and the basic user's guide every time we change the interface we just point to the Demo on the webpage. 

I think these pages could be useful but there is information in the Demo pages that I think could be appropriate here so I think consolidating the Demo and these basic user guide pages should happen if we use these pages.

ashiklom:
Is there a particularly good reason we don't just keep the Demo in the documentation? We should discuss this at the next PEcAn call. I feel like we should just deprecate the Demos and move them into the documentation, so we can keep them updated live. Especially now that the documentation can be linked directly to PEcAn releases via git tags.

tonygardella:
I believe it's because it allows a user to have a pdf to download without having to download all the documentation. Do you think we could have bookdown render the basic user's guide into a pdf on demand?

mdietze:
@ashiklom I was about to post the same point. We've slowly been moving more and more of the Demos out of Word and into Rmd each time we revise them, but hadn't taken the full plunge, in part because Rmd demos would have still been separate from our previous documentation modes. Now that our documentation itself can support Rmd (rather than the wiki and GitBook, which were just md), that really creates a strong incentive to provide better integration across the two. Let's add that to today's agenda -- e.g. do we fully integrate the demos, or leave them in a separate directory but inline them into the documentation as well (which I'd be in favor of if we could pull it off)

ashiklom:
@tonygardella It should be possible to set up bookdown's "download" button to allow downloading of a PDF copy, and since we're hosting on a separate repo and don't really care a ton about space, there's no reason we couldn't also render a PDF with each build and store it in a publicly accessible location. 

But I'm not convinced people will really need a PDF. The only reason I see for having it is if they were working offline, and it's possible to view the documentation as an HTML offline anyway (just clone the documentation repository and open the `index.html` file locally).

Actually, we could easily make PEcAn's documentation path be configurable (via `config.php`) and then have the VM and our servers point to a locally rendered version of the documentation corresponding to the version of PEcAn they are currently using (and have the git version of the documentation continue to live online).

ashiklom:
@mdietze Unless I'm mistaken, this seems like it's ready to be merged. It would be good to have our documentation be live for NACP.

PR #1278, opened 2017-03-16
dump new tables PecanProject/bety#403 PecanProject/bety#491

robkooper:
dump missing tables from migrations up to BETY 4.15

### DIFF
--- a/scripts/dump.bety.sh
+++ b/scripts/dump.bety.sh
@@ -100,7 +100,7 @@ USER_TABLES="users"
 # will be imported during creaton
 CLEAN_TABLES="benchmark_sets benchmarks"
 CLEAN_TABLES="${CLEAN_TABLES} citations covariates cultivars dbfiles"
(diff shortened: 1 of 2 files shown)


serbinsh:
@robkooper How do we implement this?  Just update PEcAn and when our systems run syncs everything is up to date or do we want to run anything manually?

robkooper:
You can just copy the dump-bety.sh and load-bety.sh script and run that, no need to really update pecan. But until people use this new version of the scripts it will not copy the right tables.

serbinsh:
Ok.  I pulled in the latest scripts which should start running with the cronjobs at BNL

[iPhone]

On Mar 17, 2017, at 10:40 AM, Rob Kooper <notifications@github.com<mailto:notifications@github.com>> wrote:


You can just copy the dump-bety.sh and load-bety.sh script and run that, no need to really update pecan. But until people use this new version of the scripts it will not copy the right tables.

ankurdesai:
Same here (also updated Bety and migrated in the experiments table). Running sync job right now.

PR #1289, opened 2017-03-20
Use explicit arguments in `get.trait.data` `lapply`

ashiklom:
Currently, these arguments are passed by order, which is just asking for trouble.


ashiklom:
Hold off on pulling this until I say so -- I'd like to test that everything still works first. I'm opening the PR so people can comment if they wish.

PR #1294, opened 2017-03-24
Replace ifelse with if because of multi-element return value

ankurdesai:
Bug fix 
## Description
ifelse statements do not return more than one element if the conditional is only one element. 

## Motivation and Context
Makes metgapfill crash. Now it works

## Types of changes


bpbond:
Note that R's `ifelse` is a demon spawn and should pretty much never be used. IMHO.

Issue #994, opened 2016-08-18
download.file and FTP gateway issues


serbinsh:
@robkooper 

I have been running into new issues with FTPs from within BNL that relate to our firewall and the need to use a FTP gateway when connecting to an external FTP such as NARR or fluxnet, but not Ameriflux which goes over https.  The issue is that I need to connect via an FTP gateway or it can't reach the external FTP site:

```
> remotefunc <- function() {
+ PEcAn.data.atmosphere::download.NARR('/data/Model_Output/pecan.output/dbfiles/NARR','2004/01/01','2004/12/31')
+ }
> remoteout <- remotefunc()
Loading required package: PEcAn.utils
Loading required package: abind
Loading required package: plyr
Loading required package: XML
Loading required package: ggplot2
Loading required package: randtoolbox
Loading required package: rngWELL
This is randtoolbox. For overview, type 'help("randtoolbox")'.
Loading required package: RCurl
Loading required package: bitops
Loading required package: lubridate

Attaching package: 'lubridate'

The following object is masked from 'package:plyr':

    here

The following object is masked from 'package:base':

    date

2016-08-18 17:17:59 DEBUG  [PEcAn.data.atmosphere::download.NARR] : 
   File ' /data/Model_Output/pecan.output/dbfiles/NARR/pres.sfc.2004.nc ' 
   already exists, skipping to next file. 
trying URL 'ftp://ftp.cdc.noaa.gov/Datasets/NARR/monolevel/dswrf.2004.nc'
Error in download.file(url, new.file) : 
  cannot open URL 'ftp://ftp.cdc.noaa.gov/Datasets/NARR/monolevel/dswrf.2004.nc'
Calls: remotefunc -> <Anonymous> -> download.file
Execution halted
[1] 1
Error in new.files$file : $ operator is invalid for atomic vectors
Calls: do.conversions ... <Anonymous> -> dbfile.input.insert -> basename -> dirname
> proc.time()
   user  system elapsed 
  4.843   0.297   5.399 

```

I noticed that download.file()  uses curl.  

```
./web/PEcAn.Network.R:    download.file(sync.url,destfile=temporaryFile, method="curl")
```

I wonder if you might have an idea of what to use instead on modex to get around FTP gateways?  Maybe it isn't possible.  I am trying to find out how to connect in one command but also using our FTP gateway, in which case I would just need to setup a special case of the download.file() function that would run for networks like ours....

robkooper:
[...]

PR #1296, opened 2017-03-24
Modex.ftp.update

serbinsh:
Updated data.atmosphere to allow for download of AmerifluxLBL data through the BNL FTP gateway.

## Description

Added function download.file to utils/R/utils.R to allow for the use of ncftget

Updated download.AmerifluxLBL() to select between base R download.file and PEcAn.utils::download.file() which uses ncftpget when on modex behind a FTP gateway

Added global FTP gateway params to modex.bnl.gov to provide correct internal FTP firewall host and parameters within BNL

## Motivation and Context

On modex behind a firewall we could not download files from ftp sites

Addresses #994 

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue) 

## Checklist:

- [ X] My change requires a change to the documentation.
- [X ] I have updated the documentation accordingly.


serbinsh:
I would like to think about adding general options that can be set in the pecan.xml to allow for the use of alternatives to R::base::download.file() for situations like FTP sites when the requesting host is behind a gateway.  That way we could have an option to put in optional inputs that get used in the download commands, such as downloading AmerifluxLBL.  Then we can get rid of my hard-coded options and replace with general solution...but I am not sure where best that would go in the tags

modules/data.atmosphere/R/download.AmerifluxLBL.R
```diff
@@ -92,7 +95,11 @@ download.AmerifluxLBL <- function(sitename, outfolder, start_date, end_date,
   
   if (download_file_flag) {
     extract_file_flag <- TRUE
-    utils::download.file(ftplink, output_zip_file)
+    if (method != "ncftpget") {
+    	utils::download.file(ftplink, output_zip_file)
+    } else {
+    	PEcAn.utils::download.file(ftplink, output_zip_file, method=method)
```
mdietze: If PEcAn.utils::download.file also has to do a check against method, why have this check here? Why not just always call PEcAn.utils::download.file and then inside that function have it use utils::download.file as the default?

modules/data.atmosphere/R/download.AmerifluxLBL.R
```diff
@@ -92,7 +95,11 @@ download.AmerifluxLBL <- function(sitename, outfolder, start_date, end_date,
   
   if (download_file_flag) {
     extract_file_flag <- TRUE
-    utils::download.file(ftplink, output_zip_file)
```
robkooper: We can just always use `PEcAn.utils::download.file(ftplink, output_zip_file, method=method)` then in there put the logic to switch, that way we can use the same code everywhere.

modules/data.atmosphere/R/download.AmerifluxLBL.R
```diff
@@ -19,10 +19,13 @@
...
 ##' @author Ankur Desai, based on download.Ameriflux.R by Josh Mantooth, Rob Kooper
 download.AmerifluxLBL <- function(sitename, outfolder, start_date, end_date, 
-                                  overwrite = FALSE, verbose = FALSE, username = "pecan", ...) {
+                                  overwrite = FALSE, verbose = FALSE, username = "pecan", method=NULL, ...) {
   # get start/end year code works on whole years only
   
-  
+  # set hostname to select appropriate download
+  #if (is.null(host)) host <- system("hostname", intern=T)
+  if (system("hostname", intern=T)=="modex") method <- "ncftpget"
```
mdietze: Not a good idea to bury machine-specific hacks into low-level functions. If method is an argument to this function, why not do this check outside the function and pass the correct value of method in? Otherwise we don't have a system that works the next time someone is behind a firewall

utils/R/utils.R
```diff
@@ -665,6 +665,25 @@ convert.expr <- function(expression) {
...
+##' Requires ncftpget and a properly formatted config file in the users 
+##' home directory
+##' @title download.file
+##' @param url complete URL for file download
+##' @param destfile destination file name
+##' @param method Method of file retrieval.  Currently the only option is ncftpget
+##' @export
+##' @author Shawn Serbin
+download.file <- function(url, destfile, method="ncftpget") {
+   if (method=="ncftpget") {
```
robkooper: leave method blank, then add `method <- if (missing(method)) getOption("download.file.method", default = "auto")` in the else we can call utils::download.file(ftplink, output_zip_file). We might want to add a special `download.ftp.method` so we can switch to that if it starts with ftp://

modules/data.atmosphere/R/download.AmerifluxLBL.R
serbinsh: OK, I agree....this was an attempt to get things working and get feedback on options to generalize

modules/data.atmosphere/R/download.AmerifluxLBL.R
serbinsh: I agree working to generalize

utils/R/utils.R
serbinsh: @robkooper but where would we then set method?  in the download.XX.R files?  That is at the function call?

I could add a switch that if the ftplink (or similar) link has ftp:// the use download.ftp.method

mdietze:
@serbinsh just a reminder that this PR is still open and waiting for changes

tonygardella:
@serbinsh Is this PR still in progress?

PR #1305, opened 2017-03-29
(WIP, PROPOSAL) Change build mechanism to devtools::remotes

ashiklom:
The `devtools::install` function provides a convenient, more R-centric
way to manage package [dependencies][deps]. In the first commit, I
demonstrate the way this works for the `PEcAn.all` package, namely by
specifying the dependencies in the `DESCRIPTION` file. Once those are
set, running `devtools::install('/path/to/pecan/all')` will install all
dependencies, and, in a `make`-like fashion, _skip packages that haven't
been modified_.

**Advantages over current "make" system**:
    - May bypass some compatibility issues and paths. Because `make` is
    external and uses `sh` for POSIX compliance, it may understand
    different environment variables than R.
    - Dependencies are internal to package, and are more logically
    nested. No need to set all dependencies in a central, external
    makefile.

What do you all think?

[deps]: https://cran.r-project.org/web/packages/devtools/vignettes/dependencies.html

### DIFF
--- a/all/DESCRIPTION
+++ b/all/DESCRIPTION
@@ -35,8 +35,21 @@ Suggests:
     PEcAn.LINKAGES,
     PEcAn.allometry,
     PEcAn.photosythesis
+Remotes:
+    local::../db,
+    local::../settings,
+    local::../modules/meta.analysis,
+    local::../utils,
+    local::../modules/uncertainty,
+    local::../modules/data.atmosphere,
+    local::../modules/data.land,
+    local::../modules/data.remote,
+    local::../modules/assim.batch,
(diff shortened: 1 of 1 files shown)


mdietze:
Dunno, I'm pretty happy with the current make system. Would this system also have the flexibility to build other things (beyond just the packages, like documentation, man, etc) and to install subsets of packages? I really like being able to build just a single model's package on a remote host and letting make sort out what packages need to be downloaded and which other PEcAn packaged need to be built -- much nicer than building "all", which installs tons of packages you never need.  Seems to be in the 'if it ain't broke' category to me.

PR #1312, opened 2017-03-31
Add NPP to BioCro output

dlebauer:
This is untested. But I added a draft of computing NPP 

## Description

this is the default variable that PEcAn uses, and it seems useful ... 

## Types of changes

## Checklist:


models/biocro/R/model2netcdf.BIOCRO.R
```diff
@@ -78,7 +79,7 @@ model2netcdf.BIOCRO <- function(result, genus = NULL, outdir, lat = -9999, lon =
                        Evap = udunits2::ud.convert(SoilEvaporation + CanopyTrans, "Mg/ha/h", "kg/m2/s"), 
                        TVeg = udunits2::ud.convert(CanopyTrans, "Mg/ha/h", "kg/m2/s"), 
                        LAI = LAI))
-    
+    RR$NPP <- udunits2::ud.convert(c(0, diff(R$TotLivBiom)), "kg/m2/h", "kg/m2/s")
```
mdietze: 1) @infotroph wnits need to adhere to the standard -- we can't have different models using different units: http://nacp.ornl.gov/MsTMIP_variables.shtml

2) Note that the standard is kg C m-2 s-1 -- just wanted to double check that the above is C not biomass

dlebauer:
* The link from @infotroph about BioCro's respiration giving nonsense refers to soil respiration, not Ra. 
* BioCro - like most models - ignores herbivory. It also ignores leaf and root turnover. Biomass increment is A-Ra.
* We capture harvest. Harvest and senescence can be accounted for by only taking the positive differences. Within the season the leaves do not fall off.

* with respect to "We need consistent variable definitions across models". What if some models simulate herbivory and others don't? I think it is clear that the role of PEcAn is not to police model structure. Can't we leave this to model skill assessment (benchmarks) and assume that this is part of model structure? If BioCro wrote out NPP rather than would it raise an eye? I am pretty sure it would not change the estimate.

Finally, even if it were possible to model NPP, it could never be be validated. In practice, measurements of NPP are (almost always) positive biomass increments and (almost always) only done for aboveground biomass. In forests biomass increments are estimated from DBH. Roots are rarely measured. At least with BioCro, the senesced leaves do not fall off.

models/biocro/R/model2netcdf.BIOCRO.R
infotroph: 1. Got it, and making a note to check the standard any time I touch an output converter.

2. It looks like this is biomass not C -- note that the k multiplier includes a division by `c2biomass` ( == 0.4).

mdietze: @infotroph thanks for pointing out the k multiplier -- given that it looks like a lot of the existing variables are in the wrong units (biomass not carbon). It also looks like this specific line will fail since RR$NPP is being calculated from R$TotLivBiom but that TotLivBiom is only in RR not R.

mdietze:
@dlebauer I am not trying to police model structure, I'm trying to police the well-agreed definition than NPP = GPP - Ra. Indeed, it is not my job to know the detailed internal structure of every single model -- I need to reply on the developers to explain what they're doing. Everything in the prior discussion from @infotroph clearly implied that the implemented calculation was an **approximation** and that BioCro included respiration. You have clarified that BioCro doesn't have a concept of autotrophic respiration or turnover, and while I find that astonishing I'm not going to police that model structure.

I'm happy to lift my "request changes" once:

1. Units are fixed to be C not Biomass as appropriate on a variable-by-variable basis
2. RR vs R bug is resolved (refraining from a request for more meaningful variable names)
3. NPP calculation is updated to exclude harvest (which you say requires just using the positive differences)

dlebauer:
I'll fix this up but briefly 
* I was wrong about litter so will add AboveLitter and BelowLitter to the calculation 
* BioCro does account for Ra. Sorry if that wasn't clear
* All models are approximations

dlebauer:
I've updated the code to address the changes requested by @mdietze . Now NPP = positive increments in biomass + litter. In the future we can modify BioCro to compute NPP directly as GPP - Ra, and update the function at that point.

Currently failing with what looks like an error caused by [`deploy.sh`](https://travis-ci.org/PecanProject/pecan/builds/218206735)

models/biocro/R/model2netcdf.BIOCRO.R
```diff
@@ -71,14 +72,19 @@ model2netcdf.BIOCRO <- function(result, genus = NULL, outdir, lat = -9999, lon =
     
     k <- udunits2::ud.convert(1, "Mg/ha", "kg/m2") / c2biomass
```
mdietze: Not following why k is divided by c2biomass since most (possibly all) of the variables multiplied by k are supposed to be reported on a carbon basis, not a biomass basis. If BioCro internal pools are on a biomass basis, you should multiply by c2biomass. If they're already on a carbon basis, you don't need this.

models/biocro/R/model2netcdf.BIOCRO.R
```diff
@@ -71,14 +72,19 @@ model2netcdf.BIOCRO <- function(result, genus = NULL, outdir, lat = -9999, lon =
...
+    result_yeari_std <- with(result_yeari, list(
+      TotLivBiom = k * (Leaf + Root + Stem + Rhizome + Grain), 
+      RootBiom = k * Root, 
+      StemBiom = k * Stem, 
+      Evap = udunits2::ud.convert(SoilEvaporation + CanopyTrans, "Mg/ha/h", "kg/m2/s"), 
+      TVeg = udunits2::ud.convert(CanopyTrans, "Mg/ha/h", "kg/m2/s"), 
+      LAI = LAI))
     
+    total_biomass <- with(result_yeari, 
+                          k * (TotLivBiom + Leaf + Root + Stem + Rhizome + Grain + AboveLitter + BelowLitter))
```
mdietze: 1) If TotLivBiom is Leaf + Root + Stem + Rhizome + Grain, doesn't that cause these terms to be double counted?
2) It's still unclear to me whether TotLivBiom is already a column in results (in which case, why redefine it above in result_yeari_std) or whether this calculation should really be done on result_yeari_std not result_yeari?

models/biocro/R/model2netcdf.BIOCRO.R
```diff
@@ -70,7 +67,8 @@ model2netcdf.BIOCRO <- function(result, genus = NULL, outdir, lat = -9999, lon =
                  TVeg = mstmipvar("TVeg", x, y, t), 
                  LAI = mstmipvar("LAI", x, y, t))
     
-    k <- udunits2::ud.convert(1, "Mg/ha", "kg/m2") / c2biomass
+    biomass2c <- 0.4
+    k <- udunits2::ud.convert(1, "Mg/ha", "kg/m2") * bioimass2c
```
mdietze: typo: biomass2c vs. bioimass2c

dlebauer: doh!

PR #1316, opened 2017-04-03
make sure we get hours back from timezone

robkooper:
the timezone in bety is now a string, not offset in hours. This will convert it from text to hours.

Did not see a function in lubridate that could do this but it is in the stringi package.


mdietze:
Note that in #1293 I had a completely different solution to the same problem. Would be good to be consistent. Mine seems a bit more kudgy, but doesn't require an extra library.

robkooper:
@mdietze question, which is correct?
```
lst = "America/New_York"
as.numeric(as.POSIXct("1970-01-01 00:00:00", tz=lst))/3600
# 5
tryCatch(stringi::stri_timezone_info(lst)$RawOffset,
                    error=function(e) NaN)
# -5
```

The offset from New York to UTC is -5, but should the code at https://github.com/mdietze/pecan/blob/master/modules/data.atmosphere/R/metgapfill.R#L191 be 5 or -5 in that case?

mdietze:
-5 is correct. Guess I have a sign error in mine -- hopefully haven't generated too many 'wrong' met files in the 1.5 weeks since my fix (but we should check the db for Ameriflux and fluxnet met over that period and just delete the gap filled and model records & files)

PR #1323, opened 2017-04-05
More tests for load.cfmet

infotroph:
## Description
Mostly checking that start and end dates are handled properly.

## Motivation and Context

Note that six(!) of these tests currently fail -- careful review requested to make sure my expectations line up with load.cfmet's design intentions.

Specific questions that should be answered before merging: 
* `load.cfmet(..., start.date, end.date)` should *always* return a dataframe that starts on `start.date` and ends on `end.date`, or throw an error if these aren't present in the input, correct?
* The lat/lon tests fail because `load.cfmet` continues execution after it notices a mismatch. Should it stop instead?
* Would it be better to move all tests of load.cfmet to a separate file named `test.load.cfmet.R`?

## Types of changes

- [x] New test

## Checklist:


mdietze:
@infotroph it's a common misconception that logger.error throws a fatal error rather than an error message. If you want a fatal error (i.e. stop), then you need to use logger.severe.  There are 4 logger functions altogether: info, warn, error, and severe. There's also a function to set the level of warnings you want to see (e.g. you can easily suppress info and warnings so you get less output). It's possible (likely?) that we might want to reconsider the name of the 3rd level to avoid this misconception. Suggestions?

infotroph:
@mdietze Got it, thanks. I don't mind "error" as the name here -- it lines up with my understanding of OS log levels, where I tend to assume error means "there's a problem in local scope, but don't kill the whole app (or PEcAn run) yet." The question then is whether `load.cfmet` is handling the problem as designed. But I'm not being the one who has to fix everyone's misconceptions, soooo... :)

I've edited my second question to clarify that logger behavior isn't at fault here.

infotroph:
@dlebauer thanks. Don't pull yet, I'll move tests to a new file.

PR #1326, opened 2017-04-06
testing build system tests

dlebauer:
do not merge

just testing that an error in the testing is caught by Travis CI


dlebauer:
@robkooper FYI

PR #1331, opened 2017-04-07
Add back to PEcAn support output pecan internal format for data

yan130:
Add back to PEcAn support output pecan internal format for data

## Description
Add code to PEcAn to add a new model called PEcAn so we can do a met2model.PEcAn which will be just a copy of the metfile passed in

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

## Checklist:


robkooper:
Background information about this pull request. In the past it was possible to download the intermediate file format in BrownDog, and use this in your code, or pass it back into BrownDog. After we refactored the code in BrownDog to use the met.process code, we can no longer download the intermediate file format. Adding this pseudo model will allow us to download the intermediate file format again.

mdietze:
I agree this shouldn't be a whole new package. I also wonder if there's a more elegant solution than adding a 'pseudomodel' to the workflow. Finally, I don't see how this would help the non-PEcAn user -- if someone wanted to use BD to convert some given met file into netCDF CF (or to have BD download that file and then convert it), it would be highly unintuitive to ask for it in 'PEcAn' (i.e. this standard isn't ours, it's a common standard which we've adopted).

robkooper:
@mdietze Any suggestions for the name? CF, so we can use met2model.CF.R?

mdietze:
@robkooper CF is a much better name. 

Also, I don't think you actually need to create an empty met2model.CF.R -- if I understand what you're saying you just need a registry file for CF, which is set to indicate that the pseudo-model "CF" uses CF standard met as it's input, so that the workflow just skips the met2model step altogether. This already happens for some of the other models that can read CF directly.

mdietze:
Oh, and you'll also need to change met.process to check the data.atmosphere package for the model registry file if there isn't a model package to get the file from, but that should be a one-line fix.

yan130:
@robkooper 
I update modules/data.atmosphere/R/met2model.module.R, this is the function to convert a.pecan.zip/nc to a.clim/ed.zip/...., right?

IF so, I don't think a single file of a.pecan.zip can provide all input parameter for met2model.module.R. for example start_date and end_date.

robkooper:
Depends, the pecan.zip should contain the nc files and we can use the start date/end dates of the nc files to find the range it covers. This range is then the tart_date and end_date.

PR #1334, opened 2017-04-11
PalEON regional sites & sitegroups

mdietze:
For archive and example purposes, doesn't need to be rerun by anyone (sites and groups are loaded on BU server and will propagate across the network). PalEON sitegroups added to facilitate running 'pseudoregional' runs at different resolutions.


modules/data.land/inst/LoadPalEONsites.R
```diff
@@ -0,0 +1,93 @@
...
+library("RCurl")
+library("XML")
+library(PEcAn.DB)
+
+nu <- function(x){as.numeric(as.character(x))}  ## simple function to convert data to numeric
+
+site.map <- data.frame(FLUX.id=rep(NA,2000),site.id=rep(NA,2000))
+
+## load up PalEON site files
+paleon <- read.csv("~/paleon/Paleon_MIP_Phase2_ED_Order_Status.csv",stringsAsFactors = FALSE)
```
robkooper: are these files available somewhere?

PR #1335, opened 2017-04-11
Fix biocro tests

infotroph:
Fixes #1320, #1328.

## Description
* Updated run.biocro and model2netcdf to retain litter variables in output
* Saved new version of sample biocro results. **CAUTION**: My change to result.RData adds dummy values (AboveLitter = BelowLitter = 0) rather than real output. If it's important for result.RData ought to contain a genuine representative BioCro result, this should be changed.
* Uncommented all tests
* Suppressed warning by not trying to unlink files that don't exist
* Suppressed password prompt/warning by changing run host to localhost

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1339, opened 2017-04-12
Catch testthat test failures in make

ashiklom:
Before, testthat errors were reported, but didn't cause `make test` to actually fail because the command itself still completed successfully.
Setting `reporter = "stop"` means that failing tests are treated as errors.

Also, running `devtools::document()` on all packages with the latest version of Roxygen made minor edits to a lot of documentation (e.g. removing trailing empty lines, updating roxygen version in DESCRIPTION files), which are also included here.


modules/emulator/NAMESPACE
```diff
@@ -4,6 +4,8 @@ S3method(calcSpatialCov,list)
 S3method(calcSpatialCov,matrix)
 S3method(p,jump)
 S3method(p,mvjump)
+S3method(plot,jump)
```
robkooper: is this correct?

ashiklom: That's a question for @mdietze , it's his function. It _looks_ like `jump` is an S4 class, so I don't think that's right. I changed it in cbdf8b0, but I can revert if necessary.

istfer:
sorry, I wasn't taking good care of this package, these functions are currently not being used, feel free to modify as you see fit

mdietze:
@ashiklom this was passing but just failed again after #1335

ashiklom:
@mdietze It's failing because #1335 added failing tests, to which Travis is now sensitive. So blame @infotroph , don't shoot the messenger :wink: 

https://travis-ci.org/PecanProject/pecan/builds/221486312#L2987

ashiklom:
@mdietze It's failing on a `data.atmosphere` test. See [line 2987 of the build log](https://travis-ci.org/PecanProject/pecan/builds/221486312#L2987).

PR #1353, opened 2017-04-14
Convert make test to make check for CI

dlebauer:
`make test` calls `devtoos::test`; `make check` calls `devtools::check`. 

`devtools::test` and `make test` are useful for development, but `devtools::check`, which calls R CMD Check, is more appropriate for continuous integration (according to devtools author [Hadley Wickham's text](http://r-pkgs.had.co.nz/check.html#check). `check`   integrates with the standard protocol for R package testing.

One side effect of using `test` is that it only calls files in `tests/testthat/test*.R` files. `check` calls `tests/testthat.R` which then runs `test_check` and checks the package. If we only use `test` for CI (which I don't think we should) the `testthat.R` files should be deprecated.

A benefit of using `check` is that it performs a standard suite of tests for R packages, and will ensure that the package itself is valid (not just the functions that it contains).

A downside is longer times for Travis CI. But ... if time is an issue perhaps we should parse `git diff` in make and only test the recent changes.

## Types of changes

## Checklist:


ashiklom:
Did you check to make sure that failing tests actually break the Travis build? AFAIK that is not the default behavior -- failing tests will be reported, but because the command itself completes successfully, the build will be listed as successful. This is the entire motivation behind my previous PR that added `reporter="stop"` to throw an error on the first failed test.

dlebauer:
@ashiklom Why do you think that this is not the default behavior? This is the recommended way to use Travis CI in the documentation http://r-pkgs.had.co.nz/check.html#travis. 

Currently the build is broken many times over by errors in R CMD Check that we have been unaware of since dropping its use. Here is an example where a failed test throws an error: https://travis-ci.org/PecanProject/pecan/builds/222208036#L3191.

ashiklom:
I thought it wasn't the default behavior because it's not the default behavior for `devtools::test`, but I may be wrong. I'm all for using `check`, especially since that's the official recommendation. I just wanted to make sure failing `testthat` tests actually broke the build. From the build log, looks like it hit several errors before the build actually broke and then actually died on a missing package rather than either of those. If it's smart enough to detect all the errors first and then fail at the end, that's awesome. But if it's informing us about errors but not actually breaking the build, that defeats the point.

infotroph:
Default devtools behavior is one thing, but note that we specifically ask Travis for *non-default* behavior. If I'm reading right, what's described in the Travis r documentation assumes *one* R package whose root is the repository root.  By invoking `make`, my understanding is that we override much of the R-specific Travis magic and need to handle it ourselves instead.

To illustrate, here's a [test repository with three active branches](https://travis-ci.org/infotroph/testtests/branches): 
* `native_build` is what I understand as a "default" setup. It uses the minimal travis config generated by `devtools::use_travis` and doesn't specify the test script. 
* `make-build` uses a Makefile that calls `devtools::test(".", reporter="stop")
* `make-check-build` uses a Makefile that calls `devtools::check(".")`

The four most recent commits are identical across all three branches. I submit that in PEcAn-land, we would probably want these commits to pass, fail, pass, fail -- so in this example the native build is the only one that gets it fully correct. Not sure how to generalize that to a multi-package repository, though.

ashiklom:
Thanks @infotroph for the comprehensive tests! If arguments are passed from `check` directly into `test`, it should just be a matter of adding the `reporter = "stop"` argument, analogously to how we just did `test`. Otherwise, we'll have to look for another way to get builds to fail.

infotroph:
A couple of thoughts, NOT tested yet:

1. Looks like the closest equivalent of `devtools::test(pkg, reporter="stop")` is to specify the reporter in testthat.R: `test_check(pkg, reporter="stop")`. Disadvantage: Still won't stop on warnings, plus need to enforce it in every package.
2. Keep track of the build path inside the Makefile, grep the log for 'WARNING' and 'ERROR', and exit with nonzero status as appropriate. I think this is [what Travis does](https://travis-ci.org/infotroph/testtests/builds/222363302#L640) in the native build. `devtools::check_failures` may make it easier to generate an in-Makefile version, but still need to know what path to point it toward.

dlebauer:
I don't think we want to stop for warnings... Probably a month of work
sorting those out ...

infotroph:
Actually! Maybe we're--or at least I am--overthinking this. The current make test works correctly, right? Why not 

```
 - make
 - echo "testing PEcAn"
 - make test
 - echo "checking for package issues" 
 - make check 
```

... and alter `make check` to pass `--no-tests`? That way a commit with failing tests can exit CI before wasting time on the checks.

ashiklom:
OK, so I wrote a script (`scripts/check_with_errors.R`) that runs some checks and fails loudly if there are any errors. By default, it will print the number of notes and warnings from `R CMD check` but will not print the actual warnings or notes, to avoid cluttering the console/Travis build log. Optionally, it can be more verbose about printing by setting the environment variable `LOGLEVEL` to `warn` or `all` to print warnings and notes, respectively. Similarly, it can be set to fail on warnings by setting the environment variable `DIELEVEL` to `warn`. For example, for full verbosity and rigor, run:

```
LOGLEVEL=all DIELEVEL=warn make check
```

I also pushed a bunch of commits with minor fixes that my local `make check` came across.

@dlebauer @robkooper @infotroph I think this is ready for your review.

modules/rtm/tests/testthat/test.invert_simple.R
```diff
@@ -1,6 +1,16 @@
...
+    library(PEcAnRTM)
 }
+
+
+# Skip computationally intensive test on TRAVIS
+travis <- Sys.getenv('TRAVIS') == 'true'
+if (travis) {
+    message('Skip computationally intensive inversion on Travis')
+} else { 
+
```
infotroph: Why not `skip_on_travis()`?

ashiklom: `skip()` just throws an error (effectively the same thing as `stop()`) when run outside of a `test_that(...)` block. The way these tests are set up, they first run simulations and then `test_that(...)` on the outputs.

infotroph: Gotcha. No objection to this approach, then.

modules/rtm/tests/testthat.R
```diff
@@ -3,4 +3,8 @@ library(testthat)
 library(PEcAnRTM)
 
 options(warn=1)
-test_check("PEcAnRTM")
+if (Sys.getenv('TRAVIS') == 'true') {
```
dlebauer: FYI see also `testthat::skip_on_travis()`

ashiklom:
@infotroph Re: [Your Gitter comment](https://gitter.im/PecanProject/pecan?at=590f6bc573b39c5f15739401): Not sure exactly whe c115f4e is being "undone", but I suspect it's because that commit isn't quite correct. `merge.met.variable` is not an S3 method (at least I don't think it is), but Roxygen thinks it is a method of `merge` unless you explicitly do `@export merge.met.variable`, as I did in 6f948dd.

infotroph:
@ashiklom That makes sense, thanks.

Should this be merged? I realize we haven't achieved `make check` yet, but it'd be nice to get all the DESCRIPTION and documentation fixes into develop.

ashiklom:
Yes, this should be merged. @mdietze or @robkooper, can you do a quick review and merge it in?

mdietze:
FWIW merge.met.variable was renamed merge_met_variable in my pending PR, to avoid the S3 method issue, but I'll resolve that conflict later

PR #1354, opened 2017-04-17
PEcAnRTM: Initialisze RNG on parallel cluster

ashiklom:
Otherwise, chains start with same seed and can end up identical if their initial conditions are the same.
This uses `parallel::setClusterRNGSeed` to set a different seed on each chain.

This hasn't been a problem for PROSPECT inversions because initial conditions are almost always generated randomly.
It has become a problem for EDR because it is often convenient to set static initial conditions.


serbinsh:
Sorry I didnt see this earlier but I like this update

PR #1356, opened 2017-04-17
Temporally Downscale Meteorology

jsimkins2:
Linear regression based temporal downscaling routines. All part of the new 'tdm' family. Uses CF convention and netCDF files throughout entire workflow. Vignette to come next.


modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+    names(dat.train) <- vars.info$CF.name
+    dat.train <- data.frame(dat.train)
+    
+    dat.train$date <- strptime(paste(dat.train$year, dat.train$doy + 1, 
+        dat.train$hour, sep = "-"), "%Y-%j-%H", tz = "GMT")
+    dat.train$time.hr <- as.numeric(difftime(dat.train$date, paste0((min(dat.train$year) - 
+        1), "-12-31 23:00:00"), tz = "GMT", units = "hour"))
+    dat.train$time.day <- as.numeric(difftime(dat.train$date, paste0((min(dat.train$year) -
```
mdietze: I don't follow these date processing steps -- standard CF files should just have one time variable named _time_. The variables year, doy, and hour are non-standard. I also don't really get the exact date manipulation you're doing here -- when revising to just use _time_ please document better.

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+    
+    dat.train$date <- strptime(paste(dat.train$year, dat.train$doy + 1, 
+        dat.train$hour, sep = "-"), "%Y-%j-%H", tz = "GMT")
+    dat.train$time.hr <- as.numeric(difftime(dat.train$date, paste0((min(dat.train$year) - 
+        1), "-12-31 23:00:00"), tz = "GMT", units = "hour"))
+    dat.train$time.day <- as.numeric(difftime(dat.train$date, paste0((min(dat.train$year) - 
+        1), "-12-31 23:00:00"), tz = "GMT", units = "day")) - 1/24
+    dat.train$time.day2 <- as.integer(dat.train$time.day + 1/(48 * 2)) + 
+        1  # Offset by half a time step to get time stamps to line up
+    dat.train <- dat.train[order(dat.train$time.hr), ]
```
mdietze: why would you possibly need to reorder the met data?

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+        "air_pressure", "specific_humidity", "wind_speed")], by = dat.train[, 
+        c("year", "doy")], FUN = mean)
+    names(train.day)[3:9] <- c("air_temperature_mean.day", "precipitation_flux.day", 
+        "surface_downwelling_shortwave_flux_in_air.day", "surface_downwelling_longwave_flux_in_air.day", 
+        "air_pressure.day", "specific_humidity.day", "wind_speed.day")
+    train.day$air_temperature_max.day <- aggregate(dat.train[, c("air_temperature")], 
+        by = dat.train[, c("year", "doy")], FUN = max)$x
+    train.day$air_temperature_min.day <- aggregate(dat.train[, c("air_temperature")], 
+        by = dat.train[, c("year", "doy")], FUN = min)$x
+    summary(train.day)
```
mdietze: If you feel the need to report this, I'd suggest sending it through logger.info since it's pretty non-essential.

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+    
+    
+    dat.train <- merge(dat.train[, ], train.day, all.x = T, all.y = T)
+    summary(dat.train)
```
mdietze: same comment about logger.

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+    # midnight NOTE: because we're filtering from the present back through
+    # the past, -1 will associate the closest hour that we've already done
+    # (midnight) with the day we're currently working on ----------
+    vars.hour <- c("air_temperature", "precipitation_flux", "surface_downwelling_shortwave_flux_in_air", 
+        "surface_downwelling_longwave_flux_in_air", "air_pressure", "specific_humidity", 
+        "wind_speed")
+    vars.lag <- c("lag.air_temperature", "lag.precipitation_flux", "lag.surface_downwelling_shortwave_flux_in_air", 
+        "lag.surface_downwelling_longwave_flux_in_air", "lag.air_pressure", 
+        "lag.specific_humidity", "lag.wind_speed")
+    lag.day <- dat.train[dat.train$hour == 23, c("year", "doy", "time.day2",
```
mdietze: Does this step assume that the training data is hourly?

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
+##' Create statistical models to predict subdaily meteorology from daily means
```
mdietze: Probably fine for a first pass, but why do we assume the input data to be downscaled is daily? e.g. is there anything in this scheme that would categorically preclude extending the approach to downscaling 6hr reanalysis?

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+##' @family tdm - Temporally Downscale Meteorology
+##' @title gen_subdaily_models
+##' @param outfolder - directory where models will be stored *** storage required varies by size of training dataset, but prepare for >100 GB
+##' @param dat.train_file - train_data file
```
jsimkins2: This is just `in.path`, but I didn't want to declare it as such because in `tdm_predict_subdaily_met`, we need to read in the `dat.trian_file` again, as `in.path` and `in.prefix` will be used by the coarse data we want to downscale

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
jsimkins2: It should be fine for downscaling 6hr reanalysis. The only thing that might not make sense are the names, (we have lag.day, next.day, etc.) but these are just the previous and next values, so as long as the dates are built properly it should be fine. I'll change this "daily" part

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+##' @param in.prefix 
+##' @param tdf_file - tdm_temporal_downscale_functions.R filepath that can be sourced i.e. '~/scripts/temporal_downscale_functions.R'
```
jsimkins2: Got it. I'll drop this

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+##' @param n.beta - number of betas to save from linear regression model
+##' @param resids - logical stating whether to pass on residual data or not
+##' @param parallel - logical stating whether to run temporal_downscale_functions.R in parallel 
+##' @param n.cores - deals with parallelization
+##' @param day.window - integer specifying number of days around a particular day you want to use data from for that
```
jsimkins2: We need our times to match up at the same time of day as the previous one to generate the statistics, so it has to be in a daily interval so we can pull stats from 13:00 on both days for example

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+##' @param outfolder - directory where models will be stored *** storage required varies by size of training dataset, but prepare for >100 GB
+##' @param dat.train_file - train_data file
+##' @param in.prefix 
+##' @param tdf_file - tdm_temporal_downscale_functions.R filepath that can be sourced i.e. '~/scripts/temporal_downscale_functions.R'
+##' @param n.beta - number of betas to save from linear regression model
+##' @param resids - logical stating whether to pass on residual data or not
+##' @param parallel - logical stating whether to run temporal_downscale_functions.R in parallel 
+##' @param n.cores - deals with parallelization
+##' @param day.window - integer specifying number of days around a particular day you want to use data from for that 
+##'                     specific hours coefficients
```
jsimkins2: Got it. I''ll add it today

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,226 @@
...
+    vars.info <- data.frame(CF.name = c("date", "year", "doy", "hour", 
+        "air_temperature", "precipitation_flux", "air_temperature_max", 
+        "air_temperature_min", "surface_downwelling_shortwave_flux_in_air", 
+        "surface_downwelling_longwave_flux_in_air", "air_pressure", "specific_humidity", 
+        "eastward_wind", "northward_wind", "wind_speed"))
+    dat.train <- list()
+    tem <- ncdf4::nc_open(dat.train_file)
+    dim <- tem$dim
+    for (j in seq_along(vars.info$CF.name)) {
+        if (exists(as.character(vars.info$CF.name[j]), tem$var) == FALSE) {
```
jsimkins2: Would that still fill the variable with NAs?

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
jsimkins2: These are used in the linear regression model. We need to keep our time of days consistent. I'll document this better

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
jsimkins2: Artifact from development, I'll delete

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
jsimkins2: Another artifact from development, I'll delete

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
jsimkins2: This just allows us to select the previous value and begin the data.frame. Even if the training data is 6hr reanalysis, this should be fine since it will be grabbing the value from the previous row, and the times will match up in the date and hr steps

dlebauer:
When this is merged, will the existing temporal downscaling functions in [modules/data.atmosphere/R/temporal.downscaling.R](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/temporal.downscaling.R) be obsolete?

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
+    # The model we're going to use
+    model.train <- function(dat.subset, n.beta, resids = resids) {
+        dat.subset$year <- as.ordered(dat.subset$year)
+
+        mod.doy <- lm(air_temperature ~ as.ordered(hour) * air_temperature_max.day * 
+            (lag.air_temperature + lag.air_temperature_min + air_temperature_min.day) + 
+            as.ordered(hour) * air_temperature_min.day * next.air_temperature_max - 
+            1 - as.ordered(hour) - lag.air_temperature - lag.air_temperature_min - 
+            next.air_temperature_max - air_temperature_max.day - air_temperature_min.day, 
+            data = dat.subset)  #
```
ashiklom: If this is large and computationally expensive, I would consider trying a faster `lm` implementation, such as `Rcpp::fastLm`.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
+            dat.list[[paste(i)]] <- dat.train[dat.train$doy >= i - day.window/2 & 
+                dat.train$doy <= i + day.window/2, ]
+        }
+    }
+    
+    # Do the computation and save a list Final list will have 2 layers per
+    # DOY: the model, and a bunch of simulated betas
+    if (parallel == T) {
+        warning("Running model calculation in parallel.  This WILL crash if you do not have access to a LOT of memory!")
+        library(parallel)
```
ashiklom: Need to get rid of the `library` call.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
```
ashiklom: Get rid of `library` call.

jsimkins2: got it, forgot to delete these after using `parallel::mclapply`

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
```
ashiklom: Remove `library` call.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
```
ashiklom: Remove `library` call.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
```
ashiklom: Remove `library` call.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
+        } else {
+            dat.list[[paste(i)]] <- dat.train[dat.train$doy >= i - day.window/2 & 
+                dat.train$doy <= i + day.window/2, ]
+        }
+    }
+    
+    # Do the computation and save a list Final list will have 2 layers per
+    # DOY: the model, and a bunch of simulated betas
+    if (parallel == T) {
+        library(parallel)
```
ashiklom: Remove `library` call.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1127 @@
...
+            dat.list[[paste(i)]] <- dat.train[dat.train$doy >= i - day.window/2 & 
+                dat.train$doy <= i + day.window/2, ]
+        }
+    }
+    
+    # Do the computation and save a list Final list will have 2 layers per
+    # DOY: the model, and a bunch of simulated betas
+    if (parallel == T) {
+        warning("Running model calculation in parallel.  This WILL crash if you do not have access to a LOT of memory!")
+        library(parallel)
```
ashiklom: Remove `library` call.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
jsimkins2: @ashiklom so `Rcpp::fastLm` can be substituted directly in for `lm` and is faster?

dlebauer: There is a lot we could do to optimize code in PEcAn. To justify adding another dependency it would help to see if this is actually a bottleneck and more rigorously profile the downscaling workflow. Hadley's chapter on code profiling is very useful http://adv-r.had.co.nz/Profiling.html

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
dlebauer: maybe check to see if there is enough memory and only warn if not.

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
dlebauer: according to our coding style ... use `TRUE` instead of `T`

recognizing that `parallel == TRUE`, `(parallel == TRUE) == TRUE`,  `((parallel == TRUE) == TRUE) == TRUE`, `(parallel == TRUE) == (parallel == TRUE)` etc (I was having fun, sorry it got out of hand), 

it should be sufficient to use

```r
if(parallel){
   ...
}
```

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
dlebauer:
```r
if(parallel)
```

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
ashiklom: :+1: to Dave's profiling comment. I don't think it's as simple as that because the `lm` return object has a lot of useful stuff that you rely on later. With `Rcpp::fastLm`, I think it would have to be calculated. Substituting isn't a requirement, just a thought.

modules/data.atmosphere/R/tdm_temporal_downscale.R
```diff
@@ -0,0 +1,655 @@
...
+##' @param lags.init - a data frame of initialization paramters to match the data in dat.mod
+##' @param dat.train - the training data used to fit the model; needed for night/day in 
+##'                    surface_downwelling_shortwave_flux_in_air
+# -----------------------------------
+#----------------------------------------------------------------------
+# Begin Script
+#----------------------------------------------------------------------
+
+# Set progress bar
+pb.index <- 1
```
dlebauer: Shouldn't this be inside the function?

modules/data.atmosphere/R/tdm_temporal_downscale.R
```diff
@@ -0,0 +1,655 @@
...
+    if (is.null(lags.init)) {
+        lags.init <- lags.list[[unique(dat.mod$ens.day)]]
+    }
+    
+    # Set up the ensemble members in a list so the uncertainty can be
+    # propogated
+    dat.sim <- list()
+    
+    # ----- Modeling surface_downwelling_shortwave_flux_in_air 
+    {
```
dlebauer: I don't understand this syntax. What do the braces enclosed?

jsimkins2: The braces enclosed divide each variable into it's own piece. Should I remove these?

dlebauer: I don't think there would be any effect of removing them, so they are superfluous and comments would be sufficient.

modules/data.atmosphere/R/tdm_temporal_downscale.R
```diff
@@ -55,8 +55,10 @@ predict.subdaily <- function(dat.mod, n.ens, path.model, lags.list = NULL,
     # propogated
     dat.sim <- list()
     
-    # ----- Modeling surface_downwelling_shortwave_flux_in_air 
-    {
+    # ----- Modeling surface_downwelling_shortwave_flux_in_air
+    # The following bracket encloses the shortwave_flux downscaling section. The same 
+    # format is replicated for each variable. 
+    {
```
dlebauer: better comments are great, but I am still confused why the braces are used. 

If they have no effect they should be removed.

But the bigger issue is that they delineate chunks of code that are mostly repeated, and in which there appear to be a fairly limited amount of variable-specific code that could be handled with `if ... else `. It will be much easier to go ahead and write a function now than after the sections evolve apart from each other. And much easier to maintain in the long run. 

I know this is a major change, but I am so confident that it will be worthwhile I've drafted a first pass at this here: https://gist.github.com/dlebauer/15bc2d58310bc4e6d76882f61da40caf and can help if you need.

modules/data.atmosphere/R/tdm_temporal_downscale.R
```diff
@@ -0,0 +1,656 @@
...
+            
+            dat.pred <- predict.met(newdata = dat.temp, model.predict = mod.surface_downwelling_longwave_flux_in_air.doy, 
+                Rbeta = Rbeta, resid.err = F, model.resid = NULL, Rbeta.resid = NULL, 
+                n.ens = n.ens)
+            dat.pred <- dat.pred^2  # because squared to prevent negative numbers
+            
+            # Hard-coding some sanity bounds by ball-parking things from NLDAS &
+            # CRUNCEP This is necessary if you have poorly constrained training
+            # models
+            dat.pred[dat.pred < 100] <- 100
```
dlebauer: at night wouldnt you expect radiation == 0? Or in artic in winter?

Is this the best place to have sanity checks like this? Or would it be possible to compare the input data with the downscaled data as sanity check (e.g. total radiation, precip, etc)

modules/data.atmosphere/R/tdm_temporal_downscale.R
dlebauer: PS here is an example dif of two chunks https://www.diffchecker.com/wjzfFylJ

modules/data.atmosphere/R/tdm_temporal_downscale.R
jsimkins2: This is longwave radiation, we can change this if you think it should be

jsimkins2: @dlebauer I can't seem to find the comments you made on Friday, but to be clear, would you suggest looping through each variable instead of doing each one separately? This can be done in tdm_temporal_downscale_functions because I _think_ that each variable is the same. I was keeping it separate so that one variable could be done at a time for a sanity check since it takes so long. Your thoughts?

dlebauer:
Looping should not prevent sanity checks. Have you ever used the
`debugonce` function?

Strange that you can't find my comments. Did they come to you via email?

modules/data.atmosphere/R/tdm_temporal_downscale.R
jsimkins2: @dlebauer Yeah I got them via email, but I can make it all a loop function if that's better. It shouldn't take too long. I ahven't used the `debugonce` function before, should I try and add this to the loop?

jsimkins2: After looking at the code more in depth, some variables have specific routines, so it couldn't be generalized in 1 for loop. For example, we only want shortwave to consider values when it's light out and temperature pulls in it's lag/next values differently because we have temperature max and mins

jsimkins2: @dlebauer so I could make a loop for a few variables but since temperature/precip/swdown are unique, I would have to leave them out or add multiple unique arguments within the for loop and that would get messy. Your thoughts?

dlebauer: @jsimkins2 yes, debugonce is very useful, and will make it possible to debug functions (wheater they are inside a for loop or not) see http://adv-r.had.co.nz/Exceptions-Debugging.html

I think "adding multiple unique arguments within the for loop" is preferable to copy-pasting, and is widely used in the PEcAn codebase. 

The basic construction

```r
if(met_varaible == 'temperature'){
   do this
} else if (met_variable == 'precip'){
   do that
} else if (met_variable == 'swdown'{
   do something else
} else {
    do standard stuff
}
```

Should work, but I suspect will need to be called a few times (to avoid copy-pasting) depending on the implementation. I put an example in the draft function that I wrote.

modules/data.atmosphere/R/tdm_generate_subdaily_models.R
```diff
@@ -0,0 +1,228 @@
+##' Generate Subdaily Models
+##' Create statistical models to predict subdaily meteorology
+# -----------------------------------
+# Description
+# -----------------------------------
+##'
+##' @title gen.subdaily.models
+##' @family tdm - Temporally Downscale Meteorology
+##' @author Christy Rollinson, James Simkins
+##' @description This is the 2nd script in the tdm workflow that takes the dat.train_file that is created from the
```
mdietze: This is a function, not a script

modules/data.atmosphere/R/tdm_predict_subdaily_function.R
```diff
@@ -0,0 +1,257 @@
...
+##' @author Christy Rollinson, James Simkins
+##' @description This script contains the functions that do the heavy lifting in the final
+##'              script of the tdm workflow titled tdm_predict_subdialy_met.R. It uses a linear
+##'              regression approach by generating the hourly values from the daily means of
+##'              the file the user selects to downscale based on the hourly models and betas
+##'              generated by tdm_generate_subdaily_models.R. 
+# ----------------------------------- 
+# Parameters
+# -----------------------------------
+##' @param dat.mod - data to be predicted at the time step of the training data
```
mdietze: ambiguous. What exactly is being passed in? A dataframe? A file?

modules/data.atmosphere/R/tdm_predict_subdaily_function.R
```diff
@@ -0,0 +1,257 @@
...
+##' @param n.ens - number of hourly ensemble members to generate
+##' @param path.model - path to where the training model & betas is stored
+##' @param lags.init - a data frame of initialization paramters to match the data in dat.mod
```
mdietze: parameters misspelled

modules/data.atmosphere/R/tdm_predict_subdaily_function.R
```diff
@@ -0,0 +1,257 @@
...
+    lags.init = NULL, dat.train) {
+    
+    # Set progress bar
+    pb.index <- 1
+    pb <- txtProgressBar(min = 1, max = 8, style = 3)
+    setTxtProgressBar(pb, pb.index)
+    
+    # Figure out if we need to extract the approrpiate
+    if (is.null(lags.init)) {
+        lags.init <- lags.list[[unique(dat.mod$ens.day)]]
```
mdietze: since lags.list defaults to NULL, seems like you need a test for ```!is.null(lags.list)``` before this line, and then possibly an error or warning for the else case

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1134 @@
...
+        # Note: played around with a log-transformation of
+        # surface_downwelling_shortwave_flux_in_air to prevent negative values,
+        # but that resulted in bias at upper range Solution was to just say
+        # anything <0 = 0 mod.doy <-
+        # lm(surface_downwelling_shortwave_flux_in_air ~
+        # as.factor(hour)*surface_downwelling_shortwave_flux_in_air.day,
+        # data=dat.subset[dat.subset$hour %in% hrs.day,]) ###
+        mod.doy <- lm(surface_downwelling_shortwave_flux_in_air ~ as.factor(hour) * 
+            surface_downwelling_shortwave_flux_in_air.day - 1 - surface_downwelling_shortwave_flux_in_air.day - 
+            as.factor(hour), data = dat.subset[dat.subset$hour %in% hrs.day,
```
mdietze: I didn't comment on this with the first model (air temp), since I assumed that was the starting point, of the model building, but don't later variable need to also depend on earlier variables if we want to capture the covariance among variables? (i.e. why isn't the instantaneous Tair a covariate here?)

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -0,0 +1,1134 @@
...
+            
+            # Save the model as a .Rdata
+            mod.save <- mod.out$mode
+            save(mod.save, file = file.path(path.out, paste0("model_surface_downwelling_shortwave_flux_in_air_", 
+                i, ".Rdata")))
+        }
+    }
+    
+    return(mod.out)
+}
```
mdietze: This function looks A LOT like the previous function -- seems like there should be a way to either make one master function (which is passed a variable specific list of covariates and linear model syntax) and has a couple special case if's, rather than having a lot of different functions. At a minimum, it looks like there are large chunks that could be pulled out to other functions so you can call then every time.

modules/data.atmosphere/R/tdm_nc2dat.train.R
```diff
@@ -0,0 +1,209 @@
...
+        stepby[1] <- 1
+    }
+    if ((nrow(raw_train_data) == 1460) | (nrow(raw_train_data) == 1464)) {
+      stepby[1] <- (1/6)
+    }
+    # Add a time stamp
+    start_time <- as.POSIXlt(paste0(yr_seq[1], "-01-01"), tz = "UTC")
+    end_time <- as.POSIXlt(paste0(yr_seq[1] + 1, "-01-01"), tz = "UTC")
+    train_df$date <- seq.POSIXt(from = start_time, by = 60 * 60/stepby[[1]], 
+        length.out = nrow(train_df))
```
jsimkins2: @mdietze so I'm creating the `date` variable in the format that `lubridate` needs for the `aggregate` step. I also thought it might be useful to have a general date variable since not all of our netcdf's have the "days since xxxx" yet, but maybe I should assume they all do and it will force us to get it done. In the mean time, do you know of any documentation on `udunits` where I can transform `days since` to a `seq.POSIXt` string?

jsimkins2: @mdietze nevermind, figured it out!

modules/data.atmosphere/R/tdm_predict_subdaily_function.R
```diff
@@ -0,0 +1,256 @@
...
+                  dat.pred[dat.pred > specific_humidity.fix] <- specific_humidity.fix
+                }
+            }
+            # ---------- End Quality Control
+            
+            # ---------- Begin propogating values and saving values
+            # Shortwave Radiaiton
+            if (v == "surface_downwelling_shortwave_flux_in_air") {
+                # Randomly pick which values to save & propogate
+                cols.prop <- sample(1:n.ens, ncol(dat.sim[[v]]), replace = TRUE)
```
jsimkins2: @mdietze this is a really good point, I need to think about this for a bit. Since we only do 1 variable at a time, I may need generate a list of which values to propagate beforehand and read that from the list so it's consistent across variables

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
jsimkins2: We tried to make air temp a predictory, but when using it as a predictor it caused problems in shortwave having unrealistic values. We ended up with a lot of values falling outside the tails of the range. If we did something more sophisticated with the covariance structure, we could probably improve this, but we think the covariance was also poorly constrained at the tails. That being said, my analysis of raw/downscaled data show little change in the covariances across variables (~7% change). I'm keeping track of all the future changes we want to make. Should I add this to the issue I'm creating?

mdietze: Yes, any requested changes that are beyond this PR should go into an issue so they aren't lost/forgotten when this PR is pulled

modules/data.atmosphere/R/tdm_temporal_downscale_functions.R
```diff
@@ -32,397 +30,693 @@
...
+        }
+        pb.index <- pb.index + 1
+        setTxtProgressBar(pb, pb.index)
+    }  # ---------- End of downscaling for loop
+    return(dat.sim)
+}
 
 
 # other functions that are used in the workflow
 predict.met <- function(newdata, model.predict, Rbeta, resid.err = F, model.resid = NULL,
```
mdietze: These functions could benifit from being in their own files as well

Issue #1347, opened 2017-04-13
Missing reference times in CF output


infotroph:
Several functions in `data.atmosphere` produce invalid CF output because they give time in units of "sec", with no internal indication what time the file started. 

## Description

```
cd ~/github_forks/pecan/modules/data.atmosphere/R
grep -R '"sec"' .
./debias.met.R:  time <- ncdf4::ncdim_def(name = "time", units = "sec", vals = (1:reso_len) *
./download.CRUNCEP_Global.R:    time <- ncdf4::ncdim_def(name = "time", units = "sec", vals = (1:ntime) * 21600,
./download.GFDL.R:    time <- ncdf4::ncdim_def(name = "time", units = "sec", vals = (1:2920) * 10800,
./download.GLDAS.R:                      units = "sec",
./download.MACA.R:    time <- ncdf4::ncdim_def(name='time', units="sec", vals=(1:365)*86400, create_dimvar=TRUE, unlim=TRUE)
./download.MsTMIP_NARR.R:                      units = "sec",
./download.NLDAS.R:    time <- ncdf4::ncdim_def(name = "time", units = "sec",
./met_temporal_downscale.Gaussian_ensemble.R:    time <- ncdf4::ncdim_def(name = "time", units = "sec", vals = seq_len(reso_len) *
./upscale_met.R:  time <- ncdf4::ncdim_def(name = "time", units = "sec", vals = (1:reso_len) *
```

## Context
The CF format conventions [require](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#time-coordinate) that time variables always include a reference time in their units attribute, and the PEcAn documentation at least implies that times should be [days since 1700-01-01 00:00:00 UTC](https://pecan.gitbooks.io/pecan-documentation/content/developers_guide/Adding-an-Input-Converter.html). Enforcing this at the CF conversion stage would simplify the logic of all the met2model.MODEL functions, and would be more reliable than assuming filenames contain start dates.

## Possible Implementation

infotroph:
An issue to be aware of while fixing this: If adding or subtracting large numbers of seconds to calculate reference time with something like `lubridate::ymd("1700-01-01") + lubridate::seconds(x$dim$time$val)`, look out for [integer overflows in the date math](https://github.com/hadley/lubridate/issues/534).

ankurdesai:
@jsimkins2 I think this is mostly in your functions. Can you take a look?

dlebauer:
[...]

PR #1357, opened 2017-04-18
Provide reference dates in download.CRUNCEP output

infotroph:
## Description
Ensures output times have CF-compliant units (#1347).

Also some code cleanup while I'm at it: 
* Add missing parameter descriptions
* Respect `overwrite` parameter (previously ignored)

## For discussion

The test I give here is slow! It downloads a full year of CRUNCEP data to do one check (so far) on the output. Is this a problem? If so, should the download be mocked out, or the test skipped in some environments?

## Types of changes

## Checklist:


modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -60,12 +65,27 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
...
+    }
+
     PEcAn.utils::logger.info(paste("Downloading",loc.file))
     ## Create dimensions
     lat <- ncdf4::ncdim_def(name = "latitude", units = "degree_north", vals = lat.in, create_dimvar = TRUE)
     lon <- ncdf4::ncdim_def(name = "longitude", units = "degree_east", vals = lon.in, create_dimvar = TRUE)
-    time <- ncdf4::ncdim_def(name = "time", units = "sec", vals = (1:ntime) * 21600, 
-                      create_dimvar = TRUE, unlim = TRUE)
+
+    timestamps <- (1:ntime) * 6/24 - 3/24 # data are 6-hourly, with timestamp at center of interval
```
dlebauer: So are you keeping the timestamp? Shifting time? What are time stamps of input and output data?

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -79,6 +99,14 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
       # This throws an error if file not found
       dap <- ncdf4::nc_open(dap_file)
 
+      # confirm that timestamps match
+      stopifnot(dap$dim$time$len == ntime)
+      dap_time <- udunits2::ud.convert(dap$dim$time$vals,
+                                       dap$dim$time$units,
+                                       paste0("days since ", start_year, "-01-01"))
```
dlebauer: Awesome that is.units can parse this

modules/data.atmosphere/R/download.CRUNCEP_Global.R
infotroph: We're keeping the timestamp. Or more precisely, we're generating a completely new time series here, then on reading each variable in the loop below we check that input times have the same values *after converting to our output units*.

Input timestamps of CRUNCEP nc files appear to always have units "days since 1700-01-01T00:00:00Z", so a file starting 2004-01-01 looks like `111033.1 111033.4 111033.6 111033.9 111034.1 ...`. Output timestamps are days since `start_date`, so the times always look like `0.125 0.375 0.625 0.875 1.125 ...`.

modules/data.atmosphere/R/download.CRUNCEP_Global.R
infotroph: Right? I should avoid ambiguity and add "00:00:00Z" to the output units, though.

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -79,6 +99,14 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
       # This throws an error if file not found
       dap <- ncdf4::nc_open(dap_file)
 
+      # confirm that timestamps match
+      stopifnot(dap$dim$time$len == ntime)
```
mdietze: Would prefer if you use PEcAn's logger functions so that we can ensure that this stop is properly caught and written out to the log file.

mdietze:
Note that setting up a test that depends on downloading from CRUNCEP will occasionally cause the build to fail when the code is fine if the connection to CRUNCEP is lost.

PR #1352, opened 2017-04-14
Cleaning up and updating "Using Git" page

tonygardella:
Updating Git workflow. Also cleaning the page up. 
   - Deleted "Quick and Easy" instructions because changes should be thoroughly documented and be on unique branches. 
   - Established only one workflow to make changes rather than have a "basic" and "advanced" workflow.
 
I also moved the order of some topics around.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [x] I have updated the CHANGELOG.md.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1360, opened 2017-04-18
Revert "Cleaning up and updating "Using Git" page"

mdietze:
Reverts PecanProject/pecan#1352


mdietze:
Approved the PR then realized that @tonygardella had not switched this to point to devel instead of master. Undoing and then Tony will need to do a new PR.

tonygardella:
@mdietze I pointed to master because documentation is built off of master right now and I wanted the latest github workflow to be in documentation sooner rather than later.

Issue #1351, opened 2017-04-14
How do we handle documenation for different PEcAn versions?


tonygardella:
Every time a new version of PEcAn comes out, documentation will be in sync with those changes. 

However, if someone has an older version or different version of PEcAn they will not have the correct documentation available to them

## Possible Implementation
Some possible solutions:

- Render all versions of documentation by having  deploy.sh point to tagged points in git history and give each version unique urls.

- Have the code available to render older versions of documentation if someone would like to.

Please discuss possible solutions .

mdietze:
So my ideal solution would be the first option -- we have unique urls generated for master, devel, and any release-*.  Hopefully there's a regexp solution to the latter that won't require us to update deploy.sh every time we create a new release.

robkooper:
We can use something like the following:
```
# find version if we are develop/latest/release and if should be pushed
BRANCH="$(git rev-parse --abbrev-ref HEAD)"
PUBLISH="NOPE"
if [ "$BRANCH" = "master" ]; then
  PUBLISH="yes"
  VERSION="latest"
elif [ "$BRANCH" = "develop" ]; then
  PUBLISH="yes"
  VERSION="develop"
elif [ "$( echo $BRANCH | sed -e 's#^release/.*$#release#')" = "release" ]; then
  PUBLISH="yes"
  VERSION="$( echo $BRANCH | sed -e 's#^release/\(.*\)$#\1#' )"
else
  PUBLISH="no"
  VERSION="local"
fi
```

Now we can use PUBLISH == "yes" to see if we need to publish and VERSION will contain where to publish it.

mdietze:
Currently we don't seem to be generating release-specific documentation, just master and develop

http://pecanproject.github.io/documentation.html

which wasn't our plan and doesn't allow people to see the docs for old versions. Or, if we are generating release documentation, we don't seem to be putting it on the webpage.

Post-AGU this needs to be part of a general recommitment to updating the documentation.

bpbond:
Might be useful to take a look at how a big project in our institute handles this.

https://github.com/JGCRI/gcam-doc
http://jgcri.github.io/gcam-doc/

Tagging my colleague @rplzzz to weigh in...

rplzzz:
[...]

PR #1363, opened 2017-04-19
Create seperate documentation for different branches

tonygardella:
In response to #1351 I'm trying to have deploy.sh handle different versions of documentation. Still need to figure out how to have different urls for different versions.

## Types of changes

## Checklist:

### DIFF
--- a/book_source/deploy.sh
+++ b/book_source/deploy.sh
@@ -37,21 +37,20 @@ if [ $PUBLISH == yes ]; then
 
   git clone https://${GITHUB_PAT}@github.com/${USER}/pecan-documentation.git book_hosted
 
-
   cp -r _book/* book_hosted
 
   cd book_hosted
   
-  ## CHECK if branch exists in pecan-documentation
-  if [ `git branch --list $BRANCH `]; then
-    git checkout $BRANCH
+  ## Check if branch named directory exists 
+  if [ -d $BRANCH]; then
+    mkdir $BRANCH
   else
-    git checkout -b $BRANCH
+    cd $BRANCH
   fi
   
   git add --all *
   git commit -m "Update the book `date`" || true
-  git push -q origin $BRANCH
+  git push -q origin master
 
 else [$PUBLISH == no]; then
   echo "Not Master, Develop, or Release Branch. Will not render Book."
--- a/book_source/deploy.sh
+++ b/book_source/deploy.sh
@@ -6,28 +6,48 @@ set -e
 #check for environment variable
 [ -z "${GITHUB_PAT}" ] && exit 0
 
-# only deploy if this is the master branch build
-branch_name=$(git symbolic-ref -q HEAD)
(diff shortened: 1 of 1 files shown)


ashiklom:
AFAIK, you won't be able to have different URL's based on different branches because GitHub only knows how to publish content from one branch.

My suggestion would be to render all the documentation in the `pecan-documentation:master` branch, but use different directories for the `master` and `develop` versions. For instance, you could have the directory structure of `pecan-documentation` look like:

```
/master/
--- figures/
--- libs/
--- tutorials/
--- index.html
--- <other HTML files>
/develop/
--- figures/
--- libs/
--- tutorials/
--- index.html
--- <other HTMLs>
```

This would mean our URLs would look like `pecanproject.github.io/pecan-documentation/master/` for the `master` branch, and `pecanproject.github.io/pecan-documentation/develop/` for the `develop` branch.

Alternatively, we could push to two different repositories, called `pecan-documentation` and, for example, `pecan-documentation-develop`. The corresponding URLs would then be `pecanproject.github.io/pecan-documentation` and `pecanproject.github.io/pecan-documentation-develop`.

tonygardella:
I like the first option of having different directories.

book_source/deploy.sh
```diff
@@ -6,28 +6,53 @@ set -e
...
+
+  cd book_hosted
+  
+  ## Check if branch named directory exists 
+  if [ -d $BRANCH]; then
+    mkdir $BRANCH
+  else
+    cd $BRANCH
+  fi
+
```
robkooper: Need to do the copy after the mkdir, so it ends up in the right folder.

book_source/deploy.sh
```diff
@@ -6,28 +6,55 @@ set -e
...
+    cd $BRANCH
+  fi
+  
+  cp -r _book/* .
+  
+  git add --all *
+  git commit -m "Update the book `date`" || true
+  git push -q origin master
+
+else [$PUBLISH == no]; then
```
ashiklom: Should probably be either just `else ...` or `elif [ ... ]`. Also, note that `bash` is quirky and defines `[` as an actual command and `]` as its final argument, so you'll need to put a space after the opening `[` and before the closing `]`.

book_source/deploy.sh
```diff
@@ -6,28 +6,62 @@ set -e
...
+  
+  ## Check if branch named directory exists 
+  if [ -d $BRANCH ]; then
+    cd $BRANCH
+  else
+    mkdir $BRANCH
+    cd $BRANCH
+  fi
+  
+  cp -r _book/* book_hosted/$BRANCH
```
robkooper: You already cd'd into the book_hosted/branch. I think this should be `cp -r ../../_book/* .`. I would suggest using: `rsync -a --delete ../../_book/ .` since that will remove any files that are no longer available in the newly created folder.

Issue #795, opened 2016-04-13
Generalize sda.enkf.R to work with other models and state variables


araiho:
Right now, the code is working for LINKAGES with four species biomass at UNDERC. To generalize, we need to do the following steps:

- Make it work for LINKAGES at other sites with more species and other state variables.

- Fix the SIPNET code to make sure that the code is generalized enough to work with two models. This will involve outlining the interface design for write.restart.MODEL and read.restart.MODEL functions.

Issue #645, opened 2015-10-02
Generalizing Kalman Filter (sda.enkf)


araiho:
These are notes from a conversation with Mike and Betsy about how to generalize the kalman filter module. More specific tasks to come. 

- sipnet specific clim split issue

- Loading Output
  right now -> read.output not sufficient
  generalize -> read.restart.[model] (path, when, what -> variable list)
    currently variables are hard coded
  reads into "forecast" nEns x nVar matrix

- Analysis
  sipnet specific snow hack
  generalize H to multiple data (observation operator generalization)
  could get EAKF working as option
  "sanity check" generalized -> Bety specify support??

- Forecast 
  generalization issue
  write.restart.[model]

- Load.data -> obs
  handling error est; QAQC
  STEPPS -> cov? MCMC samples?
  examples of "data product"
  generic structure to obs??

PR #1366, opened 2017-04-20
Changing JAGS to NIMBLE in sda.enkf and some LINKAGES specific changes in other scripts

araiho:
NIMBLE allows us to implement the tobit correctly, to run SDA faster, and to have more flexibility with observation operators in the future. We worked with the nimble team to get a toggle sampler function so that we only need to recompile the model when we need to change the sampler. 

## Description
I began working on putting fcomp in the SDA workflow in December and we quickly realized we needed to move from JAGS to NIMBLE to implement the tobit correctly. While changing from JAGS to NIMBLE I also found some places where there needs to be more generalization. Those are marked with 'TO DO.' Let me know if you want me to complete any of them before you will accept my pull request. I also added some unit changes and changes to LINKAGES stuff, so that you can do sensitivity analysis and PDA with more parameters. 

## Motivation and Context

Issue #795 may now be resolved.
This is a continuation of issue #645.

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.
- [ ] I have added tests to cover my changes. --> I could use some input on how to do this. 


tonygardella:
@mdietze Do you have time to review this?

PR #1369, opened 2017-04-20
Paleon regional met + MultiSettings bugfix

mdietze:
@rykelly could you check that you didn't have this bug in place when you did the MANDIFORE runs?


mdietze:
@rykelly I was actually hoping that bug was why we had such low correlations in MANDIFORE. But at least good to know you don't have to redo the runs.

rykelly:
That crossed my mind too. Sorry, no such luck!

PR #1370, opened 2017-04-21
Met process debugging to support PalEON regional met

mdietze:
Lots of small bugfixes and an attempt to generalize the extract step a bit.

Also added sitegroup to multisettings expansion in workflow.R

While some of the fixes are small, there are important fixes in met.process.stage and check.inputs that were causing the wrong met stage to be detected, or met.process to be skipped altogether (respectively). Importantly, now you should be able to specify a met input ID in a pecan.xml that ISN'T the final model-specific format and have check.input NOT fill in the path (the presence of the path was what caused met.process to be skipped) and have, if the met's already CF, have met.process.stage detect that both download and met2CF can be skipped. In getting skipping working, discovered that various input ID variables were not getting defined.

PalEON sitegroup runs APPEARS to now be running (though extract takes 3hr per grid cell, so will be a while till I know for sure) which should open us up to more general pseudoregional runs. My next step is thus to think about how to get met.process to run in parallel.


web/workflow.R
```diff
@@ -46,6 +46,15 @@ if("benchmarking" %in% names(settings)){
   settings <- papply(settings, read_settings_RR)
 }
 
+if("sitegroup" %in% names(settings)){
```
rykelly: 👍

mdietze:
Just realized that these changes didn't get merged before I pushed new things I was working on in this branch. Since Ryan's review, I've added to my PalEON regional scripts (not functions) examples of how I ran extract.nc remotely on geo and how I pulled it back. I've discussed with @istfer that the latter could be generalized to a more general function, since it was checking for file completeness and then creating the input record and the local and remote dbfile records.

I've also taken a first crack at providing a spin-up met function that symlinks met records, and also added the passing of spin-up setting in through met process. The spin.met function is designed to be called within the model-specific met2model.[MODEL], not as a step in met.process itself, because how different models handle spin-up is so different (e.g. currently testing with DALEC, which builds all the met into one giant file).  For some models, there will need to be additional model-specific parsing in write.configs.[MODEL]

modules/data.atmosphere/R/closest_xy.R
```diff
@@ -37,13 +40,35 @@ closest_xy <- function(slat, slon, infolder, infile) {
         }
       }
     }
+  } else {
+    ## this case may make assumptions specific to PalEON Regional
```
istfer: not sure what the Paleon lat/lon will look like here but if they are vectors, i.e. `dim(lat) = NULL`, wouldn't they meet `all(dim(lat) == dim(lon))` condition above? `all(NULL == NULL)` returns `TRUE` I think

modules/data.land/inst/LoadPalEONsites.R
```diff
@@ -90,4 +88,178 @@ for(i in seq_along(paleon.sitegroups)){
...
+                           lat = db.site.lat.lon(site_id, con = con)$lat, 
+                           lon = db.site.lat.lon(site_id, con = con)$lon,
+                           str_ns = str_ns,
+                           outfile = outfile))
+  }    
+  save(site.info,file=paste0("PalEON_siteInfo_",paleon.sitegroups[i],".RData"))
+}
+
+
+#####################################
```
istfer: from this part on is what you want to be generalized, right?

PR #1371, opened 2017-04-21
1.4.10.1 released

robkooper:
This will merge all bug fixes made to 1.4.10.1 during the release process back into the develop branch.


tonygardella:
Should I do `git revert  45bf6f0` or just open a new PR?

ashiklom:
Probably easiest to just merge this and open a new PR. @tonygardella You can grab your documentation changes by doing `git checkout <commit hash> -- <your changes>` (where "commit hash" is for a commit containing the correct version of the documentation) and then committing them.

tonygardella:
@ashiklom So these changes should have be in  master, develop (after this PR is accepted), and the release. Would it be best to do a  PR to all separately ?

ashiklom:
No, a separate PR to everything would be very confusing and would be extremely likely to introduce merge conflicts. The best solution is probably to open a PR to the branch that is the farthest behind in the history (which I think is the release candidate), then merge that branch into  `master`, and then merge the `master` branch into `develop`.

tonygardella:
@mdietze @robkooper If I click update branch, that will update master with #1369 from develop? It won't break anything but if we want to stick to keeping master branch the same as the latest release we need to make sure to approve and merge these PRs quickly in the future.

tonygardella:
@robkooper According to the [article](http://nvie.com/posts/a-successful-git-branching-model/) I think we should have done a PR from release->master and then merged release->develop (since there were changes on release) and not do a PR from master->develop.

Issue #506, opened 2015-06-24
Soil data standard


mdietze:
In order to develop a workflow that provides soil initial conditions to the models (analogous to the met workflow in operation, and the veg workflow in development), we need to first agree on a standard set of variable names and units, as well as a standard file format. The idea is that any soil database would then be converted to this standard, PEcAn would do any processing/extraction based on that standard, and then each model would convert those standard variables into what is needed for that model.

As a first pass, I think that we'll need standard names for soil texture variables (texture class, % sand, silt, clay, organic, rock), soil moisture and temperature initial conditions, and soil biogeochemical pools.

In terms of a workflow, I think we'll want to start with texture, since it's easiest and depends least on time. That said, there's an interesting interaction here between the soil texture variables and the soil physical parameters in most models, as well as a clear need to propagate uncertainties in these.

The soil moisture and temperature variables do depend on time but can be extracted from a number of reanalysis products (e.g. NARR) and field observations (e.g. Ameriflux), and thus we should discuss whether to incorporate that into the met workflow somehow (e.g. include the processing of these variables in the standard netcdf CF files).

Finally, soil biogeochemical pools will present a particular problem because the data is crappy at a regional scale and most models do not have any direct mapping between what can be measured and their internal pools. Furthermore, a lot of data will just be %C and %N (bulk), which requires that we integrate over the (usually large) uncertainty in bulk density. Rare, but important, will be data that partitions soil biochemical pools into different physical or chemical fractions. In addition, most field measurements are over specific depth ranges rather than the full soil profile, which means that the standard processing will need to deal with this extrapolation problem as well as the wide variety of soil depths used by different models. As a starting point just having a standard for %C, %N, bulk density, and their uncertainties may be good enough to get us rolling.

Finally, soil databases will present the problem that many data layers will be in vector GIS rather than nice raster layers. For sanity sake we may want to start with the various 'harmonized' raster databases, but they are often very inaccurate when applied at the site scale. Also, many harmonized databases simply drop the presence of less frequent soil types (and only report the single most common soil type) and thus they don't represent the actual variability/uncertainty that we'd want to integrate over.

mdietze:
[...]

PR #1380, opened 2017-05-01
Paleon regional: soils

mdietze:
Utilities for estimating a range of soil hydraulic and thermal properties from either texture class or sand/silt/clay.  Would eventually like to add the capacity to handle parameter uncertainties (both in the sand/silt/clay input and the relationship between texture and properties).  Variable names all added to BETY and conform to CF or CF-equivalent where not model-specific #506 

Coming next is the extraction of soil texture from PalEON / MSTMIP soil layer to a netCDF CF file that can either be read directly during write.configs (many models include soil parameters with other parameters) or converted to a model-specific format (I/O of soil2model.[model] TBD when needed)


tonygardella:
@araiho Are you unavailable this week? I added you on this as a reviewer because you are matched up with Mike this week for code review, but if you can't get to it let me know.

tonygardella:
@ankurdesai, Ann is away right now and you're next for code review for Mike so I'm adding you as a reviewer on this.

modules/data.land/R/soil_utils.R
```diff
@@ -0,0 +1,266 @@
...
+  }else if(clay > 27.5 & clay <= 40.0 & sand > 20.0 & sand <= 45.0) {
+    mysoil <-  8 #----- Clayey loam. -----------------------------------------------------#
+  }else if(clay > 35.0 & sand > 45.0) {
+    mysoil <-  9 #----- Sandy clay. ------------------------------------------------------#
+  }else if(clay > 40.0 & silt > 40.0) {
+    mysoil <- 10 #----- Silty clay. ------------------------------------------------------#
+  }else if(clay <= 70.0 & sand <= 30.0 & silt <= 30.0) {
+    mysoil <- 11 #----- Clay. ------------------------------------------------------------#
+  }else if( silt > 80.0 & clay <= 12.5) {
+    mysoil <- 14 #----- Silt. ------------------------------------------------------------#
```
ankurdesai: This functions looks fine for the most part, but one part perplexing me was how it IDed organic soils (peaty) from texture alone and I see that it doesn't. Something to consider (i.e., how to pass info to say hey this soil is mostly organic matter, assign class 13).

modules/data.land/R/soil_utils.R
```diff
@@ -0,0 +1,266 @@
...
+  }else if( clay > 70.0) {
+    mysoil <- 15 #----- Heavy clay. ------------------------------------------------------#
+  }else if( clay > 40.0 & sand > 30.0 & sand <= 45.0) {
+    mysoil <- 16 #----- Clayey sand. -----------------------------------------------------#
+  }else if( clay > 40.0 & silt > 30.0 & silt <= 40.0) {
+    mysoil <- 17 #----- Clayey silt. -----------------------------------------------------#
+  }else{
+    print("---------------------------------------------------")
```
ankurdesai: Should this print statements and stop statements be logger functions?

modules/data.atmosphere/R/met.process.R
```diff
@@ -27,6 +27,23 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
 
+  # get met source and potentially determine where to start in the process
+  if(is.null(input_met$source)){
+    if(is.null(input_met$id)){
+      PEcAn.utils::logger.warn("met.process only has a path provided, assuming path is model driver and skipping processing")
+      return(input_met$path)
+    }else {
+      logger.error("Must specify met source")
+      if(!is.null(input_met$id) & !is.null(input_met$path)){
+        met <- input_met$source <- "CRUNCEP" ## this case is normally hit when the use provides an existing file that has already been
```
ankurdesai: Wonder if we should create a generic "CFmet" source?

modules/data.atmosphere/R/spin.met.R
```diff
@@ -51,25 +52,43 @@ spin.met <- function(in.path,in.prefix,start_date,end_date,nyear,nsample,resampl
...
   
   ### define the met years to sample
+  new_year <- seq(lubridate::year(spin_start_date),by=1,length.out=nyear)
+  is.leap <- lubridate::leap_year(avail.years)
+  spin_year <- NA
   if(resample){
-    spin_year <- sample(avail.years[1:nsample],
-                        size = nyear,replace = TRUE)
+    for(t in seq_along(new_year)){
+      if(lubridate::leap_year(new_year[t])){
```
ankurdesai: I guess it makes sense to preserve distribution of leap years (esp for models that may check), but tradeoff is fewer years to sample over.

modules/data.land/R/soil_utils.R
```diff
@@ -0,0 +1,315 @@
...
+    }else if( silt[z] > 80.0 & clay[z] <= 12.5) {
+      mysoil[z] <- 14 #----- Silt. ------------------------------------------------------------#
+    }else if( clay[z] > 70.0) {
+      mysoil[z] <- 15 #----- Heavy clay. ------------------------------------------------------#
+    }else if( clay[z] > 40.0 & sand[z] > 30.0 & sand[z] <= 45.0) {
+      mysoil[z] <- 16 #----- Clayey sand. -----------------------------------------------------#
+    }else if( clay[z] > 40.0 & silt[z] > 30.0 & silt[z] <= 40.0) {
+      mysoil[z] <- 17 #----- Clayey silt. -----------------------------------------------------#
+    }else{
+      print("---------------------------------------------------")
```
ankurdesai: switch to logger statements?

modules/data.land/R/soil_utils.R
```diff
@@ -0,0 +1,315 @@
...
+      mysoil[z] <-  7 #----- Silty clay loam. -------------------------------------------------#
+    }else if(clay[z] > 27.5 & clay[z] <= 40.0 & sand[z] > 20.0 & sand[z] <= 45.0) {
+      mysoil[z] <-  8 #----- Clayey loam. -----------------------------------------------------#
+    }else if(clay[z] > 35.0 & sand[z] > 45.0) {
+      mysoil[z] <-  9 #----- Sandy clay. ------------------------------------------------------#
+    }else if(clay[z] > 40.0 & silt[z] > 40.0) {
+      mysoil[z] <- 10 #----- Silty clay. ------------------------------------------------------#
+    }else if(clay[z] <= 70.0 & sand[z] <= 30.0 & silt[z] <= 30.0) {
+      mysoil[z] <- 11 #----- Clay. ------------------------------------------------------------#
+    }else if( silt[z] > 80.0 & clay[z] <= 12.5) {
```
ankurdesai: Not sure if my last comment got added, but basically this function uses texture to get info, but doesn't have a way to use OC information to figure out peaty soils. Something to consider for future change, working in wetlands.

modules/data.land/R/soil_utils.R
infotroph: @ankurdesai Unfortunately  my understanding is that all these texture <> class relationships really only hold for mineral soil, so there's no sure way to separate peat from silt if you only have particle size (sand/silt/clay) data. Though of course If information about organic matter content *is* available, we could provide a separate function that acts upstream of this one?

models/jules/inst/template.job
```diff
@@ -20,7 +20,7 @@ if [ ! -e "@OUTDIR@/*.nc" ]; then
   # check the status
   if [ $STATUS -ne 0 ]; then
     echo -e "ERROR IN MODEL RUN\nLogfile is located at '@OUTDIR@/logfile.txt'" >&3
-    exit $STATUS
+#    exit $STATUS
```
robkooper: if there is an error the model should exit correct? So i think this should not be commented out.

modules/data.land/R/soil_utils.R
mdietze: So I believe the USDA soil classes are definitionally based on the sand/silt/clay fractions and ignore both organic and stones. More generally I agree that it would be nice to have equations that include organic and stone fractions in estimating soil physical properties, but to do so someone would need to find newer equations that include those terms. Citations welcome!

modules/data.atmosphere/R/spin.met.R
mdietze: I didn't do this by choice (or with foresight), but in response to a bug -- models that need leap years throw an error if you pass them met that's one day short

models/jules/inst/template.job
mdietze: This is, unfortunately, a hack. There's something wrong with how we're configuring dates in JULES that's causing the model to try and read meteorology for 1 year beyond the end_date. It can't find it so it throws an error, but it does this at the end of the run. If I let job.sh exit, then the model output (which is complete) never gets converted from model-specific format into PEcAn output standard. I've tried a bunch of things to try and solve the bug, and talked to Tris about it, but no luck so far -- still on my to-do list but I needed to get the runs done.  Once I sort that out I'll revert to throwing the error correctly.

modules/data.atmosphere/R/met.process.R
```diff
@@ -238,15 +250,26 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
     model.id  <- met2model.result$model.id
-    outfolder <- met2model.result$outfolder
+    model.file.info <- db.query(paste0("SELECT * from dbfiles where id = ", model.id$dbfile.id), con)
+    model.file <- file.path(model.file.info$file_path,model.file.info$file_name)
+    
   } else {
-    model.id  <- ready.id
-    outfolder <- file.path(dir, paste0(met, "_site_", str_ns))
+    PEcAn.utils::logger.info("ready.id",ready.id,machine.host)
+    model.id  <- dbfile.check("Input", ready.id, con)#, hostname=machine.host)
```
mdietze: I left that in as a reminder that I still need to get this working for the remote case

modules/data.land/R/extract_soil_nc.R
```diff
@@ -0,0 +1,195 @@
...
+  
+  ## open soils
+  nc <- ncdf4::nc_open(in.file)
+  
+  ## extract lat/lon
+  dims <- names(nc$dim)
+  lat.dim <- dims[grep("^lat",dims)]
+  lon.dim <- dims[grep("^lon",dims)]
+  soil.lat <- ncdf4::ncvar_get(nc, lat.dim)
+  soil.lon <- ncdf4::ncvar_get(nc, lon.dim)
```
mdietze: Exactly, the names can differ (some files are just lat/lon rather than writing out latitude and longitude), hence the regexp above which ensures that we fix the names when we produce our file.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -0,0 +1,195 @@
...
+  
+  ## check in range
+  dlat <- abs(median(diff(soil.lat)))
+  dlon <- abs(median(diff(soil.lon)))
+  if(lat < (min(soil.lat)-dlat) | lat > (max(soil.lat)+dlat)){
+    PEcAn.utils::logger.error("site lat out of bounds",lat,range(soi.lat))
```
mdietze: good catch!

modules/data.land/R/extract_soil_nc.R
```diff
@@ -0,0 +1,195 @@
...
+                                      "thcond1","W m-1 K-1",
+                                      "thcond2","1",
+                                      "thcond3","1",
+                                      "soil_thermal_conductivity","W m-1 K-1", 
+                                      "soil_thermal_conductivity_at_saturation","W m-1 K-1", 
+                                      "soil_thermal_capacity","J kg-1 K-1",
+                                      "soil_albedo","1"
+                                      ),
+                                    ncol=2,byrow = TRUE))
+  colnames(variables) <- c('var','unit')
```
mdietze: Since this is a tiny function that could get called a lot within the soil workflow, I think I was just trying to avoid the cost of constantly reloading the lookup table from disk every time the function is called.

modules/data.land/inst/build_soil_texture_variables.R
```diff
@@ -0,0 +1,159 @@
...
+               , 0.333, 0.075, 0.100, 0.375, 0.125)
+xclay.def = c( 0.030, 0.060, 0.110, 0.160, 0.170, 0.270
+               , 0.340, 0.340, 0.420, 0.470, 0.600, 0.200
+               , 0.333, 0.050, 0.800, 0.525, 0.525)
+
+soil.name = c("Sand","Loamy sand","Sandy loam","Silt loam","Loam","Sandy clay loam"
+              ,"Silty clay loam","Clayey loam","Sandy clay","Silty clay","Clay"
+              ,"Peat","Bedrock","Silt","Heavy clay","Clayey sand","Clayey silt")
+
+texture <- read.csv("~/pecan/modules/data.land/data/texture.csv",header=TRUE,stringsAsFactors = FALSE)
```
mdietze: It's not intended to be run by users, it's used to build an RData file of important look-up tables, the latter of which IS accessed by system.file. This script only needs to be rerun if that RData file needs to be regenerated.

PR #1387, opened 2017-05-05
Biocro: look for ppfd by full CF name

infotroph:
## Motivation and Context

`met2model.biocro` already looked for PAR and PPFD inputs by a few different names, but did not check for it by the CF standard name `surface_downwelling_photosynthetic_photon_flux_in_air` and therefore broke on some inputs that should be valid.

## Types of changes

## Checklist:


infotroph:
@dlebauer Good catch -- changed to umols, fixed the doc that made me think it was mols.

Issue #1338, opened 2017-04-12
BioCro: Upscale or error on sub-hourly met data


infotroph:
If given a met file with more-than-hourly data frequency, BioCro will return multiple lines of results per hour. It's not clear whether these have the correct units to be combined usefully.

## Possible Implementation

- [ ] In run.biocro.R, check that met input is hourly and either upscale it (with data.atmosphere::upscale_met?) or throw an error
- [ ] in either run.biocro.R or model2netcdf.BIOCRO.R, check that output is hourly and throw error if not

infotroph:
Implemented in #1394, can be closed once that's merged

PR #1394, opened 2017-05-11
upscale_met rewrite, BioCro input scaling

infotroph:
## Description

* Ensure that met2model.BIOCRO converts to hourly (#1338)
* Pass correct start/end times to BioGro when met file is not a full year long
* Respect input CF time units and base dates when upscaling (#1347)
* **Breaking changes** to `upscale_met` arguments:
    - New parameter `reso_unit`  with default "hours"
    - `reso` renamed `resolution`
    - `site.id` removed; it was unused
    - `overwrite` was ignored, now works as documented.
* Look up times from file, do not assume year is in filename
* Upscaling now works on any length of file (previously assumed one year without checking)
* Tests

## Motivation and Context

The original idea here was to ensure that BioCro met files were correctly upscaled to hourly from e.g. 15-minute Geostreams input,  but the further I dug into `upscale_met` the more changes it needed to handle generic CF files correctly.

Note that I've marked the `upscale_met` edits as a breaking change because they change the function signature, but FWIW I expect little impact -- looks like `met2model.BIOCRO` is the only place `upscale_met` is currently being used.

## Types of changes

## Checklist:


models/biocro/R/run.biocro.R
```diff
@@ -52,6 +52,32 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
+
+    # BLETCHEROUS HACK: BioCro 0.94 starts the run by subsetting weather data
+    # to day1:dayn, but it assumes the data start on DOY 1 and contain
+    # (yearlength*(24/timestep)) lines. This means that in practice, day1 and
+    # dayn are treated as "day of file" not "day of year".
+    # BioCro *does* handle DOY correctly downstream of the subsetting, so here
+    # we check if the current BioCro has fixed this assumption.
+    # If not, rescale day1 and dayn to be relative to the start of the input.
+    #   Scaling is derived by inverting Biocro's day->index equations.
+    biocro_checks_doy = tryCatch(
```
dlebauer: PEcAn style mandates `<-` instead of `=` as assignment operator

modules/data.atmosphere/R/upscale_met.R
```diff
@@ -1,87 +1,85 @@
...
-##' @param site.id
-##' @param reso - numeric value in HOURS that is the desired resolution for the dataset to be upscaled to
-##'               # might need to change this for monthly/yearly option later
-##' @param overwrite
-##' @param verbose
-##' @author James Simkins
+##' @param input_met path to netcdf file containing met dataset
+##' @param outfolder path to directory where output should be saved
+##'           Output is netcdf with named as <input_met_filename>.upscaled.nc
+##' @param resolution numeric value that is the desired resolution for the dataset to be upscaled to
```
dlebauer: what is the default resolution 6 hours? 1 hr would make more sense to me.

Does it make sense to allow users to specify the value + units, or could we stick with the PEcAn-CF standard units of days?

modules/data.atmosphere/R/upscale_met.R
```diff
@@ -1,87 +1,85 @@
...
+  step <- round(nrow(met_data) / n_times, 0)
+  rows_used <- nrow(met_data) - (nrow(met_data) %% step)
+  n_steps <- (rows_used %/% step)
+  met_data <- met_data[1:rows_used,]
+  upscaled_time = colMeans(matrix(time_data[1:rows_used], nrow=step))
   upscale_data <- data.frame()
-  for (n in seq_along(var$CF.name)) {
-    upscale_data[1:reso_len,n] <- colMeans(matrix(met_data[[n]], nrow=step))
+  for (name in names(met_data)) {
+    upscale_data[1:n_steps,name] <- colMeans(matrix(met_data[[name]], nrow=step))
```
dlebauer: @jsimkins2 and @ankurdesai I just want to confirm that it is sensible to take the means rather than sums of irradiance and precipitation fluxes when upscaling. 

This makes sense to me right now since these are rates / fluxes (units of W/m2 and kg/m2/s) rather than amounts. However I recall cases where sun and rain are summed over the interval. I believe this problem is resolved by the met2CF functions but just want to double check.

modules/data.atmosphere/R/upscale_met.R
```diff
@@ -1,87 +1,85 @@
...
-  colnames(upscale_data) <- var$CF.name
-  for (x in 1:reso_len) {
-    upscale_data$air_temperature_max[x] <- max(met_data$air_temperature[(x * step - 
-                                                                           step + 1):(x * step)])
-    upscale_data$air_temperature_min[x] <- min(met_data$air_temperature[(x * step - 
-                                                                           step + 1):(x * step)])
+  if (!is.null(upscale_data$air_temperature)
+      && is.null(upscale_data$air_temperature_max)
+      && is.null(upscale_data$air_temperature_min)) {
+    for (x in 1:n_steps) {
```
dlebauer: Are there any models that use sub-daily min / max rather than mean temperature? Or can the calculation of min/ max temp be restricted to cases where upscaled resolution = 1 day?

modules/data.atmosphere/R/upscale_met.R
dlebauer: per style conventions (apologies for being pedantic) .... 

1. add spaces ...

```r
upscale_data[1:n_steps,name] <- colMeans(matrix(met_data[[name]], nrow=step)) # NO!!!!
upscale_data[1:n_steps, name] <- colMeans(matrix(met_data[[name]], nrow = step)) # YES!!!!
```

2. use `seq_along` and `seq_len` for iterating with integers, e.g. 

```r
for(i in 1:nsteps){ #NO!!!!
for(i in seq_len(nsteps)){ #YES!!!
```

... bonus points for meaningful variable names: `for(step_i in seq_len(nsteps)){`

modules/data.atmosphere/R/upscale_met.R
```diff
@@ -1,87 +1,85 @@
...
+  time <- ncdf4::ncdim_def(name = "time", units = paste(time_unit, "since", time_base),
+                           vals = udunits2::ud.convert(upscaled_time, reso_unit, time_unit),
+                           create_dimvar = TRUE, unlim = TRUE)
   dim <- list(lat, lon, time)
   
-  for (j in seq_along(var$CF.name)) {
-    upscale.list[[j]] <- ncdf4::ncvar_def(name = as.character(var$CF.name[j]), 
-                                          units = as.character(var$units[j]), 
+  upscale.list <- list()
+  for (j in names(upscale_data)) {
```
dlebauer: 3. meaningful variable names + DRY

```r
for(j in names(foo))){# NO!!!
for(foonames in names(foo)){#YES!!!
```

when used >1x (e.g. lines 81 and 94)

```r
varnames <- names(upscale_data)

for(varname in varnames){
   ....
}

other code

for(varname in varnames){
   ....
}
```

models/biocro/R/met2model.BIOCRO.R
```diff
@@ -45,12 +44,34 @@ met2model.BIOCRO <- function(in.path, in.prefix, outfolder, overwrite = FALSE,
...
+    dt <- mean(diff(udunits2::ud.convert(
+      met.nc$dim$time$vals,
+      met.nc$dim$time$units,
+      "hours since 1700-01-01 00:00:00")))
+    if (dt < 1) {
+      # More than one obs/hour. Write upscaled hourly file and reload.
+      on.exit()
+      ncdf4::nc_close(met.nc)
+
+      upscale_result <- PEcAn.data.atmosphere::upscale_met(
```
dlebauer: @ankurdesai and @jsimkins2 also want to confirm: should upscale be done within the met2model functions? Or were we going to keep the upscaled and downscaled data in CF format?

infotroph: I agree that this is an awkward place to do the scaling, especially since upscale writes a new CF file and downscale returns a dataframe, but is the desired temporal resolution available before this stage?

If keeping this, I should at least make sure to insert a Bety record for the upscaled file.

modules/data.atmosphere/R/upscale_met.R
infotroph: 6 hours was just me leaving the previous default. 

I'll defer to others on whether units should be configurable, but have been assuming the unit check is necessary anyway -- If the upstream met2CF conversions be trusted to always produce days, this code could be simplified.

models/biocro/R/run.biocro.R
infotroph: fixed by a62ab0e9

models/biocro/R/met2model.BIOCRO.R
dlebauer: My understanding of the met workflow is that all of the scaling should be done before the model-specific conversion. That said, it is not necessary to hold up this PR before refactoring.

modules/data.atmosphere/R/upscale_met.R
infotroph: fixed in a685192c

modules/data.atmosphere/R/upscale_met.R
infotroph: style points fixed in a62ab0e9 and a685192c

modules/data.atmosphere/R/upscale_met.R
dlebauer: @infotroph even if the input met isn't days, couldn't the `resolution` argument could be defined as a numeric value with units of days? Inside the function, you could convert the input_met time series to units of days. (or otherwise the rescale the value of `resolution` to match the input. 

My only concern is that we minimize the number of arguments and resulting cognitive load.

infotroph: Yep, that works. Fixed in cff44594.

modules/data.atmosphere/R/upscale_met.R
infotroph: I don't know the answer to this. Tagging @ankurdesai  and @jsimkins2 here as well.

infotroph:
I think I've fixed the changes @dlebauer requested. Will stand by for @ankurdesai and @jsimkins2 thoughts on where scaling should happen, whether to sum or mean the irradiance and precipitation, and whether it makes sense to restrict min/max temperature calculations to the whole-day case.

models/biocro/R/met2model.BIOCRO.R
ankurdesai: Agree with David. Scaling should be in CF-space. Which doesn't prevent met2model functions for using it if model can only handle a certain timestep.

dlebauer:
Will merge. We can revisit moving the upscale_met once the met workflow is more robust.

PR #1395, opened 2017-05-11
Biocro date parsing improvements

infotroph:
## Description

Mostly intended to reduce surprises when input is a datetime rather than a bare date.

The biggest behavior change is in `load.cfmet`: it was returning only dates before the *beginning* of `end.date` (i.e. with hourly input, `end.date=2016-12-31` produced a file ending at 2016-12-30 23:00:00. Most other PEcAn functions seem to interpret `end.date` as an *inclusive* cutoff (i.e. with hourly input, `end.date=2016-12-31` should produce a file ending at 2016-12-31 23:00:00)

To resolve that discrepancy and prevent suprises from met files that are a day shorter than I expected, I've changed this to treat *dates* as a ceiling but *datetimes* as a floor: read `end.date=2015-12-31` as "through the end of the 31st" but `end.date=2015-12-31 18:00:00Z` as "stopping promptly and 18:00:01 need not apply." Reviewers, is this actually less confusing than the old system?

## Types of changes

## Checklist:


tonygardella:
@serbinsh Added you as reviewer because you're matched up with Chris this week and nobody was previously assigned to this.

dlebauer:
I've reviewed and merged

PR #1404, opened 2017-05-17
Add clowder = geostreams to available met

dlebauer:
## Description

Currently the geostreams data does not show up in the pecan web interface (https://terra-mepp.illinois.edu/pecan). 

I think this is handled at https://github.com/PecanProject/pecan/blob/develop/web/03-inputs.php#L158. It seems there are two lines, first to check the site name and the second to check the data source based on the file name. The file should have "Clowder" in the name. But It was not clear how to identify a site that is available in Clowder ... see e.g. https://terra-mepp.illinois.edu/bety/dbfiles?utf8=%E2%9C%93&DataTables_Table_0_length=25&search=energy+farm&direction=&sort= Is the option either to a) list the (currently 5 sites: 3 at UIUC and 2 at Maricopa) with data in geostreams as part of the string matching or b) query the sitegroups table or c) other?

## Types of changes

## Checklist:


robkooper:
My thinking is that we can use the key/value table. Add to each site a list of possible met products, that way we don't have to parse the sitename and include (US-NR1) to the sitename.

mdietze:
@robkooper that doesn't seem scalable -- lots of met products are associated with simple (but very large) geographic regions that are easy queries through postGIS. It doesn't make sense to me to have a key/value table that we have to update for every single site with an ever-growing list of regional and global products as sites are added and as met products come online.  key/value may work for some network-based products, but in that case can't we just leverage the sitegroups table rather than creating a new table that would be largely identical (e.g. the Ameriflux and FLUXNET sitegroups would be identical to their met key/value tables, but then we'd have to keep both up-to-date and identical -- which seems pretty redundant).

Either way, it would be good to have a DB-based list of products and the site (which may be a polygon) or sitegroup they are associated with, rather than hand coding such info into the PHP. Would be good to do that in a forward-looking way that anticipates the other input needs of models.

Issue #1415, opened 2017-05-23
Names of variables returned from read.output() not the same as load_data()


mccabete:
If you compare the names of variables from model output from load_data, the names are differnet than those of read.output. For example, load_data for an ed run will have the name "Lwdown" and read.output will have the name "surface_downwelling_longwave_flux_in_air" and no "Lwdown". 

 
## Description

Output of load_data() matches the [names in bety, but not the standard_names](https://pecan.gitbooks.io/pecan-documentation/content/developers_guide/Adding-an-Input-Converter.html#the-variable-names-should-be-standardname). Where the standard names don't match bety names for variables, load_data() just returns the original variable. This could lead to errors as some variables (ie NEE) share a common name across many standards.

read.output() returns variables that match the contents of year.nc.var files for any given model run. These match the [mstmip names](https://github.com/PecanProject/pecan/blob/e6e76a5ec36607eb3315f144e5465545ba6fb2e5/db/inst/bety_mstmip_lookup.csv), with the exception of a case change in radiation (ie LWdown -> Lwdown). 

## Context
For visualization and benchmarking the data and model output need common names and units.

mdietze:
There seem to be multiple distinct issues in this issue:

1) If ED2 is returning surface_downwelling_longwave_flux_in_air when it's supposed to be returning Lwdown then that's a bug in model2netcdf.ED2, not in read.output.

2) The statement "Where the standard names don't match bety names for variables, load_data() just returns the original variable" is genuine bug -- load_data should never return a variable that's not in the Format record. If the variable is in the Format it needs to be matched to an existing BETY variable. For any variable that's in Format it should convert the name and units to BETY. Could you give a specific reproducible example of where this is occurring? Is it an error in the Formats record or load_data itself (though regardless it seems there should be assertions in load_data or query.format.vars to catch record errors)

3) A small change in capitalization of LWdown vs Lwdown should be a quick fix -- could you just fix and PR that?

mccabete:
2. It looks like it's a bug in the format record. All the variables that aren't convertible exist in the format record, but have units that load_data wasn't expecting.   Here's an example: 

File_Path<-"/fs/data3/tmccabe/valicali/AMF_US-Dk3_BASE_HH_3-1.csv"
File_format<-query.format.vars(bety = bety, format.id = 5000000002)
site<-query.site(853,bety$con)

observations<-load_data(data.path = File_Path, format= File_format, time.row = File_format$time.row, site = site) 

[...]

PR #1419, opened 2017-05-25
query.format.vars now catches unconvertible units and throws a warning.

mccabete:
This will catch unconvertible units earlier in the workflow. It throws a warning and not an error because people could be using imperfect formats that work correctly for the variable they're interested in.  This is also a helpful fix for those making formats to double check that their formats will work with pecan. 

## Motivation and Context
 Fixes part of issue #1415, catches it earlier in the workflow. 

## Types of changes

## Checklist:


db/R/query.format.vars.R
```diff
@@ -118,5 +118,18 @@ query.format.vars <- function(bety,input.id=NA,format.id=NA,var.ids=NA){
...
   }
+  
+  # Check that all bety units are convertible. If not, throw a warning. 
+  for(i in 1:length(format$vars$bety_units)){
+    
+    if( format$vars$storage_type[i]!=""){ #units with storage type are a special case
+      # This would be a good place to put a test for valid sotrage types. Currently not implemented. 
+    }else if(udunits2::ud.are.convertible(format$vars$input_units[i], format$vars$pecan_units[i]) == FALSE){ 
+      if(misc.are.convertible(format$vars$input_units[i], format$vars$pecan_units[i]) == FALSE){
+        logger.warn(c(paste(c("Units not convertible for", ""),c(format$vars$input_name[i], format$vars$input_units[i])), paste(".  Please make sure the varible has units that can be converted to", format$vars$pecan_units[i])))
```
mdietze: technically I don't think you need to ```c()``` and ```paste``` everything together within the logger functions, they'll do that automatically. Also, more generally, you rarely need to use ```c``` within paste unless you want to vectorize the creation of a bunch of similar strings.

mccabete: Cool. Fixed it.

PR #1430, opened 2017-05-31
Update mstmip names to bety names look up table

mccabete:
I populated a table that has the mstmip names with some of their bety analogues. The table is used to convert from bety names to mstmip names during the query.format.vars step. It's worth noting that the population of this file is likely incomplete, and that there are mstmip variables with no clear bety analogues. 

## Motivation and Context
This change partially solves issue #1415, and insures that model output variable names will match names from data that will be compared to the model output. 

## Types of changes

## Checklist:


mdietze:
> It's worth noting that the population of this file is likely incomplete, 

Any reason the file shouldn't be complete? Would prefer to hold on the PR until it's complete

> and that there are mstmip variables with no clear bety analogues.

Then those mstmip variables should be added to bety so that the bety and mstmip variable are equivalent. Once they are equivalent, you won't need to populate the last column in this table.

PR #1433, opened 2017-06-01
Edits for benchmarking functions: `align_data`, `query_format_vars` and `load_data`

bcow:
## Description

`query_format_vars`
 - A section of code just needed to be moved inside an if statement

`align_data`
- More extensive changes focused on fixing unit problems and generalizing to accommodate data that may have dates but no time stamps. 

`load_data`
- Adding functionality of subsetting by year. Start and end date were in the arguments but never used.

## Types of changes


db/R/query.format.vars.R
```diff
@@ -112,6 +112,22 @@ query.format.vars <- function(bety,input.id=NA,format.id=NA,var.ids=NA){
...
+    # Check that all bety units are convertible. If not, throw a warning. 
+    for(i in 1:length(format$vars$bety_units)){
+      
+      if( format$vars$storage_type[i] != ""){ #units with storage type are a special case
+        
+        # This would be a good place to put a test for valid sotrage types. Currently not implemented. 
+        
+      }else if(udunits2::ud.are.convertible(format$vars$input_units[i], format$vars$pecan_units[i]) == FALSE){ 
+        
+        if(misc.are.convertible(format$vars$input_units[i], format$vars$pecan_units[i]) == FALSE){
```
istfer: @bcow @mccabete not super important as this gives a warning not  an error but I think there needs to be another condition here, checking for same units..

Because above input_units are populated with bety_units, and now there are cases where input_units and pecan_units the same but are not convertible according to this check therefore meets this condition.

E.g. (In Ameriflux format) for variable CO2 both inputs_units and pecan_units are "ppm (= umol mol-1)" and this check gives:

   ```
Units not convertible for CO2 with units of ppm (= umol mol-1) .  Please 
   make sure the varible has units that can be converted to ppm (= umol 
   mol-1) 
```

Comment (mdietze) [inline db/R/query.format.vars.R]: @istfer I'm not following you're request -- this already is checking if the units are convertable. If the units are convertable there's no need to check if they're the same, since that is just a special case of being convertable to itself.

As for the specific units listed, "ppm (= umol mol-1)", this is a great example of this function working as intended. That string does not represent valid units. Even though the string may be the same in both places, it's wrong in both places and needs to be fixed (rather than adding a special case that would let it through).

Comment (istfer) [inline db/R/query.format.vars.R]: @mdietze I thought this bit of code was for checking whether `input_units` and `pecan_units` are convertible when they are different, and do nothing when they're the same given these warnings:

```
2017-06-01 14:05:20 WARN   [query.format.vars] : 
   Units not convertible for CO2 with units of ppm (= umol mol-1) .  Please 
   make sure the varible has units that can be converted to ppm (= umol 
   mol-1) 
2017-06-01 14:05:20 WARN   [query.format.vars] : 
   Units not convertible for DOY with units of DOY .  Please make sure the 
   varible has units that can be converted to DOY 

```
It doesn't give such warnings for most of the other variables because for units like `cm, celcius`
```
ud.are.convertible("cm", "cm")
ud.are.convertible("celsius", "celsius")
....
```
etc. `ud.are.convertible` returns `TRUE` and passes

if checking whether the units are recognized by udunits was (also) the intended behaviour, then that's fine

PR #1437, opened 2017-06-05
IC workflow

istfer:
## Description

Changes to localize IC workflow and rearrangements to the current modules accordingly

Testing with NASA FFT inventory data.

## Motivation and Context

Unlike met process, we either have to pass the whole data during the remote execution, or copy the data files to remote in the beginning of the workflow (which makes debugging difficult), so we decided to handle IC locally and copy the files in the end if model will be run on remote.

## Types of changes

## Checklist:


modules/data.land/R/ic_process.R
```diff
@@ -119,9 +138,41 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
+        remote.execute.cmd(settings$host, "mkdir", c("-p", out.dir.remote))
+        PEcAn.utils::remote.copy.to(settings$host, pss.file.local, pss.file.remote)
+        PEcAn.utils::remote.copy.to(settings$host, css.file.local, css.file.remote)
+        PEcAn.utils::remote.copy.to(settings$host, site.file.local, site.file.remote)
+        
+        # update DB record
+        ic_input <- db.query(paste0("SELECT * from inputs where id = ", putveg.id), con)
+        fname    <- db.query(paste0("SELECT * from formats where id = ", ic_input$format_id), con)
+        mtype    <- db.query(paste0("SELECT type_string from mimetypes where id = ", fname$mimetype_id), con)
+        dbfile.input.insert(in.path = out.dir.remote, in.prefix = model_file[["file_name"]],
```
mdietze: (1) If you're adding the remote dbfile record to an existing Input record that was created during local processing, then this should be a dbfile.insert not a dbfile.input.insert, because the input file already exists and it is vital that it not be duplicated. That will also save you the trouble of not having to query the format or mime-type 
(2) This could (and probably should) be generalized to a function that copy **ANY** existing input folder from local to remote (while updating the DB records)? That would be a quite generally useful feature.

modules/data.land/R/load_veg.R
```diff
@@ -0,0 +1,107 @@
...
+##' @author Istem Fer
+load_veg <- function(new_site, start_date, end_date, 
+                     source_id, source, icmeta = NULL, format_name = NULL, 
+                     machine_host, dbparms, outfolder, overwrite = FALSE, ...){
+  
+  
+  bety <- dplyr::src_postgres(dbname   = dbparms$bety$dbname, 
+                              host     = dbparms$bety$host, 
+                              user     = dbparms$bety$user, 
+                              password = dbparms$bety$password)
```
mdietze: This snippet of code is being cut-and-pasted everywhere in PEcAn -- it really should be a simple function like db.open (indeed, db.open should probably be deprecated and replaced globally by this new function).

modules/data.land/R/load_veg.R
```diff
@@ -0,0 +1,107 @@
...
+  }else{
+    logger.severe("Can't match code to species. No valid format found.")
+  } 
+  
+  # match code to species ID
+  # no lower case
+  obs[[code.col]] <- toupper(obs[[code.col]])
+  
+  # filter dead trees
+  obs <- remove_dead_trees(obs, code.col)
```
mdietze: Do we really want to remove them? Are there no models that include snags?

modules/data.land/R/load_veg.R
```diff
@@ -0,0 +1,107 @@
...
+  
+  # merge with data
+  tmp <- spp.info[ , colnames(spp.info) != "input_code"]
+  
+  # a hack for now to have a similar structure as the FIA case
+  veg_info      <- list() 
+  
+  if(!is.null(icmeta)){
+    # the first sublist can be for the metadata maybe?
+    # to be handled by veg2model later
```
mdietze: Ok, so remind me, have we generalized the load_data call here to load separate files that have plot info and merge them together as part of the wide2long? Or do we need to handle plot info files separately?  This is a challenging design question since we don't know a priori how many (and which) formats will keep plot info in a separate file vs how many will just have that info as columns in one large file that also includes the tree info.

modules/data.land/R/put.veg.module.R
```diff
@@ -49,24 +35,29 @@
     logger.severe(paste(fcn, "does not exist."))
   }
   
+  # NOTE : match_pft is now called in veg2model
```
mdietze: I'm not sure that's a good design decision -- any function that HAS TO be called within EVERY model-specific function really seems like a function that should be in the part of the workflow that PEcAn controls, not the model-specific code.

modules/data.land/R/write_veg.R
```diff
@@ -1,28 +1,20 @@
...
-write_veg <- function(outfolder, start_date, end_date, veg_info, overwrite = FALSE, ...){
+write_veg <- function(outfolder, start_date, end_date, veg_info, site_name, source){
   
   #--------------------------------------------------------------------------------------------------#
   # Save rds file and return results data frame
   
   start_year    <- lubridate::year(start_date)
   end_year      <- lubridate::year(end_date)
-  out_file      <- paste("FIA", start_year, end_year, "veg", "rds", sep = ".")
+  out_file      <- paste(site_name, source, start_year, end_year, "veg", "rds", sep = ".")
```
mdietze: as per above, drop site_name

models/ed/R/veg2model.ED2.R
```diff
@@ -18,10 +38,9 @@ veg2model.ED2 <- function(outfolder, start_date, end_date,
   formatnames <- c("ED2.cohort", "ED2.patch", "ED2.site")
   dbfilenames <- c("css.file", "pss.file", "site.file")
   
-  file.prefix <- paste("siteid", site_id, source, start_year, end_year,
+  file.prefix <- paste(site_name, source, start_year,
```
istfer: :+1: right, I'll switch back to id, but keep the name option there

modules/data.land/R/extract_FIA.R
```diff
@@ -71,7 +72,7 @@ extract_FIA <- function(lon, lat, start_date, gridres = 0.075, dbparms){
     
   logger.debug(paste0("Found ", nrow(pss.info), " patches for coordinates lat:", lat, " lon:", lon))
   
-  fia.info[[1]] <- pss.info
+  veg_info[[1]] <- pss.info
```
istfer: oh yeah, will change

modules/data.land/R/load_veg.R
istfer: good point

modules/data.land/R/put.veg.module.R
istfer: the thing with both `match_species_id` and `match_pft` functions is that they require the veg data itself (or a particular column of it) to operate on. If we want them to be called outside the functions that write files, then we either need to pass the veg_info around (which we were trying to avoid I believe) or need to write a file after each step and pass around the file ids, then it would look like this:


- raw veg file
- load data & write: a somewhat standard veg file (like met CF files)
- match spp ids & write : veg data with matched spp ids (currently what is being written)
- match pfts & write : this file would also be model specific which makes it sort of unnecessary to write
- veg2model : final ic files (currently being written)

do we want to write all these steps, or am I missing something?

models/ed/R/veg2model.ED2.R
```diff
@@ -33,32 +52,26 @@ veg2model.ED2 <- function(outfolder, start_date, end_date,
...
+    trk     <- ifelse(!is.null(pss$trk), pss$trk, 1)
+    age     <- ifelse(!is.null(pss$age), pss$age, 100)
     
     pss <- data.frame(time = time, patch = n.patch, trk = trk, age = age)
     
     PEcAn.utils::logger.info(paste0("Values used in the patch file - time:", 
                                     pss$time, ", patch:", pss$patch, ", trk:", 
                                     pss$trk, ", age:", pss$age))
     
+    # TODO : soils can also be here, passed from settings
```
istfer: got it, haven't had the chance to dig into soils prototype yet but we can discuss how to make them work together

models/ed/R/veg2model.ED2.R
```diff
@@ -33,32 +52,26 @@ veg2model.ED2 <- function(outfolder, start_date, end_date,
...
+  # veg_info[[1]] either has 
+  # i)   full pss info (FIA case)
+  # ii)  info passed from settings
+  # iii) no info
+  pss <- as.data.frame(veg_info[[1]], stringsAsFactors = FALSE)
+  
+  # for FIA these steps are unnecessary, it already has the pss info
+  if(source != "FIA"){
+    time    <- ifelse(!is.null(pss$time), pss$time, start_year)
+    n.patch <- ifelse(!is.null(pss$n.patch), pss$n.patch, 1)
```
istfer: `veg_info[[1]]` can be i) empty ii) complete (for FIA case it's the pss information, or if all meta kind of info (age, patch, area etc.) were passed via settings) iii) have some (plot) info whatever passed via settings

not sure how to get the number of patches from `veg_info[[2]]`, you mean see if that is provided in the data and assume 1 patch if not?

modules/data.land/R/get.veg.module.R
```diff
@@ -2,107 +2,96 @@
...
+  bety <- dplyr::src_postgres(dbname   = dbparms$bety$dbname, 
+                              host     = dbparms$bety$host, 
+                              user     = dbparms$bety$user, 
+                              password = dbparms$bety$password)
+  con  <- bety$con
   
   if(input_veg$source == "FIA"){
     
-    veg_info <- extract_FIA(lon, lat, start_date, gridres = 0.075, dbparms)
+    fcn <- "extract_FIA"
```
istfer: yes that's what I had in mind, one branch for sources that have their own database (for now FIA is the only case, fcn args and name might be generalized later) and one requires using `load_data`, thought these two types would require different types of fcn args

modules/data.land/R/ic_process.R
```diff
@@ -12,12 +12,16 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
-  start_date <- settings$run$start.date
-  end_date   <- settings$run$end.date
+  run_start  <- settings$run$start.date
   model      <- settings$model$type
   host       <- settings$host
   dbparms    <- settings$database
   
+  # Handle IC Workflow locally
+  if(host$name != "localhost"){
+    host$name <- "localhost"
```
istfer: for (1) I don't overwrite `settings$host`, and use it in the end to sync the files.. for (2) all the local directories are derived from `dir` string which is reassigned to the local directories in the next line

modules/data.land/R/ic_process.R
istfer: (1) right, will fix that
(2) aren't remote.copy.to / remote.copy.from functions doing that? then updating the db record would be 1 line of code (dbfile.insert)?

modules/data.land/R/load_veg.R
istfer: agreed

models/ed/R/veg2model.ED2.R
mdietze: Yep, we already have to designate a column in veg_info[[2]] to indicate plot/patch designations and we can use that to count up the number of unique plots.

models/ed/R/veg2model.ED2.R
mdietze: I was suggesting source & start_year rather than switching back to id

modules/data.land/R/load_veg.R
istfer: no we haven't, I haven't seen such plot info files to be loaded together with the vegetation data or separately, or did I miss something? we discussed how to handle this but haven't actually generated files/formats

I included the functionality to pass some plot info via settings

models/ed/R/veg2model.ED2.R
istfer: got it, then we'll understand to which site the file belongs only from the folder name? (without actually looking into the files I mean)

modules/data.land/R/ic_process.R
mdietze: ...plus a line that queries the input to get it's ID, plus a line or two that pulls out the paths on the local and remote site, plus checking for errors and edge cases. Doesn't have to be a huge function, but I think it would be convenient to have a function that just takes in the input.id, remote folder, and con and handles both file transfer and DB updating. Might return the remote dbfile record so you have the IDs and remote path and prefix.

models/ed/R/veg2model.ED2.R
istfer: hmm I see, this will be relevant for GapMacro files, will change

modules/data.land/R/load_veg.R
mdietze: I know Josh's files are all formatted this way (separate plot and tree files -- actually, also separate files for seedlings and saplings as well).

modules/data.land/R/put.veg.module.R
mdietze: It's been a while since we discussed this, but I thought we talked about a module that would both load data and match species before writing, and then a second module that would match PFTs and then call veg2model.

modules/data.land/R/extract_FIA.R
```diff
@@ -114,9 +115,48 @@ extract_FIA <- function(lon, lat, start_date, gridres = 0.075, dbparms){
...
+                              host     = dbparms$bety$host, 
+                              user     = dbparms$bety$user, 
+                              password = dbparms$bety$password)
+  con  <- bety$con
+  
+  format.name <- "fia" 
+  code.col    <-  "spcd"
+  
+  # match code to species ID
+  spp.info <- match_species_id(input_codes = obs[[code.col]], format_name = format.name, bety = bety)
```
istfer: yes, this is actually what it is, as I mention in the comment below the first module kind of needs to have two versions load_veg and extract_veg..similar to met.process -> download.raw.met : i) site ii) regional.. only extract_FIA is a special case as it's the only DB source for now, I can refactor this

modules/data.land/R/put.veg.module.R
istfer: > a module that would both load data and match species before writing

yup, this is currently the case

> a second module that would match PFTs and then call veg2model

ah, I see, yes I kind of revert this one because it was two lines of code readRDS and match_pft to be called before veg2model.. thought they might as well be in the veg2model, and  I liked resembling veg2model to met2model by passing the path and prefix.. I can rearrange it if you prefer match_pft to be called outside

mdietze: Well, you'll want the module to still take in path and prefix, since it's the module itself that'd get passed to convert.inputs. But that module can then call the model-specific veg2model.[MODEL]. As you note earlier, no point in having it write to file in between.

istfer: sure

one question though, now that I removed `remove_dead_trees` from upstream veg data will come to `match_pft` with dead tree codes if I take match_pft outside of `veg2model`. If I don't remove them, it won't match with PFTs (in ED's FFT case I think SNAG wont). If I'm going to remove dead trees before match_pft, I have the same problem that not all mode IC want to get rid of dead trees so I'll need checks (e.g. if ED, remove dead trees : but we're trying to avoid hacks I believe, or should I pass a flag via settings and check for that flag?) how should I resolve it?

I don't have that problem if match_pft lives inside veg2model

mdietze: Ok, got it. Would it be possible to modify the check in match.species to allow DEAD/SNAG/etc to be valid unmatched species if a model hasn't set those up as being part of a PFT?

istfer: I can put a hard coded check, until now I only encountered 3 codes for dead trees..I could also look for "dead" in the commonname but there could be trees that have "dead" in their common name

PR #1444, opened 2017-06-06
Beginning work on metric_run

bcow:
Adding pseudocode for `metric_run` so it can be accessible. 
`metric_run` is a benchmarking metric function that determines whether or not a run was successful. It will be run regardless of whether or not benchmarking is turned on in settings, however, if benchmarking is turned on and "metric_run" is in the list of metrics, the score calculated by `metric_run` will be recorded in the database. This involves a slight change to `workflow.R` that I did not include in this PR because I want to make sure it doesn't break anything first.


mdietze:
@bcow need to run make document

bcow:
I've run make multiple times. I'm not sure what else I need to do?

infotroph:
@bcow looks like you need to add and commit `man/metric_run.Rd`

PR #1447, opened 2017-06-08
use 'mlegp' pkg in PDA

istfer:
Switching to use another (much faster) package for GP fitting in emulator
## Description

- Now don't need to draw from probability space, mostly changes regarding using the new package
- Few changes to accommodate updated benchmark code

## Types of changes

## Checklist:


modules/emulator/R/minimize.GP.R
```diff
@@ -227,7 +230,7 @@ mcmc.GP <- function(gp, x0, nmcmc, rng, format = "lin", mix = "joint", splinefcn
       currY <- get.y(gp, xcurr, n.of.obs, llik.fn, priors, settings)
       ycurr <- currY$posterior.prob
       pcurr <- currY$par
-      newY  <- get.y(gp, xnew, n.of.obs, llik.fn, priors, settings)
+      newY  <- get.y(gp, xnew, n.of.obs, llik.fn, priors, settings, currY$llik.par)
```
mdietze: Here and below, how does llik.par end up in currY$llik.par if it's not passed in a few lines above in the ```currY <- get.y``` line?

Wait, I just realized that you steathly snuck in the Gibbs update **before** the M-H update by burying it in the function called by get.y -- this is going to be really unintuitive to folks, so I'd add some inline documentation to the line above making this explicit.

db/R/dbfiles.R
```diff
@@ -458,9 +466,15 @@ dbfile.check <- function(type, container.id, con, hostname=PEcAn.utils::fqdn(),
     
     if(nrow(dbfiles) > 1){
```
mdietze: Here and in the previous case, rather than having an additional ```if``` nested inside this iff, couldn't this line just be
```
if(nrow(dbfiles) > 1 &&  !return.all){
```

istfer: so is it gonna be like this?

``` 
if (is.null(hostid)) {

}else if (machine.check){
    
    if(nrow(dbfiles) > 1 &&  !return.all){
           logger.warn("Returning last updated record.")
           return(dbfiles[dbfiles$updated_at == max(dbfiles$updated_at),])
    }else{
        return(dbfiles)
    }
}else{
  ...
}
```

mdietze: That's what I was thinking

istfer:
@mdietze I think I got it this time..MCMC mixes, converges, and returns the true values

modules/emulator/R/minimize.GP.R
```diff
@@ -129,6 +129,7 @@ get.y <- function(gp, xnew, n.of.obs, llik.fn, priors, settings, llik.par = NULL
   
   if(is.null(llik.par)){
     llik.par <- pda.calc.llik.par(settings, n.of.obs, SS)
+    return(llik.par)
```
mdietze: Could you add a comment in the code here explaining why you return from the function here? I'm not really following what you're trying to do, and I do find it an odd design decision to return completely different things depending on whether llik.par is null or not. That also suggests a need to update the roxygen info so that such polymorphism is well documented.  That said, I'm also fairly suspicious that this might not be what you actually want to do in the MCMC.

modules/emulator/R/minimize.GP.R
```diff
@@ -211,33 +212,45 @@ mcmc.GP <- function(gp, x0, nmcmc, rng, format = "lin", mix = "joint", splinefcn
...
+
+      newY  <- get.y(gp, xnew, n.of.obs, llik.fn, priors, settings, currll)
       ynew  <- newY$posterior.prob
+
       if (is.accepted(ycurr, ynew)) {
         xcurr <- xnew
-        pcurr <- newY$par
-        
+        #pcurr <- newY$par
+        #curll <- newY$llik.par
```
mdietze: In this version, you're only updating ```xcurr <- xnew``` but you also need to update ```SScurr <- SSnew``` (though I can't tell in the code where you're tracking the current summary statistics here, is that ycurr and ynew?)

modules/emulator/R/minimize.GP.R
```diff
@@ -211,33 +212,45 @@ mcmc.GP <- function(gp, x0, nmcmc, rng, format = "lin", mix = "joint", splinefcn
...
         accept.count <- accept.count + 1
       }
+      
+      # update likelihood parameters conditioned on model parameters
+      # when llik.par (= currll) is not passed, get.y updates ll-params, given model parameters
+      # and returns llik.par
+      currll <- get.y(gp, xcurr, n.of.obs, llik.fn, priors, settings)
```
mdietze: This is what I was worried you were doing when you made get.y polymorphic above. In addition to being confusing, this design ends up making a call to predict.gp and a random draw of SS that's different from the one used in the M-H algorithm. That is both incorrect and an unnecessary computation. What would be much simpler would be to drop the polymorphism in get,y (the return if is.null(llik.par), which should never happen), have get.y return SS as an additional arg, and then update SS in the M-H accept step as noted earlier.  By contrast, here I think all you need is
```
## update statistical parameters by Gibbs sampling
currll <- pda.calc.llik.par(settings, n.of.obs, SScurr)
```

istfer: right

how about I split up the get.y function into two, `get_ss` and `get_y`, so that it's easier to follow

I push the example, but I can put it back together if you'd prefer that

PR #1461, opened 2017-06-12
TDM Vignette & Bug Fixes

jsimkins2:
Created a vignette for Temporal Downscale Meteorology functions and updated parts of code that weren't behaving. 

## Types of changes

- New Feature -> Vignette
- Bug Fix -> added "@export" into a few functions which didn't have it so now they can be written in NAMESPACE
- Bug Fix -> nc.merge() wasn't handling the upscaling feature properly


modules/data.atmosphere/NAMESPACE
```diff
@@ -1,5 +1,8 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(predict,met)
+S3method(predict,subdaily.function)
+S3method(predict,subdaily.workflow)
```
mdietze: I'd recommend changing the names of these functions to use underscore rather than period so they export normally, rather than as S3method

modules/data.atmosphere/R/tdm_model_train.R
```diff
@@ -18,11 +18,13 @@
...
+##' @export
 # -----------------------------------
 #----------------------------------------------------------------------
 # Begin Function
 #----------------------------------------------------------------------
-model.train <- function(dat.subset, n.beta, resids = resids, threshold = NULL) {
+model.train <- function(dat.subset, v, n.beta, resids = resids, threshold = NULL, ...) {
+  dat.subset$year <- as.ordered(dat.subset$year) 
   if (v == "air_temperature") {
     dat.subset$year <- as.ordered(dat.subset$year)
```
mdietze: This line seems redundant

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
```diff
@@ -0,0 +1,55 @@
...
+### Extract Training Data and Merge All Years Into 1 File 
+
+```{r}
+library(PEcAn.data.atmosphere)
+library(PEcAn.DB)
+download.Fluxnet2015(sitename = 'Niwot Ridge Forest/LTER NWT1 (US-NR1)', username = 'pecan', overwrite = FALSE, outfolder = '~/Example', 
+   start_date = '2000/01/01', end_date = '2014/12/31')
+
+bety     <- list(user = 'bety', password = 'bety', host = 'localhost', dbname = 'bety', driver = 'PostgreSQL', write = TRUE)
+bety$con <- db.open(bety)
```
mdietze: This seems a bit of a hack around the switch to using dbplyr.

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
```diff
@@ -0,0 +1,55 @@
...
+* Fluxnet2015 30-minute/hourly
+* Ameriflux 30-minute/hourly
+
+### Extract Training Data and Merge All Years Into 1 File 
+
+```{r}
+library(PEcAn.data.atmosphere)
+library(PEcAn.DB)
+download.Fluxnet2015(sitename = 'Niwot Ridge Forest/LTER NWT1 (US-NR1)', username = 'pecan', overwrite = FALSE, outfolder = '~/Example', 
+   start_date = '2000/01/01', end_date = '2014/12/31')
```
mdietze: start_date, end_date, lat, lon, in,prefix, outfolder's root folder, etc. are all things that should be variables, not hard coded. This is less buggy and will make it far easier for others to apply this example at new sites. I'd recommend defining all these in their own r code block that'd be under the previous header.

mdietze: note: in this case by "outfolder's root folder" I mean defining ```outfolder <- "~/Example/"```, adding an explicit ```dir.create``` to make sure it exists, and then using file.path to create step-specific in and outdirs on the fly (e.g. in nc.merge it'd be ```in.path = file.path(outfolder,"gapfill")``` and ```outfolder=file.path(outfolder,"training_data")```. No need to include "NR1_" in the name of the training data folder.

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
```diff
@@ -0,0 +1,55 @@
...
+
+```{r}
+download.MACA(outfolder = "~/Example/", start_date = "2020-01-01", end_date = "2020-12-31", site_id = 772, lat.in=40.0329, lon.in=-105.546, model = "BNU-ESM", scenario = "rcp85", ensemble_member = "r1i1p1")
+
+```
+
+### Predict Subdaily Data Using Statistics From Training Data 
+
+```{r}
+predict.subdaily.workflow(outfolder = "~/Example/downscaled_data", in.path = "~/Example/MACA_site_0-772/", in.prefix = "MACA.BNU-ESM.rcp85.r1i1p1", start_date = "2020-01-01", end_date = "2020-12-31", lm.models.base = "~/Example/lm_model_output", dat.train_file = "~/Example/NR1_training_data/FLX_US-NR1_FLUXNET2015_SUBSET_HH_1998-2014_1-3_dat.train.nc", ens.hr = 3)
```
mdietze: does "workflow" really need to be in the name of the function? Could it just be "predict.subdaily"?

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
```diff
@@ -0,0 +1,55 @@
...
+```
```
mdietze: After generating the met, it would be nice to include a simple visualization (e.g. for one variable such as air temperature) that might show the original (non-downscaled) data and then the ensemble of downscaled version over some short period, so the inferred diurnal cycle and ensemble variability are really visible.  This would also have the benifit of showing the user how to **use** the resulting product.

mdietze: note: @jsimkins2 if you don't have an example of such a plot sitting around that you can just plop in, I'm fine with that coming in a later PR -- I want to get the CRUNCEP fix in ASAP.

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
jsimkins2: Is there an example using dbplyr somewhere?

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
jsimkins2: Yes, it could be predict.subdaily. This is the main workflow for this however, and we already have "predict.subdaily.function()" and "predict.met". If you think "predict.subdaily" is distinct enough I'll change it

mdietze: Well, I was going to also point out that predict.subdaily.function is a pretty non-descript function name as well, but since it appears to only be used internally I wasn't as worried about it. If you could update that, and predict.met, to more precise names I'd be in favor of that too, but it's not required.

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
mdietze: This code
```
bety <-dplyr::src_postgres(host = bety$host, user = bety$user, password = bety$password, dbname = bety$dbname)
```
has been used in lots of places around PEcAn, though I've asked someone (I'm blanking on who at the moment, to wrap this in a dplyr-compliant function that'd replace PEcAn.DB::db.open (but would be called the same way you're doing right now).

jsimkins2: thanks!

modules/data.atmosphere/vignettes/tdm_downscaling.Rmd
```diff
@@ -14,24 +14,42 @@ Examples:
...
-met2CF.csv(lat = 40.0329, lon = -105.546, overwrite = FALSE, in.path = '~/Example', in.prefix = 'FLX_US-NR1_FLUXNET2015_SUBSET_HH_1998-2014_1-3', outfolder = '~/Example/CF', start_date = '2000/01/01', end_date = '2014/12/31', format)
-
-metgapfill(lst = -7, overwrite = FALSE, in.path = '~/Example/CF', in.prefix = 'FLX_US-NR1_FLUXNET2015_SUBSET_HH_1998-2014_1-3', outfolder = '~/Example/gapfill', start_date = '2000/01/01', end_date = '2014/12/31')
-
-nc.merge(in.path = "~/Example/gapfill", in.prefix = "FLX_US-NR1_FLUXNET2015_SUBSET_HH_1998-2014_1-3", outfolder = "~/Example/NR1_training_data", start_date = "2000-01-01", end_date = "2014-12-31", upscale = "hour")
+sitename   <- 'Niwot Ridge Forest/LTER NWT1'
+outfolder  <- dir.create(file.path('~/Example'))
+start_date <- '2000/01/01'
+end_date   <- '2014/12/31'
+download.Fluxnet2015(username = 'pecan', overwrite = FALSE)
```
mdietze: @jsimkins2 here and below -- even when you define these variables, you still need to **explicitly pass them to the functions**!

Also, I think you misunderstood my suggestion about using file.path -- you're still hard coding the paths throughout this script rather than using file.path to make the **names** on the fly. Also, you don't need to keep calling dir.create on every single one of them because that'd done on the fly in the functions, it's just the root directory that you need to define and create on the fly.  Right now, you're also redefining variables all over the place (e.g. outfolder) which isn't what I intended.  Set outfolder to "~/Example" and then never redefine it. Just in the function calls themselves pass in file.path(outfolder,<subfolder>) to specify each subfolder.

Finally, I was suggesting that you define ALL constants up front -- it's much easier for users if those are set once at the top rather than having to hunt through the code to figure out all the things that have to be modified.  And if the meaning of a variable isn't obvious, please provide a comment defining what it is.

PR #1462, opened 2017-06-12
Bugfix in Makefile

ashiklom:
`make` was not creating the files used to check if devtools, roxygen,
and testthat had been performed. This was causing `make` to run in full
_every_ time for pecan intsalls that didn't already have the
.install/devtools, etc. files existing, which of course defeats that
point.


ashiklom:
Latest commit should resolve the install bug in #1444. @bcow

Issue #1459, opened 2017-06-12
Reconcile Ameriflux formats records


bcow:
We currently have three format records for the same Ameriflux files.
The discrepancies are due to the fact that query.formats.vars and met2CF.CSV are parsing and converting dates differently.
See:
https://github.com/bcow/pecan/blob/master/db/R/query.format.vars.R#L89
vs.
https://github.com/bcow/pecan/blob/master/modules/data.atmosphere/R/met2CF.csv.R#L148

ankurdesai:
@mdietze @bcow @robkooper Latest PR from me fixed met2CF.csv, so we can depreciate the use of units in format records in favor of storage_type using strptime convention (now fixed in both AMERIFLUX_HH and FLUXNET2015). Now, we need to merge formats 1000000034 and 5000000002, which are identical. How do we do this easily, PSQL expert types, so that all inputs that were pointing to one point to the other? Which table and statement is needed, and which server should do it so that sync doesn't overwrite changes?

Also, 1000000005 should be deleted.

mdietze:
The BU duplicate that needs merging is 1000000033 (not 1-34) and it is not associated with any Inputs, and it doesn't seem to be hard-coded anywhere in the code, so I'm not sure it was ever actually used for anything. @ankurdesai once you give me a 👍 that there's no information in 1-33 that's not in 5-2, I'll go ahead and delete our duplicate

Unfortunately 1000000005 is associated with 20 Input records. Can all of these be deleted? Or should their format ID's be migrated to something else (and if so, what)

```
     id     |  site_id   | notes |         created_at         |         updated_at         |     start_date      |      end_date       |         name         | parent_id | user_id | access_level | raw | format_id  
------------+------------+-------+----------------------------+----------------------------+---------------------+---------------------+----------------------+-----------+---------+--------------+-----+------------
 1000000177 | 1000000043 |       | 2015-02-05 19:09:29.83299  | 2015-02-05 19:09:29.83299  | 2001-01-01 00:00:00 | 2005-12-31 00:00:00 | Ameriflux_site_1-43  |           |         |            4 |     | 1000000005
 1000000169 | 1000000061 |       | 2015-01-22 15:49:53.02283  | 2015-01-22 15:49:53.02283  | 2001-01-01 00:00:00 | 2006-12-31 00:00:00 | Ameriflux-1-61       |           |         |            4 |     | 1000000005
[...]

PR #1464, opened 2017-06-13
Update met2CF.csv to be consistent with load.data

ankurdesai:
Changes to met2CF.csv.R and related functions to work with time formatting convention used in storage_type instead of variable units. 

## Description
Function now uses same method as load.data.R in benchmark for time converseion. query.format.vars also return site time zone to use as default time zone for CSV files.

## Motivation and Context
Deals with #1459  

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/met2CF.csv.R
+++ b/modules/data.atmosphere/R/met2CF.csv.R
@@ -26,26 +26,24 @@
 ##' @author Mike Dietze, David LeBauer, Ankur Desai
 ##' @examples
 ##' \dontrun{
-##' library(PEcAn.DB)
-##' library(lubridate)
-##' library(RPostgreSQL)
-##' bety = list(user='bety', password='bety',
-##'             host='localhost', dbname='bety', driver='PostgreSQL',
-##'             write=TRUE)
-##' con <- db.open(bety)
-##' in.path <- '/home/carya/sites/willow/'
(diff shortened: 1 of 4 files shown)


modules/data.atmosphere/R/met2CF.csv.R
```diff
@@ -158,22 +170,19 @@ met2CF.csv <- function(in.path, in.prefix, outfolder, start_date, end_date, form
...
         ## Does not match any of the known date formats, add new ones here!
         PEcAn.utils::logger.error("datetime column is not specified in format")
       }
     } else {
-      datetime_units <- format$vars$input_units[datetime_index]  #lubridate function to call such as ymd_hms
+      datetime_raw <- alldat[, format$vars$input_name[datetime_index]]
+      datetime_units <- paste(format$vars$storage_type[datetime_index],collapse = " ")  #strptime convention
       if (datetime_units == "") {
-        datetime_units <- "ymd_hm"
+        datetime_units <- "%Y%m%d%H%M" #assume ISO convention
```
mdietze: I thought the ISO had some spaces and/or punctuation. Did this test well?

ankurdesai: This is the default if no string is specified in format record storage_type. This works for Fluxnet and ameriflux (tested through web interface) which use dates like 200401011330.

PR #1468, opened 2017-06-14
linkages paleon regional runs + first pass at issue #1448 

araiho:
I changed the file format for model2netcf.linakges and also fixed soil query in write.configs.linkages in order to use multi.settings and run linkages paleon regional runs.

I also started to create a workflow for paleon sda by creating a load_data function specific to paleon data assimilation data. Right now, this function can handle tree ring data from Harvard Forest and give a user back AGB, NPP, or AGB.pft means and covariances to be used in sda.enkf. Next step will be to align with the tutorial. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue #1448 ) 

## Checklist:


modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+    }
+    
+    ### Map species to model specific PFTs
+    if(any(var.names == 'AGB.pft')){
+      spp_id <- match_species_id(unique(dataset$species_id),format_name = 'usda',bety)
+      spp_id <- spp_id[spp_id$input_code!='HAVI4',]
+      pft_mat <- match_pft(spp_id$bety_species_id, settings$pfts, con = bety$con)
+      
+      x <- paste0('AGB.pft.', pft_mat$pft)
+      names(x) <- spp_id$input_code
```
mdietze: I'm not following what's going on here -- x appears to have length pft_mat, which will <= to the length of spp_id, so unless there's a 1:1 match between species and PFT this will fail.  I think the logic here, and below, is too LINKAGES specific. More generally you need to sum species-level AGB's to the PFT level, such as with a tapply, with the 1:1 mapping being a special case.

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+      dataset$pft.cat <- x[dataset$species_id]
+      dataset <- dataset[dataset$pft.cat!='NA_AbvGrndWood',]
+      variable <- sub('AGB.pft','AbvGrndWood',variable)
+      arguments <- list(.(year, MCMC_iteration, site_id, pft.cat), .(variable))
+      arguments2 <- list(.(year, pft.cat), .(variable))
+      arguments3 <- list(.(MCMC_iteration), .(pft.cat, variable), .(year))
+    } 
+    
+    print('Now, melting data')
```
mdietze: 1) use a logger function, not print
2) maybe a more descriptive message? What if I don't know what 'melt' is -- could you explain more generally the goal of the next few steps?

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+    
+    melt_id <- colnames(cast.test)[-which(colnames(cast.test) %in% variable)]
+    melt.next <- reshape2::melt(cast.test, id = melt_id)
+    mean_mat <- reshape2::dcast(melt.next, arguments2, mean)
+    
+    iter_mat <- reshape2::acast(melt.next, arguments3, mean)
+    cov.test <- apply(iter_mat,3,function(x){cov(x)})
+   
+    for(t in seq_along(obs.times)){
+      obs.mean[[t]] <- mean_mat[mean_mat$year==obs.times[t], variable]
```
mdietze: 1) This will fail if obs.times are not in years
2) This will also fail if tree ring data isn't the first dataset processed

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+      if(any(var.names == 'AGB.pft')){
+        obs.mean[[t]] <- rep(NA, length(x))
+        names(obs.mean[[t]]) <- sort(x)
```
mdietze: The sort here seems dangerous -- aren't there instances where it would cause the order of the names to be different from the order of the means in obs.means?

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+    format_full <- format <- PEcAn.DB::query.format.vars(input.id = input.id[[i]], bety, format.id = NA, var.ids=NA)
+    
+    format$na.strings <- 'NA'
+    
+    # ---- LOAD INPUT DATA ---- #
+    time.row <- format$time.row
+    vars.used.index <- setdiff(seq_along(format$vars$variable_id), format$time.row)
+    
+    print(paste('Using PEcAn.benchmark::load_data.R on format_id',format_id[[i]],'-- may take a few minutes'))
+    obvs[[i]] <- PEcAn.benchmark::load_data(data.path, format, start_year = lubridate::year(start_date), end_year = lubridate::year(end_date), site, vars.used.index, time.row)
```
araiho: I copied this from @bcow 's benchmark workflow. So, maybe it should change there too? or maybe there's a reason why it's that why that I'm not aware of?

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+    
+    dataset <- obvs[[i]]
+    variable <- var.names
+    
+    ### Tree Ring Data Product
+    if(format_id[[i]] == '1000000040'){
+      obvs[[i]] <- obvs[[i]][obvs[[i]]$model_type=='Model RW + Census',]
+      obvs[[i]]$AbvGrndWood <- obvs[[i]]$AbvGrndWood * .48
```
araiho: The tree ring data product isn't in terms of carbon and load.data doesn't have a biomass:carbon conversion. But, I can address 1.

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+      obvs[[i]]$NPP_1_C <- obvs[[i]]$NPP_1_C * .48
```
araiho: There are a lot of variables for NPP in BETY, so I was sort of confused about which one to pick.

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: 1) Yes, the dropping of time.row needs to be done in the function itself (and then adding the pulling of time.row from the format internally in the function) and then this needs to be change everywhere load_data is called. I've asked others to do this in the past, but it seems to have not happened.
3) I suspect the benchmark workflow is like that because you're often NOT grabbing all of the variables in a file

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+      arguments3 <- list(.(MCMC_iteration), .(variable), .(year))
+    }else{
+      print('ERROR: This data format has not been added to this function (ツ)_/¯ ')
+      stop()
+    }
+    
+    ### Map species to model specific PFTs
+    if(any(var.names == 'AGB.pft')){
+      spp_id <- match_species_id(unique(dataset$species_id),format_name = 'usda',bety)
+      spp_id <- spp_id[spp_id$input_code!='HAVI4',]
```
araiho: So this is Witch Hazel. It's not a spp in linkages, so I wasn't sure how to deal with it. Not sure what other models do with witch hazel.

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: outsmarted by Markdown -- the number 2 above was referring to point 3 earlier

modules/assim.sequential/R/load_data_paleon_sda.R
araiho: Yeah, I didn't really know what to do here. I'm just trying to create a column that maps species in the data to pft in the model.

modules/assim.sequential/R/load_data_paleon_sda.R
araiho: I was trying to put all of the names on first then match them next.

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: Yes, 1 is fine short term, but could you open an issue about 2. The fundamental design of load.data is that it should return things in output standard -- if it's missing the biomass:carbon conversion then it's failing in that.

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: Yes, but there's only one NPP in the output standard, and load.data is supposed to convert everything to output standard names and units, not BETY names and units.  That said, in the FORMATS record on the web, NPP_1_C probably is the correct match. Reconciling those two is a long term goal (and the look-up table to interconvert was just updated by a recent PR by @mccabete )

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: There are currently 6 PFTs that have witch hazel, do we don't want to delete it
```
     id     |                                         definition                                          |         created_at         |         updated_
at         |             name             | parent_id | pft_type | modeltype_id 
------------+---------------------------------------------------------------------------------------------+----------------------------+-----------------
-----------+------------------------------+-----------+----------+--------------
 2000000007 | Basic all plants PFT for the MAAT model.  Cloned from an ED2 example                        | 2016-04-11 15:00:39.008867 | 2016-04-11 15:03
:10.154362 | maat.all.plants              |        95 | plant    |   2000000001
         22 | Southern half of mid-successional temperate hardwoods (e.g. oak, hickory).  No default PFT. | 2011-02-01 20:41:20        | 2011-02-01 20:41
:20        | temperate.South_Mid_Hardwood |           | plant    |            1
         39 | all plants                                                                                  | 2011-09-27 15:53:10        | 2011-09-27 15:53
:10        | plants                       |           | plant    |            1
         83 | all plants                                                                                  | 2011-09-27 15:53:10        | 2015-01-06 14:03
:33.531563 | plants2                      |           | plant    |            1
         95 | all plants                                                                                  | 2011-09-27 15:53:10        | 2015-01-06 14:03
:53.548183 | plants3                      |           | plant    |            1
         96 | Growth respiration for all plants                                                           | 2011-09-27 15:53:10        | 2013-10-07 14:51
:52        | gr.plants                    |           | plant    |            1
(6 rows)
```
I think the more general question is what to do with ANY species that doesn't match the union of the species covered by the selected PFTs -- think about writing the code that way. Ideally, we don't want to loose biomass so it would be good to lump it into some PFT -- so the default should be to list the unmatched species and throw a logger.severe. But I could see having a flag that would allow you to drop unmatches species, especially when they're a tiny fraction of the community

modules/assim.sequential/R/load_data_paleon_sda.R
```diff
@@ -0,0 +1,148 @@
...
+  
+  start_date <- settings$state.data.assimilation$start.date
+  end_date   <- settings$state.data.assimilation$end.date
+  
+  obs.mean <- list()
+  obs.cov <- list()
+  
+  start.time <- format(lubridate::ymd(start_date),settings$state.data.assimilation$forecast.time.step)
+  end.time <- format(lubridate::ymd(end_date),settings$state.data.assimilation$forecast.time.step)
+  obs.times <- start.time:end.time
```
araiho: So, I was thinking about changing it to be a format like '%Y' which would make it more obvious what one means by forecast.time.step, but I'm not sure if that's the right thing to do?

modules/assim.sequential/R/load_data_paleon_sda.R
araiho: That would go in the match_pft function no?

modules/assim.sequential/R/load_data_paleon_sda.R
araiho: I'll fix 1. But I don't understand 2.

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: That sounds like a good place to put it. If you put it in match_pft then we probably want to return NA for those records, and then screen them out after, otherwise the species ID vector will be a different length than the data itself.

modules/assim.sequential/R/load_data_paleon_sda.R
mdietze: So how would that work if the timestep was, for example, 8 days like we did in the MODIS analysis? Would you write ```8 %d```? or ```8*%d```? Would lubridate or difftime or anything else be able to parse that automatically?

Issue #1477, opened 2017-06-19
read.output returns three different measures of time when dataframe = TRUE is turned on


mccabete:
When you run read.output with the ```dataframe = TRUE ``` option on, you will get en error unless you request the variable "time" and then it will return a datraframe with ```yourModelRun$time```, ```yourModelRun$year``` and ```yourModelRun$posix```.  The ```$time``` column is just day of year with decimal hour values, and the ```$year``` column is just the year of the posix code.  The names especially of $time and $posix sound redundant enough that  they're distinction isn't immediately clear.

## Possible Implementation
We may want to deprecate the ```$time``` and ```$year``` when we can.

We could also add another option that lets the user select what kind of time they would like, with the form ```read.output(... Time_type= c("posix", "day_of_year", "year"))``` with a default to posix.  The only hang-up is that if the user doesn't put ```dataframe = TRUE``` then they only can ask for a ```$time``` column.

mccabete:
This is actually a subset of issue #1415 as load_data returns a ```$posix``` a ```$year``` and a ```$datetime```. The ```$time``` variable isn't in the load_data() function, and ```$datatime``` is just a parsed together string that matches the format storage variable ex: ```200401210730```.

PR #1479, opened 2017-06-20
Data assimilation tutorial

mccabete:
This pull request updates the data assimilation tutorial with new read.output flags, and introduced the load_data and align_data functions. In addition, there are some fixes for the analyzing output tutorial. 

## Description
This also contains a slight fix for issue #1477 so that if ```dataframe = TRUE``` is on and time isn't a requested variable pecan doesn't throw an error. 

Query.format.vars got a warning if there are duplicate columns. Will be overkill if the format is wide, but will be useful for debugging if a format gets a duplicate time column. 

## Types of changes

## Checklist:


documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -101,46 +107,40 @@ You should see a list of long numbers. Each of these is an ID. In the Files pane
...
+
+annualNEEobs<-aggregate(obs$NEE, by= list(month(obs$posix)), simplify=TRUE, FUN =mean, na.rm= TRUE)
+
+plottable<-align_data(model.calc = arun, obvs.calc = obs, align_method = "match_timestep", var= "NEE")
+head(plottable)
 ```
 
 ####C. Finally, we can finally compare model to data
-In the earlier tutorial, you also compared NEE to the ensemble model run. Here we will do the same except for each sensitivity run. Recall that we had to convert units and make a scatter plot. 
+In the earlier tutorial, you also compared NEE to the ensemble model run. Here we will do the same except for each sensitivity run. Recall that we made a scatter plot.
```
annethomas: At first I read "except for" as meaning "excluding," so maybe "but" would be slightly clearer, or remove "except" altogether.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -101,46 +107,40 @@ You should see a list of long numbers. Each of these is an ID. In the Files pane
...
+plottable<-align_data(model.calc = arun, obvs.calc = obs, align_method = "match_timestep", var= "NEE")
+head(plottable)
 ```
 
 ####C. Finally, we can finally compare model to data
-In the earlier tutorial, you also compared NEE to the ensemble model run. Here we will do the same except for each sensitivity run. Recall that we had to convert units and make a scatter plot. 
+In the earlier tutorial, you also compared NEE to the ensemble model run. Here we will do the same except for each sensitivity run. Recall that we made a scatter plot. 
 
-Here is what you did last time: We converted units, which we do with the *udunits* package. The observations are in umol m-2 s-1. The model output is read in kg C ha-1 yr-1. *udunits* can convert among many units but cannot do the conversion from grams carbon to moles carbon, so we do that separately, with the factor of 12 g C per mol.
+Since we alinged the data, we got a dataframe with the variables we requested in a ```{r}NEE.m``` and a ```{r}NEE.o``` format. The ```{r}.o``` is for observations, and the ```{r}.m ``` is for model. The posix column allows for easy plotting along a timeseries.
```
annethomas: At first I read "Since" as "In the time since," so maybe "Because" or "When" 
Typo: alinged -> aligned
Explaining .o and .m is a nice touch.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -25,7 +25,13 @@ RunDir <- "PEcAn_99000000002"
```
annethomas: Line 11: should be "through the web interface and have learned" instead of "interface, have learned"
Line 17 "an sensitivity analysis" -> "a sensitivity analysis" in two places

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -101,46 +107,40 @@ You should see a list of long numbers. Each of these is an ID. In the Files pane
 
 Let’s try to read the output from a single run id, as you did in the earlier tutorial. **read.output** is a PEcAn function to read in variables from any PEcAn compliant output files, it requires a runid, the directory, start year, end year, and an optional list of variables. Make sure you have RunDir variable set to the run number and the year to the year you ran.
 ```{r,eval=FALSE}
```
annethomas: Line 108: "from any PEcAn compliant output files, it requires a runid," -> "from any PEcAn compliant output files, **and** it requires a runid,"

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -165,9 +167,17 @@ Now that you have the hang of it, here are a few more things to try:
 
 1. Try a different error functions, given actual NEE uncertainty. You learned earlier that uncertainty in half-hourly observed NEE is not Gaussian. This makes RMSE not the correct measure for goodness-of-fit. Go to *~/pecan/modules/uncertainty/R*, open *flux_uncertainty.R*, and click on the *source* button in the program editing pane. Then you can run:
 ```{r,eval=FALSE}
-unc <- flux.uncertainty(niwotnee,QC=rep(0,17520))
+unc <- flux.uncertainty(plottable$NEE.o,QC=rep(0,17520)) 
 plot_flux_uncertainty(unc)
 ```
 The figure shows you uncertainty (err) as a function of NEE magnitude (mag). How might you use this information to change the RMSE calculation?
 2. Try a few others parameters. Repeat the above steps but with a different parameter. You might want to select one from the sensitivity PDF that has a large sensitivity or from the variance decomposition that is also poorly constrained.
```
annethomas: Line 174: others -> other

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -165,9 +167,17 @@ Now that you have the hang of it, here are a few more things to try:
...
 ```{r,eval=FALSE}
-unc <- flux.uncertainty(niwotnee,QC=rep(0,17520))
+unc <- flux.uncertainty(plottable$NEE.o,QC=rep(0,17520)) 
 plot_flux_uncertainty(unc)
 ```
 The figure shows you uncertainty (err) as a function of NEE magnitude (mag). How might you use this information to change the RMSE calculation?
 2. Try a few others parameters. Repeat the above steps but with a different parameter. You might want to select one from the sensitivity PDF that has a large sensitivity or from the variance decomposition that is also poorly constrained. 
-3. Try RMSE against daily or annual NEE instead of half-hourly. In this case, first average the values up to daily in both the model and observations. You will need to think about how to deal with missing data. For example, you could make the model output have the same data points missing as observations, and then average. Or use some other criteria.
+3. Try RMSE against monthly NEE instead of half-hourly. In this case, first average the values up to monthly in the observations. Then, use align data to match the monthly timestep in model output.
```
annethomas: Maybe "align_data" would be clearer in line 175

utils/R/read.output.R
```diff
@@ -87,7 +87,7 @@ model2netcdf <- function(runid, outdir, model, lat, lon, start_date, end_date) {
 ##' @param start.year first year of output to read (should be greater than ) 
 ##' @param end.year last year of output to read
 ##' @param variables variables to be read from model output
-##' @param dataframe A boolean that will return output in a data.frame format with posix column. Usefull for align.data and plotting. 
+##' @param dataframe A boolean that will return output in a data.frame format with a time column. Usefull for align.data and plotting.
```
annethomas: usefull -> useful

utils/man/read.output.Rd
```diff
@@ -18,7 +18,7 @@ read.output(runid, outdir, start.year = NA, end.year = NA,
 
 \item{variables}{variables to be read from model output}
 
-\item{dataframe}{A boolean that will return output in a data.frame format with posix column. Usefull for align.data and plotting.}
+\item{dataframe}{A boolean that will return output in a data.frame format with a time column. Usefull for align.data and plotting.}
 }
```
annethomas: usefull -> useful

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -101,46 +107,40 @@ You should see a list of long numbers. Each of these is an ID. In the Files pane
...
-
-Navigate to the *~/output/dbfiles* folder. Read the Ameriflux NetCDF (.nc) file in  *~/output/dbfiles/Ameriflux_site-0-722/*  here *0-722* is the database site ID, a folder that was created when you selected the site and PEcAn downloaded the Ameriflux NetCDF (.nc) file from the Ameriflux website. The *0-722* may be a different number in your system depending on the site you chose and the database contents. Also note *US-NR1*, which refers to the Fluxnet site identifier for Niwot Ridge and the *2006* for the year.
+Recall reading Ameriflux NEE in the earlier tutorial. Read in an Ameriflux csv file, stored in the bety database. Here *722* is the database site ID which allows bety to assign the correct location of the data. The *722* may be a different number in your system depending on the site you chose and the database contents. Also note *US-NR1*, which refers to the Fluxnet site identifier for Niwot Ridge.
 ```{r,eval=FALSE}
-library(PEcAn.assim.batch)
-obs <- load.L2Ameriflux.cf("~/output/dbfiles/Ameriflux_site_0-772/US-NR1.2006.nc")
-names(obs)
-```
+File_Path<-"/fs/data3/tmccabe/AMF_US-NR1_BASE_HH_9-1.csv" #This is a path to where the observations are stored.
```
mccabete: I'm not sure. @mdietze Will the tutorials from instances be able to query bety?

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -101,46 +107,40 @@ You should see a list of long numbers. Each of these is an ID. In the Files pane
...
+File_Path<-"/fs/data3/tmccabe/AMF_US-NR1_BASE_HH_9-1.csv" #This is a path to where the observations are stored. 
+File_format<-PEcAn.DB::query.format.vars(bety = bety, format.id = 5000000002) #This matches the file with a premade "format" or a template that describes how the information in the file is organized
+site<-PEcAn.DB::query.site(site.id = 772,bety$con)  
 
-Change all bad values from -9999 to not a number (NA), and plot it to see if it make sense
-```{r,eval=FALSE}
-obs[obs==-9999]=NA
-plot(obs$NEE)
-```
+obs<-PEcAn.benchmark::load_data(data.path = File_Path, format= File_format, time.row = File_format$time.row,  site = site, start_year = 2004, end_year = 2004)
```
mccabete: I figure explaining that we're reading it in at the beginning is doing a lot of that work. Also, load_data is descriptive enough that I think an inline comment would be overkill.

documentation/tutorials/AnalyzeOutput/modelVSdata.Rmd
```diff
@@ -106,7 +100,7 @@ File_format<-PEcAn.DB::query.format.vars(bety = bety, format.id = 5000000002) #T
 
 site<-PEcAn.DB::query.site(site.id,bety$con) #This tells PEcAn where the data comes from
 
-observations<-PEcAn.benchmark::load_data(data.path = File_Path, format= File_format, time.row = File_format$time.row,  site = site) #This will throw an error that not all of the units can be converted. That's ok, as the units of the varibles of interest (NEE) are being converted. 
+observations<-PEcAn.benchmark::load_data(data.path = File_Path, format= File_format, time.row = File_format$time.row,  site = site, start_year = start.year, end_year = end.year) #This will throw an error that not all of the units can be converted. That's ok, as the units of the varibles of interest (NEE) are being converted.
```
mdietze: didn't a separate PR drop start_year and end_year as args to load_data?

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -25,7 +25,13 @@ RunDir <- "PEcAn_99000000002"
 
 Within that folder, you will find the *run/* folder that has a folder for every SIPNET run and the *out/* folder, that contains the output from each of these runs. The nice thing is that PEcAn has also produced “.RData” files of these model outputs that can be easily read into R by just clicking on them!
 
-So let’s try to compare Ameriflux NEE to SIPNET NEE across all these runs to make a plot of parameter vs. goodness-of-fit. We’ll start with root mean square error (**RMSE**), but then discuss some other tests, too.
+To do some automatic querying of data, let's open up a connection to the Bety database. 
+```{r, eval = FALSE}
+bety<-list(host = "psql-pecan.bu.edu", driver = "PostgreSQL", user = "bety", dbname = "bety", password = "bety")
```
mdietze: 1) For the tutorial, they'll be using their localhost not psql-pecan
2) Probably best to get these settings from the settings file, that makes the tutorial more machine independent.
3) As I've told lots of people lots of times, we really need to stop all these redundant calls to src_postgres all over the place. Please implement a wrapper function that just takes bety as an arg and which will ultimately replace the current db_open. Then please grep/replace all these redundant calls.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -99,48 +105,41 @@ runids
...
 
-Let’s try to read the output from a single run id, as you did in the earlier tutorial. **read.output** is a PEcAn function to read in variables from any PEcAn compliant output files, it requires a runid, the directory, start year, end year, and an optional list of variables. Make sure you have RunDir variable set to the run number and the year to the year you ran.
+Let’s try to read the output from a single run id as you did in the earlier tutorial. **read.output** is a PEcAn function to read in variables from any PEcAn compliant output files. It requires: a runid, the directory, start year, end year, and an optional list of variables. Make sure you have RunDir variable set to the run number and the year to the year you ran.
 ```{r,eval=FALSE}
-arun <- read.output(runids[1],paste("~/output",RunDir,"out",runids[1],sep="/"),2006,2006,c("time","NEE"))
-plot(arun$time,arun$NEE)
+arun <- read.output(runids[1], paste("~/output", RunDir, "out", runids[1], sep="/"), start.year= 2004, end.year= 2004,"NEE", dataframe = TRUE)
```
mdietze: If we're reading settings, start/end year shouldn't be hardcoded

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mdietze: Yes, the tutorials on the VM will be able to query their local BETY. As Anne notes, we really don't want this path hardcoded to your home directory since that won't be accessible to the general user. Furthermore, the path is a bit misleading -- the file should be automatically downloaded by met.process when it's grabbing the Ameriflux met.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mdietze: I'd add that these should be the exact same steps as the visualization tutorial, so you just need to reference that rather than explain much -- but you should make that reference more explicit since 'earlier tutorial' is vague and there's no guarantee that non Flux Course users will do the tutorials in the same order as the flux course students.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -99,48 +105,41 @@ runids
...
+obs<-PEcAn.benchmark::load_data(data.path = File_Path, format= File_format, time.row = File_format$time.row,  site = site, start_year = 2004, end_year = 2004) 
 
-To get something we can compare to, we also need to only retain good NEE. At the least, a u* filter is required. Let’s apply a simple one
-```{r,eval=FALSE}
-niwotnee <- obs$NEE
-niwotustar <- obs$UST
-niwotnee[niwotustar<0.2]=NA
+names(obs) #Notice that there are several time columns ("$year", "$datetime", "$posix") for most plotting, "$posix" will be the best time axis
+
+annualNEEobs<-aggregate(obs$NEE, by= list(month(obs$posix)), simplify=TRUE, FUN =mean, na.rm= TRUE)
```
mdietze: I'm not following (1) why you've dropped the u* filtering, which we still need and (2) why you've decided to aggregate the data to monthly -- that's new and going to completely change the meaning of the subsequent error statistics. This is particularly an odd choice given that you then decide to align by match_timestep, which makes this a real apples/oranges comparison.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -99,48 +105,41 @@ runids
...
-plot(obs$NEE)
-```
+obs<-PEcAn.benchmark::load_data(data.path = File_Path, format= File_format, time.row = File_format$time.row,  site = site, start_year = 2004, end_year = 2004) 
 
-To get something we can compare to, we also need to only retain good NEE. At the least, a u* filter is required. Let’s apply a simple one
-```{r,eval=FALSE}
-niwotnee <- obs$NEE
-niwotustar <- obs$UST
-niwotnee[niwotustar<0.2]=NA
+names(obs) #Notice that there are several time columns ("$year", "$datetime", "$posix") for most plotting, "$posix" will be the best time axis
```
mdietze: The note about 'several time columns' is more for your own benifit, not the user -- I'd just drop this line altogether since, as previously noted, everything up to here is just a repeat of the previous tutorial.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -99,48 +105,41 @@ runids
...
+
+annualNEEobs<-aggregate(obs$NEE, by= list(month(obs$posix)), simplify=TRUE, FUN =mean, na.rm= TRUE)
+
+plottable<-align_data(model.calc = arun, obvs.calc = obs, align_method = "match_timestep", var= "NEE")
+head(plottable)
 ```
 
 ####C. Finally, we can finally compare model to data
-In the earlier tutorial, you also compared NEE to the ensemble model run. Here we will do the same except for each sensitivity run. Recall that we had to convert units and make a scatter plot. 
+In the earlier tutorial, you also compared NEE to the ensemble model run. Here we will do the same except we include each sensitivity run. Recall that we made a scatter plot.
```
mdietze: "Recall that we made a scatter plot." seems a bit out of place here, I'd just drop it.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
```diff
@@ -99,48 +105,41 @@ runids
...
 abline(0,1,col="red")
 ```
 
 And remember the formula for RMSE:
 ```{r,eval=FALSE}
-sqrt(mean((niwotnee-modnee)^2,na.rm = TRUE))	
+sqrt(mean((plottable$NEE.o-plottable$NEE.m)^2,na.rm = TRUE))	
 ```
 
 *na.rm* makes sure we don’t include missing or screened values in either time series.
```
mdietze: Looking back over the tutorial, I really wonder if we can make getting to this point quicker and simpler -- so much of what comes before this line is a complete repeat of what we did in the previous tutorial and thus most of it could be done as briefly and simply as possible, without repeating redundant explanations, visualizations, stats, etc.

utils/R/do.conversions.R
```diff
@@ -1,9 +1,8 @@
-##' @export
+##' @export do.conversions
```
mdietze: do we really need to name do.conversions explicitly in the export?  If this was giving S3 issues then we should just change the name to do_conversions.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mccabete: It was in the original. I can change it with the settings file.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mccabete: Yeah, that just got there because it was testing for the section below it, where we demonstrate how to do one with aggregated data, which does have the correct alignment function.

mdietze: Ok, but if you wanted to show a user how to do a different data alignment that should be part of the visualization tutorial -- that might be a useful addition there (e.g. imagine the hypothetical where we were provided monthly data, here's how we'd handle that with align_data), but it's not what we want for this tutorial.  Also, if you do aggregate flux data up you really need to use the gap-filled data, otherwise the means will be biased.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mccabete: That was in the original tutorial. I changed it because I assumed we wanted it to match the first tutorial, and show off that users could theoretically download their own data to do comparisons.  I can change it to an accessible path.

documentation/tutorials/AnalyzeOutput/modelVSdata.Rmd
mccabete: Nope, that was align_data.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mccabete: @mdietze The hypothetical was part of the original tutorial. I can move it to the first tutorial if you'd prefer.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mccabete: Maybe. It's wordier than the first one, but I think that's the the two different author's style of tutorial.  I think in general it's good to err on the side of too much info in a tutorial than too little, because people can just (and will) skip around to what they find interesting/useful. 

Also the exploratory stuff in this tutorial  different than in the other.  If there's anything that is literally the same I can edit it down.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mdietze: The first tutorial should also match "~/output/dbfiles/Ameriflux_site_0-772/US-NR1.2006.nc" or something equivalent. If it still points to something on /fs/data3/ etc then it is wrong too and should be updated as well.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mdietze: I understand that people will skip around, but I do think if there's additional exploratory stuff it should be migrated to the previous demo. Likewise, if there's wordiness here that's useful it should be migrated to the previous demo. The previous demo is about learning to do these things for visualization, analysis, etc.

By contrast, this demo has a very different learning objective -- to understand the shape of how model error varies as parameters are varied so that users have a mental model for understanding what data assimilation is doing under the hood. Right now we spend ~150 lines getting to that point and then cover it in a small code block. What I think we want is a small code block that gets us to this point, references back to the previous demo for the details, and then the real focus on the error analysis.

utils/R/do.conversions.R
mccabete: This change is just reflective of a "." changed to a "," automatically and then a "." after Istem advised me to delete an extra @export statement.  @isfer would know more about if do.conversions needs to be exported.

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mccabete: I've responded to #'s 1 and 2, but I might need a little more context for #3. What is the db_open function?

documentation/tutorials/AnalyzeOutput/modelVSdata.Rmd
```diff
@@ -70,6 +71,8 @@ outdir<- "/fs/data2/output/PEcAn_99000000001/out/99000000001"
 start.year<-1997
 end.year<-2004
 site.id<-772
+File_Path<-"/fs/data3/tmccabe/AMF_US-NR1_BASE_HH_9-1.csv" #This is a path to where the observations are stored. If you have your own observations, change to cusom file path.
```
mdietze: Got that you're moving these things earlier, but as noted previously these need to be the default paths -- where PEcAn will put the data automatically and where users will see on the VMs --  not your personal directories

utils/R/do.conversions.R
mdietze: OK, we can deal with that later. It definitely does need the **@export** and at some point soon we should switch the name to do_conversions but that doesn't have to be this PR

documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd
mdietze: Not critical for this PR. We can discuss offline and roll into a future fix.

Issue #1028, opened 2016-10-03
pecan docker container


robkooper:
Create a Dockerfile that will create a simple Docker container. This container should be usable with a compose file with BETY and PostgreSQL containers.

Right now we will only have SIPNET in this container. Later we can look at other models, or get the models somehow in the container, other container..

yan130:
my understanding is to create a Dockerfile & docker-compose.yml ( link to PostgreSQL containers)  under contrib/browndog 

and using this docker container as a converter.

right

robkooper:
yes, containers in yaml
- pecan
- postgres
- bety

kapoorabhishek24:
Hi, I see this is listed as one the projects in GSOC'17. I wanted to know if this is still available or someone has already started working on it. I have good experience in Docker and wanted to know if I can pursue this.Let me know. Thanks

yan130:
Hi, I am not clear about GSOC'17. I have some code but not finished. I can push in my fork next week.

kapoorabhishek24:
Okay, I wanted to pursue this project, Let me know if its going to be considered under GSOC'17. I was already drafting the proposal.

istfer:
@kapoorabhishek24 thanks for your interest, could you get in touch with @robkooper and check out our [application page](http://pecanproject.github.io/gsoc_apply.html)

kapoorabhishek24:
@istfer Thanks for the heads-up. I already emailed Rob, just waiting for his reply. Meanwhile, I will proceed with the application and Proposal. Looking forward to be working on this project.

yogeshdarji99:
I am interested in this project for GSOC 17, I have over 3 years of experience with Docker and have dockerized more than 50 applications. I have worked for Verizon Cloud for a year where my core task to containerize and orchestrate all the application for Verizon cloud and deploy it on DataCenters using Mesos and Marathon. Currently, I am Computer Science Graduate Student at the University of Texas at Dallas and doing part-time internship at Verizon Cloud. I look forward to working on this project full-time during Summer where my skills and expertise be made use of.

dlebauer:
@robkooper since there seems to be so much interest perhaps worth flushing out the requirements. My understanding is that there will be a group of containers. A few questions:

* is there an existing BETYdb container? 
*  is there a draft of a pecan container?
* Should models be in separate containers?
* Where should all the data or whatever is blowing the VM go?

And to everyone else ... This would be awesome to have!

yan130:
[...]

PR #1485, opened 2017-06-23
PEcAn docker container #1028

Amanskywalker:
Current VM stack converted into the docker stack with the modularization of it into smaller independent parts i.e PEcAn-core, PEcAn-web, Postgresql & Bety. And they communicate with each other as needed.

## Description
Then whole PEcAn stack is modularize into smaller independent parts i.e PEcAn-core, PEcAn-web, Postgresql & Bety. Further models are independent. SIPNET is included in PEcAn-core as default model as PEcAn-core is base for PEcAn-web so its present in PEcAn-web too. 

## Motivation and Context

issue #1028 

## Types of changes

## Checklist:


docker/bin/my_init
```diff
@@ -0,0 +1,189 @@
+#! /bin/bash
```
dlebauer: extra space after '!'?

docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Dockerfile for the pecan web
```
dlebauer: @robkooper where should this go?

tonygardella:
Added review from @mdietze because you're paried with Aman this week for review.

dlebauer:
Really excited to have this PR. Thanks for bringing this all together.

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# Set script mod +x for preprocessors
+RUN chmod 750 /build/*.sh
+
+# Run the OS System setup script
+RUN /build/system_services.sh
+
+# run update machine to update machine
+RUN /build/update_machine.sh
+
+# run inatall packages to install required packages
```
mdietze: tiny typo on install vs inatall

docker/bin/my_init
```diff
@@ -0,0 +1,189 @@
...
+    do
+     run_command $script
+   done
+
+   echo "starting rc.local scritps"
+   run_command /etc/rc.local
+}
+
+
+# function to start corn jobs
```
mdietze: Is this supposed to be cron not corn?

docker/install_packages.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+. /build/install_pecan_preprocessor.sh
+
+echo "######################################################################"
+echo "SETTING UP REPOS"
+echo "######################################################################"
+case "$OS_VERSION" in
```
mdietze: General comment -- Thinking about this from the perspective of a user adding a model, or trying to port PEcAn to a new OS, I wonder if it would be easier if you had a set of different install_packages scripts, one for each OS.

Also, maybe this is just because things are a work in progress, but I assumed that having different dockers for different models would involve those individual pieces being responsible for what model packages needed to be installed (rather than a central script for installing all the packages for all models)

robkooper: This is copied from the install script in PEcAn. At some point I think we should see if we can minimize the number of dependencies and we can then also remove all the non ubuntu stuff, since the docker container will be only Ubuntu based.

mdietze:
Ok, unless @robkooper had any more comments, all this needs for a pull is a few typo fixes. Beyond that, I'll give fair warning that later PR of this size and magnitude (introducing whole new features) should be accompanied by simultaneous updates to the documentation and release notes. Since this didn't have it, I'm hoping that the next PR about Docker explains/documents PEcAn's use of Docker.

Amanskywalker:
Thanks @mdietze, @dlebauer and @robkooper for review and suggestions.
Working on Documentations and those typo fixes. 
From next pull request I will update documentations and log accordingly.

mdietze:
@Amanskywalker still waiting for those typo fixes to finish this PR

Issue #1496, opened 2017-06-29
Solidifying output standard


annethomas:
There are several issues already discussing this goal, but hopefully this will serve as touchpoint for moving forward on establishing a more comprehensive output standard for variable names/units.
Linking to issues #1487 #1442 #1415 @mdietze @mccabete @tonygardella 

## Description
We would like to align input/output variable names more closely throughout PEcAn. The MsTMIP table is the beginning of this but there are a lot of variables not included in this standard, and the relevant variables in BETY aren't always consistent in names and units. The proposal is to have a unified table in code and documentation, probably by renaming the mstmip_var.csv table to something more general, and to add useful variables using CF naming standards. This may involve renaming some BETY variables if possible, or else using their standard name field. 

Here is a link to a table of additional variables proposed so far, to review if you'd like to give input on the decision:
https://docs.google.com/spreadsheets/d/1ETasC8Nc0zGBjzo-wAY_Tyhd-IqSLKlvq854ErVypic/edit?usp=sharing

## Notes
1.	Many of the new variables reflect the hierarchy of carbon pools/fluxes being worked on by @mccabete. 
2.     In the table so far, I've focused on input variables for pool-based models and some variables immediately related (like the litter fluxes) but this is not yet exhaustive. @mdietze Let me know if you'd like Tess and/or me to work on fleshing out the whole hierarchy in this table right away.
3.    For litter pools and fluxes, I populated with existing CF names and then divided into the subpools (leaf, root etc) and added corresponding pools/fluxes. I'm not sure how many of them we actually will need to define.
4.	Layer-based soil variables: In Mstmip, TotSoilCarb has no layer-specific version so we want to add that (as an alternative to using CarbPools). Interestingly, SoilMoist only has a layer-specific version (unless SoilWet serves as the total, though it seems like a different measurement).
5.     See the google sheet comments by each variable for more specific questions or conflicts.

annethomas:
@mdietze Can you clarify the difference between the goal/outcome of the changes we've been talking about and @dlebauer 's suggestion of populating the standard names field with the table here from 2014-2016: https://docs.google.com/spreadsheets/d/1oEiDasdTslsm0VXFPWAUKS82BYbzXLyv2CYiicXMEys/edit#gid=0
? Are we just wanting to do a more thorough overhaul of names?

mdietze:
[...]

PR #1498, opened 2017-06-29
Adding paleon variables and renaming variables.

tonygardella:
Adding Paleon variables to the previously named `mstmip_vars.csv` table. If this is the table we decide to go with I'll need to change all old calls to the `mstmip_vars.csv` table.

## Description
@mccabete and @annethomas I saw your #1496 issue. Let me know if you're ok with this.

## Types of changes

## Checklist:


annethomas:
Is there a way to see the file formatted in a more readable way?

annethomas:
Never mind, it's readable enough. So once this is created and we make decisions about other variables to add it will be to this table, correct?

mccabete:
@annethomas  Here's a more readable PDF of the variables
https://files.gitter.im/PecanProject/pecan/Fpka/PaleonModelProtocol.Phase2.v4.1.pdf

mccabete:
Is it worth going through bety and expanding the bety_to_mistmip table to include these new variables. Also: should we rename that table to be bety_to_standardnames?

utils/data/standard_vars.csv
```diff
@@ -0,0 +1,76 @@
...
+68;7;15;Yes;precipf;;kg m-2 s-1;Precipitation;1;Driver;Yes;Yes;real;3;lon;lat;time;na;"The per unit area and time precipitation representing the sum of convective rainfall, stratiform rainfall, and snowfall"
+69;8;1;Yes;BA;;m2 ha-1;Basal area;1;Diversity;No;Yes;real;4;lon;lat;time;PFT;Basal area by PFT
+70;8;2;Yes;Dens;;1 ha-1;Stem Density;1;Diversity;No;Yes;real;4;lon;lat;time;PFT;Stem Density dy PFT
+71;8;3;Yes;Fcomp;;kgC kgC-1;Aboveground Biomass Fractional Composition;1;Diversity;No;Yes;real;4;lon;lat;time;PFT;Aboveground biomass Fractional composition of each PFT within each grid cell
+72;8;4;Yes;PFT;;(-);Plant Functional Type;1;Diversity;No;Yes;character;1;PFT;na;na;na;Name of each plant function type or species included in the model
+73;8;5;Yes;Estab;;1 ha-1;New Individuals;1;Diversity;No;Yes;real;4;lon;lat;time;PFT;New Individuals
+74;8;6;Yes;Mort;;1 ha-1;Mortality;1;Diversity;No;Yes;real;4;lon;lat;time;PFT;Individuals lost through death
+75;9;1;Yes;SoilDepth;;m;Soil Depth Layer;1;other;No;no;real;1;nsoil;na;na;na;Depth to the bottom of each model-defined soil layer
```
annethomas: On the other hand, we were talking about using "depth" as a dimension, whereas SoilDepth is a whole-profile variable with nsoil as a dimension.

mdietze:
In reply to @mccabete 

> Is it worth going through bety and expanding the bety_to_mistmip table to include these new 
> variables? 

Definitely! We're really trying to enforce that the output and bety not be in conflict in names and units, but as you know some BETY names & units just don't make sense.

> Also should we rename that table to be bety_to_standardnames?

Yep.

utils/data/standard_vars.csv
mdietze: Well, I don't recall the exact discussion that resulted in adding SoilDepth to the PalEON standard, but that's one that we can't just add here since it creates two variables (SoilDepth and z_bottom) that are identically defined. I suspect the PalEON team realized what we've been discussing: 

1. That z_top, z_bottom, and z_node are pretty redundant
2. That something with 'depth' in the name makes more sense to most people

However, as you note, 'soil_depth' in BETY is the total depth, so we don't want to use that. Also unfortunately, the BETY variable 'depth' has units of cm and will have existing records as a covariate (and possibly code) that would need conversion if we switch the units to m.

I think we should do the switch on depth's units, but as an interim stopgap I propose that we 
1. follow the PalEON change of dropping the z_* variables
2. that we retain the PalEON SoilDepth variable BUT
3. we instead have SoilDepth be a dimension 
4. we explicitly flag depth and SoilDepth as redundant
5. we don't forget to go back and update depth's units and then migrate all SoilDepth records/code/etc to be depth

If we're going to make this change it will require a quick grep of the model modules to update any models outputting z_* variables.  That said, the top gap will ultimately be a pain to fix so it might just be simpler to bite-the-bullet and fix `depth` right now while we're at it.

annethomas:
@tonygardella Will you change all the "na" text values to "NA" so R will recognize them as NA?

tonygardella:
@annethomas I can! We just need to remember that when we start using this table and change the `mstmipvar` function that we change [these lines](https://github.com/PecanProject/pecan/blob/ec9b025d3fc33026878c6da2c8414e20b7613833/utils/R/utils.R#L59-L68) to recognize the `NA` values.

annethomas:
@tonygardella Right, although isn't mstmipvar going to be deprecated and replaced? I'm not entirely clear on how.

annethomas:
One other thing, I think we could change it to comma-delimited as long as the description entries with commas in them are in quotes. That would make rendering easier but I'm not sure if it's helpful otherwise.

utils/data/standard_vars.csv
```diff
@@ -0,0 +1,76 @@
...
+76;9;2;Yes;Transp;;kg m-2 s-1;Total transpiration ;1;other;Yes;Yes;real;4;lon;lat;time;PFT;Total transpiration of each PFT within each grid cell
```
annethomas: @mdietze @tonygardella Is Transp redundant of Tveg, line 33?

annethomas: The only difference is that Transp has a pft dimension

mdietze: Seems like it's be simpler to just have a pft dimension be the default and then reduce this to one variable not two. Personally I'd favor Transp over Tveg since, as noted earlier today, having both Tveg (vegetation transpiration) and vegT (vegetation temperature) is a bit confusing.

mdietze:
Superceded bu #1516

PR #1503, opened 2017-07-05
Check package dependencies in `make document`

### LINKED ISSUE #1502: Add `make depends` option to build dependencies
## Description
Because it alters the NAMESPACE, `make document` should be run before `make install`. However, `make document` does not install dependencies, which can make it fail in certain situations if all dependencies aren't satisfied. Fortunately, there is the `devtools::install_dev_deps()` function that can install all dependencies for a package, so I propose having a `make depends` option on which `document`, `install`, and `check` depend.

This should take all of 2 minutes to implement. Just putting it here as a reminder.

ashiklom:
Resolves #1502.


tonygardella:
@mccabete You're with @ashiklom for code review this week.

mdietze:
@ashiklom this is failing on RNCEP package
```
Error : .onLoad failed in loadNamespace() for 'tcltk', details:
  call: dyn.load(file, DLLpath = DLLpath, ...)
  error: unable to load shared object '/home/travis/R-bin/lib/R/library/tcltk/libs/tcltk.so':
  libtcl8.5.so.0: cannot open shared object file: No such file or directory
ERROR: lazy loading failed for package ‘RNCEP’
```

ashiklom:
@mdietze Build is now passing.

PR #1510, opened 2017-07-07
Adds namespace to get.lv function. Restores functionality to model2 n…

mccabete:
## Description
contains namespace fixes and a condition for do_conversions that allows for ED's soil flag. 

## Types of changes

## Checklist:

### DIFF
--- a/utils/R/do_conversions.R
+++ b/utils/R/do_conversions.R
@@ -53,7 +53,7 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     }
     
     # soil extraction
-    if(input.tag == "soil"){
+    if(input.tag == "soil"&& is.null(input$path)){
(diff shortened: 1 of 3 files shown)


utils/R/do_conversions.R
```diff
@@ -53,7 +53,7 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     }
     
     # soil extraction
-    if(input.tag == "soil"){
+    if(input.tag == "soil"&& is.null(input$path)){
```
mdietze: I understand that, at the moment, there's a conflict between the soil tag in ED2 and the soil processing in do_conversions. However, in the long run, this is.null check needs to go -- the equivalent check in met.process caused many bugs when it was there and even more when it was finally removed. Developers need to not rely on this check to know whether the soil data being passed in is already in model-specific format, whether it's in it's raw format, or whether it's in the intermediate 'standard' format.

PR #1529, opened 2017-07-11
Always name files as "Geostreams" not "Clowder"

infotroph:
## Description
Was saving Geostreams downloads as "Clowder.<dates>.json", but workflow expects the met name to match what's declared in register.<metname>.xml. This resulted in a confusing web workflow error (`XML content does not seem to be XML: ''`), which turns out to really mean "I'm looking for a `register.Clowder.xml` and not finding one."

Fixed by always naming the files "Geostreams".

## Types of changes

## Checklist:


infotroph:
If testing this on a clean VM, note that Geostreams downloads will fail until you set up a ~/.pecan.clowder.xml (yes, that one is still named 'clowder'), but the workflow should now give an explicit authentication error instead of a cryptic XML error.

PR #1534, opened 2017-07-12
Default for met source setting in convert.inputs

tonygardella:
Without setting the met as Null, it was messing up run that did not require that argument so it needed a default setting.

## Types of changes

## Checklist:

### DIFF
--- a/utils/R/convert.input.R
+++ b/utils/R/convert.input.R
@@ -9,7 +9,7 @@
 convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, start_date, 
                           end_date, pkg, fcn, con = con, host, browndog, write = TRUE, 
                           format.vars, overwrite = FALSE, exact.dates = FALSE, 
-                          allow.conflicting.dates = TRUE, insert.new.file = FALSE, met = NULL,...) {
+                          allow.conflicting.dates = TRUE, insert.new.file = FALSE, pattern = NULL,...) {
   input.args <- list(...)
   
   logger.debug(paste("Convert.Inputs", fcn, input.id, host$name, outfolder, formatname, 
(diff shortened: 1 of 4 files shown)


utils/R/convert.input.R
```diff
@@ -9,7 +9,7 @@
 convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, start_date, 
                           end_date, pkg, fcn, con = con, host, browndog, write = TRUE, 
                           format.vars, overwrite = FALSE, exact.dates = FALSE, 
-                          allow.conflicting.dates = TRUE, insert.new.file = FALSE,...) {
+                          allow.conflicting.dates = TRUE, insert.new.file = FALSE, met = NULL,...) {
```
mdietze: I've got to ask about the choice of name here -- it appears that this arg is only used in a call to db.input.check where it's passed as ```pattern = met```. So why not call this arg ```pattern``` as well? If I use convert.input to process initial condition data, it's pretty unintuitive to have an argument called 'met', especially when the argument itself doesn't actually have anything to do with met, but is just the 'text to search for in the file name' when looking for existing files.

tonygardella: I can change it. I called it `met` because it comes from the met.process object created [here](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/met.process.R#L47) and then is passed onto the .download.raw.met.module [here](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/met.process.R#L160) and then to convert.inputs.

I'll rename the convert.input arg to `pattern` and change [this line](https://github.com/PecanProject/pecan/blob/release/1.5.0/modules/data.atmosphere/R/download.raw.met.module.R#L25) to `pattern = met`.

PR #1542, opened 2017-07-17
Dalec IC in write.configs

annethomas:
Couples write.configs.DALEC with the pool IC workflow in development. Checks for a path to an initial conditions file in the settings and writes any valid initial conditions to the command line parameters for DALEC. Also adds a default params file taken from the C code for the model (although it isn't necessary to write the defaults; rather, not writing anything leaves the defaults).

Note: requires manually editing the xml and some user discretion on inputs but has plenty of informative error messages.

## Types of changes

## Checklist:


models/dalec/R/write.configs.dalec.R
```diff
@@ -71,17 +71,40 @@ convert.samples.DALEC <- function(trait.samples) {
...
-  # cw0 initial pool of woody carbon (g/m2)
-  # cr0 initial pool of fine root carbon (g/m2)
-  # cl0 initial pool of litter carbon (g/m2)
-  # cs0 initial pool of soil organic matter and woody debris carbon (g/m2)
-  
   return(trait.samples)
 } # convert.samples.DALEC
 
+####partition_roots: function to split root_carbon_content into fine and coarse roots by rtsize dimension at the .002 m threshold
+partition_roots <- function(roots, rtsize){
```
mdietze: There doesn't seem to be anything DALEC-specific here. Is this something that could be moved to data.land and used by other models?

models/dalec/R/write.configs.dalec.R
```diff
@@ -114,6 +137,157 @@ write.config.DALEC <- function(defaults, trait.values, settings, run.id) {
     }
   }
   
+  ### INITIAL CONDITIONS
```
mdietze: There's a good bit here that looks similar to the logic that will be in other pool based models. I wonder if there are parts of this that can be extracted and generalized for reuse in other models

annethomas:
@mdietze Yeah I was thinking partition_roots was a good candidate for pulling out. I've been thinking about the rest of it and it will be a little trickier since the pools are defined slightly differently in different models but maybe there's some pre processing that could be done generically.

PR #1543, opened 2017-07-17
Updates & speed up to temporal downscaling code

crollinson:
Manually bringing in updates to the met ensemble / met downscaling code for PalEON.  This round greatly speeds up up the temporal downscaling process by reducing the amount of information being saved about each daily met model.  See commits here for more info: https://github.com/PalEON-Project/modeling_met_ensemble/commits/master 

## Description
Rather than save the entire output of the temporal downscaling models for each day, saving just the essential bits saves a lot of space and, as a result, time.  In order to do a low-memory replication of the model.frame function, all you need are the model coefficients, a callable formula for the model, a list of the factors/column names you need, and the contrasts.  See tdm_temporal_downscale_functions.R lines 123-158 & 189-225 to see where this has been implemented.  These portions now get called in lines 32-42 to have the same functionality as the original code (but so much faster!)

I also implemented the full set of code that would be necessary to add residual error if desired.

@jsimkins2 there are two spots I'm worried about: 1) was how we pull the betas to make Rbeta (what you have is quite different from my code); and 2) I could use a double check on precipitation.  Unless you sum precipitation to the day somewhere that I didn't see, I think there might be a unit error in your code and we would end up with a fraction of what it should be.

## Motivation and Context
Executing the temporal downscaling code was quite slow.  This is because the model .Rdata files were relatively large (~5 MB) so loading them hundreds of thousands of times was very slow.  Parsing these files down to just the essential information got them down to a couple kilobytes, so the code now runs ~200 x faster.

I don't think there's a specific issue about the speed issues, but they are improvements to #1356. 

## Types of changes

I don't *think* anything should break, but Pecan's complicated, so I wouldn't guarantee it.

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/tdm_generate_subdaily_models.R
+++ b/modules/data.atmosphere/R/tdm_generate_subdaily_models.R
@@ -18,7 +18,7 @@
 # -----------------------------------
 # Parameters
 # -----------------------------------
-##' @param outfolder - directory where models will be stored *** storage required varies by size of training dataset, but prepare for >100 GB
+##' @param outfolder - directory where models will be stored *** storage required varies by size of training dataset, but prepare for >10 GB
 ##' @param dat.train_file - train_data file
 ##' @param in.prefix 
 ##' @param n.beta - number of betas to save from linear regression model
--- a/modules/data.atmosphere/R/tdm_lm_ensemble_sims.R
+++ b/modules/data.atmosphere/R/tdm_lm_ensemble_sims.R
@@ -78,15 +78,16 @@ lm_ensemble_sims <- function(dat.mod, n.ens, path.model, lags.list = NULL,
     # Read in the first linear regression model
     first_model <- ncdf4::nc_open(paste0(path.model, "/", v, "/betas_", 
                                          v, "_1.nc"))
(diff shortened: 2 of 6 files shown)


mdietze:
@crollinson thanks a ton for the update!!

I think everything looks good to me, but one of your function declarations changed so you'll need to run ```make document``` and push up the updated modules/data.atmosphere/man/gen.subdaily.models.Rd to get Travis to stop complaining.

@jsimkins2 I'm going to wait for your 'Approve' to pull this since I want you to check the two things that Christy says she has concerns with.

jsimkins2:
@crollinson Thanks for the updates! As for your concerns: 1) this explains why my recent runs have been rather jumpy, nice find!; and 2) Precipitation looks good to me!

@mdietze these changes all look good to me!

crollinson:
Hold off on this pull request for the moment... Something is looking weird in my sub daily ensembles and I need to review it.  I'm going to use this opportunity to migrate everything to Pecan, so it might be a couple days... should I just close this pull request or keep it open?

mdietze:
It's fine to keep it open. Just ping us when you ready.

crollinson:
lots of commits about met are coming, so I'm going to go ahead close this PR so I don't cause a mess as I try to get everything sorted out

PR #1549, opened 2017-07-22
Allowing to compare Model Vs (uploaded) Data for shiny workflowPlot. Add Geom Smooth

shubhamagarwal92:
## Description

1. Adding geom_smooth parameterised by dynamic 'n' value for smoothing
2. Allowing to upload data. (Similar to ModelVsData tutorial)

Main changes in 
shiny/workflowPlot/server.R
shiny/workflowPlot/ui.R
screen shot 2017-07-22 at 12 14 29 pm

screen shot 2017-07-22 at 12 14 37 pm

## Motivation and Context

Allows to visualize the model outputs compared to the actual data observations. 
Based on [ModelVsData](https://github.com/PecanProject/pecan/blob/develop/documentation/tutorials/AnalyzeOutput/modelVSdata.Rmd) tutorial

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [] All new and existing tests passed.


shiny/workflowPlot/server.R
```diff
@@ -89,7 +94,33 @@ server <- shinyServer(function(input, output, session) {
...
+  # @params: bety,settings,File_path,File_format
+  loadObservationData <- function(bety,settings,File_path,File_format){
+    start.year<-as.numeric(lubridate::year(settings$run$start.date))
+    end.year<-as.numeric(lubridate::year(settings$run$end.date))
+    site.id<-settings$run$site$id
+    site<-PEcAn.DB::query.site(site.id,bety$con)
+    observations<-PEcAn.benchmark::load_data(data.path = File_path, format= File_format, time.row = File_format$time.row,  site = site, start_year = start.year, end_year = end.year) 
+    return(observations)
+  }
+  getFileFormat <- function(bety,format.id){
```
mdietze: This function seems to just rename PEcAn.DB::query.format.vars to getFileFormat, which seems pretty pointless. I'd recommend dropping. If you do think you have a compelling reason to keep, wouldn't it be simpler to just say ``` getFileFormat <- PEcAn.DB::query.format.vars``` ?

shiny/workflowPlot/server.R
```diff
@@ -122,11 +159,38 @@ server <- shinyServer(function(input, output, session) {
...
+      File_format <- getFileFormat(bety,input$formatID)
+      # Retaining the code for getting file format using inputRecordID
+      # File_format <- getFileFormat(bety,input$inputRecordID)
+      ids_DF <- parse_ids_from_input_runID(input$all_run_id)
+      settings <- getSettingsFromWorkflowId(bety,ids_DF$wID[1])
+      inFile <- input$fileUploaded
+      externalData <- loadObservationData(bety,settings,inFile$datapath,File_format)
+      # If variable found in the uploaded file 
+      if (input$variable_name %in% names(externalData)){
+        externalData <- externalData %>% dplyr::select(posix,dplyr::one_of(input$variable_name))
```
mdietze: This code is missing a call to PEcAn.benchmark::align_data. That call needs to be done BEFORE the plotting of the model output, because it may change the timestep/averaging, start/end date, and time zone of the model output. This will also facilitate the more general use of existing Metric functions in the plotting, allowing users to easily generate a set of 'standard' plots and scores.

shiny/workflowPlot/ui.R
```diff
@@ -14,10 +14,30 @@ ui <- shinyUI(fluidPage(
...
+                   selected="scatterPlot"),
+      # uiOutput("slider"),
+      sliderInput("smooth_n", "Value for smoothing:",
+                  min=0, max=100, value=80),
+      tags$hr(),
+      tags$hr(),
+      fileInput('fileUploaded', 'Choose file to upload data',
+                accept=c('text/csv', 
+                         'text/comma-separated-values,text/plain', 
+                         '.csv')),
```
mdietze: I don't see a reason to limit the file upload type here. If you did want to restrict the type, it should be based on the FORMAT record specifying the allowable mime-type for that specific format record. Even if you don't restrict the upload that way, it seems like a pretty essential check to perform before using the data.

ashiklom:
@mdietze Regarding the custom upload vs. inputs stuff, that was my mistake. I must have missed that point in discussions, and I assumed we were following the visualization tutorial, which loads a local file rather than a database file. @shubhamagarwal92 's original version correctly used the file input ID rather than format ID, and I asked him to change it. It shouldn't be too difficult to revert.

mdietze:
@ashiklom the visualization tutorial _does_ use a file that's in the database actually, it just doesn't ask the database for that location since it's a bit of a distraction to the point, which is to show the user how to load models and data at the command line so that the user can do things with them that are beyond what PEcAn can do automatically. That said, once visualization CAN display data automatically, we'll have to update that demo.

shiny/workflowPlot/server.R
shubhamagarwal92: This function has been implemented as a wrapper to retrieve the file format either on input id or on format id. For now, it is just based on file format. In future if we allow the user to retrieve based on input id as well.

shiny/workflowPlot/server.R
shubhamagarwal92: @ashiklom we had the same discussion. Should we finally resort to align_data?

mdietze:
(1) The "revert" to Input ID now seems to be internally inconsistent. EITHER the user is providing a file and Format id OR the user is providing an Input ID. The combo of file and Input doesn't make sense. 
(2) it would be extremely helpful to prepopuate a menu of available Inputs for the site (like we do for met) rather than expect the user to know what inputs are available. For simplicity this could start with a query based on == site.id, but later on we should shift that to a postGIS intersection betweeen the (buffered) site and available inputs, as that would allow us to pick up spatial data (e.g. Modus LAI)

shiny/workflowPlot/server.R
```diff
@@ -105,21 +106,35 @@ server <- shinyServer(function(input, output, session) {
...
     return(settings)
   }
+  observeEvent(input$load,{
+    # Retrieves all site ids from multiple seleted run ids when load button is pressed
+    req(input$all_run_id)
+    ids_DF <- parse_ids_from_input_runID(input$all_run_id)
+    site_id_list <- c()
+    for(row_num in 1:nrow(ids_DF)){
+      settings <- getSettingsFromWorkflowId(bety,ids_DF$wID[row_num])
+      site.id <- c(settings$run$site$id)
```
mdietze: I'm not 100% sure I follow this code, and I think this is fine for the first version (which we can restrict to single-site runs), but this does look like it would fail if one had run a multi-site workflow

shiny/workflowPlot/server.R
```diff
@@ -105,21 +106,35 @@ server <- shinyServer(function(input, output, session) {
...
+      site_id_list <- c(site_id_list,site.id)
```
mdietze: It's probably also good to add a ```unique`` to the final site_id_list. If one selects multiple runs to compare to data, most of the time those runs will be from the same site, and thus each would have the same site ID. But each unique site ID only needs to be listed once.

Oh, and will shiny select the first thing on the list by default -- I'm thinking that 99% of the time the site ID pull down will just have one item in it and it seems a shame to require the user to manually select it.

shubhamagarwal92: Shiny does that automatically.

PR #1554, opened 2017-07-26
4 Query & Download Utils for DataONE

LiamBurke24:
The purpose of these functions is to make downloading data stored in the DataONE federation more user friendly by allowing users to search and download by DOI, package id, or resource_map. 

## Types of changes

## Checklist:


mdietze:
To make Travis happy, you'll need to run ```make document`` and add the man files it generates to your commit.

Also, the build is throwing an error
```
  /home/travis/build/PecanProject/pecan/modules/data.land/R/write_ic.R:61:0: unexpected end of input
59: 
60: 
```
It's outside of your file, so I'm not sure why it's popping up now, but you might want to double check that there's no extra or missing { } or ( ) in your code.

bcow:
What @mdietze said in his last comment about man files is exactly what we did with `roxygenize` and then running build - so you should already have those commits, you just need to push them.

LiamBurke24:
Thanks @mdietze and @bcow! `roxygenise` is returning the following error 
```
> roxygen2::roxygenize("modules/data.land")
Error in parse(text = read_lines_enc(file, file_encoding = fileEncoding)) : 
  <text>:120:0: unexpected end of input
118: 
119: 
    ^
>
```
Have either of you seen this before? I'm lucking out on google for a fix...

mdietze:
Yes, that suggests it's missing something and getting to that point unexpectedly. That's why I said to check () {} etc

LiamBurke24:
Yes, I see that now. Apologies. That is now fixed. Although it looks like I made a few Roxygen errors now. Should be resolved soon. Thanks!

LiamBurke24:
So it looks like it builds alright now. Is it possible that the data.land package would not install properly if the make went well? The reason I ask is because none of the functions in data.land (and only in data.land) can be called and when I try to look at the man pages the following is displayed: 
```
Error in fetch(key) : 
lazy-load database'/fs/data3/lpburke/R/library/PEcAn.data.land/help/PEcAn.data.land.rdb' is corrupt
```
Thanks!

LiamBurke24:
Turns out that it was failing because I accidentally changed all instances of the word dataone from dataone to DataONE. :)

modules/data.land/R/DataONE_doi_download.R
```diff
@@ -0,0 +1,119 @@
...
+  mn <<- dataone::getMNode(cn, mnId)
+  bagitFile <<- dataone::getPackage(mn, id = resource_map, format = download_format)
+  bagitFile
+  
+
+  zip_contents <<- utils::unzip(bagitFile, files = NULL, list = TRUE, overwrite = TRUE, # list files in bagitFile
+                         junkpaths = FALSE, exdir = "downloads", unzip = "internal",
+                         setTimes = FALSE)
+  
+  utils::unzip(bagitFile, files = NULL, list = FALSE, overwrite = overwrite_directory, # Unzip the bagitFile and store in directory specified under exdir
```
mdietze: Not following why you're unzipping the file twice.

modules/data.land/R/DataONE_doi_download.R
```diff
@@ -0,0 +1,119 @@
...
+        junkpaths = FALSE, exdir = directory, unzip = "internal",
+        setTimes = FALSE)
+  return(zip_contents)
```
mdietze: Would be good to format the return in the format expected by convert.inputs. That will make it easier for existing code to handle duplicate checking and database record management

modules/data.land/R/DataONE_doi_download.R
```diff
@@ -0,0 +1,119 @@
...
+                               overwrite_directory = TRUE){
+  # Finding the mnId (query)
+  cn <- dataone::CNode(CNode) 
+  locations <- dataone::resolve(cn, pid = resource_map) 
+  mnId <<- locations$data[1,"nodeIdentifier"]
+  
+  # download the bagitFile 
+  mn <<- dataone::getMNode(cn, mnId)
+  bagitFile <<- dataone::getPackage(mn, id = resource_map, format = download_format)
+  bagitFile
```
mdietze: not sure what this line is for

modules/data.land/NAMESPACE
```diff
@@ -1,10 +1,12 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(format,identifier)
```
mdietze: We've switched to naming functions with underscores rather than periods to avoid S3method misinterpretation. I'd recommend updating all function names and rerunning make document

modules/data.land/R/DataONE_doi_download.R
bcow: This relates to what we talked about yesterday - where the ziped folder may contain multiple files but we need to figure out which of the files is the actual data and what format that data is actually in.

modules/data.land/NAMESPACE
LiamBurke24: Ok. Must have missed that.  Will do.

modules/data.land/R/DataONE_doi_download.R
LiamBurke24: Ah, good catch. That was from testing. To see examine the contents of the package. Not necessary anymore. Will remove.

modules/data.land/R/DataONE_doi_download.R
LiamBurke24: Again, this was for convenience during testing. I'll remove now

modules/data.land/R/DataONE_doi_download.R
LiamBurke24: Ok. Still trying to figure out what formats the data available comes in. Will update with progress today.

modules/data.land/R/DataONE_doi_download.R
LiamBurke24: Actually, the return that I have in there is pretty pointless. It just lists the files in the package. One can easily view these files by going to the directory where they are downloaded to. So, what would be a useful return?

mdietze: The return is most definitely not pointless -- having to write a second line of code to query the contents of a directory that you already just queried the contents of is what's pointless.

As noted before, the return should be in the format expected by convert.inputs, e.g. https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/met2CF.Ameriflux.R#L101-L108

LiamBurke24:
I changed the functions from get.resource.map to get_resource_map and roxygenized and rebuilt but the functions are still showing up as having the "." separator instead of the "_". Perhaps this is why Travis failed?

LiamBurke24:
Also, I have an issue open with dataone R to see if there is a better way to download data (a.k.a. last ditch effort).

mdietze:
The build say modules/data.land/man/get_resource_map.Rd is out of date.  FYI, you should look at the build log when Travis fails.

mdietze:
You also need to do a git pull upstream develop to bring your branch up-to-date with the mainline

PR #1560, opened 2017-07-31
delete GFDL Registration siteid and chnge naming of output directory

tonygardella:
GFDL had a siteid of "1160" which was "Earth". This made it so that GFDL files from different sites were being linked to the same input record. I've deleted it here in the registration file.

Also, within the GFDL download function there is a rewriting of the outfolder argument which seemed odd to me. Importantly, the pattern "GFDL" was not in the prefix of the redefining of the outfolder argument that happens in the function and it confuses db input checks. I've added it. 

That being said I do not understand why we are redefining the outfolder such that we are putting a directory inside of another valid directory. 
The input argument "outfolder" was : /fs/data1/pecan.data/dbfiles/GFDL_site_1-9/
and turned into this: /fs/data1/pecan.data/dbfiles/GFDL_site_1-9/_site_0-1160
With my changes it will look like this: /fs/data1/pecan.data/dbfiles/GFDL_site_1-9/GFDL_site_1-9

I would see a case where we could change outfolder name to include the GFDL model and/or scenario in the prefix. If that sounds good let me know and I can add it.

It could look like this: /fs/data1/pecan.data/dbfiles/GFDL_site_1-9/GFDL_CM3_rcp45_site_1-9

## Types of changes

## Checklist:


modules/data.atmosphere/R/download.GFDL.R
```diff
@@ -29,7 +27,11 @@ download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.
   model      <- paste0(model)
   scenario   <- paste0(scenario)
   ensemble_member <- paste0(ensemble_member)
-  outfolder  <- paste0(outfolder, "_site_", paste0(site_id%/%1e+09, "-", site_id%%1e+09))
+  
+  #Fix Outfolder to include model and scenario
+  folder_name <- paste0("GFDL_",model,"_",scenario)
```
mdietze: missed ensemble_member

PR #1562, opened 2017-08-01
PEcAn RTM cleanup

ashiklom:
* Add `Remotes:` to `DESCRIPTION` packages
* Clean up documentation with markdown
* Deprecate `invert.fast` functions
* Clean up helper functions


modules/rtm/R/invert.fast.R
```diff
@@ -33,6 +30,7 @@
 #' @return Matrix (ngibbs x (length(inits) + 1)) of MCMC samples 
 #'          of parameters.
 invert.fast <- function(modname, observed, inits, cons, pmu, psd, plog, minp, ngibbs) {
+  stop('This function is deprecated. Please use `invert.auto`.')
```
tonygardella: Should we move deprecated function to a separate directory or just do as you did?

mdietze: Once upon a time there was a great guide explaining the series of steps a function should go through as it's deprecated, including formal use of the .Deprecated and .Defunct functions, but I having trouble finding it at the moment.

ashiklom: Perfectly happy to just remove these functions, since they're immortalized in the git history.

Issue #1522, opened 2017-07-10
Using remote hosts causes errors; Cannot set hostname correctly 


mccabete:
This error was generated using the web interface and BU's computing cluster geo, but should cause problems for anyone attempting to use a remote host.  Example runs are [here](http://test-pecan.bu.edu/pecan_tmccabe/05-running.php?workflowid=1000007821), and[ here](http://test-pecan.bu.edu/pecan_tmccabe/05-running.php?workflowid=1000007819). This error is caused across machines and models. The issue seems to be that the hostname cannot be assigned an array. 

```
Error executing db query 'select id from machines where hostname = 
   'geo.bu.edu' and hostname = 'tmccabe' and hostname = 
   '/projectnb/dietzelab/pecan.data/output/tmccabe' and hostname = 'module 
   load udunits R/3.3.2' and hostname = "qsub -l h_rt=24:00:00 -pe omp 8 -q 
   'geo*' -V -v OMP_NUM_THREADS=8 -N @NAME@ -o @STDOUT@ -e @STDERR@ -S 
   /bin/bash" and hostname = 'Your job ([0-9]+) .*' and hostname = 'qstat 
   -j @JOBID@ || echo DONE' and hostname = 
   '/fs/data2/output//PEcAn_1000007821/tunnel/tunnel' and hostname = 
   '/projectnb/dietzelab/pecan.data/output/tmccabe/1000007821/run' and 
   hostname = 
   '/projectnb/dietzelab/pecan.data/output/tmccabe/1000007821/out' ;' 
   errorcode=0 message='ERROR: column "qsub -l h_rt=24:00:00 -pe omp 8 -q 
   'geo*' -V -v OMP_NUM_THREADS" does not exist LINE 1: ...me = 'module 
   load udunits R/3.3.2' and hostname = "qsub -l h...  ^ ' 
[1] "Killing tunnel with PID 23403"
```

## Context
This is an issue for complex or long runs, as it prevents the use of a computing cluster and slows down the local machine.

mccabete:
The Workaround for this problem is to  wrap the array you were assigning to your hostname in another array. In my config.php file it looks like this :
```
 "geo.bu.edu" => array(
                    array("qsub"   => "qsub -l h_rt=24:00:00 -pe omp 8 -q 'geo*' -V -v OMP_NUM_THREADS=8 -N @NAME@ -o @STDOUT@ -e @STDERR@ -S /bin/bash",
[...]

PR #1568, opened 2017-08-02
Bugfix: Passing host list, when should be hostname

ashiklom:
This was causing #1522. Based on preliminary tests, this is no longer hitting that particular bug. Can't speak to other bugs yet.


ashiklom:
My latest run failed, but with a CRUNCEP writing error rather than the hostname in #1522. So ED ran -- it just was unable to find its CRUNCEP file.

PR #1574, opened 2017-08-04
Fix broken error message in jagify

annethomas:
The jagify function in meta-analysis had an error message for unreasonable standard errors which had an uninitialized variable (probably from an old copy and paste from a for loop) which prevented the error message from showing. Replaced the broken part.

## Types of changes

## Checklist:

### DIFF
--- a/modules/meta.analysis/R/jagify.R
+++ b/modules/meta.analysis/R/jagify.R
@@ -36,9 +36,11 @@ jagify <- function(result) {
               select = c("stat", "n", "site_id", "trt_id", "mean", "citation_id", "greenhouse"))
   
   if (length(r$stat[!is.na(r$stat) & r$stat <= 0]) > 0) {
+    varswithbadstats <- unique(result$vname[which(r$stat <= 0)])
     citationswithbadstats <- unique(r$citation_id[which(r$stat <= 0)])
+    
     logger.warn("there are implausible values of SE: SE <= 0 \n",
-                "for", names(result)[i], 
+                "for", varswithbadstats, 
                 "result from citation", citationswithbadstats, "\n", 
                 "SE <=0 set to NA \n")
     r$stat[r$stat <= 0] <- NA


annethomas:
@mdietze When it's working jagify actually changes those SE 0 values to NA and things proceed normally (it's actually just a warning message). Do we still want to replace the values in the database?

mdietze:
Oh, if it's doing the change on the fly then updating the DB isn't a super high priority. That said, if you've already worked out the SQL command to do the update, you might as well post that so that @robkooper or @dlebauer can run it on betydb.org

annethomas:
Ok. The other thing I noticed is that the function will replace n values with 2 if they're originally 1 and the stat is not na, which kind of seems like a bug (it first changes all is.na(n) values to 1 and then changes all 1s to 2, but in our scenario when it's actually n=1 that's not desired).

dlebauer:
Setting n=2 is not a bug, from LeBauer et al 2013:

> Unlike the mean and variance parameters, missing values of n cannot be estimated and are conservatively set either to 2 (when existence of a variance estimate indicates n   2) or to 1 (if no variance estimate is given).

http://people.bu.edu/dietze/manuscripts/12-0137.pdf
 
there may be a better approach, but I'd suggest the default remain as described unless there is a compelling reason to change this.

annethomas:
@dlebauer The problem was that there were values in the table that were originally n = 1 but had stat = 0 instead of NA, so they were being changed to n = 2 (since the check after changing all n nas to 1 was  n =1 and stat is not NA). I don't know if that causes any difference in the meta-analysis. @mdietze and I were talking about changing those 0s to NA which would prevent that scenario.

mdietze:
I agree with @annethomas, setting n=2 when stat=0 is a bug.  

I'd suggest adding a line near the start of transform.nas that sets stat = NA if stat=0 since uncertainties can never be zero, they can just asymptotically approach 0. Entries with stat=0 are an exception to the logic that the existence of an error stat implies the sample size had to be >= 2 because stat=0 is occurring for cases where someone (in this case Sunny) entered individual raw measurements and set the error to 0 since there's no sampling error. In theory, there are still other sources of observation error that could/should be accounted for, but given that others may make this mistake, it makes sense to me to add this check even if we also fix these specific DB records.

dlebauer:
Sorry I missed the fact that the bug was when stat = 0.  in addition to fixing the code and records where stat=0, could also require stat to be null or >0

dlebauer:
I've changed stat = 0 to stat = NULL on betydb.org. Will ping others to do so on gitter. 

```sql
update traits set stat = NULL where stat = 0;
```

PR #1578, opened 2017-08-10
Dataone-download-util (work in progress -- waiting for dataone release)

LiamBurke24:
New and more robust tool for downloading data from dataone. Currently built with pre-release code, so it will break travis, but the release should be out any day now. 

## Description

This unction employs the new getDataPackage function from the dataone team. 

## Motivation and Context

The last version of this tool was flawed because the code from dataone r did not communicate well with dataone. They have fixed the problem (from the ground up) so this tool had to be rebuilt. 

## Types of changes

## Checklist:


modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,54 @@
...
+    }else{
+      formats[i] <- ".xml" # this is for the unknown type... Not sure if this is a universal fix... Please advise best practices here.
+    } 
+  }
+  
+  ### read data in the packets individually
+  filenames <- names(files) # list of all files because they are stored as headers by default
+  
+  # filepath & create new directory with timestamp -- same for all files
+  fp <- paste("~/downloads/data/", destdir, "_", Sys.time(), "/", sep = "") # what should the destination directory/ filepath be for pecan?
```
mdietze: Confused by the question about where the destDir should come from since it's being passed in -- why would you assume that directory is a subpath inside a directory that you've made up? Delete this entire line and just have the following line be dir.create(destdir)

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,54 @@
...
+  dir.create(fp)
+  
+  
+  for(i in 1:n){
+    pkgMember <- datapack::getMember(pkg, filenames[i])
+    data <- datapack::getData(pkgMember)
+    base::writeLines(rawToChar(data), paste(fp, "file_", i, formats[[i]], sep = "")) # file naming is an issue... How to proceed?
```
mdietze: 1) not sure I follow why file naming is an issue since you have ```filenames[i]``` two line above
2) not sure why you're using writeLines(rawToChar(data)) -- this implies that the package is loading the data directly into R rather than downloading the file.

web/drag.and.drop.upload.php
```diff
@@ -23,8 +23,7 @@
   exit;
 }
 
-# drag and drop window. I did not build this. Still looking for the liscence. Orignial can be found at: https://html5demos.com/dnd-upload/
-
+# drag and drop window.
```
mdietze: Not sure why the note about the origin of this code was removed

web/drag.and.drop.upload.php
```diff
@@ -81,6 +80,8 @@
       formdata: document.getElementById('formdata'),
       progress: document.getElementById('progress')
     },
+    
+    # This could be problematic: What generic mimetypes should we include?
```
mdietze: As commented elsewhere, don't restrict the allowable mimetypes

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,54 @@
...
+#' @param lazyLoad "A logical value. If TRUE, then only package member system metadata is downloaded and not data. The default is FALSE." -- dataone R 
+#' @param quiet "A 'logical'. If TRUE (the default) then informational messages will not be printed." -- dataone R
+#' 
+#' @author Liam P Burke, \email{lpburke@@bu.edu}
+#' @description Adapts the dataone::getDataPackage workflow to allow users to download data from the DataONE federation by simply entering the doi or associated package id 
+#'
+#' @export
+#'
+#' @examples doi_download(id = doi:10.6073/pasta/63ad7159306bc031520f09b2faefcf87, destdir = LTER)
+doi_download = function(id, destdir = "MyDataFile", CNode = "PROD", lazyLoad = FALSE, quiet = F){
```
mdietze: is doi_download the best name for this function, since it only works with DataOne (not any DOI) and it doesn't actually take DOI as an argument

modules/data.land/R/dataone_download.R
LiamBurke24: I assumed that we would want the user to be storing the file somewhere on the PEcAn server (hence the placeholder filepath that would direct the new directory that the user creates to the right place). If you would rather the user download the file to their local machine, then I can remove this.

web/drag.and.drop.upload.php
LiamBurke24: Oh no! Not sure how this file got into this PR... I made the changes you requested on the other PR... So sorry for the confusion

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,46 @@
...
+  mnId <- locations$data[1,"nodeIdentifier"] 
+  
+  ### begin D1 download process
+  d1c <- dataone::D1Client("PROD", mnId)
+  pkg <- dataone::getDataPackage(d1c, id = id, lazyLoad = lazyLoad, quiet = quiet, limit = "1MB") # what is the standard limit for pecan downloads?
+  files <- datapack::getValue(pkg, name="sysmeta@formatId")
+  n <- length(files) # number of files
+  
+  # fileath to /dbfiles 
+  fp <- "/fs/data1/pecan.data/dbfiles/"
```
mdietze: This needs to be passed in as an argument rather than hardwired. In previous commits it was (DestDir) -- not sure why you undid that

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,46 @@
...
+  
+  # make new directory within this directory
+  newdir <- paste(fp, "NewData_", username, sep = "")
+  system(paste("mkdir", newdir))
```
mdietze: replace with dir.create

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,46 @@
...
+  
+  # switch to new directory -- unsure if I should do this in R or in unix
+  # system(paste("cd", fp, sep = " ")) 
+  setwd(newdir)
```
mdietze: This is going to really mess with the user/workflow, since it's an unexpected side effect that will cause all subsequent input/output to be somewhere really different from where the user expects. The preferred solution is to figure out how to redirect wget to a specific directory (other than it's cwd). The alternative is to do a getwd before this setwd, save that location, and then setwd back to where it was originally before exiting.

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,46 @@
...
+  
+  for(i in 1:n){
+    rename <- paste("File", i, Sys.time(), sep = "_") # new file name
```
mdietze: I'd really prefer you retain the current filename and ditch the Sys.time. So until we can figure out how to get the 'pretty' file names, that'd be something like
```
paste(i, basename(names(files[i])), sep="_")
```

modules/data.land/R/dataone_download.R
```diff
@@ -0,0 +1,46 @@
...
+    system(paste("wget", "-O", rename, names(files)[i])) # download files with wget
+  }
+  system("ls") # checks that files were downloaded to
```
mdietze: replace with list.files.  More generally, avoid system calls for things R can do internally

modules/data.land/R/dataone_download.R
LiamBurke24: I thought this was the universal filepath to the /dbfiles folder

LiamBurke24: I can change it if need be. I just didnt know if the user would have to change the destination of the files :)

modules/data.land/R/dataone_download.R
LiamBurke24: noted

modules/data.land/R/dataone_download.R
LiamBurke24: got it

modules/data.land/R/dataone_download.R
LiamBurke24: Ok, let me see what I can do

modules/data.land/R/dataone_download.R
LiamBurke24: 👍

modules/data.land/R/dataone_download.R
LiamBurke24: Got it

modules/data.land/R/dataone_download.R
```diff
@@ -24,23 +25,16 @@ dataone_download = function(id, username, CNode = "PROD", lazyLoad = FALSE, quie
...
-  
+
   # make new directory within this directory
-  newdir <- paste(fp, "NewData_", username, sep = "")
-  system(paste("mkdir", newdir))
-  
-  # switch to new directory -- unsure if I should do this in R or in unix
-  # system(paste("cd", fp, sep = " ")) 
-  setwd(newdir)
+  newdir <- paste(filepath, "NewData_", username, sep = "")
```
mdietze: Because there's no guarantee that ```filepath``` will end in a / it's better to use file.path than paste

Also, the decision to use the username in the path both breaks with PEcAn conventions and will result in everything a single user ever downloads going into the same directory. Better to use a meaningful identifier, such as the id being passed in. "NewData" also isn't particularly indentifying

How about....
```
newdir <- file.path(filepath, paste0("DataOne_", id))
```

LiamBurke24: Ok, sounds good. Will there be any issues with the creation of the newdir if the id contains '/' in it? I ran into issues with this before... (e.g. id = doi:10.6073/pasta/63ad7159306bc031520f09b2faefcf87)

mdietze: Yeah, we want to avoid slashes in folder names. 
One option is to use a gsub to replace all / with _, -, or .
The other would be to basename the id to just the last bit.

LiamBurke24: Good idea!

LiamBurke24: 👍  Just committed the changes

PR #1580, opened 2017-08-10
Tree-ring code: update for ESA2017

mdietze:


tonygardella:
@ashiklom You're matched with Mike this week for code review.

modules/data.land/R/InventoryGrowthFusion.R
```diff
@@ -224,37 +265,91 @@ model{
...
+        
+      } ## end building beta
+      
+      ## append to process model formula
+      Pformula <- paste(Pformula,
+                        paste0(" + ",myBeta,"*",covX[1],"*",covX[2]))
+      
+      ## priors
+      Xt.priors <- paste0(Xt.priors,
+                          "    ",myBeta,"~dnorm(0,0.001)\n")
```
ashiklom: Probably fine for now, but eventually, might be good to let users pass custom priors for these terms?

modules/data.land/R/InventoryGrowthFusion.R
```diff
@@ -224,37 +265,91 @@ model{
...
     for(j in seq_along(t_vars)){
       tvar <- t_vars[j]
       
-      ## grab from the list of data matrices
-      dtmp <- time_data[[tvar]]
+      if(!(tvar %in% names(data))){
+        ## add cov variables to data object
+        data[[tvar]] <- time_data[[tvar]]
+      }
+      if(any(duplicated(names(data)))){PEcAn.utils::logger.error("duplicated variable at tvar",names(data))}
```
ashiklom: This block has happened above too, so it might be worth refactoring into a simple function? I mean the whole block, not just the error statement.

Also, should this be `logger.severe`, since `logger.error` doesn't actually throw an R error?

modules/data.land/R/InventoryGrowthFusion.R
```diff
@@ -294,22 +393,42 @@ model{
...
+    jags.out <- coda.samples(model = j.model, variable.names = vnames, n.iter = n.chunk)
+    ofile <- paste("IGF",model,k,"RData",sep=".")
+    print(ofile)
+    save(jags.out,file=ofile)
+    ## could add code here to check for convergence and break from loop early
+    D <- as.mcmc.list(lapply(jags.out,function(x){x[,'deviance']}))
+    gbr <- coda::gelman.diag(D)$psrf[1,1]
+    trend <- mean(sapply(D,function(x){coef(lm(x~seq_len(n.chunk)))[2]}))
+    if(gbr < 1.005 & abs(trend) < 0.5) break
+  }
```
ashiklom: 1.005 seems like a really strict convergence criterion. Isn't it more common to use something around 1.1?

Also, don't you want to do all of this on your cumulative samples? It seems like this could lead to a lot of extra iterations if there's high autocorrelation in the chains relative to `n.chunk`.

Also, it might be nice to print `k / ceiling(n.iter/n.chunk)` occasionally, since it's a nice progress indicator.

modules/data.land/R/InventoryGrowthFusion.R
```diff
@@ -114,27 +126,35 @@ model{
...
+  ########################################################################
+  ###
+  ###        FIXED EFFECTS
+  ###
+  ########################################################################
   if(FALSE){
     ## DEV TESTING FOR X, polynomial X, and X interactions
-    fixed <- "X + X^3 + X*bob + bob + dia"
+    fixed <- "X + X^3 + X*bob + bob + dia + X*Tmin[t]" ## faux model, just for testing jags code
   }
```
ashiklom: It would be nice to pull things like this out as formal unit tests. When I develop, I usually try to write tests immediately after, or even during or before, the rest of my code, and then use the `devtools::load_all(...)` to reload all of the code in a package and run the tests there.

modules/data.land/R/InventoryGrowthFusion.R
```diff
@@ -224,37 +265,91 @@ model{
...
+  ###
+  ###        TIME-VARYING
+  ###
+  ########################################################################
+  
+  if(FALSE){ # always false...just for development
     ## DEVEL TESTING FOR TIME VARYING
-    time_varying <- "TminJuly + PrecipDec + TminJuly*PrecipDec"
+    #time_varying <- "TminJuly + PrecipDec + TminJuly*PrecipDec"
+    time_varying <- "tmax_Jun + ppt_Dec + tmax_Jun*ppt_Dec"
```
ashiklom: See my earlier comment about unit tests.

modules/data.land/R/InventoryGrowthFusion.R
```diff
@@ -114,27 +126,35 @@ model{
...
 
+  ########################################################################
+  ###
+  ###        FIXED EFFECTS
+  ###
+  ########################################################################
   if(FALSE){
     ## DEV TESTING FOR X, polynomial X, and X interactions
-    fixed <- "X + X^3 + X*bob + bob + dia"
+    fixed <- "X + X^3 + X*bob + bob + dia + X*Tmin[t]" ## faux model, just for testing jags code
```
ashiklom: It would be nice to pull things like this out as formal unit tests. When I develop, I usually try to write tests immediately after, or even during or before, the rest of my code, and then use the `devtools::load_all(...)` to reload all of the code in a package and run the tests there.

modules/data.land/R/InventoryGrowthFusion.R
mdietze: This was implemented not as a generic convergence check, but as a deliberately strict condition that would err on the side of running to completion unless the convergence is very strong.

PR #1586, opened 2017-08-16
Adding LAI output to DALEC

bcow:
Adding LAI to DALEC output.
Units are taken from DALEC documentation.
The reason I'm accessing the output directory is to get the SLA from the config files to calculate the LAI.


models/dalec/R/model2netcdf.DALEC.R
```diff
@@ -70,6 +97,7 @@ model2netcdf.DALEC <- function(outdir, sitelat, sitelon, start_date, end_date) {
     output[[14]] <- output[[1]] + output[[2]]  # Total Respiration
     output[[15]] <- output[[9]] + output[[10]] + output[[11]]  ## TotLivBiom
     output[[16]] <- output[[12]] + output[[13]]  ## TotSoilCarb
+    output[[17]] <- sub.DALEC.output[, 15] * DALEC.configs[grep("SLA", DALEC.configs) + 1][[1]]
```
mdietze: Don't follow why there's a +1 , it looks like you're using the parameter _AFTER_ SLA rather than SLA

bcow: This is an inelegant way of reading in the data from the text file - it's actually just a vector that alternates between variable name and value. So if the ith entry is the character string "SLA" then the next entry is the value attributed to SLA.

bcow:
I did also find the SLA values in the `ensemble.samples.[ensemble_id].Rdata` file. Though those are just the parameter samples and they need to be run through `convert.samples.DALEC`.

Issue #849, opened 2016-05-23
Make new Sync Host ID more automatic


mdietze:
New users (e.g. on the VM) should be encouraged to establish a Sync Host ID in a more automated fashion so that we can maintain the goal that any run done in PEcAn should be assigned a globally unique ID. This shouldn't be a strict requirement (e.g. for the demo server or for AWS VMs using in teaching) but users should be informed of this option and the resetting of the DB to a new ID should be fairly automatic.  I'd propose a system where the first page on the web would pop up a menu that would allow a user to register their VM with our central server, and that doing so would update MACHINES on one of the core nodes (which would assign them the ID) and automatically apply this new ID to the database. Completing the registration, or checking a 'don't show this again' box, would make this registration pop-up go away in future uses. For the demo and the AWS VM we could pre-check the 'go away' button.  Similarly, we should figure out a way for users to migrate their Sync Host ID if they upgrade their VM or move from a VM to a server-side install.

mdietze:
Was working on updating the documentation on how to apply a new machine ID and was thinking that it would be very useful for the 'create' option under load.bety.sh to include the capacity to MIGRATE the records under the existing set-up to the newly created database.  I envision the following steps

1. add a SYNCSHIFT variable and -s flag to load.bety.sh
2. if this and CREATE are selected, at the start of the script, call dump.bety.sh to ensure there's a local copy of the existing records
3. update all record ID's to the new machine ID range
4. let CREATE do it's thing
5. at the end, instead of loading from DUMPURL, load from the local copy

Got started on this and then got stuck at part 3. @robkooper any suggestions for an easy way to do the renumbering via shell commands?  Or to do it with DB commands during drop or load? Would prefer to not write R code to do this if there's an elegant shell option.

Issue #758, opened 2016-03-21
Admin / Sync menu


mdietze:
Create a Admin webpage as part of the PEcAn web system that will users to control general PEcAn settings. The first item that should be on this page is an easy console for turning database sync on & off. Once it's ON it would also be nice to control when it is exported, when imports are done, and who is imported from. Similarly, it would be good for this console to give a clear warning if the local database is behind the current schema and point users to info on how to do the updates. It would also be good to have the ability to change the DB ID, though this should be made hard to change after it's first moved off of 99. More advanced would be some control over _what_ is imported and exported and a special set-up menu to guide users through their first time -- this might not only walk the user through the basics more easily, but might also perform tests (e.g. is cron installed and working) and point users to relevant documentation and bugs. 

Beyond the database, it would also be good to notify users when they're behind the current PEcAn stable version and point them to info on what they need to do to upgrade (e.g. is it just a simple pull and rebuild, or do configs need to be changed?)

PR #1589, opened 2017-08-18
[GSOC 2017] Script to perform synchronisation

Amanskywalker:
Script to perform the synchronisation between two system, anyone can act as the server or client. data of the server is synced to the client.

## Description

If the machine is there is new machine which called for the first time then the details of the host is added into the machine table and send back the want id which is being used by the sync cron job to sync the data.

Note : Need to make a cron entry for the cron job 

## Motivation and Context
#758, #849  

## Types of changes

## Checklist:


web/setups/serversyncscript.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Copyright (c) 2017 University of Illinois, NCSA.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the
+ * University of Illinois/NCSA Open Source License
+ * which accompanies this distribution, and is available at
+ * http://opensource.ncsa.illinois.edu/license.html
+ */
+
+// Server side scripts to handle the client resquests
```
mdietze: How do we ensure that this script ONLY runs on one server within the PEcAn network?
How do we prevent someone who already has a host id from accidentally requesting one a second time?
How do we prevent a malicious 3rd party from setting up a loop that asks for billions of host ids?

web/setups/submitissues.php
```diff
@@ -0,0 +1,166 @@
...
+
+        <div class="alert alert-info" role="alert">
+          You can submit another one
+        </div>
+
+<?php
+      }else {
+?>
+        <div class="alert alert-danger" role="alert">Failed to submit the Bug Report</div>
+        <div class="alert alert-warning" role="alert">Please check your github credentials</div>
```
mdietze: Does this mean one has to have github credentials to submit a bug report? If that's the case, what's the advantage of this page over the current practice of just linking to "New Issue" on github itself?

web/setups/synccorn.php
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Copyright (c) 2017 University of Illinois, NCSA.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the
+ * University of Illinois/NCSA Open Source License
+ * which accompanies this distribution, and is available at
+ * http://opensource.ncsa.illinois.edu/license.html
+ */
+
+// cornjob
```
mdietze: cron, not corn.

And I'm not following how this script sets up the cron job

web/setups/sync.php
```diff
@@ -0,0 +1,22 @@
...
+// sync
+
+$outcome = NULL;
+$status = NULL;
+
+$flagfile = fopen("syncflag.txt","r");
+$want_id = fgets ($flagfile);
+fclose($flagfile);
+
+exec('../../scripts/load.bety.sh', $outcome, $status);
```
mdietze: This doesn't seem to be sufficient to sync. 

First, whenever I run the sync scripts I have to put MYSITE=1 before the script call (or use flags to specify what my host id is)

Second, my sync script always starts with a dump.bety.sh, otherwise the sync is just one way

Third, currently when running load.bety.sh I have to loop over every remote site explicitly, setting REMOTESITE=<n>, for each.

FWIW, in more advanced applications, I think it would be great to allow the user to interactively choose which sites they want to sync with, but with "all" being the default. It would also be good to eventually give the user more control over what they push out during dump.bety.sh (in terms of tables and access levels)

web/setups/feedback.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Copyright (c) 2017 University of Illinois, NCSA.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the
+ * University of Illinois/NCSA Open Source License
+ * which accompanies this distribution, and is available at
+ * http://opensource.ncsa.illinois.edu/license.html
+ */
+
+// feed back page
```
Amanskywalker: Actually i was writing this page for the feedback form. So only blank page was added.

web/setups/serversyncscript.php
Amanskywalker: This script works on the POST request with the proper JSON which consist of hostname and client-secret string provided which is being generated by clientsyncscript.php. 
If the client joined for the first time i.e client-secret is blank, this script returns a generated md5 string which is stored in the client side config.php file. this will stop from accidentally requesting another id for second time.

web/setups/feedback.php
Amanskywalker: I removed that page form this pull request. will send in future request when done.

web/01-introduction.php
```diff
@@ -76,11 +76,13 @@ function nextStep() {
         </form>
 <?php whoami(); ?>
 <p>
+  <a href="setups/" target="_blank">Basic Setup</a>
```
Amanskywalker: added a link on generic PEcAn landing page too.

robkooper:
@Amanskywalker can you fix the conflicts.

Amanskywalker:
Yes, fixed the conflicts.

PR #1590, opened 2017-08-18
LAI reallocation in prepare.pools

annethomas:
Allow prepare.pools to estimate wood and reallocate pools to match total biomass. Change _ to . in function names.

## Description
This is a temporary fix for the initial conditions workflow (prepare.pools) scenario where estimating leaf carbon content from LAI and SLA leads to a discrepancy with total biomass (ie leaf is too high) if provided, which is especially a problem when trying to derive woody carbon as the complement of leaf and roots. This fix estimates wood from coarse roots or sets it to 0 (reasonable for a non-woody system but will have to be more thoroughly vetted in the next version) and reallocates the pools to add up to the total.
Also changed the names of two functions, prepare_pools and partition_roots, to match house style (before it gets harder to fix).

## Types of changes

## Checklist:


models/dalec/R/write.configs.dalec.R
```diff
@@ -121,7 +121,7 @@ write.config.DALEC <- function(defaults, trait.values, settings, run.id) {
       sla <- default.param[which(default.param$cmdFlag == "SLA"),"val"] * 1000 #convert SLA to m2/kgC from m2/gC (dalec default)
     }
     
-    IC.pools <- PEcAn.data.land::prepare_pools(IC.path, constants = list(sla = sla))
+    IC.pools <- PEcAn.data.land::prepare.pools(IC.path, constants = list(sla = sla))
```
annethomas: Oh, that was based on this documentation https://pecanproject.github.io/pecan-documentation/coding-style.html#dont-use-shortcuts . I was wondering why so many functions were still _

mdietze: It's the other way around. We _were_ originally using . but discovered that led to a lot of S3 conflicts and more recently have been switching to underscore, thus functions that use . are legacy and in need of updating.

@tonygardella could you fix the documentation ASAP so that the style guide is in line with our actual practices

annethomas: Ah. Working on the fix. 
The other thing that confused me was that do_conversions changed to do.conversions recently.

Issue #1520, opened 2017-07-08
reapply DB clean-up


mdietze:
repeat cleanup steps from PR #1091 to latest code. As @dlebauer notes, at this point that will be easier than resolving conflicts.

PR #1600, opened 2017-08-23
Full `PEcAn.DB` package cleanup

ashiklom:
## Description
I went through and manually fixed the following:
- Pass arguments by name, rather than just by position.
- Fix some lingering NAMESPACE issues.
- Move remaining packages from `Depends` to `Imports` in the `DESCRIPTION` field.
- Code formatting suggestions from `lintr::lint()` (the same program used for code "Diagnostics" by RStudio).
- Highlighted a few outstanding tasks as TODOs in the code

## Motivation and Context

Addresses #1520 

## Types of changes

## Checklist:


tonygardella:
@ashiklom Thanks for doing this! Part of package cleanup also involves running formatR::tidy_dir with args (recursive = TRUE, arrow=TRUE, width.cutoff = 500, indent = 2)

It should fix things aesthetically, but can also make some messes you'll have to clean manually. If you don't have time for that let me know. I can do that after I finish with fixing the model packages.

ashiklom:
I'm not at all convinced that using `formatR` is necessary or even desirable. It seems like the benefit from it is pretty minor, and it has a tendency to break things in unexpected ways. I'm still finding lots of places in other packages where it mangled multi-line comments or did other weird things.

tonygardella:
Yeah I agree it can make a mess of things. That's mostly why the original PR on this cleanup got nixed. One good thing it did was help with the readability of code when there are long lines of code/comments which I see you fixed anyways. It also helped with using arrows when assigning things, which seems to not be an issue here. Did you manually fix lines on your own or did you use some other function to help go through all the files?

ashiklom:
I went through each file and fixed them by hand. The default Rstudio code linter does a pretty good job of highlighting aesthetic issues like missing spaces between operators, and also warns about stuff like partial matching of argument names.

If anything, I would recommend running `lintr::lint(<filename>)` on each file to identify outstanding style issues and fix those by hand. RStudio conveniently lets you click on each issue raised by that command and takes you to to the corresponding line in the source code.

Issue #1452, opened 2017-06-09
SDA ensemble adjustment


mdietze:

mdietze:
# General problem

In an ensemble filter, we start with an initial (prior) _forecast_ that has a sample mean, $\mu_f$ and a sample covariance, $P_f$. After data assimilation that is updated via _analysis_to a posterior mean, $\mu_a$, and covariance, $P_a$. In a standard ensemble filter the next ensemble is simply sampled from the multivariate posterior (typically a multivariate Normal). Alternatively, the original ensemble members can instead be _nudged_ so that they have the proper posterior mean and covariance, which is called **Ensemble Adjustment**. If other factors are associated with each ensemble member (drivers, parameters, non-updated state variables, etc.), ensemble adjustment has been argued to be a more appropriate updating method.

# Diagonal case

Consider the simplifying case where $P_f$ and $P_a$ are diagonal matrices (no covariances) or the equivalent case of a univariate model. We can think of the adjustment occurring in two steps.

### Normalize

First, each of the original ensemble members, $X_i$ can be normalized to a Z-score (mean = 0, standard deviation = 1).

$$Z_i = {{X_i - \mu_f}\over{\sqrt{P_f}}}$$

### Update

Second, the standardized score can be rescaled to the new mean and covariance simply as

$$X_{i,a} = Z_i \sqrt{P_a} + \mu_a$$
Putting these two steps together gives an overall adjustment of

$$X_{i,a} = (X_i - \mu_f)\sqrt{{P_a}\over{P_f}}  + \mu_a$$
### Simulation

```{r}
n = 100
## forecast (prior)
mu_f_TRUE <- c(-3,5)
P_f_TRUE  <- diag(c(0.5,2))
## forecast ensemble
X_f <- mvtnorm::rmvnorm(n,mu_f_TRUE,P_f_TRUE)
plot(X_f)
mu_f <- apply(X_f,2,mean)
P_f  <- diag(apply(X_f,2,var))

## analysis
mu_a <- c(10,-3)
P_a  <- diag(c(3,1))

## analysis ensemble
X_a <- X_f*0
for(i in seq_len(n)){
  X_a[i,] <- (X_f[i,]-mu_f)*sqrt(diag(P_a)/diag(P_f))+mu_a
}

plot(X_a)
## check if ensemble mean is correct
cbind(mu_a,apply(X_a,2,mean))
## check if ensemble var is correct
cbind(diag(P_a),apply(X_a,2,var))
```

[...]

PR #1605, opened 2017-08-25
Ensemble Adjustment Filter in sda.enkf

araiho:
In this pull request, the biggest thing is that I added the math for the ensemble adjustment filter. Other things that I did are smaller updates to LINKAGES and SIPNET packages. 

## Description
Units Changes:
- Changing units to standard in write.restart and read.restart
- adding udunits to sample.ic.sipnet
- fixing split_inputs to get julian day for end.day
- adding posixt format line per @istfer suggestion to pda.get.model.output 
- changing workflow.treering to use udunits

General Updates:
- Adding tags to TreeRingSDA.Rmd 
- Making linkages load from my home directory in case of updates per @ashiklom suggestion
- Adding abovegroundwood to model2netcdf.LINKAGES
- Fixing param names in write.configs.LINKAGES
- Changing names of species to take out white space

SDA Changes:
- Fixing tobit2space.model to only estimate means for variables that need data imputation.
- Added ensemble adjustment math and now analysis update comes from weighted ensembles
- Adding stem density constraint to read.restart.LINKAGES
- Changing colors on bias diagnostics to not match time series

## Motivation and Context

Mainly addressing #1452. 
Still working on #1448. I may need some more advice on how to fully resolve this issue. 
I was able to run the tree ring demo, so I think #1451 is also resolved.

## Types of changes

## Checklist:

- [x] I have added *some* tests to cover my changes.


tonygardella:
@mccabete You're matched this week with Ann for code review.

models/linkages/R/model2netcdf.LINKAGES.R
```diff
@@ -39,9 +39,9 @@ model2netcdf.LINKAGES <- function(outdir, sitelat, sitelon, start_date = NULL, e
   
   ### Loop over years in linkages output to create separate netCDF outputs
   for (y in seq_along(years)) {
-    if (file.exists(file.path(outdir, paste(years[y], "nc", sep = ".")))) {
-      next
-    }
+    # if (file.exists(file.path(outdir, paste(years[y], "nc", sep = ".")))) {
+    #   next
+    # }
```
mccabete: Anne is this also supposed to be turned on for debugging or should it be deleted? If there's a third option would you mind just clarifying the situation in which you would un-comment it?

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -36,9 +36,9 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
   
   ### Loop over years in SIPNET output to create separate netCDF outputs
   for (y in years) {
-    if (file.exists(file.path(outdir, paste(y, "nc", sep = ".")))) {
-      next
-    }
+    # if (file.exists(file.path(outdir, paste(y, "nc", sep = ".")))) {
```
mccabete: Again, it's not immediately clear to me why this is commented out.

modules/assim.sequential/inst/paleon_sda.R
```diff
@@ -1,26 +1,49 @@
 
 library(PEcAn.all)
 library(PEcAn.SIPNET)
+library(PEcAn.LINKAGES)
```
mccabete: I thought we were deprecating library calls? Is there any reason not to use namespace calls?

documentation/tutorials/StateAssimilation/TreeRingSDA.Rmd
```diff
@@ -40,48 +40,66 @@ Perform a site-level SIPNET run using the following settings
...
-  <process.variance>FALSE</process.variance>
-  <sample.parameters>TRUE</sample.parameters>
+   <state.data.assimilation>
+   <n.ensemble>35</n.ensemble>
+   <process.variance>FALSE</process.variance>
+   <sample.parameters>TRUE</sample.parameters>
+   <data>
+    <format_id>
+    <format_id>1000000040</format_id>
+    </format_id>
```
mdietze: 1) If we have input.id, we shouldn't need the user to specify format_id as well.
2) we definitely shouldn't need to nest format_id within format_id

documentation/tutorials/StateAssimilation/TreeRingSDA.Rmd
```diff
@@ -40,48 +40,66 @@ Perform a site-level SIPNET run using the following settings
...
+    <input.id>
+          <input.id>1000013298</input.id>
+    </input.id>
```
mdietze: Likewise, definitely shouldn't need to nest input.id within input.id. Also, should make sure we're consistent in naming tags between here an other parts of the code where input id's are specified.

documentation/tutorials/StateAssimilation/TreeRingSDA.Rmd
```diff
@@ -40,48 +40,66 @@ Perform a site-level SIPNET run using the following settings
...
-  </state.variables>
-  <spin.up>
-    <start.date>1979/01/01</start.date>
-    <end.date>1979/12/31</end.date>
-  </spin.up>
-  <forecast.time.step>1</forecast.time.step>
+   <variable>
+   <variable.name>NPP</variable.name>
+   <unit>MgC/ha/yr</unit>
+        <min_value>-9999</min_value>
```
mdietze: minor thing, but all these ```<variable.name>``` and ```<unit>``` specifications should be indented since they're nested within ```<variable>```

models/linkages/R/model2netcdf.LINKAGES.R
mdietze: 👍 second what @mccabete said

models/linkages/R/write_restart.LINKAGES.R
```diff
@@ -212,16 +214,19 @@ write_restart.LINKAGES <- function(outdir, runid, start.time, stop.time, setting
       new.ntrees[s] = sample(size = 1, x = 50:150)
     } 
   }
+  
+  if(sum(new.ntrees) > 198) new.ntrees <- round((new.ntrees / sum(new.ntrees)) * runif(1,160,195))
+
```
mdietze: This seems like a hack -- could you at least provide a comment explaining the rational for this and whether it's intended to be temporary or not

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -309,20 +314,14 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
...
+    
+    y.censored[1:N] ~ dmnorm(muf[1:N], prec = pf[1:N,1:N])
+    
+    for(i in 1:N){
+      y.ind[i] ~ dconstraint(y.censored[i] > 0)
     }
-
-    #Priors
-    pf[1:N,1:N]  ~ dwish(aq[1:N,1:N],bq)
-    muf[1:N]  ~ dmnorm(mu.prior[1:N],cov.prior[1:N,1:N])
```
mdietze: Not following how this works now that the priors on muf and pf have been removed

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -332,11 +331,15 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
...
+  purple       <- col2rgb("purple")
+  alphapurple <- rgb(purple[1], purple[2], purple[3], 75, max = 255)
+  brown       <- col2rgb("brown")
+  alphabrown <- rgb(brown[1], brown[2], brown[3], 75, max = 255)
   
   ###-------------------------------------------------------------------###
   ### loop over time                                                    ###
   ###-------------------------------------------------------------------###  
-  for (t in seq_len(nt)) {#
+  for(t in seq_len(nt)) {#seq_len(nt)
```
mdietze: looks like this needs clean up

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -482,83 +481,73 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
...
+          constants.tobit2space = list(N = length(mu.f),
+                                       muf = mu.f,
+                                       pf = solve(Pf))
+          data.tobit2space = list(y.ind = x.ind,
+                                  y.censored = x.censored)
           #set.seed(0)
           #ptm <- proc.time()
           tobit2space_pred <- nimbleModel(tobit2space.model, data = data.tobit2space,
-                                          constants = constants.tobit2space, inits = inits.tobit2space)
+                                          constants = constants.tobit2space)
```
mdietze: would nimble convergence be quicker if we retained the specification of inits? I'd guess that sample mean and cov would be a decent starting point

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -739,8 +735,48 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
       enkf.params[[t]] <- list(mu.f = mu.f, Pf = Pf, mu.a = mu.a, Pa = Pa)
     }
     
-    ## update state matrix
-    analysis <- as.data.frame(rmvnorm(as.numeric(nens), mu.a, Pa, method = "svd"))
+    ###-------------------------------------------------------------------###
```
mdietze: Glad to have ensemble adjustment added, and I agree it should be the default, but I wonder if we want to retain a flag and an ```if``` to toggle between the two options rather than simply deleting the original EnKF?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -739,8 +735,48 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
...
+    for(i in seq_len(nens)){
+      Z[i,] <- 1/sqrt(L_f) * t(V_f)%*%(X[i,]-mu.f)
+    }
+    Z[is.na(Z)]<-0
+    
+    ## analysis
+    #mu_a <- c(10,-3)
+    #D  <- sqrt(diag(c(3,1)))
+    #R  <- matrix(c(1,-0.75,-0.75,1),2,2)
+    #P_a  <- D%*%R%*%D
```
mdietze: You can delete these comments, which were just for testing

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -817,6 +870,7 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL) {
         # analysis
         ciEnvelope(as.Date(obs.times[t1:t]), XaCI[, 1], XaCI[, 2], col = alphapink)
         lines(as.Date(obs.times[t1:t]), Xa, col = "black", lty = 2, lwd = 2)
+        #legend('topright',c('Forecast','Data','Analysis'),col=c(alphablue,alphagreen,alphapink),lty=1,lwd=5)
```
mdietze: reason the legend is commented out?

modules/assim.sequential/inst/paleon_sda.R
mdietze: While I don't disagree with @mccabete, it's worth noting that this is a script in the inst folder, not a function definition, so I'm not as worried about library calls here

modules/assim.sequential/inst/paleon_sda.R
```diff
@@ -1,26 +1,49 @@
...
 library(PEcAn.assim.sequential)
+library(nimble)
+library(lubridate)
+
+ciEnvelope <- function(x,ylo,yhi,...){
+  polygon(cbind(c(x, rev(x), x[1]), c(ylo, rev(yhi),
+                                      ylo[1])), border = NA,...) 
+}
```
mdietze: I don't think you need to redefine this since it's exported by the visualization module, which should load when you load PEcAn.all (if not, that's a bug)

models/linkages/R/model2netcdf.LINKAGES.R
araiho: I think it was originally in here so that nc files wouldn't be remade if they already existed. But, that was a problem sometimes if I wanted to rerun an sda time step. So, I guess I'm not sure what to do. I thought about adding an overwrite flag, but that seemed like a pecan design thing more than a choice I should make.

models/linkages/R/write_restart.LINKAGES.R
araiho: So, this is to enforce stem density in linkages. I could estimate it with sda maybe? or have it as a parameter like carrying capacity?

mdietze: In the short term I wouldn't add stem density to the SDA since in the current example we don't have data to estimate it, plus you might still need some constraint if the SDA itself pushes the density too high. I'll leave the long term solution to you (e.g. carrying capacity or other option) but in the short term I think I'd be happy if you just added some comments here explaining.

modules/assim.sequential/R/sda.enkf.R
araiho: The purpose of this short model is to impute the means for the state variables that are zero inflated, so that they are in tobit space in the analysis model. I don't think it was correct to estimate them before.

modules/assim.sequential/R/sda.enkf.R
araiho: according to the nimble manual 'Nodes following a dinterval (or dconstraint) distribution should normally be set as data with known values. Otherwise, the node may be simulated during initialization in some algorithms (e.g.,MCMC) and thereby establish a permanent, perhaps unintended, constraint.' So, I took out the inits to avoid that problem.

modules/assim.sequential/R/sda.enkf.R
araiho: 👍

modules/assim.sequential/R/sda.enkf.R
araiho: it was covering the whole figure. I'll google a way to fix that.

modules/assim.sequential/R/sda.enkf.R
araiho: I added the flag but for some reason it's not changing here

modules/assim.sequential/R/sda.enkf.R
mdietze: I understand the purpose of the model, what I don't understand is how a model that fits ```muf``` and ```pf``` can run without priors on those variables.

modules/assim.sequential/R/sda.enkf.R
mdietze: Did you forget to make a commit or to push it? I don't see the flag

modules/assim.sequential/R/sda.enkf.R
araiho: They aren't fit. when y.censored = 0, y.censored is imputed with mean muf and covariance pf

modules/assim.sequential/R/sda.enkf.R
araiho: yeah it's in one of the commits below https://github.com/PecanProject/pecan/pull/1605/commits/58112a0e4285cd77ee4f98438cbb5ce927f7c045
But maybe there is more that I need to change?

modules/assim.sequential/R/sda.enkf.R
mdietze: Ok, but then where does muf and pf come from? They aren't the sample mean and cov because those are zero-inflated and will give you the wrong value. I thought the whole point of this model was was to estimate those variables so we can pass them on to main filter model?

araiho: I'm getting confused because if we estimate them we change the sample means and covariances for the state variables that are not zero inflated. At least that is what was happening before. I think we only want to impute the state variables that are all zeros and in those cases the mean is zero and we were picking the minimum covariance to be the stand in covariance. But, you're saying we want to estimate the whole muf and pf before passing it into the filter model and assume that all of the state variables are zero inflated?

mdietze: * I've always assumed that the thing we were passing into the filter model was muf and pf as a prior on X, not passing in the actual ensemble members. If you were doing the latter I don't see why we need the extra step of this model, since we could just do it all as one model.
* We cannot impute the state variables of the zeros without also estimating their mean and variance! Their sample mean and variance is still wrong. For example, if a state is >50% of it's observations at zero, then the true mean has to be negative, but the sample mean can never ever be negative. This is an extreme case, but the bias in mu will still be there in all cases and will just accumulate to be larger and larger the more zeros in the data.
* I agree that it's odd for non-zero bound state variables to have their means and variances shift away from the sample mean and variance based on their covariance with zero-bound state variables. But it seems like we should figure out why that's happening, and whether it's actually the correct answer or not. To me, this code doesn't feel like the correct solution.
* State variables where all or almost all ensemble members are zero definitely seems like a tough case. But in those cases the mean definitely isn't zero -- it's some unknown negative number. Problem is there's almost no information in the ensemble to figure out what that negative number SHOULD be. Seems like whatever the model comes up with would be pretty dependent upon the prior (which presumably would make -1 much more likely than -1000). Would LIKE to deal with these cases explicitly, but I could see needing a kluge/hack in place that excludes such terms from the state estimation and instead inserts a mean that's set to have high uncertainty, zero covariances, and a mean that results in negligible probability of positive values (the exact value might scale with ensemble size). We'd want this to be set so that the data can easily nudge the model, but not be so strong as to allow it to follow noise in the data (after all, the model did reach a consensus that the species should be extinct)

modules/assim.sequential/R/sda.enkf.R
mdietze: Hmm -- very odd. Could you just try re-adding the changes? I wonder if in some combo of merging you accidentally overwrote the change?

PR #1614, opened 2017-08-30
Changes to Fates build script

tonygardella:
Update Fates build script 

## Types of changes

## Checklist:

### DIFF
--- a/models/fates/inst/create_1x1_ref_case.sh
+++ b/models/fates/inst/create_1x1_ref_case.sh
@@ -59,8 +59,8 @@ echo "*** Modifying xmls  ***"
 ./xmlchange -file env_mach_pes.xml -id MAX_TASKS_PER_NODE -val 1
 ./xmlchange -file env_mach_pes.xml -id TOTALPES -val 1
 # Modifying : env_build.xml
-#Updated variable of CESMSCRATCHROOT?#./xmlchange -file env_build.xml -id CIME_OUTPUT_ROOT -val ${CASEROOT}
-./xmlchange -file env_build.xml -id CESMSCRATCHROOT -val ${CASEROOT}
+./xmlchange -file env_build.xml -id CIME_OUTPUT_ROOT -val ${CASEROOT}
+#Outdated Var ./xmlchange -file env_build.xml -id CESMSCRATCHROOT -val ${CASEROOT}
 ./xmlchange -file env_build.xml -id GMAKE -val make
 #./xmlchange -file env_build.xml -id MPILIB -val openmpi
 #./xmlchange -file env_build.xml -id OS -val Linux
--- a/models/fates/inst/create_1x1_ref_case.sh
+++ b/models/fates/inst/create_1x1_ref_case.sh
@@ -60,7 +60,6 @@ echo "*** Modifying xmls  ***"
 ./xmlchange -file env_mach_pes.xml -id TOTALPES -val 1
 # Modifying : env_build.xml
 ./xmlchange -file env_build.xml -id CIME_OUTPUT_ROOT -val ${CASEROOT}
-#Outdated Var ./xmlchange -file env_build.xml -id CESMSCRATCHROOT -val ${CASEROOT}
 ./xmlchange -file env_build.xml -id GMAKE -val make
 #./xmlchange -file env_build.xml -id MPILIB -val openmpi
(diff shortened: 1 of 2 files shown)


models/fates/inst/create_1x1_ref_case.sh
```diff
@@ -18,15 +18,20 @@
 # site_lat_c =   9.1543, 5.07389,  -2.60909722,  1.4368,   4.1865,  15.6324
 # site_lon_c = 280.1539, 8.85472, 299.7907,     28.5826, 114.017,   99.217
 #=============================================================================================
-#export NETCDF_HOME=/usr/local/  # is this nescessary?
+export NETCDF_HOME=/usr/local/  
+export NETCDF_PATH=${NETCDF_HOME}
```
serbinsh: @tonygardella @mdietze I think we need to be clearer with this build script that it is a template and is not generic enough to download and run everywhere.  For example ,I don't need to export my netCDF paths....but it was clear from the other day that for your current setup you do.  So this should be something listed as "optional"

models/fates/inst/create_1x1_ref_case.sh
```diff
@@ -52,7 +57,7 @@ echo "*** Modifying xmls  ***"
 ./xmlchange -file env_mach_pes.xml -id MAX_TASKS_PER_NODE -val 1
 ./xmlchange -file env_mach_pes.xml -id TOTALPES -val 1
 # Modifying : env_build.xml
-./xmlchange -file env_build.xml -id CIME_OUTPUT_ROOT -val ${CASEROOT}
+./xmlchange -file env_build.xml -id CESMSCRATCHROOT -val ${CASEROOT}
```
serbinsh: Im still uncertain about this because it is my understanding that CESMSCRATCHROOT is an old variable and as of a few re-factors ago it is now CIME_OUTPUT_ROOT.  This is what I use, not scratch root and it works for me.  So again this is unclear. I pointed this out to Ryan and he didnt really know...he was just having you try various things.  This should be clarified.

models/fates/inst/create_1x1_ref_case.sh
```diff
@@ -102,7 +107,7 @@ echo "*** Running case.setup ***"
 
 cat >> user_nl_clm << \EOF
 hist_empty_htapes = .true.
-hist_fincl1='EFLX_LH_TOT','TSOI_10CM','QVEGT','NEP','GPP','AR','ED_bleaf','ED_biomass','NPP_column’,'NPP','MAINT_RESP','GROWTH_RESP'
+hist_fincl1='EFLX_LH_TOT','TSOI_10CM','QVEGT','NEP','GPP','AR','ED_bleaf','ED_biomass','NPP_column','NPP','MAINT_RESP','GROWTH_RESP'
```
serbinsh: Ryan told me that NPP is the output to use, and that NPP_column wasn't useful but I don't remember why.  NEP can stil cause PIO write errors and I would recommend we don't set that until we can confirm that NEP wont crash runs

models/fates/inst/create_1x1_ref_case.sh
serbinsh: For example, @tonygardella look at the config_machines.  In there the XML tag is CIME_OUTPUT_ROOT and used to be CESMSCRATCHROOT that is why I think that var has changed, but again Ryan wasn't clear about that/

serbinsh: Yeah, take a look in here: https://github.com/NGEET/fates-clm/blob/master/cime/config/cesm/machines/config_machines.xml

tonygardella: That link doesn't work for me. Is it a private repo? 

That being said, a grep for "CESMSCRATCHROOT" only can find it in "ngeet-config_machines.xml". I'll switch it back to "CIME_OUTPUT_ROOT".

serbinsh: Yeah its private, I thought you had access.  But yeah try with CIME_OUTPUT_ROOT in your create run script and see if you can then case.submit the model and that it works.....

tonygardella: Tried it with just the CIME_OUTPUT_ROOT and it worked fine so I  removed CESMSCRATCHROOT.

PR #1616, opened 2017-08-31
Update SIPNET netcdf

annethomas:
Updated some of the output variables to standard variables in SIPNET's model2netcdf (and switched to the to_ncvar function from mstmipvar for those), as well as adding an LAI output calculated from the leaf_carbon_content output and SLA from the model run's param file. Added SoilResp to the standard_vars table. A couple of other minor fixes.

## Types of changes

## Checklist:


annethomas:
@mdietze I left the mstmipvar function in place for the Time and FracJulianDay variables in model2netcdf.SIPNET because they aren't in either table but apparently mstmipvar can still convert nonstandard vars to ncvar, just without units. Should I add that functionality to to_ncvar? Or switch those over to direct ncvar_def?

mdietze:
So long as ```time``` is properly defined as a dimension, I don't think we need to retain Year and FracJulianDay in the output at all. No code downstream is expecting these variables (since they're nonstandard) and they're redundant with time -- we can regenerate them on the fly with simple calls to base POSIX and lubridate functions.

annethomas:
@mdietze Took them out

annethomas:
@mdietze Oh, one more thing, should AbvGrndWood in the sipnet output be wood_carbon_content? I came to understand that sipnet, like dalec, had its wood pool as all-plant both in and out but haven't found as clear of an explanation as I have for dalec.

annethomas:
Also, will benchmarking be able to identify the appropriate variable names for old outputs?

mdietze:
@annethomas I think there's been so few benchmarks set up & run, so I'm not too worried about old names unless the benchmark itself is set up using an old name

I was pretty sure that SIPNET had an internal fine and coarse root partitioning -- we looked at that once I thought.  While it may need updating for coarse:stem, but it would be good to retain AbvGrndWood since I think that will be used in benchmarks

Issue #801, opened 2016-04-21
incorrect identification of leap years; multiple implementations. 


dlebauer:
The logic to test if a year is a leap year has been implemented perhaps a dozen times independently [within the PEcAn code base](https://github.com/PecanProject/pecan/search?utf8=%E2%9C%93&q=leap), including once as the function [`is.leap`](https://github.com/PecanProject/pecan/blob/1743b942fdba4adb7cd6cc7716f2f3ae258eb2c0/models/ed/inst/timeutils.R).
2. there are [many cases where `year %% 4 i== 0`](https://github.com/PecanProject/pecan/search?utf8=%E2%9C%93&q=%22year+%25%25+4%22&type=Code) is used although it should be `year %% 400 == 0 | (year %% 4 == 0 & year %% 100 != 0)` (`is.leap` and `leap_year` implement this correctly).
1. The lubridate package has a function `leap_year` that returns T/F. We already use this function [in many places](https://github.com/PecanProject/pecan/search?utf8=%E2%9C%93&q=%22leap_year%22&type=Code) and I suggest that this is a good solution
3. (just a comment / food for thought) testing if a year is a leap year is often followed by calculating time steps (e.g. days or seconds/year). `lubridate` also provides convenience functions for this that would be less error prone than declaring seconds / year or days / month constants repeatedly in the code.

TODO
- [ ] _bugfix:_ All of the invalid uses of `year %% 4 i== 0` should be replaced by a valid function (`leap_year`)
- [ ] _enhancement:_ deprecate `is.leap` and replace it with `lubridate::leap_year`
- [ ] _documentation_ add some notes about existing 'helper' functions to the model template README

robkooper:
Other handy functions to add:

```
days_in_year <- function(year) {
  ifelse(lubridate:: leap_year(year), 366, 365)
}

hours_in_year <- function(year, every=24) {
  24/every * ifelse(lubridate:: leap_year(year), 366, 365)
}
```

now we can simply call `24/6 * days_in_year(1999)`` to get the value for every 6 hours.

ashiklom:
Closed by #1623.

PR #1618, opened 2017-08-31
Model Package cleanups

### LINKED ISSUE #1129: biocro, clm45, dalec, ed, gday, jules, maespa, preles, sipnet package cleanup
Following #1116 @tonygardella will clean up the biocro, clm45, dalec, ed, gday, jules, maespa, preles, and sipnet packages.

tonygardella:
- Closes #1129 
- Checked for wrong leap year implementations #801 
- Deprecating mstmipvar function in favor of PEcAn.utils::to_ncvar
- Checking for library() calls

**Not done testing yet** Still need to double check for things I missed.

## Types of changes

## Checklist:


tonygardella:
@ankurdesai You're paired with me for code review this week.

models/gday/R/model2netcdf.GDAY.R
```diff
@@ -85,32 +85,34 @@ model2netcdf.GDAY <- function(outdir, sitelat, sitelon, start_date, end_date) {
...
+    var[[2]] <- PEcAn.utils::to_ncvar("HeteroResp", dims)
+    var[[3]] <- PEcAn.utils::to_ncvar("TotalResp",dims)
+    var[[4]] <- PEcAn.utils::to_ncvar("GPP", dims)
+    var[[5]] <- PEcAn.utils::to_ncvar("NEE", dims)
+    var[[6]] <- PEcAn.utils::to_ncvar("NPP", dims)
     
     ## C-State
-    var[[7]] <- ncvar_def("AbvGrndWood", "kgC/m2", list(lon,lat,t), -999)
-    var[[8]] <- ncvar_def("TotSoilCarb","kgC/m2", list(lon,lat,t), -999)
-    var[[9]] <- ncvar_def("LAI","m2/m2", list(lon,lat,t), -999)
```
ashiklom: This is different from the other calls because it has the unit in the second slot. Is that intentional? In general, we should probably use named arguments, rather than relying on positions. But more generally, is it safe to assume that the default unit is "kgC/m2/s"? I would probably be more comfortable with no default argument and keeping the unit explicit.

mdietze: Not sure I follow what you mean by "no default argument and keeping the unit explicit". The units look pretty explicit to me, and those are the output standard's units, so I'm not sure I follow what you think is unsafe?

ashiklom: I wasn't sure how the new `PEcAn.utils::to_ncvar` worked. I've since checked, and I see now that it doesn't take a unit argument because it gets those from the standard variables list. So the call `to_ncvar("LAI", "m2/m2", dims)` will not work.

https://github.com/PecanProject/pecan/blob/develop/base/utils/R/to_nc.R#L42

tonygardella: Not intentional. Missed deleting this unit.

PR #1620, opened 2017-09-01
Biocro: Suppress "no visible binding" messages in package check output

### LINKED ISSUE #1309: Clean up biocro check output
`devtools::check("models/biocro")` currently produces a mess of NOTES about documentation and imports/depends problems. Should clean these up at least to the point of being able to see more important warnings.

## Description

- [x] convert in-function `library()` calls to `@import`, qualify out-of-package functions with `::`, etc. 
- [x] Add missing @param descriptions, fix misc Roxygen errors
- [ ] ? suppress "global variable" notes from data.table calls? The messages are spurious (all variables are defined in the table they're operating on), but annoying. May not be worth the trouble unless anyone's planning a CRAN submission.

I'll tackle this after the changes for #1283 are completed.

infotroph:
## Description
Fixes an annoying and potentially-problem-obscuring volume of NOTEs about "no visible binding for global variable" in the devtools::check() output. This is the last remaining item to close #1309.

The root cause of these notes is that the check doesn't know about nonstandard evaluation inside data.table calls. In #1309 I claimed the changes needed to fix these would make the code "WAY uglier", but now I don't think this is so bad.

Please check my deletion in acbaa51b73ca88656bb2f1e941098a19372d29db -- I can't see any way these tranforms aren't no-ops, but the "HAAAACK" comments make me suspect they were there for a reason. @dlebauer do you remember what these calls were supposed to do?

## Types of changes

## Checklist:


models/biocro/R/write.configs.BIOCRO.R
```diff
@@ -46,18 +46,10 @@ convert.samples.BIOCRO <- function(trait.samples) {
   ## transform values with different units cuticular conductance - BETY default is
   ## umol; BioCro uses mol
   if ("b0" %in% trait.names) {
-    trait.samples <- transform(trait.samples, b0 = udunits2::ud.convert(b0, "umol", "mol"))
+    trait.samples$b0 = udunits2::ud.convert(trait.samples$b0, "umol", "mol")
   }
   if ("Sp" %in% trait.names) {
-    trait.samples <- transform(trait.samples, Sp = udunits2::ud.convert(Sp, "kg/m2", "g/cm2"))
-  }
-  if ("vmax" %in% trait.names) {
```
dlebauer: FWIW the following HAAAKs don't don't do anything. They were originally introduced in https://github.com/infotroph/pecan/commit/ecee6fd45c889ce10151fd7ae8abda53c658c2b4#diff-fb09778b34d1b5614bf2c73c9b2b0591 to account for bugs in the biocro code by rescaling Rd and vmax. When the bug was resolved I removed the scaling in https://github.com/infotroph/pecan/commit/4b0340b1605198832107fdbde1ef27be860d5f2e  but should have just removed these lines alltogether.

dlebauer: So OK to merge - thanks for catching this @infotroph!

models/biocro/R/met2model.BIOCRO.R
```diff
@@ -173,16 +173,16 @@ cf2biocro <- function(met, longitude = NULL, zulu2solarnoon = FALSE) {
   }
   
   ## Convert RH from percent to fraction BioCro functions just to confirm
-  if (met[, max(relative_humidity) > 1]) {
-    met[, `:=`(relative_humidity = relative_humidity/100)]
+  if (met[, max(met$relative_humidity) > 1]) {
+    met$relative_humidity = met$relative_humidity/100
```
dlebauer: FWIW I think 'data.table' way was faster, as would the dplyr mutate function, though this is definitely more standard and readable.

models/biocro/R/met2model.BIOCRO.R
```diff
@@ -173,16 +173,16 @@ cf2biocro <- function(met, longitude = NULL, zulu2solarnoon = FALSE) {
...
   }
   newmet <- met[, list(year = lubridate::year(date),
                        doy = lubridate::yday(date),
                        hour = round(lubridate::hour(date) + lubridate::minute(date) / 60, 0),
                        SolarR = ppfd, 
-                       Temp = udunits2::ud.convert(air_temperature, "Kelvin", "Celsius"), 
-                       RH = relative_humidity, 
+                       Temp = udunits2::ud.convert(met$air_temperature, "Kelvin", "Celsius"),
```
dlebauer: Is the `met$` necessary? Why is it used for some variables but not others (like wind_speed and ppfd)?

dlebauer:
Thanks for cleaning this up. Overall it is easier to read and probably clear because it is more explicit. I made a few comments but none are required.

The only thing I would suggest is to use consistent style - using the `dfname$variablename` for all variables in run.biocro.R,

PS The data.table author Matt Dowle and Hadley consider this a problem with R CMD Check. Here is a discussion of the issue: https://stackoverflow.com/questions/9439256/how-can-i-handle-r-cmd-check-no-visible-binding-for-global-variable-notes-when

PR #1625, opened 2017-09-05
Further Biocro data.table style cleanup

infotroph:
## Description

This is a continuation of #1620 -- @dlebauer made some good suggestions in review that seemed worth following up on even though that PR has been merged.

Also of note: Removed an unnecessary conversion from ratio to percentage and back to ratio when computing relative humidity inside `met2model.BIOCRO`.

## Motivation and Context

In [https://github.com/PecanProject/pecan/pull/1620#pullrequestreview-60151116](https://github.com/PecanProject/pecan/pull/1620#pullrequestreview-60151116), @dlebauer said:

> FWIW I think 'data.table' way was faster, as would the dplyr mutate function, though this is definitely more standard and readable. 

I got interested in this speed comparison and [may have gone a *tad* overboard](https://gist.github.com/infotroph/fd76379d8a0e2aa3d7a8ed4ac76dbd2c). Bottom line: data.table is indeed a bit faster, dplyr was way slower, and using base R instead of data.table inside the conditional test gave a surprisingly consistent speedup. I've undone my previous switch to base assignment, but made the `:=` calls consistent throughout.

In [https://github.com/PecanProject/pecan/pull/1620#pullrequestreview-60151578](https://github.com/PecanProject/pecan/pull/1620#pullrequestreview-60151578), @dlebauer said:

> Is the met$ necessary? Why is it used for some variables but not others (like wind_speed and ppfd)?

The met$ is only necessary to shut up R CMD check, with the dual nice side-effects of (1) being a tiny tiny bit faster, and (2) making it clearer that newmet is being constructed from a mix of existing met columns (`met$date`, `met$air_temperature`, `met$relative_humidity`, `met$precipitation_flux`) and newly computed vectors (`wind_speed`, `ppfd`). I've added `met$` to all the local columns to make this more clear.

## Types of changes

## Checklist:

### DIFF
--- a/models/biocro/R/met2model.BIOCRO.R
+++ b/models/biocro/R/met2model.BIOCRO.R
@@ -150,7 +150,7 @@ cf2biocro <- function(met, longitude = NULL, zulu2solarnoon = FALSE) {
         qair = met$specific_humidity,
         temp = udunits2::ud.convert(met$air_temperature, "Kelvin", "Celsius"),
         press = udunits2::ud.convert(met$air_pressure, "Pa", "hPa"))
-      met[, `:=`(relative_humidity = rh * 100)]
+      met[, `:=`(relative_humidity = rh)]
     } else {
       PEcAn.logger::logger.error("neither relative_humidity nor [air_temperature, air_pressure, and specific_humidity]", 
                    "are in met data")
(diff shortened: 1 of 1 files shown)


dlebauer:
Thanks!

PR #1633, opened 2017-09-09
Compile qaqc

### LINKED ISSUE #1632: PEcAn.qaqc not in Makefile
Probably "just" needs to be added to definition of $(BASE) and have appropriate $(call depends,base/qaqc) line.

infotroph:
Fixes #1632, but **discussion needed** before merging.

### Done:
* [x] Add Makefile entries
* [x] Remove space in vignette filename that breaks the whole Make process (!!!)
* [x] Roxygenize (previously hand-documented)

### TODO
* [ ] Add `tests/testthat/Rplots.pdf`, or preferably update tests to remove it when test is complete
* [ ] Clean up code to match current style: Remove `attach`, use explicit `::` namespaces (not essential to make build pass, but ought to do now while we're looking at it)

## Motivation and Context

PEcAn.qaqc is not currently listed in the Makefile, and apparently never has been. This patch builds and installs it alongside the other base packages, but while implementing it I noted that: 

* the package contains only one function (`new.taylor`), 
* it's not used anywhere else in PEcAn, and 
* it has not been updated in several years. 

It seems odd to be adding it to 'base' alongside packages that are heavily used for core functionality. Is the intention that `qaqc` is for core functionality that we just haven't written yet?

So: **Is this patch actually a desirable change right now?**

Pro:
* It's confusing for some packages to be part of "base" but not actually installed
* QA/QC is a desirable functionality that should be expanded, and some users may in fact want to make some Taylor plots.

Con:
* The package is out of date and needs further updates before it will build successfully
* Maybe Taylor plots belong as rightfully in some other package (PEcAn.visualization?) as they do here

Discuss, please. @mdietze can you weigh in on the long-term design part? 

## Types of changes

## Checklist:


tonygardella:
@infotroph Looks like you'll have to follow @ashiklom 's advice and add `Rplots.pdf` to a `.gitignore` for Travis to be happy.

infotroph:
@tonygardella Agreed, but I'm waiting to hear a consensus on whether we actually want to merge this. 
Y'all won't hurt my feelings either way.

mdietze:
I hadn't gone back to see what was currently in QA/QC when I suggested adding it. I agree that if there's just a Taylor diagram function, that probably would be better in _visualization_ or as a metric function in _benchmarking_.  That said, we just had a long discussion during this week's call about writing more QA/QC functions for keeping the database clean, so I don't see us wanting to drop the package.

Sounds like adding to .gitignore is a pretty simple short term fix, so let's do that for now and then merge this in.

PR #1647, opened 2017-09-13
Dataone download util 

LiamBurke24:
Now retains filename when downloading with wget

## Description

- Added `--content-disposition` to retain filename when downloading with wget

- Removed erroneous comment

- updated example

## Motivation and Context

Simplifies the process of identifying downloaded files

## Types of changes

## Checklist:


modules/data.land/R/dataone_download.R
```diff
@@ -20,7 +22,7 @@ dataone_download = function(id, filepath = "/fs/data1/pecan.data/dbfiles/", CNod
   
   ### begin D1 download process
   d1c <- dataone::D1Client("PROD", mnId)
-  pkg <- dataone::getDataPackage(d1c, id = id, lazyLoad = lazyLoad, quiet = quiet, limit = "1MB") # what is the standard limit for pecan downloads?
+  pkg <- dataone::getDataPackage(d1c, id = id, lazyLoad = lazyLoad, quiet = quiet, limit = "1MB")
```
mdietze: Does limit have to be specified? If so I'd make it larger (other places in the system [e.g. met] routinely do much larger downloads than this)

modules/data.land/R/dataone_download.R
```diff
@@ -12,6 +12,8 @@
 #' @export
 #'
 
+#' @examples doi_download(id = "doi:10.6073/pasta/63ad7159306bc031520f09b2faefcf87", filepath = "/fs/data1/pecan.data/dbfiles/")
```
mdietze: This probably need to be labeled as not run, otherwise it might interfere with the package build since most systems won't have that specific filepath.

modules/data.land/R/dataone_download.R
LiamBurke24: Yes, it defaults to 1MB. What would you recommend? 1TB :)

mdietze: I think that's fine for now. As you work on the Shiny part, it would be good (at some point in the future) to see if we could query the download size and ask the user to confirm the download if it's over some threshold (i.e. so someone doesn't accidentally do a 1TB download that will take days and fill up their drive)

PR #1648, opened 2017-09-13
Data upload (VERY ROUGH DRAFT)

LiamBurke24:
This Shiny app will allow users to upload data directly from dataONE via a doi or other identifier OR manually upload data with the fileInput method.

## Description

This is still very rough. The only functioning aspect of this app is dataone download side. 

## Motivation and Context

Ingesting is painful. This should make it easier. 

## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ section for the next release.
 For more information about this file see also [Keep a Changelog](http://keepachangelog.com/) .
 
 ## [Unreleased]
+- Defined Min_upload_level=3
```
mdietze: This needs more context:

1) outside of this PR no one will have any clue what this is referring to
2) I don't think the captures what this PR is adding to the system

shiny/Data-Ingest/app.R
```diff
@@ -0,0 +1,76 @@
...
+      
+    )
+  )
+)
+
+server <- function(input, output) {
+  
+  # d1d <- eventReactive(input$D1Button, { input$id }) #print doi on click
+  
+  # How do I force R to load the dependencies before I run dataone_download? brew, redland, datapack, dataone
```
mdietze: R should handle it's own dependencies. Do I infer from this comment that it wasn't?

shiny/Data-Ingest/app.R
```diff
@@ -0,0 +1,76 @@
...
+   d1d <- eventReactive(input$D1Button, { PEcAn.data.land::dataone_download(input$id) }) #run dataone_download with input from id on click
```
mdietze: How is dataone_download running without an output directory specified?

web/import.data.php
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
```
mdietze: Not following the goal of this page -- from above it looks like you've switched entirely to Shiny.  Should this file be dropped entirely?

CHANGELOG.md
LiamBurke24: Got it. I will update this. Thank you!

shiny/Data-Ingest/app.R
LiamBurke24: It wasn't before, but I think it is now. I think this had something to do with the redland error... So I'll delete this comment now

shiny/Data-Ingest/app.R
LiamBurke24: It defaults to the dbfiles directory path (fs/data1/pecan.data/dbfiles) Should I include the option for the user to change this on the web version? I thought that we were reserving the ability to select the destdir for console users. I can add an output selector option to the dashboard if need be.

web/import.data.php
LiamBurke24: Yes this file should be dropped. I was unaware that it was in the PR. My apologies.

shiny/Data-Ingest/app.R
mdietze: You should be grabbing the dbfile path from the config.php and then passing it in explicitly. Having it only work on the BU node in the network, and not for any collaborators or users, is a pretty big design limitation. This doesn't require a completely fluid destdir output selector -- that seems dangerous.

shiny/Data-Ingest/app.R
```diff
@@ -58,30 +58,37 @@ ui <- dashboardPage(
...
   })
   
-   # output$debug <-
-  
- # 
-  
-  output$upload <- renderTable({
-    if(is.null(data())){return()}
-    input$file
-  })
```
mdietze: Is this chunk deleted because it was the wrong approach? Cause it was superseded by something else?

Issue #1649, opened 2017-09-13
Error in PEcAn.data.atmosphere::download.CRUNCEP()


serbinsh:
When using the latest develop branch I am getting this error when trying to use PEcAn.data.atmosphere::download.CRUNCEP()

```
> remotefunc <- function() {PEcAn.data.atmosphere::download.CRUNCEP(site_id=1000000105, lat.in=9.154, lon.in=-79.848, model=NULL, scenario=NULL, ensemble_member=NULL, overwrite=FALSE, outfolder='/data/Model_Output/pecan.output/dbfiles/CRUNCEP_site_1-105/', start_date='1990-01-01', end_date='2004-12-31')}
> remoteout <- remotefunc()
2017-09-13 08:30:22 INFO   [PEcAn.logger::logger.info] : 
   Downloading 
   /data/Model_Output/pecan.output/dbfiles/CRUNCEP_site_1-105//CRUNCEP.1990.nc 
2017-09-13 08:30:22 INFO   [PEcAn.logger::logger.info] : 
   https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_tair_1990_v1.nc4 
2017-09-13 08:30:32 INFO   [PEcAn.logger::logger.info] : 
   https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_lwdown_1990_v1.nc4 
2017-09-13 08:30:40 INFO   [PEcAn.logger::logger.info] : 
   https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_press_1990_v1.nc4 
2017-09-13 08:30:47 INFO   [PEcAn.logger::logger.info] : 
   https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_swdown_1990_v1.nc4 
2017-09-13 08:30:52 INFO   [PEcAn.logger::logger.info] : 
   https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_uwind_1990_v1.nc4 
2017-09-13 08:30:59 INFO   [PEcAn.logger::logger.info] : 
   https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_vwind_1990_v1.nc4 
[...]

PR #1651, opened 2017-09-13
Proposed bugfix to address #1649

serbinsh:
Added new function and modified data.atmosphere::download.CRUNCEP function to handle slow responses from THREDDS server

## Motivation and Context

Bug fix

## Types of changes

## Checklist:


modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -97,8 +99,9 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
       PEcAn.logger::logger.info(dap_file)
 
       # This throws an error if file not found
-      dap <- ncdf4::nc_open(dap_file)
-
+      #dap <- ncdf4::nc_open(dap_file, verbose=FALSE)
```
infotroph: Old version of line can be deleted instead of commented out

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -113,16 +116,16 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
       }
 
 
-      dat.list[[j]] <- ncdf4::ncvar_get(dap,
+      dat.list[[j]] <- retry.func(ncdf4::ncvar_get(dap,
```
infotroph: Will we need to wrap a retry around *every* ndcf4 function that could potentially be used on a remote file? That seems like it will get messy.

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -113,16 +116,16 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
...
                                  as.character(var$DAP.name[j]),
                                  c(lon_grid, lat_grid, 1),
-                                 c(1, 1, ntime))
+                                 c(1, 1, ntime)), maxErrors=maxErrors, sleep=sleep)
 
-      var.list[[j]] <- ncdf4::ncvar_def(name = as.character(var$CF.name[j]),
+      var.list[[j]] <- retry.func(ncdf4::ncvar_def(name = as.character(var$CF.name[j]),
```
infotroph: My understanding is that ncvar_def works locally (it sets up a variable in memory that can later be written to the output file), so maybe this one can stay unwrapped?

modules/data.atmosphere/R/download.CRUNCEP_Global.R
mdietze: I don't think it will get too messy, there's really not THAT many remote file connections in the system and while it would be nice to be preventative, only a subset of those have been a source of time out issues.  I think this PR is a really useful addition for those cases, and there are places we know this will be an issue that we can get ahead of now (e.g. @annethomas fixes to the MODIS function)

PR #1653, opened 2017-09-13
dvmdostem

serbinsh:
Initial dvmdostem push to PEcAn_develop

## Description
Initial working version (though still rough) of the dvmdostem wrappers

## Motivation and Context
Working toward dvmdostem integration

## Types of changes

## Checklist:

- [X ] My change requires a change to the documentation.


serbinsh:
FYI -  this is a rough working version and we have plans for further generalization and cleanup but was worried about keeping this hidden and making the PR even larger

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+                           vals=c(1), # <=== read from dvmdostem file!
+                           longname="coordinate_longitude")
+
+  latd <- ncdf4::ncdim_def(name='lat', 
+                           units="degrees_north",
+                           vals=c(1), # <=== read from dvmdostem file! 
+                           longname="coordinate_latitude")
+
+  # Odd - not sure what PEcAn uses this for as there should be one
+  # file for each year. Maybe monthly or daily resolution outputs?
```
mdietze: correct, many models are outputting variables at daily or subdaily resolution

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+##' @export
+##'
+##' @author Tobey Carman, Shawn Serbin
+##' @importFrom ncdf4 ncdim_def ncvar_def ncatt_get ncvar_add
+model2netcdf.dvmdostem <- function(outdir) {
+
+  # Define PEcAn style dimensions
+  lond <- ncdf4::ncdim_def(name='lon',
+                           units="degrees_east",
+                           vals=c(1), # <=== read from dvmdostem file!
```
mdietze: If these values have not yet been read in, but will be read from the model's output file, then you should just wait to define the dimensions until you have the data in hand so that you don't have to figure out how to change it to the correct values later.

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+                            units='years since 2000-01-01 00:00:00',
+                            vals=c(0),
+                            unlim=TRUE,
+                            longname="time",
+                            calendar='standard')
+  
+  # Define PEcAn style variables
+  nppv <- ncdf4::ncvar_def(name="NPP",
+                           units="kgC m-2 s-1",
+                           dim=list(lond, latd, timed)) # WTF? dimension spec seems to be backwards...
```
mdietze: Time is last because look-up is optimized for plotting time-series at locations, rather than maps across locations

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+  
+
+  # Open a dvmdostem output file  
+  d <- ncdf4::nc_open(file.path(outdir, "NPP_yearly_tr.nc"))
+  #d = ncdf4::nc_open('NPP_yearly_tr.nc')
+  a <- ncdf4::ncvar_get(d, 'NPP')
+  nppvals <- a[10,10,]
```
mdietze: Could you document what specifically the dimensions of ```a``` are and why the first two are hard-coded as 10,10? The may be a good reason for this, but without access to the file itself I get nervous about hard-coded constants.

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+  
+  print(paste0("Length of time dim: ", d$dim$time$len))
```
mdietze: would prefer PEcAn.logger::logger.debug over a print statement (since we can turn the message off when not in debug mode)

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+  ctr <- 1
+  #out_yr <- 1998
+  out_yr <- 1901
```
mdietze: This strikes me as the sort of thing that should come from the output file rather than being hard coded

models/dvmdostem/R/model2netcdf.dvmdostem.R
```diff
@@ -0,0 +1,76 @@
...
+  for (yr in d$dim$time$val) {
+
+
+    # Write to file
+    new_pecan_style_file = ncdf4::nc_create(file.path(outdir, paste0(out_yr, ".nc")), list(nppv))
+
+    ncdf4::ncvar_put(nc=new_pecan_style_file, varid=nppv, vals=nppvals[ctr], verbose=TRUE)
+    ctr <- ctr + 1
+    out_yr <- out_yr + 1
```
mdietze: This structure is pretty surprising, since yr is never used anywhere. Does ```d$dim$time$val``` really not refer to a time dimension that can be coerced into the out_year?  Why not set the loop up as ```ctr in seq_along(d$dim$time$val)```?

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+  # that make the loops in write.config.dvmdostem easier to 
+  # understand/deal with
+  trait_names <- colnames(trait_values)
+  # for debugging
+  PEcAn.logger::logger.info(trait_names)
+  
+  # Not really sure how this should work yet? 
+  # Example from FATES model
+  # trait.names[trait.names == "leaf_respiration_rate_m2"]   <- "atref.rd"
+  #                                pecan notion?                dvmdostem?
```
mdietze: Correct, this would be an example of changing a name from PEcAn standard to model-specific

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+  # trait_names[trait_names == "CMT04.Salix.SLA"] <- ????? jsondata$pft0$sla
+  colnames(trait_values) <- trait_names
+
+  ### Conversions (for example, convert SLA to m2/g?)
+  ### TODO: !! Need to remove PFT name from here and just look for variable names !!
+  ###       !! Should only have train name here, such as SLA or cuticular_conductance
+  if ("CMT04.Salix.SLA" %in% names(trait_values)) {
+    # Convert from m2 / kg to m2 / g
+    trait_values[["CMT04.Salix.SLA"]] <- trait_values[["CMT04.Salix.SLA"]] / 1000.0
```
mdietze: Not sure why this is "CMT04.Salix.SLA" -- seems like it should just be "SLA" because that's the PEcAn standard name and you haven't converted SLA to a different name. In general, we haven't found the need for PFT-specific unit conversions in other models.

Furthermore, we recommend using udunits for unit conversions over just dividing by 1000. The conversion's pretty trivial in this case, but udunits makes it self documenting and avoids errors (e.g. udunits2::ud_convert(trait_values[['SLA']],'m2 kg-1','m2 g-1')

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+  }
+  if ("CMT04.Salix.cuticular_cond" %in% names(trait_values)) {
+    # Convert from umol H2O m-2 s-1 to ??? (need to find dvm-dos-tem-units)
+    trait_values[["CMT04.Salix.cuticular_cond"]] <- trait_values[["CMT04.Salix.cuticular_cond"]] / 10^9
```
mdietze: likewise, PEcAn variable name is just cuticular_cond

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+  # place. The basic flow will be like this:
+  #  - Read dvmdostem parameter file into json object, load into memory
+  #  - Update the in-memory json object
+  #  - Write the json object back out to a new dvmdostem parameter file
+
+  # Start by figuring out which community number we are working on.
+  # This is important because we need to pull the correct "community block"
+  # from the dvmdostem parameter files. Each trait in the traits dataframe
+  # should have a name like this:
+  #   CMT04.Salix.extinction_coefficient_diffuse
```
mdietze: not sure how/why this would be true. If "CMT04.Salix" is your pft and "extinction_coefficient_diffuse" is your trait, then one of these would be the row and the other would be a column. Or, if you leave traits in a list (recommended), the PFT will be the elements of the list and the parameters will be the named elements in the vectors.

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+  # !! hard coding in this function. 
+  cmtnum <- strsplit(unlist(strsplit(names(trait_df)[1], '.', fixed=TRUE))[1], "CMT")
+  cmtnum <- unlist(cmtnum)[2]
+  cmtnum <- as.numeric(cmtnum)
+
+  # Next, use a helper script distributed with dvmdostem to read the dvmdostem
+  # parameter data into memory as a json object, using a temporaroy json file
+  # to hold a representation of each dvmdostem parameter file.
+  dimveg_params <- paste(appbinary_path, "parameters", 'cmt_dimvegetation.txt', sep="/")
+  envcanopy_params <- paste(appbinary_path, "parameters", 'cmt_envcanopy.txt', sep="/")
```
mdietze: These look like they should be file.path not paste

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+
+  dimveg_jsonfile <- '/tmp/dvmdostem-dimveg.json'
+  envcanopy_jsonfile <- '/tmp/dvmdostem-envcanopy.json'
```
mdietze: This seems dangerous, as different sites and ensemble members would all be using the same tmp file.

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+          args=(c("--dump-block-to-json", envcanopy_params, cmtnum)),
+          stdout=envcanopy_jsonfile, wait=TRUE)
+
+  # Read the json file into memory
+  library("rjson")
+  dimveg_jsondata <- fromJSON(paste(readLines(dimveg_jsonfile), collapse=""))
+  envcanopy_jsondata <- fromJSON(paste(readLines(envcanopy_jsonfile), collapse=""))
+
+  # (2)
+  # Overwrite certain parameter values with (ma-posterior) trait data
```
mdietze: "(ma-posterior)" seems misleading, recommend dropping.

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+    #print(cur_t_name)
+
+    # In converting trait.values into a dataframe, we get 
+    # these long column names that are dot separted and seem
+    # to contain cmtkey.pftname.traitname
+    # So the following splits out the various components into 
+    # convenient local variables
+    cmtkey <- unlist(strsplit(cur_t_name, '.', fixed=TRUE))[1]
+    pftname <- unlist(strsplit(cur_t_name, '.', fixed=TRUE))[2]
+    tname <- unlist(strsplit(cur_t_name, '.', fixed=TRUE))[3]
```
mdietze: If you do need to do this (which still seems wrong to me), I'd at least do the ```unlist(strsplit(cur_t_name, '.', fixed=TRUE))``` once to a variable, and then use that variable to extract the 3 parts you need.

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+
+        # Then pick out only the entry for the correct pft 
+        if (identical(cur_pft_name, pftname)) {
+          print(cur_t_name)
+          print(tname)
+          if (tname == "SLA") {
+            dimveg_jsondata[[jd_name]]$sla = trait_df[[cur_t_name]]
+          } else if (tname == "frprod_perc_10") {
+            dimveg_jsondata[[jd_name]]$`frprod[0]` = trait_df[[cur_t_name]]
+          } else {
```
mdietze: 1) The point of doing the earlier conversion from PEcAn names to Model names is to avoid having to so a big set of if's here -- if everything is in model names you should be able to just access the right variable directly
2) If you *do* end up with a big set of if's here (which isn't uncommon, I've done something similar in lots of models), then I really don't see the point of having a convert.samples function, since you can just do the name matching and unit conversions right here. Given what you've got here so far, it sure seems to me that the decision to write a convert.samples function has made things harder on you.

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+            # pass...variable not in this file or datastructure
+            PEcAn.logger::logger.info("pass...variable not in this file or datastructure")
+          }
+        }
+      }
+    } # end loop over json for env canopy
+  }
+
+  # Write it back out to disk (overwriting ok??)
+  dimveg_exportJson <- toJSON(dimveg_jsondata)
```
mdietze: namespace needed on toJSON

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+  write(envcanopy_exportJson, "/tmp/envcanopy_newfile.json")
+
+
+  # (3)
+  # Format a new dvmdostem parameter file using the new json file as a source.
+
+  if (dir.exists(file.path(rundir, "parameters/"))) {
+    # pass
+  } else {
+    print("No parameter/ directory in run directory! Need to create...")
```
mdietze: should use a logger function

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+
+
+  # TODO:
+  #  [x] finish with parameter update process
+  #  [x] dynamically copy parameters to right place
+  #  - dynamically copy the output_spec from insts folder to the
+  #    right place (see variable above for getting stuff from inst)
+  #  - figure out how to handle the met.
+  #     -> step one is symlink from raw data locations (Model install folder)
+  #        into pecan run folder, maybe do this within job.sh?
```
mdietze: If you symlink to the met location, then yes you need to do that in job.sh since the model and met may be on a remote server, and thus you can't create that link right now. If you copy the met then you can do that now since for remote runs the whole rundir will be rsynced to the remote machine.

models/dvmdostem/R/write.config.dvmdostem.R
```diff
@@ -0,0 +1,353 @@
...
+
+  # Get a copy of the config file written into the run directory with the
+  # appropriate template parameters substituted.
+  if (!is.null(settings$model$configtemplate) && file.exists(settings$model$configtemplate)) {
+    config_template <- readLines(con=settings$model$config_template, n=-1)
+  } else {
+    config_template <- readLines(con=system.file("config.js.template", package = "PEcAn.dvmdostem"), n=-1)
+  }
+
+  config_template <- gsub("@INPUT_DATA_DIR@", file.path(dirname(appbinary), "DATA/SewardPen_10x10"), config_template)
```
mdietze: seems like something that needs generalization

models/dvmdostem/inst/config.js.template
```diff
@@ -0,0 +1,85 @@
...
+  },
+
+  "IO": {
+    "parameter_dir":      "parameters/",
+    "hist_climate_file":  "@INPUT_DATA_DIR@/historic-climate.nc",
+    "proj_climate_file":  "@INPUT_DATA_DIR@/projected-climate.nc",
+    "veg_class_file":     "@INPUT_DATA_DIR@/vegetation.nc",
+    "drainage_file":      "@INPUT_DATA_DIR@/drainage.nc",
+    "soil_texture_file":  "@INPUT_DATA_DIR@/soil-texture.nc",
+    "co2_file":           "@INPUT_DATA_DIR@/co2.nc",
```
mdietze: Of the things in this list, PEcAn should be able to provide climate, veg, soil texture, and co2 at the moment -- most everything else will need to have model defaults in the near-term. However, you probably want to template the file names themselves as well since there's no guarantee that the name of the file PEcAn is passing in will be the same as the one listed here. Also, does TEM strictly require that all these files be in the same directory? PEcAn will have them in different directories by default and thus it'd be simpler to just have these paths be where pecan has the files stored (which it will provide you under settings$run$inputs), rather than adding the extra step of creating a bunch of symlinks in INPUT_DATA_DIR

models/dvmdostem/inst/dvmdostem.pecan.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
```
mdietze: If this fill is just for your own personal testing, it probably doesn't have to be committed. If it's a more general test, it should be in the tests directory. That said, this specific file won't run since it lacks run, site, and input information.

models/dvmdostem/inst/workflow_small.R
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env Rscript
```
mdietze: not sure this script needs to be committed either

mdietze:
Also, need to update CHANGELOG and documentation. Specifically, the latter should make explicit the existence of the tem coupling, the stage you're at in the process (I think there's a table for that),  information/links about how to download and install the model, and probably some basic citations. 

https://pecanproject.github.io/pecan-documentation/master/pecan-models.html

https://pecanproject.github.io/pecan-documentation/master/installing-pecan.html

base/utils/data/trait.dictionary.csv
```diff
@@ -91,4 +91,6 @@
 "SPRTMX";"SPRTMX";"cm";""
 "DMAX";"DMAX";"degree days";""
 "DMIN";"DMIN";"degree days";""
-"leaf_longevity";"leaf longevity";"year";""
\ No newline at end of file
+"leaf_longevity";"leaf longevity";"year";""
+"frprod_perc_10";"fine root allocation percentage vertically in soils (every 10 cm)";"0-100";"frprod_perc_10"
+"gcmax";"Maximum canopy conductance";"m s-1";"gcmax"
```
mdietze: Would be good for this to have the same units as other conductances (kg m-2 s-1)

base/utils/data/trait.dictionary.csv
```diff
@@ -91,4 +91,6 @@
 "SPRTMX";"SPRTMX";"cm";""
 "DMAX";"DMAX";"degree days";""
 "DMIN";"DMIN";"degree days";""
-"leaf_longevity";"leaf longevity";"year";""
\ No newline at end of file
+"leaf_longevity";"leaf longevity";"year";""
+"frprod_perc_10";"fine root allocation percentage vertically in soils (every 10 cm)";"0-100";"frprod_perc_10"
```
mdietze: This strikes me as a poor (and too model specific) choice for a trait name. First, the 'every 10cm' part implies that there's not one allocation, but a whole host of allocation parameters that have a joint, sum to 100%, constraint. Second, I'd think about this from a data-driven perspective -- (A) if we wanted to store empirical root profile and root growth data by depth, how would we do that and what should we call it? (B) given that data, how do we translate those posteriors into the thing this specific model needs? Finally, I tend to find proportion data easier to work with than percent.

tonygardella:
@serbinsh Can this be closed in favor of #1814?

tonygardella:
Closing because of #1814

PR #1655, opened 2017-09-13
Fix netcdf writing in call_MODIS

annethomas:
## Description
The call_MODIS function now writes a netcdf file with dimensions corresponding to the dates and spatial grid requested. The time dimension is in units days since the start year (which starts at 0, so it may be shifted one from what you expect when translated into standard dates). The spatial dimensions are simply nrow and ncol, which is derived from the size requested (currently the function only does squares; each additional cell is 1 km2). Future implementations should convert this to actual lat/long or eastings/northings. Because of a discrepancy between row and column-dominant layouts in the MODIS and netCDF data structures, the dimension listed as "nrow" actually corresponds to the columns when the variable is printed out and vice versa. The time dimension is the highest level dimension. Based on the correspondence between MODIS raw number and geotiff formats, the true rows (nrow; look like columns) are latitudinal and the true columns are longitudinal.

## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - Cleanup of leap year logic, using new `PEcAn.utils::days_in_year(year)` function (#801).
 - Replace many hard-coded unit conversions with `udunits2::ud.convert` for consistency, readability, and clarity
 - Added a new retry.func() to base/utils to provide ability to re-try a function X times before stopping.  Currently using this function in the download.CRUNCEP() function to handle slow responses from THREDDS.
+-Reformatted call_MODIS netcdf output
```
mdietze: In the future, I'd recommend a bit more detail -- you're not saying anything about how it was reformatted

annethomas:
Shoot, I wrote a whole comment and it didn't post. So the discrepancy is between MODIS and both R/Python, and nrow is referring to the original MODIS array, which Python and R interpret as columns. I'm thinking changing the names of the dims would help (nrow would correspond to lat and ncol to lon, if I'm thinking right, except right now only the center coordinate is provided so the calculation would need to be added in).

Here's an example (in a rectangle even though currently the function only produces squares):
if time (dim 3) is length 4, rows (in the MODIS sense; dim 2 here, which appears as columns) are length 3 and cols (dim 1) are 2, and the grid from MODIS' perspective is [[[1,2],[3,4],[5,6]], [[1,2],[3,4],[5,6]], [[1,2],[3,4],[5,6]], [[1,2],[3,4],[5,6]]] 
(except without any brackets or commas which is what makes it hard to transform)
then ncdf4 reads the file out like this: (so it’s tipped on its side)
> ncvar_get(np3,"LAI")
, , 1
     [,1] [,2] [,3]
[1,]    1    3    5
[2,]    2    4    6

, , 2
     [,1] [,2] [,3]
[1,]    1    3    5
[2,]    2    4    6

, , 3
     [,1] [,2] [,3]
[1,]    1    3    5
[2,]    2    4    6

, , 4
     [,1] [,2] [,3]
[1,]    1    3    5
[2,]    2    4    6

Is there a better way?

mdietze:
1) It is less important to me whether the data is [lat, lon ,time] or [lon, lat, time] so long as it's labeled correctly
2) If you have a matrix
```
1 3 5
2 4 6
```
and you want to flip it to
```
1 2
3 4
5 6
```
then that's just a transpose (```t()``` in R, ```numpy.transpose``` in Python)

annethomas:
Ok, so as it stands nrow does correctly label the dimension that is rows in the original data, even though it becomes columns in the netcdf. The main problem is probably that I'm not as nimble in python as R, but the reason it's not as easy as transposing is because each date's grid comes from MODIS squashed into this linear form that doesn't match the row/column scheme of netcdf (or numpy's matrix) so I would need to do some manipulating to fundamentally reorder the numbers and get it to a transposable format before netcdf creation (which is currently the only thing that creates a matrix from each linear array).  

Would you rather I leave it, rename the dims lat/lon (even though it's not the actual coordinates), rename them something else, or work on getting the matrix stuff figured out?

mdietze:
@annethomas what matters to me is that the dimensions are labeled correctly.  Definitely **don't** have rows labeled 'cols' and vice versa. If the grid is NOT lat/lon then I don't want it labeled lat/lon either. But row/col is fundamentally the wrong label too since it tells me nothing about what the dimension / orientation of the data actually is. I'm guessing 'northing' and 'easting' are closer to the correct labels.

annethomas:
Yeah, that makes sense. Do you know if there's an easy python way to add 1 km to a given lat/lon (the center cell)? Otherwise I can go with easting/northing, although for that to be technically accurate would also probably need some conversion.

dlebauer:
@annethomas the terraref project has Python utilities for converting among a local grid in meters, utm, and lat lon. https://github.com/terraref/terrautils/blob/master/terrautils/spatial.py The parameters may need to change but the undertook transformations may help.

annethomas:
Ok, @mdietze another problem: when I went to label units on Northing, I realized the numbers currently being put in, 1 to the total number of rows, would then actually be "km south of the top of the grid". Is that too confusing? Should I not label the units or switch the #s to go from the bottom so that it's actually north of the reference point?

annethomas:
I just pushed the basic label change without units so let me know if you want me to do anything about that

PR #1668, opened 2017-09-15
Changes to model2netcdf.ed for reading monthly files

istfer:
Currently PEcAn can only read Tower files of ED and perform analysis on those output variables. I added functionality for reading monthly `-E-` files and writing values to the ncdf files.

## Description

- Took out current code to their own functions, namely, `read_T_files`, `put_T_values`
- Added two other functions to do the same thing for `E` files (`read_E_files`, `put_E_values`)
- Modified read.output, sensitivity, and write.configs.ED code to accommodate changes, it should be possible to perform SA on `DBH` output now
- These changes shouldn't break the current functionality, will test it thoroughly in the next couple of days, but wanted to PR so that I can work on the requested changes in the meantime

## Motivation and Context

I need to be able to conduct SA and PDA with demographic outputs of ED

## Types of changes

## Checklist:


models/ed/R/write.configs.ed.R
```diff
@@ -506,6 +506,9 @@ write.config.jobsh.ED2 <- function(settings, run.id) {
 
   jobsh <- gsub("@START_DATE@", settings$run$start.date, jobsh)
   jobsh <- gsub("@END_DATE@", settings$run$end.date, jobsh)
+  
+  jobsh <- gsub("@PFTNAMES@", sapply(settings$pfts, `[[`, "name"), jobsh)
+  jobsh <- gsub("@DHBREAKS@", unlist(settings$pfts$dbh.breaks), jobsh)
```
istfer: note to self: this needs changing, `gsub` uses the first element only

base/utils/R/read.output.R
```diff
@@ -141,6 +141,11 @@ read.output <- function(runid, outdir, start.year = NA, end.year = NA, variables
       for (v in variables) {
         if (v %in% c(names(nc$var), names(nc$dim))) {
           newresult <- ncdf4::ncvar_get(nc, v)
+          if("pft.numbers" %in% names(nc$dim)){
```
mdietze: Standard dimension name is pft not pft.numbers
https://github.com/PecanProject/pecan/blob/develop/base/utils/data/standard_vars.csv#L5

base/utils/R/read.output.R
```diff
@@ -141,6 +141,11 @@ read.output <- function(runid, outdir, start.year = NA, end.year = NA, variables
...
+            # means there are PFT specific outputs we want
+            pft.ind <- strsplit(nc$var$PFT$longname, ",")[[1]] == pft.name
```
mdietze: 1) Intuitively, I'd expect this code to grab all PFTs if pft.name isn't specified explicitly (default = NULL). I'm pretty skeptical that this code will do that correctly
2) nc$var$PFT is also not referencing a standard name. ```pft``` should be a dim not var
3) attribute should be long_name not longname
4) long_name should be accessed using ncdf4::ncatt_get

base/utils/R/sensitivity.R
```diff
@@ -51,7 +51,10 @@ read.sa.output <- function(traits, quantiles, pecandir, outdir, pft.name = "",
       run.id <- sa.run.ids[[pft.name]][quantile, trait]
       
       for(var in seq_along(variables)){
-        out.tmp <- read.output(run.id, file.path(outdir, run.id), start.year, end.year, variables[var])
+        out.tmp <- read.output(runid = run.id, outdir = file.path(outdir, run.id), 
+                               start.year = start.year, end.year = end.year, 
+                               variables = variables[var], 
+                               pft.name = pft.name)
```
mdietze: This change causes a subtle, but non-trivial, change in functionality. For example, if I previously ran a multi-PFT SA for NPP, pecan would have extracted the sensitivity of TOTAL NPP to the pft's parameters one at a time. Now we're going to get the PFT-SPECIFIC NPP's sensitivity to the pft's parameters. There doesn't appear to be any backward compatibility to get to the original functionality.  

Maybe we could change the way we SA/UA/EA variables to allow dimension specifications on variables? e.g. a SA on NPP would do total NPP while a SA on NPP[early_hardwood] would run a SA on that specific PFT?

models/ed/inst/template.job
```diff
@@ -45,7 +45,7 @@ if [ ! -e "@OUTDIR@/history.xml" ]; then
 
   # convert to MsTMIP
   echo "require (PEcAn.ED2)
-model2netcdf.ED2('@OUTDIR@', @SITE_LAT@, @SITE_LON@, '@START_DATE@', '@END_DATE@')
+model2netcdf.ED2('@OUTDIR@', @SITE_LAT@, @SITE_LON@, '@START_DATE@', '@END_DATE@', '@PFTNAMES@', '@DHBREAKS@')
```
mdietze: Do we really want model2netcdf.ED2 to only process specific PFTs and DBH bins? I'd expect the default to be all PFTs and size classes

models/ed/R/write.configs.ed.R
mdietze: Is this supposed to be DBHBREAKS not DHBREAKS?

Since this seems like an optional arg, it would be good to first check that settings$pfts$dbh.breaks exists before doing this substitution, and having a sensible alternative if it's not specified

A change like this requires an update to the documentation

base/utils/R/sensitivity.R
mdietze: Any change of functionality at this basic level requires an update to the SA documentation.

istfer: as long as there is no pft dimension in the netcdf file (we almost never use pft as a dimension in `model2netcdf`s if I'm not mistaken) read.output won't read the outputs by PFT (it will never get into the `if("pft.numbers" %in% names(nc$dim))` block)

but I see that, that check is not going to be enough because we can have both T and E output files in ED (and have pft specific outputs in LPJ-GUESS in addition to cumulative outputs) and we should be able to request an SA on NPP and DBH on the outputs of the same set of runs, then the ncdf will have a pft dimension and extract the wrong thing for NPP. 

So, we can either use variable names to specify this as you suggest or we can check the variable dimensions in the netcdf (as opposed to the broad check whether a pft dimension is specified or not in the ncdf). e.g. NPP variable dimensions will be [lat,lon,t], we'll read the whole thing, whereas DBH will have pft in its dimension, we can check  if that's the case, and only read output for that pft

models/ed/R/model2netcdf.ED2.R
```diff
@@ -118,497 +226,661 @@ model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date, end_date) {
...
+put_E_values <- function(yr, nc_var, out, begins, ends, pft.names, dbh.breaks, ...){
+  
+  s <- length(nc_var)
+  
+  data(pftmapping, package = "PEcAn.ED2")
+  pfts <- sapply(pft.names, function(x) pftmapping$ED[pftmapping$PEcAn == x]) 
+  
+  # ----- fill list
+  
+  t <- ncdf4::ncdim_def(name = "dtime", units = paste0("days since ", yr, "-01-01 00:00:00"),
```
istfer: @mdietze how should I define the time dimension for monthly outputs? if we have both T and E files, their time steps are different..should we align their time steps (writing half hourly DBH sounds weird), but then the alternative would be defining two time dimensions, what do you think?

models/ed/inst/template.job
istfer: it processes all, these arguments have multiple elements e.g.

```
pft.names = c("temperate.Early_Hardwood", "temperate.Late_Conifer", "temperate.Late_Hardwood")
dbh.brakes = c(0,50,100)
```
but I think pft.names should be possible to extract from the output file itself on the fly

mdietze: Yes, I understood that these variables can be vectors, but the code doesn't seem to allow them to have a default. Because it's not a 'standard' pecan setting, if someone doesn't manually add DHBREAKS to their pecan.xml via Advanced Edit, it appears that job.sh will throw an error.

models/ed/R/model2netcdf.ED2.R
mdietze: Define two time dimensions

istfer: what should those time dimension names be? currently one is `time` the other is `dtime`

mdietze: Not 100% sure. Maybe ```month```?

Here's some things to take a look at, but neither is a definitive answer
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html
http://www.unidata.ucar.edu/software/netcdf/docs/BestPractices.html

mdietze:
@istfer are you still working on this PR or are you just waiting for review?

istfer:
no sorry, this still needs some work, will finish up next week

tonygardella:
@istfer Is this something you'll come back to in the coming weeks?

istfer:
yes I will @tonygardella

base/utils/R/sensitivity.R
istfer: So the block in `read.output` is changed from `if("pft.numbers" %in% names(nc$dim))` to `if("pft" %in% sapply(nc$var[[v]]$dim, `[[`, "name"))` so it won't try to read pft specific SA outputs for NPP because it won't have `pft` as a dimension for models like ED2

but models like LPJ-GUESS (and I guess, Linkages as well?) can give outputs noth in total NPP and NPP per PFT. As we don't have a "NPP per PFT" variable (so that we cannot write NPP and NPP-per-PFT in our standard netcdfs), it makes sense to write NPP variable per pft for a model by default when possible, and read.output would read it per pft by default (when the variable has  `pft` as a dimension). Then the SA/EA code should know whether we want total NPP or pft specific NPP, as you suggest @mdietze:

>Maybe we could change the way we SA/UA/EA variables to allow dimension specifications on variables? e.g. a SA on NPP would do total NPP while a SA on NPP[early_hardwood] would run a SA on that specific PFT?

So it should be a tag in the xml right? e.g.

```
<sensitivity.analysis>
  <quantiles>
    <sigma>-1</sigma>
    <sigma>0</sigma>
    <sigma>1</sigma>
  </quantiles>
  <variable>DBH</variable>
  <perpft>TRUE</perpft>
</sensitivity.analysis>
```
When it doesn't see `<perpft></perpft>` tag it would default to original functionality. What do you think?

Also, does it make sense to do pft-specific ensemble? I guess we want that functionality in as well?

tonygardella:
@istfer Is this still in progress?

istfer:
Not sure about updates to the std vars table, but I think I'm mostly done with this PR, it's ready for review again @tonygardella 

in the meantime I'll test it more thoroughly

tonygardella:
@mdietze Could you review this again?

base/utils/R/read.output.R
```diff
@@ -80,6 +80,32 @@ read.output <- function(runid, outdir, start.year = NA, end.year = NA, variables
       for (v in variables) {
         if (v %in% c(names(nc$var), names(nc$dim))) {
           newresult <- ncdf4::ncvar_get(nc, v)
+          # begin per-pft read
+          if("pft" %in% sapply(nc$var[[v]]$dim, `[[`, "name")){
+            # means there are PFT specific outputs we want
+            # parse pft names and match the requested
+            pft.string <- ncdf4::ncatt_get(nc, "PFT")
```
mdietze: When referencing pft/PFT, seem to be switching case between this line and the `if` right above. Is there a reason we're not consistent?

base/utils/data/standard_vars.csv
```diff
@@ -104,7 +105,8 @@ Root carbon content, optionally by size class; alternatively specify fine_ and c
 "PAR","surface_downwelling_photosynthetic_photon_flux_in_air","mol m-2 s-1","Photosynthetically Active Radiation","Driver","real","lon","lat","time",NA,"Photosynthetically Active Radiation"
 "precipf",NA,"kg m-2 s-1","Precipitation","Driver","real","lon","lat","time",NA,"The per unit area and time precipitation representing the sum of convective rainfall, stratiform rainfall, and snowfall"
 "BA",NA,"m2 ha-1","Basal area","Diversity","real","lon","lat","time","pft","Basal area by PFT"
-"Dens",NA,"1 ha-1","Stem Density","Diversity","real","lon","lat","time","pft","Stem Density dy PFT"
+"Dens",NA,"1 ha-1","Stem Density","Diversity","real","lon","lat","time","pft","Stem Density by PFT"
+"DBH",NA,"cm","Diameter at Breast Height","Diversity","real","dbhsize","time","pft","NA","DBH by PFT"
```
mdietze: This variable is ambiguous. If DBH has a size class dimension that is cm, what numbers are going into `DBH` itself that could also have dimensions of cm?

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -217,7 +218,7 @@ Only if this section is defined a sensitivity analysis is done. This section wil
 * **end.date** : [required?] end date of the sensitivity analysis (in YYYY/MM/DD format)
 * **_NOTE:_** start.date and end.date are distinct from values set in the run tag because this analysis can be done over a subset of the run.
 * ** variable** : [optional] name of one (or more) variables the analysis should be run for. If not specified, sensitivity.analysis variable is used, otherwise default is GPP.
-
+* ** perpft** : [optional] if `TRUE` a sensitivity analysis on PFT-specific outputs will be run. This is only possible if your model provides PFT-specific outputs for the `variable` requested.
```
mdietze: Do we really want the default for this to be TRUE? For some models that could cause a very large number of runs to be started. It seems to me that we'd want to default this to FALSE.

models/ed/R/model2netcdf.ED2.R
```diff
@@ -21,25 +22,155 @@
 ##'
 ##' @author Michael Dietze, Shawn Serbin, Rob Kooper, Toni Viskari, Istem Fer
 ## modified M. Dietze 07/08/12 modified S. Serbin 05/06/13
-model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date, end_date) {
-
-  flist <- dir(outdir, "-T-")
-  if (length(flist) == 0) {
-    print(paste("*** WARNING: No tower output for :", outdir))
+model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date, end_date, pft_names = NULL, dbh_breaks = 0) {
```
mdietze: Added two new arguments but didn't update the roxygen tags to define what those variables are.

models/ed/R/model2netcdf.ED2.R
```diff
@@ -21,25 +22,155 @@
...
+  flist[["-T-"]] <- dir(outdir, "-T-") # tower files
+  flist[["-E-"]] <- dir(outdir, "-E-") # monthly files
+  
+  # check if there are files
+  file.check <- sapply(flist, function (f) length(f) != 0)
+  
+  if (!any(file.check)) {
+    
+    # no output files
+    print(paste("*** WARNING: No output files found for :", outdir))
```
mdietze: warnings should use use the PEcAn logger warning function, not print

models/ed/R/model2netcdf.ED2.R
```diff
@@ -118,497 +250,721 @@ model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date, end_date) {
...
-  # TODO - remove this function and replace with ifelse statements inline below (SPS)
-  conversion <- function(col, mult) {
-    ## make sure only to convert those values that are not -999
-    out[[col]][out[[col]] != -999] <- out[[col]][out[[col]] != -999] * mult
-    return(out)
+  # note that there is always one Tower file per year
+  ysel <- which(yr == yfiles)
+  
+  if (yr < strftime(start_date, "%Y")) {
+    print(paste0(yr, "<", strftime(start_date, "%Y")))
```
mdietze: should be a logger, not a print

models/ed/R/model2netcdf.ED2.R
```diff
@@ -118,497 +250,721 @@ model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date, end_date) {
...
+    next
   }
-
-  checkTemp <- function(col) {
-    out[[col]][out[[col]] == 0] <- -999
-    return(out)
+  
+  if (yr > strftime(end_date, "%Y")) {
+    print(paste0(yr, ">", strftime(end_date, "%Y")))
```
mdietze: logger

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
istfer: I'm confused, this tag doesn't change the number of sensitivity runs, just enables the reading of the outputs per pft. Current SA workflow by default does start a larger number of runs if there are larger number of PFTs, no?

models/ed/R/model2netcdf.ED2.R
istfer: these are remnants from the current code, will change

base/utils/R/read.output.R
mdietze: Ok, that should be documented inline since I just demonstrated that someone else going to read/modify that code will miss that subtlety

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
mdietze: Ok, that's not clear from the documentation you wrote. I'd recommend updating to make it clear that this only affects the output processing, not the execution.

I expect/hope the need for this flag to be deprecated in the not-to-distant future, since we really should be doing the output post-processing and analysis on the fly (e.g. in the Shiny apps) rather than requiring that it all be specified up front.

models/ed/inst/ED2IN.rgit_istfer
```diff
@@ -1292,7 +1292,7 @@ $ED_NL
    !---------------------------------------------------------------------------------------!
    ! IOPTINPT -- Optimization configuration. (Currently not used)                          !
    !---------------------------------------------------------------------------------------!
-   NL%IOPTINPT = ''
+   ! NL%IOPTINPT = ''
```
ashiklom: I would just delete this block (including the comment above it) so it's not confusing. I think this has been dropped from the ED2 mainline.

mdietze: You'll want to verify that it has been completely removed from the ED2 mainline, not just that it's currently not used, as FORTRAN namelists are not flexible to missing values the way that XML and JSON are -- if a tag is missing in the namelist it will typically throw a cryptic error by misparsing the next or previous tag.

ashiklom: Considering I've had multiple ED runs fail on me recently because the tag _was_ in there when ED wasn't expecting it, I think it's safe to say it's gone.

istfer: Yes, it has been dropped from the mainline (it actually happened in my PR) I was keeping it just in case I need to go back and do some tests with older ED2 versions, but I can also delete it

base/utils/data/standard_vars.csv
istfer: hmm `dbhsize` is something like `c(10,20,30,40,50)` referring to the break points which are in cm, then DBH is the average of trees that are falling into that bin, so it's also cm.

btw, actually the read.output code is currently not compatible with anything other than `dbh_breaks = 0` and I don't think anyone other than me will use it in the near future but I put it there any way to prepare for future use (and because Ryan already implemented it). I will revisit this idea when pushing the ESM code

mdietze: Why is the average DBH within a DBH bin something interesting/useful to output? What does this tell you? I don't think I've ever seen anyone report this before. Reporting stem density by DBH bin, by contrast, does seem potentially useful because that would give you a size distribution. Growth and mortality by size would also be useful. But I think you need a better justification for why we need size by size.

istfer: IIRC we were rebinning to common DBH bins to be able to assimilate Jim's growth estimates by size class.. can discuss tomorrow

istfer:
I am going to punt the implementation of `dbh.breaks` until it's actually needed. The code currently reads only two variables of ED's monthly outputs, but the functionality is there for who wants to expand that list. 

When other parameter functional groupings such as `soil` is in the xml, code doesn't break. But it doesn't read any variable that has `soil` as a dimension yet. Which means, soil-SA will be run on totals for that pft-specific variable (the default behaviour). For example, when pft-specific SA is requested on `DBH`, which ED outputs per pft, for soil quantile-runs the total DBH will be read as sensitivity output whereas for temperate.EH quantile-runs only the temperate.EH DBH values will be read as sensitivity outputs. 

I was hesitant to add further functionality for other parameter functional groupings at this point because `read.output` is a very generic and agnostic function (and I don't think anyone need it for now?) Maybe in the future we might think of more than one type of `read.output` function?

So the PR should be good to go now.

However, I want to remind that this issue #1795 still persists. I think tower file values are being shifted for one time step by ED. It's a little more subtle when runs are stared mid-year. But the one below is an example from a run that starts from `2005/01/01`. An `analysis-T-2004.h5` file is also being written by ED which has 1 value that belongs to `2005`. But  `model2netcdf.ED2` naturally considers `start.date` for deciding which files to read so it ignores that value in the  `analysis-T-2004.h5` file, therefore everything is being shifted for 1 time-step (although, even the correctly read version seems to be not matching the  timing well for this particular run). This PR doesn't do anything to address this issue and continues the previous behaviour, because I think it should be resolved somewhere else, but I don't know how. Maybe people who are interested in aligning ED outputs with tower data (me, @bcow, @mccabete  etc.) should take a look at this together some time.
plot_zoom_png

tonygardella:
@mdietze If you have time today could you give this a review?

base/utils/R/read.output.R
```diff
@@ -30,7 +30,7 @@
 ##' @return vector of output variable
 ##' @export
 ##' @author Michael Dietze, David LeBauer
-read.output <- function(runid, outdir, start.year = NA, end.year = NA, variables = "GPP", dataframe= FALSE) {
+read.output <- function(runid, outdir, start.year = NA, end.year = NA, variables = "GPP", dataframe= FALSE, pft.name = NULL) {
```
mdietze: pft.name is missing from roxygen

PR #1670, opened 2017-09-18
CMIP5 local extraction fix

crollinson:
Fix a bug in dealing with missing leap year in GCM output.  Also add my local NLDAS extraction script which seems to have gotten lost somewhere along the way.

## Description

A fix to a mistake/bug when dealing with GCMs that are missing leap year.  Also added more robust functionality so that you can extract small time periods from a multi-year raw GCM file and also don't have to know up front if the model is missing leap year or not.

Both the CMIP5 and NLDAS scripts require you to have previously downloaded the raw data files and converted them to netcdf (if necessary), but otherwise retains the original file structure of the download.

## Motivation and Context

In previous versions of extract.local.cmip5 there was a mistake that improperly dealt with models missing leap year.  This wasn't a big problem for short periods of times, but when you miss leap year for 1,000 years, you end up with your date stamps eventually being off by ~6 months.  This was not good and caused some really odd patterns when we then try to combine data products.  These fixes address that bug as well as fix/implement some features to make it easier (& faster) to extract shorter time periods from large files.

## Types of changes

## Checklist:


modules/data.atmosphere/R/extract_local_CMIP5.R
```diff
@@ -40,15 +40,18 @@ extract.local.CMIP5 <- function(outfolder, in.path, start_date, end_date, site_i
   # Some GCMs don't do leap year; we'll have to deal with this separately
   no.leap <- c("bcc-csm1-1", "CCSM4")
   
+  if(scenario == "p1000" | GCM=="MPI-ESM-P") date.origin=as.Date("850-01-01")
+  if(scenario == "historical" & GCM!="MPI-ESM-P") date.origin=as.Date("1850-01-01")
+
```
ankurdesai: I think with both no.leap and date.origin, the user should also be able to pass these in optionally as well, assuming new CMIP5 (or other scenarios) or future CMIP6 runs have different configurations. This also makes it clear to someone using this function that these are things to watch out for.

modules/data.atmosphere/R/extract_local_CMIP5.R
```diff
@@ -153,10 +173,34 @@ extract.local.CMIP5 <- function(outfolder, in.path, start_date, end_date, site_i
...
       nc.time <- ncdf4::ncvar_get(ncT, "time")
-      
+
       # splt.ind <- ifelse(GCM %in% c("MPI-ESM-P"), 4, 3)
       # date.origin <- as.Date(str_split(ncT$dim$time$units, " ")[[1]][splt.ind])
- 
+      nc.date <- date.origin + nc.time
+      date.leaps <- seq(as.Date(paste0(files.var[[var.now]][i,"first.year"], "-01-01")), as.Date(paste0(files.var[[var.now]][i,"last.year"], "-12-31")), by="day")
+      # Figure out if we're missing leap dat
+      no.leap <- ifelse(is.null(no.leap) & length(nc.date)!=length(date.leaps), TRUE, FALSE)
```
mdietze: FWIW, the ifelse seems unnecessary here -- is there a reason you couldn't rewrite this as:
```
no.leap <- is.null(no.leap) & length(nc.date)!=length(date.leaps)
```

Also, be aware that @ashiklom rewrote/standardized a lot of the leap processing code throughout the rest of PEcAn.  Alexey: (1) could you check Christy's code to see if your functions would be useful here as a later PR, (2) we should probably add those functions to the "code standards" part of the documentation, else we'll all keep reinventing them

PR #1674, opened 2017-09-19
Slight tweak to file path to facilitate parallelization + cleanup

crollinson:
Slight tweak to file path to facilitate parallelization; cleanup old parallelization flags; fix consistent direction.filter naming

## Description

Previously, the way file paths were passed, trying to run things is
parallel (or at least the way I do it) got a little tricky.  It didn’t
seem much sense to have some redundant file paths, at least with my
file structure, so I think it’ll be okay.  *If this doesn't work with Pecan file structures, just reject the PR and I'll keep this branch for my personal use.*

Also, I did some clean-up along the way: 
1. Remove old flags for parallelization. These were a hold-over from my original paleon-specific workflow.  Further parallelization within the tdm family *could* be accomplished by parallelizing the computation of variables in lm_ensemble_sims, but I'm not prepared to implement those changes now.  For my purposes, I think prioritizing the number of ensemble members being fed in makes the most sense, but those working with a single input time series may feel differently.
2. Cleaned up inconsistently labeled direction.filter and added case to make sure one or the other is being used so it doesn't do forward filtering if you forgot/added an "s" or something like that.

## Motivation and Context

Need to be able to pass a list for one, but not multiple (I don't think) arguments to allow ensembles to be temporally downscaled in parallel. An example of how this is implemented can be found here: [6_predict_subdaily_met.R](https://github.com/PalEON-Project/modeling_met_ensemble/blob/master/scripts/6_predict_subdaily_met.R)

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/tdm_predict_subdaily_met.R
+++ b/modules/data.atmosphere/R/tdm_predict_subdaily_met.R
@@ -18,19 +18,18 @@
 # Parameters
 # -----------------------------------
 ##' @param outfolder - directory where output file will be stored
-##' @param in.path - path to model dataset you wish to temporally downscale
+##' @param in.path - base path to dataset you wish to temporally downscale; Note: in order for parallelization 
+##'                  to work, the in.prefix will need to be appended as the final level of the file structure.  
+##'                  For example, if prefix is GFDL.CM3.rcp45.r1i1p1, there should be a directory with that title in in.path.
 ##' @param in.prefix - prefix of model dataset, i.e. if file is GFDL.CM3.rcp45.r1i1p1.2006 the prefix is 'GFDL.CM3.rcp45.r1i1p1'
 ##' @param path.train - path to CF/PEcAn style training data where each year is in a separate file.
-##' @param direction.filter - Whether the model will be filtered backwards or forwards in time. options = c("backward", "forward")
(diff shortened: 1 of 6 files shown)


modules/data.atmosphere/R/tdm_predict_subdaily_met.R
```diff
@@ -67,6 +67,9 @@ predict_subdaily_met <- function(outfolder, in.path, in.prefix, path.train, dire
   
   n.ens <- length(ens.labs)
   
+  # Update in.path with our prefix (seems silly, but helps with parallelization)
+  in.path <- file.path(in.path, in.prefix)
+  
   # Extract the lat/lon info from the first of the source files
   fnow <- dir(in.path, ".nc")[1]
   ncT <- ncdf4::nc_open(file.path(in.path, fnow))
```
mdietze: Looking at the line added above, I think this will now do something a bit different from what PEcAn has expected up to now, as it will make the file name ```in.path/in.prefix/fnow``` while we've more typically thought of prefix as the start of the filename (e.g. ```in.path/in.prefix-year.nc```) . I'm not sure that this is necessarily 'wrong' because we also haven't been dealing with ensembles yet, and I can't recall off the top of my head whether we'd settled on ensemble number being a subfolder or part of the filename. I'm guessing this code is doing the latter.

PR #1675, opened 2017-09-19
read_web_config upgrades

mdietze:
```read_web_config``` wasn't returning the variable that @LiamBurke24 needs from it, ```dbfile_folder```, because it involved a cross-reference to output_folder, and because it had a previous kluge of only returning the first 14 lines (which was enough to get BETY info but avoided dealing with other parsing challenges)

Currently finding and replacing references is a bit of a kluge that would be nice to generalize (will probably fail if there are multiple replacements to make), but it works for the config.php I tested it on.


tonygardella:
@mdietze I'm paired with you this week.

PR #1686, opened 2017-09-25
Benchmarking documentation 

bcow:
First drafts of documentation for

Benchmarking:
- Using benchmarking through the interface
- The beginnings of documentation on clone runs

BETY:
- Creating new formats
- Creating new inputs
- Reference Runs

load.data and query.format.vars


book_source/adv_user_guide_web/Benchmarking.Rmd
```diff
@@ -0,0 +1,84 @@
...
+     <ensemble_id>1000004796</ensemble_id> 
+     <input_id>1000008121</input_id>
+     <variables>
+       <variable_id>411</variable_id> 
+     </variables>
+     <metrics>
+       <metric_id>1000000003</metric_id> 
+     </metrics>
+     <new_run>TRUE</new_run>
+   <reference_run_id>1000000050</reference_run_id>
```
mdietze: This section refers to the tag as ```<benchmark>``` while the section above says it's ```<benchmarking>```. One of these is wrong. I'd prefer the former.

This section also uses the ```<reference_run_id>``` tag without documenting it.

The text above refers to various things as 'required' that I don't think are (or at least I don't think should be) and continues to confuse the distinction between run DEFINITION and run EXECUTION.  I'd recommend having an very explicit separation of these two uses  in the documentation since it has become a major source of confusion for both users and developers and I feel the current documentation will just reinforce that confusion. Since the primary use of benchmarking is for the EXECUTION of existing benchmarks, and because few users will create their benchmarks via XML, I think that should go first. In which case, as far as I understand, we just need ```<reference_run_id>``` documented and to clarify whether a workflow that runs multiple reference runs should have ```<benchmark>``` sections (each with only one ```<reference_run_id>```, or just one section that has multiple ```<reference_run_id>```'s.  Personally, I think the latter is simpler, but if that's not how the code is set up now, then you should document the way the code is set up now.

book_source/adv_user_guide_web/Benchmarking.Rmd
```diff
@@ -0,0 +1,84 @@
...
+Below are all the possible benchmarking sections. 
+If you provide the ensemble id, then you do not need to provide the reference run id and vice versa.
+
+`<benchmarking>` settings:
+
+- `ensemble_id`: the id of the ensemble that you will be using - the settings from this ensemble will be saved in a reference run record and then `ensemble_id` will be replaced with `reference_run_id` 
+- `input_id`: the id(s) of the benchmarking data  (required)
+- `variable_id`: the id(s) of the variable(s) of interest within the data. If you leave this blank, all variables that are shared between the input and model output will be used. 
+- `metric_id`: the id(s) of the metric(s) to be calculated. If you leave this blank, all metrics will be used. 
+- `new_run`: TRUE = create new run, FALSE = use existing run (required)
```
mdietze: Seems like this tag should have a default (FALSE) rather than being required

book_source/adv_user_guide_web/Benchmarking.Rmd
```diff
@@ -0,0 +1,84 @@
...
+- When the run has completed, navigate to the workflow visualization Shiny app. 
+- Load model data
+  + Make sure that your model output is loading properly (i.e. you can see plots of your data)
+- Load benchmarking data
+  + Again make sure that you can see the uploaded data plotted alongside the model output. In the future there will be more tools for double checking that your uploaded data is appropriate for benchmarking, but for now you need to do the sanity checks by  hand. 
+- Navigate to the Benchmarking tab
+  + The first step is to register the new model run as a reference run in the database. Benchmarking cannot be done before this step is completed. When the reference run record has been created, additional menus for benchmarking will appear. 
+  + Next from the menu select
+    + The variables in the uploaded data that you wish to compare with model output. Note: this list shows all available variables regardless of whether or not there is a comparable variable in the model output. Thus if you select a variable to be benchmarked and it does not show up in the results, double check that it is also in the model output!
```
mdietze: Seems like this warning could be avoided by creating the benchmark from the existing Visualization -- i.e. we should already have verified that the data can be loaded, matched to a model variable, aligned, and the thing we want computed before we update the database to say that we want to make that thing (figure, stat, etc) persistent and repeatable.

book_source/adv_user_guide_web/Benchmarking.Rmd
```diff
@@ -0,0 +1,84 @@
...
+- Navigate to the Benchmarking tab
+  + The first step is to register the new model run as a reference run in the database. Benchmarking cannot be done before this step is completed. When the reference run record has been created, additional menus for benchmarking will appear. 
+  + Next from the menu select
+    + The variables in the uploaded data that you wish to compare with model output. Note: this list shows all available variables regardless of whether or not there is a comparable variable in the model output. Thus if you select a variable to be benchmarked and it does not show up in the results, double check that it is also in the model output! 
+    + The numerical metrics you would like to use in your comparison. 
+    + Additional comparison plots that you would like to see.
+  + All these selections populate the benchmarking section of the xml witch is then saved in the same location as the original run output.
```
mdietze: which

Also, writing this out to an xml just so you can read back in seems unnecessary and overly complicated. Can't you just call the required R function directly, passing it in the required variables that you go from the pull-down menus?

book_source/adv_user_guide_web/Benchmarking.Rmd
```diff
@@ -0,0 +1,84 @@
...
+  + Currently all results are stored in the `benchmarking.output.Rdata` object which is stored in the same folder as the original model run. This loads a list  called `results`  which contains the following:
+    + `input` a list element for each input file that was used for benchmarking
+      - `bench.results`: a data frame of all numeric benchmarking scores
+      - `format`: a data frame that can be used to see how the input data was transformed to make it comparable to the model output. This involves converting from the original variable names and units to the internal pecan standard. 
+      - `aligned.dat`: a data frame of the final aligned model and input values. 
+  + All plots are saved as pdf files (with names beginning with "benchmarking.")
+
+## 2. Clone Run
+
+It may be the case that you want to exactly recreate a run that was done on a different server. Or that you want to take the settings that were used to set up a run of ED2 at Harvard Forrest and use them to set up a run at the same site but using SIPNET instead. For both of these cases, instead of recreating the runs by hand every time, we are developing a way to "clone" runs.
```
mdietze: 1) For the first sentence, why would you have to 'clone' a benchmark from a different server just to rerun it? Seems like that should work just by providing the reference run id. Admittedly at the moment it might fail if it doesn't have the required inputs or data, and can't get them automatically, but that's a completely separate issue having to do with moving files among servers.

2) In the second sentence, I'd describe the general concept (you want to repeat a run done in one model with another model) then use HF ED2 -> SIPNET as an example.

That said, this whole section seems to convolve cloning runs and cloning benchmarks. The former probably needs to be documented elsewhere and in greater detail (the text below gives no indication HOW to run 'the cloning workflow') so that this section can focus on cloning benchmarks (but can link to the more general doc on cloning runs).

book_source/adv_user_guide_web/Benchmarking.Rmd
```diff
@@ -0,0 +1,84 @@
...
+      - `format`: a data frame that can be used to see how the input data was transformed to make it comparable to the model output. This involves converting from the original variable names and units to the internal pecan standard. 
+      - `aligned.dat`: a data frame of the final aligned model and input values. 
+  + All plots are saved as pdf files (with names beginning with "benchmarking.")
+
+## 2. Clone Run
+
+It may be the case that you want to exactly recreate a run that was done on a different server. Or that you want to take the settings that were used to set up a run of ED2 at Harvard Forrest and use them to set up a run at the same site but using SIPNET instead. For both of these cases, instead of recreating the runs by hand every time, we are developing a way to "clone" runs. 
+
+The cloning workflow takes in the reference run id and imports all the saved settings from the database. This populates the entire pecan.xml file with the information necessary to do a run. Thus the benchmarking section of the pecan.xml only requires the reference run id and the input id. An ensemble id cannot be put in because the run hasn't happened yet.
```
mdietze: Pure run cloning should't require an input id.
Benchmark cloning should default to cloning all the benchmarks and their metrics.
When a specific benchmark or benchmarks to clone are specified explicitly, doesn't it make more sense to require the benchmark ID not the input ID?

book_source/adv_user_guide_web/Benchmarking.md
```diff
@@ -0,0 +1,81 @@
+# Benchmarking
```
mdietze: This file is generated on-the-fly from the Rmd version and thus shouldn't be committed.

mdietze:
Just a reminder that this PR needs a few small fixes before we pull, and we really want this documentation in Tuesday's release candidate.

mdietze:
@bcow I wanted to remind you that this PR, has been open for months and only needed minor edits. Could you spend 30 min on this so we can finish it?

mdietze:
@dlebauer there has been documentation on the pecan-specific portions of BETY for many years (eg adding models). These only get used within PEcAn and thus I think are best left in the pecan documentation

tonygardella:
@bcow Is this in progress or do you want someone to take a look at it again?

mdietze:
This isn't ready for re-revision and still needs work from @bcow.  I think this needs to be a high priority. I'd really like this settled by the end of **THIS WEEK** so that this documentation is in place before next week's conversation with the ForestGEO team.

bcow:
This is ready for re-review.

book_source/04_advanced_user_guide/02_adding_to_pecan/01_case_studies/02_add_formats.Rmd
```diff
@@ -0,0 +1,127 @@
...
+
+This information is stored in a Format record in the bety database. Make sure to read through the current Formats before deciding to make a new one. 
+
+# Creating a new Format record in BETY
+
+If the Format you are looking for is not available, you will need to create a new record. Before entering information into the database, you need to be able to answer the following questions about your data:
+
+- What is the file MIME type? We have a suit of functions for loading in data in open formats such as CSV, txt, netCDF, etc. 
+     - PEcAn has partnered with the [NCSA BrownDog project](http://browndog.ncsa.illinois.edu/) to create a service that can read and convert as many data formats as possible. 
+     - If your file type is less common or a proprietary type, you can use the [BrownDog DAP](http://dap.ncsa.illinois.edu/) to convert?
```
mdietze: This ends with a question mark, but isn't really phrased as a question. It's also not clear whether you're asking the user to convert their data, or to just check whether the data can be converted. Finally, don't forget that Formats loaded by `load_data` are a subset of Formats, since not all data is loaded via load_data. For example, model inputs need to be registered with Formats, even if PEcAn never reads the file, since the Model_Type -> Formats link is what creates the list of optional and required model inputs and allows PEcAn to find existing model input files in in the Inputs table.

book_source/04_advanced_user_guide/02_adding_to_pecan/01_case_studies/02_add_formats.Rmd
```diff
@@ -0,0 +1,127 @@
...
+
+- What is the file MIME type? We have a suit of functions for loading in data in open formats such as CSV, txt, netCDF, etc. 
+     - PEcAn has partnered with the [NCSA BrownDog project](http://browndog.ncsa.illinois.edu/) to create a service that can read and convert as many data formats as possible. 
+     - If your file type is less common or a proprietary type, you can use the [BrownDog DAP](http://dap.ncsa.illinois.edu/) to convert? 
+     - If BrownDog cannot convert your data, you will need to contact us about writing a data specific load function.
+  
+- What variables does the file contain?
+     - What are the variables named?
+     - What are the variable units?
+     - How do the variable names and units map to PEcAn variables?
```
mdietze: In the future you might provide a link to a more detailed explanation of how one goes about figuring this out. When I've done it, it's required a lot of searching the Variables table for any conceivably related variable names and keywords. Might also be a good place to remind folks to NOT create redundant variables (same meaning but different name or different units) as that will break the ability to automate model/data and data/data comparisons

book_source/03_intermediate_users_guide/03_hidden_analyses/04_benchmarking.Rmd
```diff
@@ -1,3 +1,110 @@
...
+  - `bench.results`: a data frame of all numeric benchmarking scores
+  - `format`: a data frame that can be used to see how the input data was transformed to make it comparable to the model output. This involves converting from the original variable names and units to the internal pecan standard. 
+  - `aligned.dat`: a data frame of the final aligned model and input values. 
+- All plots are saved as pdf files with names with "benchmark_plot-type_variable_input-id.pdf"
+  
+- To view interactive results, naviage to the Benchmarking Plots tab in the shiny app.
+
+
+
+### Benchmarking in pecan.xml
```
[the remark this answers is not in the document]
bcow: Should the benchmarking section of the xml just be in the pecan.xml documentation page or stay on the benchmarking page?

book_source/04_advanced_user_guide/02_adding_to_pecan/01_case_studies/02_add_formats.Rmd
```diff
@@ -0,0 +1,127 @@
+# Formats records in BETY {#NewFormat}
```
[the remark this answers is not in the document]
bcow: I did a rough search and added links to the new formats documentation in quite a few files. I think I'd like to finish off this PR and ultimately open a new one in which we systematically go through and fix all the broken links in the documentation.

book_source/03_intermediate_users_guide/03_hidden_analyses/04_benchmarking.Rmd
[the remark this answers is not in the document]
mdietze: We don't have an official policy, but some things are separate and others are not. The trend is that more complex modules (SDA, PDA) tend to keep their xml doc separate. Kinda makes sense as it keeps the file from ending up huge.

Issue #1604, opened 2017-08-24
Reconcile `dplyr` and `DBI`-based database methods


ashiklom:
## Description

The latest version of `dplyr` (and its child database package -- `dbplyr`) now allows the standard `dplyr` verbs to work with objects returned by `DBI`. In other words, it's no longer necessary to use the `src_postgres` and similar functions to create database connections to work with `dplyr`; instead, code like this should work:
```r
# Connect to database
con <- PEcAn.DB::db.open(...)
# Do dplyr stuff
tbl(con, "dbfiles") %>% select(id, machine_id) %>% <etc...>
```

## Context

See #1600 , particularly [this line](https://github.com/PecanProject/pecan/pull/1600/files#diff-3f7d8d26196fbd0b4986397299b910b6R17). This affects the `query.dplyr.R` code in `PEcAn.DB` and, I suspect, code in `PEcAn.benchmark`, `PEcAn.data.land`, and other places that use `dplyr` syntax.

mccabete:
Are we deprecating the use of passed "bety" objects being generated by src_postgress as well? 

For example query.format.vars takes a "bety" object that will throw an error when it's passed PEcAn.DB::db.open(settings$database$bety).

ashiklom:
Yes, that's exactly the idea. We need to be consistent about our database interfaces, so we should avoid using `dplyr::src_*` functions in favor of `PEcAn.DB::db.open` (which calls `DBI`-based functions).

mccabete:
Great! Do I rewrite the function to take a "PEcAn.DB::db.open(settings$database$bety)", or is there a different function PEcAn.DB to set bety equal to? 

Also, what is the difference between "bety" vs "con"?

ashiklom:
As far as I know, `bety` generally refers to things returned by `src_*` functions. The output of the `src_*` functions is a list containing `con` (a database connection, exactly the same as from `db.open` and some other information about the connection. In other words, `bety$con == con`.

Really, the change I was envisioning is to just change the `bety` argument of `query.format.vars` and other functions to `con`, and to remove any references inside those functions to `bety$con`. The `dplyr::tbl(bety, ...)` function should, with recent versions of `dplyr`, work exactly the sam as `dplyr::tbl(con, ...)`, so that's as easy as a find-replace. Otherwise, I think a lot of functions that take `bety` as an argument have a line like `con <- bety$con` somewhere near the beginning, which you can just remove once you replace the function argument.

Does that make sense?

[...]

PR #1687, opened 2017-09-25
Benchmarking Shiny app 

bcow:
First draft of the Benchmarking shiny app. 

Through the app you can:
- register a reference run
- load input data for benchmarking 
- select variables and metrics for benchmarking 

Everything is written to the settings xml - and the code technically calls the benchmarking workflow and thus should return results but is crashing at the workflow step so this section of the code is commented out.


tonygardella:
@serbinsh You're matched with @bcow this week for code review. Let me know if you don't have time for review. Also, @bcow feel free to add any other reviewers you would like to see your PR.

shiny/workflowPlot/bm.R
```diff
@@ -0,0 +1,275 @@
...
+  })
+})
+
+
+##### Setup benchmarks
+observeEvent(input$load_data,{
+  req(input$all_input_id)
+  
+  bm$metrics <- dplyr::tbl(bety,'metrics') %>% dplyr::select(one_of("id","name","description")) %>% collect()
+  bm$vars_list <- c("NPP", "LAI")
```
mdietze: I'm assuming the list is this tiny just for testing?  Isn't the list of variables already being handled by a different part of the Shiny visualization code?

shiny/workflowPlot/bm.R
```diff
@@ -0,0 +1,275 @@
...
+  output$reportmetrics <- renderText(paste(bm$bm_metrics))
+  
+  inputs_df <- getInputs(bety,c(input$all_site_id)) %>% 
+    dplyr::filter(input_selection_list == input$all_input_id)
+  output$inputs_df_table <- renderTable(inputs_df)
+  
+  config.list <- PEcAn.utils::read_web_config("../../web/config.php")
+  output$config_list_table <- renderTable(as.data.frame.list(config.list))
+  
+  bm$bm_settings$info <- list(userid = 1000000003) # This is my user id. I have no idea how to get people to log in to their accounts through the web interface and right now the benchmarking code has sections dependent on user id - I will fix this.
```
mdietze: wasn't userid among the things that you deleted back on L25-30? Seems like you should save that for here. Though I guess that's not a perfect solution because that's the userid of the person that did the run, not the person that set it up as a benchmark.

shiny/workflowPlot/bm.R
```diff
@@ -0,0 +1,275 @@
...
+    bm$bm_settings$benchmarking <- append(bm$bm_settings$benchmarking,list(benchmark = benchmark))
+  }
+  
+  output$calc_bm_button <- renderUI({})
+  output$print_bm_settings <- renderPrint(bm$bm_settings)
+  
+  basePath <- dplyr::tbl(bety, 'workflows') %>% dplyr::filter(id %in% bm$ens_wf$workflow_id) %>% dplyr::pull(folder)
+  
+  settings_path <- file.path(basePath, "pecan.BENCH.xml")
+  saveXML(PEcAn.utils::listToXml(bm$bm_settings,"pecan"), file = settings_path)
```
mdietze: This may be useful for debugging, but ultimately I don't think you need to write this out. Can't you just call define_benchmark(bm$bm_settings) right here?

shiny/workflowPlot/bm.R
```diff
@@ -0,0 +1,275 @@
...
+  bm$settings_path <- settings_path
+  
+  bm$calc_bm_message <- sprintf("Benchmarking settings have been saved here: %s", bm$settings_path)
+  
+})
+
+# This sources the benchmarking workflow, which at the moment seems simpler than pulling functions out of it.
+# observeEvent(bm$settings_path,{ #Mostly so this is in a separate section which makes it asier to debug.
+#   settings <- PEcAn.settings::read.settings(bm$settings_path)
+#   source(system.file("scripts/benchmark.workflow.R",package = "PEcAn.benchmark"))
```
mdietze: As per previous comment, I really think it's a better design to just call the functions that register benchmarks directly, rather than calling benchmark.workflow.R.

shiny/workflowPlot/server.R
```diff
@@ -246,6 +247,9 @@ server <- shinyServer(function(input, output, session) {
     # Not able to add icon over ggplotly
     # add_icon()
   })
+  
+  source("bm.R", local = TRUE)
```
mdietze: would be good to have a bit of doc here explaining what the source to bm.R is doing

shiny/workflowPlot/bm.R
bcow: Agreed.
Eventually will the benchmarking workflow file itself be deprecated? Once the shiny app is done and cloning is implemented, there is no need for anyone to actually use the workflow. 
That being said, I want a version of everything that works outside of shiny (just in case we decide not to use shiny at some point...)

modules/benchmark/R/define_benchmark.R
```diff
@@ -41,71 +41,103 @@ define_benchmark <- function(settings, bety){
...
+      var_ids <- benchmark$variable_id
+    }else{
+      format <- PEcAn.DB::query.format.vars(bety, input.id = benchmark$input_id)
+      var_ids <- format$vars$variable_id
     }
     
+    # Create benchmark records using all metrics
+    # Unless individual metrics are specified in settings
+    # This can be expanded to "suits" of metrics 
+    # (for example, all plots or all regression based metrics)
```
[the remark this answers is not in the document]
bcow: Ok I see where I misunderstood.

modules/benchmark/R/metric_Frechet.R
```diff
@@ -1,12 +1,15 @@
...
 ##' 
 ##' @author Betsy Cowdery
 
-metric_Frechet <- function(dat, ...) {
-  PEcAn.logger::logger.info("Metric: Frechet Distance")
-  Fdist <- SimilarityMeasures::Frechet(as.matrix(dat$obvs), as.matrix(dat$model))
+metric_Frechet <- function(metric_dat, ...) {
+  logger.info("Metric: Frechet Distance")
+  dat.no.na <- na.omit(metric_dat)
+  Fdist <- Frechet(as.matrix(dat.no.na$obvs), as.matrix(dat.no.na$model))
```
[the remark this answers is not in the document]
bcow: I'm not entirely sure I understand this question. I can explain the circumstance under which I had to include na.omit - which may actually be opening a bigger can of worms than it initially looks like. Maybe we can talk about this one in person.

shiny/workflowPlot/bm.R
bcow: Yes that's a typo - it should have been deleted.

shiny/workflowPlot/bm.R
mdietze: I think it's fine to have the script stick around in /inst for folks who, for whatever reason, want to add benchmarks from the command line. May provide a handy example if we needed, for example, to set up a large multi-site benchmark (e.g. all of FLUXNET) to test if models capture certain emergent across-site patterns.

modules/benchmark/R/bm_settings.R
```diff
@@ -0,0 +1,108 @@
...
+  }
+  return(settings)
+}
+
+##------------------------------------------------------------------------------------------------##
+##' @name add_workflow_info
+##' @title Move benchmarking settings back in to original pecan settings object
+##' @param settings settings or multisettings object
+##' @export 
+##' @author Betsy Cowdery
```
mdietze: documentation doesn't match function

shiny/workflowPlot/server.R
```diff
@@ -18,6 +18,9 @@ options(shiny.maxRequestSize=100*1024^2)
 # Define server logic
 server <- shinyServer(function(input, output, session) {
   bety <- betyConnect()
+  source("workflowPlot_fcns.R", local = TRUE) # Load all functions that need to be dfined for this script
```
mdietze: "defined"

modules/benchmark/R/define_benchmark.R
```diff
@@ -41,71 +41,103 @@ define_benchmark <- function(settings, bety){
...
       # bm.settings$ensemble_id <- NULL
       
-    }else{logger.error("Cannot find or create_BRR")}
-  } 
+    }else{logger.error("Cannot find or create benchmark reference run")}
+  }else{logger.debug("Reference run already created")}
   
   
   # Retrieve/create benchmark entries
+  which.bm <- which(names(bm.settings) == "benchmark")
```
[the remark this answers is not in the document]
bcow: Agreed. I'm going to punt this one to my next pass through the benchmarking code.

shiny/workflowPlot/bm.R
bcow: Yes - right now I'm keeping it for debugging but it's not necessary.

shiny/workflowPlot/bm.R
```diff
@@ -0,0 +1,275 @@
...
+
+observeEvent(input$load,{
+  req(input$all_run_id)
+  ids_DF <- parse_ids_from_input_runID(input$all_run_id)
+  button <- FALSE
+  if(nrow(ids_DF) == 1){
+    
+    # Check to see if the run has been saved as a reference run 
+    # (note this is straight from the beginning of create.BRR and could 
+    # potentially be turned in to its own mini function)
```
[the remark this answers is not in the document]
bcow: Now a new function: clean_settings_BRR

modules/benchmark/R/bm_settings.R
```diff
@@ -0,0 +1,108 @@
...
+
+read_settings_BRR <- function(settings){
+  
+  # Check database connection
+  if (is.null(settings$database$bety)) {
+    PEcAn.logger::logger.info("No databasse connection, can't get run information.")
+    return (settings)
+  }
+  
+  bety <- dplyr::src_postgres(dbname   = settings$database$bety$dbname,
```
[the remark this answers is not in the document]
bcow: This ties back into issue #1604 - we really shouldn't be using `dplyr::src_postgres` now that the new version of `dplyr` is compatible with DBI objects. In other words, we should move back to using `PEcAn.DB::db.open` and `PEcAn.DB::db.close`.
I'll make these changes in the next pass of the benchmarking code. 

Meanwhile, `dplyr` connection objects close when the object is removed. So there are no explicit disconnect functions. So I don't think there should be a problem of stray open connections created using `dplyr::src_postgres`.

modules/benchmark/R/bm_settings.R
mdietze: still not fixed

shiny/workflowPlot/server.R
mdietze: not fixed yet

shiny/workflowPlot/server.R
mdietze: not fixed yet

PR #1689, opened 2017-09-26
Fix PDA Gaussian likelihood precision parameter draw

istfer:
The draw for `tau` was wrong.

## Description

`tau` parameter was previously drawn like this:
```
rgamma(1, 0.001 + n[k]/2, 0.001 + error.stats[k]/2)
```
however, the error.stats for the data I'm using (Soil Respiration) is way smaller than `0.001` (e.g. `5.120905e-15`). As a result `tau` was swamping the likelihood, and `SoilResp` data was not informing the PDA at all (hence, I was still getting constraints from NEE and LE, and nothing was jumping out at me until I checked post-PDA model soil respiration predictions in more detail) 

now `tau` is drawn like this:
```
rgamma(1, n[k]/2, error.stats[k]/2)
```
is this correct? should it be wider @mdietze?

this fails when emulator suggests a negative SS (error.stats), therefore I had to force emulator to estimate positive values for Gaussian likelihood SS (which shouldn't be negative anyway)

Now I get more contribution from `SoilResp` to the overall likelihood, and still able to retrieve back true values from a test with synthetic data.

But also, now in tests with real data, parameters are less constrained. Also, still, post-PDA performance is worse for SoilResp. I think parameter values which improve NEE and Qle in SIPNET, doesn't work for SoilResp. Anyway, need to discuss with @mdietze


mdietze:
@istfer yeah, you just rediscovered one of the two "gotchas" of the gamma prior -- assuming a default 'uninformative' prior SS that is in fact highly informative. FYI the other arises when n is too small and the distribution switches from unimodal to something that looks more exponential.

Now, the problem with your solution, setting the prior to gamma(0,0), is that it's an improper prior. What we've done elsewhere (meta-analysis) is to scale the prior SS lower if we can detect that the units of the the data are going to give a very low SS https://github.com/PecanProject/pecan/blob/develop/modules/meta.analysis/R/run.meta.analysis.R#L92

istfer:
not sure I'm following
```
tau <- rgamma(1, a, b)
```
so, should the `a` remain to be `0.001 + n[k]/2`? 

and how do you suggest I calculate the `b`? like `0.001 * mean(soil_data_obs) ^ 2`

In my case, `mean(soil_data_obs)` is something like `5e-08`,  with this calculation `b` would be `2.5e-18`, and `tau` would be around `[2.8e+18, 1.4e+19]`, does that sound about right?

mdietze:
So the Gibbs sampling of tau arises from the conjugacy of a gamma(a,b) prior and a Normal likelihood, that gives the posterior of gamma(a + n/2, b + SS/2), which is conditional on all the other model parameters that give rise to the SS. Which is why I said gamma(n/2,SS/2) implied that the prior was gamma(a=0, b=0).

Given the range you report for the SS, yes the proposed prior looks to be in the correct range. Now that the SS >> b, the SS should dominate the posterior since the implied prior sample size (2*a) is tiny. The prior of b is chosen based on the mean of the data, but not it's variance, so it's probably a fair grey area for not being double dipping (since the priors shouldn't be chosen based on the data, but here we're essentially just scaling our prior by the units of the data, and could in theory get back to a data-invariant prior specification simply by changing the __units__ of the data appropriately)

istfer:
thanks so much for the explanation, now I think I follow.

how does the implementation I suggest look?
```
        get_order <- log10(error.stats[k])
        # tau prior : gamma(a, b)
        a  <- 1e-3
        b  <- 1e-3 * (10^get_order) # scale prior, make SS >> b
        tau <- rgamma(1, a + n[k]/2, b + error.stats[k]/2)
```
I'll re-run some tests

mdietze:
Two points:
1) I think you missed an important subtlety in what I tried to explain earlier. Setting the scale of the prior based on the mean of the data isn't double dipping. Setting the prior based on the SS is most definitely double dipping. I'd strongly suggest your previous plan of setting ```b = 0.001 * mean(soil_data_obs) ^ 2```. Or, better yet, make ```a``` and ```b``` parameters input to the function by the user (e.g. via settings), but allow it to default to the value specified.
2) the actual code you implemented
```
        get_order <- log10(error.stats[k])
        b  <- 1e-3 * (10^get_order)
```
simplifies to
```
b <- 1e-3*error.stats[k]
```
which is much more clear and doesn't require the calculation of a log of something that you immediately then just exponentiate.

PR #1690, opened 2017-09-26
Check for Rpreles and Maeswrap packages

tonygardella:
Made Rpreles and Maeswrap suggested packages in PEcAn model packages so that if they are not available package building will not fail

## Types of changes

## Checklist:

### DIFF
--- a/models/maespa/R/write.config.MAESPA.R
+++ b/models/maespa/R/write.config.MAESPA.R
@@ -24,8 +24,11 @@
 ##' @author Tony Gardella
 ##-------------------------------------------------------------------------------------------------#
 write.config.MAESPA <- function(defaults, trait.values, settings, run.id) {
-  
-  library(Maeswrap)
+
+  if(!require("Maeswrap")){
+    logger.severe("The Maeswrap package is not installed. 
+                  Please consult PEcAn documentation for install notes")
+  }
   
   # find out where to write run/ouput
   rundir <- file.path(settings$host$rundir, as.character(run.id))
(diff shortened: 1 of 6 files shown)


models/maespa/R/write.config.MAESPA.R
```diff
@@ -24,8 +24,11 @@
 ##' @author Tony Gardella
 ##-------------------------------------------------------------------------------------------------#
 write.config.MAESPA <- function(defaults, trait.values, settings, run.id) {
-  
-  library(Maeswrap)
+
+  if(!require("Maeswrap")){
+    logger.severe("The Maeswrap package is not installed. 
+                  Please consult PEcAn documentation for install notes")
```
mdietze: would be better if it included the link to the relevant page in the documentation

models/preles/R/runPRELES.jobsh.R
```diff
@@ -18,6 +18,10 @@
 ##' @author Tony Gardella, Michael Dietze
 runPRELES.jobsh <- function(met.file, outdir, parameters, sitelat, sitelon, start.date, end.date) {
   
+  if(!require("Rpreles")){
+    logger.severe("The Rpreles package is not installed. 
+                  Please contact the PEcAn team to obtain source code")
```
mdietze: likewise, provide link to documentation and/or instructions on HOW to contact the PEcAn team. That said, if we have permissions to redistribute we should just provide a weblink. If we don't have permissions, they should be contacting the model team not us.

mdietze:
More general question: for purposes of Travis we want to verify that all the model packages build, but for user purposes we might not want ```PEcAn.[model]``` to be required for any of the models within PEcAn.all. Should we consider not building the model packages by default (e.g. so you'd have to call ```make models``` explictly or name individual models) or having simple make instructions for someone who just wants to work with single models, e.g. 
```
make base
make modules
make ./install/models/sipnet
```

Issue #1630, opened 2017-09-07
Create a Data Base clean up script/ set of funcitons


mccabete:
It would be good to have a script or set of functions or both that round up records that need to be fixed. This could apply to: 

inputs with no file records
file-formats without files 
etc. 

This  could just be some queries that puts records into a generic readable file that could be hand edited, and then re-fed into a function that would cull records that remain.

mccabete:
@tonygardella What DB errors do we tend to get?

ashiklom:
Let's make an editable list! I'll start a checklist here -- feel free to edit this comment (pencil button in the top right of the issue) to add stuff, or check stuff off that has been implemented.

* [ ] Orphan `dbfiles` records -- i.e. `dbfiles` records with no `input`, `model`, etc. records 
* [ ] Childless `input` records -- i.e. `input` records that don't have corresponding `dbfiles` records
* [ ] `formats` records that don't have any matching `dbfiles`

dlebauer:
* constraints can be used to prevent orphaned / childless records. 
* There are a bunch of cleanup tasks based on constraint violations listed here: https://github.com/PecanProject/bety/issues?utf8=✓&q=is%3Aopen%20%22needed%20cleanup%22%20

ashiklom:
Ankur -- Function for a given site and met product that would purge everything except original data download.

mccabete:
Mike- A function that takes away old runs, and especially met made by old runs

tonygardella:
It use to be that failed downloads of raw met data created a db record even though actual files are not there. For example if a site only had 2000- 2010 and you ran 1998-2002 it would update the input record to say 1998-2002 was there and only download 2000-2002. No sure how to fix this mismatch other than to delete old runs and their associated file as much as possible.

ashiklom:
> failed downloads of raw met data created a db record even though actual files are not there

Input records, especially met records, should have a start date, end date, and format specification, right? Perhaps we can use the format record and `load_data` functionality to get the actual bounds of the data, compare them to the start and end date, and if there is a mismatch, then flag the file for deletion? Similarly, would it be too aggressive to delete any input file that is missing any of these specifications? I.e. Are there circumstances where it's OK for an input to be missing a start date, end date, or format?

dlebauer:
@ashiklom soil datasets often do not have a start and end date

tonygardella:
[...]

PR #1700, opened 2017-09-28
Database cleanup

mccabete:
This is a template script for a Pre-release database cleanup script. It contains functions that query the database for "orphaned" records, a function that reads out database queries into a hand-editable-file, and a function that will delete all remaining records in that file. 

It also contains documentation for how additional functions could (and should!) be added to the pre-release script. 

## Motivation and Context
It partially resolves issue #1630. The idea being that several bugs are caused by records that were created incorrectly. It's also probably good practice to round up database records that aren't being used/ aren't connected to other records in a way that allows them to  be used. 

It's worth noting that the current script does not find or delete the records that were causing met bugs.  It also doesn't find all of the possible "orphaned" files. 

## Types of changes

## Checklist:


tonygardella:
Are these function meat to be used exported, as in be called `PEcAn.qaqc::function`? If yes, you'll want to add `@export ` in the headers of your functions.

tonygardella:
I'm paired with you for code review this week.

base/qaqc/vignettes/Pre-release-database-cleanup.Rmd
```diff
@@ -0,0 +1,63 @@
+---
+title: "Pre Release database Cleanup"
+output: html_notebook
+---
+
+This is an quick scriptfor cleaning up the database. For further documentation see the README.rmd file in the main `qaqc` folder.
```
mdietze: script for

base/qaqc/vignettes/Pre-release-database-cleanup.Rmd
```diff
@@ -0,0 +1,63 @@
...
+
+inputs<-find_inputs_without_formats(con=con, created_after = "2014-01-01",updated_after =NULL, user_id = NULL)
+
+```
+
+Since just a dump of every column can be hard to read, just choose the columns that are important. 
+```{r}
+column_names<-get_table_column_names(table = formats, con = con)
+column_names$formats
+column_names<-column_names$formats[-c(4,6:8)]
```
mdietze: doing this by number seems a bit confusing -- maybe less so if you're doing it interactively. Maybe change this to by name, or split this code block in two so that you first display the list of formats and then explain explicitly _how_ you're dropping certain columns.

base/qaqc/vignettes/Pre-release-database-cleanup.Rmd
```diff
@@ -0,0 +1,63 @@
...
+column_names
+```
+
+
+```{r}
+write_out_table(table = formats, outdir = outdir, relevant_table_columns = column_names, table_name = "formats")
+write_out_table(table = inputs,outdir = outdir, relevant_table_columns =c("id", "created_at", "name"), table_name = "inputs")
+```
+
+**Step 4: Check the entries, and edit the files **
```
mdietze: Presumably one could also view the results and drop rows in R **before** calling ```write_out_table```

base/qaqc/vignettes/Pre-release-database-cleanup.Rmd
```diff
@@ -0,0 +1,63 @@
...
+This is the most important step! Navigate to the written out tables and *delete entries that should remain in the database*. If the tables are dificult to read, change what columns are retained by editing the "relevant_table_columns" parameter. 
+
+**Step 5: Cull all remaining entries in the files**
+
+This will delete all entries remaining in the file, and place a delete log in the outdir directory. 
+```{r}
+### Note: Do not run before the editing step ###
+
+#cull_database_entries(table=formats, outdir = outdir,file_name ='query_of_formats' , con=con)
+#cull_database_entries(outdir = outdir,file_name =query_of_inputs , machine_id = NULL)
```
mdietze: don't you need quotes on the file_name here?  Any why specify machine_id explicitly if it's NULL?

mccabete: machine_id is null in this case, but could be used as a parameter for a dbfile table  deletion. I set it explicitly because I thought it would be nice to demonstrate to whomever was reading my script that it was a parameter that existed. It will eventually probably be best practice to set it.

base/qaqc/R/cull_database_entries.R
```diff
@@ -15,26 +15,54 @@
 ##' 
 ##' 
 
-cull_database_entries<-function(table, outdir, file_name, con, machine_id = NULL){
+cull_database_entries<-function(table = NULL, outdir, file_name = NULL, con, machine_id = NULL, table_name = NULL){
```
mdietze: argument table_name has been added but no documented. table isn't documented either.

base/qaqc/R/cull_database_entries.R
```diff
@@ -15,26 +15,54 @@
...
-  table<-read.table(file = file, header = TRUE, sep = "|")
+  if(is.null(table)){
+    
+    if(is.null(file_name)){
+      
+      PEcAn.logger::logger.severe("If a table object hasn't been provided, a file_name must be set.")
+      
+    }else{
+      file<-paste(outdir,"/",file_name, sep = "")
+      table<-read.table(file = file, header = TRUE, sep = "|") # '|' chosen because it's unlikely to overlap with "notes" text.
```
mdietze: choice of | as delimiter seems odd and risky, plus will be a pain for folks who want to go through the table in a spreadsheet, why not just do csv?

base/qaqc/R/cull_database_entries.R
```diff
@@ -15,26 +15,54 @@
...
+  
   if("dbfile" %in% table$table_name){
     
-    table<-table[table$machine_id == machine_id] #prevents deletion of files form other databases
-    
+    if(!is.null(machine_id) && "machine_id" %in% colnames(table)){
+     
+      table<-table[table$machine_id == machine_id] #prevents deletion of files from other databases
+    }else{
+      PEcAn.logger::logger.warn("Either machine_id is set to NULL or machine_id isn't a column name in table. No subssetting by machine will occur. ")
```
mdietze: this seems like 'info' not 'warn'

base/qaqc/R/find_formats_without_inputs.R
```diff
@@ -13,53 +15,36 @@
 ##' or look at the README
 
 
-find_formats_without_inputs <- function(con, user_id = NULL, created_after = NULL, updated_after = NULL){
+find_formats_without_inputs <- function(con, user_id_code = NULL, created_after = NULL, updated_after = NULL, created_before = NULL, updated_before = NULL){
```
mdietze: addition of "code" to user_id seems superfluous. Just update the documentation to make sure it's clear that this is the _numeric_ id not the username.

base/qaqc/R/find_formats_without_inputs.R
```diff
@@ -13,53 +15,36 @@
...
+    format_command<-dplyr::filter(format_command,created_at < created_before)  
+  }
+  if(!is.null(created_after)){
+    format_command<-dplyr::filter(format_command,created_at > created_after)
+  }
+  if(!is.null(updated_before)){
+    format_command<-dplyr::filter(format_command, updated_at < updated_before)
+  }
+  if(!is.null(updated_after)){
+    format_command<-dplyr::filter(format_command, updated_at > updated_after)
```
mdietze: nice! much cleaner 👍

base/qaqc/vignettes/Pre-release-database-cleanup.Rmd
```diff
@@ -14,11 +14,13 @@ password = 'bety',
...
 outdir<-as.character(getwd()) #set as prefered out directory
-bety_backup_directory<-as.character(getwd()) #set as prefered directory for bety backup. 
+bety_backup_directory<-as.character(getwd()) #set as prefered directory for bety backup.
+
+options(scipen=999) #To view full id values
 ```
 
 **Step 2: Back up bety**
-Before any major deletion processes, it makes sense to make a back-up version of the database. 
+Before any major deletion processes, it makes sense to make a back-up version of the database. Don't skip this step.
```
mdietze: 😉

base/qaqc/vignettes/Pre-release-database-cleanup.Rmd
```diff
@@ -28,36 +30,56 @@ system(backup_dir)
...
+formats<-formats[colnames(formats) %in% column_names] #subset for easy viewing
+View(formats)
+
+formats<-formats[grep("test",formats$name),]
+View(formats)
+```
+
+** Option 2: Edit a file  **
+
+This is also the most important step! Navigate to the written out table and *delete entries that should remain in the database*. If the tables are dificult to read, change what columns are retained by editing the "relevant_table_columns" parameter.
```
mdietze: difficult

base/qaqc/R/cull_database_entries.R
mccabete: I choose | just because when I did "," I got error when I encountered comments with comma's  in them.

mccabete: like 90% sure that was why

mccabete:
@mdietze I think I hit all of your points.

PR #1705, opened 2017-09-28
Update to check_qsub function so that it waits properly

serbinsh:
bug fix


ashiklom:
I don't see any problems with it, but why exactly is it necessary?

serbinsh:
@ashiklom it wasn't seeming to wait for status, just plowing forward.  I set it to how it was before when it worked......

PR #1717, opened 2017-10-07
Release/v1.5.1

mdietze:


mdietze:
@robkooper given the plan discussed today for a 1.5.1.1 minor release, my interpretation is that this PR should be allowed to go through. Could you either approve or let me know what you think needs to be done.

PR #1724, opened 2017-10-12
Makefile: Enforce installation order without rebuilding unchanged packages

infotroph:
## Description

Streamlines the make process by changing most build rules from from "normal" prerequisites (`PEcAn.DB: PEcAn.logger`) to order-only prerequisites (`PEcAn.DB: | PEcAn.logger`).

## Motivation and Context

Because PEcAn is a collection of R packages with varying degrees of functional coupling, the dependencies between packages are mostly expressed at run time rather than build time*, and therefore the build process can be relatively atomic as long as it happens in the right order. 

The current Makefile achieves the ordering by declaring some packages as prerequisites of others, which means a change in a heavily-used package requires fully reprocessing all the downstream packages even if their code is unchanged. This works, but makes debugging the base packages very tedious because every test build takes forever. 

Fortunately, Make has a syntax for that! This patch changes most of the build rules from "normal" prerequisites ("If logger changes, DB must be rebuilt", as we'd want if compiling C headers or the like) to "order-only" prerequisites  ("logger must be installed before DB is, but don't reinstall DB every time logger changes"). In my testing with these changes applied, it is now reliably true that if I make changes in one package and rerun Make, only the package I changed is rebuilt.

*Note the one exception to "run time rather than build time": The models whose documentation uses Roxygen templates have a build-time dependency on the *installed* version of PEcAn.ModelName (c.f. #1692), so `models/template` is still a normal prerequisite to all the other models. I have tested and confirmed that changing anything in the template package will correctly result in all models being re-roxygenized and reinstalled.

## Types of changes

## Checklist:


ankurdesai:
I so ❤️  this change! Thank you.

PR #1728, opened 2017-10-18
Bd 1834 cf to model

yan130:
## Description
Use met.process to convert CF met to model.
in this case, 
 stage = (download.raw = FALSE, met2cf = FALSE, standardize = FALSE, met2model = TRUE)
met.process will pass "inputfiles" (pecan.nc or pecan.zip) to 
.met2model.local.module and convert to model specific file without write an entry to bety.

## Motivation and Context
https://opensource.ncsa.illinois.edu/jira/browse/BD-1556

## Note
failed to update man of data.atmosphere: 
Dependency package PEcAn.logger not available.


tonygardella:
@robkooper Can you take a look at this one?

tonygardella:
@robkooper Can you take a look at this one as well?

modules/data.atmosphere/R/met.process.R
```diff
@@ -234,30 +249,47 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
+                                            str_ns = str_ns,
+                                            site = site, 
+                                            start_date = start_date, end_date = end_date, 
+                                            browndog = browndog, 
+                                            new.site = new.site,
+                                            overwrite = overwrite$met2model,
+                                            exact.dates = reg.model$exact.dates,
+                                            spin = spin)
+      
+      model.id  <- met2model.result$model.id
```
robkooper: is this section not needed later in the code?

modules/data.atmosphere/R/met.process.R
```diff
@@ -109,13 +109,28 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
+    if(input_met$source == "CUSTOM"){
+      # internal file is provided
+      stage <- list(download.raw = FALSE, met2cf = FALSE, standardize = FALSE, met2model = TRUE)
+      format.vars <- list()
+      format.vars$lat <- NULL
+      format.vars$lat <- NULL
+      format.vars$lat <- NULL
+    } else {
+      if(is.null(model)){
+        stage <- list(download.raw = TRUE, met2cf = TRUE, standardize = TRUE, met2model = FALSE)
```
robkooper: you changed the code from original, is this correct to set met2model to TRUE/FALSE based model.

modules/data.atmosphere/R/met.process.R
yan130: used in the next line

modules/data.atmosphere/R/met.process.R
yan130: is.null(model) == true mean we don't need to convert to model format, we just need the internal format,

is.null(model) == false mean to do the convert old way, from XML to model.

yan130:
cannot find the original branch, will create a new one.

PR #1729, opened 2017-10-19
Jules initial condition for LAI

tonygardella:
First pass at reading in initial conditions for JULES. This only handles LAI as a uniform condition for all pfts. 

Next, more initial conditions need to be added if applicable and calculation of those initial conditions weighted by a given fractional area of the pft in a plot.

## Types of changes

## Checklist:


tonygardella:
@istfer You're partnered with me for review this week. @mdietze Let me know if the implementation of this in regards to applying the lai value across pfts needs to be different.

models/jules/R/write.config.JULES.R
```diff
@@ -585,6 +590,33 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
+        # Read in netcdf file
+        IC.nc <- ncdf4::nc_open(IC.path) 
+        #Get LAI value
+        lai <- try(ncdf4::ncvar_get(IC.nc,"LAI"))
+        #Read in default initial conditions dat file
+        ic.dat  <- file.path(local.rundir, "initial_conditions.dat")
+        ic.dat.text <- readLines(con = ic.dat, n = -1)
+        #For now if number of LAI values does not match number of pfts set them set them across
+        if (npfts!= length(lai)) {
+          ic.dat.text <- gsub("@LAI_DAT@", paste0(rep(lai,npfts), collapse = " "), ic.dat.text)
```
mdietze: The approach taken of subbing in a match against this tag will fail because most cases currently leave the tag in place. Specifically, the previous code above, and the non-TRIFFID case, as well as the case here were LAI isn't in the IC file. In addition, if the user passes in a non-default reference case this tag won't be present. Given that the text file isn't actually hard to parse, I'd ditch the tag approach.

models/jules/R/write.config.JULES.R
```diff
@@ -585,6 +590,33 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
+          writeLines(ic.dat.text, con = ic.dat)
+        }
+        
+        
+      }else{
+        PEcAn.logger::logger.error("Bad initial conditions filepath")
+      }
+      
+    }
   }
```
mdietze: Why is LAI only set for the TRIFFID case? There's no ELSE here that sets the LAI for the default JULES case.

models/jules/R/write.config.JULES.R
```diff
@@ -585,6 +590,33 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
+      if(!is.null(IC.path)){
+        # Read in netcdf file
+        IC.nc <- ncdf4::nc_open(IC.path) 
+        #Get LAI value
+        lai <- try(ncdf4::ncvar_get(IC.nc,"LAI"))
+        #Read in default initial conditions dat file
+        ic.dat  <- file.path(local.rundir, "initial_conditions.dat")
+        ic.dat.text <- readLines(con = ic.dat, n = -1)
+        #For now if number of LAI values does not match number of pfts set them set them across
+        if (npfts!= length(lai)) {
```
istfer: is `npfts== length(lai)` a case?

models/jules/R/write.config.JULES.R
```diff
@@ -554,13 +556,19 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
   ## set npft to the value needed for surface type definition
   npft <- max(c(npft, 5))
+  
+  ## Edit jules_surface_types.nml to set correct number of PFTS
 
   ## --------------------------- END PFTS ------------------------------------------
-
-  ## Edit jules_surface_types.nml to set correct number of PFTS
+  
+  ## For JULES and, Enter defaults for IC.DAT
```
mdietze: comment doesn't make sense

models/jules/R/write.config.JULES.R
```diff
@@ -554,13 +556,19 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
 
   ## Edit INITIAL_CONDITIONS.NML soil carbon LAI
+  
+  
   if(useTRIFFID){
+    
+    PEcAn.logger::logger.warning("No Initial Conditions, using defaults")
```
mdietze: Why is there a warning about no initial conditions BEFORE the check to see if there are initial conditions?

models/jules/R/write.config.JULES.R
```diff
@@ -579,12 +587,51 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
-    ic.text <- readLines(con = ic.dat, n = -1)
-    ic.text[2] <- paste(ic.text[2]," 5.0 5.0 0.5 0.5 0.5 0.2 0.2 0.2 0.2 0.2 0.0 0.0 0.0 0.0")
-    writeLines(ic.text, con = ic.dat)
+    ## also load and parse IC dat file
+      ic.dat <- file.path(local.rundir, "initial_conditions.dat")
+      ic.text <- readLines(con = ic.dat, n = -1)
+      ic.text[2] <- paste(ic.text[2]," 5.0 5.0 0.5 0.5 0.5 0.2 0.2 0.2 0.2 0.2 0.0 0.0 0.0 0.0")
+      writeLines(ic.text, con = ic.dat)
+    
+    } else if (!is.null(settings$run$inputs$poolinitcond$path)){
```
mdietze: The ```if``` here in the ```else if``` is completely redundant with the previous ```if``` (if the last check what if the path was null and was false, then you already know that this check has to be true).  A better check would be ```file.exists```

models/jules/R/write.config.JULES.R
```diff
@@ -579,12 +587,51 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
+      
+      #Read in path to standard initial conditions file
+      IC.path <- settings$run$inputs$poolinitcond$path
+      
+      ## LAI m2/m2
+      if(!is.null(IC.path)){
```
mdietze: This is now the third time that you've checked whether ```settings$run$inputs$poolinitcond$path``` is null -- completely redundant. drop this ```if``` and reindent

models/jules/R/write.config.JULES.R
```diff
@@ -579,12 +587,51 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
+        # Read in netcdf file
+        IC.nc <- ncdf4::nc_open(IC.path)
+        #Get LAI value
+        lai <- try(ncdf4::ncvar_get(IC.nc,"LAI"))
+        #Read in default initial conditions dat file
+        ic.dat  <- file.path(local.rundir, "initial_conditions.dat")
+        ic.dat.text <- readLines(con = ic.dat, n = -1)
```
mdietze: This exact code was on the other side of the ```if``` above. Should be moved to before the ```if (is.null(settings$run$inputs$poolinitcond$path))```

models/jules/R/write.config.JULES.R
```diff
@@ -579,12 +587,51 @@ write.config.JULES <- function(defaults, trait.values, settings, run.id) {
...
+  }else{ 
+  # If not TRIFFID
+  ic.dat <- file.path(local.rundir, "initial_conditions.dat")
+  ic.dat <- file.path(".", "initial_conditions.dat")
+  ic.text <- readLines(con = ic.dat, n = -1)
+  ic.dat.text <- gsub("@sthuf_dat@", paste0("0.749  0.743  0.754  0.759", collapse = " "), ic.dat.text)
+  ic.dat.text <- gsub("@t_soil@", paste0("276.78  277.46  278.99  282.48", collapse = " "), ic.dat.text)
+  ic.dat.text <- gsub("@snow_tile@", paste0("0.0  0.46  0.0  0.0  0.0  0.0  0.0  0.0  0.0", collapse = " "), ic.dat.text)
+  ic.dat.text <- gsub("@lai_dat@", paste0("5.0  4.0  2.0  4.0  1.0", collapse = " "), ic.dat.text)
+  writeLines(ic.text, con = ic.dat)
```
mdietze: Not following this code at all. If you're not using TRIFFID and don't have a pool IC file, then you **don't need to open/modify the initial_conditions.dat file at all**, so it's not clear why you're inserting a bunch of values here.  If you're not using TRIFFID and you **DO** have a pool IC file, you're ignoring that file when you should be reading it and inserting LAI values.

tonygardella:
Getting back to this in the coming weeks.

tonygardella:
Closing because I won't come back to this for at least another month.

PR #1735, opened 2017-10-24
Soil texture variable name update

mdietze:
updating code for change in soil texture variable names


modules/data.land/R/extract_soil_nc.R
```diff
@@ -113,10 +113,10 @@ extract_soil_nc <- function(in.file,outdir,lat,lon){
...
   variables <- as.data.frame(matrix(c("soil_depth","m",
                                       "soil_cec","meq/100g",
-                                      "volume_fraction_of_clay_in_soil","m3 m-3",
-                                      "volume_fraction_of_sand_in_soil","m3 m-3",
-                                      "volume_fraction_of_silt_in_soil","m3 m-3",
-                                      "volume_fraction_of_gravel_in_soil","m3 m-3",
+                                      "fraction_of_clay_in_soil","m3 m-3",
+                                      "fraction_of_sand_in_soil","m3 m-3",
+                                      "fraction_of_silt_in_soil","m3 m-3",
+                                      "fraction_of_gravel_in_soil","m3 m-3",
```
annethomas: So does the m3 m-3 unit allow for mass? It seems like it should be kg, or are we keeping it mass/volume-agnostic?

annethomas: In bety it's unitless, right?

mdietze: Good catch, I was focused on the variable names and completely forgot about the units. Yes bety is correct (unitless)

dlebauer:
Sorry for the late reply, but shouldn’t these be mass_fraction_of_XXXX_in_soil to avoid the volume vs mass ambiguity?

This appears to be the CF approach (eg search for “fraction of in” at http://cfconventions.org/Data/cf-standard-names/47/build/cf-standard-name-table.html)

Issue #210, opened 2014-09-15
clone.run.settings allows repeating of existing run with a new model


mdietze:
Proposed function to assist with benchmarking, but which could be used more generally to repeat any run that was done by one model with a new model

Inputs:
Ensemble.id of previous run
Model.id of new model
Optional: list additional input IDs

Output: settings object for new run 

Behavior: 
  Uses clean settings to make a clean copy of the setting file from a previous run
  Removes the previous model and inserts the setting for the new model
  For the run, looks up the parents of the inputs. If those inputs are required or optional for the new model then find the correct child input or generate it using convert inputs. If there is an input that will be dropped then give a warning. If there is an input that is required but not in the original model nor in the optional inputs argument then stop with an error. Optional inputs overwrite existing inputs when specified.  The new run is NOT started

mdietze:
New idea, what if we add machine.id to the function args (optional) so that we could use the function to move a settings file between servers (either for the same model or a different one)

PR #1737, opened 2017-10-27
Clone run pseudocode WIP

bcow:
This is a first pass at the pseudocode for `clone_runs` as initially outlined in #210


bcow:
We originally called it `clone_run_settings` is that the name we should use?

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+  
+  #------ Update model ------#
+  
+  settings.new$model$id <- model.new$id
+  
+  #------ Update inputs ------#
+  
+  for(tag in names(settings.old$run$inputs)){
+    
+    # Check that the given input is required for the new model
```
mdietze: I think the logic below is right, but the comment here suggests you're only matching 'required' tags, rather than converting all tags you can regardless of whether they're optional or required

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+    if(tag %in% inputs_all$tag){
+      
+      # Check lookup table to see what format input needs to be in 
+      # For example for met we need the CF standard 
+      # For IC it depends on pool vs cohort - may need special cases
+      tags_std <- read.csv(
```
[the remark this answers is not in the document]
mdietze: Also, load the look up table before the loop, not every iteration

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+      
+      if(!is.na(old.input$parent_id)){
+        parent.input <- tbl(bety, "inputs") %>% 
+          filter(old.input$parent_id == id) %>% 
+          select( one_of("id", "format_id", "parent_id")) %>% collect()
+      }else{
+        parent.input <- list(format_id = NA)
+      }
+      
+      if(old.input$format_id == new_format){
```
mdietze: At this point new_format is undefined, which suggests you should add format_id to the variables queried when you do the input_all query.

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+        # The input id points to a file that is already in the right format to be converted
+        settings.new$run$inputs[[tag]]$id  <- old.input$id 
+      }else if(parent.input$format_id == new_format){
+        # The parent id points to a file that is in the right format to be converted
+        settings.new$run$inputs[[tag]]$id  <- parent.input$id 
+      }else if(is.na(old.input$parent_id)){
+        # If the input has no parent id, assume it's raw data (is that safe???)
+        settings.new$run$inputs[[tag]]$id  <- old.input$id
```
mdietze: Not safe. An input could be in a model specific format that was directly inserted into inputs. This is becoming less common for things with a standard (e.g. met) but will be very common for things where there's a model-specific default database (e.g. soils in ED and CLM)

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+      }else{
+        # This shouldn't ever happen
+        # I assume model specific input shoud always have the standard as the parent
```
mdietze: I think this case will happen a lot. 

1) I think you should change the if's above to check both the new_format **OR** the standard format. I don't think this will change the number of cases above.

2) You could still get here if a given tag doesn't have a standard format

3) You could also get here for the IC case of going from pool to cohort. old could be a model specific pool, parent could be the std pool, but you need to go further back upstream until either you get to the raw data or the cohort std

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+      # and be commented out inthe XML, but that would be difficult
+      # to implement with a list object. I propose moivng it to an "omitted" list 
+      # that could be written out in an XML or in some sort of report
+    }
+  }
+  
+  # Check that all required inputs are in settings
+  missing_inputs <- setdiff(names(settings.new$run$inputs), inputs_req)
+  if(length(missing_inputs)>0){
+    PEcAn.utils::logger.warn("Required inputs were not specified in the old run settings")
```
mdietze: I'd also make sure to spit out a list of the missing inputs so the user / log knows that certain inputs were inferred. We might also want to think carefully about the copy-of-a-copy problem. If we clone A -> B -> C and there's some input that A had that C needs, but B didn't need it, it would have gotten lost. I don't know if we can automate those checks yet, but we should definitely be recording the provenance of cloned runs in the <info> block of the settings, so the user (and perhaps eventually the code) preserves what was cloned

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+##' @param settings settings or multisettings object 
+##' @param model model id 
+##' @param con database connection
+##' @export 
+##' 
+##' @author Betsy Cowdery 
+##' @importFrom dplyr tbl filter rename collect select 
+
+clone_run <- function(settings, model, con){
+
```
mdietze: first thing needs to be a detection of Multisettings and recursive call to papply

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+  missing_inputs <- setdiff(names(settings.new$run$inputs), inputs_req)
+  if(length(missing_inputs)>0){
+    PEcAn.utils::logger.warn("Required inputs were not specified in the old run settings")
+    # If there are obvious defaults, use them to fill in missing inputs
+    # Make sure the XML can't be run if there are missing inputs
+  }
+  
+  #------ Update PFTs ------#
+
+  old.pfts <- unlist(lapply(seq_along(pfts), function(x) settings$pfts[[x]][["name"]]))
```
mdietze: I think we want id, not name. If only name is present we probably have to look up ID since name is not unique (two different models can have PFTs with the same name)

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+  # Get all possible pfts for the model 
+  new.pfts.all <- tbl(bety,"pfts") %>% filter(model.new$modeltype_id == modeltype_id) %>% pull(name)
```
mdietze: likewise should be ID

modules/benchmark/inst/clone_run_inputs_lookup.csv
```diff
@@ -0,0 +1,15 @@
...
+
+"soil", "soil", "soil_std", NA
+
+
+"IC.raw", "IC.cohort", "raw", NA
+"IC.raw", "IC.pool", "raw", NA
+
+"IC.cohort", "IC.cohort", "cohort_std", NA
+"IC.cohort", "IC.pool", "cohort_std", NA
+
```
mdietze: read.csv will get very upset about all these blank lines.

modules/benchmark/R/clone_run.R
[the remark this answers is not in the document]
infotroph: Specifically, see 83b8d8790e4e0371baca5272e8dbf0353ae2ba79 for all the steps to lazy-loading in one place. Ping me if you get stuck or want details.

modules/benchmark/R/align_pft.R
```diff
@@ -57,26 +57,32 @@
...
-align_pft<-function(con, observation_one, observation_two, custom_table=NULL, format_one, format_two, subset_are_ok=FALSE, comparison_type="data_to_data", ...){
+align_pft<-function(con, observation_one, observation_two, custom_table=NULL, 
+                    format_one = NULL, format_two = NULL, subset_are_ok=FALSE, 
+                    comparison_type="data_to_data", ...){
 
   if(comparison_type == "data_to_model"){
     
     #align_data_to_model_pft(settings_one, observations_1)
-    PEcAn.logger::logger.severe("data_to_model alignment not yet implemented. Returning NULL.")
+    PEcAn.utils::logger.severe("data_to_model alignment not yet implemented. Returning NULL.")
```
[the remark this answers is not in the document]
bcow: Oops I went converted in the wrong direction.

modules/benchmark/R/clone_run.R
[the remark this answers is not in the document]
bcow: Which of these files do I need to write by hand and which are automatically generated? I went through and see that for standard_vars:

In DESCRIPTION
- LazyData is set to TRUE
- LazyLoad is deleted

In the data directory there needs to be 
- the data file
- and R script with the same name that loads the data file

In the package R directory:
- a dataset.R script with roxygen documentation that calls the R function

I tried writing these files and the package builds but the data isn't getting loaded. Maybe some of this should be written automatically?

modules/benchmark/R/clone_run.R
```diff
@@ -0,0 +1,115 @@
...
+  settings.old <- settings 
+  settings.new <- list()
+  
+  # Copy over settings that won't change
+  settings.new$meta.analysis <- settings.old$meta.analysis
+  settings.new$ensemble <- settings.old$ensemble
+  settings.new$run$site <- settings.old$run$site
+  settings.new$run$start.date <- settings.old$run$start.date
+  settings.new$run$end.date <- settings.old$run$end.date
+
```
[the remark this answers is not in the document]
bcow: I tried it the other way and ended up changing my mind - I'll try to explain my reasoning. The point of this function is to create a `pecan.xml`, not a `pecan.CHECKED.xml` which just means that a lot of the information that's imported from the old settings really doesn't need to be copied over because it will be automatically generated when the suit of settings functions are run (like `prepare.settings` and `check.settings`). This covers things like  `info`, `host`, `database`, `rundir`, `modeloutdir`.
Also, in some cases, like `pfts`, all the possible sub tags are filled in `pecan.CHECKED.xml` but one only needs to specify the id and the rest of the information is filled in automatically. The list of things to delete seems longer than what needs to be saved at this point. 
So we might as well start with a clean palate and only copy over the bare minimum of information and let everything else get filled in automatically, reducing the chance of errors/conflicts. 

Alternatively, we could take the approach of moving everything over and "commenting out" sections that are problematic - but I'm still not clear on how to do that well when writing to xml from a list ... I'd just rather all that other information were in a log file as we discussed.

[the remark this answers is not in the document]
mdietze: @bcow I strongly disagree with your logic here. By my count we have at least 83 xml tags defined in our documentation, and this was without anything from benchmarking (which is missing from the documentation) and the link to the PDA tags broken (@tonygardella please fix this). There are inevitably additional tags that are undocumented and more important there are new tags added constantly. The idea of being able to construct a settings from scratch is simply unsustainable — it would be too hard to ensure the new settings is constructed correctly and not missing anything. Furthermore, only a handful of those tags would need to be changed, most will be identical, so why not just copy the whole object?

I also really don’t follow why you think you can delete large sections. A pecan.xml will not run without host, database, etc and I don’t see where these come from. I also don’t see why you want to loose that info as there may be custom settings that need to be preserved and provenance that needs to be retained and augmented (ie info tag) rather than deleted. 

Finally, I don’t understand what you think needs to be commented out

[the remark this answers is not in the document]
bcow: Point taken about new tags being added regularly. I agree my approach doesn't make sense if we want to be able to keep on top of continuing changes. 

Tags like host and database can get filled in using the config list that is generated from config.php. I assumed we would want to delete them because the new cloned run could be on a new server. Those would be filled in anyway if the old settings had already been cleaned / were from a reference run. What kind of custom settings should I be looking out for? 

My thought on commenting things out was half baked and doesn't even really make sense to me either.

tonygardella:
@bcow Do you think you'll get back to his in the coming weeks?

mdietze:
@bcow at today's PEcAn meeting we decided to close this PR since no one has worked on this for 5 years and it's not anything anyone is planning on working on soon. Feel free to reopen if/when you ever get back to this.

PR #1741, opened 2017-10-30
Fix for porposing and post-processing of scaling factor in emulator

istfer:
Small fixes to the code that I used in production runs.

The handling of scaling factors was causing errors in the second round

## Types of changes

### DIFF
--- a/models/sipnet/R/write.configs.SIPNET.R
+++ b/models/sipnet/R/write.configs.SIPNET.R
@@ -287,10 +287,15 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     ### ----- Soil parameters soil respiration Q10.
     if ("soil_respiration_Q10" %in% pft.names) {
       param[which(param[, 1] == "soilRespQ10"), 2] <- pft.traits[which(pft.names == "soil_respiration_Q10")]
+      # assume soil resp Q10 cold == soil resp Q10
+      param[which(param[, 1] == "soilRespQ10Cold"), 2] <- param[which(param[, 1] == "soilRespQ10"), 2]
     }
     # soil respiration rate -- units = 1/year, reference = 0C
     if ("som_respiration_rate" %in% pft.names) {
       param[which(param[, 1] == "baseSoilResp"), 2] <- pft.traits[which(pft.names == "som_respiration_rate")]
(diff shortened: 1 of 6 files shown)


models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -287,10 +287,15 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+      # assume soil resp Q10 cold == soil resp Q10
+      param[which(param[, 1] == "soilRespQ10Cold"), 2] <- param[which(param[, 1] == "soilRespQ10"), 2]
     }
     # soil respiration rate -- units = 1/year, reference = 0C
     if ("som_respiration_rate" %in% pft.names) {
       param[which(param[, 1] == "baseSoilResp"), 2] <- pft.traits[which(pft.names == "som_respiration_rate")]
+      # default SIPNET prior of baseSoilRespCold was 1/4th of baseSoilResp
+      # assuming they will scale accordingly
+      param[which(param[, 1] == "baseSoilRespCold"), 2] <- param[which(param[, 1] == "baseSoilResp"), 2] * 0.25
     }
```
istfer: Sorry for mixing these write.configs changes into this PR but these were two assumptions I was using with SIPNET to reduce dimensionality in my PDA runs 

When we turn on the functionality for seasonally varying soil respiration in SIPNET, it uses `baseSoilRespCold` and `soilRespQ10Cold` instead of `baseSoilResp` and `soilRespQ10`, respectively, when the temperature is below some `coldSoilThreshold`

I felt like `soilRespQ10` == `soilRespQ10Cold` is not too bad of an assumption (or @mdietze?)
and I let `baseSoilRespCold` vary as 1/4th of `baseSoilResp` because the default prior ranges in SIPNET were suggesting that

These resulted in much better winter soil respiration values, and they don't affect anything when the seasonal soil respiration functionality is turned-off

mdietze: I think these were fine for you PDA, but let's not pull them into the mainline (for the sake of repeatability, possibly put within a flag that defaults off and a note referencing your paper). My strong preference is for code to not do things that are unexpected (in this case change parameters other than the ones the user specified) and that parameters stay at defaults unless specified.

Issue #1713, opened 2017-10-06
Sort and document base/utils/data


infotroph:
## Description

Part of #1681. many of the files in `base/utils/data/` do not seem to be used in any code or examples, and several are large enough to trigger warnings from `R CMD check`. The whole folder should be reviewed and cleaned up as needed.

## Completion criteria:

* [x] Remove truly unused data files
* [x] Document all remaining datasets by editing base/utils/R/datasets.R
* [x] Confirm that all existing data calls work with lazy-loading
* [x] Recompress remaining files as needed to optimize installed package size

mdietze:
Some of those (e.g. mstmip_*) can be deprecated, but I think a lot of the others are (or were) used in tests. So I'd take a close look at the tests to figure out what's 'truly unused' as well as whether certain tests could be streamlined to work with smaller datasets.

Do test datasets need to work with lazy-loading?

infotroph:
> Do test datasets need to work with lazy-loading?

No if they're moved to the test directory, yes if they stay in `data/`. When lazy-loading is on, everything in data/ is converted to Rdata files at build time.

I expect we'll want to retain the raw form of everything used to test file-loading code. For datasets that get processed into an R object before we used them, lazy-loading will probably Just Work.

infotroph:
Bumping this, because setting `LazyData: True` in #1684 puts all of these datasets into the package namespace, leading to nasty bugs like #1720, where PEcAn.utils::prior.distns was shadowing a local `prior.distns` when the workflow deleted it. 

There could well be more bugs like this. To verify that there aren't, would need to look up all the variable names stored inside `base/utils/data/*.RData` and see where else those names are used in the PEcAn code at times `PEcAn.utils` might be loaded.

infotroph:
(assigning myself, but help welcome)

infotroph:
- [ ] `base/utils/data/madata.RData` is not used anywhere. Delete, or move to `modules/meta.analysis/tests/testthat/data/` and write tests that use it?
- [ ] `base/utils/data/output.RData` is not used anywhere, AND is binary-identical to `modules/uncertainty/data/output.RData`. Delete.
- [ ] `base/utils/data/samples.RData` is not used anywhere, AND is binary-identical to `modules/uncertainty/data/samples.RData`. Delete.
- [ ] `base/utils/data/sensitivity.analysis.RData` is not used anywhere. Delete, or move to `modules/uncertainty/tests/testthat/data/` and write tests that use it?
[...]

PR #1744, opened 2017-10-31
WIP: delete outdated RData files (#1713)

infotroph:
## Description

Deletes unused files in `base/utils/data/`, to avoid polluting the lazy-load namespace and reduce installed package size. 

<del>Probably don't merge until discussion in #1713 (about fates of `mstmip_*` and `trait.*`) is resolved.</del><ins>OK to merge now, Will handle CSV files in separate PRs</ins>

Question: Does this require any documentation changes?

## Types of changes

## Checklist:


infotroph:
My reasoning in 07b527b: Existing tests don't need to do anything with the full MCMC result, they just check that we're manipulating it as expected, so no need for a 6-MB file. Since the manipulations require compatible structures in all three of `prior.distns.RData`, `post.distns.RData`, and `trait.mcmc.RData` and I didn't see an easy way to shrink the MCMC results without rerunning them, I replaced all three files with smaller versions all taken from a single new BioCro run with settings$meta.analysis$iter = 50.

Note that this also re-enables a previously skipped test in PEcAn.MA.

infotroph:
@mdietze I think this is ready to merge if it looks OK to you -- All RData files are now deleted or moved, and CSV cleanup bounced to their own issues (#1746, #1747).

PR #1758, opened 2017-11-04
maespa code failed with str_trim

robkooper:
When running a maespa example it would fail in the met2model and model2netcdf, specifically in the maesawrap code. It could not find str_trim call. Adding require(stringr) solved this problem.

Problem seems to be that Maeswrap Depends on stringr, resulting in stringr being loaded into the environment. Since PEcAn.MAESPA only imports Maeswrap it will not load stringr into the environment. Either we need to make PEcAn.MAESPA depend on Maeswrap, or require stringr.

Can somebody else see if this fails for them as well. This is the the pecan.xml file used:
```xml
<?xml version="1.0"?><pecan>
 <outdir>/data/pecan</outdir>
 <database>
  <bety>
   <user>bety</user>
   <password>bety</password>
   <host>postgres</host>
   <dbname>bety</dbname>
   <driver>PostgreSQL</driver>
   <write>TRUE</write>
  </bety>
 </database>
 <pfts>
  <pft>
   <name>eucalyptus</name>
   <constants>
    <num>1</num>
   </constants>
  </pft>
 </pfts>
 <meta.analysis>
  <iter>3000</iter>
  <random.effects>FALSE</random.effects>
  <threshold>1.2</threshold>
  <update>AUTO</update>
 </meta.analysis>
 <ensemble>
  <size>1</size>
  <variable>NPP</variable>
  <start.year>2008</start.year>
  <end.year>2009</end.year>
 </ensemble>
 <model>
  <id>1000000009</id>
  <type>MAESPA</type>
  <revision>1</revision>
  <delete.raw>FALSE</delete.raw>
  <binary>/usr/local/bin/maespa</binary>
 </model>
 <run>
  <site>
   <id>1180</id>
   <met.start>2008/02/16</met.start>
   <met.end>2009/03/20</met.end>
   <name>Botanical Garden of South Wales University</name>
   <lat>-33.880449</lat>
   <lon>151.188354</lon>
  </site>
  <inputs>
   <met>
    <source>CRUNCEP</source>
    <output>MAESPA</output>
   </met>
  </inputs>
  <start.date>2008/02/16</start.date>
  <end.date>2009/03/20</end.date>
  <dbfiles>/data/dbfiles</dbfiles>
  <host>
   <name>localhost</name>
  </host>
 </run>
</pecan>
```
Also checked in is the DESCRIPTION file used when installing Maeswrap, this moves the rgl from depends to recommended.

## Types of changes

## Checklist:

### DIFF
--- a/models/maespa/R/met2model.MAESPA.R
+++ b/models/maespa/R/met2model.MAESPA.R
@@ -174,6 +174,9 @@ met2model.MAESPA <- function(in.path, in.prefix, outfolder, start_date, end_date
     lonunits <- "W"
   }
 
+  # need to load stringr in memory for Maeswrap
+  require(stringr)
+
   ## Write output met.dat file
   metfile <- system.file("met.dat", package = "PEcAn.MAESPA")
   met.dat <- Maeswrap::replacemetdata(out, oldmetfile = metfile, newmetfile = out.file.full)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 ## [Unreleased]
 
 ### Fixes
+- Fixed issue where Maeswrap would fail complaining about missing str_trim
 - Fixed issue #1752 by updating the site.lst() function to include site.id=site$id instead of site.id=site, as site is an object not just the id
 - Update to PEcAn.ED2::met2model.ED2 to fix issue with rhdf5::h5write. Bug fix to #1742
 - Fixed write.config.xml.ED2 parsing of data/history* files
--- a/docker/DESCRIPTION.maeswrap
+++ b/docker/DESCRIPTION.maeswrap
@@ -1,0 +1,15 @@
+Package: Maeswrap
+Type: Package
+Title: Wrapper functions for MAESTRA/MAESPA.
+Version: 1.7
+Author: Remko Duursma
+Maintainer: Remko Duursma <remkoduursma@gmail.com>
+Suggests:
+    rgl
+Depends:
+    lattice,geometry,stringr,tools
+Description: A bundle of functions for modifying MAESTRA/MAESPA input files,
+    reading output files, and visualizing the stand in 3D. Handy for running
+    sensitivity analyses, scenario analyses, etc.
+License: GPL
+LazyLoad: yes
--- a/models/maespa/R/model2netcdf.MAESPA.R
+++ b/models/maespa/R/model2netcdf.MAESPA.R
@@ -24,6 +24,8 @@
 ##' @author Tony Gardella
 model2netcdf.MAESPA <- function(outdir, sitelat, sitelon, start_date, end_date, stem_density) {
   
+  # need to load stringr in memory for Maeswrap
+  require(stringr)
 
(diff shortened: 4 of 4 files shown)


models/maespa/R/met2model.MAESPA.R
```diff
@@ -174,6 +174,9 @@ met2model.MAESPA <- function(in.path, in.prefix, outfolder, start_date, end_date
     lonunits <- "W"
   }
 
+  # need to load stringr in memory for Maeswrap
+  require(stringr)
```
mdietze: @robkooper if ```stringr``` is listed as a Depends in Maeswrap, why do we need to ```require``` it here and in model2netcdf? If maeswrap is calling it's libraries incorrectly, I'd prefer to PR a fix to them than to revert our code to using ```library``` calls, which we've moved away from

robkooper: This is a nice description of the problem, following is quoted from https://stackoverflow.com/a/8638902

There is one unfortunately common exception to the advice above: if your package relies on a package A which itself "Depends" on another package B, your package will likely need to attach  A with a "Depends directive.

robkooper: So options are:
1. use the stringr as done here
2. add Maeswrap as a depends in PEcAn.MAESPA
3. work with maintainer of Maeswrap to make it so he does not have depends, but all imports.

I picked the stringr since that is the smallest change and only adds stringr to namespace. Adding the depend line seems to be overkill and adds more packages to the namespace.

Let me know which you prefer.

mdietze: So it appears that model2netcdf.MAESPA requires the Maeswrap package in order to function, and write.configs (our only truly required function for wrapping a model) requires model2netcdf, which makes Maeswrap required. Therefore Maeswrap should not be in the ```Suggests``` part of the package DESCRIPTION but in the ```Imports``` (which would be preferable to the ```Depends```).  That said, adding Maeswrap to the Imports will cause the PEcAn build to fail unless (A) Maeswrap is installed or could be installed automatically (which is not what we want UNTIL Maeswrap is updated to move rgl to Suggests) or (B) PEcAn.MAESPA is itself not required in the build. Neither are good options. @tonygardella or @robkooper have either of you been talking to @RemkoDuursma or @mdekauwe about a PR to Maeswrap moving rgl to Suggests -- that's the best solution for PEcAn, but I haven't heard whether it'd be acceptable on their side.

@robkooper it sounds like you also think Maeswrap should likewise use ```Imports``` over ```Depends```, so I guess I'm coming down in favor of your option (3) of working with Maeswrap maintainers as our Plan A.  Seems like you could test the Imports/Depends shift in your docker/DESCRIPTION.maeswrap the same way you tested moving rgl to Suggests.

robkooper: Pull request created: https://bitbucket.org/remkoduursma/maeswrap/pull-requests/1/use-imports-suggests-instead-of-depends/diff

robkooper: Used my branch of maeswrap and not changes to PEcAn and it ran successfully.

robkooper:
This is now merged into maeswrap.

PR #1166, opened 2016-11-03
Added Tobit Model

araiho:
## Description
The Tobit model was added to help with left and right censored data. Specifically, data like biomass that has positive support and true zeros. 

I tested with simulated and real data.

PR #1766, opened 2017-11-08
Adding verified tobit2space model for sda.enkf.R and small fixes

araiho:
Adding verified tobit2space model for sda.enkf.R and small fixes

## Description
@mdietze and I worked with Chris Paciorek to verify the tobit2space model with a simulation script. We need the tobit2space model to map state variables to tobit space before running the generalized ensemble filter. 

In this pull request, there are also some small changes caught while working with @istfer on calibration of LINKAGES model. 

## Motivation and Context
The model specification was incorrect in past versions
This addresses issue past pull request #1166 

## Types of changes

## Checklist:


modules/assim.sequential/R/sda.enkf.R
```diff
@@ -208,7 +208,7 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
       append = FALSE)
   
   ## start model runs
-  PEcAn.remote::start.model.runs(settings, settings$database$bety$write)
+  start.model.runs(settings, settings$database$bety$write)
```
mdietze: why'd you drop the namespace from start.model.runs?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -530,26 +548,66 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
...
           }
           
         }
-        
+         
         set.seed(0)
-        dat.tobit2space <- runMCMC(Cmcmc_tobit2space, niter = 10000, progressBar=FALSE)
+        dat.tobit2space <- runMCMC(Cmcmc_tobit2space, niter = 50000, progressBar=TRUE)
+        
+        assessParams <- function(dat, Xt, mu_f_TRUE, P_f_TRUE){
```
mdietze: Is there a reason this function is defined inline within another function like this rather than defining it outside as a stand-alone function as we typically do?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -330,6 +330,38 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
     
   })
   
+  assessParams <- function(dat, Xt, mu_f_TRUE, P_f_TRUE){
```
mdietze: You seem to have moved the definition of this function from one place within sda.enkf to a different place within sda.enkf, but haven't really explained why this function is defined within sda.enkf rather than by itself.

araiho: Should I put it in sda.enkf.R as another function or move it to its own file?

mdietze: Generally better practice to put it in its own file. Don’t forget to export and roxygenize

models/linkages/R/write.config.LINKAGES.R
```diff
@@ -155,7 +156,7 @@ write.config.LINKAGES <- function(defaults = NULL, trait.values, settings, run.i
           if ("AGEMX" %in% names(vals)) {
             spp.params[spp.params$Spp_Name == group, ]$AGEMX <- vals$AGEMX
           }
-          if ("G" %in% names(vals)) {
+          if ("Gmax" %in% names(vals)) {
             spp.params[spp.params$Spp_Name == group, ]$G <- vals$G
```
istfer: does this also need to change to 
```
             spp.params[spp.params$Spp_Name == group, ]$G <- vals$Gmax
```
then?

PR #1773, opened 2017-11-14
Small fixes for missing values in soil2netcdf and soil_params (soil_utils) in data.land

ankurdesai:
soil2netcdf relies on soil_params (in soil_utils) to estimate parameters. This fix allows 1) passing soil type instead of texture to soil2netcdf to be consistent with recent change @mdietze made to soil_params, 2) passing empty texture variables to soil_params results in them being NULL not missing, and this needs to be tested in estimating texture, and 3) some soil types (like "peat") output NULL or NA values for some variables and soil2netcdf was incorrectly writing those out to the NetCDF file. These changes fixes those, so we can make wetlands great again.

## Types of changes


mdietze:
Thanks for the tweaks

PR #1778, opened 2017-11-17
Benchmarking changes for the next release

bcow:
This pull request has lots of small changes that were starting to pile up so I wanted to open the PR before the list got too long:

I've created two new functions, neither of which contain much new code but deserve to be made in to new functions because they will need to be improved/expanded in the future.

`check_BRR`
This function should determine whether or not a run has already been registered in the database as a reference run. There is no good way of checking this right now. The settings filed in the reference_runs table is just a text field containing a cleaned XML string from a settings file. Checking whether a new XML string exactly matches any of the settings in the  table is not a good comparison, but the alternative is loading all the XML entries in and comparing them which would take a long time. Other options we have discussed is changing the settings field to jsonb - which may provide more options for comparisons. 

`bety2pecan`
This is a function for converting bety variables to pecan standard via MsTMIP standards. We may want to move away from using a lookup table to do these conversions and by making a separate function for the conversions, any additional changes will only involve changing this function instead of `query.format.vars`.
This is also where I am putting the Ameriflux hack for FC and NEE mapping to NEE in pecan standard. This is only a temporary fix. But it does bring up a larger question of what to do when variable conversions are not one to one and can result in duplicate columns in the loaded data. 

Other changes

- Fix to align_data that takes care of another date range problem
- Removing more machine specific information with clean_settings_BRR
- Reorganizing the way that benchmarking output is stored in directories (this system is more scalable, and I think more easy to navigate)
- code hardening
- improvements to the shiny app


bcow:
Is the reason other people's commits are showing up in my PR because this is a branch off of develop? Would it be better to do this as a PR to develop and then have that merged in to the release branch?

mdietze:
It looks like this pull in from ```develop``` after ```release/v1.5.2``` was forked, so there's a number of previously merged PR changes from @ashiklom and @serbinsh in here. I don't know if there's an easy way to revert those, but we need to be more careful about this in the future cause we could make a mess of things. @ashiklom and @serbinsh is there anything here that you specifically _don't_ think should go in 1.5.2 (e.g. things that might destabalize the release code)

mdietze:
And @bcow I didn't see your earlier message -- in general we're really supposed to avoid pulling ```develop``` into ```release```, we should only go the other way, in case there things in develop that shouldn't go in the release.

base/db/R/query.format.vars.R
```diff
@@ -174,3 +167,50 @@ query.format.vars <- function(bety, input.id=NA, format.id=NA, var.ids=NA) {
...
+    
+    keep <- which(vars_full$bety_name[which(vars_full$pecan_name == "NEE")] == "NEE")
+    if(length(keep) == 1){
+      discard <- vars_full$bety_name[which(vars_full$pecan_name == "NEE")][-keep]
+      vars_full <- vars_full[!(vars_full$bety_name %in% discard),]
+      dups <- unique(vars_full$pecan_name[duplicated(vars_full$pecan_name)])
+    }
+  }
+  if(length(dups) > 0){
+    PEcAn.logger::logger.warn(paste("The variable(s)", paste(dups, collapse = ", "),"are duplicated.
```
mdietze: I don't think you need to use paste within logger functions since they already do that

bcow:
That makes sense. This time around, I could branch off of release and then cherry pick my benchmarking commits over  ... unless the RTM stuff is ok to commit.

PR #1779, opened 2017-11-17
Add cohort2pool function

sbs2019:
Function convert Pecan standard veg data with DBH data into a net cdf file with aboveground biomass

## Motivation and Context
Workflow needed the ability to convert from cohort to carbon pool.

## Types of changes
New function

## Checklist:


modules/data.land/R/cohort2pool.R
```diff
@@ -0,0 +1,55 @@
...
+  path <- dirname(veg_file)
+  last_dir <- basename(path)
+  nums_id <- strsplit(last_dir,"[^[:digit:]]")
+  base_id <- nums_id[[1]][length(nums_id[[1]])]
+  suffix <- nums_id[[1]][(length(nums_id[[1]])-1)]
+  if (suffix == 1) {
+    siteid <- as.numeric(paste0(suffix, "0000", base_id))
+  } else if (siteid ==0) {
+    siteid <- as.numeric(base_id)
+  }
```
mdietze: This if/else doesn't make sense to me. 

First, why are you only building the site_id if the server is BU (this code needs to work for all servers, not just 1). There should never be server-specific hacks like this in the code

Second, why are you assuming a fixed number of digits between the suffix and base ID?  siteid should just be
```
as.numeric(suffix)*10^9 + as.numeric(base_id)
```
and that should work pretty consistently across servers

Third, the ```else if``` cannot ever be true because it's based on the siteid but the siteid is not yet defined unless the previous if was true. I think you meant suffix == 0, but as noted previously you don't need to handle this as a separate case

modules/data.land/R/cohort2pool.R
```diff
@@ -0,0 +1,55 @@
...
+  
+  #Grab DBH
+  
+  dbh <- dat[[2]]$DBH
+  if(is.null(allom_param)){
+    a = 2                        
+    b= 0.3
+  }
```
mdietze: this case DOES need an else because right now the allom_param argument is never used to define a and b

modules/data.land/R/cohort2pool.R
```diff
@@ -0,0 +1,55 @@
...
+  } else if (siteid ==0) {
+    siteid <- as.numeric(base_id)
+  }
+  
+  #Grab DBH
+  
+  dbh <- dat[[2]]$DBH
+  if(is.null(allom_param)){
+    a = 2                        
+    b= 0.3
```
mdietze: What's the origin of these default parameters? If you're using some sort of global allometry provide a citation.

modules/data.land/R/cohort2pool.R
```diff
@@ -0,0 +1,55 @@
...
+  }
+  
+  #Calculate AGB
+  
+  for (i in 1: length(dbh)) {
+    biomass = a + b*log(dbh)
```
mdietze: 1. use seq_along(dbh) not 1:length
2. most allometries are base 10
3. if you're calculating the allometry in the log domain, you need to unlog the result (i.e. ```biomass = 10^(a+b*log10(dbh))``` though do check that the ```a``` you are using is appropriate, and not actually the linear domain parameter ```biomass = a*dbh^b``` -- the default you use sure looks like this version. 
4. you need to calculate more than just AGB!  Where's foliar biomass for LAI? Roots?

modules/data.land/R/cohort2pool.R
```diff
@@ -0,0 +1,55 @@
...
+    b= 0.3
+  }
+  
+  #Calculate AGB
+  
+  for (i in 1: length(dbh)) {
+    biomass = a + b*log(dbh)  
+  }
+  
+  tot_biomass <- sum(biomass)
```
mdietze: missing division by the plot area. Or, if stem density is in the file on a tree-by-tree basis, then you want to do sum(biomass*density)

modules/data.land/R/cohort2pool.R
```diff
@@ -0,0 +1,55 @@
...
+  
+  #Prep Arguments for pool_ic function
+  
+  input <- list()
+  dims <- list(time =1)
```
mdietze: Assigning time=1 is a bit dangerous -- if any downstream function actually attempts to use the time variable it will interpret it as POSIXct (seconds since the beginning of 1970)

tonygardella:
Closed in favor of #1836

PR #1787, opened 2017-12-04
Small fixes release

robkooper:
- update version to 1.5.2
- fix how function is computed when using logger function
- enable BIOCRO in use with BrownDog

## Types of changes

## Checklist:


mdietze:
```
 M modules/data.atmosphere/man/browndog.met.Rd
```

modules/data.atmosphere/R/met.process.R
```diff
@@ -371,8 +372,22 @@ browndog.met <- function(browndog, source, site, start_date, end_date, model, di
                           startdate = start_date, enddate = end_date,
                           dbfile.name = basename(outputfile), 
                           stringsAsFactors = FALSE)
+  } else if (model == "BIOCRO") {
+    metinfo <- db.query(paste0("select mimetypes.type_string, formats.name from mimetypes, formats, modeltypes, modeltypes_formats",
+                               " where modeltype_id=modeltypes.id and format_id=formats.id and formats.mimetype_id=mimetypes.id",
+                               " and tag='met' and modeltypes.name='", model, "'"), con)
+
+    formatname <- tolower(paste0("met.", model))
+    outputfile <- file.path(folder, "browndog_generated.out") # TODO is there a better name?
```
dlebauer: Is this outputfile the file that will contain the model met data? If so, then biocromet.csv seems appropriate. But aren't met files generally per-year? Thoughts @infotroph ?

infotroph: Is the intention that Browndog should return a file whose format is identical to the output of `met2model.BIOCRO()`? That [uses 'biocromet' as the formatname](https://github.com/PecanProject/pecan/blob/develop/models/biocro/R/met2model.BIOCRO.R#L88), puts each year in its own file, and always [ends filenames with `.yyyy.csv`](https://github.com/PecanProject/pecan/blob/develop/models/biocro/R/met2model.BIOCRO.R#L43). `run.biocro` currently doesn't care about the prefix but does require yearly files and `yyyy.csv`. Though maybe it shouldn't...

Issue #1400, opened 2017-05-12
Enable PEcAn to run BioCro v>= 1.0


dlebauer:
The new version of BioCro has different inputs and outputs.

It is run thus: 

```r
library(BioCro)
Gro(initial_state              = sorghum_initial_state, 
       parameters              = sorghum_parameters, 
       varying_parameters = weather05, 
       modules                   = sorghum_modules)
```

Key changes:

1. No longer crop-specific functions
2. parameters are now two lists: initial conditions and parameters (instead of many, as before)
3. 'WetDat' changed to varying_parameters 
   * this is pretty cool because parameters can be moved from parameters to varying_parameters if, e.g., they vary.
4. Output is different. Need to check names 

I _think_ everything can be handled by just updating the `run.biocro` function to check if the version is < or >= to 1.0

### TODO

* [x] need to register new version of BioCro in models page
    * can be linked to same biocro.Rscript files
* [ ] update run.biocro to handle new parameterization

justinmcgrath:
Another big difference is that the `Gro` function does not change the time period in the weather data. `BioGro` and `willowGro` would truncate the period, and you could use `day1` and `dayn` to specifiy which days are truncated. With `Gro`, you pass only the times at which you want a solution, so you would truncate first. The number of rows in the result is always the same as the number of rows in varying_parameters. To reproduce the default behavior of `BioGro`, the function `get_growing_season_climate` is provided for convenience.

```
Gro(initial_state              = sorghum_initial_state, 
       parameters              = sorghum_parameters, 
       varying_parameters = get_growing_season_climate(weather05), 
       modules                   = sorghum_modules)
```

dlebauer:
@infotroph do you have a reproducible workflow for download,Geostreams met2CF.Geostreams and met2model.BIOCRO?

I've got this so far. 

```r
library("PEcAn.data.atmosphere")
library("dplyr")

ne <- download.Geostreams(
  outfolder="~/gstest",
  sitename="UIUC Energy Farm - NE",
  start_date="2016-03-01",
  end_date="2016-03-31")

met2CF.Geostreams(in.path = "~/gstest", 
                  in.prefix = "Clowder.UIUC Energy Farm - NE", 
                  outfolder = ".gstest/cf",
                  start_date = "2016-03-01 06:00:00", 
                  end_date = "2016-03-31 05:00:00")

```

which fails because date / time matching in the filename (is this fix on a fork/branch?)

infotroph:
[...]

PR #1788, opened 2017-12-05
Support Biocro V1

infotroph:
## Motivation and Context

Addresses #1400. 

Notable changes: 

* Uses default parameters from the BioCro package when available and no local version provided
* `settings$model$revision` is now mandatory
* In multi-year runs, all end-of-year conditions are carried over to start of next year (was a hard-coded subset)

## Types of changes

## Checklist:


infotroph:
TODO before merging:

- [ ] Update documentation and changelog
- [ ] Fix Travis failures (probably caused by non-numeric Bety record for BioCro revision)
- [ ] Remove support for coppicing when using BioCro >= 1.0 -- this is an old hack that should be handled elsewhere

models/biocro/R/run.biocro.R
```diff
@@ -51,111 +44,177 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
       day1 <- NULL
       dayn <- NULL
     }
+    if (!is.null(day1) && day1 > lubridate::yday(starti)) {
+      lubridate::yday(starti) <- day1
+    }
+    if (!is.null(dayn) && dayn < lubridate::yday(starti)) {
+      lubridate::yday(plant_date) <- dayn
```
justinmcgrath: What is `plant_date`? I would think it means the date the crop was planted, but it's being assigned the value of `dayn`, which is `config$simulationPeriod$dateofharvest` at this point.

models/biocro/R/run.biocro.R
```diff
@@ -51,111 +44,177 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
       day1 <- NULL
       dayn <- NULL
     }
+    if (!is.null(day1) && day1 > lubridate::yday(starti)) {
```
justinmcgrath: This if statement would be easier to understand as
lubridate::yday(starti) = max(day1, lubridate::yday(starti))

models/biocro/R/run.biocro.R
justinmcgrath: Also, the conditional is a check against `starti`. If this is right, it should be commented to explain it.

models/biocro/R/run.biocro.R
justinmcgrath: `starti` and `endi` are used between this assignment and their initialization. It could all be in one place.
```
lubridate::yday(starti) = max(c(day1, start.date, lubridate::ymd(paste0(yeari, "-01-01")))
```

models/biocro/R/run.biocro.R
```diff
@@ -51,111 +44,177 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
-      class(m) == "BioGro"},
-      error = function(e){FALSE})
-    if (!biocro_checks_doy && min(WetDat[,"doy"])>1) {
-      if (!is.null(day1)){
-        # Biocro calculates line number as `indes1 <- (day1 - 1) * 24`
-        indes1 <- Position(function(x)x==day1, WetDat[,"doy"])
-        day1 <- indes1/24 + 1
+    HarvestedYield <- 0
+
+    if (utils::packageVersion('BioCro') >= 1.0) {
```
justinmcgrath: The conditional flow in this loop makes it very difficult to understand. Presumably, the two paths of the conditional perform the same fundamental calculations, and could be dependent on a class instead of control flow as follows.

Define the functions
results = biocro_0_wrapper(some arguments)
results = biocro_1_wrapper(some arguments)

and then write the loop as
```pseudocode
if (BioCro < version 1) {  # This if-else statement could and should be replaced by a lookup table if you continue to add more wrappers.
   loop_function = biocro_0_wrapper
} else {
   loop_function = biocro_1_wrapper
}
for (...) {
  do prep stuff
  tmp.result = loop_function(arguments)  # Now it's straightforward function call instead of branched control flow, and there's separation of responsibility between iteration and calculation.
}
```

The purpose of the loop is then clearer, and supporting new versions then requires only writing new wrappers instead of making an ever more complex for loop.

models/biocro/R/run.biocro.R
```diff
@@ -24,22 +24,15 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
-
-    # Check that all variables are present in the expected order --
-    # BioGro accesses weather vars by position and DOES NOT check headers.
-    stopifnot(identical(colnames(WetDat), c("year", "doy", "hour", "SolarR", "Temp", "RH", "WS", "precip")))
     stopifnot(all(sapply(WetDat, is.numeric)))
-    WetDat <- as.matrix(WetDat)
-    
+    starti <- max(start.date, lubridate::ymd(paste0(yeari, "-01-01")))
+    endi <- min(end.date, lubridate::ymd(paste0(yeari, "-12-31")))
     if (!is.null(config$simulationPeriod)) {
```
justinmcgrath: This is not a good place to sort out parameters. It requires that this function have explicit knowledge about parts of the program it shouldn't need to know about. The desired parameters should be set before calling this function.

dlebauer: @justinmcgrath do you mean line 39 `stopifnot(identical(colnames(WetDat), c("year", "doy", "hour", "SolarR", "Temp", "RH", "WS", "precip")))`?

I believe that this is just an assertion to ensure that the met file meets model requirements. BioCro v 0.9x requires that a met file has this specific sequence of column names. So this could be wrapped in a conditional that only applies to BioCro v 0,9x. 

If BioCro v1.x is more flexible, there could be a different condition.

models/biocro/R/run.biocro.R
dlebauer: (but using `<-` instead of `=` per PEcAn style guide)

models/biocro/R/run.biocro.R
dlebauer: that is a good idea. +1

models/biocro/R/run.biocro.R
```diff
@@ -51,111 +44,177 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
-    if (!biocro_checks_doy && min(WetDat[,"doy"])>1) {
-      if (!is.null(day1)){
-        # Biocro calculates line number as `indes1 <- (day1 - 1) * 24`
-        indes1 <- Position(function(x)x==day1, WetDat[,"doy"])
-        day1 <- indes1/24 + 1
+    HarvestedYield <- 0
+
+    if (utils::packageVersion('BioCro') >= 1.0) {
+
+      if ("SolarR" %in% names(WetDat)) {
```
dlebauer: @justinmcgrath should this move to BioCro v1 so that it accepts ( `solar` or `SolarR`)  and (`WS` or `windspeed` ), allowing backward compatibility w/ existing weather datasets? (or does it already accept these synonyms?)

models/biocro/R/run.biocro.R
```diff
@@ -51,111 +44,177 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
-        iplant$iRhizome <- data.table::last(tmp.result$Rhizome)
-        iplant$iRoot <- data.table::last(tmp.result$Root)
-        iplant$iStem <- data.table::last(tmp.result$Stem)
-        
-        if ((i - 1)%%coppice.interval == 0) {
+        initial_values <- config$pft$initial_values
+      }
+
+      if (i > 1) { # TODO HarvestedYield is never used and coppice not applicable to all crops. Rethink?
+        if ((i - 1) %% coppice.interval == 0) {
```
dlebauer: @justinmcgrath this is logic for handling perennial crops like willow ... I implemented it in PEcAn to generalize the original approach (https://github.com/ebimodeling/biocro/blob/master/inst/extdata/willowGrotestv0.912.R) in which coppice cycles for a specific analysis are written into a script. 

I always thought this type of logic should go in the model. Does the rewrite do anything to support perennial crops, or could this logic be incorporated into the model itself?

models/biocro/R/run.biocro.R
```diff
@@ -51,111 +44,177 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
+
+      tmp.result <- BioCro::Gro(
+        initial_values = initial_values,
+        parameters = config$pft$parameters,
+        varying_parameters = WetDat,
+        modules = config$pft$modules)
+
+      # save final state as initial values for next year
+      # TODO: Some pools should NOT start at 100% of previous season --
+      # need to account for harvest, decomposition, etc
```
dlebauer: @justinmcgrath does the v1 model handle harvest, decomposition, senescence, etc.

models/biocro/R/run.biocro.R
```diff
@@ -193,19 +252,19 @@ run.biocro <- function(lat, lon, metpath, soil.nc = NULL, config = config, coppi
...
-        .fun = max),
-      dplyr::summarize_at(
-        .tbl = daily_grp,
-        .vars = c("SoilEvaporation", "CanopyTrans", map="precip"),
-        .fun = sum),
-      dplyr::summarize_at(
-        .tbl = daily_grp,
-        .vars = c(mat = "Temp"),
-        .fun = mean))
+    dplyr::summarize_at(
```
dlebauer: @justinmcgrath can you think of other model outputs that will be of interest for model validation and prediction and should be available by default by the PEcAn workflow?

e.g. is there energy balance? soil respiration?

models/biocro/R/write.configs.BIOCRO.R
```diff
@@ -1,92 +1,99 @@
...
+  if ("SLA" %in% trait.names) {
+    trait.samples$SLA = udunits2::ud.convert(trait.samples$SLA, "kg/m2", "g/cm2")
   }
-  if ("Sp" %in% trait.names) {
-    trait.samples$Sp = udunits2::ud.convert(trait.samples$Sp, "kg/m2", "g/cm2")
+
+  ## rename bety variables to match active version of biocro
+  names(trait.names) <- trait.names #looks weird, but naming this vector simplifies indexing below
+
+  name_lookup <- dplyr::tribble(
```
dlebauer: @justinmcgrath similar to weather data, does the model v1 support these parameter synonyms?

I don't think that parameter synonyms are as useful for backward compatibility as met since the parameter files are fundamentally different.

models/biocro/R/write.configs.BIOCRO.R
```diff
@@ -1,92 +1,99 @@
...
+    "Jmax", "jmax", "jmax",
+    "leaf_respiration_rate_m2", "Rd", "Rd",
+    "cuticular_cond", "b0", "b0",
+    "stomatal_slope.BB", "b1", "b1",
+    "SLA", "Sp", "iSp",
+    "growth_respiration_coefficient", "GrowthRespFraction", "growth_respiration_fraction",
+    "extinction_coefficient_diffuse", "kd", "kd",
+    "chi_leaf", "chi.l", "chil",
+    "quantum_efficiency", "alpha", "alpha"
+  )
```
dlebauer: @justinmcgrath in addition, this would be a good time to suggest any other parameters that you think would be useful to analyze in sensitivity analysis / variance decomposition or estimate via calibration (e.g. partitioning and allocation parameters?)

models/biocro/R/run.biocro.R
justinmcgrath: It will break if the wrong names are given, which is in line with the single responsibility principle. I think this is an issue because the weather data sets in BioCro do not use consistent names, so we could unify the names and simplify the code, which I'd quite like to do. If the preferred names in the data sets that Pecan uses are 'WS' and 'SolarR', and thus they cannot be renamed in the data set, then Pecan is the appropriate place for the name change, since it is the only part of the code that should know about both where the parameters come from and where they will be used. This particular spot does not seem like a great place for it though, but it's not immediately obvious to me what place would be better.

As a general solution to this problem, I would define sets of mapping between names that have the same meaning. For any new model that uses different names, provide a mapping to a set of standardized names. Then it's straightforward to write a function that performs the conversion as follows:
parameters_for_b = read(parameter_set, as='set_a').as('set_b')

The mapping could also include conversions for units. The mapping has the attractive feature of being self documenting. As long as a consistent format is used, the mapping could be provided anywhere: in the original package, or elsewhere.

models/biocro/R/run.biocro.R
justinmcgrath: Harvest isn't handled, but could be handled in the same manner as coppicing. Tissues senesce and litter accumulates. There is no decomposition, but at this point if someone has a model, it would be easy to add.

models/biocro/R/run.biocro.R
justinmcgrath: Hopefully we'll have LAI data, which would be a perfect proxy for SLA in this case, so that would be good to include. We'll have height data, and that's somewhere in the model. I'll have to get it to output it. There is no energy balance or soil respiration.

I think I would write this differently though. Presumably whatever calls this function knows what it wants from the output, so it could pass the list of names as a parameter and this function could subset based on that, or this function could return all the values, and the caller could subset it.

Even those solutions aren't ideal. Currently, this violates single responsibility. The function runs the model and summarizes the result.

models/biocro/R/write.configs.BIOCRO.R
justinmcgrath: I don't think there is any cuticular conductance. Only one parameter name is accepted though. These sorts of mappings are not part of the model, or the caller, and would be better designed as its own entity, with only a specific mapping defining how the caller and model interact.

models/biocro/R/run.biocro.R
dlebauer: > If the preferred names in the data sets that Pecan uses are 'WS' and 'SolarR', 

PEcAn doesn't care what the names are, and already has a general solution to this problem: each model wrapper has a function analogous to met2model.BIOCRO function that handles name and unit conversions with varying degrees of elegance. 

I forgot that BioCro v0.9x didn't care what the column names are. In that case, we can make the met2model.BIOCRO output provide BioCro v1.0 names (solar and windspeed) and remove this code since v0.9x won't care.

models/biocro/R/run.biocro.R
dlebauer: @justinmcgrath  great I see it here: https://github.com/ebimodeling/biocro-dev/issues/124

>  but doing that will require Pecan to examine the solution to determine how much was removed

Harvested biomass is certainly an output of interest, not only to PEcAn but to any user of BioCro for crop simulation - wouldn't it make sense for BioCro to keep track of what was removed as a separate variable 'HarvestedYield' (as I originally implemented here since it doesn't currently do so?)?

models/biocro/R/write.configs.BIOCRO.R
justinmcgrath: Ah, well here's a mapping. If you made this more general, you could use it elsewhere.

Potentially we could examine anything that's differentiable.  For now, I've changes the growth model and would be interested in these parameters.

resistance_leaf_to_stem
resistance_stem_to_grain
resistance_stem_to_root
resistance_stem_to_rhizome
rate_constant_leaf_senescence
rate_constant_stem_senescence
rate_constant_root_senescence
rate_constant_rhizome_senescence
KmLeaf_senescence
KmStem_senescence
KmRoot_senescence
KmRhizome_senescence

models/biocro/R/run.biocro.R
justinmcgrath: I think that v0.9 does use the column names in some places and indexes in other places - rather annoying. I know I changed some of the indexes to column names, but I know it didn't change everything because it was easier to just get rid of that stuff entirely.

models/biocro/R/run.biocro.R
justinmcgrath: Yes, that can be saved.

models/biocro/R/run.biocro.R
dlebauer: @justinmcgrath 

* Do you mean compute LAI on the fly from leaf biomass * SLA?
* once you export height that would be great + easy to add

> Currently, this violates single responsibility. The function runs the model and summarizes the result.

You are correct; Typically PEcAn takes whatever a model's defaut output is and converts it to the PEcAn format in model2netcdf.<MODEL> functions. For BioCro I added this step for convenience: 1) so a rectangular .csv could be written and 2) to allow daily and annual summaries that are much smaller when runs are very large (regional + met + parameter uncertainty). At this point rather than refactoring I would drop the daily and annual summaries since PEcAn (currently) only uses the hourly output. Then if / when there is a need for daily or hourly summaries it should work on the PEcAn standard format and work for all models.

justinmcgrath: LAI is one of the outputs. The name is 'lai'. It is currently calculated as leaf_mass * SLA, but I think it would be better to use the output directly.

models/biocro/R/write.configs.BIOCRO.R
dlebauer: @justinmcgrath for these could you please fill out this table so that I can add these to the table: https://docs.google.com/spreadsheets/d/1GVPcEtcj2oTMT6N_4CUK0nX3VBOjkXOLTo5Rzz4MH3Y/edit?usp=sharing

Note: 
* min and max (optional) should be the range of possible values (generally lower bound is 0, this is for crude qa/qc)
* the prior distribution, param a, param b follow R conventions in https://cran.r-project.org/doc/manuals/R-intro.html#R-as-a-set-of-statistical-tables

models/biocro/R/write.configs.BIOCRO.R
dlebauer: if I understand correctly, this is the specific mapping defining how the caller and model interact. We have discussed using standalone tables in files or the database, but these tables are generally small. I was mostly wondering if there is backward compatibility between v0,9x and v1 parameter names; sounds like there is not.

models/biocro/R/run.biocro.R
dlebauer: OK Looks like both Leaf biomass and LAI are included in the hourly output given to PEcAn (line 91 and 231 above )

models/biocro/R/run.biocro.R
infotroph: Good catch -- `plant_date` isn't a thing and that line should have been `endi`.

models/biocro/R/run.biocro.R
infotroph: The tricky bit with assigning all 3 at once is that start.date could be in a different year, in which case we don't care about its DOY.  I think it's probably better to keep these as separate calculations, but have cleaned up the logic  so it's more clear how they fit together.

models/biocro/R/run.biocro.R
infotroph: I'm afraid I don't follow either of your comments here. Can you elaborate?

* @justinmcgrath, all the parameter checks in this section are of things that might change every year, so shouldn't the code that manages the time loop be responsible for ensuring they have sane values at each step? To what function would you suggest moving them?

* @dlebauer, you're correct that the `stopifnot(...)` is an assertion to catch misformatted met files, but it *already* applies only to BioCro 0.9x -- note that in this diff it's deleted from line 39 and added at line 103, inside the BioCro 0.9 section.

justinmcgrath: I put this comment is a stupid spot. This refers to the if else statements that select day1 and dayn.

The function needs to know day1 and dayn, but it doesn't need to sort out where it's getting them from. An easier to use design is for the function to accept only one set of parameters, and the caller should decide what parameters to use. You could set day1 and dayn from `config`, get rid of the other choices, and make it an error to not provide day1 and dayn. The same reasoning applies to the use of soil.nc at the beginning.

The order of columns does not matter in the new version. Variables are always identified by their names. Not supplying a required name is an error, and it should list the missing names. There aren't built in functions in C++ for that, so sometimes it just gives a generic error if I missed something.

dlebauer: @infotroph my comment is that for BioCro v0.9, the column names do not matter, only the position does. Therefore, if there is a single BioCro met standard that uses v1.0 names and 0.9 order then this condition could just test that the dataset has v1.0 names in v0.9 order. (PEcAn never needs to know about the names WS and  SolarR that only appear in a few sample datasets, and should not be stopped if it has a dataset with v1 names).

models/biocro/R/run.biocro.R
infotroph: Fixed in 1ee5c70fb423491dd559603f0da75d0df4ba5b68

models/biocro/R/run.biocro.R
infotroph: Fixed in a3d2e5a9f84f31847672a7bf934363fbc4794da2, but I pass it a bazillion arguments  to avoid changing any of the 0.9 logic, and I'm not sure I like the resulting kitchen-sink function signature. Further thoughts?

models/biocro/R/run.biocro.R
infotroph: Column names fixed in 4604149501c4cf17b31607ea919a42fdc3b5eccc

models/biocro/R/write.configs.BIOCRO.R
infotroph: @justinmcgrath Is there a table somewhere with names and definitions of the parameters Biocro 1 recognizes? I constructed these synonyms by hand-mapping the changes between v0.9 and v1 Micanthus parameters, and could easily have missed something or gotten things wrong.

infotroph: @justinmcgrath Also, by "I don't think there is any cuticular conductance" do you mean BioCro has no input parameter for it? In that case what is b0? Do you see any others that are misidentified here?

dlebauer:
nice work. Thanks for cleaning this up. v0.9 is a mess; Kitchen sink for v0.9 is better than refactoring legacy code.

infotroph:
@dlebauer Well, it's a little worse because this design commits to kitchen sink for the V1 wrapper as well, but maybe it's good enough.

Is my decision not to export or document the wrappers a reasonable one? My hope is that nobody ever needs to call them from outside of run.biocro.

dlebauer:
Yep. No need to export

scripts/add.models.sh
```diff
@@ -25,6 +25,8 @@ addModelFile "${FQDN}" "Preles"   "PRELES"   ""          "true"           "/bin"
 
 # special case for R models
 addModelFile "${FQDN}" "BioCro"   "BIOCRO"   "git"       "biocro.Rscript" "${R_LIBS_USER}/PEcAn.BIOCRO"
+addModelFile "${FQDN}" "BioCro"   "BIOCRO"   "0.95"      "biocro.Rscript" "${R_LIBS_USER}/PEcAn.BIOCRO"
+addModelFile "${FQDN}" "BioCro"   "BIOCRO"   "1.0"       "biocro.Rscript" "${R_LIBS_USER}/PEcAn.BIOCRO"
```
infotroph: @robkooper is this change reasonable? Same script gets called for both versions, we just use the Bety version record to keep track of which formats to give it.

models/biocro/R/write.configs.BIOCRO.R
justinmcgrath: There's this out-of-date file: https://github.com/ebimodeling/biocro-dev/blob/master/data/biogro_parameters.r

Updating that is on my list of things to do.

I guess I would consider b0 and cuticular conductance to be different things, but it looks like many articles make no distinction.

models/biocro/R/run.biocro.R
justinmcgrath: Classes do this (in this specific case, a functor). Parameters specific to each model are passed upon initilization and each subclass returns a common interface. As a pseduocode example,
```
call_biocro_0.9 <- function(genus, ...) {  # ... means other parameters, not necessarily R's "the dots", but you could use them if they're convenient.
   function(WetData) {
      if (genus == 'Saccarhum') {
        tmp.result = CaneGro(...)
      } else if (genus == 'Salix') {
        tmp.result = willowGro(...)
      }
      return (tmp.result)
   }  # End of returned function.
}

call_biogro_1.0 <- function(...) {
   function(WetData) {  # This function has the same signature and return value as the one returned by call_biogro_0.9
      ...
   return (tmp.result)
   }
}

# In run.biocro
  if (utils::packageVersion('BioCro') >= 1.0) {
    caller_fn <- call_biocro_1(...)
  } else {
    caller_fn <- call_biocro_0.9(genus, ...)
  }
```

 BioCro 0.9 makes that someone annoying because of the way it picks start and end times, so you'd have subset WetDat before calling (in the same way you do for v1.0) and have the class return a function that sets day1 and dayn to the first and last days of the WetDat parameter, so you don't have to pass any time variables.

infotroph: Come to think of it... I already *do* subset WetDat before calling! So this is actually pretty straightforward to do.

justinmcgrath:
Cool. In the v0.9 version, you'll have to get day1 and dayn from WetDat and
pass it to the various functions or it will try to guess them, and it will
further subset the data.

On Dec 11, 2017 3:27 PM, "Chris Black" <notifications@github.com> wrote:

*@infotroph* commented on this pull request.
------------------------------

In models/biocro/R/run.biocro.R
<https://github.com/PecanProject/pecan/pull/1788#discussion_r156203648>:

> -    # we check if the current BioCro has fixed this assumption.
-    # If not, rescale day1 and dayn to be relative to the start of the input.
-    #   Scaling is derived by inverting Biocro's day->index equations.
-    biocro_checks_doy <- tryCatch(
-      {m <- BioCro::BioGro(WetDat = matrix(c(0,10,0,0,0,0,0,0),nrow = 1),
-                   day1 = 10, dayn = 10, timestep = 24);
-      class(m) == "BioGro"},
-      error = function(e){FALSE})
-    if (!biocro_checks_doy && min(WetDat[,"doy"])>1) {
-      if (!is.null(day1)){
-        # Biocro calculates line number as `indes1 <- (day1 - 1) * 24`
-        indes1 <- Position(function(x)x==day1, WetDat[,"doy"])
-        day1 <- indes1/24 + 1
+    HarvestedYield <- 0
+
+    if (utils::packageVersion('BioCro') >= 1.0) {

Come to think of it... I already *do* subset WetDat before calling! So this
is actually pretty straightforward to do.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/PecanProject/pecan/pull/1788#discussion_r156203648>,
or mute
the thread

tonygardella:
@dlebauer @infotroph How far out are we on getting support for biocro v1?

infotroph:
@tonygardella Probably a week or so. I've made all the code changes I know about, but need to check whether documentation is up to date and maybe add some tests.

infotroph:
I've now made all the changes that were on my list. I'm sure it could use more testing and more documentation, but what's done covers most of the changes I made.

@dlebauer Still looking OK?
@justinmcgrath Does this address all the changes you wanted? If so, can you update your review?

PR #1800, opened 2018-01-04
Met Downscaling Updates

crollinson:
Substantial updates to both the bias correction and temporal downscaling workflows. 

## Description

Both the bias correction and temporal downscaling workflows have been overhauled to achieve several improvements 

1. Fix lingering bugs that essentially doubled the anomaly in the doubling of anomalies in the bias-correction step. (see debais_met_regression.R)
2. Pass met through sanity checks, preferably trying again and/or duplicating reasonable sims to forcing hard-coded bounds.  This occurs in both the bias correction & temporal downscaling steps. (See flags for force.sanity and sane.attempts). Some of these sanity checks may need to be adjusted, particularly temperature flags if we're running with future met scenarios.
3. Adjusting rain distribution in the temporal downscaling step to make the number of rainy hours resemble what's observed in the empirical data.  NOTE: this is currently hard-bounded at not raining more than half the day. (This arose from constant drizzle causing major fails in ED.)
4. Added methods of propagating uncertainty: One option is to pull a random sub-ensemble member (should encourage ensemble spread), but another option is to just use the mean (more suitable if you don't want met ensembles).
5. Also fix some minor bugs with labeling (force 4-digit years), filtering, and met alignment.

## Motivation and Context

The largest overhauls come from discovering that some generated ensemble members left sanity behind by producing temperatures greater than typically found in the Sahara &/or Antartica at Harvard Forest and/or other quirks that caused the met not to run in ED.

Note: this version of the code generated the 1166-year, 90+ member met ensemble used for C. Rollinson & I. Fer's 2017 AGU talks.  (Ensemble generated 5 Dec 2017.)

## Types of changes

## Checklist:


modules/data.atmosphere/R/align_met.R
```diff
@@ -47,6 +47,7 @@
 ##' @param n.ens  - number of ensemble members to generate and save
 ##' @param pair.mems - logical stating whether ensemble members should be paired in 
 ##'                    the case where ensembles are being read in in both the training and source data
+##' @param mems.train - (optional) ensemble identifiers so that the training data is read in a specific order
```
mdietze: This documentation is insufficient for a user to know what they're supposed to put in here or what exactly this optional flag does. Could you provide a bit more explanation in the Details? And maybe an Example?

ankurdesai:
@crollinson I'm going to be out of the country doing fieldwork for 2 weeks. So won't be able to look in detail ASAP, but soon.

crollinson:
Sorry for the beast of the pull request -- I'll try to do a better job of making individual PR for small bugs and separate out the big overhaul.  Unfortunately in this case, the small bug fixes revealed bigger issues that needed to fixed for sane output to be produced.  Most of what won't load by default are the sanity check bounds.  At the moment, those sanity checks generally align with what's in ED2, but it might make sense to adjust them if people are running with climate change scenarios.

PR #1810, opened 2018-01-16
Benchmarking Update

bcow:
A bit of a grab bag of changes implemented to get results for AGU. 
All of this is still work in progress but should be added in before conflicting changes are made. 
I will revisit some of this code very soon in response to some of the most recent conversations on gitter/slack. 

Benchmarking related functions
- `read_output`: Added a check to keep from loading multi-dimensional variables (until we can better handle them - they should be dealt with in all the `model2netcdf` functions but this is just to prevent causing errors in visualization/benchmarking if they haven't.)
- `align_data`: Added case where a run is only one year long and also is a leap year
- `calc_benchmark`: Added  start & end year, general cleanup
- `calc_metrics`: Removes all leading and trailing NA's from the data (not the same as running `na.omit` which would remove all rows with NA's and could potentially change metric scores.)
- Changes to multiple metric functions to account for NA's

Model related function
- JULES
   - output sums over surface types
   - currently ignores soil related output because we haven't decided how to report values at multiple depths
   - additional outputs variables added to `output.nml`
- LPJ-GUESS
   - fixed time step error


base/db/R/query.dplyr.R
```diff
@@ -235,6 +234,72 @@ var_names_all <- function(bety, workflow_id, run_id) {
...
+#   globalDF$title <- as.character(globalDF$title)
+#   globalDF$xlab <- as.character(globalDF$xlab)
+#   globalDF$ylab <- as.character(globalDF$ylab)
+#   globalDF$var_name <- as.character(globalDF$var_name)
+#   return(globalDF)
+# } #load_data_single_run
+
+# Currently commenting out previous version of load_data_single_run
+# The following version ususe read.output instead of alternative code 
+
```
ashiklom: If the below code works, can you just delete this? It's safely stored in our git history if we ever need it.

mdietze: @ashiklom do you mean delete all of load_data_single_run or @bcow's globalDF changes?

ashiklom: I just mean delete the commented-out old version of `load_data_single_run` function so it isn't taking up space.

base/utils/R/read.output.R
```diff
@@ -102,6 +102,22 @@ read.output <- function(runid, outdir, start.year = NA, end.year = NA, variables
...
+    
+    for(var in names(result)){
+      c <- dim(result[[var]])[2]
+      r <- dim(result[[var]])[1]
+      if(!is.na(c) & r > 1){
+        PEcAn.logger::logger.warn("Variable", var, "has", r, "dimensions,
+      it cannot be loaded and will be omitted.")
+        result[[var]] <- NULL 
+      }
+    }
```
ashiklom: I think Mike's suggestion was that for variables that have multiple levels (e.g. Soil Moisture), to sum across the layers as a first pass (`colSums`).

bcow: @mdietze should I include that?

mdietze: I'd say that it probably does make sense for this to be consistent with the earlier PR where I recommended that switch. That said, the end goal should be the ability to safely return matrices and/or a specifically selected layer without downstream code dying

PR #1812, opened 2018-01-17
Temporary hotfix for Shiny

ashiklom:
Several potential issues:

- `PEcAn.DB::load_data_single_run`:
  - Currently can't handle matrix output (e.g. multiple layers of soil for moisture and temperature). This has a hack to only grab the top layer.
  - If variable is not found, returns a 0-length data frame, which then crashes on the `cbind`. This skips the variable with a warning if it has 0-length output.
- `shiny::server.R`
  - Shiny crashes if no inputs are available for a given site and format. This hotfix returns `NULL` if no inputs are found, which allows Shiny to draw the output, even if it can't find data to plot with.


ashiklom:
Note that in 3efe291 I added some documentation for debugging Shiny. Not sure how well this works on actual remote servers, but it works very well on the VM.

Issue #1722, opened 2017-10-11
Move workflow functions from utils to own package


infotroph:
## Description

As discussed in #1681: Move do_conversions, get.results, run.write.configs etc from PEcAn.utils to a new package possibly named PEcAn.workflow.

## Context
These are the functions that integrate calls from many other packages, so by design they depend on the rest of PEcAn. By including them here we guarantee circular dependencies with any other package that uses PEcAn.utils. Should also help resolve #1719. 

## Possible Implementation

 Functions to move (list is incomplete, suggestions needed):
* [ ] `do_conversions`
* [ ] `get.results`
* [ ] `run.write.configs`
* [ ] `ensemble`
* [ ] `get.parameter.samples`

infotroph:
Also to move: PEcAn.DB::runModule.get.trait.data (#1815)

infotroph:
While moving each function, should also review documentation to make sure it's clear what the purpose of the function is and how you're supposed to use it in a workflow.

Issue #1681, opened 2017-09-23
PEcAn.utils package cleanup


infotroph:
## Description

The utils package currently throws a LOT of messages at check time, contains a number of functions that appear to be vestigial or broken, and contains undeclared dependancies on other packages that in turn depend on utils. I propose to clean up as many of these as we can.

This is a big topic -- discuss, please. I'll link to sub-issues as we identify them.

### Delete?

* [ ] Everything in `inst/get_mstmipvars.R` -- this file looks like an old demo.
* [ ] `model2netcdfdep` and `model2netcdf` -- unused, have been marked as deprecated since 1.3.7!
* [ ] `theme_border` -- unused, relies on a hack that doesn't work in current versions of ggplot
* [ ] Are any datasets in `base/utils/data/` unused? they add a LOT to the installed size of the package

### Deprecate now, delete in a few releases?

* [ ] `counter` -- assigns to a global that nothing reads anymore.
* [ ] `create.base.plot` -- just wraps ggplot2::ggplot()
* [ ] Most or all of `status.R` will be unneeded once #1660 is completed, but is essential until then.

### Keep? 

Apparently unused right now, but look potentially useful. Perhaps they should be more visibly documented.

* [ ] `as.SafeList` 
* [ ] `as.sequence`
* [ ] `distn.table.stats`

### Move to other packages?

* [ ] date-time utils (to their own new package)
* [ ] plotting tools (to an existing package, maybe visualization?)
* [ ] file-handling tools (maybe even just NetCDF-handling tools?)

### Clean up existing code

Main goal: Reduce trivial output from `devtools::check("base/utils")` until we can tell it doesn't mask important messages

* [ ] Namespace janitorial work
      - Remove all library() calls
      - Move imported packages from Depends to Imports
      - `::` for all out-of-package functions
      - Adress any remaining `no visible binding for global...` warnings (tasks above should fix most of them)
* [ ] Identify and resolve circular dependencies 
      - Do any packages no longer depend on utils now that logger is separate?
      - `do_conversions` and `convert.input` both import a lot of other PEcAn packages. Can some of these move to Suggests instead of Depends?
* [ ] Roxygen janitorial work
    - Document all function parameters
    - Check that \usage agrees with  code
    - Provide more examples

infotroph:
[...]

PR #1817, opened 2018-01-18
Create new package PEcAn.workflow

infotroph:
...and move `PEcAn.DB::runModule.get.trait.data` into it

## Motivation and Context
As discussed in #1681, #1722. The idea here is to collect in one package the functions that Actually Do The Work™, meaning:

* Intuitively, PEcAn.workflow is for the functions that get called directly from workflow.R
* More broadly, if you'd call a function exactly once in a normal workflow and comment it out to skip some major chunk of your analysis, it probably belongs here
* Pragmatically, this package is for functions that create circular dependencies between base packages. If a function uses many different PEcAn packages to do its job, it might belong here.

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation


ashiklom:
Looks good.

PR #1823, opened 2018-01-24
Post-AGU and Tansley SDA changes

istfer:
This is the branch including my AGU changes and @mdietze's updates to the SDA code

Shouldn't affect other workflows, except there are few -nonbreaking- changes to the SIPNET code

Changes need to be in before we start refactoring SDA workflow

## Types of changes


modules/assim.sequential/R/met_filtering_helpers.R
```diff
@@ -0,0 +1,34 @@
+
+sample_met <- function(settings,nens=1){
+  
+  # path <- settings$run$inputs$met[["path"]]
+  # developing
+  path = "/fs/data3/istfer/AGU17/met_ensembles/SIPNET/"
```
mdietze: This AGU hack probably shouldn't stay as part of the PR

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -75,18 +77,25 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
     sampleIDs <- c(1:n.inputs,sample.int(n.inputs, (nens - n.inputs), replace = TRUE))
   }
   
-  ens.inputs <- list()
+  source('~/pecan/modules/assim.sequential/R/met_filtering_helpers.R')
```
mdietze: the call to source is a code development hack that shouldn't be necessary -- the met_filtering_helpers.R content will load as part of the package. It's also not stable since some people don't install pecan at ~/pecan (e.g. on my laptop I have pecan at ~/git/pecan)

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -75,18 +77,25 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
...
+  } else {
+    ens.inputs <- restart$ens.inputs
+  }
+  #ens.inputs <- list()
   inputs <- list()
   for(i in seq_len(nens)){
-    ### get only nessecary ensemble inputs. Do not change in anaylysis
-    ens.inputs[[i]] <- get.ensemble.inputs(settings = settings, ens = sampleIDs[i])
-    
+    print(i)
```
mdietze: remove debugging print statement

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -133,29 +145,58 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
...
-  if (sample_parameters == TRUE) {
-    settings$ensemble$size <- settings$state.data.assimilation$n.ensemble
+  if(is.null(restart) & is.null(restart$params)){
+    if (sample_parameters == TRUE) {
+      settings$ensemble$size <- settings$state.data.assimilation$n.ensemble
+    } else {
+      settings$ensemble$size <- 1
+    }
+    cumulative_ensemble_samples <- numeric(0)
+    repeat{ # temporary SIPNET hack, I want to make sure sum <1 for SIPNET
```
mdietze: This should at a minimum be generalized to something like ```if(model == 'SIPNET'){  repeat...```

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -350,24 +403,33 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
     ###-------------------------------------------------------------------###
     ### read restart                                                      ###
     ###-------------------------------------------------------------------###  
+    X_tmp <- vector("list", 2) 
     X <- list()
+    new.params <- params
+    source('~/pecan/models/sipnet/R/read_restart.SIPNET.R')
```
mdietze: as noted earlier, remove development hack

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -802,11 +872,10 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
     
     ## in the future will have to be separated from analysis
     new.state  <- analysis
-    new.params <- params
     
     ANALYSIS[[t]] <- analysis
     
-    if (interactive() & t > 1) { #
+    if (FALSE & interactive() & t > 1) { #
```
mdietze: This FALSE is a bit of a development hack too. Would be nice if there was a way to detect not only if the session is interactive but also whether it has a graphics capability. In lieu of that, probably fine to just drop the ```FALSE &```

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -880,7 +949,7 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
         # analysis
         ciEnvelope(as.Date(obs.times[t1:t]), XaCI[, 1], XaCI[, 2], col = alphapink)
         lines(as.Date(obs.times[t1:t]), Xa, col = "black", lty = 2, lwd = 2)
-        legend('topright',c('Forecast','Data','Analysis'),col=c(alphablue,alphagreen,alphapink),lty=1,lwd=5)
+        #legend('topright',c('Forecast','Data','Analysis'),col=c(alphablue,alphagreen,alphapink),lty=1,lwd=5)
```
mdietze: why was the legend removed?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -907,6 +977,8 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
       ### write restart by ensemble                                         ###
       ###-------------------------------------------------------------------### 
       
+      source('~/pecan/models/sipnet/R/write_restart.SIPNET.R')
```
mdietze: debug hack to remove

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -8,13 +8,14 @@
...
 ##' @param adjustment  flag for using ensemble adjustment filter or not
+##' @param restart
 ##' 
 ##' @description State Variable Data Assimilation: Ensemble Kalman Filter
 ##' 
 ##' @return NONE
 ##' @export
 ##' 
-sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustment = TRUE) {
+sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustment = TRUE, restart=NULL) {
```
araiho: Why is the restart param added? Maybe add description above?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -75,18 +77,25 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
...
+    print(i)
+    ### get only necessary ensemble inputs. Do not change in analysis
+    #ens.inputs[[i]] <- get.ensemble.inputs(settings = settings, ens = sampleIDs[i])
     ### model specific split inputs
     inputs[[i]] <- do.call(my.split_inputs, 
                            args = list(settings = settings, 
                                        start.time = settings$run$start.date, 
-                                       stop.time = settings$run$end.date,
-                                       inputs = ens.inputs[[i]]))
+                                       stop.time = as.Date(names(obs.mean)[1]),#settings$run$end.date,
```
araiho: That's a good idea!

modules/assim.sequential/R/sda.enkf.R
araiho: Are these checks that should be added to get.parameter.samples instead?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -173,12 +214,19 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
...
-    do.call(what = my.write.config, args = list(defaults = NULL, 
+    if(is.null(restart)){
+      do.call(what = my.write.config, args = list(defaults = NULL, 
                                          trait.values = params[[i]], 
                                          settings = settings, 
                                          run.id = run.id[[i]], 
                                          inputs = inputs[[i]], 
                                          IC = IC[i, ]))
+    } else {
+      ## copy over old run's forecast
```
araiho: By old runs to you mean spin up?

modules/assim.sequential/R/sda.enkf.R
istfer: this was a nasty hack, I probably should just comment it out, it also assumes the pft is `temperate.deciduous_SDA` 

I think in the end I moved towards defining the priors such that their sum won't exceed 1, but for now we can leave it as a comment as a reminder

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -350,24 +403,33 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
...
+      X_tmp[[i]] <- do.call(my.read_restart, args = list(outdir = outdir, 
                                                      runid = run.id[[i]], 
                                                      stop.time = obs.times[t], 
                                                      settings = settings, 
                                                      var.names = var.names, 
                                                      params = params[[i]]))
+      # states will be in X, but we also want to carry some deterministic relationships to write_restart
+      # these will be stored in params
+      X[[i]]      <- X_tmp[[i]]$X
+      new.params[[i]] <- X_tmp[[i]]$params
```
araiho: I'll write a note to fix read_restart.linkages to match this

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -929,7 +1001,9 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
     ###-------------------------------------------------------------------###
     ### save outputs                                                      ###
     ###-------------------------------------------------------------------### 
-    save(t, FORECAST, ANALYSIS, enkf.params, file = file.path(settings$outdir, "sda.output.Rdata"))
+    #save(t, FORECAST, ANALYSIS, enkf.params, file = file.path(settings$outdir, "sda.output.Rdata"))
+    save(list = ls(envir = environment(), all.names = TRUE),
```
araiho: This could get kind of huge right? Do we need the whole environment?

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -19,59 +19,74 @@
...
                      end.year = lubridate::year(stop.time),
                      variables = var.names)
   
-  last <- length(ens$NPP)
+  last <- length(ens$GWBI)
   
   forecast <- list()
   
-  # unit.conv <- (10000/1)*(1/1000)*(365.25*24*60*60) ## kgC m-2 s-1 -> MgC/ha/yr
+  kgms2Mghayr <- (10000/1)*(1/1000)*(365.25*24*60*60) ## kgC m-2 s-1 -> MgC ha-1 yr-1
```
araiho: do you use these or only ud.convert?

modules/assim.sequential/R/sda.enkf.R
istfer: yeah, I was only working with ensemble kalman filter so it wasn't getting as big as your. I changed this for development, will change it back

modules/assim.sequential/R/sda.enkf.R
istfer: it was blocking the most recent timestep, will change it back

modules/assim.sequential/R/sda.enkf.R
istfer: @mdietze do you want something specific?

modules/assim.sequential/R/sda.enkf.R
mdietze: probably could just move it to topleft instead

models/sipnet/R/read_restart.SIPNET.R
mdietze: good point, these sort of conversions should definitely be superseded by udunits

modules/assim.sequential/R/sda.enkf.R
istfer: I think @mdietze refers to the previous assimilation period (or both?) but maybe he can clarify

modules/assim.sequential/R/sda.enkf.R
mdietze: Yeah, I put a lot of that detail in the PR I sent to @istfer but you are right that more documentation should go directly into the code. Istem, I'll write that up and send it to you so you can insert it into the PR (probably easier than me doing another PR to you).

In a nutshell, restart is to allow forecasts to be iterative. e.g you ran a forecast from 2017-2027, it's now 2018 and you have new data, you want to restart from what you previously forecasted would happen in 2018. The code starts a completely new workflow (ID, folders, etc) but pulls over the previous forecast for that timestep, previous parameters, previous pointers to which met ensemble member you were using, etc. so that you restart the ensemble from the right place. The code then skips the first Forecast and dives directly into the first Analysis

modules/assim.sequential/R/sda.enkf.R
mdietze: As per comment above, I mean a previous workflow

Issue #227, opened 2014-10-08
Update query.trait.data to restrict query based on pfts.cultivar_id field


dlebauer:
In order to conduct a cultivar-level meta-analysis, I will need to restrict the trait data query to the cultivar associated with a PFT.

related to PecanProject/bety#144

infotroph:
@dlebauer Am I right that #1826 (cultivars in PFTs) resolved this / OK to close?

infotroph:
@dlebauer Closing this as probably resolved; Please reopen if needed.

Issue #224, opened 2014-10-01
add entity_id and cultivar_id to query.trait.data


dlebauer:
- entity_id is required for multivariate meta-analysis
- cultivar_id is required for cultivar level runs

infotroph:
@dlebauer OK to close?

PR #1826, opened 2018-01-26
Cultivars in pfts

infotroph:
Allows meta-analysis to operate on PFTs defined as sets of cultivars rather than whole species (#227, #224). 

## Description

Essentially all of these changes are internal to the low-level DB functions -- at the workflow level, just pass in the name of a PFT that's defined on cultivars and it should Just Work.

* New function `query.pft_cultivars` works essentially like `query.pft_species`, but implemented with dplyr calls intead of raw SQL statements.
* The bulk of the changes are in `get.trait.data.pft`, which now considers `pft_type` in addition to ID when looking up a PFT name, queries either `cultivars_pfts` or `pfts_species` as appropriate for this PFT, and adjusts posterior file names to match.

Breaking changes:
* The `trait.names` argument to `get.trait.data.pft` no longer has a default value, but this will have no effect because `get.trait.data`, the only function that currently calls it directly, still provides one.
* `query.traits`,`query.data`, and `query.trait.data` all gain a new `ids_are_cultivars` parameter (FALSE by default for backwards compatibility) to indicate whether the IDs to be queried are `specie_id` or `cultivar_id`.
* If a PFT name returns multiple IDs and no modeltype is specified to disambiguate them,  `get.trait.data.pft` now stops with logger.severe.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


infotroph:
TODO before merging:
* [ ] changelog
* [ ] fix `query.yields`, which I missed in the first push and is still assuming species.

infotroph:
@bcow You're paired with me for code review this week

infotroph:
To test this, you'll need some cultivar-PFTs. There's no Web interface to create these yet; here's how I made my testing set:

```

-- Invent some new cultivar PFTs
INSERT INTO pfts (name, definition,  modeltype_id, pft_type) VALUES
    ('Pavi_9005438', 'Switchgrass cultivar 9005438', 2, 'cultivar'),
    ('Pavi_alamo', 'Switchgrass cultivar "Alamo"', 2, 'cultivar'),
    ('Pavi_cir', 'Switchgrass cultivar "Cave-In-Rock"', 2, 'cultivar'),
    ('Pavi_all', 'All switchgrass cultivars 20180125', 2, 'cultivar');

-- associate them with the relevant cultivar IDs
INSERT INTO cultivars_pfts (pft_id, cultivar_id) VALUES
    ((SELECT id AS pft_id FROM pfts WHERE name='Pavi_9005438'), 1);
INSERT INTO cultivars_pfts (pft_id, cultivar_id) VALUES
    ((SELECT id AS pft_id FROM pfts WHERE name='Pavi_alamo'), 3);
INSERT INTO cultivars_pfts (pft_id, cultivar_id) VALUES
    ((SELECT id AS pft_id FROM pfts WHERE name='Pavi_cir'), 10);
INSERT INTO cultivars_pfts (pft_id, cultivar_id)
    SELECT
        (SELECT id FROM pfts WHERE name='Pavi_all') AS pft_id,
        id AS cultivar_id
    FROM cultivars WHERE specie_id = 938;

-- copy SLA prior from species-level switchgrass PFT `pavi`
SELECT prior_id INTO TEMP psla
    FROM priors
        JOIN pfts_priors ON (priors.id=prior_id)
        JOIN variables ON (variable_id=variables.id)
        JOIN pfts ON (pft_id=pfts.id)
    WHERE pfts.name='pavi' AND variables.name='SLA';

-- associate copied pavi prior with cultivar PFTs
INSERT INTO pfts_priors (pft_id, prior_id) VALUES
    -- Pavi_9005438 currently has no SLA records, so not assigning a prior
    ((SELECT id FROM pfts WHERE name='Pavi_alamo'),
        (SELECT prior_id FROM psla)),
    ((SELECT id FROM pfts WHERE name='Pavi_cir'),
        (SELECT prior_id FROM psla)),
    ((SELECT id FROM pfts WHERE name='Pavi_all'),
        (SELECT prior_id FROM psla));

DROP TABLE psla;
```

infotroph:
Documentation update to match, submitted as a separate pull request to avoid merge complications: #1831

dlebauer:
I've put the new pfts in the master database for testing.

infotroph:
I've assembled a demo package that should make it easier to verify that this works correctly on your machine: https://github.com/infotroph/pecan_cvpft_demo.

tonygardella:
@ankurdesai You're up for code review this week with Chris. I know you are quite busy so if you can't take a quick look for any glaring errors can you comment so @mccabete , who is paired with Chris next, can take a look? Tess, if you don't have time let me know and I'll review.

dlebauer:
@infotroph on a new 1.5.2 VM I ran https://github.com/infotroph/pecan_cvpft_demo It is a nicely prepared demo but at line 87 `settings_54 <- runModule.run.write.configs(settings_54)` I get 

> Error in read.table(file = file, header = header, sep = sep, quote = quote,  : 
  first five rows are empty: giving up

base/db/R/get.trait.data.R
```diff
@@ -56,37 +63,49 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon,
...
-        "SELECT pfts.id FROM pfts, modeltypes WHERE pfts.name='", pft$name,
-        "' and pfts.modeltype_id=modeltypes.id and modeltypes.name='", modeltype,
-        "'"
-      ),
-      con = dbcon
-    )[['id']]
+  pftres <- (dplyr::tbl(dbcon, "pfts")
+    %>% dplyr::filter(name == pft$name))
+  if (!is.null(modeltype)) {
+    pftres <- (pftres %>% dplyr::semi_join(
```
dlebauer: Does this need to be enclosd in quotes? Is there any effect of doing so?

base/db/R/query.pft.R
```diff
@@ -45,6 +47,57 @@ query.pft_species <- function(pft, modeltype, con) {
...
+  if (!is.null(modeltype)) {
+    pft_tbl <- (pft_tbl
+      %>% dplyr::inner_join(
+        dplyr::tbl(con, "modeltypes"),
+        by = c("modeltype_id" = "id"),
+        suffix = c("", ".mt"))
+      %>% dplyr::filter(name.mt == modeltype))
+  }
+
+  (pft_tbl
```
dlebauer: shouldn't this be captured as a new object and explicitly returned?

base/db/R/get.trait.data.R
infotroph: Do you mean enclosed in parentheses? It makes no difference here, I just got in the habit of wrapping my pipelines so that the `%>%` could be at the beginning of the line instead of the end.

base/db/R/query.pft.R
infotroph: Not if we're using Adv-R style: ["Only use return() for early returns."](http://r-pkgs.had.co.nz/style.html#style-fun)

dlebauer: OK. thanks for checking

base/db/R/get.trait.data.R
dlebauer: I'd vote for sticking w/ convention of pipes at end of line and not using parentheses (though unlike spaces, not enough to insist on, or hold up this PR)

PR #1828, opened 2018-01-28
Use binary R packages in Travis build

### LINKED ISSUE #1806: Build-breaker: bioconductor-mirror does not exist anymore
## Description

```
Rscript -e "devtools::document('"models/ed"')"
Updating PEcAn.ED2 documentation
Loading PEcAn.ED2
Loading required package: PEcAn.utils
Attaching package: ‘PEcAn.utils’
The following object is masked from ‘package:utils’:
    download.file
Loading required package: coda
Error in (function (dep_name, dep_ver = NA, dep_compare = NA)  : 
  Dependency package BiocInstaller not available.
Calls: <Anonymous> -> load_all -> load_imports -> mapply -> <Anonymous>
Execution halted
make: *** [.doc/models/ed] Error 1
```

## Context

`PEcAn.ED2` depends on `rhdf5`, which is only available from BioConductor. The current build system attempts to install it from https://github.com/bioconductor-mirror, which has been deleted. Builds without a cached copy of rhdf5 -- i.e. all of them as soon as our Travis cache gets clobbered for any reason -- now fail. Here's a sample [failing build](https://travis-ci.org/infotroph/pecan/builds/327282968) from my fork.

## Possible Implementations

1. Switch entirely from `rhdf5` to `h5 ` as proposed in #1797 
2. Change remote URLS, possibly to something like `github::bioconductor/BiocInstaller@RELEASE_3_5` and `git::https://git.biocondutor.org/packages/rhdf5.git`?
3. Temporary workaround: tell Travis to install the precompiled version of rhdf5 from [c2d4u](https://launchpad.net/~marutter/+archive/ubuntu/c2d4u/+packages?field.name_filter=rhdf5) and make users wrestle Bioconductor on their own until we deploy a more permanent fix
4. ???

ashiklom:
This dramatically reduces build times by using pre-compiled binaries of many R packages PEcAn depends in the Travis build. Major kudos to @infotroph for figuring out how to do this.

This also indirectly resolves #1806, since the `rhdf5` package is among the binaries installed.


Makefile
```diff
@@ -87,23 +87,23 @@ clean:
...
 
 .install/mockery: | .install
-	Rscript -e "if(!require('mockery')) install.packages('mockery', repos = 'http://cran.rstudio.com', Ncpus = ${NCPUS})"
+	time Rscript -e "if(!require('mockery')) install.packages('mockery', repos = 'http://cran.rstudio.com', Ncpus = ${NCPUS})"
 	echo `date` > $@
 
-depends_R_pkg = Rscript -e "devtools::install_deps('$(strip $(1))', threads = ${NCPUS});"
-install_R_pkg = Rscript -e "devtools::install('$(strip $(1))', Ncpus = ${NCPUS});"
+depends_R_pkg = time Rscript -e "devtools::install_deps('$(strip $(1))', threads = ${NCPUS});"
+install_R_pkg = time Rscript -e "devtools::install('$(strip $(1))', Ncpus = ${NCPUS});"
```
infotroph: All of these `times` were intended to be testing code and I forgot to remove them, but votes please: Is knowing the build times for individual packages useful for general debugging? If so, they only add a couple of output lines per Rscript -- I wouldn't be opposed to leaving these two `times`s in place and dropping the ones on devtools, roxygen, testthat, mockery.

ashiklom: I was wondering why I started seeing timings during the build process. I'd say keep 'em -- can't hurt to know what's taking up lots of time, and gives us a useful benchmark for further optimizing our build later.

PR #1835, opened 2018-02-01
Updating fates - in progress

serbinsh:
- [ X] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


serbinsh:
Steps:

1) Confirm all new FATES param names and correct param file
2) Check CLM params - update write.configs as needed
3) New PFTs?
4) Can we setup PEcAn to use cesm_inputs param files instead of storing in /inst?
5) How do we fix the ability to run at custom sites with custom met?  Need to link to correct domain and surf, also provide correct datm in info.  Need to check Ryan's site-level (BCI) examples to see what might be going wrong

tonygardella:
@serbinsh Would you like someone to review this or are you still adding to this?

serbinsh:
All updates assume (for now) that the user will be running CTSM-CLM5 as the host land model with FATES. Perhaps longer term we can be this flexible enough to use ELM or CLM

Also we still need to set up the wrapper to grab the FATES parameter file (and CLM) from the refcase folder.  We can make sure this happens by updating the create case script to copy the input param file from a custom folder OR the cesm data directory

serbinsh:
@tonygardella I am now ready for a review

serbinsh:
[ensemble.analysis.NOENSEMBLEID.GPP.2000.2001.pdf](https://github.com/PecanProject/pecan/files/1797711/ensemble.analysis.NOENSEMBLEID.GPP.2000.2001.pdf)

[ensemble.ts.NOENSEMBLEID.GPP.2000.2001.pdf](https://github.com/PecanProject/pecan/files/1797713/ensemble.ts.NOENSEMBLEID.GPP.2000.2001.pdf)

Changes seem to be working on modex

serbinsh:
bump @tonygardella, ready for a review...who is in line for me?

tonygardella:
@dlebauer is your code reviewer this week. @dlebauer if you don't have time, comment and I'll find someone else. Thank you!

serbinsh:
@dlebauer this PR is to bring the FATES wrappers in line with the use of CLM5-FATES and to operate as it had been previously.  I am working on a new PR to allow site-level inventory starts with FATES next and then when the new parameter file is released with new FATES updates and more PFTs I will need to create another PR to work with that version, and that would be the supported version moving forward.

models/fates/R/write.configs.FATES.R
```diff
@@ -183,14 +196,20 @@ write.config.FATES <- function(defaults, trait.values, settings, run.id){
 #   ## Write PARAMETER file
    
    ## COPY AND OPEN DEFAULT PARAMETER FILES
+   # TODO: update this to read param files (CLM and FATES) out of the refcase directory, not the PEcAn package
```
dlebauer: We've discussed in the past trying to avoid using "TODO's" in the text ... or at least making sure there is an open github issue and putting the link next to the TODO comment

models/fates/inst/create_CLM5-FATES_1x1br_refrun.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# Define output options
+export date_var=$(date +%s)
+echo "*** Build start: ${date_var} "
+export TAG='PEcAn_ref_run03'
+export MODEL_VERSION=CLM5-FATES
+export CLM_HASH=`(cd ${MODEL_SOURCE};git log -n 1 --pretty=%h)`
+export FATES_HASH=`(cd ${MODEL_SOURCE}/src/fates;git log -n 1 --pretty=%h)`
+export GIT_HASH=C${CLM_HASH}-F${FATES_HASH}
+export CASE_NAME=${CASEROOT}/${MODEL_VERSION}_${TAG}
```
ashiklom: Unless you need these variables outside of this script (i.e. in the parent process), you shouldn't need `export`, and in fact should probably avoid it to not pollute/confuse the environment.

I would try dropping `export` and seeing if the script still works.

models/fates/inst/template.job
```diff
@@ -13,101 +16,105 @@ mkdir -p "@OUTDIR@"
...
+#./xmlchange --file env_mach_pes.xml --id NTASKS_ATM --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_LND --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_ICE --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_OCN --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_CPL --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_GLC --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_ROF --val 1
+#./xmlchange --file env_mach_pes.xml --id NTASKS_WAV --val 1
+#./xmlchange --file env_mach_pes.xml --id MAX_TASKS_PER_NODE -val 1
+#./xmlchange --file env_mach_pes.xml --id TOTALPES --val 1
```
ashiklom: Commented values should be removed. They will live on forever in the Git history.

models/fates/inst/template.job
```diff
@@ -13,101 +16,105 @@ mkdir -p "@OUTDIR@"
...
-  ## DATES -> ENV_RUN
-  ./xmlchange -file env_run.xml -id RUNDIR -val @CASEDIR@/run
-  ./xmlchange -file env_run.xml -id RUN_STARTDATE -val @START_DATE@
-  ./xmlchange -file env_run.xml -id STOP_OPTION -val ndays
-  ./xmlchange -file env_run.xml -id STOP_N -val @STOP_N@
+## ENV_BUILD update configurations
+./xmlchange --file env_build.xml --id CIME_OUTPUT_ROOT --val @CASEDIR@
+#./xmlchange --file env_build.xml --id EXEROOT --val @BLD@  # this is the way it should be set, long term
+./xmlchange --file env_build.xml --id EXEROOT --val @CASEDIR@/bld # temporary fix
+#./xmlchange --file env_build.xml --id BUILD_COMPLETE --val TRUE  # TEMPORARY!  This eventually needs to be uncommented so we don't build the model each time !
```
ashiklom: Remove commented code.

models/fates/inst/template.job
```diff
@@ -13,101 +16,105 @@ mkdir -p "@OUTDIR@"
...
+## SURFDATA
+rm @INDIR@/lnd/clm2/surfdata_map/surfdata*
+ln -s @RUNDIR@/surfdata_@SITE_NAME@_simyr2000.nc @INDIR@/lnd/clm2/surfdata_map/
+#export SURFMAP=@INDIR@/lnd/clm2/surfdata_map/surfdata_@SITE_NAME@_simyr2000.nc
+
+# update namelist
+#cat >> user_nl_clm << EOF
+#fsurdat = '@INDIR@/lnd/clm2/surfdata_map/surfdata_@SITE_NAME@_simyr2000.nc'
+#finidat = ' '
+#EOF
```
ashiklom: Remove commented code.

models/fates/inst/template.job
```diff
@@ -116,77 +123,82 @@ fates_paramfile = '@INDIR@/lnd/clm2/paramdata/fates_params.@RUN_ID@.nc'
...
   
-  ## ADDITIONAL MODS THAT ARE JUST ASSOCIATED WITH REFCASE - removed 'NEP' 'NPP_column'
-cat >> user_nl_clm << EOF
+## ADDITIONAL MODS THAT ARE JUST ASSOCIATED WITH REFCASE
+#cat >> user_nl_clm << EOF
+#hist_empty_htapes = .true.
+#hist_fincl1='EFLX_LH_TOT','TSOI_10CM','QVEGT','GPP','AR','ED_bleaf','ED_biomass','NPP','MAINT_RESP','GROWTH_RESP'
+#hist_mfilt             = 8760
+#hist_nhtfrq            = -1
+#EOF
```
ashiklom: Remove commented code.

models/fates/inst/template.job.createclone
```diff
@@ -48,47 +51,46 @@ export GFORTRAN_UNBUFFERED_PRECONNECTED=yes
...
+./xmlchange --file env_run.xml --id DIN_LOC_ROOT --val @INDIR@
+./xmlchange --file env_run.xml --id DIN_LOC_ROOT_CLMFORC --val '@INDIR@'
+./xmlchange --file env_run.xml --id DOUT_S --val TRUE
+./xmlchange --file env_run.xml --id DOUT_S_ROOT --val '@CASEDIR@/run'
+./xmlchange --file env_run.xml --id PIO_DEBUG_LEVEL --val 0
+./xmlchange --file env_run.xml --id ATM_DOMAIN_FILE --val 'domain.lnd.@SITE_NAME@.nc'
+./xmlchange --file env_run.xml --id ATM_DOMAIN_PATH --val '@INDIR@/share/domains/domain.clm/'
+./xmlchange --file env_run.xml --id LND_DOMAIN_FILE --val 'domain.lnd.@SITE_NAME@.nc'
+./xmlchange --file env_run.xml --id LND_DOMAIN_PATH --val '@INDIR@/share/domains/domain.clm/'
+./xmlchange --file env_run.xml --id CLM_USRDAT_NAME --val '@SITE_NAME@'
```
ashiklom: A lot of this seems redundant across multiple files. Not necessary for this PR, but it would be nice if there was a way to use `source` or something else to keep these commands in only one place.

models/fates/R/write.configs.FATES.R
serbinsh: Just addressed this @dlebauer

models/fates/inst/template.job
ashiklom: Unless they are for upcoming features that haven't been implemented yet, in which case you can leave them.

serbinsh: More so, I dont know yet if these need to be explicit or not on all machines.  The tricky part is that depending on the machine-specific configs, these may need to be set but on Modex I have these set automatically.  So to be honest I dont know yet if we need these set or not until someone else tries running on a different machine

models/fates/inst/template.job.createclone
serbinsh: @ashiklom I dont understand this comment, could you elaborate?  This is the approach used to set CLM settings...

serbinsh: Also create clone is still a work in progress since I still cannot get the actual CIME command ./create_clone to work and not much help from the CLM community....so its not really implemented yet

ashiklom: I just meant that, for example, lines 77-81 of this file are, character for character, identical to lines 74-78 the `template.job` file, so any changes we might want to make in the future would have to be changed twice. Again, not having the perspective on the full code, I don't know how simple it would be to remove this redundancy with some kind of wrapper or template (which is why it's just a comment and not a "Request changes"), but something to think about.

models/fates/inst/create_CLM5-FATES_1x1br_refrun.sh
ashiklom: My understanding is that if you run this script followed by another script that does depend on these settings, then you need to keep `export`. I.e. If the workflow is something like:

```
./create_CLM5-FATES...
./run_model.sh   # This script depends on CLM_HASH, FATES_HASH, etc...
```

However, if these variables are only used _within_ the `create_CLM45...` script, then you should remove the export.

models/fates/inst/create_CLM5-FATES_1x1br_refrun.sh
```diff
@@ -0,0 +1,112 @@
...
+#
+# --- Last updated by Shawn P. Serbin <sserbin@bnl.gov>
+#=============================================================================================
+
+# Define CTSM-FATES options
+export MODEL_SOURCE=/data/software/CTSM/ctsm                   	# CTSM/CLM/ELM source directory - USER UPDATE
+export PROJECT=PEcAn_simulations                               	# project or run tag
+export COMPSET=I2000Clm50FatesGs				# define compset for run
+export RES=1x1_brazil						# set grid resolution
+export MAC=modex                                             	# Name your machine - USER UPDATE
```
tonygardella: @serbinsh Is the modex machine settings a default in the machine config file or did you make your own machine settings?

serbinsh: Shoot, that needs to be switched back to a default machine name there like before and the end user needs to then either leave it or switch it to match their local machine.  I will switch it back to eddi, which is what it was before

models/fates/inst/template.job.createclone
serbinsh: @ashiklom Ah, ok.  Well the issue is that we dont want to template.job files but until I can figure out how to properly use ./create_clone or the bugs are sorted we need to continue to rebuild the model exe. But the goal is to deprecate the current template.job for the version that has the create clone command so that we dont rebuild every time but instead re-use the existing exe.  So in the end there are redundancies not but I hope to get rid of one at some point

models/fates/inst/template.job
ashiklom: Fair enough.

In general, I don't like that so much of this is hard-coded within the `template.job` file. Given that all of these commands follow a very regular pattern, I think it's a much better idea to do this all from within R so that the user can set any or all of these tags from within the `pecan.xml` file. For instance, something like this:

```r
xmlchange <- function(file, id, value) {
  system2("./xmlchange", c("--file", file, "--id", id, "--value", value))
}

set_xml_values <- function(file, tag_list) {
  for (i in seq_len(tag_list)) {
    xmlchange(file, names(tag_list)[i], tag_list[[i]])
  }

write_fates_xml <- function(casedir, site_name, xml_file = "env_build.xml" ...) {
  dots <- list(...)
  default_tags <- list(NTASKS_ROF = 1, NTASKS_WAV = 1, MAX_TASKS_PER_NODE = 1, TOTALPES = 1, <etc>)
  new_tags <- modifyList(list(CIME_OUTPUT_ROOT = casedir, EXEROOT = file.path(casedir, bld)), dots)
  use_tags <- modifyList(default_tags, new_tags)
  set_xml_vaues(xml_file, use_tags)

settings$run$other_settings <- list(NTASKS_ROF = 1, NTASKS_WAV = 1)
do.call(write_fates_xml, c(list(casedir = settings$run$casedir, site_name = settings$site_name), settings$run$other_settings))
```

...but that's probably best saved for another time.

models/fates/inst/create_CLM5-FATES_1x1br_refrun.sh
tonygardella: Cool! At some point could I see your machine settings? I don't think eddi would work for us. Also, in the future we'll need to make some documentation on how a user would update these settings.

models/fates/inst/template.job
serbinsh: Actually, that is a good plan for a targeted PR that would benefit both CLM and FATES.  I am going to take this nugget and create an issue focused on this sort of "upgrade" to the code

tonygardella:
@serbinsh Did you make all the changes you wanted?

PR #1845, opened 2018-02-06
Add documentation for modifying Bety schema

ashiklom:
If this exists in the Bety docs already, I couldn't find it. The Bety docs may be a better place for this -- if so, let me know where to put it and I'll happily stick it there. Though, for what it's worth, because PEcAn and Bety are so tightly integrated, I think it's a good idea to merge the documentation into a single document.


dlebauer:
Thanks for putting this together. This definitely goes in the BETYdb docs, but can be linked to from the PEcAn docs.

ashiklom:
OK, closing this PR and opening one for the `betydb-documentation`. See https://github.com/PecanProject/betydb-documentation/pull/6

PR #1854, opened 2018-02-13
 Making met2model.ED2 stop enforcing leap years. 

mccabete:
This puts in an optional flag (leap_years= FALSE) that stops met2model_ED2 from enforcing leap years. Works the same flag into the days_in_year and seconds_in_year base function for other models. 

The default is set to enforcing leap years, because the majority of met sources do include leap years and deviations would represent genuine bugs in the met.  Since ED2 also works with leap years, there was no need to enforce 365 days. 

## Motivation and Context
Some met sources don't include leap years, and some models don't enforce leap years. We decided as a group in this issue that we should modify the met2model to either enforce or not enforce leap years based on the model's ability to process leap years. To do that for ED2 (doesn't require leap years) I included a leap year flag. 

## Review Time Estimate

- [x] Not pressing, but < 5 min PR
- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/utils/R/days_in_year.R
```diff
@@ -11,9 +11,9 @@
 #' days_in_year(2010)  # Not a leap year -- returns 365
 #' days_in_year(2012)  # Leap year -- returns 366
 #' days_in_year(2000:2008)  # Function is vectorized over years
-days_in_year <- function(year) {
+days_in_year <- function(year, leap_year = TRUE) {
```
ashiklom: This argument should be documented. Add something like the following to the Roxygen block above this:

```r
#' @param leap_year Whether or not to enforce leap years. If `FALSE`, always return a 365 day year.
```

base/utils/R/seconds_in_year.R
```diff
@@ -7,8 +7,8 @@
 #' seconds_in_year(2001)  # Regular year -- 365 x 24 x 60 x 60 = 31536000
 #' seconds_in_year(2000:2005)  # Vectorized over year
 #' @export
-seconds_in_year <- function(year) {
-    diy <- days_in_year(year)
+seconds_in_year <- function(year, leap_year = TRUE) {
+    diy <- days_in_year(year, leap_year)
```
ashiklom: Add the following to document this parameter (if it's not already in here).

```r
#' @inheritParams days_in_year
```

models/ed/R/met2model.ED2.R
```diff
@@ -54,12 +55,18 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
   dl <- c(0, 32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336, 367)
   month <- c("JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC")
   mon_num <- c("01", "02", "03", "04", "05", "06", "07", "08", "09", "10", "11", "12")
-  day2mo <- function(year, day) {
-    leap      <- lubridate::leap_year(year)
-    mo        <- rep(NA, length(day))
-    mo[leap]  <- findInterval(day[leap], dl)
-    mo[!leap] <- findInterval(day[!leap], dm)
-    return(mo)
+  day2mo <- function(year, day, leap_year_agnostic) {
```
ashiklom: You have `leap_year_agnostic` here, but use `leap_year` in the function. Assuming this is a typo?

models/ed/R/met2model.ED2.R
```diff
@@ -54,12 +55,18 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
+      mo        <- rep(NA, length(day))
+      mo <- findInterval(day, dm)
+      return(mo)
+    }else{
+      leap      <- lubridate::leap_year(year)
+      mo        <- rep(NA, length(day))
+      mo[leap]  <- findInterval(day[leap], dl)
+      mo[!leap] <- findInterval(day[!leap], dm)
+      return(mo)
+    }
```
ashiklom: `if (x == FALSE)` can be more efficiently written as `if (!x)`.

Also, `mo <- rep(NA, length(day))` is the same in both blocks, so it can be pulled out of the `if-else` altogether.

Finally, the Tidy style guide technically requests spaces between `if/else` and `(/{` (i.e. `if (...) {` and `} else {`.) 

Personally, I would recommend enabling RStudio's "Diagnostics" highlighting for editing code -- it does a good job of keeping you honest about style things like this. To do that, go to "Tools / Global Options / Editing / Diagnostics" and check the things you want it to check for you (I like to have every box in that section checked).

base/utils/R/days_in_year.R
ashiklom: Also, once you do this, you should re-run `make document` to make sure the documentation is updated.

models/ed/R/met2model.ED2.R
ashiklom: Also, since you are modifying the function here, you need to change its call later in the script ([line 177](https://github.com/PecanProject/pecan/blob/8f0e23dd2a25568fd62effec78d8c32ace071c78/models/ed/R/met2model.ED2.R#L177)).

base/utils/R/days_in_year.R
mccabete: Woops. Thanks for catching that.

models/ed/R/met2model.ED2.R
```diff
@@ -26,9 +26,11 @@
 ##' @param lst timezone offset to GMT in hours
 ##' @param overwrite should existing files be overwritten
 ##' @param verbose should the function be very verbose
+##' @param leap_year Enforce Leap-years? If set to TRUE, will require leap years to have 366 days. If set to false, will require all years to have 265 days. Default = TRUE.
```
ashiklom: 365 days, I think, not 265

models/ed/R/met2model.ED2.R
```diff
@@ -146,7 +153,7 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
     hr   <- NULL
     asec <- sec
     for (y in seq(year, year + nyr - 1)) {
```
istfer: @mccabete  I'm curious, does this`seq(year, year + nyr - 1)` work for others as it's supposed to? when I try to process years 2005 and 2006, this becomes `2004` and `2005` for the 2005 loop and `2005` and `2006` for the 2006 loop

feel like `nyr <- floor(udunits2::ud.convert(length(sec) * dt, "seconds", "years"))` should be `round` or something?

istfer: looks like that's due to @ashiklom's [changes](https://github.com/PecanProject/pecan/commit/bd1ce911eb14282caeabe7dfb4affd6714de18d8)

previous version returns a `1`
```
# for an example Ameriflux met `length(sec) = 17520`, `dt = 1800`
floor(length(sec) / 86400 / 365 * dt)
```

new version returns `0`
```
floor(udunits2::ud.convert(length(sec) * dt, "seconds", "years"))
```

ashiklom: I don't know why those two would return different values. `sec / 86400 = days` and `days / 365` gives years, so that unit conversion should be the same.

ashiklom: Ah, `udunits` is more precise...

```
udunits2::ud.convert(1, "years", "seconds")
# [1] 31556926
86400 * 365
# [1] 31536000
```

So yeah, we should probably change that back. My mistake! I didn't think to check the calculation.

ashiklom: A bit more detail: apparently `udunits2` recognizes that an astronomical year is actually ~0.25 days longer than a calendar year.

```
udunits2::ud.convert(1, "years", "days")
# [1] 365.2422
``

ashiklom: Indeed, buried in the [UDUnits documentation](https://www.unidata.ucar.edu/software/udunits/udunits-2.0.4/udunits2lib.html) is this:

> CAUTION: The timestamp-unit was created to be analogous to, for example, the degree celsius—but for the time dimension. I've come to believe, however, that creating such a unit was a mistake, primarily because **users try to use the unit in ways for which it was not designed (such as converting dates in a calendar whose year is exactly 365 days long)**. Such activities are much better handled by a dedicated calendar package. Please be careful about using timestamp-units.

ashiklom: So, in summary, @mccabete would you mind cleaning up the mess I made by changing `nyr <- floor(...` on line 150 back to `nyr <- floor(length(sec) * dt / 86400 / 365)`?

mccabete:
@ashiklom Yes, but only because you were so kind about my typos.

tonygardella:
@mccabete Could you do a `make document` and push?

tonygardella:
@mccabete Could you `make document` please?

ashiklom:
This PR is fixed and ready to be merged.

PR #1856, opened 2018-02-15
Define package location of teardown function

tonygardella:
A few builds on TRAVIS are failing with error:
```
Testing PEcAn.BIOCRO
Error in teardown(PEcAn.logger::logger.setLevel(loglevel)) : 
  could not find function "teardown"
```
## Description
Added `testthat::` so that it knows where `teardown` is coming from.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
So the underlying problem here is that teardown only exists in testthat >= 2.0.0, which is already listed in the DESCRIPTION but apparently not enforced at install time because testthat is in suggests.

PR #1860, opened 2018-02-21
RTM: Add resampling functions

ashiklom:
## Description

Basically, these are just a convenient wrapper around base R's `splinefun`, but they are particularly useful for working with spectra.

NOTE that this PR is a superset of #1859. Once that gets merged, this will only show the new functions and none of the formatting revisions.

## Motivation and Context

Not quite a formal approach to observation error, but a convenient shortcut to the common situation of receiving spectra at non-integer wavelengths, or for quickly comparing multiple different instruments.

## Review Time Estimate

- [ ] Immediately
- [X] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/rtm/R/resample.R
```diff
@@ -0,0 +1,57 @@
+#' Resample vector, matrix, or spectra
+#'
+#' Conveinent wrapper around base R's `splinefun`. See [stats::splinefun()]
```
serbinsh: Should we enable linear as well?

modules/rtm/tests/testthat/test.resample.R
```diff
@@ -0,0 +1,51 @@
...
+    expect_equal(true_m, resample_m)
+  }
+)
+
+data(testspec)
+test_that(
+  "Spectra resampling works",
+  {
+    spec <- spectra(testspec_ACRU[, 1:5], 400:2500)
+    new_wl <- seq(400, 1300, 10)
```
serbinsh: Can we add a test to convert from "coarse" to 1nm or finer spec as well?   Or I could just test that myself by reading in some AVIRIS or g-liht spec.

modules/rtm/tests/testthat/test.resample.R
```diff
@@ -0,0 +1,51 @@
...
+  }
+)
+
+test_that(
+  "Multivariate resampling works",
+  {
+    m <- matrix(1:20, ncol = 2)
+    true_m <- cbind(seq(1, 10, 0.5), seq(11, 20, 0.5))
+    resample_m <- resample(m, 1:10, seq(1, 10, 0.5))
+    expect_equal(true_m, resample_m)
```
serbinsh: @ashiklom is this so you can resample the spec res of a matrix (i.e. image)?

serbinsh:
This looks pretty useful

modules/rtm/R/resample.R
ashiklom: Yeah, I suppose that could be useful. I'll add a `if (method %in% c("linear", "constant")) <use approxfun> else <use splinefun>`.

modules/rtm/tests/testthat/test.resample.R
ashiklom: An image would probably be a cube, which this doesn't quite cover (I suppose I could try to extend this to arrays with 3+ dimensions, but it's a bit more complicated), but generally, yeah, the idea is if you have multiple observations, all at the same spectral resolution, you can pass them directly into `resample` without having to manually `apply` or loop and `cbind`.

modules/rtm/tests/testthat/test.resample.R
ashiklom: Yeah, this is probably a good idea. I'll just need to figure out a good error tolerance for `expect_equal`.

ashiklom:
@serbinsh Implemented your recommendations. You can now do `resample(..., method = "linear")`, which will call `approxfun(..., method = "linear")` instead of `splinefun(...)` under the hood. I also fixed some NAMESPACE issues (forgot to `@export` these functions), and a minor bug related to arguments to `plot.spectra`.

PR #1867, opened 2018-03-05
HPDA code

istfer:
## Description

Hierarchical PDA code that fits model parameters at multiple sites locally, globally, and hierarhically.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


istfer:
Initial code push after ESA. Needs some more cleanup and testing.

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
```diff
@@ -0,0 +1,646 @@
...
+   <id>1000000022</id>
+</sitegroup>
+```
+
+In the meantime, your `<run>` section should specify `<inputs>`, `<start.date>` and `<end.date>` information accordingly. E.g.:
+
+
+```
+ <run>
+  <settings.1>
```
mdietze: I didn't think you needed the .n at the end of settings. I've never used that before when setting up multisettings runs

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
```diff
@@ -0,0 +1,646 @@
...
+
+If requested, your `<ensemble>` and `<sensitivity.analysis>` sections would look similar to this in your `pecan.CONFIGS.xml`:
+
+
+```
+ <ensemble>
+  <settings.1>
+   <size>100</size>
+   <variable>NEE</variable>
+   <ensemble.id>1000017326</ensemble.id>
```
mdietze: Is this really necessary? I've never specified ensemble.id within the ensemble block before and I've never specified anything nested within settings tags. Why can't you just use the standard notation and provide a single size and variable tag and let it apply those to all sites using the default run start/end dates. It seems like you only need to do this if the start/end are different from the run's

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
```diff
@@ -0,0 +1,646 @@
...
+  </param.names>
+</assim.batch>
+```
+
+Different from the PDA settings, this time the main function will be the `emulator.ms`. This function can be run in three modes, which will be selected under the `<mode>` tag above: local, global and hierarchical. If left empty, the function will go through all three modes. But note that, for both global and hierarhical calibration, we need site-level runs. We will start with local mode, i.e. site-level calibration. 
+
+Now we will extend the settings above to include information about data to assimilate and likelihoods, similar to the standard-PDA setup, except for multiple sites. First, add `<assim.batch>` tag to the multisettings in the bottom of your `pecan.HBPDA.xml`:
+
+```
+ <multisettings>
```
mdietze: I don't recall this tag before, are you adding it?

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
```diff
@@ -0,0 +1,646 @@
...
+  <multisettings>assim.batch</multisettings>
```
mdietze: Why would the tag within multisettings be multisettings again? What is this telling us that's different from the overall pecan.xml

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
```diff
@@ -0,0 +1,646 @@
...
+  <multisettings>sensitivity.analysis</multisettings>
+  <multisettings>run</multisettings>
+ </multisettings>
+```
+
+Now we can insert multi-site settings in the `<assim.batch>` section accordingly (note that now the chunk above where we defined `method`, `mode`, `param.names` etc. will go under each `setting.*`):
+
+```
+  <assim.batch>
+  <settings.1>
```
mdietze: The design below seems to have a lot of redundant information. Inputs are site specific, but why/how would method, mode, n.know, iter, chain, param.names, be different between sites?

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
istfer: hmm, I can't remember whether I came up with this numbering or just mimicking what pecan workflow was automatically producing but it doesn't break anything and I thought it makes things easier to follow

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
istfer: I haven't specified these myself either, workflow produce them automatically? I'm just letting users know what it looks like in the CONFIGS.xml at this point. Otherwise my original pecan.xml has this only:
```
 <ensemble>
  <size>100</size>
  <variable>NEE</variable>
 </ensemble>
```

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
istfer: this is also automatically being added by pecan workflow at the end of the CONFIGS.xml.  I added the tag below so that multi.settings changes also apply to `assim.batch` tags in the xml.

So, at the end of CONFIGS.xml there was this chunk (created by the workflow):
```
 <multisettings>
  <multisettings>ensemble</multisettings>
  <multisettings>sensitivity.analysis</multisettings>
  <multisettings>run</multisettings>
 </multisettings>
```
If the module is listed there, `read.settings` reads the tag as `multi.settings` for that module. As assim.batch tags are added manually currently, so I also added `<multisettings>assim.batch</multisettings>` there manually.

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
istfer: they wouldn't be different but otherwise I was not able to load `assim.batch` tags as multi settings properly. 

I'm not familiar with multi.settings usage, and couldn't see any documentation. So I was just mimicking what is being written when we write a multi.settings list back to an xml (e.g. the ensemble tag above) to make it work. I can play around a little bit more.

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
mdietze: Ok, I just went back through earlier multisettings runs and it's something that was already there, it was just something that I'd never looked at before because my sitegroup runs expand to that automatically without me going and looking at them.

I agree the numbering is useful.

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
mdietze: Got it. It's probably good to have the vignette focus on what you put in originally, otherwise it leaves the impression that the user has to do a LOT more work than they actually do.

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
mdietze: Like with the other multisettings (e.g. ensemble), I bet there's a way to get it to do a lot of the expansion automatically. Probably OK for it to be this way in the current PR, but let's try to make things easier for the user by the next release

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
rykelly: IIRC the numbering doesn't serve any practical purpose, except to not have a bunch of identically named blocks. As in, I don't know what XML2list does if it finds a bunch of identically named list items at the same level in the hierarchy, but I think I just wanted to have control over that.

modules/assim.batch/vignettes/MultiSitePDAVignette.Rmd
rykelly: Hehe. Yeah, maybe not the best naming choice I've made. But yes, this is just a bit of metadata the MultiSettings code uses to know which blocks are meant to be replicated for each Settings. Looks kludgey to me (even aside from the naming). But I vaguely remember being convinced there needed to be something in there to make it explicit which settings were to be expanded.

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+    local <- TRUE
+    global <- hierarchical <- FALSE
+  }else if(pda.mode == "global"){
+    global <- TRUE
+    local <- hierarchical <- FALSE
+  }else if(pda.mode == "hierarchical"){
+    hieararchical <- TRUE
+    local <- global <- FALSE
+  }else{
+    local <- global <- hierarchical <- TRUE
```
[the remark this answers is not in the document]
istfer: for hierarchical calibration we need local runs anyway, and global is not really too much work in between.. I thought if someone forgot or omitted to specify the `<mode>` tag it's safest to assume they want all three

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+  ## -------------------------------- Prepare for Global and Hierarchical ----------------------------------------- 
+  
+  
+  # we need some objects that are common to all calibrations
+  obj_names <- c("init.list", "rng", "jmp.list", "prior.fn.all", "prior.ind.all", "llik.fn", 
+    "settings", "prior.ind.all.ns", "sf", "prior.list", "n.param.orig", "pname", "prior.ind.orig",
+    "hyper.pars")
+  
+  need_obj <- load_pda_history(workdir = multi.settings$outdir,  
+                   ensemble.id = multi.settings[[1]]$assim.batch$ensemble.id,
```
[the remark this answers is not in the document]
istfer: all these objects are common to all site-level runs as well as global and hierarchical.. could either call the same functions (in pda.emulator) to extract them, or can get them from one of the previous runs' saved history..1st site is as good as any

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+      jmp.list  <- norm_transform$jmp
+
+      
+    }
+    
+
+    ########### hierarchical MCMC function with Gibbs ##############
+    
+    
+    hier.mcmc <- function(settings, gp.stack, nstack, nmcmc, rng,
```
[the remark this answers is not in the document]
istfer: the only reason this is a function was that so that I can use `parLapply` on it later and I kinda found it convenient to have it in the same script while developing but now I can move it to its own document

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+      #
+      #      tau_df    : Wishart degrees of freedom
+      #      tau_V     : Wishart scale matrix
+      #
+      #      tau_global ~ W (tau_df, tau_scale)
+      #      sigma_global <- solve(tau_global)
+      #
+      
+      tau_df <- nsites + nparam + 1
+      tau_V  <- diag(1, nparam)
```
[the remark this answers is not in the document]
istfer: yeah, tau_df is only used once to initialize tau_global which is probably unnecessary because tau_global is immediately updated in the loop. That's why I initialized it to `nsites + nparam + 1` 

some of these initializations were to wrap my head around dimensions, hyperparameters etc. I'll try to document it more

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+    # start the clock
+    ptm.start <- proc.time()
+    
+    # prepare for parallelization
+    dcores <- parallel::detectCores() - 1
+    ncores <- min(max(dcores, 1), settings$assim.batch$chain)
+     
+    logger.setOutputFile(file.path(settings$outdir, "pda.log"))
+    
+    cl <- parallel::makeCluster(ncores, type="FORK", outfile = file.path(settings$outdir, "pda.log"))
```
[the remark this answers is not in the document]
istfer: it appends to it. maybe I can back it up, but pda.log is only useful while mcmc is going on to see whether it's stuck or something, otherwise I almost never go check it back, maybe I should delete the previous one to save space

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+          repeat{ # make sure to stay in emulator boundaries, otherwise it confuses adaptation
+            mu_site_new[ns,] <- mvtnorm::rmvnorm(1, mu_site_curr[ns,], jcov.arr[,,ns])
+            check.that <- (mu_site_new[ns,] > rng[, 1, ns] & mu_site_new[ns,] < rng[, 2, ns])
+            if(all(check.that)) break
+          }
+        }
+        
+        
+        # re-predict current SS
+        currSS    <- sapply(seq_len(nsites), function(v) get_ss(gp.stack[[v]], mu_site_curr[v,], pos.check))
```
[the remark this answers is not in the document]
istfer: it is, if there are multiple emulators per site they're sublists under the site-level list

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+
+  
+  # lists to collect emulators and run MCMC per site later
+  gp.stack      <- vector("list", nsites) 
+  SS.stack      <- vector("list", nsites) 
+  #nstack       <- vector("list", nsites) 
+  
+  ## -------------------------------------- Local runs and calibration ------------------------------------------
+  
+  for(s in seq_along(multi.settings)){ # site runs - loop begin
```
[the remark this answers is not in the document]
istfer: yeah I'll change these to papply, but I want to note that it's a little too tolerant for failures. I only realized after three sites ran that I had a typo in the xml that would normally cause error in the first pass otherwise

>Speaking of parallelization, if pda.emulator does not return until it's done then we're really missing the ability to leverage remote by putting all runs for all sites into the queue at once.

This is true, I will look into this later.

>Also, I assume pda.emulator can be configured to do multiple rounds of knots per site automatically?

It doesn't do it automatically yet. If user decides they need another round they need to re-read the xml written by the pda-workflow and re-start pda.emulator. But automation shouldn't be difficult.

istfer:
@rykelly it _is_ awesome! I saved myself many headaches and time using your code, thanks!

modules/assim.batch/R/pda.emulator.ms.R
```diff
@@ -0,0 +1,518 @@
...
+        # tau_global   : error precision matrix
+        
+        # sum of pairwise deviation products
+        pairwise_deviation <- apply(mu_site_curr, 1, function(r) r - mu_global)
+        sum_term <- pairwise_deviation %*% t(pairwise_deviation)
+        
+        tau_sigma <- solve(V_inv + sum_term)
+        
+        # update tau
+        tau_global <- rWishart(1, df = tau_df, Sigma = tau_sigma)[,,1] # site precision
```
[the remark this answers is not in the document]
istfer: as it needs to be done once, I did it before the loop, but as it's not a big calculation I can write it our explicitly

istfer:
@para2x I just added my global SA function to the vignette if it helps (rest of the vignette is still not up-to-date). As global SA is not fully integrated with pecan workflow, I was using my custom function (sorry it's not fully commented). 

I'd be happy to use a more integrated/generalized version if you/we decide some version of our approaches should go into the mainline/shiny app.

modules/assim.batch/R/hier.mcmc.R
```diff
@@ -0,0 +1,244 @@
...
+  #
+  #      tau_global ~ W (tau_global_df, tau_global_sigma)
+  #      sigma_global <- solve(tau_global)
+  #
+  
+  sigma_global_df   <- nparam + 1
+  sigma_global_scale   <- (cov(mu_init_samp)/sigma_global_df) 
+  
+  # initialize sigma_global (nparam x nparam)
+  sigma_global <-  riwish(sigma_global_df, sigma_global_scale)
```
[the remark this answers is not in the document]
istfer: @mdietze  by reasonably well I mean, if I generate some values from these priors and check the ranges of values 95% looks OK

```
> allp <- list()
> for(i in 1:1000){
     sigma_global <-  MCMCpack::riwish(sigma_global_df, sigma_global_scale)
     mu_global <- rmvnorm(1, mu_global_mean, mu_global_sigma)
     allp[[i]] <- rmvnorm(1000, mu_global, sigma_global)
     
 }
> allsamp <- do.call("rbind", allp)
> apply(allsamp,2,quantile, c(0.025,0.975)) #these values are OK-ish
           [,1]       [,2]      [,3]      [,4]      [,5]
2.5%   103.3031 0.01836558  6.204159  11.34932  2.953825
97.5% 1399.7945 0.18640106 38.356880 234.00614 48.601049
```

But there are really long tails that extend to unrealistic values, e.g.
```
>  apply(allsamp,2,range)
          [,1]      [,2]      [,3]      [,4]      [,5]
[1,] -29746.92 -2.923556 -585.8785 -6048.621 -766.2857
[2,]  29529.44  3.029907  559.2427  5789.496  866.6587
```
pairs

modules/assim.batch/R/hier.mcmc.R
```diff
@@ -0,0 +1,244 @@
...
+    # propose new site parameter vectors
+    repeat{ # make sure to stay in emulator boundaries, otherwise it confuses adaptation
+      thissite <- g %% nsites
+      if(thissite == 0) thissite <- nsites
+      proposed <- mvtnorm::rmvnorm(1, mu_site_curr[thissite,], jcov.arr[,,thissite])
+      check.that <- (proposed > rng_orig[, 1] & proposed < rng_orig[, 2])
+      if(all(check.that)) break
+    }
+
+    mu_site_new <- matrix(rep(proposed, nsites),ncol=nparam, byrow = TRUE)
```
istfer: @mdietze  Now I propose the same value for all sites, because I think this way (when a proposed value is accepted by more than one site) `pairwise_deviation <- apply(mu_site_curr, 1, function(r) r - t(mu_global))` calculation has less scatter (and I'm more precise). This slightly increases the influence of the global priors later in the accept/reject part. What do you think?

The implementation is a bit hacky, every iteration new value is proposed by one site, but it mixes fine if run the chains a little longer. Normally if site-to-site variability is not huge the proposed value might as well be accepted by other sites anyway.

istfer:
@Dietze this is still not done, but @infotroph and I worked on it so that it can pass travis. I can continue working on this PR, but if you prefer to merge this monster, I'll owe you two more PRs: one for finishing up the HPDA work, and one for the package clean up

istfer:
@mdietze just pinging you about this just in case you missed it, @infotroph will push the roxygen update soon and it will create conflicts with this PR. 

PR passes travis but not all comments were addressed, but it also has some general fixes to the PDA workflow which others can make use of. So I'm leaving the decision to you, we can either merge this and I will open up a new PR, or it can stay. I'm happy to resolve conflicts and keep working on this one.

PR #1870, opened 2018-03-10
Additions and revisions to ED/EDR API in PEcAn.ED2 and PEcAnRTM

ashiklom:
## Description

I created a bunch of utilities that facilitate setting up and running ED2 directly from R, without having to go through the entire PEcAn workflow. The three main classes of utilities are for working with ED meteorology (`ed_metheader`), ED vegetation inputs (`ed_veg`), and ED2IN file (`ed2in`). All three of these classes come with methods for reading, writing, modifying, and checking the associated objects.

The new vignette demonstrates an example workflow for running ED offline from PEcAn.

This PR also includes some related changes to the EDR system. The process of doing the EDR PDA has made me realize that I made a lot of poor design decisions in my first pass at the EDR R wrapper (particularly, combining EDR setup and execution into one function) that these changes try to resolve.

I imagine a review of this will take some time since it's a pretty hefty PR, but **please hold off on merging this** until I've tested it with the full PEcAn workflow. I'll leave a comment when that's been done.

## Motivation and Context

Mostly, this is to facilitate working with ED offline from the full PEcAn system, which you may want to do for a number of reasons. However, the utilities for writing and checking ED inputs are valuable to the PEcAn workflow.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


models/ed/R/met2model.ED2.R
```diff
@@ -274,31 +277,33 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
+      nlat = 1,
+      dx = 1,
+      dy = 1,
+      xmin = lon,
+      ymin = lat,
+      variables = metvar_table
+    ))
+
+    check_ed_metheader(ed_metheader)
+    write_ed_metheader(ed_metheader, met_header)
```
istfer: now these two object names are two similar, maybe we can call the latter `met_header_file` or something?

models/ed/R/write_ed_metheader.R
```diff
@@ -0,0 +1,41 @@
...
+    block_lines[6] <- paste(metformat$variables$flag, collapse = " ")
+    blocks[[i]] <- block_lines
+  }
+  format_names <- paste("format", seq_len(nformats), sep = "", collapse = ", ")
+  file_lines <- c(format_names, as.character(nformats), Reduce(c, blocks))
+  writeLines(file_lines, filename)
+}
+
+# TODO: First line does actually matter -- maybe a list of format names?
+# Regardless, need to set it to something
```
istfer: the first line of my header is `"header"`, will ED expect something there?

ashiklom: As far as I can tell, those are just metadata that are ignored by ED. I had a version of a header file at one point whose header was something like `#"header"`, which should fail if ED uses it for anything useful.

That said, we may be able to put some useful information in there as metadata. Perhaps info about the original met product?

models/ed/R/met2model.ED2.R
ashiklom: Addressed in 7857f5c.

models/ed/R/write.configs.ed.R
```diff
@@ -211,77 +229,63 @@ write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings
...
-
   ##---------------------------------------------------------------------
   # Modify any additional tags provided in settings$model$ed2in
-  ed2in.text <- ed2in_set_value_list(settings$model$ed2in, ed2in.text, 
-                                     "PEcAn: Custom argument from pecan.xml")
+  ed2in.text <- modify_ed2in(ed2in.text, .dots = settings$model$ed2in)
   
   ##----------------------------------------------------------------------
-  writeLines(ed2in.text, con = file.path(settings$rundir, run.id, "ED2IN"))
+  check_ed2in(ed2in.text)
```
istfer: So I guess you created these sets of functions to work locally. And write.configs work locally, but ED2IN paths could be remote paths, right? so I assume this function will give errors because it tries to read remote files for some of the checks, maybe pass a flag for remote and skip those checks?

ashiklom: Good point, I didn't consider that. I'll add a `check` argument to all of these functions, and then they can be called with something like `met2model.ED2(..., check = !is_remote)`.

models/ed/R/write_restart.ED2.R
```diff
@@ -128,26 +128,19 @@ write_restart.ED2 <- function(outdir,
...
-
-  writeLines(ed2in_new, file.path(ed2in_path))
+  ed2in_orig <- read_ed2in(ed2in_path, check = FALSE)
+
+  ed2in_new <- modify_ed2in(
+    ed2in_orig,
+    start_date = start.time,
+    end_date = stop.time,
+    RUNTYPE = "HISTORY",
+    IED_INIT_MODE = 5,
```
istfer: I'm asking this out of curiosity, why is this 5 but not 4?

models/ed/R/write.configs.ed.R
ashiklom: Addressed in d463b1f.

models/ed/R/write_restart.ED2.R
ashiklom: Oof, I did this ages ago, back at the Berkeley hackathon, and I don't remember why I picked this exact value. Maybe because it's more flexible than 4? The ED2IN documentation says this:

```
   !    4.  Same as 1, but using ED2.1 H5 history/state files that take the form:          !
   !        'dir/prefix-gxx.h5'                                                            !
   !        Initialization files MUST end with -gxx.h5 where xx is a two digit integer     !
   !        grid number.  Each grid has its own initialization file.  As an example, if a  !
   !        user has two files to initialize their grids with:                             !
   !        example_file_init-g01.h5 and example_file_init-g02.h5                          !
   !        NL%SFILIN = 'example_file_init'                                                !
   !                                                                                       !
   !    5.  This is similar to option 4, except that you may provide several files         !
   !        (including a mix of regional and POI runs, each file ending at a different     !
   !        date).  This will not check date nor grid structure, it will simply read all   !
   !        polygons and match the nearest neighbour to each polygon of your run.  SFILIN  !
   !        must have the directory common to all history files that are sought to be used,! 
   !        up to the last character the files have in common.  For example if your files  !
   !        are                                                                            !
   !        /mypath/P0001-S-2000-01-01-000000-g01.h5,                                      !
   !        /mypath/P0002-S-1966-01-01-000000-g02.h5,                                      !
   !        ...                                                                            !
   !        /mypath/P1000-S-1687-01-01-000000-g01.h5:                                      !
   !        NL%SFILIN = '/mypath/P'                                                        !
```

istfer: yeah as we only have one grid (g001, is that correct?) and `N_POI =1` I was wondering if ours is case-4 and selecting 5 would confuse restart, but maybe it doesn't matter..anyway, I'll be testing it soon

ashiklom: Makes sense. The last time I touched this code was when I first developed it, and it _seemed_ to do what I wanted then, but I don't think I ever tried to do any real state data assimilation. Feel free to change it to whatever makes it work (or might even make sense to tweak `write_restart` to take it as an argument).

models/ed/DESCRIPTION
```diff
@@ -24,11 +24,17 @@ Imports:
     stringr(>= 1.1.0),
     udunits2 (>= 0.11),
     XML (>= 3.98-1.4),
-    hdf5r
+    hdf5r,
```
serbinsh: so have we dropped the old HDF5 package now?

serbinsh: We may need to update the install.depends script to account for this change.

models/ed/R/write_restart.ED2.R
serbinsh: I though we always use INIT mode 3? @mdietze

models/ed/DESCRIPTION
ashiklom: Yes, `rhdf5` is more difficult to install (not on CRAN), and may not be actively maintained anymore. This one is actively maintained, and IMHO provides a more convenient interface.

models/ed/R/write_restart.ED2.R
ashiklom: My understanding is that we need 4 or 5 for state data assimilation, so ED reads more information about the last state from the history files.

ashiklom:
Tested this on the VM and it seems to work fine, so I think this is good to go.

PR #1889, opened 2018-04-02
Fixes #1886, assignment to vector when NA error

### LINKED ISSUE #1886: NA in VPD makes metgapfill fail.
@ankur In metgapfill, I'm getting an error: 
```
Error in VPD[VPD > es] <- es : NAs are not allowed in subscripted assignments
```  
and then it exits met.process. Looking at VPD there are 11 values in a row that are NA.  

## Context
I was doing a run at Niwot Ridge from 2003-2006 using Ameriflux and the error shows up when processing 2005.

## Possible Implementation
For solutions should I tell the user to select new dates and tell which date is problematic, or should there be fix to the code to somehow bypass the NAs?

ankurdesai:
Closes #1886 

## Description
Assignment subsetting when the assigning values are a vector fail if the subset includes NA. Using the which statement fixes that. metgapfill should work better now.

### DIFF
--- a/modules/data.atmosphere/R/metgapfill.R
+++ b/modules/data.atmosphere/R/metgapfill.R
@@ -302,10 +302,12 @@ metgapfill <- function(in.path, in.prefix, outfolder, start_date, end_date, lst 
 
     ##Once all are filled, do one more consistency check
     es <- get.es(Tair_degC) * 100
+
     rH[rH < 0] <- 0
     rH[rH > 100] <- 100
     VPD[VPD < 0] <- 0
-    VPD[VPD > es] <- es
+    badVPD_es <- which(VPD > es)
+    VPD[badVPD_es] <- es[badVPD_es]
     sHum[sHum < 0] <- 0
     
     ## one set of these must exist (either wind_speed or east+north wind)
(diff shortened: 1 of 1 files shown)


tonygardella:
Thank you! Can you make the changes to the release branch instead? 
```
git fetch upstream
git checkout -b release/v1.5.3 upstream/release/v1.5.3
#make your changes 
git push origin release/v1.5.3
```
Then open a PR comparing to the `release/v1.5.3` branch?

modules/data.atmosphere/R/metgapfill.R
```diff
@@ -302,10 +302,12 @@ metgapfill <- function(in.path, in.prefix, outfolder, start_date, end_date, lst
 
     ##Once all are filled, do one more consistency check
     es <- get.es(Tair_degC) * 100
+
     rH[rH < 0] <- 0
     rH[rH > 100] <- 100
     VPD[VPD < 0] <- 0
-    VPD[VPD > es] <- es
+    badVPD_es <- which(VPD > es)
+    VPD[badVPD_es] <- es[badVPD_es]
```
infotroph: If you're using which just for the purpose of excluding NAs, you could consider doing the exclusion explicitly: `VPD[VPD > es & !is.na(VPD)] <- es`.

ankurdesai: Your solution doesn't work, VPD>es & !is.na(VPD) will return TRUE/FALSE/NA, and since es is a vector of more than one element, will trigger the same error. I tested this explicitly.

ankurdesai:
@tonygardella I followed your instructions. When creating the pull request it seemed to add a bunch of other commits from other folks. Did I do this right? See #1890

tonygardella:
Sorryy I should of added the `git add` and `git commit` after you make your changes to the code. Also when I say compare your PR to `release/v1.5.3`, you can pick which branch is the base branch you are comparing your code changes to. The default is "develop" but for this you'll need to select `release/v1.5.3` from the drop down menu in the PR github interface.

modules/data.atmosphere/R/metgapfill.R
infotroph: Gotcha -- I missed that es had more than one element. Sorry for the noise.

tonygardella:
I'm going to close this in favor of #1890

PR #1890, opened 2018-04-03
Release/v1.5.3

### LINKED ISSUE #1886: NA in VPD makes metgapfill fail.

ankurdesai:
Not sure I'm doing this right. Following Tony's instructions to move my commit to close #1886 from develop to 1.5.3.

### DIFF
--- a/modules/data.atmosphere/R/metgapfill.R
+++ b/modules/data.atmosphere/R/metgapfill.R
@@ -305,8 +305,8 @@ metgapfill <- function(in.path, in.prefix, outfolder, start_date, end_date, lst 
     rH[rH < 0] <- 0
     rH[rH > 100] <- 100
     VPD[VPD < 0] <- 0
-    badVPD_es <- which(VPD > es)
-    VPD[badVPD_es] <- es[badVPD_es] 
+    badVPD_es <- which(VPD > es)  
+    VPD[badVPD_es] <- es[badVPD_es]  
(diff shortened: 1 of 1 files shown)


mdietze:
@ankurdesai this PR seems to just be moving the contents of PecanProject:release/v1.5.3 into develop via your personal branch, but without the fixes discussed in #1889. Because it's not different from release Github won't let me change the branch the PR is going into. However, when I try to change #1889 to pull into release instead of develop, then there are >100 changes rather than just the one. So... I think you need to manually apply the changes you did in #1189 into this branch, then commit and push, after which point we should be able to change the PR to be into release and be done.

ankurdesai:
That's what I thought I did, but I'll try again.

tonygardella:
@ankurdesai I changed the base and it looks like everything is fine now.

ankurdesai:
Thanks. I think it's ready now.

PR #1897, opened 2018-04-06
read_restart and write_restart for ED-SDA coupling

istfer:
## Description

This is a first running version of ED-SDA workflow with four important things to note:
1) The current implementation is far from a sensible SDA: The variable associations (i.e. which ED variables corresponds to the assimilated data variables) were my best initial guesses but not thought through, the variables passed to the state matrix are not definitive, the adjustment to the ED2's state variables (actually, only one variable `NPLANT`) in write_restart is super naive etc.
2) We plan to run SDA workflow locally (i.e. current sda.enkf function will run and complete on wherever the host is: either local machine or remote machine, and only objects for visualization will be returned, or something similar to that logic). But this version for ED2 submits runs to remote machine, syncs back outputs to local machine, runs the analysis on local machine and submits next time step runs again to the remote.
3) Because of these remote/local differences I didn't bother to adjust the met ensemble sampling code (`sample_met` function works locally and it's a temporary function anyway).
4) I haven't tackled the IC part yet.

But the SDA workflow should be able to run back-to-back for ED2 now. I'm going to do a test over the weekend.

EDIT: The test was successful. It runs back-to-back, but GWBI is problematic at this point. 
plot_zoom_png

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


tonygardella:
@araiho Adding you because of your familiarity with SDA and pecan.

mdietze:
@tonygardella who's supposed to be paired with @istfer this week? I'm happy to give this a final look before it's merged, but I don't think we should restrict the initial review to @araiho and I.

tonygardella:
@robkooper You're paired with Istem this week for review.

models/ed/R/ed_varlookup.R
```diff
@@ -0,0 +1,50 @@
+#' Lookup function for translating commonly used ED variables
+#' returns out list, readvar variables to read from file, expr if any derivation is needed
+#' @export
+ed.var <- function(varname) {
+  if(varname == "AGB") {
+    out = list(readvar = "AGB_CO", 
+               type = 'co', units = "kgC/plant", 
+               drelated = NULL,                    # other deterministically related vars?
+               expr    = "AGB_CO")                                                 } else
```
robkooper: can you move the else to the same line as the next if so it is easier to read.

models/ed/R/met2model.ED2.R
```diff
@@ -137,8 +148,9 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
     slen <- seq_along(SW)
     Tair <- c(rep(Tair[1], toff), Tair)[slen]
     Qair <- c(rep(Qair[1], toff), Qair)[slen]
-    U    <- c(rep(U[1], toff), U)[slen]
-    V    <- c(rep(V[1], toff), V)[slen]
+    # U  <- c(rep(U[1], toff), U)[slen]
```
robkooper: just remove this code that is commented out, we always have git history to look at why it is not here. Also the comment above is fine.

models/ed/R/met2model.ED2.R
```diff
@@ -114,14 +114,25 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
-    V    <- ncdf4::ncvar_get(nc, "northward_wind")
+    U    <- try(ncdf4::ncvar_get(nc, "eastward_wind"),  silent = TRUE)
+    V    <- try(ncdf4::ncvar_get(nc, "northward_wind"), silent = TRUE)
     Rain <- ncdf4::ncvar_get(nc, "precipitation_flux")
     pres <- ncdf4::ncvar_get(nc, "air_pressure")
     SW   <- ncdf4::ncvar_get(nc, "surface_downwelling_shortwave_flux_in_air")
     LW   <- ncdf4::ncvar_get(nc, "surface_downwelling_longwave_flux_in_air")
     CO2  <- try(ncdf4::ncvar_get(nc, "mole_fraction_of_carbon_dioxide_in_air"), silent = TRUE)
 
+    use_UV <- is.numeric(U) & !is.numeric(V)
```
robkooper: does it make sense to first check and see if wind_speed is available and use that, if not available use U and V, and print error if neither is available.

models/ed/R/other.helpers.ED2.R
```diff
@@ -2,9 +2,33 @@
 #' 
 #' @author Alexey Shiklomanov
 #' @inheritParams base::list.files
+#' @export
```
robkooper: is this  a function that is needed outside of the ED code?

models/ed/R/other.helpers.ED2.R
```diff
@@ -2,9 +2,33 @@
...
     allfiles <- list.files(path, ...)
     dirs <- list.dirs(path, ...)
     outfiles <- setdiff(allfiles, dirs)
     return(outfiles)
 }
+
+
+#' Function translating pecan vars to ED vars
+#' var.names <- c("DBH", "AGB", "AbvGrndWood")
+#' @export
```
robkooper: is this a function that is needed outside of the ED code?

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -80,25 +82,42 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
     sampleIDs <- c(1:n.inputs,sample.int(n.inputs, (nens - n.inputs), replace = TRUE))
   }
   
- 
+  # currently not focusing on met ensembles for ED-SDA coupling, and recycling same met if not localhost
+  # sample_met is a temporary function anyway 
   if(is.null(restart) & is.null(restart$ens.inputs)){
-    ens.inputs <- sample_met(settings,nens)
-  } else {
+    if(settings$host$name != "localhost"){ # temporary HACK..temporary HACK..temporary HACK
```
robkooper: can you make an issue for this, so we don't loose it. Maybe in that issue also specify what a solution is for this hack.

models/ed/R/met2model.ED2.R
istfer: thanks for pointing this out, I commented them out instead of deleting as a reminder, I was meaning to ask @mdietze whether this was how he wanted me to implement met2model changes for wind

models/ed/R/met2model.ED2.R
istfer: yeah it does, the thing is we (almost always?) write `U` and `V` in the standard CFs, PalEON data is a special case. I'll revise this also after I make sure Mike is OK with this implementation overall

models/ed/R/other.helpers.ED2.R
istfer: not currently, but it could be

models/ed/R/other.helpers.ED2.R
istfer: no..but if I don't export it I have to source it everytime while I'm developing the code. Even if I load  PEcAn.ED2 it can't find the function when it's called through other functions. Maybe I'm missing something?

modules/assim.sequential/R/sda.enkf.R
istfer: I wish I could loose it 😛 

Let me see, I think this issue is relevant #1167 and also this #1455

models/ed/R/met2model.ED2.R
mdietze: This is not what intended. The CF standard is to write U & V. ED2 wants U & V. If U & V are absent then we can approximate U = wind speed and V = 0, but we definitely shouldn't be doing that by default.

Really confused by the logic of is.numeric(U) & !is.numeric(V) and when that would apply -- seems like we're either going to have both or we'll have neither, but I can't imagine why we'd have U but not V and how we could calculate sqrt(U ^ 2 + V ^ 2) if we don't have V. 

If we have both then use them. If we have neither check for wind_speed.

models/ed/R/met2model.ED2.R
mdietze: per above, this is not how I wanted this implemented. Let's revert to the previous code

models/ed/R/met2model.ED2.R
istfer: one with `!` one without was definitely a typo

OK I'm going to change it back to writing both U and V

models/ed/R/ed_varlookup.R
```diff
@@ -0,0 +1,50 @@
...
+    out = list(readvar = "DLNAGB_DT", # this is actually rate of change in AGB 
+               type = 'co', units = "1/yr", 
+               drelated = NULL,
+               expr    = "DLNAGB_DT")                                              
+  } else { # No Match!
+    warning(paste0("Couldn't find varname ", varname, "!"))
+    out = NULL
+  }
+  return(out)
+}
```
ashiklom: Personally, I think it's cleaner to use `tibble::tribble` to make this a lookup data frame:

```r
ed_var <- tibble::tribble(
    ~varname, ~readvar, ~type, ~units, ~drelated, ~expr,
    "AGB", "AGB_CO", "co", "kgC/plant", NA, "AGB_CO",
...
```

...but that's just my opinion -- no reason to change this since it works fine.

models/ed/R/met2model.ED2.R
```diff
@@ -114,14 +114,25 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
     sec  <- nc$dim$time$vals
     Tair <- ncdf4::ncvar_get(nc, "air_temperature")
     Qair <- ncdf4::ncvar_get(nc, "specific_humidity")  #humidity (kg/kg)
-    U    <- ncdf4::ncvar_get(nc, "eastward_wind")
-    V    <- ncdf4::ncvar_get(nc, "northward_wind")
+    U    <- try(ncdf4::ncvar_get(nc, "eastward_wind"),  silent = TRUE)
+    V    <- try(ncdf4::ncvar_get(nc, "northward_wind"), silent = TRUE)
```
ashiklom: No need to change, but I recently figured out how to do this with `tryCatch`. In general, it looks something like this:

```r
U <- tryCatch(
    ncdf4::ncvar_get(...),
    error = function(e) {
        logger.info("Wind is absent, so returning NULL")
        return(NULL)
    }
)
```

Here, it's a more complicated nested check, so your approach is probably better, but just wanted to point this out.

istfer:
this is ready for another review from @robkooper or @mdietze

modules/assim.sequential/R/met_filtering_helpers.R
```diff
@@ -15,8 +25,18 @@ sample_met <- function(settings,nens=1){
   tmp_members <- gsub(paste0(".", end_date),   "", tmp_members) 
   member_names <- unique(dirname(ens_members))
   
-  #this will change from model to model, generalize later
-  ens_ind <- unlist(sapply(paste0(member_names, ".clim"), grep, tmp_members))
+  # this will change from model to model, generalize later
+  # This function is temporary but if we will continue to use this approach for met ensembles (instead of met process workflow)
```
mdietze: This whole function seems to be a bit of a hack -- any time we write a function that as a hardcoded if on model type we've done something wrong. The next PR to SDA needs to get rid of (or correctly generalize) this function all together, and NOT replace it with sample_met.model, because if done correctly there's should be nothing about the process of sampling from a list of inputs that should require model-specific code. Likewise, the sampling needs to be generalized to be able to sample over all inputs, not just met, as I really hope we'll have soil ensembles in place within the year.

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -57,17 +57,19 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
   my.read_restart  <- paste0("read_restart.", model)
   my.write_restart <- paste0("write_restart.", model)
   my.split_inputs  <- paste0("split_inputs.", model)
+  # models that don't need split_inputs OR some flag passed via settings
+  # or should there be a split_inputs.ED2 doing nothing?
+  nosi <- c("ED2")
+  no_split <- model %in% nosi
```
mdietze: hack to be fixed. If necessary a split_inputs.ED2 that does nothing is a better solution than hard coding this. The other alternative is to put this info in the model registry file in the model package.

models/ed/R/write.configs.ed.R
```diff
@@ -106,7 +106,7 @@ convert.samples.ED <- function(trait.samples) {
 ##' @export
 ##' @author David LeBauer, Shawn Serbin, Carl Davidson, Alexey Shiklomanov
 ##-------------------------------------------------------------------------------------------------#
-write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings$constants, check = FALSE) {
+write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings$constants, check = FALSE, inputs = NULL, ...) {
```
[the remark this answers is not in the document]
istfer: both of our other SDA-coupled models have `inputs` as an argument in their write.configs

the modified settings bit (`settings$run$inputs$met$path <- inputs$met$path`) is used downstream to update the ED2IN:

```
  ed2in.text <- modify_ed2in(
    ed2in.text,
    latitude = as.numeric(settings$run$site$lat),
    longitude = as.numeric(settings$run$site$lon),
    met_driver = settings$run$inputs$met$path, *** HERE ***
    start_date = startdate,
    end_date = enddate,
    MET_START = metstart,
    MET_END = metend,
    IMETAVG = -1,   # See below,
    add_if_missing = TRUE,
    check_paths = check
  )
```

models/ed/R/write_restart.ED2.R
```diff
@@ -1,100 +1,90 @@
...
-  sda_suffix <- paste0("SDA.", sda_datestr)
-
-  # Get history restart file path
-  histfile <- get_restartfile.ED2(mod_outdir, runid, start.time)
-  if (is.null(histfile)) {
-    PEcAn.logger::logger.severe("Failed to find ED2 history restart file.")
-  }
-
+write_restart.ED2 <- function(outdir, runid, start.time, stop.time,
+                              settings, new.state, new.params, inputs, RENAME = TRUE) {
```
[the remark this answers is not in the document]
istfer: Yes, all other write_restarts had the same arguments, I was trying to preserve compatibility, although I did misorder them, will fix.. 

I should update `PEcAn.ModelName::write_restart.ModelName ` too then I guess?

models/ed/R/write_restart.ED2.R
```diff
@@ -1,100 +1,90 @@
...
+  
+  restart <- new.params$restart
+  
+  old.state <- restart$oldstate # hack: this will probably change in the near future, it's currently just AbvGrndWood 
+  histfile  <- restart$histfile # Get history restart file path
+  restart   <- restart$restart
```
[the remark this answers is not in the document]
istfer: I agree that it's a bit unintuitive.. But write_restart.ED2 (at least the way we do things currently) requires reading back the old state, and to read the old state we need to find out the histfile name again etc.. I thought it's easier to just pass them around and new.params was convenient

In SIPNET we were passing just params to update things deterministically but as with ED2 I can imagine other write_restarts might want things that are created/read in read_restarts? Maybe should change the name of `new.params` to something like more practical like `pass.to.wr`?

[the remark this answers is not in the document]
mdietze: "new parameters" and "other misc info that write restart needs" are categorically very different things, which is why I objected to passing things that are not parameters along with the parameter list. If there are other things that need to be passed then we should discuss how/where that occurs, rather than renaming `new.params` and muddling the development direction of joint SDA/PDA.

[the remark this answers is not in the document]
istfer: I'd be down for that..I think something for "other misc info that write restart needs" will make life a lot easier

models/ed/R/write_restart.ED2.R
```diff
@@ -1,100 +1,90 @@
...
-  names(pftnames) <- pftnums
-  names(pftnums) <- pftnames
 
-
-  nc <- ncdf4::nc_open(histfile)
-
-  #### Read common variables
+  # rename history file 
+  # (IF: At this point in order not to swamp directories I decided to write annual historu files
+  # but this causes mismatches in start-date because ED2 writes them as 1961-01-01 not 1961-12-31
```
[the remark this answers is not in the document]
istfer: yes it uses `iyeara,imontha,idatea,itimea` to generate the filename https://github.com/EDmodel/ED2/blob/master/ED/src/io/h5_output.F90#L293

I was punting ED2 fixes for now, I'm hoping to work on a more general time tracking-stamping PR for ED2 when I can find the time to do comprehensive tests

models/ed/R/write_restart.ED2.R
```diff
@@ -125,26 +115,46 @@ write_restart.ED2 <- function(outdir,
...
+  if(settings$host$name == "localhost") check_ed2in(ed2in_new)
   write_ed2in(ed2in_new, ed2in_path)
 
   # Remove old history.xml file, which job.sh looks for
-  file.remove(file.path(mod_outdir, runid, "history.xml"))
-
+  file.remove(file.path(mod_outdir, runid, "history.xml"))  # this is local
+  # have job.sh delete the old history.xml: this is temporary, this script will eventually run on remote
+  jobsh <- readLines(file.path(rundir, runid, "job.sh"),-1)
+  jobsh[17] <- paste0("rm -f ", file.path(settings$host$outdir, runid, "history.xml"))
```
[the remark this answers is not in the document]
istfer: agreed.. will do changes to the template, I was trusting that the whole SDA loop will eventually run on the remote w/o syncing back and forth, but would be more robust to have the template changes

istfer:
@mdietze  apart from the things that I'll be punting as we discussed, I think only changes to book_source remains unadressed..what exactly needs to go to book_source for this PR?

mdietze:
@istfer for the book_source you need to check to see if anything you've changed rises to the level of needing to document changes to how SDA works, what models are supported in SDA, etc.

istfer:
The only relevant part I could see in the book_source was [this table](https://github.com/PecanProject/pecan/blob/develop/book_source/06_reference/02_models/model_index.Rmd)

I tested  that it runs w/o errors with the last changes, I think the PR is ready to re-review and/or close @mdietze

models/ed/R/write.configs.ed.R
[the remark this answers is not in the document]
mdietze: > both of our other SDA-coupled models have inputs as an argument in their write.configs

Ok, but that's not actually an explanation for why you added it to `write.config.ed`, where it's not needed.

The addition of inputs to `write.config.sipnet` is a legacy of the first early attempts at SDA code, which was done in a hurry (PalEON summer course deadline) and without a solid plan in place for generalization). At that time we didn't have read/write restart functions as part of the design, so I needed a way to pass the updated state back in. I think LINKAGES, which also doesn't have a dedicated restart file, followed the same pattern to allow write.restart leverage write.configs. By contrast ED2 has a dedicated restart file and doesn't use write.configs to restart, and thus `inputs` can be dropped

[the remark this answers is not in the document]
istfer: but it is needed, not in restart loop but for the initial ensemble run at the beginning of the [SDA workflow](https://github.com/PecanProject/pecan/blob/develop/modules/assim.sequential/R/sda.enkf.R#L233_L240)

Issue #1880, opened 2018-03-21
errror in browndog shiny app


robkooper:
When viewing the browndog shiny app it will not show any sites.

## Context
Following text shows in the console:
```
Uncaught TypeError: Cannot read property 'lat' of null
    at Object.project (leaflet.js:6)
    at Object.latLngToPoint (leaflet.js:6)
    at e.project (leaflet.js:6)
    at e.latLngToLayerPoint (leaflet.js:6)
    at e.update (leaflet.js:7)
    at e.onAdd (leaflet.js:7)
    at e._layerAdd (leaflet.js:6)
    at e.addLayer (leaflet.js:6)
    at e.addLayer (leaflet.js:7)
    at LayerStore.add (binding.js:462)
```

## Possible Implementation
Looks like we need to make sure to filter those locations that have a null for either lat or lon.

PR #1900, opened 2018-04-10
remove site without geometry; fix sites.length==0; remove old map mar…

yan130:
## Description
remove site without geometry from SQL;
fix sites.length==0, for most input$type cannot be found in bety
remove old map markers when changing input$type 
fix agreement bug, if changing from "Ameriflux" without check the agreement, then switch back "AmerifluxLBL"


robkooper:
See issue #1880

PR #1903, opened 2018-04-10
Cultivar-PFT support for priordupe

infotroph:
## Description

Updates the confusingly named `PEcAn.DB::priordupe`, which is an unexported function for duplicating existing *PFTs* (but that does include their *associations* with priors):

* Added support for cultivar-PFTs.
* Overhauled the queries to use dplyr
* Modified the return value: Formerly returned nothing, now returns the Bety ID of the newly created PFT. I've marked this a "breaking change" below, but shouldn't hurt anything because priordupe is unused in the current PEcAn code.

## Motivation and Context

Necessary because of #1826 (cultivars in PFTs), for two different reasons: 80% because there's not yet a web interface for creating new cultivar PFTs from scratch, so duplicating them from existing ones is the easiest way to do it. The other 20% is just that `priordupe` is the last code (that I know of) that was broken by #1826, so even thought it's rarely used I'll give myself a gold star when it's fixed.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


dlebauer:
p.s. I would have added priordupe as an alias, except that I am pretty sure no one has been actively using this function since the button to do the same was added web interface a very long time ago.

infotroph:
👍 I've updated the filenames to match.

base/db/R/clone_pft.R
```diff
@@ -0,0 +1,105 @@
...
+##' @examples \dontrun{
+##' clone_pft(parent.pft.name    = "tempdecid",
+##'           new.pft.name       = "mytempdecid",
+##'           new.pft.definition = "mytempdecid is a new pft",
+##'           settings = pecan_settings_list)
+##' }
+clone_pft <- function(parent.pft.name,
+                      new.pft.name,
+                      new.pft.definition,
+                      settings){
```
ashiklom: I still think that, in general, we should move away from `settings` arguments in favor of more explicit arguments, since doing it this way obscures the things actually required for this to work if you want to do it outside of the PEcAn workflow. But that's just, like, my opinion, man.

base/db/R/clone_pft.R
```diff
@@ -0,0 +1,105 @@
...
+  if (length(new.pft.name) > 1 || length(parent.pft.name) > 1) {
+    PEcAn.logger::logger.severe("multiple PFT names given, and clone_pft only knows how to handle one at a time.")
+  }
+
+  con <- db.open(settings$database$bety)
+  on.exit(db.close(con))
+
+  parent.pft <- (dplyr::tbl(con, "pfts")
+    %>% dplyr::filter(name == parent.pft.name)
+    %>% dplyr::collect())
```
ashiklom: No need to change, because this works and it's perfectly clear what's going on...but why wrap these pipes in parentheses? I don't think I've ever seen anyone do this. Just curious.

base/db/R/clone_pft.R
```diff
@@ -0,0 +1,105 @@
...
+                      new.pft.definition,
+                      settings){
+
+  if (new.pft.name == parent.pft.name) {
+    PEcAn.logger::logger.severe("new.pft.name must not be the same as parent.pft.name")
+  }
+
+  if (length(new.pft.name) > 1 || length(parent.pft.name) > 1) {
+    PEcAn.logger::logger.severe("multiple PFT names given, and clone_pft only knows how to handle one at a time.")
+  }
```
ashiklom: A little while ago, I added functions `PEcAn.logger::logifnot/.../severeifnot`. Definitely no need to change this, but I'm curious what you think about that approach? I thought it might be a more concise way to do this frequent construct, but it might not be in practice:

```r
PEcAn.logger::severeifnot(
    length(new.pft.name) > 1 || length(parent.pft.name) > 1,
    msg = "Multiple PFT names given...
)
```

base/db/R/clone_pft.R
```diff
@@ -0,0 +1,105 @@
...
+    %>% dplyr::collect())
+
+  if(nrow(new_members) > 0){
+    DBI::dbWriteTable(
+      conn = con,
+      name = member_tbl,
+      value = as.data.frame(new_members),
+      append = TRUE,
+      row.names = FALSE)
+  }
```
ashiklom: Also no need to change, but this sort of "create new table, anti-join against SQL table, and `dbWriteTable` on remainer" operation is what [`PEcAn.db::db_merge_into`](https://github.com/PecanProject/pecan/blob/develop/base/db/R/db_merge_into.R) is meant to do. Any good reason not to use that here? (Again, no need to change -- just interested in limits of that function's utility, with the eventual goal of refactoring a lot of our DB code to use higher level utilities like that).

base/db/R/clone_pft.R
infotroph: I completely missed that you'd added these! I like them for clarity, even if it is verbose like all the logger calls are

base/db/R/clone_pft.R
infotroph: In multiline expressions I strongly prefer to have infix operators at the left end of the line, so the parens are just to make sure the whole pipeline is evaluated as a unit. Do we have a formal style standard about this? If not I'll argue that we should adopt my way :)

dlebauer: In lieu of formal style we should defer to convention. And PEcAn style should defer to the broader community conventions by default - unless there is a compelling reason to do otherwise.

ashiklom: To the extent that we follow the [Tidyverse style guide](http://style.tidyverse.org/pipes.html), the style is this:

```r
rabbit %>%
  hop() %>%
  jump()
```

That said, I have no desire to be dogmatic about it -- I think other aspects of coding style are much more important, and the code is perfectly clear and readable as is.

Issue #1476, opened 2017-06-19
NARR THREDDS


ankurdesai:
Extract site/region level NARR surface variables using their THREDDS interface, re: discussion on 6/19 telecon with @dlebauer and @mdietze . 

## Description
Replaces our current direct download of NARR, which is timeconsuming. Here is the NARR THREDDS location:
  https://www.esrl.noaa.gov/psd/thredds/catalog/Datasets/NARR/catalog.html

Monolevel contains the surface variables, and subsurface are soil variables.

## Context
Similar to CRUNCEP, this would be another way to acquire a regularly used met variable that is 3-hourly and 32 km resolution spatial from 1979 to present. CRUNCEP is essentially a merger of NARR temporal and CRU gridded data.

## Possible Implementation
Should follow CRUNCEP-MSTIMIP pretty closely.

PR #1904, opened 2018-04-10
data.atmosphere: Download NARR using THREDDS

ashiklom:
## Description

Implements downloading of NARR via THREDDS.

## Motivation and Context

Downloading all 50 bajillion terabytes of NARR is dumb and slow. This leverages THREDDS to make the download efficient. See #1476.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/download.NARR_thredds.R
```diff
@@ -0,0 +1,260 @@
...
+    msg = paste("Latitude must be a scalar, but has length", length(latitude))
+  )
+
+  PEcAn.logger::severeifnot(
+    length(lon.in) == 1,
+    msg = paste("Longitude must be a scalar, but has length", length(longitude))
+  )
+
+  narr_start <- lubridate::ymd("1979-01-01")
+  narr_end <- lubridate::ymd("2018-03-31")
```
ankurdesai: So NARR is updated pretty frequently, monthly I think. For checking end date, I'd rather it be something like if date < (current date - one month) or something like that. Otherwise, this code will prevent runs past March 2018.

modules/data.atmosphere/R/download.NARR_thredds.R
```diff
@@ -0,0 +1,260 @@
...
+#' @param start_date Start date for meteorology
+#' @param end_date End date for meteorology
+#' @param lat.in Latitude coordinate
+#' @param lon.in Longitude coordinate
+#' @param progress Whether or not to show a progress bar (default = `TRUE`)
+#' @return `tibble` containing time series of NARR data for the given site
+#' @author Alexey Shiklomanov
+#' @examples
+#' dat <- get_NARR_thredds("2008-01-01", "2008-01-15", 43.3724, -89.9071)
+get_NARR_thredds <- function(start_date, end_date, lat.in, lon.in,
```
ankurdesai: To be consistent with met.process functions , it should be named download.NARR_threads (like the name of the filename), no?  Or is the idea, that download.NARR would call this function? If not, then an outfolder needs to be specified download.XXX <- function(outfolder, start_date, end_date, site_id, lat.in, lon.in, overwrite = FALSE, verbose = FALSE) and a "results" array needs to be returned.

ashiklom:
Yup, the idea was to write/modify download.NARR later to just be like 5
lines that call this function and save the output. Just haven't gotten to
that yet, but wanted you to take a look at this first. I like having this
be a separate function because I've run into use cases where I just want to
be able to grab met data for a late/lon independent of the pecan workflow.

On Apr 10, 2018 9:47 PM, "Ankur Desai" <notifications@github.com> wrote:

*@ankurdesai* commented on this pull request.
------------------------------

In modules/data.atmosphere/R/download.NARR_thredds.R
<https://github.com/PecanProject/pecan/pull/1904#discussion_r180613929>:

> +    length(end_date) == 1,
+    msg = paste("End date must be a scalar, but has length", length(end_date))
+  )
+
+  PEcAn.logger::severeifnot(
+    length(lat.in) == 1,
+    msg = paste("Latitude must be a scalar, but has length", length(latitude))
+  )
+
+  PEcAn.logger::severeifnot(
+    length(lon.in) == 1,
+    msg = paste("Longitude must be a scalar, but has length",
length(longitude))
+  )
+
+  narr_start <- lubridate::ymd("1979-01-01")
+  narr_end <- lubridate::ymd("2018-03-31")

So NARR is updated pretty frequently, monthly I think. For checking end
date, I'd rather it be something like if date < (current date - one month)
or something like that. Otherwise, this code will prevent runs past March
2018.
------------------------------

In modules/data.atmosphere/R/download.NARR_thredds.R
<https://github.com/PecanProject/pecan/pull/1904#discussion_r180614821>:

> @@ -0,0 +1,260 @@
+#' Retrieve NARR data using thredds
+#'
+#' @param start_date Start date for meteorology
+#' @param end_date End date for meteorology
+#' @param lat.in Latitude coordinate
+#' @param lon.in Longitude coordinate
+#' @param progress Whether or not to show a progress bar (default = `TRUE`)
+#' @return `tibble` containing time series of NARR data for the given site
+#' @examples
+#' dat <- get_NARR_thredds("2008-01-01", "2008-01-15", 43.3724, -89.9071)
+get_NARR_thredds <- function(start_date, end_date, lat.in, lon.in,

To be consistent with met.process functions , it should be named
download.NARR_threads (like the name of the filename), no? Or is the idea,
that download.NARR would call this function? If not, then an outfolder
needs to be specified download.XXX <- function(outfolder, start_date,
end_date, site_id, lat.in, lon.in, overwrite = FALSE, verbose = FALSE) and
a "results" array needs to be returned.

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/PecanProject/pecan/pull/1904#pullrequestreview-111064074>,
or mute the thread

ashiklom:
And good to know NARR is updated monthly -- I'll change that check
accordingly.

ashiklom:
Also, just realized I forgot to `@export` the main function, so I'll need to do that before this gets merged.

modules/data.atmosphere/R/NARR_csv.R
```diff
@@ -0,0 +1,315 @@
...
+#' @param overwrite Overwrite existing files?  Default=FALSE
+#' @param verbose Turn on verbose output? Default=FALSE
+#' 
+#' @examples
+#' download.NARR_csv(tempdir(), "2001-01-01", "2001-01-12", 43.372, -89.907)
+#' 
+#' @export
+#'
+#' @author Alexey Shiklomanov
+download.NARR_csv <- function(outfolder, start_date, end_date, lat.in, lon.in, overwrite = FALSE, verbose = FALSE, ...) {
```
ankurdesai: Is there a particular reason you'd want to download a CSV (other than it is useful now and then). If want to plug into met.process, this adds the extra step of then running met2CF.csv, which then needs the format defined in Formats table. I

ashiklom: Nope. I'll change it accordingly.

ashiklom:
@ankurdesai OK, just changed this to download to NetCDF, modeled on the way CRUNCEP does this, but with a more functional programming approach.

modules/data.atmosphere/NAMESPACE
```diff
@@ -81,6 +85,7 @@ export(wide2long)
 importFrom(PEcAn.DB,db.close)
 importFrom(PEcAn.DB,db.query)
 importFrom(PEcAn.DB,dbfile.input.insert)
+importFrom(magrittr,"%>%")
```
infotroph: For imports like this that are likely to get used in multiple places, I'd prefer we make a habit of putting them in a place that clearly applies to the whole package, e.g. in  PEcAn.DB we put the importFrom in the [package documentation block](https://github.com/PecanProject/pecan/blob/develop/base/db/R/zzz.R#L8). 

That said, maybe it makes sense to import it here first and then move it when it's actually being used multiple places in the package.

modules/data.atmosphere/R/download.NARR.R
```diff
@@ -7,6 +7,7 @@
 ##' @param verbose Turn on verbose output? Default=FALSE
 ##' @param method Method of file retrieval. Can set this using the options(download.ftp.method=[method]) in your Rprofile.
 ##' example options(download.ftp.method="ncftpget")
+##' @importFrom magrittr %>%
```
infotroph: For imports like this that are likely to get used in multiple places, I'd prefer we make a habit of putting them in a place that clearly applies to the whole package, e.g. in  PEcAn.DB we put the importFrom in the [package documentation block](https://github.com/PecanProject/pecan/blob/develop/base/db/R/zzz.R#L8). 

That said, maybe it makes sense to import it here first and then move it when it's actually being used multiple places in the package.

tonygardella:
@ashiklom Is this good to merge or are you going to add anything?

ashiklom:
Since it's still disconnected from the PEcAn workflow, and it seems to have
worked for me, I don't think I'm gonna mess with it. There were a few
significant changes that happened since it was last reviewed, so it might
be worth another look by someone if they have the time.

tonygardella:
@ashiklom Just the two conflicts and I can merge.

ashiklom:
Addressed.

PR #1907, opened 2018-04-13
added new traits to dictionary

dlebauer:
Added variables to trait dictionary so plots have useful labels 

This will no longer be necessary after we implement https://github.com/PecanProject/pecan/issues/729


tonygardella:
@dlebauer Could you resolve the conflict so we can merge?

PR #1908, opened 2018-04-16
Geostreams update

infotroph:
## Description

* <del>Call Clowder `geostreams/sensors/:id/update` endpoint before checking data availability. </de>
* Pass start/end dates in strict T-separated ISO 8601 form to avoid surprising date range results
* Fix swapped lat/lon 

## Motivation and Context

Per @max-zilla, the Geostreams API only recalculates min_start_time and max_end_time for a sensor when specifically requested, so they are not guaranteed to be accurate -- or even populated at all -- if queried without first calling update.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/met2CF.Geostreams.R
```diff
@@ -53,12 +53,12 @@ met2CF.Geostreams <- function(in.path, in.prefix, outfolder,
 
     if (length(unique(dat$geometry.coordinates)) == 1) {
       # all lat/lons are are identical-- no need to store extra copies
-      raw_lat <- dat$geometry.coordinates[[1]][[1]]
```
robkooper: Just making sure, what version of clowder is this tested with? We had a bug that was fixed in 1.3.3 in which lat/lon was reversed.

modules/data.atmosphere/R/download.Geostreams.R
```diff
@@ -56,6 +56,25 @@ download.Geostreams <- function(outfolder, sitename,
   sensor_txt <- httr::content(sensor_result, as = "text", encoding = "UTF-8")
   sensor_info <- jsonlite::fromJSON(sensor_txt)
   sensor_id <- sensor_info$id
+
+  # Request that Geostreams recalculate data availability for us
+  # otherwise start and end dates may be null or stale
+  update_result <- httr::GET(url = paste0(url, "/sensors/", sensor_id, "/update"),
```
robkooper: can this be made optional, maybe with default being true. If I call this function a few times in a row I might want to update it the first time, but subsequent calls can just use the precomputed values.

modules/data.atmosphere/R/met2CF.Geostreams.R
infotroph: I don't know which version I'm testing against -- is there a version endpoint I can check? -- but I can confirm that the files I downloaded today use [lon, lat] everywhere whereas I just deleted some files from last July that used [lon, lat] in their sensor info followed by datapoints in [lat, lon]. So yes, this is changing to the newer standard.

dlebauer: @robkooper we are using geostreams in the terraref database (1.x#420 branch:develop sha1:e91643f).

modules/data.atmosphere/R/download.Geostreams.R
infotroph: I went with the update-every-time theory after a discussion in the TERRA project Slack where @dlebauer posed that it's not PEcAn's job to know whether Clowder needs to recompute a value before sending it, and @max-zilla replied that `sensors/:id/update` is 'basically harmless'. If it's not harmless, we should revisit whether to call it at all. I see three possible alternatives:

* "Servers don't lie": Never call update, treat the reported availability dates as authoritative, and let users complain to Clowder admins if stale timestamps mean they can't see the data they want. 
* "Are you _sure_ you don't have that?": Compare user dates against the availability reported in the initial site lookup, and if user dates fall outside the reported range then call update and check again before giving up.
* "Servers don't lie but they do lag sometimes": Assume that data are added in ~real time and that the server updates at some regular interval, and call update only if `sensor_info$max_end_time < (Sys.time() - ASSUMED_UPDATE_INTERVAL)`. 

It doesn't change any of the above, but re "If I call this function a few times in a row": Note that for the use case of downloading multiple years from one site, the update already happens only once per `download.Geostreams` call and isn't repeated for each year.

robkooper: Most of the code that writes to the database will also update the sensor information, they know best when things up date. As @dlebauer said we can call this endpoint from a cronjob every so often if need to. In which case I'd go with option 1 "Servers don't lie".

As for the option it is not a no-op, however calling it for a single sensor does cut down the compute time, but the more datapoints there are the more expensive the function is. (it has to loop through all datapoints for a sensor, finding start time, end time and variables and combine all those values).

infotroph:
I've implemented the "servers don't lie" approach by removing the update calls and requesting that the Terraref team set up a cron job on their Clowder server -- frequency and degree-of-guarantee on the update are up them, but from PEcAn's perspective it's back to being the API's problem.

PR #1927, opened 2018-05-06
download.GFDL: CF-compliant time units, download only requested dates

infotroph:
## Description

* Record time units as seconds since current_year-01-01, not "secs" (#1347). (cc @dlebauer this fixes GFDL->BioCro conversion errors)
* Adds basic tests for GFDL download, skipped on Travis for speed
* Download only the observations from the year of current interest (was downloading whole 5-year block and then subsetting)

Please review commits 13e8956 and d544380 extra-critically -- My changes pass all of my tests, but I feel like that `if(year %%5...` chain is a great example of "don't let me delete it until I understand why it was there in the first place"

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/download.GFDL.R
+++ b/modules/data.atmosphere/R/download.GFDL.R
@@ -24,6 +24,7 @@ download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.
 
   start_year <- lubridate::year(start_date)
   end_year   <- lubridate::year(end_date)
+  obs_per_year <- 365 * 24 /3 # 3-hr intervals, leap days ignored
 
   #Fix Outfolder to include model and scenario
   folder_name <- paste0("GFDL_", model, "_", scenario, "_", ensemble_member)
@@ -75,7 +76,8 @@ download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.
 
   for (i in seq_len(rows)) {
     year <- ylist[i]
-    ntime <- 14600
+    # find start position of currently-wanted year in the 5-year DAP file
+    time_offset <- 1 + ((year-1) %% 5) * obs_per_year
 
     PEcAn.logger::logger.debug(
       sprintf(
@@ -144,8 +146,8 @@ download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.
       dap_file <- paste0(dap_base, dap_end)
       dap <- ncdf4::nc_open(dap_file)
       dat.list[[j]] <- ncdf4::ncvar_get(dap, as.character(var$DAP.name[j]),
-                                 c(lon_GFDL, lat_GFDL, 1),
-                                 c(1, 1, ntime))
+                                 start = c(lon_GFDL, lat_GFDL, time_offset),
+                                 count = c(1, 1, obs_per_year))
       var.list[[j]] <- ncdf4::ncvar_def(name = as.character(var$CF.name[j]),
                                  units = as.character(var$units[j]),
                                  dim = dim,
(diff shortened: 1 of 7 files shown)


modules/data.atmosphere/R/download.GFDL.R
```diff
@@ -142,10 +144,21 @@ download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.
...
-      dap <- ncdf4::nc_open(dap_file)
+      dap <- ncdf4::nc_open(dap_file, suppress_dimvals = TRUE)
+
+      # Sanity check:
+      # We're saving the data with timestamps at the end of the interval,
+      # while GFDL-supplied timestamps vary slightly -- some vars are
+      # timestamped in middle of interval, others at end.
+      # But if these disagree by more than 3 hours, we have a problem.
+      raw_time <- ncdf4::ncvar_get(dap, "time", start = time_offset, count = obs_per_year)
+      converted_time <- udunits2::ud.convert(raw_time, dap$dim$time$units, dim$time$units)
```
ashiklom: I _think_ this is fine here, but you want to be _very_ careful using `ud.convert` on times. `ud.convert` is not designed for calendar conversions, but rather for _physical_ conversions, so it knows, for example, that there are not exactly 365 days in a year:

```
ud.convert(1, "year", "days")
# [1] 365.2422
```

We may want to invest some time into our own wrapper around `lubridate` and other utilities that know how to take into account some of these met-specific quirks (e.g. 360 day years, no leap years, etc.).

modules/data.atmosphere/tests/testthat/helper.R
```diff
@@ -0,0 +1,37 @@
...
+	expect(
+		grepl(regexp, msg,  ...),
+		sprintf(
+			"%s does not match %s.\nActual value: \"%s\"",
+			label,
+			encodeString(regexp, quote = "\""),
+			encodeString(msg)))
+
+	invisible(val)
+}
```
ashiklom: Eventually, I would move this function to `PEcAn.utils` or `PEcAn.logger`. Doesn't have to hold up this PR though.

modules/data.atmosphere/R/download.GFDL.R
infotroph: This is a good point, but also one I'm sore about: the the udunits documentation [specifically discourages](https://www.unidata.ucar.edu/software/udunits/udunits-current/doc/udunits/udunits2lib.html#Time) using it for time, but the netCDF time units were [literally designed to be converted by udunits](https://www.unidata.ucar.edu/software/netcdf/time/recs.html) and the CF conventions [incorporate the udunits definitions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch04s04.html) but then recommend "caution" with units like year and month that vary in length. Bottom line: a better wrapper would be great, but I'm not sure it can save us from an overflexible standard.

modules/data.atmosphere/R/download.GFDL.R
```diff
@@ -142,10 +144,21 @@ download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.
...
+      PEcAn.logger::errorifnot(all(diff(converted_time) == 3 * 60 * 60))
+      PEcAn.logger::errorifnot(all(abs(dim$time$vals - converted_time) < (3 * 60 * 60)))
```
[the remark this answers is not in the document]
infotroph: ...I just realized the `PEcAn::logger::*ifnot` function take `msg` as their first argument and only test `...` as conditions, so  these errors will never fire. 

1. Do not merge until I fix this.
2. This argument order seems backwards both when comparing to `stopifnot` (accepts conditions first and only) and to `logger.error` (... is for more message details), and I'd suggest we change it soon before these functions are deployed any more widely.

modules/data.atmosphere/R/download.GFDL.R
```diff
@@ -14,7 +14,7 @@
 #' @param scenario Which scenario to run (options are rcp26, rcp45, rcp60, rcp85)
 #' @param ensemble_member Which ensemble_member to initialize the run (options are r1i1p1, r3i1p1, r5i1p1)
 #' @author James Simkins, Alexey Shiklomanov, Ankur Desai
-download.GFDL <- function(outfolder, start_date, end_date, site_id, lat.in, lon.in,
+download.GFDL <- function(outfolder, start_date, end_date, lat.in, lon.in,
```
ankurdesai: Are you sure removing this is safe with met.process?

infotroph: @ankurdesai Does met.process have any expectations about how download.* will use the argument? I'm sure it's safe to remove in the sense that it won't throw an error (a `site_id` passed from upstream will be accepted, treated as part of `...`, and ignored) or change the behavior of  `download.GFDL` (it was already ignoring `site_id`), so removing it is only a problem if ignoring it is also a problem.

mdietze: Just checked a few other met download functions and `site_id` doesn't occur in most of them, and in the few that contain it may be in error (GLDAS, NLDAS) as they're using `site_id` to append to the outfolder name, but I suspect (hope) that's already part of the outfolder name that's being passed in.

PR #1930, opened 2018-05-09
Release/v1.5.3

robkooper:
This is release 1.5.3 (before the accidentally merged develop). This is the release that is on the VM.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible


robkooper:
@infotroph thanks for noticing, happened to have that sitting on my laptop so managed to recover that specific commit as well. I think we are now good to go.

tonygardella:
Error in build
```
Error : package ‘commonmark’ was installed by an R version with different internals; it needs to be reinstalled for use with this R version
ERROR: lazy loading failed for package ‘roxygen2’
* removing ‘/home/travis/R/Library/roxygen2’
```

infotroph:
@robkooper I think you'll need to specify R 3.4 (like in 266c95d31)  until apt packages catch up. Note that `ppa:marutter/c2d4u` is still 3.4 only and its 3.5 equivalent is being (slowly) populated at `ppa:marutter/c2d4u3.5`

PR #1940, opened 2018-05-18
load.bety.sh: In the test that checks whether DATABASE exists, avoid …

gsrohde:
## Description
This is a one-line fix to `load.bety.sh` that fixes the test of whether the database being loaded into exists.

## Motivation and Context
The current test issues false positives.

For example, if we run the script using the default database name "bety" and "bety" does not
exist but "bety_mepp" does, we don't get the helpful error message on how to
create "bety" (but we get a mess of less helpful error messages later on when
the script doesn't find "bety").

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/scripts/load.bety.sh
+++ b/scripts/load.bety.sh
@@ -244,7 +244,7 @@ fi
 ID_RANGE=1000000000
 
 # before anything is done, check to make sure database exists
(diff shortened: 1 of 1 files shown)


gsrohde:
@robkooper Can you merge?  I don't have write access.

PR #1948, opened 2018-05-29
DataONE Download Util -- DO NOT MERGE (work in progress)

LiamBurke24:
The core functionality of the util is unchanged. I replaced the function that uses wget to download files from the dataONE federation with functions from dataoneR and datapack packages. 

## Motivation and Context

This solves the problem of authentication issues with wget. In my testing, wget would consistently download xml files but would throw errors when downloading .csv files. This PR consistently downloaded all file types that I tested (xml, csv) and other (e.g. eml). 

There is still room to improve the naming of the downloaded files. I already have some ideas. This PR is more focused on validating the core funcitonality. 

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -159,12 +159,20 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - Added PEcAn.utils::download.file() to allow for use of alternative FTP programs
 - Updated downloadAmeriflux and downloadNARR to make use of PEcAn.utils::download.file()
 - Added -w flag to load.bety.sh script to specify the URL to fetch the data from
+<<<<<<< HEAD
```
mdietze: need to resolve merge conflict and also make sure your changes are in the CHANGELOG

modules/data.land/R/dataone_download.R
```diff
@@ -39,9 +33,19 @@ dataone_download = function(id, filepath = "/fs/data1/pecan.data/dbfiles/", CNod
...
+    data <- datapack::getData(pkgMember) 
+    
+    if(files[i][1] == "text/csv"){ # format is stored as the first entry of the list "files"
+      base::writeLines(rawToChar(data), paste0(newdir, "/", "Data", i, ".csv")) #  If csv, add .csv 
+      
+    } else if(files[i][1] == "text/xml"){
+      base::writeLines(rawToChar(data), paste0(newdir, "/", "MetaData", i, ".xml")) # If XML add .xml and add Metadata title
+      
+    } else 
+      base::writeLines(rawToChar(data), paste0(newdir, "/", "MetaData", i)) # If other, process without extension (typically eml)
```
mdietze: I think the approach you've taken has succeeded so far because all 3 cases are fundamentally text-based formats. This rawToChar approach will fail with binary formats (netCDF, geoTIFF, etc)

web/cofig.php
```diff
@@ -0,0 +1,113 @@
+<?php
```
mdietze: I think this whole file was added accidentally -- you should never commit your local config.php because it will overwrite everyone else's local config

LiamBurke24:
Alexey had another idea about the download mechanic. He thinks that curl is supported widely enough to use one of the curl packages during the download. This might fix the filename issue. I'll commit with any progress.

web/cofig.php
LiamBurke24: Yes this was a mistake. Sorry about that. I'll fix this.

modules/data.land/R/dataone_download.R
LiamBurke24: Ok! Good to know. Alexey gave me some ideas to check out using curl based R packages instead. I'll push some commits soon.

CHANGELOG.md
LiamBurke24: Ok. Noted. Will do for next time.

LiamBurke24:
I am opening a new PR that addresses these issues today. I will delete this PR when I am done.

mdietze:
@LiamBurke24 you don't need to delete your current PR and open an new one, you just need to fix the issues within the same branch and then push them to Github. The PR will automatically be updated.

LiamBurke24:
Ok. I was unsure how to remove some of the files that I accidentally committed in this PR. So I already made a new branch and am almost done with the PR...  Is that ok?

Issue #1939, opened 2018-05-18
FATES output time mismatch with metadata


serbinsh:
**Describe the bug**
Presently in the FATES pecan standard output we define the time variable as something like: 

```	
double time(time) ;
		time:units = "days since 1995-01-01 00:00:00" ;
		time:long_name = "time" ;
		time:calendar = "standard" ;

```

However the time variable data we use from FATES output actually counts up continuously from the start year, e.g. :

```
[sserbin@modex SA-median]$ ncdump -h case.clm2.h0.1995-01-01-00000.nc | grep time
	time = UNLIMITED ; // (8760 currently)
	float time(time) ;
		time:long_name = "time" ;
		time:units = "days since 1995-01-01 00:00:00" ;
		time:calendar = "noleap" ;
		time:bounds = "time_bounds" ;
	int mcdate(time) ;
	int mcsec(time) ;
	int mdcur(time) ;
	int mscur(time) ;
	int nstep(time) ;
		nstep:long_name = "time step" ;

[sserbin@modex SA-median]$ ncdump -h case.clm2.h0.1996-01-01-00000.nc | grep time
	time = UNLIMITED ; // (8760 currently)
	float time(time) ;
		time:long_name = "time" ;
		time:units = "days since 1995-01-01 00:00:00" ;
		time:calendar = "noleap" ;
		time:bounds = "time_bounds" ;
	int mcdate(time) ;
	int mcsec(time) ;
	int mdcur(time) ;
	int mscur(time) ;
	int nstep(time) ;
		nstep:long_name = "time step" ;

```

As such our PEcAn nc metadata does not match the FATES output.  We either need to 1) replace the time since part and instead set it the same as FATES such that each successive year has "days since" of the start year OR 2) stop using ```        nc.time <- ncin$dim$time$vals             # days since "start_date"``` or modify such that each year correctly has days since that year 

**To Reproduce**
Run PEcAn-FATES

**Expected behavior**
We should have consistency between the time variable definition and the time data 

**Screenshots**
NA

**Machine (please complete the following information):**
 - Server [e.g. BU, NCSA, VM, etc.]
modex but would apply to all

 - OS: [e.g.Linux]
Linux but applies to all

 - Browser(if applicable) [e.g. chrome, safari]
NA

develop

serbinsh:
You can see here what I mean.  The current time var we use increases linearly rather than restarts each new simulation year.
```
[...]

PR #1953, opened 2018-05-31
Updated FATES output time variable to fix issue #1939. 

serbinsh:
Current fix is to keep the same time time_since from the first year of the run (CLM default).  However we could change this so it resets for each year if that is preferred

## Review Time Estimate

- [X ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

- [X ] I have updated the CHANGELOG.md.


serbinsh:
Or should we have time restart for each year of the year loop such that the time since is restarted each year?

serbinsh:
Does anyone mind if I merge this? @mdietze ?

PR #1957, opened 2018-06-01
Loading animations on page load and plot reclculations

Chaitanyachavali:
## Description
User's should know when to page is ready to use and while plots are being recalculated. We can intimate users by showing some animations while these background processes are going on

## Animations included
* Loading animation till page is ready to use
* Loading animation while plots are being recalculated

### Example (Animation on plot recalculation):
vid


bcow:
This looks great. Because there is code that is repeated in multiple places in the app, I am thinking this would be a good opportunity to start modularizing the code. 

I am willing to pull this in as is, but @Chaitanyachavali could you look at https://shiny.rstudio.com/articles/modules.html and decide if this is a feasible short term task? If so, I would wait to pull in this PR until it is modularized.

mdietze:
@Chaitanyachavali to keep GSoC on pace I'm going to pull this now, but I do encourage you to consider the modularization that @bcow suggests as part of your next PR

PR #1963, opened 2018-06-05
Emulator paper updates

istfer:
## Description

Changes accompanying emulator paper revisions. Main changes are:
1) Sampling from the MCMC in between rounds and in post-PDA ensembles (the latter is in progress)
2) Using log-normal prior for the bias parameter instead of uniform

## Review Time Estimate

- [ ] Immediately
- [X] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/utils/R/get.parameter.samples.R
```diff
@@ -74,21 +101,22 @@ get.parameter.samples <- function(settings,
     priors <- rownames(prior.distns)
     
     if (exists("trait.mcmc")) {
-      ma.traits <- names(trait.mcmc)
+      ma.traits[[i]] <- names(trait.mcmc)
```
mdietze: improved naming here too?

modules/assim.batch/R/pda.emulator.R
```diff
@@ -178,72 +177,29 @@ pda.emulator <- function(settings, external.data = NULL, external.priors = NULL,
...
   if (run.round) {
     
-    # loads the posteriors of the the previous emulator run
-    temp.round <- pda.load.priors(settings, con, run.round)
-    prior.round.list <- temp.round$prior
-    
-    
-    prior.round.fn <- lapply(prior.round.list, pda.define.prior.fn)
-    
     ## Propose a percentage (if not specified 80%) of the new parameter knots from the posterior of the previous run
```
mdietze: comment is not consistent change in default parameter

modules/assim.batch/R/pda.utils.R
```diff
@@ -754,7 +732,7 @@ return_hyperpars <- function(assim.settings, inputs){
     for(k in seq_along(assim.settings$inputs)){
       hyper.pars[[k]] <- list()
       hyper.pars[[k]]$parama <- 0.001
-      hyper.pars[[k]]$paramb <- 0.001 * mean(inputs[[k]]$data[,1], na.rm = TRUE) ^ 2
+      hyper.pars[[k]]$paramb <- 0.001 * sd(inputs[[k]]$data[,1], na.rm = TRUE) ^ 2
```
mdietze: instead of squaring the sd, why not just use `var`

base/utils/R/ensemble.R
```diff
@@ -130,15 +136,35 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
     col.i <- 0
     for (pft.i in seq(pft.samples)) {
       ensemble.samples[[pft.i]] <- matrix(nrow = ensemble.size, ncol = length(pft.samples[[pft.i]]))
+      
+      if(length(pft.samples[[pft.i]])>0){
+        # coming from anything that fits parmeters jointly (PDA / global-MA), we want to preserve correlations
+        same.i <- stats::quantile(pft.samples[[pft.i]][[1]], random.samples[, 1]) # use 1st
+        # will use these same indices for samples that will go together
+        same.i <- which(round(pft.samples[[pft.i]][[1]], digits = 4) %in% round(same.i, digits=4))
+        same.i <- same.i[!duplicated(round(pft.samples[[pft.i]][[1]], digits = 4)[same.i])]
```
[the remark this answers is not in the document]
istfer: Yeah I wasn't sure about how to do this, I will change it to uniform case only

istfer:
@mdietze I think this is ready for re-review. The only thing that I didn't address was this:
>Really seems like there should be an else here for handling the univariate case. Actually, it might just be better to handle the switch between univariate and multivariate sampling at a high level in the code

not sure what you're referring to here. Currently it should be able to handle the univariate case (do you mean, 1 parameter constrained by the PDA?) When it's univariate and another method (than uniform) is requested, we don't have to fall back to uniform if it's a post-PDA ensemble, is that what you want to be enabled?

Also, one more thing @bcow needs to check. `load_data()` doesn't load FC in PDA unless I change it from `NEE,kg C m-2 s-1,FC` to `FC,kg C m-2 s-1,FC` in `bety_mstmip_lookup.csv`, would it break your benchmarking experiments immediately? Or can it stay until next time we tackle this issue?

mdietze:
@istfer by univariate I was referring to methods that sample parameters independently from their marginal distributions. By multivariate I meant methods that sample parameters from the joint posterior distribution. AFAIK for the other methods that have been implemented (LHC, sobol, etc) we don't currently have a way of sampling them from the joint posterior. We only have one method (random sampling) for taking draws from the joint distribution.  Because of that I was suggesting that there be a high-level switch between joint and marginal methods rather than handling that conditional within every single sampling method individually (which seems inefficient)

base/utils/R/ensemble.R
```diff
@@ -130,15 +134,32 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
     col.i <- 0
     for (pft.i in seq(pft.samples)) {
       ensemble.samples[[pft.i]] <- matrix(nrow = ensemble.size, ncol = length(pft.samples[[pft.i]]))
+      
+      # meaning we want to keep MCMC samples together
+      if(length(pft.samples[[pft.i]])>0 & !is.null(param.names)){ 
+        # TODO: for now we are sampling row numbers uniformly
+        # throw a warning if other methods were requested 
+        if(method != "uniform"){
+          PEcAn.logger::logger.warn("Other approaches are not implemented yet for sampling jointly. Defaulting to uniform sampling of the chain.")
```
mdietze: Probably want this to be a logger.severe -- if downstream code is expecting a different sampling method (e.g. sobol SA) then returning random could cause a but that's hard to decode.

bcow:
@istfer I'm fine with that change to the lookup  table. I can always change it locally if I need to use NEE.

PR #1965, opened 2018-06-07
Added progress messages to dataone_download_fn

LiamBurke24:
Added a few progress messages to the dataone_download function since this function can take a few minutes to interface with the member nodes via the dataoneR API. These messages will be especially valuable in the Shiny download app. 

## Description

Its pretty simple. I just added some messages that will print on the console announcing the completion of various parts of the download process. I intend to print these messages in the web application. 

## Motivation and Context

The dataone_download function can take up to a few minutes to interface with the dataONE member nodes. I wanted to add some messages so that the user has some feedback to affirm that the function is working. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/data.land/R/dataone_download.R
```diff
@@ -27,24 +27,28 @@ dataone_download = function(id, filepath = "/fs/data1/pecan.data/dbfiles/", CNod
   ### automatically retrieve mnId
   cn <- dataone::CNode(CNode) 
   locations <- dataone::resolve(cn, pid = id) 
+    print("Connecting to Member Node")
```
mdietze: Use the PEcAn logger package, rather than print, to post messages as this will go to the terminal or correct log file at the correct level of verbosity (e.g. these mostly look like logger.info, not warnings or errors)

LiamBurke24: Got it. Just pushed these changes.

Issue #1949, opened 2018-05-30
Timeline for package checks in CI (+ list of current check errors)


infotroph:
**Describe the bug**
Current Travis builds run unit tests (`devtools::test`) on each package in PEcAn, but do not run any package checks (`R CMD check` or near-equivalently `devtools::check`), which would run tests *and* check for many possible errors in package structure, code quality, and documentation completeness.

I propose to implement full package checks on a four-stage schedule:

* Immediately: 
    - begin running checks on all packages every Travis build
    - Exempt those that cause the build to fail, using a temporary whitelist in .travis.yml. Continue running devtools::test on these.
* By September 1: 
    - Correct all ERRORs in check output, removing packages from whitelist as we go
    - Once whitelist is empty, all packages will be checked every build and an ERROR from the check will break the build. 
    - WARNINGs in check output will not yet break the build, but should have open issues to track them.
* Starting 2019:
   - Begin treating WARNINGs in check output as errors. 
   - Packages that produce warnings are exempted using a temporary whitelist. Continue checking these and breaking the build on ERRORs.
* By PEcAn 2.0:
    - Travis treats WARNINGs in check output as build-breaking errors.

**To Reproduce**

Packages that currently produce at least one ERROR from package checks (As determined from [this experimental Travis build](https://travis-ci.org/infotroph/pecan/builds/385187751)):

* [x] base/db: Fails with two check errors: One syntax error in examples, one test failure that doesn't occur in devtools::test() on this Travis build or at all (devtools::test, devtools::check, or R CMD build && R CMD check) on my machine. To correct, fix example syntax then look for differences in dependency calculation between $(make document) and $(make install).
	
* [x] base/utils: Fails with "Packages suggested but not available: ‘PEcAn.data.atmosphere’ ‘PEcAn.data.land’ ‘PEcAn.emulator’ ‘PEcAn.priors’ ‘PEcAn.settings’ ‘PEcAn.DB’". To correct, need to finish fixing circular dependencies so these can be removed from Suggests in base/utils.
	
* [x] models/ed: Fails with check error "Namespace dependency not required: ‘magrittr’". To correct, add magrittr to Imports.
	
* [x] models/linkages: Fails with check error "Package suggested but not available: ‘linkages’". To correct: fix syntax in Remotes field
	
[...]

PR #1973, opened 2018-06-14
Run (most) package checks on Travis

infotroph:
## Description

As proposed in #1949: Runs devtools::check on all packages during the Travis build, and marks the build as broken if any check returns an ERROR. WARNINGs and NOTEs do not break the build (for now), but I encourage all PEcAn authors to routinely run `devtools::check` locally and fix all WARNINGs and as many NOTEs as possible.

N.B. Currently skips checking of PEcAn.benchmark and PEcAn.data.land, until failures in their examples are resolved.

## Motivation and Context

Note that this PR simply adds `make check` into the existing test script, and does *not* implement the "build stages" approach I demoed a few meetings ago -- that's in [this branch](https://github.com/infotroph/pecan/tree/travis-build-stages-grouped) and needs more work before merging.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

- [x] build system tweak

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ script:
   # TODO: Would be nice to add `-Otarget` after `-j"$NCPUS"` below,
   # but not supported in Make 3.x.
   # When Travis updates, check for Make 4 and add -O if available.
-  - make -j"$NCPUS"
+  - make # TODO: can't use -j here; rtm compilation not parallel-safe
   - echo 'Testing PEcAn packages'
   - make -j"$NCPUS" test
   - make -j"$NCPUS" document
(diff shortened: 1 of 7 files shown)


infotroph:
This is finally passing Travis and ready for another look. Caveats:

* `make document`, `make test`, and `make check` are all now run with 2 cores in parallel. Not a huge speedup (that's all the cores a Travis container has), but enough to prevent the job timing out.
* I had to skip checking of RTM too (#1976). Once this is fixed, `make install` can use `-j2` as well, which should speed things up even further.
* Processes running in parallel means interwoven log entries. This has the potential to be pretty annoying; if necessary we might consider upgrading the system Make to version >= 4.0 -- it supports `-Otarget` to keep output from each job together in the console.

PR #1979, opened 2018-06-18
Ldramko developdownloads: adding download.NOAA.R function

Luke-Dramko:
## Description

Finished the download.NOAA.R function, which saves NOAA GEFS weather data, processes it, and returns information related to its location, type, etc.

## Motivation and Context

This feature adds new functionality by adding a way to interact an additional data source.  This data can be used for new types of models, such as for carbon flux models.

This is new.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:

How and with what should I update the CHANGELOG.md document?   I am unsure of whether my change requires a change in documentation.  Finally, all the tests I ran passed, which tested all bounds in the input I was aware of.  However, I am unsure of if there are more tests out there to run.


modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+    var_dat = cbind(var_dat, rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = 1, forecast_time = "1800", date=gefs_date)$data)
+    j = 1
+    while (j < num_days) {
+      current_date = current_date + lubridate::days(1)  #increments the date by one
+      
+      gefs_date = POSIX.to.GEFSDate(current_date)
+      var_dat = cbind(var_dat, rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = 1, forecast_time = "0000", date=gefs_date)$data)
+      var_dat = cbind(var_dat, rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = 1, forecast_time = "0600", date=gefs_date)$data)
+      var_dat = cbind(var_dat, rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = 1, forecast_time = "1200", date=gefs_date)$data)
+      var_dat = cbind(var_dat, rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = 1, forecast_time = "1800", date=gefs_date)$data)
```
mdietze: I'm really not following what you're trying to do here, but I'm pretty sure it's not what we're looking for. First, because time_idx is set to 1 it seems like you're not grabbing the full forecast, just the first 6hr, which isn't what we want. We want the full timeseries, start date to the end date, which requires setting date and time_idx. Second, setting `forecast_time` requires that we need to know the `start_time`, not just the `start_date`, and find the nearest start time to the one requested. Third, the NOAA server will be much more efficient if you grab large chunks of data at once rather than grabbing a lot of little data and cbinding it together on our side.

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+      j = j + 1   
+    }
+    noaa_data[[i]] = var_dat
+    i = i+1
+  }
+  
+  last_data_set_at = "1800"
+  
+  #Data requests that end on the current day have to be handled in their own special loop because not all data for a given day is necessarily
```
mdietze: I think this whole `if` can be dropped if time_idx is set correctly.

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+      }
+    }
+  }
+  
+  if (verbose && include_forecast) {
+    PEcAn.logger::logger.info("Downloading forecast data...")
+  }
+  
+  #This loop attaches 16 day forecast data onto the data that currently exists.
+  if (include_forecast) {
```
mdietze: still don't see the need for this separate case. Indeed, this case look the most like what we want -- I think you made everything above more complicated than it needed to be. Also, time_idx should max at 65 not 64

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+  }
+  
+  ###################################################
+  # Not all NOAA data units match the cf data standard.  In this next section, data are processed to
+  # confirm with the standard when necessary.
+  # The following is a list of variables which need to be processed:
+  # 1. NOAA's relative humidity must be converted to specific humidity
+  # 2. NOAA's measure of precipitation is the accumulation over 6 hours; cf's standard is precipitation per second
+  
+  if (verbose) {
```
mdietze: as above, drop if verbose

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+    PEcAn.logger::logger.info("Converting units.")
+  }
+  
+  #Convert NOAA's relative humidity to specific humidity
+  humid_index = cf_var_names == "specific_humidity"
```
mdietze: `humid_index` will be a vector of TRUE/FALSE. I think you might want to make this be 
```
humid_index = which(cf_var_names == "specific_humidity")
```

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+  
+  humid_data = noaa_data[humid_index][[1]]
+  temperature_data = noaa_data[cf_var_names == "air_temperature"][[1]]
+  pressure_data = noaa_data[cf_var_names == "air_pressure"][[1]]
+   
+  #Depending on the volume and dimensions of data you download, sometimes R stores it as a vector and sometimes
+  #as a matrix; the different cases must be processed with different loops.
```
mdietze: I'm not sure why this would be the case -- aren't we always grabbing all of the ensemble members by default.

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+  # 2. NOAA's measure of precipitation is the accumulation over 6 hours; cf's standard is precipitation per second
+  
+  if (verbose) {
+    PEcAn.logger::logger.info("Converting units.")
+  }
+  
+  #Convert NOAA's relative humidity to specific humidity
+  humid_index = cf_var_names == "specific_humidity"
+  
+  humid_data = noaa_data[humid_index][[1]]
```
mdietze: not sure why these are being accessed as `noaa_data[humid_index][[1]]` rather than just `noaa_data[[humid_index]]` etc. same with the re-insert

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+  }
+  
+  #Update the noaa_data list with the correct data
+  noaa_data[humid_index][[1]] <- humid_data
+  
+  # Convert NOAA's total precipitation (kg m-2) to precipitation flux (kg m-2 s-1)
+  #NOAA precipitation data is an accumulation over 6 hours.
+  precip_index = cf_var_names == "precipitation_flux"
+  
+  noaa_data[precip_index][[1]] = noaa_data[precip_index][[1]] / 21600  #There are 21600 seconds in 6 hours
```
mdietze: don't hard-code conversions, use `udunits2`

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+  if (verbose) {
+    PEcAn.logger::logger.info("Writing data to netcdf file")
+  }
+  
+  
+  if (!dir.exists(outfolder)) {
+    dir.create(outfolder, recursive=TRUE)
+  }
+  
+  setwd(outfolder)
```
mdietze: Don't do this within a function! Just use full path names below

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+  st = lubridate::day(start_date)
+  ed = lubridate::day(end_date)
+  dycomponent = st
+  if (st != ed) { dycomponent = paste0(st, "-", ed) }
+  
+  flname <- if (include_forecast) {
+    paste("NOAA", "GEFS", yrcomponent, moncomponent, dycomponent, "f", "nc", sep=".")
+  } else {
+    paste("NOAA", "GEFS", yrcomponent, moncomponent, dycomponent, "nc", sep=".")
+  }
```
mdietze: Per earlier comment about ISO dates, this whole bit of code can be simplifed to one case, and you don't need to hand-parse year, month, and day as `format` can do that for you. Also, seems like we probably need to include time as well not just date.

That said, right now it looks like you're setting up to write one file, rather than one file per ensemble member -- I think we need to include the ensemble member in the file name

modules/data.atmosphere/R/download.NOAA.R
```diff
@@ -0,0 +1,387 @@
...
+    formatname = "CF Meteorology",        #Type of data
+    startdate = paste0(as.character(start_date), " 00:00:00"),                                     #starting date and time, down to the second
+    enddate = paste0(as.character(end_date), " ", substring(last_data_set_at, 1, 2), ":00:00"),    #ending date and time, down to the second
+    dbfile.name = "NOAA.GEFS",            #Source of data
+    stringsAsFactors = FALSE
+  )
+  
+  #Begin working with the netcdf4 saving format
+  #define dimensions.  Note that all variables will have the same dimension, so we can reuse them for all variables.  The data is two dimensional: one dimension
+  #is time (measured in periods of six hours), while the other one is which particular ensemble the data is drawn from.
```
mdietze: 1) PEcAn standard requires lat and lon dimensions, even if they are length 1.
2) I believe our current standard is to write different ensemble members to different files. @ankurdesai @crollinson can you confirm and point @Luke-Dramko to an example of where/how you're writing out met ensembles in the downscaling code, as well as how you're storing ensembles in the database (i.e. is the input & dbfiles ID by site or by ensemble member?)

mdietze:
Also, the package documentation needs to be rebuilt because you've added new functions, and those files committed as well.

Luke-Dramko:
All issues discussed above have been fixed and pushed to github.

modules/data.atmosphere/R/download.NOAA.R
crollinson: Currently, I use a hierarchical naming scheme where it's ID.ENS (adding on another .ENS if you're doing both bias correction & temporal downscaling) and each member gets its own folder.  

In my workflow, I specify the [ensemble member tags up front](https://github.com/crollinson/MetDownscaling_Manuscript/blob/505a6a602aeb686618e78943c420e585192227c1/Workflow_Downscale/3_bias_correct.R#L57-L59) (allowing me to easily add on), which then get passed to the downscaling argument ([bias correction example](https://github.com/crollinson/MetDownscaling_Manuscript/blob/505a6a602aeb686618e78943c420e585192227c1/Workflow_Downscale/3_bias_correct.R#L108))

I don't know how this is getting implemented in PEcAn's met workflow.  I did try to make sure it would be compatible with what I know of PEcAn's arrangements, but also make it flexible enough for my standalone use.  Hope this helps!

Luke-Dramko:
Latest version of NOAA_GEFS download function.  Updated to include new ensemble handling standards from issue #1167, including updating file names and returning a list of data frames.  Also included code to compensate for the delay in data posting to the NOAA database.

tonygardella:
@Luke-Dramko From the pecan directory, on the command line, do `make document` and push the documentation files that it creates.

mdietze:
Also, before you run `make document` you should take a closer look at the available ROxygen tags, as there's a lot of info that's in a large text blob and much of that maps onto specific tags.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -43,6 +43,39 @@
...
+##' args = commandArgs(trailingOnly = TRUE)
+##' if (length(args) > 0) {
+##'  test_no = as.integer(args[1])
+##'}
+##'
+##' source("~/pecan/modules/data.atmosphere/R/download.NOAA_GEFS.R")
+##'
+##'
+##' if (test_no == 1) { #Default case
+##'  download.NOAA_GEFS("~/Working/results", lat.in= 46.2420, lon.in = -89.3476, "willow creek")
```
mdietze: 1) There needs to be an `@example` tag before the example code
2) I I think this test case is too complicated -- you really just need this one line example within the `\dontrun` block

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,289 @@
...
+  
+  nc_var_list = list()
+  for (i in 1:length(cf_var_names)) { #Each ensemble member will have data on each variable stored in their respective file.
+    nc_var_list[[i]] = ncdf4::ncvar_def(cf_var_names[i], cf_var_units[i], dimensions_list, missval=NaN)
+  }
+  
+  #For each ensemble
+  for (i in 1:21) { # i is the ensemble number
+    #Generating the file name.  File name is specific to the range of dates and times of the forecast.
+    flname = file.path(outfolder, paste("NOAA_GEFS", site_id, i, format(start_date, "%Y-%m-%dT%H:%M"),
```
mdietze: Each ensemble member needs to be in it's own folder, so you need to make an ensemble-specific folder between the general outfolder and the filename

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,289 @@
...
+##' that, but if the end date is beyond 16 days, only 16 days worth of forecast are recorded.  Times are rounded down to the previous 6 hour forecast.  NOAA
+##' GEFS weather data isn't always posted immediately, and to compensate, this function adjusts requests made in the last two hours
+##' back two hours (approximately the amount of time it takes to post the data) to make sure the most current forecast is used.
+##' 
+##' @section Data Save Format
+##' Data is saved in the netcdf format to the specified directory.  File names reflect the precision of the data to the given range of days.
+##' NOAA.GEFS.willow creek.3.2018-06-08T06:00.to.2018-06-24T06:00.nc specifies the forecast, using ensemble nubmer 3 at willow creek on
+##' June 6th, 2018 at 6:00 a.m. to June 24th, 2018 at 6:00 a.m.
+##' 
+##' @section Dependencies
```
mdietze: As mentioned previously, this Dependencies section can be dropped, that info is all in the DESCRIPTION

modules/data.atmosphere/R/download.NOAA_GEFS.R
mdietze: It's actually `@examples` not example

modules/data.atmosphere/R/download.NOAA_GEFS.R
Luke-Dramko: In that case, what should the folder name be?

mdietze: `NOAA_GEFS.[site_id].[ens_id].[start_datetime].[end_datetime]`

Luke-Dramko:
Function has been updated to include separate folders for each ensemble.  The `@examples` tag was added.  Additionally, roxygen documentation has been updated for the package.

modules/data.atmosphere/inst/scripts/runtest-download.NOAA_GEFS.R
```diff
@@ -0,0 +1,38 @@
...
+#It accepts a command line argument specifying the case test number to run, otherwise, it
+#defaults to the first case.
+
+test_no = 1
+args = commandArgs(trailingOnly = TRUE)
+if (length(args) > 0) {
+  test_no = as.integer(args[1])
+}
+
+source("~/pecan/modules/data.atmosphere/R/download.NOAA_GEFS.R")
```
ashiklom: For the future, instead of this, look at the [`devtools::load_all`](https://www.rdocumentation.org/packages/devtools/versions/1.13.5/topics/load_all) function.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,260 @@
...
+  increments = increments + ((lubridate::hour(end_date) - lubridate::hour(start_date)) %/% 6) #These calculations are required to use the rnoaa package.
+  
+  end_hour = sprintf("%04d", ((forecast_hour + (increments * 6)) %% 24) * 100)  #Calculating the starting hour as a string, which is required type to access the 
+                                                                        #data via the rnoaa package
+  forecast_hour = sprintf("%04d", forecast_hour * 100)  #Having the end date as a string is useful later, too.
+  
+  #Recreate the adjusted start and end dates.
+  start_date = as.POSIXct(paste0(lubridate::year(start_date), "-", lubridate::month(start_date), "-", lubridate::day(start_date), " ", 
+                                 substring(forecast_hour, 1,2), ":00:00"), tz="UTC")
+  end_date = start_date + lubridate::hours(increments * 6)
```
ashiklom: I won't let it block the PR, but in general, you should try to make sure your code follows the [R style guide](http://style.tidyverse.org/).

For example, you want to use `<-` for assignment (but _not_ for function arguments). There are a few other minor things in there that I would also fix in the future (e.g. spaces between infix operators and argument assignments, indentation and line length guidelines).

modules/data.atmosphere/inst/scripts/runtest-download.NOAA_GEFS.R
```diff
@@ -0,0 +1,38 @@
+#This program is to test the R script donload.NOAA.R during development.
```
ashiklom: Again, this shouldn't hold up the PR, but it would be nice to convert this to some kind of [unit test](http://r-pkgs.had.co.nz/tests.html).

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,260 @@
...
+  
+  dimensions_list = list(time_dim, lat_dim, lon_dim)
+  
+  nc_var_list = list()
+  for (i in 1:length(cf_var_names)) { #Each ensemble member will have data on each variable stored in their respective file.
+    nc_var_list[[i]] = ncdf4::ncvar_def(cf_var_names[i], cf_var_units[i], dimensions_list, missval=NaN)
+  }
+  
+  #For each ensemble
+  for (i in 1:21) { # i is the ensemble number
```
ashiklom: This looks like a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)). I would assign this to an informatively-named variable early in the function (e.g. `N_weather_ensembles`). That will also make it easier to change later on.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,260 @@
...
+    
+    flname = file.path(ensemble_folder, paste(identifier, "nc", sep = "."))
+    
+    #Each ensemble member gets its own unique data frame, which is stored in results_list
+    #Object references in R work differently than in other languages. When adding an item to a list, R creates a copy of it
+    #for you instead of just inserting the object reference, so this works.
+    results$file = flname
+    results_list[[i]] = results
+    
+    if (!file.exists(flname) | overwrite) {
```
ashiklom: Inside of `if` statements, you almost always want `||` instead of `|`. The difference is that `||` (and its analogue `&&`) always return a single logical value and operate strictly from left to right, whereas `|` and `&` are vectorized across the entire expression. See [`?base::Logic`](http://stat.ethz.ch/R-manual/R-devel/library/base/html/Logic.html) for details.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,260 @@
...
+  #The statement udunits2::ud.convert(1, "kg m-2 6 hr-1", "kg m-2 s-1") is equivalent to udunits2::ud.convert(1, "kg m-2 hr-1", "kg m-2 s-1") * 6,
+  #which is a little unintuitive. What will do the conversion we want is what's below:
+  noaa_data[[precip_index]] = udunits2::ud.convert(noaa_data[[precip_index]], "kg m-2 hr-1", "kg m-2 6 s-1")  #There are 21600 seconds in 6 hours
+  
+  #############################################
+  # Done with data processing.  Now writing the data to the specified directory. Each ensemble member is written to its own file, for a total
+  # of 21 files.  
+  if (!dir.exists(outfolder)) {
+    dir.create(outfolder, recursive=TRUE)
+  }
```
ashiklom: FWIW, `dir.create(outfolder, recursive = TRUE, showWarnings = FALSE)` will create the directory if it doesn't exist and succeed silently if it does (without `showWarnings = FALSE`, it will throw a warning but still succeed).

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,260 @@
...
+  
+  results_list = list()
+  
+  #Each ensemble gets its own file.
+  #These dimensions will be used for all 21 ncdf4 file members, so they're all declared once here.
+  #The data is really one-dimensional for each file (though we include lattitude and longitude dimensions
+  #to comply with the PEcAn standard).
+  time_dim = ncdf4::ncdim_def("Time", "6 Hours", 1:ncol(noaa_data[[1]]))
+  lat_dim = ncdf4::ncdim_def("Lattitude", "Degrees North", lat.in)
+  lon_dim = ncdf4::ncdim_def("Lattitude", "Degrees East", lon.in)
```
ashiklom: Shouldn't this be `"Longitude"`? Also, "latitude" is misspelled (unless it's like that in the PEcAn standard?).

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -0,0 +1,260 @@
...
+    PEcAn.logger::logger.severe(sprintf('Start date (%s) exceeds the NOAA GEFS range (%s to %s).',
+                                        start_date,
+                                        NOAA_GEFS_Start_Date, Sys.Date()))
+  }
+  
+  if (lubridate::hour(start_date) > 23) {
+    PEcAn.logger::logger.severe(sprintf("Start time %s is not a valid time", lubridate::hour(start_date)))
+  }
+  
+  if (lubridate::hour(end_date) > 23) {  #Done separately from the previous if statement in order to have more specific error messages.
```
ashiklom: Shouldn't both of these be `>=`? Why can't I get data starting at 23:30 (11:30PM)?

Luke-Dramko: You can.  This statement only looks at the hour, not the minute.  So, as long as the hour is 23, you're good to go.  At 23:30, the hour is 23, so the error statement won't be triggered.

modules/data.atmosphere/R/download.NOAA_GEFS.R
Luke-Dramko: Good catch.  I thing that was a copy and paste thing.  Fixed.

PR #1980, opened 2018-06-18
Progress Spinner (for D1 download) and Editing Directory Name for Local Upload

LiamBurke24:
I added two main functionalities to this app. 1. Spinner that displays when the "shiny-busy" class is active in the DataONE download app. 2. A text input field in the local upload panel that allows the user to enter a custom name for the directory in `dbfiles` where the files will be stored. 

## Description

1. I added a spinner from  [http://www.lettersmarket.com/uploads/lettersmarket/blog/loaders/common_green/ajax_loader_green_64.gif](url) to show when app is processing the download dataONE request. Thanks to @para2x  I employ a `conditionalPanel` to display the loader and a loading message only when the "shiny-busy" class is active in the DataONE download panel. Because of this, users can be certain that the process is running when the loader is active.

2. I added a input field in the local files panel so that users can name the destination directory where their local uploads will be moved to. The app takes this input and creates a destination directory within `dbfiles` and displays the file path after the files have been moved. 

## Motivation and Context

Without the spinner, users cannot be certain that their download request is being processed. Since the spinner only displays when the `shiny-busy` class is active, the user can be certain that the app is processing the download while the spinner is displayed. 

The input directory name box solves the issue of creating a name for the destination directory inside of dbfiles. Perhaps we may want to add the user's username to the directory name to see who uploaded what data. We may want to create some sort of standardized naming scheme for these directories instead. In the meantime, this allows us to test uploading files to multiple directories within dbfiles. 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


LiamBurke24:
Currently trying to figure out how to install a couple dependencies for the `redland` binary (via `.travis.yml` that are preventing redland, datapack, and by extension dataone from being installed in the build.

ashiklom:
Tried to fix this build for a while but gave up. The ultimate roadblock I ran into is a fundamental incompatibility between the following pairs of packages:
- `libgeos-c1v5` vs. `libgeos-c1`
- `libgdal1h` vs. `libgdal20`

You can review my commits to experience the fun I had trying to get this to work.

In general, it _seems_ like it should be possible to make everything work if we pulled everything from the `ubuntugis-unstable` PPA, but that seems pretty complicated because of all the dependencies among spatial libraries (`gdal` and `postgis`).

I think a lot of this would be resolved almost instantly if only Travis supported Xenial (16.04), but seems they've been dragging their feet on this for _years_.

modules/data.atmosphere/NAMESPACE
```diff
@@ -1,5 +1,8 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(extract,local.CMIP5)
+S3method(extract,local.NLDAS)
+S3method(extract,nc)
```
infotroph: This change shouldn't happen -- it's caused by Roxygen getting confused about things you didn't touch. To fix, delete your local copy of this NAMESPACE file and then regenerate it by running Roxygen from a clean session -- probably easiest from a terminal:

```
cd /path/to/your/copy/of/pecan
rm modules/data.atmosphere/NAMESPACE
make .doc/modules/data.atmosphere
```

modules/data.land/DESCRIPTION
```diff
@@ -14,7 +14,10 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
 Depends:
     PEcAn.DB,
     PEcAn.utils,
-    dbplyr
+    dbplyr, 
+    redland,
+    datapack,
+    dataone,
```
infotroph: Consider moving dataone from the `Depends` second to the `Imports` section just below, then include the namespace in the spelling each time you use a `dataone` function inside `data.land`. So for example if you're using the (hypothetical) dataone function `some_func`, you'd write `dataone::some_func(...)`.

LiamBurke24: Sounds good. I already use that format in my dataone related functions. To confirm, I can move dataone to `Imports`? Should I do anything to its dependencies (`redland` and `datapack`)?

LiamBurke24:
@mdietze I have pushed the new UI commits to this PR (so this is essentially two PR's in one). I am still unable to get it to build. I am hoping to sort this out soon. Here is the current bottleneck. I can't get past data.atmosphere: ``` Error: object 'extract' not found whilst loading namespace 'PEcAn.data.atmosphere'
Execution halted
Makefile:137: recipe for target '.doc/modules/data.atmosphere' failed
make: *** [.doc/modules/data.atmosphere] Error 1``` 

I tried deleting and remaking the NAMESPACE file to no avail. I will be able to look at this more closely on Monday morning. I am not sure if this PR is of any value until this gets sorted out. Are you able to download and run my code without it breaking your version of pecan? Perhaps you can look this over on a sacrificial branch? Thanks!

LiamBurke24:
Not sure what is causing this download.NARR_site conflict. @ashiklom Do you think this is a byproduct of our troubleshooting last week? I tried to resolve the merge conflict and push the commit but the conflict persists.

modules/data.land/DESCRIPTION
infotroph: Yes, you can move dataone to Imports. If redland and datapack are never called directly from any data.land functions, they should be able to go unmentioned in this file and dataone will handle its own dependency on them for itself.

CHANGELOG.md
```diff
@@ -35,8 +35,11 @@ For more information about this file see also [Keep a Changelog](http://keepacha
   
 - shiny/Data-Ingest 
   - Download data from DataONE to a temporary directory, display the contents of that directory, and then move files from the temporary directory to dbfiles
-  - Upload files from local machines (via drag and drop on some browsers), display files, and then move these files to a directory in dbfiles. 
-
+  - Upload files from local machines (via drag and drop on some browsers), display files, and then move these files to a directory in dbfiles.
+  - Spinner displays when class "shiny-busy" is invoked during the dataONE download process. In this way, users can be sure that the app is live and is processing their request. 
+  - Users can now input the name of the destination directory that they wish to create within dbfiles. 
+  - Updated Travis.yml to include librdf0-dev so that it can download redland, datapack, and dataone.
```
infotroph: For the record it's usually OK to leave changes to .travis.yml unmentioned in the changelog -- think of them as a necessary part of the user-facing changes described in the lines above, not a separate feature in themselves. No need to change this one, though -- the extra detail doesn't hurt.

LiamBurke24: Got it. Thanks!!

PR #1985, opened 2018-06-20
Bug Fix, Debias routine: Bound rainless day distribution

crollinson:
## Description

Bound the number of rainless days to at least 1 and no more than 364.
At sites with a high standard deviation, you can get random values for
n.now that cause the quantile to be negative or greater than 1 (which
is impossible.)

## Motivation and Context

In sites or datasets with high inter annual variability in the number of rainless days and when processing thousands of years (via ensemble generation), an impossible number of rainless days can be generated.  This causes the quantile function too ind the precip cutoff to fail because it's trying to find a number outside of the 0-1 bound.  

I've added a bound where the minimum number of days is as much as 1 sd less than the minimum observed in the training dataset, allowing for the dataset to drift into novel conditions.  If that produces a negative number, then try 0.5 sd less and if that fails, set the minimum number of rainless days to 1.  A similar logic is used for an upper bound.  These cases should be invoked rarely, but they do happen.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/debias_met_regression.R
+++ b/modules/data.atmosphere/R/debias_met_regression.R
@@ -185,6 +185,8 @@ debias.met.regression <- function(train.data, source.data, n.ens, vars.debias=NU
       
       # Hard-coding in some sort of max for precipitaiton
       rain.max <- max(train.data$precipitation_flux) + sd(train.data$precipitation_flux)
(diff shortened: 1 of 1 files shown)


crollinson:
I debated that, but was afraid it would introduce some unforeseen error and I wasn't sure if there were places where it rains every day in a year or never rains in a year.  I *think* it should be with 0 & 365 though, so there's a commit coming with that change.

PR #1986, opened 2018-06-21
Missing roxygen2

serbinsh:
Added roxygen2 to depends list

### DIFF
--- a/scripts/install.dependencies.R
+++ b/scripts/install.dependencies.R
@@ -21,7 +21,7 @@ if (!("Rgraphviz" %in% installed.packages()[, "Package"])) {
 }
 
 # install packages needed from CRAN
-list.of.packages <- c("BayesianTools","data.table", "doSNOW", "dplR", "earth", 
+list.of.packages <- c("roxygen2", "BayesianTools","data.table", "doSNOW", "dplR", "earth", 
                       "emulator", "ggmap", "ggplot2", "gridExtra", "Hmisc", "httr", "kernlab", 
                       "GPfit", "knitr", "Maeswrap", "MASS", "MCMCpack", "mvtnorm",
                       "plotrix", "raster", "randtoolbox", "rjags", "rgdal", "tgp", "DBI", 


serbinsh:
Are we still using that file to try and get many of the depends installed?  The error for my PEcAn make came when building atmosphere since roxygen2 was a depends....or maybe all

mdietze:
I didn't realize the install.dependencies script was still used/maintained. It's not called anywhere from the Makefile

serbinsh:
Indeed @mdietze I just noted that and as see

```[sserbin@modex pecan]$ grep -ri 'roxygen2' .
Binary file ./.git/index matches
./.travis.yml:      - r-cran-roxygen2
./Makefile:.install/roxygen2: | .install
./Makefile:	time Rscript -e "if(!require('roxygen2')) install.packages('roxygen2', repos = 'http://cran.rstudio.com', Ncpus = ${NCPUS})"
./Makefile:$(ALL_PKGS_I) $(ALL_PKGS_C) $(ALL_PKGS_T) $(ALL_PKGS_D): | .install/devtools .install/roxygen2 .install/testthat```

but for some reason when building PEcAn in a R lib path that didnt have roxygen2 the build for atmosphere crashed....could be that a depends for roxygen2 was also missing causing that not to work so we can probably kill this PR....

**Comment by serbinsh** [comment]
Yeah, I think it was missing roxygen2 depends that didnt automagically update which ultimately killed the build. Closing this unless we think we need it.

**Comment by serbinsh** [comment]
ugh...stringi, then Rcpp, then xml2, ....what next! this R 3.5 thing is a bit annoying

**Comment by serbinsh** [comment]
More

```
Rscript -e "devtools::document('"modules/data.atmosphere"')"
Updating PEcAn.data.atmosphere documentation
Loading PEcAn.data.atmosphere
Error in (function (dep_name, dep_ver = NA, dep_compare = NA)  :
  Dependency package car not available.
Calls: <Anonymous> -> load_all -> load_imports -> mapply -> <Anonymous>
Execution halted
make: *** [.doc/modules/data.atmosphere] Error 1

```

really not sure why my update.packages(checkBuild= true) didnt catch all of this?

PR #1991, opened 2018-06-25
Update PEcAnRTM

serbinsh:
Updated PEcAnRTM package to include a new function for running the 4SAIL model using measured or modeled input reflectance, transmittance, and soil reflectance/ground cover reflectance spectra

## Description
Created a new R function "foursail()" to interface with the 4SAIL Fortran code 

## Motivation and Context
To enable the use of 4SAIL with any version of PROSPECT (i.e. 4, 5, 5b, D) and custom soil/background reflectance inputs

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X ] When possible
## Types of changes

## Checklist:

- [ X] My change requires a change to the documentation.
- [ X] I have updated the CHANGELOG.md.
- [ X] I have added tests to cover my changes.


modules/rtm/inst/scripts/testing_foursail.R
```diff
@@ -0,0 +1,35 @@
...
+
+# get leaf refl/trans
+LRT <- PEcAnRTM::prospect(c(2,55,10,3,0.1,0.007,0.007), 'D')
+plot(LRT[,1])
+refl <- LRT[,1]
+length(refl)
+tran <- LRT[,2]
+
+# generate 4SAIL canopy spectra
+sail_spec <- foursail(refl, tran, rsoil, params)
```
ashiklom: It would be nice to convert this to a unit test. You could basically keep all of the code, with the following changes:
* Drop the `library` call.
* Add a `context("SAIL canopy RTM with custom inputs")
* Add one or two very simple `expect_*` statements, such as `expect_equal(nrow(sail_spec), 2101); expect_equal(ncol(sail_spec), 4); expect_true(all(sail_spec) > 0); expect_true(all(is.finite(sail_spec)))`

modules/rtm/NAMESPACE
```diff
@@ -39,6 +39,7 @@ export(params.prospect4)
 export(params.prospect5)
 export(params.prospect5b)
 export(params2edr)
+export(plot.spectra)
```
ashiklom: Suggested this on Slack, but re-iterating here:

Not sure why this S3 method is being recognized all of a sudden. I would try one or more of the following in the `plot.spectra` definition:
* `#' @importFrom graphics plot` -- maybe the issue is that it's not finding the `plot` S3 generic in the PEcAnRTM NAMESPACE, and that's why it's not recognizing it as an S3 method. If you do this, you may also have to add `graphics` to the `Imports:` in the DESCRIPTION.
* `#' @method plot spectra` -- explicitly declare it as a method. I would only do this if the previous suggestion fails -- in general, `roxygen` should be smart enough to figure this out.

modules/rtm/inst/scripts/testing_foursail.R
ashiklom: I can take care of this. I just pulled the PR and can add a patch.

serbinsh: OK thats cool @ashiklom thanks!

serbinsh:
@ashiklom updated with test.  Let me know what else could use tweaking, if anything

modules/rtm/NAMESPACE
serbinsh: Done

modules/rtm/inst/scripts/testing_foursail.R
serbinsh: Updated with new test [that you prepared for me :) ]

PR #1992, opened 2018-06-26
Travis: use Docker postgis

infotroph:
## Description

* Provide GDAL 2 by getting all GDAL binaries from ubuntugis PPA
* Run PostGIS from a Docker image and turn off local Postgres entirely
* Add `\dontrun{}` to all examples in `modules/data.atmosphere/R/download*.R` that didn't yet have them

## Motivation and Context

The underlying issue here is a GDAL version conflict:
* We use `rgdal`, which now requires GDAL 2.
* The Travis-blessed way of installing Postgis (`postgresl-9.6-postgis-2.3`, retrived from apt.postgres.org) involves a binary linked against GDAL 1. this appears to be true for *all* the postgres.org Trusty binaries (i.e. we could easily install a newer Postgres but it wouldn't help), and I haven't found anyone else distributing a Trusty binary (i.e. we can't just grab a different binary until Travis offers a Linux less than five years old).

I tried a number of avenues to resolve this, but the changes here are the simplest option I see that actually resolves the problem.

**Running Postgres from a Docker image is a big change to the Travis build**: It both adds a major complexity layer and means we have to use "sudo: required", which is Travis's obtuse way of spelling "Run in a full clean VM instead of from a container, with accordingly slower boot times." Please review suspiciously and with an eye for unstated implications.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


mdietze:
Looks good to me, but I'd like @robkooper and @ashiklom to think about this since they're much closer

.travis.yml
```diff
@@ -98,7 +99,7 @@ notifications:
 
 ## list of services to be running
 services:
-  - postgresql
+  - docker
```
robkooper: do we need docker even if we use `sudo: required`? Would it be enough to install libgdal-dev?

.travis.yml
```diff
@@ -21,20 +24,18 @@ cache:
   - packages
 
 addons:
-  postgresql: 9.6
   apt:
     sources:
       - sourceline: 'ppa:marutter/c2d4u'
       - sourceline: 'ppa:nschloe/hdf5-backports'
+      - sourceline: 'ppa:ubuntugis/ppa' # for GDAL 2 binaries
```
robkooper: Why not also install libgdal-dev, which is what provides the right libraries to compile rgdal.

infotroph: We're installing it already (line 49) -- it was just finding[libgdal-dev 1.10 from trusty/universe](https://travis-ci.org/PecanProject/pecan/builds/396665460#L1083) before, and after adding this line it now finds [libgdal-dev 2.1 from ubuntugis](https://travis-ci.org/PecanProject/pecan/builds/396861337#L1110) without compiling.

.travis.yml
infotroph: I'm not sure I follow -- we install libgdal-dev already (see my comment below), the problem is that there's no version of it that's acceptable to both postgis and rgdal. Are you suggesting that we install both gdal 1 and gdal 2 alongside each other?

.travis.yml
robkooper: I think you might be able to install `libgdal1-dev`

infotroph: At some point I tried that with apt binaries and found that it always removed the first-installed version when I tried to the other, but I don't think I revisited it with sudo enabled. Would you be willing to take a stab at it?

infotroph:
Since we'll be asking @robkooper to fix things if this breaks horribly, please don't merge until he approves (or nixes!).

PR #2002, opened 2018-07-11
ED tree-ring SDA with better variables

istfer:
## Description

- Biomass changes are now updated in terms of BDEAD (instead of density, NPLANT)
- DBH and BSTORAGE are also updated proportionally to BDEAD
- FAST_SOIL_C and and STRUCTURAL_SOIL_C are updated in the state matrix

Initial condition sampling, and updateing biomass and/or density optionally are not in yet.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


mdietze:
Would there be a reason to have a flag that would let you toggle the updating between updating nplant (previous), bdead (current), and both (planned) so that you could run experiments seeing which performed best?

PR #2007, opened 2018-07-23
Data-Ingest New Features and attempt at linking to BETY (Work in Progress) 

LiamBurke24:
Added new features to app while waiting for BETY docker issues earlier in the week. Users can now create new input records, db file records, and format records by entering all the inputs in the inputs tab and then selecting `Bety Test`. This should allow us to quickly progress to a final version of this app. It is still necessary for users to manually check that the input was inserted in the various tables. 

## Description

New Features: 
- Hide `Download` button until `doi` is entered. 
- New Radio Buttons ot toggle between download methods 
- Pass Filename from selected file from input method to `Name` field
- Pass new format name and mimetype from the formats box to the inputs box
- `Bety Test` Button to test `dbfiles.input.insert`

## Screenshots

screenshot_2018-07-22 pecan data ingest
screenshot_2018-07-22 pecan data ingest 1
screenshot_2018-07-22 pecan data ingest 2

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/shiny/Data-Ingest/server_files/ingest_workflow_svr.R
+++ b/shiny/Data-Ingest/server_files/ingest_workflow_svr.R
@@ -98,13 +98,13 @@ observeEvent(input$createInput, {
 
 observeEvent(input$testBety, {
   
-  Shared.data$input_record_df <- PEcAn.DB::dbfile.input.insert(in.path = d1_tempdir,
-                                                                in.prefix = "hf103-01-eddy-2000-01.csv", #inputsList$Name,
+  Shared.data$input_record_df <- PEcAn.DB::dbfile.input.insert(in.path = c(local_tempdir),
+                                                                in.prefix = inputsList$Name,
                                                                 siteid =  "1000004955", # inputsList$siteID,
-                                                                startdate = "2000-07-01", #inputsList$StartDate,
-                                                                enddate =   "2014-07-01", #inputsList$EndDate,
+                                                                startdate = "2004-01-01", #inputsList$StartDate,
+                                                                enddate =   "2018-01-01", #inputsList$EndDate,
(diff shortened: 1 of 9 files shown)


tonygardella:
@LiamBurke24 general question: What does the Test Bety button do?

tonygardella:
Nevermind I see it adds a test input. Maybe add a way to delete the test input as well.

mdietze:
The latter. Do a dedicated query to create the format, as that is very distinct and quite complicated (once you add the formats_variables). Don't recall if there's an existing function for creating that record, but there's definitely one for querying it which you'll want to look at.  This also makes things simpler --- since a new format is optional (one might be using an existing format), by creating the format record first the input insert never has to deal with the case of whether format exists or not (as it always will)

LiamBurke24:
Want me to include the separate format record query in this PR?

shiny/Data-Ingest/server_files/ingest_workflow_svr.R
```diff
@@ -31,40 +28,103 @@ updateSelectizeInput(session, "InputParentName",  choices = input_names, server
...
+  inputsList$Notes <<- input$InputNotes
+  
+  ## Print List
+  output$summInputs <- renderPrint({inputsList})
+})
+
+
+observeEvent(input$testBety, {
+ output$TestPUT <- renderPrint(inputs$List)
+ inputsList$Path <- "/tmp/RtmpolQsqK/local_tempdir/hf103-01-eddy-2000-01.csv"
```
mdietze: this seems like an odd choice for a default

LiamBurke24: NOOOOOOOO. I missed deleting this while testing. :( Hang on.

PR #2009, opened 2018-07-23
Add ED2 SAS script

crollinson:
Function for Semi-Analytical Solution for speeding up finding steady-state in model spinup.

## Description

Runs for the PalEON project have been using a Semi-Analytical Solution to accelerate the rate at which slow pools achieve steady-state.  This is based on Xia et al. 2012 GMD 5:1259-1271 and was first developed for ED2 by Jaclyn Matthes and then Christy Rollinson.  The script uses ED's internal model structure to solve soil carbon pools and initialize the landscape mosaic aboveground vegetation structure based on disturbance rates and the observed pathway of succession within a single path.

Although this has been used by PalEON and some other work by Rollinson, it should still be considered developmental and further refinement of parameterization, etc. may be necessary.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


models/ed/R/SAS.ED2.R
```diff
@@ -0,0 +1,502 @@
...
+  #---------------------------------------
+  # Finding the mean soil temp & moisture
+  # NOTE:  I've been plyaing around with finding the best temp & soil moisture to initialize things
+  #        with; if using the means from the spin met cycle work best, insert them here
+  # This will also be necessary for helping update disturbance parameter
+  #---------------------------------------
+  slmsts <- calc.slmsts(slxsand, slxclay)
+  slpots <- calc.slpots(slxsand, slxclay)
+  slbs   <- calc.slbs(slxsand, slxclay)
+  soilcp <- calc.soilcp(slmsts, slpots, slbs)
```
mdietze: * These functions don't seem to be defined anywhere
* They seem redundant with the soil_params function https://github.com/PecanProject/pecan/blob/develop/modules/data.land/R/soil_utils.R#L26

models/ed/R/SAS.ED2.R
```diff
@@ -0,0 +1,502 @@
...
+  
+  # Calculating Soil fire characteristics
+  soilfr=0
+  if(abs(sm_fire)>0){
+    if(sm_fire>0){
+      soilfr <- smfire.pos(slmsts, soilcp, smfire=sm_fire)
+    } else {
+      soilfr <- smfire.neg(slmsts, slpots, smfire=sm_fire, slbs)
```
mdietze: more undefined functions

models/ed/R/SAS.ED2.R
```diff
@@ -0,0 +1,502 @@
...
+  }
+  # ========================
+  
+  A_decomp <- temperature_limitation * water_limitation # aka het_resp_weight
+  # *************************************
+  
+  # *************************************
+  # Calculate the steady-state pools
+  # NOTE: Current implementation weights carbon input by patch size rather than using the 
+  #       carbon balance from the oldest state (as was the first implementation)
```
mdietze: Not sure I follow the reason for switching to using a weighted average. Presumably you've also done a weighted average of outputs too not just inputs?  Does this give a result that's very different from the original and/or one that goes to steady state more quickly after the accelerated spin up?

models/ed/R/SAS.ED2.R
```diff
@@ -0,0 +1,502 @@
...
+  # *************************************
+  # -------------------
+  # Do the carbon and fast nitrogen pools
+  # -------------------
+  fsc_ss <- fsc_in_y[length(fsc_in_y)]/(fsc_loss * A_decomp)
+  ssl_ss <- ssl_in_y[length(ssl_in_y)]/(ssl_loss * A_decomp * Lc) # Structural soil C
+  ssc_ss <- ((ssl_loss * A_decomp * Lc * ssl_ss)*(1 - r_stsc))/(ssc_loss * A_decomp )
+  fsn_ss <- fsn_in_y[length(fsn_in_y)]/(fsc_loss * A_decomp)
```
mdietze: Despite the above comment talking about a weighted average, this code appears to just use the oldest age class, and doesn't weight by area.dist

models/ed/R/SAS.ED2.R
```diff
@@ -0,0 +1,502 @@
...
+  # -------------------
+  # *************************************
+  
+  # *************************************
+  # Replace dummy values in patch matrix with the steady state calculations
+  # *************************************
+  # Figure out what steady state value index we shoudl use 
+  # Note: In the current implementaiton this should be 1 because we did the weighted averaging up front, 
+  #       but if something went wrong and dimensions are off, use this to pick the last (etc)
+  p.use <- 1
```
mdietze: Not following this comment. To the extent that I can follow the code above, the variables below look like they should be scalars. And if they're not, I don't follow why we'd want to use the first value.

models/ed/R/SAS.ED2.R
crollinson: I didn't know about the soil_params function, so I'll check that out.  These are from ED's internal routines, so if they aren't consistent with soil_params, would it be better to add a new script with ED soil functions or tack it on here?

mdietze: FWIW soil_params is based on some R code that Marcos wrote specifically for ED2. That doesn't guarantee the params will be exactly the same, but I'm hopeful that's the case

mdietze:
Now that we survived ESA, just checking in to see what your schedule was for finishing up this PR?

crollinson:
Goal is by end of month.  Nothing will happen this week due to institutional obligations (undergrad program wrap-up, politicians, other visitors...).

When I get a chance, I'll make the requested changes and include diagnostic tests of weighted average versus oldest age.  Based on the regional spins, it'll be important to do these for a couple climates/sites since the original scheme worked well at Harvard Forest, but less so at other sites.   Is there a list of of PEcAn priority test sites that I should use?  I might not be able to do them all, but I could do maybe top 3 or so.

mdietze:
@robkooper I'm confused by the errors in @crollinson old PR, since they seem to be showing up in the met code, but I don't see any changes in that part of the code in this PR

```
test.download.NARR.R:22: skip: NARR download works as expected
Reason: On CI
────────────────────────────────────────────────────────────────────────────────

2020-11-01 16:57:47 SEVERE [load.cfmet.R#60: PEcAn.logger::logger.severe] : 
   run end date 253373615999 after met data ends 299770200 
⠏ |   0       | loading data from PEcAn-CF met drivers
2020-11-01 16:57:47 SEVERE [load.cfmet.R#57: PEcAn.logger::logger.severe] : 
   run start date -62167219200 before met data starts 284013000 
2020-11-01 16:57:47 SEVERE [load.cfmet.R#57: PEcAn.logger::logger.severe] : 
   run start date -599702400 before met data starts -599529600 
2020-11-01 16:57:47 SEVERE [load.cfmet.R#60: PEcAn.logger::logger.severe] : 
   run end date -567993601 after met data ends -568080000 
2020-11-01 16:57:47 SEVERE [load.cfmet.R#23: PEcAn.logger::logger.severe] : 
   lat / lon ( 39 , 20 ) outside range of met file ( 40.25 40.25 , -88.25 
   -88.25 
2020-11-01 16:57:47 SEVERE [load.cfmet.R#23: PEcAn.logger::logger.severe] : 
   lat / lon ( 9 , -88 ) outside range of met file ( 40.25 40.25 , -88.25 
   -88.25 
⠴ |  16       | loading data from PEcAn-CF met drivers
✔ |  21       | loading data from PEcAn-CF met drivers [0.2 s]
```

That said, @crollinson there are a bunch of errors that you need to fix:
```
R check of models/ed returned new problems:
Execution halted
checking Rd \usage sections ... WARNING: Undocumented arguments in documentation object 'SAS.ED2' 'prefix' 'kh_active_depth' 'decay_rate_fsc' 'decay_rate_stsc' 'decay_rate_ssc'
checking R code for possible problems ... NOTE: SAS.ED2: no visible binding for global variable 'blckyr'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'calc.slmsts'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'calc.slpots'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'calc.slbs'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'calc.soilcp'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'smfire.pos'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'smfire.neg'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'dgeom'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'complete.cases'
checking R code for possible problems ... NOTE: SAS.ED2: no visible global function definition for 'write.table'
```

mdietze:
closed by #2738

PR #2010, opened 2018-07-23
Function to sample ED IC from DBH posteriors

istfer:
## Description

Functionality for sampling @andydawson's Paleon DBH posteriors to produce IC files for ED.

The workflow runs back to back, but I'm aware that the implementation is not ideal. I had to conform to the old design and I didn't delve too much into design questions on purpose as I know @para2x is working on refactoring and ensemble generator. I'm happy to come back to this later and make it compatible with the new design.

Still submitting this PR to open it to discussion with @mdietze 

also updated read/write restart functions to nudge all bpools after the analysis because ED kills trees due to negative carbon balance, but still testing this

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


models/ed/R/sample.IC.ED2.R
```diff
@@ -0,0 +1,114 @@
+##' Samples intial conditions for ED2
+##' 
+##' NOTE: almost all of the fcn args can be extracted from pecan settings list, just wnated to be explicit
+##' 
+##' @param samples a dataframe with columns: tree id, spp info, Measurement_Year, DBH, plot id (if applicable), could come from load_data
```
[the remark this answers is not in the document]
mdietze: In addition, this looks like the type of dataframe that I'd expect for raw data, not for posterior samples of the latent state -- where is the column for MCMC rep?

models/ed/R/sample.IC.ED2.R
```diff
@@ -0,0 +1,114 @@
...
+    # prepare info for css/pss/site
+    css_info    <- list()
+    pss_info    <- list()
+    # site_info <- list()
+    
+    # css
+    patch.list <- list()
+    for(np in seq_len(n.patches)){
+      patch_sub  <- samples[samples$plot == np,] # these colnames might need generalization
+      samp_ind   <- tapply(seq_along(patch_sub$tree), patch_sub$tree, sample, 1)
```
mdietze: Not 100% sure I follow what this line is doing. Could use a comment and a bit of an explanation about the underlying assumptions.

models/ed/R/sample.IC.ED2.R
```diff
@@ -0,0 +1,114 @@
...
+  # remove rows with NAs (we don't want DBH to be NA but do we want to allow missing taxa?)
+  samples <- samples[complete.cases(samples), ]
+
+  # match spp info with species names 
+  # the matching can go into the ic-loop below, depending which is easier to handle sample size vs ensemble size
+  # for sample sizes of 10^5 this is still fast without sampling
+  spp_match <- match_species_id(input_codes = samples[[format_name]], format_name = format_name)
+  
+  # match pfts
+  pft_match   <- match_pft(bety_species_id = spp_match$bety_species_id, pfts = pfts)
```
mdietze: At a high level, this function seems to replicate, rather than build on, the existing veg workflow. This first bit is redundant. The middle bit that does the sampling isn't actually model specific, but should apply to all cohort models (and generalizable to all pool-based models using the cohort2pool function). The last bit seems easily generalizable to other veg2models.

models/ed/R/sample.IC.ED2.R
```diff
@@ -0,0 +1,114 @@
...
+    ic.list[[icn]] <- veg2model.ED2(outfolder, veg_info, start_date, new_site, paste0(source, "_ens", icn))
+    localfiles     <- paste0(sub("(.*)[.].*", "\\1", ic.list[[icn]]$filepath), c(".css", ".pss", ".site"))
+    remotefiles    <- paste0(inputs_path, "/", sub("(.*)[.].*", "\\1", ic.list[[icn]]$filename), c(".css", ".pss", ".site"))
+    # copy to remote and delete local
+    for(fls in seq_along(localfiles)){
+      PEcAn.remote::remote.copy.to(host_info, localfiles[fls], remotefiles[fls])
+      file.remove(localfiles[fls])
+    }
+    
+    ic.list[[icn]] <- data.frame(css = remotefiles[1], pss = remotefiles[2], site = remotefiles[3], stringsAsFactors = FALSE)
```
mdietze: To be able to insert new records into the DB, this return seems pretty different from what convert.inputs is expecting

models/ed/R/write.configs.ed.R
```diff
@@ -102,12 +102,14 @@ convert.samples.ED <- function(trait.samples) {
...
-##' @param inputs updated input paths coming from SDA workflow, will modify ED2IN
+##' @param inputs updated met paths coming from SDA workflow, will modify ED2IN
+##' @param IC updated css/pss/site paths coming from SDA workflow, will modify ED2IN
+##' 
 ##' @return configuration file and ED2IN namelist for given run
 ##' @export
 ##' @author David LeBauer, Shawn Serbin, Carl Davidson, Alexey Shiklomanov
 ##-------------------------------------------------------------------------------------------------#
-write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings$constants, check = FALSE, inputs = NULL, ...) {
+write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings$constants, check = FALSE, inputs = NULL, IC = NULL, ...) {
```
mdietze: Do not add an IC arg to write.configs.ED2. The IC info is all coming from the settings.

models/ed/R/sample.IC.ED2.R
```diff
@@ -0,0 +1,114 @@
+##' Samples intial conditions for ED2
+##' 
+##' NOTE: almost all of the fcn args can be extracted from pecan settings list, just wnated to be explicit
+##' 
+##' @param samples a dataframe with columns: tree id, spp info, Measurement_Year, DBH, plot id (if applicable), could come from load_data
+##' @param format_name how the spp info provided: usda, fia, latin_name, custom
```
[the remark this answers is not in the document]
istfer: sure, would you like me to rename this variable in the [match_species_id](https://github.com/PecanProject/pecan/blob/develop/modules/data.land/R/match_species_id.R#L38) as well then?

models/ed/R/sample.IC.ED2.R
```diff
@@ -0,0 +1,114 @@
...
+##' @param start_date settings$run$start.date
+##' @param pfts settings$pfts
+##' @param ne number of ensembles requested
+##' @param latitude settings$run$site$lat
+##' @param longitude settings$run$site$lon
+##' @param source the source of the samples, will appear as prefix in the IC filenames e.g. "HF_lyford.PalEON"
```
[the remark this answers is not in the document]
istfer: yeah, I was again trying to use the same argument name as IC workflow, there [source](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/veg2model.ED2.R#L7) is like FIA, FFT that appears in the name of the files. but I can change this too

models/ed/R/sample.IC.ED2.R
istfer: Yes, I am currently not writing records to the DB, it's `veg2model.ED2` function that returns convert.input compatible values. I'll see if I can incorporate this function into the IC workflow as you mention in the comment above

models/ed/R/write.configs.ed.R
istfer: I agree, I was following the SIPNET example. I guess, I can populate the settings$inputs list and disintegrate [somewhere here](https://github.com/PecanProject/pecan/blob/develop/modules/assim.sequential/R/sda.enkf.R#L79_L111). I just thought this will change after refactoring and ensemble generator anyway

models/ed/R/sample.IC.ED2.R
istfer: my only concern with complete integration with IC workflow is that IC workflow currently works for generating one IC file per site from one data source (naming files and inserting to the DB accordingly), but in this case we are generating multiple files for a single site from a single dataframe. I'll check how I can do that

on other ensemble naming and DB insertion, what have you decided? currently I'm not writing anything to the DB, I was just writing files like `HF_lyford.PalEON_ens*.1961.lat42.53lon-72.18.css` to a folder

mdietze: Yep, understood. @Luke-Dramko just tacked that problem for met.process and generalized `convert.input` to work with ensembles. 

The relevant code starts here:
https://github.com/PecanProject/pecan/blob/1041156104b544d0d31024f0028d75c66aac7eea/base/utils/R/convert.input.R#L34

An example of that being called is here:
https://github.com/PecanProject/pecan/blob/1041156104b544d0d31024f0028d75c66aac7eea/modules/data.atmosphere/R/download.raw.met.module.R

And here for an example of the downstream workflow using the ensemble (where the default of n=1 is just a special case):
https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/met.process.R#L253

models/ed/R/write.configs.ed.R
mdietze: Yeah, I think @para2x has refactored that approach away and hopefully the PR with that in place will be submitted very soon. 

The addition of an IC arg in write.configs.SIPNET started as a quick hack in the final sprint when I was trying to get SDA working for the first PalEON summer course. Deeply regret not fixing it long ago.

istfer:
@mdietze I think this PR is ready for re-review. I now got rid of the sample.IC.ED2 function and the IC and inputs args in write.configs.ED2

my main question is how to re-populate the settings after the ensembles, currenty I'm just returning sublists of inputs (and then loop over them in SDA):
```
 <inputs>
   <ens1>
    <css>...</css>
    <pss>...</pss>
    <site>...</site>
    <met>...</met>
    <lu>...</lu>
    <soil>...</soil>
    <thsum>...</thsum>
    <veg>...</veg>
   </ens1>
   <ens2>
    <css>...</css>
    <pss>...</pss>
    <site>...</site>
    <met>...</met>
    <lu>...</lu>
    <soil>...</soil>
    <thsum>...</thsum>
    <veg>...</veg>
   </ens2>
   ...
   <ensN>
    <css>...</css>
    <pss>...</pss>
    <site>...</site>
    <met>...</met>
    <lu>...</lu>
    <soil>...</soil>
    <thsum>...</thsum>
    <veg>...</veg>
   </ensN>
</inputs>
```
should I follow the multi.settings tags (e.g. <settings.1>) instead of these `<ens*>`  tags? what is the ensemble generator code supposed to return?

mdietze:
@para2x and/or @Luke-Dramko could you provide @istfer with an example of what the ensemble code returns as its updated settings.  My understanding was that you had ensemble members within inputs, while Istem's version above appears to have inputs within ensembles.

You should definitely NOT use the MultiSettings code to run an ensemble or SDA at a single site. A MultiSettings SDA should be reserved for when you're running SDA over multiple sites (e.g. when there is covariance across sites)

istfer:
the grouping and naming of tags are now compatible with the way ensemble generator handles things

I'm gonna need small tweaks when the ensemble generator and refactored SDA is fully coupled, but shouldn't be a big deal

ready for re-review

book_source/04_advanced_user_guide/02_adding_to_pecan/01_case_studies/04_input_conversion.Rmd
```diff
@@ -237,6 +237,56 @@ With the input id  in hand you can now edit a pecan XML so that the PEcAn functi
...
+        <source.id>1000015682</source.id>
+        <useic>TRUE</useic>
+        <ensemble>20</ensemble>
+        <metadata>
+          <area>1256.637</area>
+          <n.patch>3</n.patch>
+        </metadata>
+      </css>
+```
+Here the `source.id` should point to a file that has MCMC samples to generate the ensemble from. The number between the `<ensemble>` tag defines the number of ensembles requested. The workflow will populate the settings list `run$inputs` tag with ensemble member information. E.g.:
```
mdietze: In all other contexts, the input ID referenced within a specific in put is just called `<id>`. In the next version I think we should revert to that here, rather than deviating from that norm, once the veg IC workflow is capable of distinguishing raw data, pecan standard, and model-specific file IDs (which should be discernable from the format ID associated with that input ID)

models/ed/R/veg2model.ED2.R
```diff
@@ -39,12 +39,33 @@ veg2model.ED2 <- function(outfolder, veg_info, start_date, new_site, source){
...
+    }
+    if(!is.null(pss$age)){
+      age <- as.numeric(pss$age)
+    }else{
+      age <- 100
+    }
+    if(!is.null(pss$area)){
+      area <- as.numeric(pss$area)
+    }else{
+      area <- 1000 # m2
```
mdietze: This silent assumption is dangerous. I think this should be an error.

modules/assim.sequential/R/sda.enkf.R
```diff
@@ -225,6 +229,9 @@ sda.enkf <- function(settings, obs.mean, obs.cov, IC = NULL, Q = NULL, adjustmen
   
   for (i in seq_len(nens)) {
     
+    # is this gonna break other model runs? inputs is usually just the met path which is all they need anyway?
+    settings$run$inputs <- inputs[[i]]
```
mdietze: This could break things, but hopefully it's a very temporary change since much of this will change with the refactor

modules/data.land/R/ens.veg.module.R
```diff
@@ -0,0 +1,53 @@
+.ens.veg.module <- function(getveg.id, dbparms,
```
mdietze: Undocumented and un-exported function. Also, why are we starting these function names with a `.`?

modules/data.land/R/ic_process.R
```diff
@@ -83,93 +86,163 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
 
-    getveg.id <-.get.veg.module(input_veg = input, 
-                              outfolder = outfolder, 
-                              start_date = start_date, end_date = end_date,
-                              dbparms = dbparms,
-                              new_site = new.site,
-                              host = host, 
-                              machine_host = machine.host,
-                              overwrite = overwrite$getveg)
+    for(i in 1:nsource){
```
mdietze: seq_length preferable to 1:nsource

modules/data.land/R/ic_process.R
```diff
@@ -83,93 +86,163 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
-      # IF: For now IC workflow is only working for ED and it's the only case for copying to remote
-      # but this copy to remote might need to go out of this if-block and change
+      model_file <- db.query(paste("SELECT * from dbfiles where container_id =", putveg.id[[i]], "and machine_id =", machine$id), con)
       
-      # Copy to remote, update DB and change paths if needed
-      if (settings$host$name != "localhost") {
-        
-        remote_dir <- file.path(settings$host$folder, paste0(input$source, "_site_", str_ns))
+      # now that we don't have multipasses, convert.input only inserts 1st filename
+      # do we want to change it in convert.inputs such that it loops over the dbfile.insert?
```
mdietze: Not following this comment -- wasn't `put.veg.module` called in a loop so that putveg.id is a whole list of input IDs? Each ID should point to it's own filename, not the filename of the first ensemble member.

modules/data.land/R/ens.veg.module.R
istfer: as these functions are entirely for within package use, I thought it was kind of a convention to do that, but no other particular reason behind. I can get rid of them

modules/data.land/R/ic_process.R
istfer: this was an older comment, the comparison ("now that we have...") was with respect to an initial version of this workflow where I was going through the whole IC multiple times separately for css, pss and site. Now we get into IC-workflow for `css` initially but we end up writing `pss` and `site` too. 

convert.input only inserts `css` to the DB, so when we query file name only the one with `.css` extension comes up, everytime we need pss or site we need to swap for `.css` in the filename. but I think at one point we also discussed having only one entry for ED css/pss/site in the DB, so that they always go together.. although never really forced it I think

anyway, that's what this comment was about, otherwise each ensemble member has their own IDs pointing to their filenames as you said

modules/data.land/R/ens.veg.module.R
mdietze: More important (holding up the PR) is the lack of documentation and export

infotroph: I'll jump in to defend unexported as a good thing here -- If it's for in-package use, not exporting it means we can move/rename/delete it later without a deprecation rigamarole. And if we find later that we do need to export it, it's just as easy to do then as now.

(I do vote for a line or two of documentation, though)

PR #2014, opened 2018-07-24
Add component biomass calculations to cohort2pool

tonygardella:
Add the calculation of ratios of component biomass to total biomass in order to get leaf, woody, and coarse root biomass from DBH.

## Motivation and Context
Needs to expand the capability of stand data to inform initial conditions of models. Integration into veg inputs process is being tested right now.

@infotroph You're partnered with me for code review this week. Let me know if you do not have time to review.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/data.land/R/cohort2pool.R
```diff
@@ -30,27 +30,63 @@ cohort2pool <- function(veg_file, allom_param = NULL) {
...
+    b <- 2.5
   } else {
     print("user provided allometry parameters not yet supported")
     return(NULL)
   }
   
-  #Calculate AGB
+  #Calculate Total Biomass
+  #Need to Convert the total kg of biomass to (kg/m^2) but data on area of land does not exist
+  #In it's place,dividing by number of trees.
```
mdietze: This is not a sound assumption. The area of the land needs to be passed in explicitly and should be part of the veg workflow already.

mdietze: Also not clear why you switched from AGB to total biomass? There's going to be a lot less constraint on the total biomass allometries (much greater uncertainty) because of the rarity of belowground data

modules/data.land/R/cohort2pool.R
```diff
@@ -30,27 +30,63 @@ cohort2pool <- function(veg_file, allom_param = NULL) {
...
+  component_names <- c("foliage","coarse_root", "stem_bark","stem_wood")
+  
+  # Hardwood params
+  hwood_0params <-c(-4.08113,-1.6911,-2.0129,-0.3065)
+  hwood_1params <- c(5.8816, 0.8160, -1.6805,-5.4240)
+  hlist <-list(hwood_0params, hwood_1params)
+  # Softwood params
+  swood_0params <-c(-2.9584,-1.5619,-2.0980,-0.3737)
+  swood_1params <- c(4.4766,0.6614,-1.1432,-1.8055)
+  slist <-list(swood_0params, swood_1params)
```
mdietze: These allometries should not be hard coded as the only possible values, but should come in as part of the allom_param object, which itself should ultimately come from the PEcAn.Allom package. The values listed here may be included as defaults if other values are not specified, similar (and combined with) the AGB default earlier.

modules/data.land/R/cohort2pool.R
```diff
@@ -30,27 +30,63 @@ cohort2pool <- function(veg_file, allom_param = NULL) {
...
+  
+  
+  comp_ratios <- purrr::pmap(hlist,function(B0,B1){
+    
+    exp(B0 + (B1/dbh))
```
mdietze: why are you dividing B1/dbh. Power Law allometries take the form exp(B0 + B1*dbh). Again, we should probably just switch to using the allom module's predict function.

PR #2016, opened 2018-07-25
data.atmosphere: Draft of check input met file function

ashiklom:
## Description

- `check_met_input_file` checks whether target met input file conforms
  to the PEcAn standard.
- `check_met_variable` checks whether a variable from a NetCDF file is
  known to the CF standard.
- `get_cf_variables_table` downloads the current CF variables list as
  a `data.frame` from the CF conventions website.
- `pecan_standard_met_table` is a code version of the CF standard
  table from the PEcAn documentation ("17.5.1.1 Meteorological Data conversion")

## Motivation and Context

Per discussions from the PEcAn meeting on 25 July 2018.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/get_cf_variables_table.R
```diff
@@ -0,0 +1,24 @@
+#' Retrieve the current CF variables table from `cfconventions.org`
+#' and convert it into a `data.frame`
+#'
+#' @param cf_url URL of CF variables table XML
+#' @return CF variables table, as a `tibble`
+#' @author Alexey Shiklomanov
+#' @export
+get_cf_variables_table <- function(cf_url = "http://cfconventions.org/Data/cf-standard-names/57/src/cf-standard-name-table.xml") {
```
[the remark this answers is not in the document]
infotroph: (1) they have a "latest" version of the human-readable HTML table, but all XML versions have explicit versions numbers

(2) I'd argue this is correct -- We need to know what version we're supporting and be aware when we change it.

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+        testthat::expect_match(
+          ncdf4::ncatt_get(nc, "time", "units")[["value"]],
+          "^days since [[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2}T[[:digit:]]{2}:[[:digit:]]{2}:[[:digit:]]{2}Z$"
+        )
+         testthat::expect_true("latitude" %in% names(dimensions))
+         testthat::expect_equal(
+           ncdf4::ncatt_get(nc, "latitude", "units")[["value"]],
+           "degrees_north"
+         )
+         testthat::expect_true("longitude" %in% names(dimensions))
```
infotroph: Do we want to enforce dimension order? If so, all three of these could be bundled to `expect_equal(c("longitude", "latitude", "time"), names(dimensions))`

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+         testthat::expect_equal(
+           ncdf4::ncatt_get(nc, "longitude", "units")[["value"]],
+           "degrees_east"
```
infotroph: Another place where we can save typing if we trust the nc dimvals: `expect_equal(nc$dim$longitude$units, "degree_east")`

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+#'
+#' @param metfile Path of met file to check, as a scalar character.
+#' @param variable_table `data.frame` linking standard names to their
+#'   units. Defaults to [pecan_standard_met_table].
+#' @return `TRUE` (invisibly) if successful, otherwise throw an informative error.
+#' @author Alexey Shiklomanov
+#' @export
+check_met_input_file <- function(metfile, variable_table = pecan_standard_met_table) {
+
+  metfile <- normalizePath(metfile)
```
infotroph: Just for my curiousity: Why the normalization? I don't see any manipulation of the path itself and would have expected file.exists and nc_open be able to handle any path that normalizePath can, but I know you wouldn't do this if it wasn't needed.

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+    purrr::transpose()
+
+  test_vars_which_errors <- !purrr::map_lgl(test_vars[["error"]], is.null)
+  test_vars_errors <- test_vars[["error"]][test_vars_which_errors] %>%
+    purrr::map(conditionCall)
+  all_errors <- c(list(dimensions = test_dims), test_vars_errors)
+
+  if (length(all_errors) > 0) {
+    cat(unlist(all_errors), sep = "\n\n")
+    stop("At least one error was found.")
```
infotroph: If we need PEcAn.logger::severeifnot at the top of the function, shouldn't we use it here too? Or conversely, if stop() is OK here can we get away with stopifnot() up top?

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  invisible(TRUE)
```
infotroph: Why not return the actual result?

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+#' correct units.
+#' @param nc NetCDF object (as returned by [ncdf4::nc_open]).
+#' @param variable Variable to check, as a character scalar.
+#' @return `TRUE` (invisibly) if successful, otherwise throw an informative error.
+#' @author Alexey Shiklomanov
+#' @inheritParams check_met_input_file
+check_met_variable <- function(nc, variable, variable_table = pecan_standard_met_table) {
+  tryCatch(
+    testthat::test_that(
+      glue::glue("Variable '{variable}' has correct name and units."),
```
infotroph: I... did not realize you could glue together a test description! Cool!

modules/data.atmosphere/R/pecan_standard_met_table.R
```diff
@@ -0,0 +1,24 @@
...
+  "surface_downwelling_shortwave_flux_in_air"             , "W m-2"       , "solar_radiation"   , "rsdsAdjust"   , "swdown" , "dswrf" , "Rg"               ,
+  "surface_downwelling_photosynthetic_photon_flux_in_air" , "mol m-2 s-1" , "PAR"               , NA             , NA       , NA      , "PAR *(NOT DONE)*" ,
+  "precipitation_flux"                                    , "kg m-2 s-1"  , "cccc"              , "prAdjust"     , "rain"   , "acpc"  , "PREC (mm/s)"      ,
+  "wind_to_direction"                                     , "degrees"     , "wind_direction"    , NA             , NA       , NA      , "WD"               ,
+  "wind_speed"                                            , "m/s"         , "Wspd"              , NA             , NA       , NA      , "WS"               ,
+  "eastward_wind"                                         , "m/s"         , "eastward_wind"     , NA             , NA       , NA      , "CALC(WS+WD)"      ,
+  "northward_wind"                                        , "m/s"         , "northward_wind"    , NA             , NA       , NA      , "CALC(WS+WD)"
+)
```
infotroph: Is the intention to export this as a dataset? If so, I'd handle it something like PEcAn.utils::standard_vars, i.e. create the object in `data/pecan_standard_met_table.R`, document it in something like `R/datasets/.R`. If it's not intended to be user-visible and is strictly for internal use, I think Hadley recommends `R/sysdata.rda`

modules/data.atmosphere/R/get_cf_variables_table.R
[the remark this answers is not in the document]
ashiklom: I agree with @infotroph. For one, this isn't actually used anywhere yet -- I just thought it would be a nice helper. Second, you can already give it whatever URL you want in your scripts. Third, I agree with Chris -- we should be _explicit_ about which version we are currently supporting.

That said, since this does change somewhat frequently, I added a function for quickly building this URL from a version number.

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+
+  PEcAn.logger::severeifnot(
+    file.exists(metfile),
+    msg = glue::glue("File '{metfile}' does not exist.")
+  )
+
+  nc <- ncdf4::nc_open(metfile)
+
+  test_dims <- tryCatch(
+    testthat::test_that(
```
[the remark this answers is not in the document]
ashiklom: Good catch. I assumed it was already being imported, but it was in-fact in Suggests. I changed this accordingly.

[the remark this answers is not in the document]
dlebauer: Thanks for taking this on Alexy ... But, would it be more appropriate to use assertions rather than tests here? Eg https://cran.r-project.org/web/packages/assertthat/README.html

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,97 @@
...
+      "Dimensions are correct.",
+      {
+        dimensions <- nc[["dim"]]
+        testthat::expect_type(dimensions, "list")
+        testthat::expect_equal(length(dimensions), 3)
+        testthat::expect_true("time" %in% names(dimensions))
```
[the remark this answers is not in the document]
ashiklom: The problem I have with `$` subsetting is that it silently and "helpfully" accommodates incomplete names. So, for instance, `nc$nd` will happily return `3` because it completes to `ndim`. I think this behavior is maybe OK for interactive use, but I'd rather avoid it in rigorous tests or production code. In the past, I found a few places in PEcAn where `$` used the wrong name of a column or list element that didn't trigger an error because the new name was a superset of the old name (e.g. `days` vs. `days_in_year`).

modules/data.atmosphere/R/check_met_input.R
ashiklom: I'm not actually sure if the order of dimensions in `nc[["dim"]]` matters. I _think_ (and I may be wrong) that it can be different for every variable, and _there_ we _should_ actually check that the dimension order is correct because it affects interpretation of the grid (e.g. `time` in the first vs. last dimension). But I very well may be wrong about that.

modules/data.atmosphere/R/check_met_input.R
ashiklom: See my above comment on `$` subsetting. This is actually a great example, because `nc...$unit` and `nc...$units` will both work, but `nc...[["unit"]]` will fail.

modules/data.atmosphere/R/check_met_input.R
ashiklom: Ah, good catch. I should have used `PEcAn.logger::logger.severe` here. I'll change that accordingly.

ashiklom: I don't want to use the shorthand `severeifnot` because I want to neatly print out all the errors first.

modules/data.atmosphere/R/check_met_input.R
ashiklom: You can glue together anything, as long as you have access to variable in the environment!

In case you're curious, the environment thing bit me once when I tried to write code like this:

```r
my_logger <- function(string) {
  my_message <- glue::glue(string)
  cat(<some formatting> my_message)
}
my_logging_function <- function(var_value) {
  my_variable <- var_value
  my_logger("The value of my variable is {my_variable}")
}
my_logging_function(5)
```

This failed because `my_variable` was not propagated through to the `my_logger` function properly. There are probably clever ways to code around this, but I didn't think it was worth my time to figure them out.

modules/data.atmosphere/R/pecan_standard_met_table.R
ashiklom: So this is designed to be both useful internally and user-facing. As for how exactly it's stored, I'm not sure I see the benefit of storing it in `data/` or `systdata.rda`. Things in `data` either have to be in `.Rdata` format -- which requires them to be compiled elsewhere, and therefore adds another step to installing this package whenever that changes -- or CSV -- which admittedly is basically what this is already...but this way, the dataset documentation and raw data live in the same location. 

In general, I was following [this recommendation](http://r-pkgs.had.co.nz/data.html) from the "R Packages":

> A simple alternative to these three options is to include it in the source of your package, either creating by hand, or using dput() to serialise an existing data set into R code.

...which I took to mean the way this is presented here.

modules/data.atmosphere/R/check_met_input.R
ashiklom: What would the result be? `testthat::test_that` simply returns `TRUE` if all the tests pass, so I would just end up with a named list of `TRUE`s, which I suppose is informative in its own way, so I could do that.

modules/data.atmosphere/R/check_met_input.R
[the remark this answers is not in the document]
ashiklom: I was considering using `assertthat` (, but I was concerned that if I introduced another dependency, @serbinsh would actually murder me.

More seriously, besides introducing another dependency, one advantage of `testthat` over `assertthat` is that `testthat` expectations, when run inside of a `test_that` block, proceed with evaluation, whereas `assertthat::assert_that` is like `stopifnot` in that it fails on the first failing assertion. This is good here because it allows you to diagnose multiple problems at the same time -- e.g. if both `latitude` and `longitude` have the wrong unit, this code should spit out both errors, whereas using `assert_that`, you would discover one error, then fix the file, then return only to discover the second error, so you'd have to go back _again_ and fix that too (etc., etc.).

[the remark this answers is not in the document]
ashiklom: Case in point:

```r
metfile <- "~/Downloads/nobackup/CRUNCEP.2004.nc"
check_met_input_file(metfile)
# Test failed: 'Dimensions are correct.'
# * ncdf4::ncatt_get(nc, "latitude", "units")[["value"]] not equal to "degrees_north".
# 1/1 mismatches
# x[1]: "degree_north"
# y[1]: "degrees_north"
# * ncdf4::ncatt_get(nc, "longitude", "units")[["value"]] not equal to "degrees_east".
# 1/1 mismatches
# x[1]: "degree_east"
# y[1]: "degrees_east"
# 2018-07-26 10:40:07 SEVERE [check_met_input.R#62: PEcAn.logger::logger.severe] : 
#    At least one error was found. 
# Error in PEcAn.logger::logger.severe("At least one error was found.") : 
#   At least one error was found.
```

I like this output (1) the context of the error is clear ("Dimensions are correct"), and (2) all the errors show up at once, with informative error messages.

modules/data.atmosphere/R/check_met_input.R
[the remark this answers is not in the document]
infotroph: silent completion is a valid concern, though I submit it's less of a concern here because these names are controlled by the ncdf4 class, not assigned by a file's author. In fact I think I've answered my "is it safe..." question: Whichever one we actually check, I'd say one is enough and that if `length(dimensions) != nc$ndim` that means there's something wrong beyond the scope of our validation.

modules/data.atmosphere/R/check_met_input.R
infotroph: Testthat also documents that it supports limited quasiquotation, but I haven't bothered to make it work for anything more advanced than `for (i in ...){expect_equal(foo()[[!!i]], !!i + 10)}` I wonder if that works in the descriptions too...

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -0,0 +1,102 @@
...
+        time_regex <- paste0(
+          "^(seconds|minutes|hours|hours|days) since ",
+          "[[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2}",
+          "T[[:digit:]]{2}:[[:digit:]]{2}:[[:digit:]]{2}Z$"
+        )
+        testthat::expect_match(
+          ncdf4::ncatt_get(nc, "time", "units")[["value"]],
+          time_regex
```
infotroph: This is probably the place to check for a `calendar` attribute too

modules/data.atmosphere/R/check_met_input.R
infotroph: Good catch, I was conflating those.

modules/data.atmosphere/R/check_met_input.R
infotroph: I reluctantly concede your point :)

modules/data.atmosphere/R/check_met_input.R
infotroph: Yeah, named list of bools seems right -- all `TRUE` if OK, `all_errors` otherwise? Or am I misunderstanding the contents of `all_errors`? In general, I just think it ought to always *return* a result rather than / in addition to side-effecty error messages.

PR #2017, opened 2018-07-26
Ensemble generator

para2x:
## Description

This pull request contains a major change in how we generate ensembles in pecan. The revised `write.ensemble.configs ` function handles both making ensembles for new fresh simulations as well as continuation of a sequence (like what we are dealing with in  SDA) given a set of initial conditions, parameters and drivers. 
Within the pecan xml this new structure is designed for managing different sources of uncertainty. 
```
<samplingspace>
<parameters>
   <method>lhc</method>
</parameters>
<met>
   <method>sampling</method>
</met>
<soil>
   <method>sampling</method>    
</soil>
<vegetation>
   <method>sampling</method>
   <parent>met</parent>
</vegetation>
</samplingspace>
```
Each piece in the sampling space can potentially have a method tag and a parent tag. Method refers to the sampling method and parent refers to the cases where we need to link the samples of two components. When no tag is defined for one component,  one sample will be generated and used for all the ensembles. This allows for partitioning/studying different sources of uncertainties. For example, if no met tag is defined then one met path will be used for all the ensembles and as a result the uncertainty will come from the variability in the parameters.
At the moment no sampling method is implemented for soil and vegetation.

## Motivation and Context

The main motivation behind this change is that it offers a generalized solution for generating ensembles which can be used in any other analysis such as SDA. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


base/utils/R/ensemble.R
```diff
@@ -191,96 +214,194 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+      )
+      cat(run.id, file = file.path(settings$rundir, "runs.txt"), sep = "\n", append = TRUE)
+     
     }
-    runs[counter, "id"] <- run.id
+    return(invisible(list(runs = runs, ensemble.id = ensemble.id, samples=samples)))
+    #------------------------------------------------- if we already have everything ------------------        
+  }else{
+    #browser()
+    #reading retsrat inputs
```
mdietze: typo

base/utils/R/ensemble.R
```diff
@@ -191,96 +214,194 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+  
 } # write.ensemble.configs
+
+
+
+#' Function for reurning the met ensuble based on sampling method, parent or etc
+#'
+#' @param settings list of PEcAn settings
+#' @param method Method for sampling - For now looping or sampling with replacement is implemented
+#' @param parenids If met's ids are read from a parent what are the ids
```
mdietze: Is there are reason to call this `parenids` rather than `parent_ids`? The latter is much easier to understand. Also, because every db table has IDs, clarify exactly what IDs you need here (input? dbfile? ensemble?). Finally, no need to reference 'met' specifically if this has been generalized.

base/utils/R/ensemble.R
```diff
@@ -191,96 +214,194 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+#' @param ...
```
mdietze: Given usage below, you're expecting something very specific to be passed via `...`. Please document/explain what that is.

base/utils/R/ensemble.R
```diff
@@ -191,96 +214,194 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+#' @export
+#'
+#' @examples
+input.ens.gen<-function(settings,input,method="sampling",parenids=NULL,...){
+  #-- reading the dots and exposing them to the inside of the function
+  samples<-list()
+  dots<-list(...)
+  if (length(dots) > 0 ) lapply(names(dots), function(name){assign(name, dots[[name]], pos=1 )})
+
+  if ( tolower(input)=="met" ){
```
mdietze: bit confused -- this function looks like it will only run if input == met, but I don't see anything in the code below that is met specific other than `settings$run$inputs$met`, which could easily be abstracted away, e.g. `settings$run$inputs[[input]]`

mdietze: That said, aren't parameters a special case (need to be sampled from one file and passed as a list, rather than sampling file paths or IDs). I don't recall the earlier code handling that special case, so either there needs to be that case in this function or in the code that calls it.

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -165,20 +165,43 @@ Only if this section is defined an ensemble analysis is done.
...
+    <method>sampling</method>
+    <parent>met</parent>
+  </vegetation>
+  
+  </samplingspace>
 </ensemble>
 ```
 
 * **size** : [required] the number of runs in the ensemble.
-* **variable**: [optional] name of one (or more) variables the analysis should be run for. If not specified, sensitivity.analysis variable is used, otherwise default is GPP.
```
mdietze: Why has the `variable` tag been removed? At the moment that still needs to be there for the postprocessing code

base/utils/R/ensemble.R
para2x: I was actually trying to reserve the possibility that this function could handle other components like soil and vegetation later on with different variable types. For the met we have just a vector of paths but just like parameters for soils we may have a list of dataframes or a list of lists. That's why I didn't want to assume that the rest of components also would have similar variable type as met. I felt like this would give the next developers the peace of mind that their work would be independent of met and parameters.
Parameters are also handled in line 262-263 of ensemble.R

mdietze: Yes, I see that there's code that handles the case where parameters samples is null, but that comes after the loop over everything that's in `samplingspace` so I wanted to make sure that this function handled being called for the `parameter` case correctly. It's not obvious to me that it does -- if parameters are handled later why not have an explicit return right at the start when input == parameters (and add a comment to that effect in the loop around line 254.

I do think you can assume that all other inputs are handled the same way as met -- this is why I was pushing you to not have input specific functions. Soils, veg, etc will all exist as ensembles of files since write.configs.[MODEL] always expects inputs to be passed in as file paths.

para2x: Cool,... I'll change the code accordingly then.

base/utils/R/ensemble.R
```diff
@@ -184,7 +184,7 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
 ##' @param restart In case this is a continuation of an old simulation. restart needs to be a list with name tags of runid, inputs, new.params (parameters), new.state (initial condition), ensemble.id (ensemble id), start.time and stop.time.See Details.
 ##' @return list, containing $runs = data frame of runids, $ensemble.id = the ensemble ID for these runs and $samples with ids and samples used for each tag.  Also writes sensitivity analysis configuration files as a side effect
 ##' @details Resatrt functionality here is developed using model specific functions called write_restart.modelname . You need to make sure first that this function is already exsit for your dersired model.
```
mdietze: typo

base/utils/R/ensemble.R
```diff
@@ -182,13 +181,17 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
 ##' @param settings list of PEcAn settings
 ##' @param write.config a model-specific function to write config files, e.g. \link{write.config.ED}  
 ##' @param clean remove old output first?
-##' @return list, containing $runs = data frame of runids, and $ensemble.id = the ensemble ID for these runs. Also writes sensitivity analysis configuration files as a side effect
+##' @param restart In case this is a continuation of an old simulation. restart needs to be a list with name tags of runid, inputs, new.params (parameters), new.state (initial condition), ensemble.id (ensemble id), start.time and stop.time.See Details.
+##' @return list, containing $runs = data frame of runids, $ensemble.id = the ensemble ID for these runs and $samples with ids and samples used for each tag.  Also writes sensitivity analysis configuration files as a side effect
+##' @details Resatrt functionality here is developed using model specific functions called write_restart.modelname . You need to make sure first that this function is already exsit for your dersired model.
```
mdietze: multiple typos

base/utils/R/ensemble.R
```diff
@@ -182,13 +181,17 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
+##' new state is mainly a dataframe with a different column for different variables for n rows and n sample size. new.params also has similar structure to ensemble.samples which is sent as an argument.
```
mdietze: Needs more detail. What sorts of variables? How are columns named? Units? Is there other code are you expecting to generate this that you should be cross-referencing?

Also, it doesn't make sense to me that `new state` would be a square n-by-n matrix. Why would the number of states and the number of ensemble members be the same? Also are variables really in the rows not the columns?

base/utils/R/ensemble.R
```diff
@@ -212,96 +215,214 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+      myparent<-samp.ordered[[i]]$parent # do I have a parent ?
+      #call the function responsible for generating the ensemble
+      samples[[names(samp.ordered[i])]] <- input.ens.gen(settings=settings,
+                                                         input=names(samp.ordered)[i],
+                                                         method=samp.ordered[[i]]$method,
+                                                         parent_ids=if( !is.null(myparent)) samples[[myparent]] # if I have parent then give me their ids - this is where the ordering matters making sure the parent is done before it's asked
+                                                         )
+    }
+
+    # if there is a tag required by the model but it is not specified in the xml then I replicare n times the first element
```
mdietze: typo

base/utils/R/ensemble.R
```diff
@@ -212,96 +215,214 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+                                                         parent_ids=if( !is.null(myparent)) samples[[myparent]] # if I have parent then give me their ids - this is where the ordering matters making sure the parent is done before it's asked
+                                                         )
+    }
+
+    # if there is a tag required by the model but it is not specified in the xml then I replicare n times the first element 
+    required_tags%>%
+      purrr::walk(function(r_tag){
+        if (is.null(samples[[r_tag]]) & r_tag!="parameters") samples[[r_tag]]$samples <<- rep(settings$run$inputs[[tolower(r_tag)]]$path[1], settings$ensemble$size)
+      })
+    # if no ensemble piece was in the xml I replicare n times the first element in params
```
mdietze: typo

base/utils/R/ensemble.R
```diff
@@ -212,96 +215,214 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
-
-  return(invisible(list(runs = runs, ensemble.id = ensemble.id)))
+  
+  
+  
 } # write.ensemble.configs
+
+
+
+#' Function for reurning the met ensuble based on sampling method, parent or etc
```
mdietze: typos. Also, no longer met specific

base/utils/R/ensemble.R
```diff
@@ -212,96 +215,214 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+#'
+#' @examples
+#' \dontrun{input.ens.gen(settings,"met","sampling")}
+#'  
+input.ens.gen<-function(settings,input,method="sampling",parent_ids=NULL){
+
+  #-- reading the dots and exposing them to the inside of the function
+  samples<-list()
+  samples$ids<-c()
+      #
```
mdietze: Unnecessary indent

base/utils/R/run.write.configs.R
```diff
@@ -165,8 +166,24 @@ runModule.run.write.configs <- function(settings, overwrite = TRUE) {
     return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = FALSE))
   } else if (PEcAn.settings::is.Settings(settings)) {
     write <- settings$database$bety$write
-    ens.sample.method <- settings$ensemble$method
-    return(run.write.configs(settings, write, ens.sample.method, overwrite = overwrite))
+
+    #Old version of pecan xml which don't have a sampling space if it's not specified either it's an old xml or it's just not especified.
```
mdietze: 1. As noted by @infotroph this function has been deprecated (see line 13 above) and changes should be made in PEcAn.workflow.
2. typos
3. the catching and fixing of old pecan.xml tags should be done in the PEcAn.settings module.

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -13,7 +13,23 @@ runModule.run.write.configs <- function(settings, overwrite = TRUE) {
     return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = FALSE))
   } else if (PEcAn.settings::is.Settings(settings)) {
     write <- settings$database$bety$write
-    ens.sample.method <- settings$ensemble$method
+    #Old version of pecan xml which don't have a sampling space if it's not specified either it's an old xml or it's just not especified.
```
mdietze: see previous comment about typos and migration to PEcAn.settings

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -13,7 +13,23 @@ runModule.run.write.configs <- function(settings, overwrite = TRUE) {
...
+        settings$ensemble$samplingspace$parameters$method <- "uniform"
+        
+        ens.sample.method <-"uniform"
+      }
+    }else{
+      
+      #This is where I find the sampling method and pass it to the run.write.config which is responsible for generating samples for both ENS and SA.
+      ens.sample.method <- settings$ensemble$samplingspace$parameters$method
+    }
+    browser()
```
mdietze: I assume this is here for debugging and should be removed

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -165,20 +165,36 @@ Only if this section is defined an ensemble analysis is done.
...
 
-`write.configs()`, `run.ensemble.analysis()`
+`write.configs()`,`write.ensemble.configs`, `run.ensemble.analysis()`
 
 #### tags
 
 ```xml
 <ensemble>
 	<size>5</size>
-  <variable>GPP</variable>
```
mdietze: As noted previously, `variable` tag should not be removed

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -165,20 +165,36 @@ Only if this section is defined an ensemble analysis is done.
...
+  </samplingspace>
 </ensemble>
 ```
 
 * **size** : [required] the number of runs in the ensemble.
 * **variable**: [optional] name of one (or more) variables the analysis should be run for. If not specified, sensitivity.analysis variable is used, otherwise default is GPP.
+* **samplingspace**:[optional] Contains tags for defining how the ensembles will be generated.
+
+Each piece in the sampling space can potentially have a method tag and a parent tag. Method refers to the sampling method and parent refers to the cases where we need to link the samples of two components. When no tag is defined for one component, one sample will be generated and used for all the ensembles. This allows for partitioning/studying different sources of uncertainties. For example, if no met tag is defined then, one met path will be used for all the ensembles and as a result the outpus's uncertainty will come from the variability in the parameters. At the moment no sampling method is implemented for soil and vegetation.
+Availbe sampling methods for `parameters` can be found in the documentation of the `get.ensemble.samples` function in `PEcAn.utils` package.
```
mdietze: typo

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -165,20 +165,36 @@ Only if this section is defined an ensemble analysis is done.
...
+For the cases where we need simulations with a predefined set of parameters, met and initial condition we can use the restart argument. Restart needs to be a list with name tags of runid, inputs, new.params (parameters), new.state (initial condition), ensemble.id (ensemble ids), start.time and stop.time.
+
+The resatrt functionality is developed using model specific functions by called write_restart.modelname. You need to make sure first that this function is already exsit for your dersired model.
```
mdietze: multiple typos

base/db/R/db_merge_into.R
```diff
@@ -11,7 +11,7 @@
 #' db_merge_into(iris[1:12,], "iris", irisdb)
 #' dplyr::tbl(irisdb, "iris") %>% dplyr::count()
 db_merge_into <- function(values, table, con, by = NULL, drop = FALSE, ...) {
-  values_fixed <- match_dbcols(values, table, con, drop = FALSE)
+  values_fixed <- match_dbcols(values, table, con, drop = drop)
```
infotroph: These changes to `db_merge_into` seem like they deserve their own PR. I can't infer from the code whether this is a bug fix, an interface extension, or an ensemble-specific change, and even if it is ensemble-specific it will be helpful to future maintainers of the DB functions if this has its own history that documents:

* Why the change was needed 
* How it changes function behavior
* If fixing a bug, there should be a link to an associated issue number and new tests that cover the bug case
* If changing the existing behavior of `db_merge_into` to accommodate  the ensemble case, how do we demonstrate that `db_merge_into` still behave correctly for non-ensemble cases? Tests would be ideal here, but I realize adding them post facto is often hard to do.

para2x: Well... This was not specific to ensemble really and it was mainly a bug fix. The two arguments of by and drop are never used in the code. I ran into a problem with the last inner join in this function which without the by tries to join the two tables using all the columns.
You're right though, I stopped using this function because it needs a separate PR with more changes and also it slows down the write.ensemble.configs.

PR #2021, opened 2018-08-01
Ensemble mods

Luke-Dramko:
## Description

This pull request is the sum of a number of tweaks to the PEcAn workflow which allow NOAA_GEFS ensemble runs to work properly.  Several changes were made; of note, a custom gapfilling function (`gapfill.NOAA_GEFS`) was added, and the `do_conversions` sub-workflow was modified and tweaked to accompany this change.

Several other changes were made to SIPNET's template files and config writing script.

This pull request includes a version of the new version of ensemble.R and run.write.configs.R because these are necessary for it to work.  However, the final versions of these pieces of code are handled in #2017 .

## Motivation and Context

These changes are part of an attempt to adapt PEcAn to handle ensembles and forecasts.  This, in turn, is part of an initiative to use PEcAn in NEFI.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -15,19 +15,22 @@
...
   writeLines(config.text, con = file.path(settings$rundir, run.id, "sipnet.in"))
 
   ### WRITE *.clim
-  template.clim <- settings$run$input$met$path  ## read from settings #typo in inputs?
-  if (!is.null(inputs)) {
-    ## override if specified in inputs
-    if ("met" %in% names(inputs)) {
-      template.clim <- inputs$met$path
-    }
+  template.clim <- settings$run$inputs$met$path$path1     ## read from settings #typo in inputs? Fixed.
```
infotroph: If fixed, remove comment

modules/data.atmosphere/R/met2model.module.R
```diff
@@ -41,7 +41,7 @@
     fcn <- paste0("met2model.", model)
     lst <- site.lst(site.id=site$id, con=con)
     
-    model.id <- PEcAn.utils::convert.input(input.id = input.id,
+    model.id <- convert.input(input.id = input.id,  ###PEcAn.utils::
```
infotroph: Old version was correct: Please use `pkgname::function()` consistently for all functions that live outside the current package.

modules/data.atmosphere/R/metgapfill.R
```diff
@@ -677,6 +677,6 @@ metgapfill <- function(in.path, in.prefix, outfolder, start_date, end_date, lst
   ## 5b. May need to have an edge case for precipitation
   ## Step 6.Replace gaps with debiased time series (perhaps store statistics of fit somewhere as a measure of uncertainty?)
   ## Step 7. Write to outfolder the new NetCDF file
-
+
```
infotroph: This seems to be a whole bunch of whitespace changes in an otherwise untouched file. Are they necessary? If not, I'd prefer the changes be removed so it's easier to understand the rest of the PR.

base/workflow/man/run.write.configs.Rd
```diff
@@ -2,39 +2,34 @@
...
+Main driver function to call the ecosystem model specific (e.g. ED, SiPNET) 
+run and configuration file scripts
 }
 \details{
-The default value for \code{posterior.files} is NA, in which case the
-   most recent posterior or prior (in that order) for the workflow is used.
-   When specified, \code{posterior.files} should be a vector of filenames with one entry for each PFT.
-   Specify filenames with no path; PFT outdirs will be appended. This forces use of only
-   files within this workflow, to avoid confusion.
+DEPRECATED: This function has been moved to the PEcAn.workflow package and will be removed from PEcAn.utils.
```
infotroph: 1. Why are there changes to run.write.configs.Rd in a PR that doesn't change run.write.configs.R? 

2.  This is the man page for `PEcAn.workflow::run.write.configs`, which is the version that isn't deprecated, so it should definitely NOT contain the deprecation notice from `PEcAn.utils::run.write.configs`.

modules/data.atmosphere/R/metgapfill.NOAA_GEFS.R
```diff
@@ -0,0 +1,199 @@
+##'@title Gapfill NOAA_GEFS weather data
+##'@section Purpose:
+##'This function uses simple methods to gapfill NOAA GEFS met data
+##'Temperature and Percipitation are gapfilled with spline; other data sources are gapfilled with
```
infotroph: typo "Percip"

modules/data.atmosphere/R/met2model.module.R
Luke-Dramko: Whoops, that's something I did for debugging.  I'll definitely change that back.  I can't believe I missed that.

base/workflow/man/run.write.configs.Rd
Luke-Dramko: Ugh.  That's from a mix of bad version controlling and the switch to roxygen2 6.1.0.  After I ran `make` after updating to 6.1.0, there were about a large number of .Rd files to commit.  Additionally, I'd been using an intermediate version of run.write.configs in for code that was being developed concurrently.  During part of that process, run.write.configs was moved to PEcAn.utils.  

In any case, it's a mistake, so I'll get that fixed.

modules/data.atmosphere/R/metgapfill.R
Luke-Dramko: I originally thought I had to modify the function, but then found out it wouldn't be suitable for what I wanted to do and had to write a different one.  I thought I had reverted back to the original version, but I can commit a version copy-and-pasted from develop.

modules/data.atmosphere/R/metgapfill.NOAA_GEFS.R
Luke-Dramko: Good catch.

base/utils/R/convert.input.R
```diff
@@ -10,7 +10,7 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
                           end_date, pkg, fcn, con = con, host, browndog, write = TRUE, 
                           format.vars, overwrite = FALSE, exact.dates = FALSE, 
                           allow.conflicting.dates = TRUE, insert.new.file = FALSE, pattern = NULL,
-                          forecast = FALSE, ensemble = FALSE,...) {
+                          forecast = FALSE, ensemble = FALSE, ensemble_name = NULL, ...) {
```
mdietze: Undocumented new argument. What is this and do we really need it? Looks like we're just using it as `paste(filename_pattern, ensemble_name, sep="\\.")` which suggests to me that it's just simpler to ask the function calling this to do that paste when it creates the pattern it's looking for.

base/utils/R/convert.input.R
```diff
@@ -523,8 +525,12 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
     
     if (forecast && !is.null(input.id) && !is.na(input.id)) { # for met2model coversion, arguments will be extraneous otherwise.
       fcn.args$year.fragment = TRUE
-      parent.inputfile.name <- PEcAn.DB::db.query(paste0("SELECT name FROM inputs WHERE id =", input.id), con)
-      fcn.args$in.data.file = parent.inputfile.name$name #Sends the file name (minus the extension)
+      if (grepl("NOAA_GEFS\\.[0-9]+", fcn.args$in.prefix)) { # Protects legacy database entries.
```
mdietze: Given that the 'legacy' is just a few weeks, I would go fix the legacy database entries rather than introduce a permanent hack into convert.input

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -15,19 +15,22 @@
...
   writeLines(config.text, con = file.path(settings$rundir, run.id, "sipnet.in"))
 
   ### WRITE *.clim
-  template.clim <- settings$run$input$met$path  ## read from settings #typo in inputs?
-  if (!is.null(inputs)) {
-    ## override if specified in inputs
-    if ("met" %in% names(inputs)) {
-      template.clim <- inputs$met$path
-    }
+  template.clim <- settings$run$inputs$met$path$path1
```
mdietze: we should not be getting input paths in this format

modules/data.atmosphere/R/download.raw.met.module.R
```diff
@@ -38,7 +38,7 @@
     
   } else if (register$scale == "site") {
     # Site-level met
-    raw.id <- PEcAn.utils::convert.input(input.id = NA,
+    raw.id <- convert.input(input.id = NA,  ### PEcAn.utils::
```
mdietze: Should not be removing package names from functions

base/utils/R/convert.input.R
Luke-Dramko: That is what it's used for.  The issue is that sometimes convert.input handles all ensembles at once (during the raw data download) and sometimes it handles individual ensemble members (during gapfilling and met2model processing), where it's being called iteratively by met.process.  In the latter case, convert.input is not aware of which ensemble member it's trying to process, and needs that information to properly check if the file already exists in the database.

If I put this code in `metgapfill.module` and `met2model.module`, convert.input wouln't be able to properly build its regular expressions when processing ensembles iteratevely in the download stage.  convert.input is only called once in the download stage; download.NOAA_GEFS generates the entire ensemble of files at once, which need to be iterated over.  convert.input wouldn't be able to build its regular expression because the code for that would have been moved to outside the function.

In short, the problem is that the iteration is happening in two different places: sometimes in convert.input and sometimes in met.process.

The regular expression thing is a workaround the fact that the inputs table doesn't have a column for ensemble, so it's impossible to uniquely identify by column content alone.  The only place where ensemble identification is found is in the file name, (and folder name), which is why a regex is used to match the appropriate file name.  Even if an ensemble column did exist, however, convert.input would still need the information about which ensemble it's trying to process.

None of the arguments to convert.input have documentation on top of the function header.  I can write some.  Where else am I supposed to add documentation for new arguments?

modules/data.atmosphere/R/download.raw.met.module.R
Luke-Dramko: Ah, whoops, that was for debugging.  I can fix that.

models/sipnet/R/write.configs.SIPNET.R
Luke-Dramko: The decision to add list depth and use a list instead of a vector was discussed here: https://pecanproject.slack.com/archives/C8RD9BRLK/p1531508672000023?thread_ts=1531504235.000165&cid=C8RD9BRLK .  All input paths were wrapped in an additional list, even if there was only one, for consistency.  This was part of pull request #2004 .  

The path settings remain(ed) unchanged throughout the workflow, and write.configs.SIPNET was changed to reflect this.

Is it the responsibility of `met.process` to return file paths in settings differently, or is it the responsibility of `run.write.configs` to format the arguments differently? (Relevant thread here: https://pecanproject.slack.com/archives/C8RD9BRLK/p1533301225000306?thread_ts=1533246209.000194&cid=C8RD9BRLK).

mdietze: I understand that met.process and the other do.conversions functions will be generating ensemble Inputs that have multiple paths and multiple ID's, but it is the responsibility of `run.write.configs` to pass every `write.configs.[model]` a single specific path for every Input, and that input is expected to be at `settings$run$inputs$<input>$path`. This change doesn't just break SIPNET, it breaks every single model and add unnecessary complexity/confusion to write.configs.[model]. No model needs to know whether it's being run individually or in an ensemble or which ensemble member it is.

Luke-Dramko: This code happened to be causing a merge conflict anyway, so I resolved the conflict in favor of the develop version.  So long as the changes to `run.write.configs` in #2025 send a single input path at `settings$run$inputs$<input>$path`, this should be good to go.

base/utils/R/convert.input.R
```diff
@@ -60,10 +60,12 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
       # site.id, a regex placeholder is used instead.
       
       filename_pattern = pattern #pattern is always the name of the meteorological data product (met).
-      if (!is.null(input.args$sitename)) {
+      if (!is.null(input.args$sitename) || !is.null(site.id)) {
```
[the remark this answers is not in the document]
Luke-Dramko: It does.  `filename_pattern` is interpreted by `grepl` in `dbfile.input.check`.  `grepl` interprets filename_pattern as a regular expression, and returns TRUE if there are any matches anywhere in the string.  In order to specify a regular expression that must begin at the start of the string, there would have to be a `^` character at the start of the regular expression.  

If pattern = NULL, this is essentially just searching for an ensemble number/name surrounded by `.`.  (The filename standard for ensembles indicates that they should be surrounded by `.`).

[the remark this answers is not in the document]
infotroph: Got it, thanks.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -15,6 +15,7 @@
 ##' @author Michael Dietze
 write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs = NULL, IC = NULL,
                                 restart = NULL, spinup = NULL) {
+
```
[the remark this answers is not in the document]
Luke-Dramko: Extra comment has been removed.

Luke-Dramko:
Hypothetically, we could only have one ensemble argument.  Right now, what `ensemble` does is it's an integer equal to the number of ensembles if the data produce is an ensemble, and FALSE if the data product is not an ensemble.  This information is used to set up a loop in the existing-input checking stage of `convert.input`.  If convert.input is being called iteratively, it doesn't need this information, because it's only processing one file.  It does, however, need a way to identify the file that it's processing.  In this case, `ensemble` could be the ensemble identifier, which is the role ensemble_name is playing.

The problem is that sometimes, the name/identifier of the ensemble member _is_ an integer (as it is for our only ensemble data product thus far).  You could do something clever with the data types, (perhaps the number of ensembles must be an integer and the ensemble identifier must being a string/character), but where data types are so fluidly coerced in R, often as a side effect to other operations, this seems like a very easy way to get mixed up and have a tricky-to-track-down bug.

I thought of this relatively recently, but still think the extra argument is the way to go for the sake of clarity - you can depend on what each variable is doing.  I can definitely change it if you think differently.

Luke-Dramko:
As far as tests are concerned, adding a test for convert.input does sound like a good idea.  Currently, the existing `pecan/tests/travis.sipnet.xml` doesn't even touch convert.input - its file path is recognized and returns from met.process before convert.input is called.  

The tricky part about making tests for convert.input is that its behavior is dependent on the state of the database.  The downloading/inserting new inputs part of the code would only occur once, and once the record was in the database, it would continue to be return early without testing most of the code.  

It would be most beneficial to
1. Remove the record of the test file from the database.
2. Test to see if convert.input can properly download the right file and insert it into the database.
3. See if convert.input recognizes the new file.

Is this something that would go in the `travis.yml`?  I think (but am not sure) that Travis only looks for failed commands (nonzero exit statuses) as reasons to fail a build, and if step 3 was messed up, Travis could still pass it.

Of course, because there's only one database entry, multiple Travis instances running this concurrently could mess each other up on step 3.

There are a lot of different routes convert.input can take too - is the existing input file checking done in the forecast section, the exact.dates section, or the default file appending section?  Is data converted with browndog or locally?

PR #2023, opened 2018-08-02
Nefi scripts

Luke-Dramko:
## Description

This collection of scripts help run PEcAn every morning and plug directly into a CRON job.  They are not part of the workflow, but I've found them useful or necessary for developing and running iterative forecasting.  All of these files are in `pecan/modules/assim.sequential/inst/`.  The request was to put these here so they wouldn't be lost.

For reference, the CRON job looks like:
`0 3 * * * /absolute/path/to/script/run.gefs.sipnet.sh`

## Motivation and Context

These scripts are part of the NEFI initiative and aim to iteratively call the PEcAn workflow, including its SDA section.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/assim.sequential/inst/NEFI/generate.gefs.xml.R
```diff
@@ -0,0 +1,135 @@
...
+# but good practice anyway).
+settings$run$site$met.start <- format(Sys.time() - lubridate::days(12), "%Y-%m-%d")
+settings$run$site$met.end <- format(Sys.time() + lubridate::days(16), "%Y-%m-%d")
+
+settings$ensemble$start.year <- format(start_date, "%Y")
+settings$ensemble$end.year <- as.character(end_date, "%Y")
+
+# Determine the workflow ID by looking in the database
+ids <- PEcAn.DB::db.query("SELECT id FROM workflows", con)
+ids <- ids[ids<2000000000] # There are several placeholder ids that are much too large in the database, so this screens
```
mdietze: FYI, it's more efficient to have SQL do this sort of filtering than to query a large table and then post-process it in R

modules/assim.sequential/inst/NEFI/generate.gefs.xml.R
```diff
@@ -0,0 +1,135 @@
...
+                           # those out.  Kind of hack-y, but it works.
+workflowid <- as.character(max(ids) + 1)
+settings$workflow$id <- workflowid
```
mdietze: Don't do this. Create the record in the workflow table, then have that insert return the new ID. This is much much safer than what you're doing here, which is creating the ID of the record before you create the record.

modules/assim.sequential/inst/NEFI/generate.gefs.xml.R
```diff
@@ -0,0 +1,135 @@
...
+
+# Create the output directory.  PEcAn does not do this for you.  It's normally the job of the
+# web interface's php code.
+if (!dir.exists(outdir)) {
+  dir.create(outdir, recursive=TRUE, showWarnings = FALSE)
+}
+
+# Register new workflow id with the database
+hostname <- PEcAn.remote::fqdn()
+query <- paste0("INSERT INTO workflows (site_id, model_id, notes, folder, hostname, start_date,",
```
mdietze: This should have a `RETURNING id` at the end and doesn't need the started_at and created_at

Luke-Dramko: Looks like all three requested changes have to do with the workflow id/database.  I didn't know the query returned with workflow id after insertion; that's certainly much better.  Not a hard change to make!

modules/assim.sequential/inst/NEFI/generate.gefs.xml.R
infotroph: Also, IDs over 2e9 aren't placeholders, they're from [other nodes in the BETY network](https://pecan.gitbook.io/betydb-documentation/betydb-system-administration/distributed-betydb#primary-key-allocations). As written, this won't work for any workflow created outside the Dietze or Serbin labs.

modules/assim.sequential/inst/NEFI/generate.gefs.xml.R
Luke-Dramko: Workflow ID bad practices have been fixed.

PR #2025, opened 2018-08-03
Making write.ensemble.configs compatible with other models

para2x:
After merging the #2017 pull request, it was found that `write.ensemble.configs` is not compatible with any model but SIPNET because of the extra argument sent to `write.config.model` called input. In this pull request, I changed the function in a way that before sending the settings to `write.config.model`, it is modified to have one path like this 
`settings$run$inputs$met$path`.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/utils/R/ensemble.R
```diff
@@ -341,13 +341,16 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
           "outdir      : ", file.path(settings$host$outdir, run.id), "\n",
           file = file.path(settings$rundir, run.id, "README.txt"))
       
+      #changing the structure of input met to what the models are expecting
+      settings$run$inputs$met$path <- samples$met$samples[[i]]
```
mdietze: This needs to be done for every input, not just met

PR #2035, opened 2018-08-08
Sipnet SDA Preprocessing Code

Luke-Dramko:
## Description

This code provides an interface between the main PEcAn workflow and the current version of the SDA code.  It sets up `obs.mean` and `obs.cov`, which are required arguments, and handles the call to the SDA entry point `PEcAn.assim.sequential::sda.enkf`.

The changes made in this pull request do not change the SDA code.  The SDA code is currently too specific towards tree ring predictions, but is undergoing refactoring.

This pull request also includes a dependency to `prep.data.assim` (the function that handles the call to the SDA code), `download.US_WCr`.  `download.US_WCr` is a dependency for `prep.data.assim` (and for the graphs functions in `pecan/modules/assim.sequential/inst/NEFI/`).

Code is written to the level of generalization currently possible with existing standards.

## Motivation and Context

These pieces of code link the main workflow with the SDA code, and provide real flux tower data to use.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/uncertainty/R/prep.data.assim.R
```diff
@@ -0,0 +1,82 @@
...
+  
+  # Creates a proxy row for rbinding
+  sums <- NULL
+  
+  # One vector holds the mean for each variable.
+  obs.mean <- NULL
+  
+  for (i in 1:length(field_data)) {
+    AMF.params <- PEcAn.uncertainty::flux.uncertainty(field_data[[i]], QC = rep(0, length(field_data[[i]])))
+    numvals <- 10;
```
mdietze: numvals should be a function argument

modules/uncertainty/R/prep.data.assim.R
```diff
@@ -0,0 +1,82 @@
...
+    random_mat = NULL
+    new_col = rep(0, length(field_data[[i]]))
+    
+    # Create a new column
+    # i: the particular variable being worked with
+    # j: the column number
+    # k: the row number
+    for (j in 1:numvals) { # number of random numbers
+      obs <- field_data[[i]][!is.na(field_data[[i]])]
+      plot(obs)
```
mdietze: probably don't want to make a plot during a workflow run, and definitely not 5000 times in a loop

modules/uncertainty/R/prep.data.assim.R
```diff
@@ -0,0 +1,82 @@
...
+      res[pos]  <- rexp(length(obs[pos]),
+                        1 / (AMF.params$intercept[[1]] + (AMF.params$slopeP[[1]] * obs[pos])))
+      res[!pos]  <- rexp(length(obs[!pos]),
+                         1 / (AMF.params$intercept[[1]] + (AMF.params$slopeN[[1]] * obs[!pos])))
+      
+      random_multiplier <- sample(c(-1,1), length(res), replace = TRUE)
+      simulated <- obs+(random_multiplier*res)
+      
+      sim_idx = 1; # Because NA's are excluded, real time points don't match up exactly.
+      for (k in 1:length(field_data[[i]])) { # k for each real time point
```
mdietze: not following why this step is needed -- the presence of NAs shouldn't change the length of the vector

Luke-Dramko: That's because there's a `!is.na() on line 36 (`obs <- field_data[[i]][!is.na(field_data[[i]])]`), which removes the NA's from the vector.  @istfer recommended removing `NA`s before applying the `rexp` functions, perhaps because they're not positive nor negative.  Once the `NA`s are removed, the vector's size is smaller than the original array, and it's no longer a guarantee that the i'th element of the real data corresponds to the i'th element of the simulated data.

PR #2039, opened 2018-08-09
NARR download folder

yan130:
## Description
remove site number for regional download raw data

## Motivation and Context
To avoid duplicate downloading


modules/data.atmosphere/R/download.raw.met.module.R
```diff
@@ -1,7 +1,11 @@
 .download.raw.met.module <- function(dir, met, register, machine, start_date, end_date, str_ns,
                                      con, input_met, site.id, lat.in, lon.in, host, site, username, overwrite = FALSE) {
   
-  outfolder <- file.path(dir,paste0(met, "_site_", str_ns))
+  if(register$scale == "regional"){
+    outfolder <- file.path(dir,paste0(met, "_site"))
```
mdietze: This won't work because of the number of regional products that do a site-level extraction as part of the download. Also, if the download really is regional I don't see why you need to include "_site" in the name

yan130: @robkooper 

for 1), in this case, we can change the IF clauses to only match NARR data.
for 2), we can have 
outfolder <- file.path(dir,paste0(met, "_ regional"))

mdietze: I’d suggest using the registry to make this distinction more general, rather than inserting a hack for a specific data set

robkooper: Since it is regional, doesn't any download function download data for a region, not just a site? I think we should capture in the registry file what region this download is related to (in case of NARR it would be continental USA).

mdietze: Because of how we store them in BETY postGIS regions can be 'sites' in PEcAn, and there are regional sites in there already (including a site for the whole planet). That info is already in the Inputs record, and it's already in the registry (e.g. https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/inst/registration/register.NARR.xml). I guess the question is whether we want/need to include that site ID in the folder name as well?

My suggestion for the registry was to add a variable that distinguishes regional products that are downloaded from those that are directly extracted. Not only could that be used here, but it would also allow is to get rid of an annoying hack in met.process where they're a hard-coded list of regional products that are told to skip over the met2CF and extract steps.

robkooper: The other option is to let the site_id in the registration file take precendence over the siteid provided. So in the case of NARR it would still download it to NARR_1135_site. Then once we get to site specific information it would use the actual site. That way we stay consistent.

yan130: sorry, not clear what to do with "Then once we get to site specific information it would use the actual site"

yan130: done

PR #2045, opened 2018-08-15
Refactored SDA

para2x:
This pull request contains refactoring the SDA module or `PEcAn.assim.sequential` package. The core functionality of the code is the same however, it takes advantage of the the new ensemble generation capability implemented in the pecan.

 I tried to keep most of the original code either as a separate functions or in their original format (the original SDA function is still exported as `sda.enkf.original`), whereas some pieces were developed from scratch (such as post analysis visualizations) trying to improve the efficiency of the code. 

There has been also changes to the functions indirectly involved in the SDA process (more extensive changes in read_restart/write_restart of linkages and SIPNET) which will be presented in a separate PR.

sda_modular 2

So, this is also how I setup my workflow for SDA and the xml which I assume is pretty compatible with the main workflow.
```
unlink(c('run','out','SDA'),recursive = T)
rm(list=ls())

#Reading in the settings
settings <- read.settings("pecan.SDA.xml")

#sample from parameters used for both sensitivity analysis and Ens
get.parameter.samples(settings, ens.sample.method = settings$ensemble$samplingspace$parameters$method) 

obs.list <- load_data_paleon_sda(settings)

sda.enkf (settings,
                obs.mean = obs.list$obs.mean,
                obs.cov = obs.list$obs.cov,
                control=list( trace=T,
                                     interactivePlot=F,
                                     TimeseriesPlot=T,
                                     BiasPlot=F,
                                     plot.title="With IC (analysis result) for the forecast",
                                     debug=F)
                                   )
```
and the XML (The only new thing about it is that the adjustment is not an argument anymore ):

```
   <state.data.assimilation>
   <process.variance>FALSE</process.variance>
   <adjustment>TRUE</adjustment>
    .
    .
    .
  <start.date>1960/01/01</start.date>
  <end.date>2000/12/31</end.date>
  </state.data.assimilation>

```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,397 @@
...
+    # do we have obs for this time - what year is it ?
+    obs <- which(!is.na(obs.mean[[t]]))
+    obs.year <- year(names(obs.mean)[t])
+    ###-------------------------------------------------------------------------###
+    ###  Taking care of Forecast. Splitting /  Writting / running / reading back###
+    ###-------------------------------------------------------------------------###-----  
+    #- Check to see if this is the first run or not and what inputs needs to be sent to write.ensemble configs
+    if (t>1){
+      #removing old simulations
+      unlink(list.files(outdir,"*.nc",recursive = T,full.names = T))
```
para2x: This is not the most efficient way possible for forcing the run.model converts it's out to *.nc specially in large ensembles which takes a lot of time to clean. I didn't know of any option forcing `start.model.runs` to behave how we wanted it . Other ideas are appreciated.

modules/assim.sequential/R/Adjustment.R
```diff
@@ -0,0 +1,57 @@
...
+  ## analysis
+  S_a  <- svd(Pa)
+  
+  L_a  <- S_a$d
+  V_a  <- S_a$v
+  
+  ## analysis ensemble 
+  X_a <- X*0
+  for(i in seq_len(nrow(X))){
+    # she decomposed Pa - then it's putting it back together but with a different Z which comes from the likelihood of that ens    X_a[i,] <- V_a %*%diag(sqrt(L_a))%*%Z[i,] + mu.a
```
istfer: is there a reason why this is turned off? or if it's not needed anymore is there a reason why the loop left open?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,393 @@
...
+##' @return Returns whatever the FUN is returning. In case of EnKF and GEF, this function returns a list with estimated mean and cov matrix of forecast state variables as well as mean and cov estimated as a result of assimilation/analysis .
+##' @export
+
+Analysis.sda<-function(settings,
+                       FUN,
+                       Forcast=list(Pf=NULL,mu.f=NULL,Q=NULL,X=NULL),
+                       Observed=list(R=NULL,Y=NULL),
+                       ...
+){
+  if (is.null(FUN)) stop('Analysis function needs to be defined !')
```
istfer: maybe use logger functions?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,393 @@
...
+  FUN(settings, Forcast, Observed,...)
```
istfer: `Forecast` is misspelled here and in other places using this arg

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,393 @@
...
+##' 
+##' @return It returns a list with estimated mean and cov matrix of forecast state variables as well as mean and cov estimated as a result of assimilation/analysis .
+##' @export
+GEF<-function(setting,Forcast,Observed,...){
+  #------------------------------Setup
+  #-- reading the dots and exposing them to the inside of the function
+  dots<-list(...)
+  if (length(dots)>0) lapply(names(dots),function(name){assign(name,dots[[name]], pos=1 )})
+  #General
+  var.names <- unlist(sapply(settings$state.data.assimilation$state.variable,
```
istfer: this could be reduced to `sapply(settings$state.data.assimilation$state.variable, '[[', "variable.name")`

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,393 @@
...
+##' @param Observed A list containing the observed variables including R (cov of observed state variables) and Y (vector of estimated mean of observed state variables)
+##' @param ... Extra argument sent to the analysis function.
+##’ @details
+##’  
+##' 
+##' @description Given the Forecast and Observed this function performs the Generalized Ensemble Kalamn Filter. The generalized ensemble filter follows generally the three steps of sequential state data assimilation. But, in the generalized ensemble filter we add a latent state vector that accounts for added process variance. Furthermore, instead of solving the analysis analytically like the EnKF, we have to estimate the mean analysis vector and covariance matrix with MCMC.
+##' 
+##' @return It returns a list with estimated mean and cov matrix of forecast state variables as well as mean and cov estimated as a result of assimilation/analysis .
+##' @export
+GEF<-function(setting,Forcast,Observed,...){
```
istfer: was this supposed to be `settings`?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,393 @@
...
+  aqq[t + 1, , ]   <- V
+  bqq[t + 1]       <- n
+  return(list(mu.f = mu.f,
+              Pf = Pf,
+              mu.a = mu.a,
+              Pa = Pa,
+              q.bar = q.bar,
+              n = n,
+              X.new=X.new,
+              CIX1=quantile(dat[, iX[2]], c(0.025, 0.5, 0.975)),  #7
```
istfer: I guess here it should be `iX[1]`

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,397 @@
...
+      #-Splitting the input for the models that they don't care about the start and end time of simulations and they run as long as their met file.
+      inputs.split <- list()
+      for(i in seq_len(nens)){
+        if(!no_split){ 
+     #---------------- model specific split inputs
+          inputs.split$samples[i] <-do.call(my.split_inputs, 
+                                      args = list(settings = settings, 
+                                                  start.time = (ymd_hms(obs.times[t-1],truncated = 3) + second(hms("00:00:01"))), 
+                                                  stop.time = obs.times[t],
+                                                  inputs = inputs$samples[[i]]))
```
istfer: where does `inputs` come from now?

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,397 @@
...
+      restart.arg <- NULL
+      new.params <- params
+    }
+  #-------------------------- Writing the config/Running the model and reading the outputs for each ensemble
+    write.ensemble.configs(defaults = settings$pfts, 
+                           ensemble.samples = ensemble.samples, 
+                           settings = settings,
+                           model = settings$model$type, 
+                           write.to.db = settings$database$bety$write,
+                           restart = restart.arg)->outconfig
```
istfer: direction of arrow

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,397 @@
...
+    ### Interactive plotting ------------------------------------------------------   
+    if (t > 1 & control$interactivePlot) { #
+      print(interactive.plotting.sda(settings,t,obs.times,obs.mean,obs.cov,obs,X,FORECAST,ANALYSIS))
+    }
+    ###-------------------------------------------------------------------###
+    ### save outputs                                                      ###
+    ###-------------------------------------------------------------------###---- 
+    save(t, FORECAST, ANALYSIS, enkf.params,new.state,new.params,run.id,ensemble.id,ensemble.samples,inputs, file = file.path(settings$outdir,"SDA", "sda.output.Rdata"))
+    #writing down the image - either you asked for it or nor :)
+    post.alaysis.ggplot(settings,t,obs.times,obs.mean,obs.cov,obs,X,FORECAST,ANALYSIS,plot.title=control$plot.title)
```
istfer: is the fcn name supposed to be `post.analysis.ggplot`?

modules/assim.sequential/R/sda_plotting.R
```diff
@@ -0,0 +1,546 @@
...
+  pink       <- col2rgb("deeppink")
+  alphapink  <- rgb(pink[1], pink[2], pink[3], 180, max = 255)
+  green      <- col2rgb("green")
+  alphagreen <- rgb(green[1], green[2], green[3], 75, max = 255)
+  blue       <- col2rgb("blue")
+  alphablue  <- rgb(blue[1], blue[2], blue[3], 75, max = 255)
+  purple       <- col2rgb("purple")
+  alphapurple <- rgb(purple[1], purple[2], purple[3], 75, max = 255)
+  brown       <- col2rgb("brown")
+  alphabrown <- rgb(brown[1], brown[2], brown[3], 75, max = 255)
```
istfer: this code chunk is repeated so many times here and in other plotting fcns, probably worth having its own fcn

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -0,0 +1,85 @@
...
+    run = function() {
+      if(toggle == 1)
+        nested_sampler_list[[1]]$run()
+    },
+    methods = list(
+      reset = function()
+        nested_sampler_list[[1]]$reset()
+    )
+  )
+}
```
istfer: this fcn doesn't return anything explicitly, was this intended? I mean, wouldn't it return only the last fcn?

modules/assim.sequential/R/sda.enkf_refactored.R
istfer: Why wouldn't `start.model.runs` create new nc files? while SDA progresses iteratively, each new year's output should be converted to nc, what was happening in your case?

para2x: Well I don't really know why, but after converting the outputs of the first run, it was kinda skipping the conversion and always reading those first nc files.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -468,7 +472,7 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
   }else{
     #some stuff about IC file that we can give in lieu of actual ICs
   }
-
+  if(file.exists(file.path(settings$rundir, run.id, "sipnet.param"))) file.rename(file.path(settings$rundir, run.id, "sipnet.param"),file.path(settings$rundir, run.id, paste0("sipnet",Sys.time()%>%as.numeric,".param")))
```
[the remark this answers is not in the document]
para2x: I take this out, It was just for myself trying to double check the results.

modules/assim.sequential/R/Nimble_codes.R
istfer: oh ok I missed the `<<-` operator 👍

modules/assim.sequential/R/sda.enkf_refactored.R
para2x: input is defined in line 232. The first round the met is split just to the bounds of SDA so I read it from the original settings but in the second round on I don't change the original met path and I just produce and use the split ones in each iteration.

modules/assim.sequential/R/sda.enkf_refactored.R
istfer: have you checked whether you still experience this without this line of command? I mean, you were skipping split input before so sipnet was running all years at every iteration, so it wouldn't recreate new nc files. but now that's fixed, maybe this is not an issue anymore?

para2x: Sure I can give it a try.

para2x: This is weird. NO it doesn't convert the new nc files. I don't know why ?!! 
capture

@araiho : any idea ? Did you say, you saw the same behavior from `PEcAn.remote::start.model.runs` too ?

istfer: weird, this plot again looks like a forward run only? is split_inputs.SIPNET not working somehow?

para2x: Yup, this is when I get without deleting nc files. 
So let's say I'm doing SDA for 1960-1965. In the first run I split the met to 1960-1965 so I run the model for all of those years and there is a nc file for each in the out dir. For the second round which I wanna do 1961 the model creates 1961.out but since the nc file is there form the first run (which covered the whole simulation period) 1961.nc does NOT get updated. 
I think this is what's happening here. But if I delete the nc files in each iteration it works all fine.

istfer: then yes, this is an expected behaviour, model2netcdf wouldn't overwrite the existing nc unless `overwrite=TRUE`.. but is there a reason why you're running the whole period in the beginning? if you're doing SDA for 1960-1965, shouldn't the initial forecast be run up until 1960? 

if you have to run the whole 1960-1965 period from the beginning, then you can utilize the overwrite argument of model2netcdf instead of unlinking maybe

para2x: That makes sense. 
I'll see how can I modify it using your comments before using it in the multi site.

araiho: The original sda was getting the initial run from settings$run$model$start.date etc. So, one had to make sure they changed that so it wouldn't run the whole time period in the spin up. But, you could change that so it gets the spin up time period from the settings$sda instead? I can't remember if there was a specific reason we wanted to keep the spin up dates in the run section of the xml.

araiho:
@para2x just wondering. Did you test GEF here for linkages or sipnet?

modules/assim.sequential/R/sda_plotting.R
araiho: Is there a way to put it in pecan.data.viz?

para2x:
I feel like I'm almost done with all the edits @mdietze @istfer @araiho , except for the first one (forcing model2netcdf to overwrite), which I think it's not crucial to have that at the moment. (I couldn't find a workaround at least for now But we can keep that in mind for the next versions. I'll be happy to discuss that in person if you like.)

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,393 @@
...
+  outdir     <- settings$modeloutdir # currently model runs locally, this will change if remote is enabled
+  rundir     <- settings$host$rundir
+  host       <- settings$host
+  forecast.time.step <- settings$state.data.assimilation$forecast.time.step  #idea for later generalizing
+  nens       <- as.numeric(settings$ensemble$size)
+  processvar <- settings$state.data.assimilation$process.variance
+  var.names <- sapply(settings$state.data.assimilation$state.variable, '[[', "variable.name")
+  names(var.names) <- NULL
+  #filtering obs data based on years specifited in setting > state.data.assimilation
+  assimyears <- year(settings$state.data.assimilation$start.date) : year(settings$state.data.assimilation$end.date) # years that assimilations will be done for - obs will be subsetted based on this
```
istfer: would you like to `lubridate::year` these (and the other two lines below)? or extract `year <- lubridate::year` above like the others?

modules/assim.sequential/R/sda_plotting.R
```diff
@@ -0,0 +1,502 @@
...
+  generate_colors_sda()
+  t1         <- 1
+  var.names <- unlist(sapply(settings$state.data.assimilation$state.variable, 
+                             function(x) {
+                               x$variable.name
+                             }, 
+                             USE.NAMES = FALSE), 
+                      use.names = FALSE)
+  #----
+  t1 <- 1
```
istfer: `t1 <- 1` repeats

modules/assim.sequential/R/sda_plotting.R
```diff
@@ -0,0 +1,502 @@
...
+  par(mfrow = c(2, 1))
+  colmax<-2
+  for (i in 1:ncol(FORECAST[[t]])) { #
+    
+    Xbar <- plyr::laply(FORECAST[t1:t], function(x) { mean(x[, i]/rowSums(x[,1:colmax]), na.rm = TRUE) })
+    Xci  <- plyr::laply(FORECAST[t1:t], function(x) { quantile(x[, i]/rowSums(x[,1:colmax]), c(0.025, 0.975), na.rm = TRUE) })
+    
+    Xa <- plyr::laply(ANALYSIS[t1:t], function(x) { mean(x[, i]/rowSums(x[,1:colmax]), na.rm = TRUE) })
+    XaCI <- plyr::laply(ANALYSIS[t1:t], function(x) { quantile(x[, i]/rowSums(x[,1:colmax]), c(0.025, 0.975), na.rm = TRUE) })
+
```
istfer: is there a reason why these are different than the previous version (there was no division bu `rowSums()` before)?

```
        Xbar <- plyr::laply(FORECAST[t1:t], function(x) { mean(x[, i], na.rm = TRUE) })
        Xci  <- plyr::laply(FORECAST[t1:t], function(x) { quantile(x[, i], c(0.025, 0.975), na.rm = TRUE) })
        
        Xa <- plyr::laply(ANALYSIS[t1:t], function(x) { mean(x[, i], na.rm = TRUE) })
        XaCI <- plyr::laply(ANALYSIS[t1:t], function(x) { quantile(x[, i], c(0.025, 0.975), na.rm = TRUE) })
```

para2x: This is interesting, I have no memory of changing this and why I did that .

para2x: I'm gonna go back to what it was before.

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,393 @@
...
+        print(R)
+        cat ("\n --------------Forecast mean ----------- \n")
+        print(enkf.params[[t]]$mu.f)
+        cat ("\n --------------Forecast Cov ----------- \n")
+        print(enkf.params[[t]]$Pf)
+        cat ("\n --------------Analysis mean ----------- \n")
+        print(t(enkf.params[[t]]$mu.a))
+        cat ("\n --------------Analysis Cov ----------- \n")
+        print(enkf.params[[t]]$Pa)
+        cat ("\n ------------------------------------------------------\n")
```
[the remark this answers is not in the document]
para2x: pecan logger functions flatten the matrices and dataframes out. I'll use a mix of `print` and logger for here.

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,415 @@
...
+##' @param Y vector of observations
+##' @param X Dataframe or matrix of forecast state variables for different ensembles.
+##’ @details
+##’  
+##' 
+##' @description This function creates a mtrix mapping obsereved data to their forecast state variable.
+##' 
+##' @return This returns a mtrix specifying which observation go with which state variables.
+##' @export
+Construc_H <- function(choose, Y, X){
```
mdietze: Reason that there's no second t in Construct?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -0,0 +1,415 @@
...
+
+##' @title GEF
+##' @name  GEF
+##' @author Michael Dietze \email{dietze@@bu.edu}, Ann Raiho and Hamze Dokoohaki
+##' 
+##' @param settings  pecan standard settings list.  
+##' @param Forecast A list containing the forecasts variables including Pf (cov of forecast state variables), mu.f (vector of estimated mean of forecast state variables),
+##' Q (process variance) and X (a dataframe of forcats state variables for different ensemble)
+##' @param Observed A list containing the observed variables including R (cov of observed state variables) and Y (vector of estimated mean of observed state variables)
+##' @param ... Extra argument sent to the analysis function. This function requires an extra argument called `extraArg` that needs to be passed through ellipsis.
```
mdietze: As noted previously, extraArg should  be an explicit argument here, not passed via ellipsis.

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -0,0 +1,85 @@
+#' @import nimble
+load_nimble <- function(){
```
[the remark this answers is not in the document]
para2x: @araiho

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -0,0 +1,85 @@
...
+    X.mod[1:N] ~ dmnorm(muf[1:N], prec = pf[1:N,1:N]) ## Model Forecast ##muf and pf are assigned from ensembles
+    
+    ## add process error
+    X[1:N]  ~ dmnorm(X.mod[1:N], prec = q[1:N,1:N])
+    
+    #observation operator
+    y_star[1:YN] <- y_star_create(X[1:YN])
+    
+    ## Analysis
+    y.censored[1:YN] ~ dmnorm(y_star[1:YN], prec = r[1:YN,1:YN]) #is it an okay assumpution to just have X and Y in the same order?
```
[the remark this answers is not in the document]
para2x: @araiho

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,393 @@
...
+  ###-------------------------------------------------------------------###
+  ### set up for data assimilation                                      ###
+  ###-------------------------------------------------------------------###-----  
+  nt          <- length(obs.times)
+  if (nt==0) stop('There has to at least one Obs')
+  FORECAST    <- ANALYSIS <- list()
+  enkf.params <- list()
+  aqq         <- NULL
+  bqq         <- numeric(nt + 1)
+  CI.X1       <- matrix(0, 3, nt) # it was taken care of
```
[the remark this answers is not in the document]
para2x: @araiho

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,393 @@
...
+    )
+    params<-new.params
+  }else{
+    t<-0
+  }
+  
+  ###------------------------------------------------------------------------------------------------###
+  ### loop over time                                                                                 ###
+  ###------------------------------------------------------------------------------------------------###---- 
+  while(t<nt){
```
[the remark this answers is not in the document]
para2x: I believe one of the best ways of feeding observed data to the SDA code is through directories. We feed the code with a path to a directory (through pecan xml) which is structured like our lists. This allows us to write into the directory for high frequency data in one hand and have the SDA code to scan the directory on the other hand. In short, the advantage of `while` here over `for` is that in a scenario like above we can extend our loop size whenever there is a new obs in the folder.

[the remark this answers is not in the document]
para2x: This mindset also makes the SDA code more compatible with the rest of the main workflow because it won't need user specific arguments/inputs anymore.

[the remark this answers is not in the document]
mdietze: I'm not sure I follow or agree with this logic. Nowhere else in the code do we have code 'find' new data in a directory, the data is always specified in the settings file and registered in the database. I also don't really see a use case for the arrival of new data causing an update of nt while the code is actively running. Used in an iterative forecast mode the Analysis step is only performed at the very first step (t=1) and all subsequent steps are in the future and thus not constrained by observations. When used in iterative forecast mode what the arrival of new data does trigger is the initiation of a NEW call to sda.enkf in the restart mode that takes the most recent forecast as the IC and clones it to a new workflow directory.

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -0,0 +1,393 @@
...
+#' @name  sda.enkf
+#' @author Michael Dietze and Ann Raiho \email{dietze@@bu.edu}
+#' 
+#' @param settings    PEcAn settings object
+#' @param obs.mean    list of observations of the means of state variable (time X nstate)
+#' @param obs.cov     list of observations of covariance matrices of state variables (time X nstate X nstate)
+#' @param Q           process covariance matrix given if there is no data to estimate it
+#' @param restart     Used for iterative updating previous forecasts. When the restart is TRUE it read the obejct in SDA folder writen from previous SDA.
+#' @param control    List of flags controling the behaviour of the SDA. trace for reporting back the SDA outcomes, interactivePlot for ploting the outcomes after each step, 
+#' TimeseriesPlot for post analysis examination, BiasPlot for plotting ..., plot.title is the title of post analysis plots and debug mode allows for pausing the code and examinign the variables inside the function.
```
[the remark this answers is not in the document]
para2x: @araiho

modules/assim.sequential/R/Nimble_codes.R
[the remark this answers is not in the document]
araiho: @para2x @mdietze I did not make this function so I'm not exactly sure what you want. Do you want me to point you to documentation for all of the nimble functions contained? The tobit stuff should be in the sda methods. and the y_star_create we haven't quite finished thinking through but is for doing agb versus fcomp. I think what @mdietze wants for this documentation is just what the load_nimble function does. Is that correct? In that case all you need to say is that it loads the nimbe functions that are used in the GEF.

modules/assim.sequential/R/Nimble_codes.R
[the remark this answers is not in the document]
araiho: Sorry old comment in there. The order should be taken care of earlier in the code.

modules/assim.sequential/R/Nimble_codes.R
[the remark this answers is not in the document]
para2x: I do have that down now, I didn't know if we need to say what these function would provide to GEF. But if @mdietze thinks it's enough to mention that this function just loads the required functions, it's all good then.

para2x:
@mdietze,  I think I have most of the comments down.

modules/assim.sequential/R/Nimble_codes.R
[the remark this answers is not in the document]
mdietze: If the comment is old and no longer applies it should be removed

That said, I'm not sure I saw where H was actually being used earlier, and it's not explicitly in this code. I'm not going to hold up this PR for the integration of H, but it should be on the to-do list

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -96,7 +97,7 @@ EnKF<-function(setting, Forecast, Observed, H, extraArg=NULL, ...){
 ##' @author Michael Dietze \email{dietze@@bu.edu}, Ann Raiho and Hamze Dokoohaki
 ##' 
 ##' @param settings  pecan standard settings list.  
-##' @param Forecast A list containing the forecasts variables including Q (process variance) and X (a dataframe of forcats state variables for different ensemble)
+##' @param Forecast A list containing the forecasts variables including Q (process variance) and X (a dataframe of forecats state variables for different ensemble)
```
mdietze: still a spelling error

para2x:
@mdietze outdated ones were taken care of before and typos were just resolved.

modules/assim.sequential/R/Nimble_codes.R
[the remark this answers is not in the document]
mdietze: This comment is still here. It has not yet been removed or revised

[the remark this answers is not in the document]
para2x: I just removed the comment.

Issue #1976, opened 2018-06-16
PEcAnRTM compilation not parallel-safe


infotroph:
**Describe the bug**
Building PEcAn with parallel jobs (e.g. `make -j2`) fails for me because of the following error compiling RTM -- I haven't looked in detail, but my best guess is that some object files have dependencies on `*.mod` were being satisfied by the ordering of the file rather than declared explicitly.

**To Reproduce**

Successful build with only one core:
```
0> make clean
rm -rf .install .check .test .doc
find modules/rtm/src \( -name \*.mod -o -name \*.o -o -name \*.so \) -delete
0> make .install/modules/rtm

[...snip...]

Rscript -e "devtools::document('"modules/rtm"')"
Updating PEcAnRTM documentation
Loading PEcAnRTM
Re-compiling PEcAnRTM
'/Library/Frameworks/R.framework/Resources/bin/R' --no-site-file --no-environ  \
  --no-save --no-restore --quiet CMD INSTALL  \
  '/Users/chrisb/github_forks/pecan/modules/rtm'  \
  --library='/var/folders/qr/mbw8xxpd45jdv_46b27914280000gn/T//RtmpwJ2XYS/devtools_install_ae68660420f1'  \
  --no-R --no-data --no-help --no-demo --no-inst --no-docs --no-exec  \
  --no-multiarch --no-test-load --preclean 

* installing *source* package ‘PEcAnRTM’ ...
** libs
gfortran  -fPIC -Wall -g -O2  -c  RTM/modules/mod_types.f90 -o RTM/modules/mod_types.o

[...snip...]

* DONE (PEcAnRTM)
echo `date` > .doc/modules/rtm
time Rscript -e "devtools::install('modules/rtm', Ncpus = 1);"
Installing PEcAnRTM
'/Library/Frameworks/R.framework/Resources/bin/R' --no-site-file --no-environ  \
  --no-save --no-restore --quiet CMD INSTALL  \
  '/Users/chrisb/github_forks/pecan/modules/rtm'  \
  --library='/Library/Frameworks/R.framework/Versions/3.4/Resources/library'  \
  --install-tests 

* installing *source* package ‘PEcAnRTM’ ...
** libs
make[1]: Nothing to be done for `all'.
installing to /Library/Frameworks/R.framework/Versions/3.4/Resources/library/PEcAnRTM/libs
** R
** data
** inst
** tests
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
[...]

PR #2059, opened 2018-08-21
Build in parallel where possible

infotroph:
## Description

* When building on Travis, all Make invocations now get `-j2` unless otherwise specified.
* Despite my efforts here, package installation does *not* seem to be fully parallel-safe. Instead, Travis build sets NCPUS=2 to at least allow the dependencies or each package to download/install in parallel.
* Restructured RTM build; now specifies module prerequisites explicitly (#1976).
* Fixed target-ordering ambiguity in book compilation
* When trying to load a package for the purpose of installing it if missing, avoid complaining about missingness.
* Added '+' modifiers to Makefile lines that invoke recursive Make calls, including the ones hidden inside the package compilation step of `install.packages`. **Caution: opaque magic!** I don't understand why this works, but it seems necessary because without it, parallel installations of any dependency needing compilation were often failing deep inside the internal Make process spawned by `install.packages`, with the message `make[1]: *** read jobs pipe: No such file or directory.  Stop.`

### Overall goal

Speed up builds by using all available cores as much as possible, especially on Travis (where the container has two cores but we've been mostly single-threading).

### More notes about '+' lines in Make recipes

* By "+ lines" I mean any line of a Make recipe that has a plus sign as its first non-whitespace character, which has special behavior but apparently no official name.
* The special behavior is that when a recipe is called, plus lines are always executed even if Make is in dry-run (-n), touch (-t), or question (-q) mode. This might be desirable when, say, that line calls a recursive make process which will inherit and correctly handle the dry-run/question/touch flags.
* But by my reading of the Make documentation, adding `+` to a recipe line is only *stated* to change the behavior of the -n, -t, and -q flags, none of which are visibly in use in the places I'm adding plus lines!
* More informally, it seems like `+` could be glossed as "Execute this line no matter what; it invokes a recursive make that will handle any conditional flags." By that gloss, it's reasonable -- but
maybe still not fully expected -- that + lines would make subprocesses better-behaved in general. But this is hand-waving and the documentation doesn't actually say this.
* An undesirable side effect: `make -n install`, which ought to be a dry run, will still not *install* any dependencies but will now *download* all dependency packages before skipping their compilation.
* Even after all this, some packages were still throwing errors when run in parallel, so installation on Travis uses `make -j1` for installation and `make -j2` for documentation/testing/checking. 
* Lots more details in my [over-long debugging notes](https://gist.github.com/infotroph/bf522f5959a0ff709377919131bf23d8).

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


book_source/Makefile
```diff
@@ -13,4 +11,5 @@ deploy:
 	./deploy.sh
 	
 pdf:
+	./check_bkd_pkg.R
```
robkooper: why not add build to pdf:?

infotroph: as in `pdf: build`? I assume because the render_book calls are doing different things such that it's not necessarily desirable to wait for both when calling one.

But now that you mention it I should have done `build: bkdcheck` and `pdf: bkdcheck` instead of removing the rule.. Will change.

infotroph:
Merge fixed, plus one more change while I was doing it: Since our default make rule is `all: install document` and `devtools::check` also rebuilds documentation before checking, `make document` was twice redundant. 

@robkooper @mdietze Assuming the build passes and you agree with my logic on `make doc`, this is ready to merge.

PR #2063, opened 2018-08-23
Fix the defining of args for db.file.insert in convert.inputs

tonygardella:
## Description
Was getting an error:
```
Error in basename(in.prefix) : a character vector argument expected
```
And found that in convert.inputs, when trying to insert the result into the database, it wasn't handling the result object correctly to define `in.prefix`. The id `existing.input` arg wasn't being defined properly as well
## Motivation and Context
This comes from testing the VM

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/utils/R/convert.input.R
+++ b/base/utils/R/convert.input.R
@@ -678,8 +678,8 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
       
       if (insert.new.file && id_not_added) {
         dbfile.id <- PEcAn.DB::dbfile.insert(in.path = dirname(result[[1]]$file[1]),
-                                             in.prefix = result$dbfile.name[1], 
-                                             'Input', existing.input$id, 
+                                             in.prefix = result[[1]]$dbfile.name[1], 
+                                             'Input', existing.input[[1]]$id, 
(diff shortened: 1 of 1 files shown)


robkooper:
Thing that scares me a little with this pull request is that we use this function in 21 locations and this is the first time we see this issue.

What was the cause of this issue to be discovered. Could this be that another piece of code returns a wrong argument? I don't want to fix this in one place and then have 20 other places that are broken.

tonygardella:
@robkooper @mdietze existing.input([Line 626](https://github.com/PecanProject/pecan/blob/develop/base/utils/R/convert.input.R#L626)) and result([Line 608](https://github.com/PecanProject/pecan/blob/develop/base/utils/R/convert.input.R#L608)) are explicitly turned into lists after the data is downloaded. I believe Luke did that to make it easier to handle ensembles. @mdietze Yeah I forgot to change to `i` as the result is looped over. Other scenarios dealing with these lists of data.frames are working and for some reason this was the only part of the code after those objects get turned into lists that didn't treat them as lists. I think Luke just forgot this chunk or didn't test this scenario.

robkooper:
looked at the rest of the code around the patch and this now matches better with the code.

Issue #2064, opened 2018-08-24
Possible bug in settings$run$inputs$met$path for sensitivity analysis runs


serbinsh:
**Describe the bug**
When creating the write.configs for SA runs, settings$run$inputs$met$path returns a list of 1 instead of a character string, for example: 

```
> settings$run$inputs$met$path
$path1
[1] "/data/Model_Output/pecan.output/dbfiles/AmerifluxLBL_MAAT_site_1-146/AMF_US-Dix_BASE_HH_2-1.2005-01-01.2006-12-31.csv"

> str(settings$run$inputs$met$path)
List of 1
 $ path1: chr "/data/Model_Output/pecan.output/dbfiles/AmerifluxLBL_MAAT_site_1-146/AMF_US-Dix_BASE_HH_2-1.2005-01-01.2006-12-31.csv"
```

This instead should be a simple character string of the full path

serbinsh:
This can cause a crash in PEcAN workflows that need to parse this path, for example

```
2018-08-24 15:24:51 INFO   [write.config.MAAT] :
   *** Convert input trait values to MAAT parameters and units ***
2018-08-24 15:24:51 INFO   [write.config.MAAT] :
   Running with met drivers:
   /data/Model_Output/pecan.output/dbfiles/AmerifluxLBL_MAAT_site_1-146/AMF_US-Dix_BASE_HH_2-1.2005-01-01.2006-12-31.csv
List of 1
 $ path1: chr "/data/Model_Output/pecan.output/dbfiles/AmerifluxLBL_MAAT_site_1-146/AMF_US-Dix_BASE_HH_2-1.2005-01-01.2006-12-31.csv"
2018-08-24 15:24:51 INFO   [write.config.MAAT] :
Error in dirname(settings$run$inputs$met$path) :
  a character vector argument expected
Calls: <Anonymous> ... <Anonymous> -> do.call -> write.config.MAAT -> dirname
```


but when calling WC from ensemble it works fine, so this ONLY happens for SA runs

```
2018-08-24 15:26:50 INFO   [write.config.MAAT] :
   *** Model mimic not selected ***
2018-08-24 15:26:50 INFO   [write.config.MAAT] :
   *** Convert input trait values to MAAT parameters and units ***
2018-08-24 15:26:50 INFO   [write.config.MAAT] :
   Running with met drivers:
[...]

PR #2067, opened 2018-08-27
Fix the met for SA

para2x:
This is a fix for issue #2064. This makes sure that one met path is sent to write.sa.config.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/utils/R/sensitivity.R
+++ b/base/utils/R/sensitivity.R
@@ -301,7 +301,7 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
             # This checks for met.id tag in the settings under sensitivity analysis - if it's not there it creates it. Then it's gonna use what it created.
             if (is.null(settings$sensitivity.analysis$met.id))  settings$sensitivity.analysis$met.id <- 1
              
-              settings$run$inputs$met$path <- settings$run$inputs$met$path[settings$sensitivity.analysis$met.id]
+              settings$run$inputs$met$path <- settings$run$inputs$met$path[[settings$sensitivity.analysis$met.id]]
             
           }
           
--- a/base/utils/R/sensitivity.R
+++ b/base/utils/R/sensitivity.R
@@ -298,8 +298,9 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
           
           # I check to make sure the path under the met is a list. if it's specified what met needs to be used in 'met.id' under sensitivity analysis of pecan xml we used that otherwise, I use the first met.
           if (is.list(settings$run$inputs$met$path)){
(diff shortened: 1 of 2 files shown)


base/utils/R/sensitivity.R
```diff
@@ -296,6 +296,15 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
               file = file.path(settings$rundir, run.id, "README.txt"), 
               sep = "")
           
+          # I check to make sure the path under the met is a list. if it's specified what met needs to be used in 'met.id' under sensitivity analysis of pecan xml we used that otherwise, I use the first met.
+          if (is.list(settings$run$inputs$met$path)){
+            if (is.null(settings$sensitivity.analysis$met.id))  settings$sensitivity.analysis$met.id <- 1
+            settings$run$inputs$met$path <- settings$run$inputs$met$path [settings$sensitivity.analysis$met.id]
```
mdietze: Small thing, but would be good if you made the `else` explicit here and if you didn't have a space in `path [`

para2x: @mdietze I corrected the `path [`.
In the logic of the code I didn't think there is a need for an `else` statement (if it's `else` then it's in my desired format/shape).  
Let me know if you think this definitely needs to have an `else`.

serbinsh:
@para2x what is the output supposed to be following this update?  A single character string or a list of 1?

base/utils/R/sensitivity.R
```diff
@@ -296,6 +296,17 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
               file = file.path(settings$rundir, run.id, "README.txt"), 
               sep = "")
           
+          # I check to make sure the path under the met is a list. if it's specified what met needs to be used in 'met.id' under sensitivity analysis of pecan xml we used that otherwise, I use the first met.
+          if (is.list(settings$run$inputs$met$path)){
+            # This checks for met.id tag in the settings under sensitivity analysis - if it's not there it creates it. Then it's gonna use what it created.
+            if (is.null(settings$sensitivity.analysis$met.id))  settings$sensitivity.analysis$met.id <- 1
+             
+              settings$run$inputs$met$path <- settings$run$inputs$met$path[settings$sensitivity.analysis$met.id]
```
serbinsh: I guess then makes sure the result is just one string with the final path, so if if there is say a path$path1 this will return path1 as path as a single string so I think this should be good, if its a string

para2x: It's gonna be a string for sure, if there is a bunch of paths under met$path like `<path1>, <path2> ` then either it uses the `met.id` tag in the settings or it just takes the first one and puts one path string in the met$path.

serbinsh:
@mdietze this was merged into release but not develop?

PR #2077, opened 2018-08-29
Fixed error when format or input id is NULL but other is not.

ankurdesai:
Addresses an issue in db query format vars when input or format is NULL, related to #2072 but doesn't entirely solve the other issues in convert inputs.


mdietze:
I think @infotroph request for this second check is reasonable -- @ankurdesai do you have the bandwidth to add that to this PR or should we open a new PR?

I'd also add that in both cases (no format ID or invalid format ID) I wonder it should be a 'severe' stopping error rather than just an error message? For the main workflow I think that makes sense, but the thing that gives me pause is the data ingest app by @LiamBurke24, where you might be querying to see if a format exists already and then be creating one if it doesn't.

That said, all of this begs the question as to why invalid and NULL formats are suddenly being passed into the workflow when this never was a problem before.

infotroph:
I agree it's concerning, but don't take my example as strong evidence of a workflow problem -- I found it while running an outdated demo against an unsynced local copy of Bety, so the invalid ID isn't in itself surprising.

ankurdesai:
I can try to look but it's not happening today most likely with start of semester looming. It does seem to be a larger issue that somewhere NULL ids are being sent out from a bad or failed query (and seems to be somewhere in met process/convert inputs - or possibly an issue with BETY tables errors). I think the function was designed to find the format from input. But @infotroph does suggest an interesting case that should be trapped. I really wish R error messages told you line numbers...

serbinsh:
good to merge then since @infotroph created a separate issue? @mdietze, @ankurdesai ?

PR #2089, opened 2018-08-31
Initial base code to create batch Runs

tonygardella:
This script contains a  function to create and execute a batch of basic runs by creating a pecan.xml from scratch, and then executing workflow.R.
Function( takes in a vector of run specifications:
Models -  finds models on user defined machine
met - User defined type. Right now its CRUNCEP and Ameriflux.
Site id  provided by user
Dates provided by user
)

expand.grid creates combination table of run specifications so that if multiple models, sites, dates, etc. exist, all combinations are made into each row.  Then the function is applied to those rows of  run specifications.

## Currently  In progress/Debugging (Coming in the next few days)
- Status and error report. Read STATUS file and direct console output to a logfile. Grep error statement and append this info next to it's row in the run table. Export Run table to a separate file.
- Optional Inputs from models and their requisite settings

## Next Steps
- Optional Control of sensitivity and ensemble run settings
- More flexibility on run specifications (Add if statements on settings that could be defined by user that hard coded right now)
- Cleaning test runs. 

@serbinsh Added you because you're my reviewer this week. Could be useful for you as well so I'd like your thoughts on what you'd like me to add.
@mdietze Need your thoughts on whether or not this is ok as a base strategy to do things.
@bcow Took some of your ideas from the clone runs pseudo code, and would like your input as well

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/workflow/inst/batch_runs.R
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+
+##Create Run Args
+pecan_path <- "~/pecan"
+config.list <- PEcAn.utils::read_web_config(paste0(pecan_path,"/web/config.php"))
+bety <- betyConnect(paste0(pecan_path,"/web/config.php"))
+con <- bety$con
+machineid <- 99000000001 #ID of the VM
```
serbinsh: will this be a non hard-coded option?

tonygardella: Yep

base/workflow/inst/batch_runs.R
```diff
@@ -0,0 +1,120 @@
...
+pecan_path <- "/fs/data3/tonygard/work/pecan"
+config.list <- PEcAn.utils::read_web_config(paste0(pecan_path,"/web/config.php"))
+bety <- betyConnect(paste0(pecan_path,"/web/config.php"))
+con <- bety$con
+
+## Find name of Machine R is running on
+mach_name <- Sys.info()[[4]]
+mach_id <- tbl(bety, "machines")%>% filter(grepl(mach_name,hostname)) %>% pull(id)
+  
+model_ids <- tbl(bety, "dbfiles") %>% filter(machine_id == mach_id) %>%
```
serbinsh: @tonygardella should we assume all models will work or that we may have models "registered" with a machine but that it may not work at the moment and instead we provide a list of specific models to test OR automatically generate based on BETY

serbinsh:
Would a similar approach work for doing runs where we compare the same model at the same site with the same PFT but different met drivers, or same except PFTs differ?  or do we have such functions already in benchmarking @bcow @mdietze ?

base/workflow/inst/batch_runs.R
tonygardella: It's not the assumption that all models work. Part of the reason for making this is to easily identify what's not working. that being said I could make this part into an if statement where you could provide a vector of model ids and if you don't then it will run all models on your machine.

I'm beginning to see this this could be used for not just testing, but an easy way to do a batch of experiments. I'll take that into account as I continue defining args and adding options.

mdietze:
@ashiklom since you recently pulled this branch and got it to run, could you merge your changes into @tonygardella and review what else needs to happen so that we can get this PR in

base/workflow/R/create_xml.R
```diff
@@ -0,0 +1,113 @@
+##' @export
+##' @aliases create_xml
+##' @name create_xml
+##' @title create_xml
+##' @description Function to create a viable PEcAn xml from a table containing specifications of a run
+##' @param run_list PEcAn table containing specifications of runs
```
mdietze: The format for this table needs to be documented  better

base/workflow/R/create_xml.R
```diff
@@ -0,0 +1,113 @@
...
+##' @param overwrite.met,overwrite.fia,overwrite.ic logical
```
mdietze: These don't seem to actually be used as params

base/workflow/inst/batch_runs.R
```diff
@@ -0,0 +1,431 @@
...
+#' @param ensemble_size
+#' @param ensemble_variable
+#' @param sensitivity
+#' @param db_bety_username
+#' @param db_bety_password
+#' @param db_bety_hostname
+#' @param db_bety_driver
+#' @return 
+#' @author Alexey Shiklomanov
+create_execute_test_xml <- function(model_id,
```
mdietze: Isn't there a function of the same name inside the package that's part of this same PR? Can we decide which we like better (or merge the two) so that there's only one version.

base/workflow/inst/batch_runs.R
```diff
@@ -0,0 +1,431 @@
...
+##Create Run Args
+met_name <- c("CRUNCEP", "AmerifluxLBL")
+startdate <- "2004-01-01"
+enddate <- "2004-12-31"
+out.var <- "NPP"
+ensemble <- FALSE
+ens_size <- 1 # Run ensemble analysis for some models?
+sensitivity <- FALSE
+user_id <- 99000000002   # TODO: Un-hard-code this
+dbfiles_folder <- normalizePath("~/output/dbfiles") # TODO: Un-hard-code this
```
mdietze: Generally it's better to have library loads and run args at the top of a script, not this far down. The function def'n above this really would be better in it's own file (which is consistent with previous comment about resolving duplication)

base/workflow/inst/batch_runs.R
```diff
@@ -0,0 +1,431 @@
...
+print(site_id_noinput)
+
+site_id <- site_id_noinput$site_id
+site_name <- gsub(" ", "_", site_id_noinput$sitename)
+
+
+
+#Create permutations of arg combinations
+options(scipen = 999)
+run_table <- expand.grid(
```
mdietze: I don't really know how expand.grid works, but it's unclear to me from the syntax below which things are being expanded in all possible permutations and which are held fixed (and if they're fixed, how you designate which attribute they line up with). So for global models we want to permute over models x sites, but not all models make sense to run at all sites (LINKAGES only for east temperate, crop models only on crop sites, etc). Similarly, we probably don't want to permute all sites to run all possible met combinations, since AmerifluxLBL won't exist for arbitrary sites on the globe and as we expand the list of met products we want to test we really just need a handful of runs for the gridded products to verify they are working. And the start/end dates are specific to a site x met combination -- how does this function know that (and, for example, that they're not model attributes)? Oh, and PFTs are also a model x site specific setting, where do those come in?  And where do you specify other inputs (IC, soils, etc)? All in all, to me it just seems simpler to have a table of what runs you want to do at what sites with what inputs than to try to write code that creates all possible permutations

base/workflow/inst/batch_runs.R
```diff
@@ -0,0 +1,431 @@
...
+result_table <- as_tibble(run_table) %>%
+  mutate(
+    outdir = NA_character_,
+    workflow_complete = NA,
+    has_jobsh = NA,
+    model_output_raw = NA,
+    model_output_processed = NA
+  )
+
+for (i in seq_len(nrow(run_table))) {
```
mdietze: probably fine for a first pass, but in general I don't think we should be limited to doing these runs sequentially

ashiklom: Moved to new issue #2399

base/workflow/inst/batch_runs.R
ashiklom: `expand.grid` generates a full permutation of everything in it. So this is a full factorial combination of all arguments. However, some of them are scalar, so they are fixed.

This structure was in the original code, and I took as my responsibility to just get it to run and produce some kind of output. I can revise the structure of this script to read in a table of runs, but the first draft of that table will be pretty limited because I don't have time to figure out which combination of sites, PFTs, models, and other settings makes sense.

mdietze: That's fair. I'm also fine with pulling this PR with expand.grid still there and without deeper documentation under the promise that those things are coming soon. Happy to help figuring out what model, site, met, date combinations make sense (e.g. in a Google Sheet)

ashiklom:
@mdietze OK, I think I addressed all of your comments.

- Documentation is [here](https://github.com/PecanProject/pecan/pull/2089/files#diff-fa1450edfafe47cf01585f28feb069c3R363).
- `create_xml_...` function is in its [own file](https://github.com/PecanProject/pecan/pull/2089/files#diff-90e2d8e8738c4945682abe023c463dcfR1), with Roxygen documentation, etc.
- Script [`batch_run.R`](https://github.com/PecanProject/pecan/pull/2089/files#diff-624cd7951970689a283652c2d620afb1R1) takes an input table and runs all of the tests.
- Experimental script [`permutation_tests.R`](https://github.com/PecanProject/pecan/pull/2089/files#diff-2d265a20f089fbd96752a1c628e76416R1) generates a CSV table that can be used as input to `batch_run`. It's basically a lightly tweaked version of the old run script.

Obviously, there's a lot more to be done in the way of automated integration testing, but I think this is a decent start. Unless there are glaring, fundamental flaws (or minor fixes), I think we should get this in so folks can start putting it to the test on their machines. That will also be a good way of identifying edge cases that the current (limited) tests don't cover.

Issue #2047, opened 2018-08-17
SIPNET errors not being caught


serbinsh:
**Describe the bug**
It *looks* like SIPNET runs will continue through then workflow without throwing an error even if no sipnet.clim file is created/available and all model runs fail without producing output.  

```
prerun
Error reading from climate file: first location must be loc. 0
mv: cannot stat `/data/Model_Output/pecan.output/PEcAn_2000000989/run/2000075060/sipnet.out': No such file or directory
```

```
> # Do conversions
> settings <- PEcAn.utils::do_conversions(settings)
2018-08-17 11:00:43 DEBUG  [PEcAn.utils::do_conversions] : 
   do.conversion outdir /data/Model_Output/pecan.output/dbfiles 
2018-08-17 11:00:43 INFO   [PEcAn.utils::do_conversions] : PROCESSING:  met 
2018-08-17 11:00:43 INFO   [PEcAn.utils::do_conversions] : 
   calling met.process: 
2018-08-17 11:00:44 INFO   [browndog.met] : 
   browndog download url : 
   http://dap.ncsa.illinois.edu:8184/file/20278776_pecan.clim 
2018-08-17 11:00:56 WARN   [PEcAn.DB::dbfile.input.insert] : 
   Multiple input files found matching parameters format_id = 24, startdate 
   = 2004/01/01, enddate = 2004/12/31, parent = .  Selecting the last input 
   file.  This is normal for when an entire ensemble is inserted 
   iteratively, but is likely an error otherwise. 
2018-08-17 11:00:56 WARN   [PEcAn.DB::db.close] : 
   Connection created outside of PEcAn.DB package 
2018-08-17 11:00:56 DEBUG  [PEcAn.utils::do_conversions] : 
   updated met path: 
   /data/Model_Output/pecan.output/dbfiles/BD-b99e48b494b5/sipnet.clim 
> 
> # Query the trait database for data and priors
> if (PEcAn.utils::status.check("TRAIT") == 0){
+   PEcAn.utils::status.start("TRAIT")
+   settings <- PEcAn.workflow::runModule.get.trait.data(settings)
+   PEcAn.settings::write.settings(settings, outputfile='pecan.TRAIT.xml')
+   PEcAn.utils::status.end()
+ } else if (file.exists(file.path(settings$outdir, 'pecan.TRAIT.xml'))) {
+   settings <- PEcAn.settings::read.settings(file.path(settings$outdir, 'pecan.TRAIT.xml'))
+ }
[...]

PR #2096, opened 2018-09-06
Add check for file size and exit if empty or NA in convert.input

tonygardella:
## Description
Takes the result list of dataframes that contains the file names of downloaded/converted files from convert.input and checks for whether or not the files or empty or a file was downloaded or not. If there is a file of size 0 or it's empty, logger.severe is used to exit out of pecan.

## Motivation and Context
There was a problem that files could be downloaded/converted that were actually empty, or within the list of files that were supposed to be downloaded/converted some were successfully done and others were not, but the since some worked things would continue on.

@bcow You're my code reviewer this week.
@serbinsh This should help with making runs fail at the correct step when met is empty
@ashiklom @infotroph  Want some advice on creating a better log statement. I have a data.frame of result files and when it fails I want to print out the data.frame, but current logger seems to just flatten everything into a string. I know you guys have thought of these things and would like advice on how to improve the logger message I'm returning. 

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/utils/R/convert.input.R
+++ b/base/utils/R/convert.input.R
@@ -631,7 +631,7 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
       )
     })
   
-  if (any(result_sizes[[i]]$missing) || any(result_sizes[[i]]$empty)){
+  if (any(result_sizes$missing) || any(result_sizes$empty)){
     log_format_df = function(df){
       df  %>%
         format() %>%
--- a/base/utils/R/convert.input.R
+++ b/base/utils/R/convert.input.R
@@ -620,7 +620,7 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
   #--------------------------------------------------------------------------------------------------#
   # Check if result has empty or missing files
   
-  result_sizes <- result %>% purrr::map(~ .x$file) %>% purrr::map(~ .x %>% file.size) %>% as.data.frame() %>% 
+  result_sizes <- result %>% purrr::map(~ .x$file) %>% purrr::tmap(~ .x %>% file.size) %>% as.data.frame() %>% 
     `colnames<-`(c("file_size")) %>%
     mutate(missing = ifelse(is.na(file_size),NA,TRUE)) %>%
     mutate( empty = ifelse(file_size==0,"EMPTY",TRUE))
--- a/base/utils/R/convert.input.R
+++ b/base/utils/R/convert.input.R
@@ -618,6 +618,18 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
   }
   
   #--------------------------------------------------------------------------------------------------#
+  # Check if result has empty or missing files
+  
+  result_sizes <- result %>% purrr::map(~ .x$file) %>% map(~ .x %>% file.size) %>% as.data.frame() %>% 
+    `colnames<-`(c("file_size")) %>%
+    mutate(missing = ifelse(is.na(file_size),NA,TRUE)) %>%
+    mutate( empty = ifelse(file_size==0,"EMPTY",TRUE))
+  
+  if( NA %in% result_sizes$missing | "EMTPY" %in% result_sizes$empty){
+    
+    PEcAn.logger::logger.severe(msg = paste0("Files Processed with function: ", fcn," have resulted in empty or missing files."))
+  }
+  
   # Insert into Database
   outlist <- unlist(strsplit(outname, "_"))
   
--- a/base/utils/R/convert.input.R
+++ b/base/utils/R/convert.input.R
@@ -620,7 +620,7 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
   #--------------------------------------------------------------------------------------------------#
   # Check if result has empty or missing files
   
-  result_sizes <- result %>% purrr::map(~ .x$file) %>% map(~ .x %>% file.size) %>% as.data.frame() %>% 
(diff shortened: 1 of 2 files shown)


infotroph:
> Want some advice on creating a better log statement. I have a data.frame of result files and when it fails I want to print out the data.frame, but current logger seems to just flatten everything into a string. I know you guys have thought of these things and would like advice on how to improve the logger message I'm returning.

@tonygardella Can you give an example of an input that triggers the flattened message? I still don't have a great handle on setting up convert.input calls, so it would be slow for me to come up with one.

serbinsh:
This looks relevant for issue #2047

tonygardella:
@infotroph Sending you the call to convert.input I'm testing on slack. Probably will not work depending on what files and records are associated with your machine.

tonygardella:
@infotroph I'll create a result object that's the same structure as expected and send that to you too.

base/utils/R/convert.input.R
```diff
@@ -618,6 +618,18 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
   }
   
   #--------------------------------------------------------------------------------------------------#
+  # Check if result has empty or missing files
+  
+  result_sizes <- result %>% purrr::map(~ .x$file) %>% purrr::map(~ .x %>% file.size) %>% as.data.frame() %>% 
+    `colnames<-`(c("file_size")) %>%
```
serbinsh: use of piping requires magrittr yes?  Is that a depends for utils already?

ashiklom: - [ ] Style guide says to split pipes into separate lines.
- [ ] `map(~.x$file)` I think can be rewritten as `map("file")` -- that's some nice syntactic sugar from `purrr`.
- [ ] `map(.x %>% file.size)` should just be `map(file.size)` (taking into account the pipe, it's just `map(.x, file.size)`).
- [ ] `c("file_size")` in `colnames` can just be `"file_size"`. No need to use `c` for length 1 vectors

All that said, if `result` is already a `data.frame`, you should be able to do this much more simply as:

```
result %>%
  dplyr::mutate(file_size = purrr::map_dbl(file, file.size)) # _dbl makes sure output is a numeric vector 
```

base/utils/R/convert.input.R
```diff
@@ -618,6 +618,18 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
...
+    mutate(missing = ifelse(is.na(file_size),NA,TRUE)) %>%
+    mutate( empty = ifelse(file_size==0,"EMPTY",TRUE))
```
ashiklom: These should be `dplyr::mutate`. Also, `mutate` evaluates sequentially, so you can combine these into one statement. I would write this as:

```
result %>%
  dplyr::mutate(
    file_size = purrr::map_dbl(file, file.size),
    missing = is.na(file_size),
    empty = file_size == 0
)
```

Explanation:
- `dplyr::if_else` is better than `base::ifelse` because it does stricter type checking. `ifelse` does some implicit coercion that can give unpredictable results.
- Logical values are easier to work with than custom values. Leaving `missing` and `empty` as logical will make it easy later on to do things like `dplyr::if_else(result$missing, val_if_true, val_if_false)` or `if (all(result$missing))`
- Note that in your `empty` column, because a column can only have a single type, `TRUE` will be coerced to the string `"TRUE"` if at least one file is `"EMPTY"`, due to R's coercion rules. You almost definitely want to avoid this.

base/utils/R/convert.input.R
```diff
@@ -618,6 +618,18 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
...
+  
+  if( NA %in% result_sizes$missing | "EMTPY" %in% result_sizes$empty){
```
ashiklom: Related to my above comment, by using logical columns, this can be simplified to:

```
if (any(result_sizes$missing) || any(result_sizes$empty)) {...}
```

Also, you _almost never_ want to use single-character forms of `&` and `|` in `if/else` blocks, because `if` and `else` conditions are expected to be scalar whereas single character`&` and `|` are vectorized and can return a vector. `if/else` will take only the first item of vector conditions with a warning, which is almost never what you want.

If you have vector conditions that you're testing, you'll probably want to use them in combination with either `any(<condition>)` or `all(<condition>)`, and if you have multiple scalar conditions, you'll want to use the double-character forms (`&&` and `||`), which evaluate once left to right and evaluate as early as possible.

To see what I mean, compare the output of the following:

```
x <- c(TRUE, TRUE, FALSE)
y <- c(TRUE, FALSE, FALSE)
x | y # Result is length 3 vector
x || y # Result is SCALAR -- performs only the first comparison
if (x | y) print ("yes") # Should throw a warning about using only the first element
if (any(x | y)) print("yes") # Should print "yes"
if (all(x | y)) print("yes") # Should not print anything
```

base/utils/R/convert.input.R
tonygardella: The result object is now a list of multiple dataframes which makes applying things a bit different than just handling a dataframe. Changing the map calls to map("file") and map(file.size) wouldn't get the right elements. I probably can make this cleaner though. So right now, the map(~ .x$file) gets just the file column from each dataframe in the result list (maintains a list of dataframe structure) and then the purrr::map(~ .x %>% file.size) applies file.size function to each dataframe in that list and I put that into a data frame with as.data.frame. Now that I'm thinking about it I can flatten the list of dataframe with the file columns from results and apply filesize to that. Not sure if that will simplify things much though.

tonygardella:
@ashiklom When you get a chance can you take a look again? Thank you!

Issue #2053, opened 2018-08-20
GlobalSensitivity Shiny Loading Error - VM 


tonygardella:
![image](https://user-images.githubusercontent.com/8570691/44356992-ecc57c00-a47e-11e8-84d1-f2b2eed06e94.png)
To Reproduce

1. Open Web GUI with pecan VM running by going to http://localhost:6480/ in local browser
2. Click on 'Output Visualization'
3. Click on "GlobalSensitivity"
4. See error when loading.

tonygardella:
Loads Fine once I disabled add-ons in my browser.

mdietze:
How would a user know that they need to disable add-ons in their browser for this app to work? Is there a way to make our app work without requiring users to turn off their browser add-ons? I'm not sure I agree that this issue can be considered fixed at this point

robkooper:
@tonygardella do you know what addon was the one that made it fail? Does it fail for other people?

serbinsh:
@tonygardella @tonygardella @mdietze My Shiny apps are working without doing any sort of disabling of add ons, however this app is crashing for other reasons

```
2018-08-27 09:09:41 INFO   [read.output] :
   GPP NPP TotalResp AutoResp HeteroResp SoilResp NEE AbvGrndWood
   leaf_carbon_content TotLivBiom TotSoilCarb Qle Transp SoilMoist
   SoilMoistFrac SWE litter_carbon_content LAI fine_root_carbon_content
   coarse_root_carbon_content GWBI AGB Mean: 1.09e-07 5.23e-09 1.51e-07
   1.04e-07 4.77e-08 5.34e-08 4.25e-08 13.1 0.149 17.9 16.1 62.2 5.19e-05
   75.4 0.628 0.29 0.28 2.44 0.724 3.91 NaN 13.2 Median: 9.26e-10 -1.94e-09
   8.57e-08 7.36e-09 3.56e-08 4.09e-08 3.33e-08 12.6 0.00797 16.8 16.5 49.9
   3.24e-07 86.6 0.721 0 0.28 0.131 0.318 3.81 NA 12.6
Rbind ensemble output...
Load parameter values...
Get run samples...
Cbind ensemble samples...
Cbind ensemble output and samples...
plotEnsemble x:
plotEnsemble y:
Error in `[.data.frame`(ensemble.out, , x) : undefined columns selected
Error in `[.data.frame`(ensemble.out, , y) : undefined columns selected
[...]

PR #2097, opened 2018-09-08
Roxygen percent escaping + consequences

infotroph:
## Description

* Fixed five Rd syntax errors caused by unescaped percent characters in Roxygen comments
* Wrapped prior.fn example in `dontrun` -- it uses DEoptim::DEoptim, which isn't installed by default.
* Fixed namespace error in vwReg ("function `.()` not found"), fixed typos in param docs, and removed unneeded library() calls
* Removed unused dependencies from viz namespace 

## Motivation and Context
Another rabbit hole from a simple-looking patch. Original motivation was just to fix a warning about "mismatched braces in Rd", but those were masking more errors in the function examples for viz::vwReg and priors::prior.fn, so I kept fixing things until the tests passed. 

While I was at it, I also did a cleanup pass through the PEcAn.visualization dependencies, mostly removing Shiny packages and their indirect dependencies that are now handled by the Shiny apps themselves. Net result: A fresh bare-metal install now needs 16 (!) fewer packages.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


infotroph:
@mdietze Your understanding is correct, but to elaborate:

* Make doesn't currently touch any of the Shiny apps. Instead, each Shiny app is responsible for checking / installing-if-needed its own dependencies at server startup.
* But Make *could* touch Shiny if we wanted! I don't have a sense how complicated the Shiny server setup process is -- If it's a hassle and folks want a shortcut, we would add an e.g. `make run-shiny` that's available for interactive use but need not be run by `make all`.
* Each package I removed from Visualization is indeed checked for and installed by at least one Shiny app, so even though Make no longer handles them they still wind up installed when needed.
* I did my best to verify<sup>[1]</sup> that all the deps I removed from the Visualization package were *only* there as a means of installing them for Shiny -- none of their functions were used inside Visualization itself.


<sup>[1]</sup>Since I haven't seen this written down anywhere, here's my go-to procedure for answering "somePkg claims to depend on otherPkg, but does it actually use it for anything?": 
* First find the easy stuff (namespace-qualified calls, importFrom statements, and sometimes useful comments): `grep -R 'otherPkg' path/to/somePkg/`
* Then search for any functions called without namespace (usually attached by `Depends: otherPkg` or `library(otherPkg)`): `Rscript -e 'devtools::check("somePkg")' | grep 'no visible global function definition'` and figure out which package provides each of the functions reported. If they're from otherPkg, then it's in use. (And all of these messages should be fixed no matter which package they're from!)
* Last, edit somePkg/DESCRIPTION to remove all references to OtherPkg, recheck for "no visible global function" messages, rerun tests to make sure nothing broke.

mdietze:
> This was a holdover from before we formalized the each-app-responsible policy.

I don't recall when this became policy (I'm guessing at a meeting that occurred during my recent summer travels) but I wasn't on that call and I don't agree that this is a good choice. I've seen some of the recent PR that include package installs and the approach strikes me as hacky. I've used similar approaches in labs and found it unreliable. At a higher level, conceptually I don't like the idea of run-time installation, particularly for Shiny, which has always been buggy for us (why create more things that can break), and is designed to be quick and interactive (a first time user may give up on an app that is unresponsive the first time they use it because it's off downloading a bunch of libraries in the background). It also opens up issues of system-specific user permissions that could cause trouble (e.g. if the www or apache account that's running the app doesn't have permissions to install libraries). As I noted a few weeks ago in #2053 I think we should really move to having the shiny apps only depend on PEcAn packages, so that the general PEcAn build ensures that packages are installed. 

In my mind PEcAn.visualization is the exact right place for these general Shiny libraries to be installed since a lot of the core functionality of these visualization apps really should be in this package.

infotroph:
> I wasn't on that call

My memory of the gist of the discussion: 
    1. The point of encapsulating things into packages is to divide up responsibilities cleanly, including keeping track of the necessary and sufficient dependencies the functionality *in that package*. Declaring dependencies so they can be used elsewhere is no less hacky than omitting dependencies on the assumption they're provided elsewhere. And R's package checks correctly yell at us for doing either one.
    2. If a Shiny app is distinct enough to live in its own directory, it's also distinct enough to handle its own dependencies<sup>1</sup>. If some of its functions belong in a different package, we should move them there and declare their dependencies *at the time we move them*, but not before.
    3. Like it or not, running Shiny adds a lot of dependencies, and someone who wants to install PEcAn without them ought to have a way of doing that.
    4. None of the above principles obligate any runtime code to contain `if(!requireNamespace("some_dep"){install.packages("some_dep")`, we just thought that was the easiest way to do it. It just means each Shiny app is required to know what its dependencies are and is not allowed to assume Some Other Package Took Care Of That Already. 

I think I remember @bcow, @robkooper, and @serbinsh having input here -- any of y'all want to amend my summary?


> It also opens up issues of system-specific user permissions

This is a really good point, I don't think we discussed it on the call, and it should be fixed.

> conceptually I don't like the idea of run-time installation

(1) Bear in mind this is *server* runtime, not app runtime. If I understand the Shiny model correctly, this means if I install PEcAn and check once that a particular app loads at all, then I can be confident all its dependencies are installed.
(2) I think this is another argument for `make install-shiny`.

<sup>1</sup>Is there any reason a Shiny app can't *also* be an R package? Seems to me like that would (1) match the conceptual separation-of-responsibility model, and (2) automate the dependency issues.

mdietze:
In theory I get the idea of dividing responsibilities cleanly, but AFAIK you can't make a Shiny app a R package (@infotroph brings up this idea too).  In practice then, to me it makes sense to push as much of the app's guts into packages as possible so that they can be subject to continuous integration, testing, and package dependency installation.

I'm less convinced by the argument that someone is going to do a full install of everything in PEcAn **EXCEPT** Shiny -- doesn't seem like a plausible use case. What we see in practice is that by far the largest use case for partial installs is installing specific model packages and data ingest packages for the sake of remote execution. Those cases don't involve installing PEcAn.visualization or benchmarking, or any of the other webservices that only run on the head node.

robkooper:
Actually right now with the dockerization of PEcAn we do want a slim version of PEcAn that does not include all the shiny dependencies, and an image that is shiny.

One option that gets to both, is maybe add a PEcAn.shiny package, this contains all the dependencies for all the shiny packages, as well as any shiny specific pieces of code we want to share between shiny apps (if we want to use it outside of shiny it should move to an appropriate package). At that point we can do a `R INSTALL PEcAn.shiny` which will install all shiny dependencies as well.

infotroph:
After thinking more and experimenting a little, I take back my "server runtime" vs "app runtime" distinction and I understand @mdietze's performance concern better. 

I think the core issue I was misunderstanding here is that Shiny apps have no clear installation process ("They're just webpages! Put them in shiny/ and they run when you load the URL!"), so runtime is the only time it's *possible* to enforce the dependencies. I agree that this argues for one `library("PEcAn.shiny")` that provides everything rather than each app needing to check a bazillion imports.

That said, I'll return to the patch and and reiterate that *right now*, PEcAn.visualization isn't using the deleted packages and has no way of distinguishing "dependency used elsewhere" from "unused cruft". Even if some of these will be added back in later by relocated code, I'd strongly prefer that they be deleted until then.

Would it be helpful if I add a Make rule that runs all the Shiny apps once at build time, so that they find and install packages then rather than when run by the www account? Maybe literally just `for app in `ls shiny/`; do Rscript shiny/$app/server.R; done`

infotroph:
Per today's PEcAn call: I will implement a very simple first pass at `make shiny`, open an issue for strategies to add Shiny into integration tests, and ping @mdietze here when ready for review.

infotroph:
@mdietze this is ready for another look. I've punted the Shiny discussions elsewhere: CI strategy in #2121, dependency installation is in #2120.

PR #2101, opened 2018-09-13
Check provided met path provided by user WIP

tonygardella:
## Description
This takes in the provided met path, checks if it's a file in the database, checks if that file is associated with an input record, and then checks if the format of the input it's associated with is the format associated with model being used. Unless the path meets all these conditions, it will exit with a specific message notifying the user why the path provided is invalid.

## Motivation and Context
Previously, if a path was put into the met section of pecan.xml, we would use that path without regard to it being in the database or if it was of the correct format or not. This has led to some confusion to users.

@araiho You're my reviewer this week so I'm tagging you. Let me know if you do not have time. I can assign someone else.
@serbinsh This should stop anyone running into the problem of providing a met path.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


tonygardella:
Just realized I forgot to check the date too. Added the WIP until I add that and test.

tonygardella:
@robkooper So in this code, if a user provides a met path to a file it checks the database for that file to make sure it's associated with the correct records. However, in the Travis testing, the clim file that gets used to test run Sipnet isn't in the database so it fails. What machine is the sipnet.clim file on? I can create the records for the file.

robkooper:
The location for the clim file in the travis.sipnet.xml file in tests, see https://github.com/PecanProject/pecan/blob/develop/tests/travis.sipnet.xml#L52

tonygardella:
CLOSING this so we can maintain the functionality of adding met without having it within the database.

PR #2108, opened 2018-09-20
Missing index in convert.inputs

tonygardella:
## Description
The object `existing.input` is turned into list and so one instance of it being called did not take into account that structure. 

## Motivation and Context
Made for a weird bug that was hard to track down because it would make met not work  and then work at different stages because the correct records were actually being created but the result object sometimes did not contain an input.id to pass on. I'd run it again and it would work, but since the conversion actually worked and the records in BETY were there it would pass through the step. Took a bit to test and make sure this worked  because I had to keep cleaning the DB before testing. It's a small thing but it messed up every step when there was an existing input on other machine and a new file needed to be inserted after the initial stage of met.process.

@ankurdesai tagging you as reviewer because I think this solves many of the problems you and I were seeing.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/utils/R/convert.input.R
+++ b/base/utils/R/convert.input.R
@@ -709,7 +709,7 @@ convert.input <- function(input.id, outfolder, formatname, mimetype, site.id, st
                                              in.prefix = result[[i]]$dbfile.name[1], 
                                              'Input', existing.input[[i]]$id, 
                                              con, reuse=TRUE, hostname = machine$hostname)
-        newinput$input.id  <- c(newinput$input.id, existing.input$id)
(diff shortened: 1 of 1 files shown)


tonygardella:
Checked the rest of the instances of existing.input after it's turned into a list that need this and they all are called correctly.

serbinsh:
@tonygardella is this fix also relevant to develop?  If so can we also get this fix into that branch?

tonygardella:
@serbinsh Yep totally relevant to develop. We'll get the release branch pulled in today.

PR #2111, opened 2018-09-24
Bd 2251 polyglot to fence

yan130:
## Description
use browndog URL instead of polyglot

Not sure where to add BrownDog R libaray install information. 
Not sure about the test


mdietze:
library should be added to modules/data.atmosphere/DESCRIPTION
unit tests should be added to modules/data/atmosphere/tests

modules/data.atmosphere/R/met.process.R
```diff
@@ -459,13 +459,12 @@ browndog.met <- function(browndog, source, site, start_date, end_date, model, di
...
-  userpass <- paste(browndog$username, browndog$password, sep = ":")
-  curloptions <- list(userpwd = userpass, httpauth = 1L, followlocation = TRUE)
-  result <- RCurl::postForm(paste0(browndog$url, formatname, "/"),
-                     fileData = RCurl::fileUpload("pecan.xml", xmldata, "text/xml"), .opts = curloptions)
-  url <- gsub(".*<a.*>(.*)</a>.*", "\\1", result)
-  PEcAn.logger::logger.info("browndog download url :", url)
-  downloadedfile <- PEcAn.utils::download.url(url, outputfile, 600, curloptions)
+  pecanxml <- tempfile("pecan.xml")
+  write(xmldata, pecanxml)
+  result.url <- BrownDog::convert_file(browndog$url, pecanxml, formatname, "./", browndog$token, download=FALSE)
```
robkooper: make sure to update the config.example.php
is ./ the current folder? if so please use a temp folder, since you might not always be able to write to the current folder.

yan130: which config.example.php? didn't find it. 

is ./ the current folder?  -- yes, but it's not used, since download is false.

tonygardella:
@yan130 Can you run `make document` from the pecan main pecan directory and push up the .Rd files it generates? That seems to be the thing holding up the Travis build.

yan130:
Thanks @tonygardella . just push the changes.

tests/pecan2.tests.xml
```diff
@@ -4,13 +4,10 @@
 
  <browndog>
    <!--
-   <url>http://dap.ncsa.illinois.edu:8184/convert/</url>
-   <username>browndog.user</username>
-   <password>GritGin3</password>
+     token expire after 24 hours
```
robkooper: Does the BrownDog package allow to generate a token easily? If so can you put the command here?

modules/data.atmosphere/R/met.process.R
robkooper: what is the ./ represent, if this is a location on disk please use tempfolder just in case the working directory is read-only.

yan130: no, the file doesn't download so we don't need a folder to hold it. 
it just a placeholder.

tests/pecan2.tests.xml
yan130: done

robkooper:
Talked about this in NCSA BrownDog meeting and decided to change this code to use a key instead of a token and add a special screen where the user can add this token, just as we do for the NARR page.

mdietze:
@yan130 this PR has been sitting untouched as a WIP all year. What's the plan for when this will get finished up?

robkooper:
@yan130 is no longer working at NCSA. I will take this work over.

robkooper:
closing PR for now

PR #2115, opened 2018-09-26
WIP Use pkgdown to make package html doc pages

tonygardella:
## Description
Create Package specific documentation using `pkgdown::buildsite()` which creates html pages we can host on our website. Lot's to do to make this work:
- [ ] Add `Authors@R` and requisite info to each packages Decription file. 
     - [It's a known issue](https://github.com/r-lib/pkgdown/issues/727) that without Authors@R, the build messes up.
- [ ] Cleanup vignettes so they do not get run when building. Settings examples = FALSE allows examples to not be run but not vignettes
- [ ] Follow some advice [here](https://github.com/r-lib/pkgdown/issues/715) about making multi-package docs. Essentially move each packages new "docs" directory to a cleaner directory tree.
- [ ] Add the building script to develop branch and push the clean tree to gh-pages branch. Make sure those directories are added to gitignore so they are not part of  develop.

@ashiklom Don't need to look closely now but would like feedback . I Just wanted to open a PR as I add things. Mostly looked through how pkgdown works today and tried it on our list of packages.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


.dockerignore
```diff
@@ -0,0 +1,12 @@
...
+.install
+
+tests
+contrib
+
+modules/rtm/src/RTM/*.o
+modules/rtm/src/RTM/*/*.o
+modules/rtm/src/RTM/*/*/*.o
+modules/rtm/src/*.mod
+modules/rtm/src/*.so
```
ashiklom: Why did you move all of these files into `gh-pages`? All that should be there is the HTML required to host the website. I would delete all of these, since they unnecessarily bloat the website, and introduce confusion.

scripts/build_staticdocs.R
```diff
@@ -1,37 +1,26 @@
...
+pkg_list<- list.dirs(path = ".", full.names = FALSE, recursive = TRUE) %>%
+           grep("man$",., value = TRUE) %>%
+           dirname(.)
+pkg_list <- pkg_list[-1]
+
+lapply(pkg_list, function(x){
+  print(x)
+  tryCatch(
+    {
+      pkgdown::build_site(x,examples = FALSE)
```
ashiklom: I would add the arugment `override = list(destination = file.path("docs", x))` or something to put the HTML files in something like `docs/base/utils/`, `docs/models/ed`, etc.

tonygardella:
@ashiklom I should've been more clear by what I mean by clean directory tree in my list of to do's. That refers to what you say by having just html files. I'm waiting to do this until I can get html files that actually look ok. Right now the html files have the skeleton of the webpage but do not have information in them because our tags and function headers aren't the cleanest. I was going to wait to worry about pushing to different branches until I can get at least one html that looks decent.

base/all/DESCRIPTION
```diff
@@ -0,0 +1,48 @@
+Package: PEcAn.all
+Type: Package
+Title: PEcAn functions used for ecological forecasts and
+    reanalysis
+Version: 1.6.0
+Date: 2018-08-20
+Authors@R:
+Author: David LeBauer, Mike Dietze, Xiaohui Feng, Dan Wang,
+    Mike Dietze, Carl Davidson, Rob Kooper, Shawn Serbin
```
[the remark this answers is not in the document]
tonygardella: Yeah that's part of to do. I was experimenting with just adding it to see what happens after. It results in other issues we have in our tags/headers.

ashiklom:
Right, but there's no need to pollute the git revision tree of this branch by committing the entire PEcAn source code to it. It also has the side effect of making this PR really hard to review because I have to wade through several years worth of code differences that are irrelevant to this PR.

I wouldn't push any HTML files anywhere at all until you have ones that you're satisfied with. In the absence of those, the only changes that should be in this PR are changes to all the package `DESCRIPTION` files (setting `Authors@R:...`) and your revisions to `build_staticdocs`.

tonygardella:
Closing this until I finish all the cleaning. Then I'll put the cleaned packages to develop and the htmls separately to gh-pages.

PR #2113, opened 2018-09-25
Revise Docker documentation order and introduction

ashiklom:
## Description

- Fix documentation headings for Docker section so they're all under the "Docker" heading
- Make "Docker quickstart" the second section. I feel like "quickstart" things usually come before detailed descriptions.
- Add links to main Docker documentation and two YouTube tutorials I found _really_ useful for getting started with Docker
- Tighten up some of the text in the Docker introduction section. 

## Motivation and Context

As we move towards replacing the VM with Docker containers, we should try to have better documentation for it.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2118, opened 2018-09-26
Documentation for PEcAn's `docker-compose.yml` file

ashiklom:
NOTE: This PR builds on (i.e. includes) #2114 , which in turn builds on #2113. I did this so these PRs could be reviewed and merged incrementally if desired. If you want to only review the unique changes in this PR (relative to the other two), select the last commit in the "Changes from all commits" drop-down menu on the "Files changed" tab of this PR.

## Description

This PR adds lots of documentation about PEcAn's `docker-compose.yml` file, which doubles as a more comprehensive description of PEcAn's Docker architecture.

I filled in every section, but I only have a rudimentary understanding of Traefik, Minio, and RabbitMQ, so I left notes in those sections indicating that more details were forthcoming (most likely from @robkooper).

A compiled version is available here: https://ashiklom.github.io/pecan-documentation/docker-index.html#pecan-docker

## Motivation and Context

You should never need motivation for documentation.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


docker-compose.yml
```diff
@@ -104,7 +104,7 @@ services:
 
   # PEcAn executor, executes jobs launched from web
   executor:
-    image: pecan/executor:${PECAN:-latest}
+    image: pecan/executor:${PECAN:-develop}
```
ashiklom: This is a bit misleading, because our `develop` container version is 3 months behind the 1.6.0 release. However, I put it here anticipating that we should start building containers more frequently in the future, so that `pecan/executor:develop` actually tracks the GitHub source code develop branch.

However, it's no problem to change it to `RC1.6.0` if that's preferred.

ashiklom:
I just tested these instructions locally and they generally work from a clean setup (with the caveat that you have to follow the manual troubleshooting steps for initializing BETY. I added those instructions in here as well).

book_source/06_reference/04_docker/03_architecture.Rmd
```diff
@@ -17,5 +19,186 @@ The PEcAn framework containers consist of multiple unique ways to interact with
...
+#### Top-level structure {#pecan-dc-structure}
+
+The root of the `docker-compose.yml` file contains three sections:
+
+- `services` -- This is a list of services provided by the application, with each service corresponding to a container.
+When communicating with each other internally, the hostnames of containers correspond to their names in this section.
+For instance, regardless of the "project" name passed to `docker-compose up`, the hostname for connecting to the PostgreSQL database of any given container is _always_ going to be `postgres` (e.g. you should be able to access the PostgreSQL database by calling the following from inside the container: `psql -d bety -U bety -h postgres`).
+The services comprising the PEcAn application are described below.
+
+- `networks` -- This is a list of networks used by the application.
```
robkooper: Only if two containers are connected to the same network can they communicate with each other using ports and hostnames. If two containers on different networks they will not be able to talk to each other directly, only using  ports exposed to hosting docker machine.

book_source/06_reference/04_docker/03_architecture.Rmd
```diff
@@ -17,5 +19,186 @@ The PEcAn framework containers consist of multiple unique ways to interact with
...
+We just provide the network name (`pecan`) and resort to Docker's default network configuration.
+Note that the services we want connected to this network include a `networks: ... - pecan` tag.
+For more details on Docker networks, see the [official documentation](https://docs.docker.com/network/).
+
+- `volumes` -- Similarly to `networks`, this just contains a list of volume names we want.
+Briefly, in Docker, volumes are directories containing files that are meant to be shared across containers.
```
robkooper: Volumes are also used to persist data between restarts of containers. Any data not stored in a volume (or in a folder mounted from the docker machine into the container) is lost when the container is relaunched. For example keeping the database in a folder allows the data to be saved between different runs of the database container. If we don't use the volumes, we would start with a blank database every time we restart the containers.

book_source/06_reference/04_docker/03_architecture.Rmd
```diff
@@ -17,5 +19,186 @@ The PEcAn framework containers consist of multiple unique ways to interact with
...
+
+The traefik service configuration looks like this:
+
+```{r, echo = FALSE, comment = ''}
+yaml::write_yaml(dc_yaml$services["traefik"], stdout())
+```
+
+#### `minio` {#pecan-dc-minio}
+
+[Minio](https://github.com/minio/minio) is a browser-based file manager that facilitates exploration of PEcAn files without the need for CLI tools.
```
robkooper: minio is service that will provide access to the a folder on disk. This data is made available using different protocol, such as S3 buckets, web based, etc. We mainly use this service to allow for easy access to the data using a web browser.

book_source/06_reference/04_docker/03_architecture.Rmd
```diff
@@ -17,5 +19,186 @@ The PEcAn framework containers consist of multiple unique ways to interact with
...
+[RabbitMQ](https://www.rabbitmq.com/) is a message broker service.
+In PEcAn, RabbitMQ functions as a task manager and scheduler, coordinating the execution of different tasks (such as running models and analyzing results) associated with the PEcAn workflow.
+
+Our configuration is as follows:
+
+```{r, echo = FALSE, comment = ''}
+yaml::write_yaml(dc_yaml$services["rabbitmq"], stdout())
+```
+
+Note that the `traefik.frontend.rule` indicates that browsing to http://localhost:8000 (with no further tags) leads to the RabbitMQ management console.
```
robkooper: Since the default username and password is used (guest/guest) for rabbitmq we highly recommend changing this for the production instances of PEcAn, as well as removing access to the rabbitmq interface using traeffic.

docker-compose.yml
robkooper: my thinking is that we want to use latest so people can use `docker-compose up` to bring up the latest official version of PEcAn. I would like to start pushing automatically both latest and develop to docker hub. If people would like to use develop, they should be able to use `PECAN=develop docker-compose up` to bring up the latest develop stack.

No matter what we need to be consistent and make sure we use either latest or develop for all images (except for the externally managed images).

ashiklom: I agree that we should use `latest` to be consistent...but I'm not going to make this change until we have a `pecan/executor:latest` container up on Docker Hub. Currently, the only options are `RC1.6.0` and `develop`, hence this change.

book_source/06_reference/04_docker/03_architecture.Rmd
ashiklom: Addressed in b85069e35.

book_source/06_reference/04_docker/03_architecture.Rmd
ashiklom: Addressed in 8745aeb9a.

ashiklom:
@robkooper I revised everything as you suggested _except_ changing `pecan/executor:develop` back to `latest`. As I said, I agree in principle, but I think we should have a working `docker-compose` file out of the box, and right now, there is no `pecan/executor:latest` image available, so running that will fail.

If you can create a `pecan/executor:latest` container (or link one to `RC1.6.0` -- not sure how Docker Hub does these things), then I'm happy to change it back.

ashiklom:
@robkooper If I addressed all your changes, please mark this as "Approved" so it can be merged.

PR #2120, opened 2018-09-28
Shiny dependency installation

infotroph:
## Description
Adds a new make command, `make shiny`, which checks for dependency packages declared in `shiny/*/DESCRIPTION` and installs them or dies trying, similar to the standard installation process for R package dependencies. In the process, moves responsibility for dependency checking from runtime to installation time (but beware that "installation" is loosely defined for Shiny apps - see below).

Note that Shiny dependencies still differ from R package dependencies in several important ways:
* Listing a package in `Depends` will *not* cause it to be attached at runtime, so there is no difference between `Imports` and `Depends` -- they both act like `Imports` does in an R package. If the app code needs packages loaded, it still needs to load them itself.
* Nothing checks whether the declared dependencies are in sync with the Shiny code -- there is no easy equivalent to `devtools::check()`. Authors need to manually update the dependency declarations in DESCRIPTION as they add and remove dependencies in the app code.
* Nothing enforces that dependency checks be run when the app is installed/upgraded -- but this is because Shiny apps don't have a clear installation step in the first place! If in doubt, run `make shiny` again.

## Motivation and Context

As previously discussed in #2097 and several related conference calls, our dependency handling for Shiny apps has been inconsistent and we've often relied on declaring imports in packages that don't actually use them. This PR tries to localize responsibility within each app and provide a consistent method for checking that all needed packages are installed.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


scripts/install_shiny_deps.R
```diff
@@ -0,0 +1,16 @@
+
+# Install dependencies declared by a Shiny app
+# Relies on the undocumented behavior that devtools::install_deps will work
+# on any directory containing a DESCRIPTION file, even if not an R package.
+
+path <- commandArgs(trailingOnly = TRUE)[[1]]
```
ashiklom: No need to change, just curious: Why use double brackets here? Doesn't `commandArgs` always return a character vector?

scripts/install_shiny_deps.R
```diff
@@ -0,0 +1,16 @@
+
+# Install dependencies declared by a Shiny app
+# Relies on the undocumented behavior that devtools::install_deps will work
+# on any directory containing a DESCRIPTION file, even if not an R package.
```
ashiklom: Cool! Thanks for documenting this -- perfectly anticipated my question.

scripts/install_shiny_deps.R
```diff
@@ -0,0 +1,16 @@
...
+
+path <- commandArgs(trailingOnly = TRUE)[[1]]
+
+devtools::install_deps(path)
+dep <- desc::desc_get_deps(file.path(path, "DESCRIPTION"))
+purrr::walk(
+	dep$package,
+	~if (system.file(package = .) == "") {
```
ashiklom: An interesting way to check that a package is installed. Elsewhere in PEcAn (or at least in all the Dockerfiles), we tend to use `pkg %in% rownames(installed.packages())`. Any particular reason you're doing it this way here? Again, no need to change, just curious. Though it might be nice to add a quick comment that this is checking if a package is installed.

Also, does `devtools::install_deps` throw warnings about failure to install anything? If so, would it be easier to run that with warnings temporarily converted to errors?

shiny/Data-Ingest/DESCRIPTION
```diff
@@ -5,4 +5,19 @@ Date: 2018-08-20
...
+	PEcAn.utils,
+	PEcAn.visualization,
+	shinydashboard,
+	shinyjs,
+	shinyTime,
+	shinytoastr,
+	shinyWidgets,
+	shinyjs,
+	stringr,
+	tibble
```
ashiklom: 👍 👍 👍 Thank you for adding all of these (in all the DESCRIPTION files, not just here)! This is a huge help!

shiny/Elicitation/DESCRIPTION
```diff
@@ -1,9 +1,15 @@
 Type: Shiny
 Title: Expert Prior Elicitation
 License: FreeBSD + file LICENSE
-Author: Mike Dietze <dietze@bu.edu> 
-Author: Rob Kooper <kooper@illinois.edu>
+Author:
+	Mike Dietze <dietze@bu.edu>,
+	Rob Kooper <kooper@illinois.edu>
```
ashiklom: 👍 for better formatting! For future reference, I think we might want to move to the slightly more formal (or at least `pkgdown`-friendly) form:

```
Authors@R: c(person("Mike", "Dietze", email = "dietze@bu.edu"), person(first = "Rob", last = "Kooper", email = "kooper@illinois.edu"))
```

However, since the Shiny apps aren't being `pkgdown`ed, might not be necessary here.

Makefile
```diff
@@ -10,15 +10,20 @@ MODULES := allometry assim.batch assim.sequential benchmark \
 				 data.mining data.remote emulator meta.analysis \
 				 photosynthesis priors rtm uncertainty
 
+SHINY := BenchmarkReport BrownDog Data-Ingest Elicitation Pecan.depend \
```
robkooper: if possible replace this with:

```
SHINY_DIRS := $(dir $(wildcard shiny/*/.))
SHINY := $(SHINY_DIRS:shiny/%/=%)
```

This will automatically generate the list of folders in the shiny folder, this way we don't have to remember to update the list in the Makefile

scripts/install_shiny_deps.R
infotroph: I think it was just habit here ("I'm sure I only want one value back, better use double brackets"), but now that I think about it `commandArgs()[[1]]` has the virtue of failing loudly if you forget to pass any arguments.

scripts/install_shiny_deps.R
infotroph: > Also, does devtools::install_deps throw warnings about failure to install anything?
No, alas, that's exactly what I'm trying to work around here. `install_deps` returns nothing and never throws warnings, just lists on stdout the packages it doesn't know how to install. My first draft used capture.output and grepped for the string "Not on CRAN", but that seemed awfully brittle.

> we tend to use pkg %in% rownames(installed.packages())
The docs for `installed.packages` specifically recommend against this usage -- and recommends `system.file` instead! -- because `installed.packages` is expensive and has to read *multiple* files from every package in the library. It usually hides the slowness by caching its result for the whole session, but since we use it here right after installing packages the cache is usually invalid and won't help.

> might be nice to add a quick comment that this is checking if a package is installed
This is a good call and I'll do it.

shiny/Elicitation/DESCRIPTION
infotroph: I approve of this in principle but haven't yet written an authors@R and decided I'd learned enough new things for one PR 😉

scripts/install_shiny_deps.R
infotroph: explanatory comments added in #2130

Makefile
infotroph: Done in #2130. Guess we should eventually do the same for base/models/modules, once all their contents are ready to be included in the build...

ashiklom: I don't disagree, but I did want to quickly mention that, IIRC, using `dir`, `wildcard`, and similar functions in Makefiles makes them "non-portable" for Windows users. I vaguely recall getting this flagged when running `R CMD check` on `PEcAnRTM`, which at some point used similar constructions in its Makevars file. Personally, I think bending over backwards to support Windows users isn't worth our time, at least at the level of PEcAn as a whole (individual packages should still try to work on Windows), especially now that the Docker workflow is maturing.

scripts/install_shiny_deps.R
ashiklom: Huh, good to know! We should use this in the Dockerfiles and elsewhere at some point.

Makefile
infotroph: I've been assuming that we threw out portability as soon as we started using secondary expansion. Am I confusing "portability between OSes" and "portability to non-GNU build systems"?

(also note that we already rely heavily on `wildcard` before this change, c.f. the bottom third of the Makefile)

PR #2126, opened 2018-10-01
Linking site with PFT

para2x:
## Description

In this PR, I have tried to take advantage of the relation between `sites_cultivars` and `cultivars_pfts` tables in bety and see if there is a specific PFT that is registered for a site. If this information is registered then I subset our parameters and keep just the samples form that PFT. If no information is found then function would not change anything. 
If we find this approach useful for establishing a link between the siteid and pft then the next step would be to populate the `sites_cultivars` table (because it's empty now and even with this change `write.ensemble.configs` works as business as usual.).

site_cultivar

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


para2x:
@mdietze @dlebauer  I was hoping if I could get your thoughts on this.

robkooper:
I like the idea of having PFT's associated with sites. This list of PFT's could be at the top of the list indicating that most likely these are PFT's that work on this site. This should prevent a user (me) from picking a PFT that is not applicable to the site.

para2x:
@mdietze I agree almost with all of your comments. I can still think of few more ways to generalize this a bit more. 
There is actually a table called `pft_species` in bety which has records in it for linking these two. Having a `site_species` table could help a lot replacing all the cultivar related tables (Or this could be even an option for users to choose between species/cultivar link to site)
For a data layer that would spatially map the plant species, are you referring to a specific product?

mdietze:
@para2x  yes I know about `pft_species`, my point is that it's not unique (one pft can contain many species, but each species can also be in many pfts)

para2x:
I'm looking into different data products that they have spatial distribution of plant species, see how can I use them.

mdietze:
@para2x don't get bogged down on any specific data products at this point, as there are potentially many. Big picture is the larger design, not the ingest of specific products (we've already got a workflow for ingest)

infotroph:
I haven't looked at this in detail yet, but to amplify @mdietze's third point: `cultivars_pfts` is explicitly intended to be analogous to `species_pfts` and to work as much like it as possible, and I'm pretty leery of any change that adds differences in their behavior. What is the special case that calls for it here?

para2x:
@infotroph as part of our cms project, we are working on developing a multi-site state data assimilation workflow in assim.sequantial package. At the moment with the current functionality of pecan we are forced to use same PFT for all the sites around the U.S. Now, we are thinking about different ways that a user can directly specify what PFT needs to be used for each site.
I have been trying so hard to come up with ideas that first would affect the other modules the least and then solve our problem at the meantime.

para2x:
But whatever method we end up using needs to be implemented inside the `write.ensemble.config` because that's what the SDA is using.

dlebauer:
this functionality has been commented out, and at the meeting today we decided to remove the preceding database query and messages in PR https://github.com/PecanProject/pecan/pull/3005

We can revisit the use case and potential implementation for site level PFTs in the future based on need.

PR #2133, opened 2018-10-02
Linking site with PFT using XML tags

para2x:
## Description
This is another effort for providing a different method for linking sites with PFTs. The idea is to use xml tags inside site tag in pecan xml. However, this could be complemented with separate family of functions that they can read in different data sources (such as database, lookup tables or raster files) and they would do the real linkage by updating the pecan xml and adding the required tag to the site tag. 

I don't really expected this PR to be merged real soon, however I thought it would be a practical example and a place to keep discussing about it. 

I will appreciate all ideas and comments. 
The tag in site would look like this:
```
    <site>
      <id>772</id>
      <met.start>2004/01/01</met.start>
      <met.end>2004/12/31</met.end>
      <site.pft>
      <pft.name>temperate.broadleaf.deciduous</pft.name>
      </site.pft>
    </site>

```
The code I have added in this PR will read in these tags and subsets the ensemble samples for each site given its PFT.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
  - [X] When possible
## Types of changes

## Checklist:


para2x:
This PR will change the current behavior and functionality of the `write.ensemble.function` even if the tag is not provided.

para2x:
I just pushed a new function to this branch which can read in a look up table and modify a single setting pecan xml and/or a multi-setting pecan xml file. This function adds the required xml tags linking sites to pfts and writes it down to the `outdir`.

para2x:
I believe there could be a series of functions then to create the look up table based on available data sources and send it to this new developed `site.pft.linkage` function.

mdietze:
@para2x rather than hacking around the edges, I'd really really prefer that you stop and go back to the big picture. I'd like to see a high-level plan and discussion of the pros/cons of different alternatives. I don't want to see any functions or code implemented until we decide on what the plan is -- it's not efficient to spend time coding until we agree on the high level approach we want to take.

PR #2136, opened 2018-10-04
WIP Gh-pages html

tonygardella:
## Description
Initial Renderings of pkgdown produce package descriptions. Strips everything away except the html pages.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


ashiklom:
In theory, to link to home, you should just be able to add a relative link to the higher working directory (since you're in `base/all/docs/`, the link would be `../../../index.html`), but I'm not sure if it's that simple. It would be cool to get it up in the navigation bar so it was visible from every page. In the meantime, you could just add something like this to each package's README:

```markdown
[Return to package index](../../../index.html)
```

tonygardella:
True that could work in the meantime and yeah I want add it to the navigation bar because the "home" button should  bring someone back to the head index page. I think I can write a html file with a  chunk that can live in wherever the pages will live that overrides the home button to link to the head index page.

tonygardella:
WIP until finalize decision on where they should go.

tonygardella:
Unnecessary now: http://pecanproject.github.io/pkgdocs.html
Next steps will be to finish the automation script that pushed to the website repository.

PR #2144, opened 2018-10-11
Linking sites with PFTs

para2x:
## Description

This PR is my latest effort for establishing a link between PFTs and sites. The first graph conceptually shows what is envisioned for this process. From the bottom, there would be a series of functions handling the processes of reading in different data sources and creating a Lookup Table (LUT). Then another function is responsible for parsing the LUT and adding a xml tag to the site in pecan xml specifying the exact pft that needs to be used for this site/simulation. Finally, write.ensemble.config will recognize the site/pft xml tag and subsets the parameters samples based on the defined PFT.
site-pft 1
Below you can see the flowchart implemented with the real function names.  The processes of reading in the LUT and adding the tags will be part of prepare settings functions and consequently the modified pecan xml will be written down afterwards as the `pecan.CHECKED.xml`.
site_pft flowchart

**The whole linkage process is optional at the moment and if you don't have the required tag this is not going to affect you at all.** 
If you're interested to you use this feature you need to add the following tag under run > inputs.

```
      <pft.site>
      <path>site_pft.csv</path>
      </pft.site>
``` 
At the moment I have functions that support `.csv` and `.txt` files which are comma separated and have the following format:

```
site, pft
1000025731,temperate.broadleaf.deciduous
764,temperate.broadleaf.deciduous
```
## Review Time Estimate

- [ ] Immediately
    - [X] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/settings/R/loadPath_sitePFT.R
```diff
@@ -0,0 +1,24 @@
...
+#' @description The csv or the text file needs to have a header and be separated using comma. Under the first column in the text file, one needs to specify the site id and in the second column there has to be the name of the PFT.
+#'
+#' @examples
+loadPath.sitePFT<-function(settings, Path){
+  #finding the file extension.
+  ext <- tools::file_ext(Path)
+  
+  if (ext=="csv" | ext=="txt"){
+    # reading in the links
+    links <- read.table(file.path(settings$outdir,Path), header = T, sep = ",")
```
mdietze: I don't understand the assumption that the file will be located in the outdir. That seems pretty unlikely, as the outdir is run specific, while inputs are usually stored in a completely different directory and registered with the DB. Furthermore, our input processing code is set up to resolve the full paths of inputs. Seems like it would be simpler to just have this function take a single argument, Path.

base/settings/R/loadPath_sitePFT.R
```diff
@@ -0,0 +1,24 @@
+#' Title loadPath.sitePFT
+#'
+#' @param settings pecan setting list.
+#' @param Path  Character of file name with extension. The path will be generated using the outdir tag in pecan settings.
+#'
+#' @return a dataframe of two columns of site and pft
+#' @export loadPath.sitePFT
+#' @description The csv or the text file needs to have a header and be separated using comma. Under the first column in the text file, one needs to specify the site id and in the second column there has to be the name of the PFT.
```
mdietze: There's ambiguity here about how the system handles multi-PFT runs. Since you're strictly assuming that there are two columns then I assume you have to enter the site ID multiple times (e.g. a site with 12 PFTs would have 12 rows).

base/settings/R/pft_site_linker.R
```diff
@@ -0,0 +1,66 @@
+#'
+#' @title site.pft.linkage
+#' 
+#' @param pecan.xml.address pecan settings list.
```
mdietze: this doesn't match the function itself, where this is just `settings`
```suggestion
#' @param settings pecan settings list.
```

base/settings/R/pft_site_linker.R
```diff
@@ -0,0 +1,66 @@
...
+#' @param site.pft.links dataframe. Your look up table should have two columns of site and pft with site ids under site column and pft names under pft column.
+#'
+#' 
+#' @description This function creates the required tags inside pecan.xml to link sites with pfts given a look up table. If the required tags are already defined in the pecan xml then they will be updated.
+#' 
+#' @return NONE
```
mdietze: This says `return NONE` but the function returns `return(new.mset)`. Update documentation

base/settings/R/pft_site_linker.R
```diff
@@ -0,0 +1,66 @@
...
+#'  "1000025731", "temperate.broadleaf.deciduous11",
+#'  "1000000048", "temperate.broadleaf.deciduous2",
+#'  "772", "temperate.broadleaf.deciduous3",
+#'  "763", "temperate.broadleaf.deciduous4"
+#' )
+#'  
+#' # sending a single setting xml to the function
+#' site.pft.linkage(settings,site.pft.links)
+#' # sending a multi- setting xml file to the function
+#' site.pft.linkage(settings,site.pft.links)
```
mdietze: Given that the code is identical, I don't think you need different examples for sending single settings vs multiSettings

base/settings/R/pft_site_linker.R
```diff
@@ -0,0 +1,66 @@
...
+  if(!is.MultiSettings(settings)) settings<- list(settings) 
+  
+  #for each site in this setting
+  new.mset <- settings%>%
+    purrr::map(function(site.setting){
+      
+      site.pft <- NULL
+      site.id <- (site.setting[['run']])$site$id
+      #if no site id was found
+      if (is.null(site.id)) PEcAn.logger::logger.severe('Your site needs to have a site id.')
```
mdietze: * do you really want to stop a full multiSettings run if one site is missing a site ID?
* Would be nice to have a more informative error message -- if you're running 1000's of sites and 1 is missing and ID, this would be a pretty uninformative message.

base/settings/R/pft_site_linker.R
```diff
@@ -0,0 +1,66 @@
...
+      # see if we can find this site id in the LUT
+      if (site.id %in% site.pft.links$site) site.pft <-site.pft.links$pft[which(site.pft.links$site %in% site.id)]
+      # if there was a pft associated with that
+      if (!is.null(site.pft)) site.setting[['run']]$site$site.pft <- as.list(site.pft) %>% setNames(rep('pft.name', length(site.pft)))
+      site.setting
```
mdietze: what is this line doing? should this be 
```suggestion
      return(site.setting)
```

base/settings/R/pft_site_linker.R
```diff
@@ -0,0 +1,66 @@
...
+    })
+  
+  #putting it in the right format depending if it's multisetting or not 
+  if (is.MultiSettings(settings)) {
+    new.mset <- MultiSettings(new.mset)
+    outdir <-new.mset[[1]]$outdir
```
mdietze: outdir is grabbed but never used. Looks like you were previously trying to write out the xml but decided that was unnecessary, but haven't cleaned up these outdir or the `#write.settings(new.mset, paste0("pecan.site_pft.xml"), outdir)` below

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -399,6 +399,19 @@ The following are the most common types of inputs, along with their correspondin
 ##### (Experimental) `veg`: Vegetation initial conditions {#xml-run-inputs-veg}
 
 (Under construction. Follow developments in the `PEcAn.data.land` package, located in `modules/data.land` in the source code).
+##### `pft.site` If you're interested to use a specific PFT for your site/sites you can use the following tag to tell interanl functions, which PFT needs to be used for what site.
```
mdietze:
```suggestion
##### `pft.site` Multi-site site / PFT mapping

When performing multi-site runs, it is not uncommon to find that different sites need to be run with different PFTs, rather than running all PFTs at all sites. If you're interested to use a specific PFT for your site/sites you can use the following tag to tell PEcAn which PFT needs to be used for what site.
```

book_source/03_intermediate_users_guide/01_pecan_xml.Rmd
```diff
@@ -399,6 +399,19 @@ The following are the most common types of inputs, along with their correspondin
...
+  <path>site_pft.csv</path>
+</pft.site>
+```
+For example using the above tag, user needs to have a csv file named `site_pft` stored in the pecan folder. At the moment we have functions supporting just the `.csv` and `.txt` files which are comma separated and have the following format:
+```
+site_id, pft_name
+1000025731,temperate.broadleaf.deciduous
+764,temperate.broadleaf.deciduous
+```
+Then pecan would use this lookup table to inform `write.ensemble.config` function about what PFT need to be used for what site.
```
mdietze:
```suggestion
Then pecan would use this lookup table to inform `write.ensemble.config` function about what PFTs need to be used for what sites.
```

base/settings/R/loadPath_sitePFT.R
para2x: That's correct. 
I'm hoping the process of creating the lookup table will be all automated by a series of functions like `loadPath.sitePFT` which makes this just a matter of formating the lookup table.

PR #2149, opened 2018-10-13
WIP: Document standards for dependency declaration 

infotroph:
## Description
This is a first stab at writing down our current approach to importing dependencies... or at least the framework I've been working from in my last few months of package cleanup PRs, so if it's not our approach I need a copy of the memo 😄 

I *do not* think this is ready to merge as-is -- Please help me edit it! Known deficiencies:

* Way too wordy compared to the rest of the chapter
* Contains advice that isn't really "coding style". Should some or all of it live elsewhere?
* Should probably update the `library vs require` section while we're at it, to strongly discourage both. (OK to use `library` in interactive code, but in package code use `::` nearly everywhere or `requireNamespace(..., quietly = TRUE)` when absolutely necessary).

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
@araiho You're paired with me for code review this week -- are you willing to read and suggest improvements?

tonygardella:
Definitely agree on your third point.

tonygardella:
I think your advice is fine here under "Coding Style" as it is a part of "Coding Practice".

book_source/05_developer_guide/04_coding_practices/Coding_style.Rmd
```diff
@@ -71,13 +71,24 @@ When another package is required by a function or script, it can be called in th
...
+
+In some of the PEcAn model interface packages, we push the definition of "not necessarily needed" by declaring the coupled model package in `Suggests` rather than `Imports`. This is a compromise to simplify installation of PEcAn for users who only plan to use one or a few models: For example, the `PEcAn.BIOCRO` package cannot do anything useful when the BioCro model is not installed, but it lists BioCro in Suggests because *PEcAn as a whole* can work without it. This way users can avoid the bother of installing BioCro if they only wanted to run, say, SIPNET.
```
ashiklom: Not necessarily for this PR, but it might be good to mention here the circumstances under which packages in `Suggests` are or are not installed. For instance, `install.packages` has the `dependencies` argument, and by default installs `Depends`, `Imports`, and `LinkingTo`, but _not_ `Suggests`.  I'm pretty sure `devtools::install*` also defaults to this, based on its documentation. However, in PEcAn's Makefile, we currently have `dependencies = TRUE`, which forces `Suggests` to also be installed. It might actually be worth changing that, since Suggests packages often give us a lot of problems.

infotroph: >mention here the circumstances under which packages in Suggests are or are not installed

Good call. I was thinking I should be explicit that DESCRIPTION isn't magic (especially: `Imports:` declares your /intention/ to import from those packages, it doesn't do anything to the namespace), and I can try to work that in as well.



> we currently have dependencies = TRUE, which forces Suggests to also be installed. It might actually be worth changing that, 

R CMD check is pretty insistent that it wants all Suggests installed. I think it'd be /possible/ to work around that, but given how many of our suggests are in fact used for checking I don't know if we'd save any hassle. Worth a try, though.

ashiklom:
Bump. Personally, I think this is worth merging as is. It already reads well, and dependencies are an important and difficult conversation that justifies a more wordy discussion. We can punt my comments about installing "Suggests" packages into its own issue.

That said, to save some headache, it may be worth closing this PR and adding this content into @tonygardella's big documentation PR (#2310). Since this is only one commit that is limited to a single file, it should be pretty painless to apply it via `git rebase` or `git cherry-pick` (or maybe even `git merge`, if done carefully?) on top of @tonygardella's existing changes. You could even consider just copy-pasting the content from this commit as a new commit, and tagging @infotroph as the "real" author in the commit message.

infotroph:
From a discussion on #2358 that seems potentially worth adding to this documentation as a policy:

> Both specifically here and as a general rule to keep in mind, I think maybe we should try to avoid adding packages to the Travis installation list as part of routine PRs. Unless devtools can't figure out how to install it automatically or it takes so long to compile that it makes the build fail, I think it makes more sense to merge the within-package changes and then evaluate whether to tweak the Travis build as a separate PR.
>
> My reasoning is that using C2D4U binaries seems to come with more chance of version conflicts than I was originally expecting, and that in many cases it's probably faster on average to compile once and then use the Travis cache rather than wait for the download from Launchpad every time.
>
> I'm only 70% sure this is a good idea, and am open to counterarguments -- maybe this is just asking for trouble later when we know the build has gotten slow but have forgotten which packages got added?

mdietze:
@infotroph @ashiklom @tonygardella This seems like a relatively small documentation PR that's been lingering. Can we get it out of WIP and get it merged in soon?

infotroph:
@mdietze I should be able to come back to this over the weekend and clean it up

infotroph:
So uh... I took another pass at editing this and now it's a 2909-word screed. Vicious editors needed, please.

PR #2152, opened 2018-10-16
DOCKER: Make web post RabbitMQ messages even after edit

### LINKED ISSUE #2151: Dockerized web workflow tries to run directly if "Edit XML" or "Advanced Edit"
**Describe the bug**
Using the Dockerized web interface -- if the user tries to edit the PEcAn XML through the web interface, the web container tries to execute the workflow itself (i.e. by calling `$Rbinary`) rather than posting a message to RabbitMQ to trigger the `executor` container. This silently fails (i.e. web workflow hangs) because none of the required components (including R itself) are available on the web container.

**To Reproduce**
1. Run the Dockerized PEcAn web workflow.
2. Select "Edit PEcAn XML"
3. Save the XML and try to proceed with the workflow.

The "Job executing" page should hang with no output. A tiny little message in the web container logs will show something like `sh: /usr/bin/R not found`.

**Expected behavior**
Web container should post a message to RabbitMQ after PEcAn edit, which will be picked up by the executor and the workflow will proceed as before.

**Machine (please complete the following information):**
PEcAn Docker.

**Proposed solution**

Add the RabbitMQ submission code from `04-runpecan.php` to whatever happens after `pecan_edit`, `model_edit`, etc.
As a band-aid, this can be copy-pasted, but a better solution would involve writing a PHP function (in `common.php`) to send a RabbitMQ message, which can then be re-used more easily.

ashiklom:
Also, refactor RabbitMQ functionality in web interface into its own functions.

Also, allow `docker/receiver.py` to receive custom applications via `custom_application` message.

This PR fixes #2151.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/web/07-continue.php
+++ b/web/07-continue.php
@@ -73,10 +73,15 @@ if (file_exists($folder . DIRECTORY_SEPARATOR . "STATUS")) {
 }
 
 # start the workflow again
-chdir($folder);
-pclose(popen("$exec &", 'r'));
+if ($rabbitmq_host != "") {
+  $msg_exec = str_replace("\"", "'", $exec);
+  $message = '{"folder": "' . $folder . '", "custom_application": "' . $msg_exec . '"}';
+  send_rabbitmq_message($message, $rabbitmq_queue);
+} else {
+  chdir($folder);
+  pclose(popen("$exec &", 'r'));
+}
 
 #done
 header("Location: $path");
 ?>
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 ## Unreleased
 
 ### Fixes
-- Fixed error in read_web_config which would filter out all variables.
+- Fixed error in `read_web_config` which would filter out all variables.
+- Docker:
+  - Make sure web interface posts RabbitMQ messages even after editing files (fixes #2151)
 
 ### Added
 - Lots of new documentation for running PEcAn using Docker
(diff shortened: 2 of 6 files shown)


web/07-continue.php
```diff
@@ -73,10 +73,15 @@
 }
 
 # start the workflow again
-chdir($folder);
-pclose(popen("$exec &", 'r'));
+if ($rabbitmq_host != "") {
+  $msg_exec = str_replace("\"", "'", $exec);
```
ashiklom: This replaces all double quotes with single quotes in the command. So, `R_LIBS_USER="/path/to/R/library" "/usr/bin/R"` becomes `R_LIBS_USER='/path/to/R/library' '/usr/bin/R'`. This is because JSON doesn't support single-quoted strings, which would normally be the preferred way to escape this.

NOTE that bash _does not_ perform expansion inside single quotes, so any variables (i.e. `$var`) included in the command will _not_ be expanded.

web/07-continue.php
```diff
@@ -73,10 +73,15 @@
...
+  $message = '{"folder": "' . $folder . '", "custom_application": "' . $msg_exec . '"}';
```
robkooper: problem with this is that I can send now arbitrary code to the container to be executed. I'd rather have flags that we can pass along and use these flags to do specific things in the executor container.

ashiklom: I get where you're coming from, but I'm tempted to punt on it because that fix is a bit more involved. Also, the current system _already_ allows you to execute any arbitrary code on the containers via a custom `job.sh` script. IIRC, we can even edit that directly from the web interface, via "Edit Model Configs".

ashiklom: Case in point:

screen shot 2018-10-18 at 9 42 10 am

robkooper: yeah, we should figure out how to fix that :) maybe only allow edit if logged in.

PR #2153, opened 2018-10-16
Add download method argument to `download.CRUNCEP`

ashiklom:
## Description
New argument `method` defaults to "opendap" (current). However, specifying `method = "ncss"` (a.k.a. `<method>ncss</method>` in XML) uses the slower but much more reliable NetCDF subset method, which involves subsetting the file on the server first, then downloading the subsetted file (to `tempfile()`), and then reading it locally.

## Motivation and Context

OpenDAP is a notoriously unreliable download method. In my experience, NetCDF subset, although somewhat slower and not quite as elegant (because files have to be downloaded locally), fails _much_ more rarely. In my tests, it has never once failed, whereas OpenDAP fails at least 50% of the time.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Added
 - Lots of new documentation for running PEcAn using Docker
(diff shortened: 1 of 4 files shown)


ashiklom:
Gah, I tried a test run and `<method>` tag seems to get lost at some point and doesn't make it to the `remote.execute.R` call. I'll have to investigate.

Issue #1265, opened 2017-03-02
CRUNCEP ocean mask excludes coastal sites


mccabete:
The met driver CRUNCEP has a 0.5 * 0.5 resolution, so certain coastal sites (Like Kennedy Space Center, for example, but I'm sure there are others) are treated as ocean and fail to provide met data. 

## Description

The two suggestions brought up at todays meeting were to automatically shift to the nearest land grid cell instead of just returning an error. That, or interpolate from multiple surrounding land grid cells. 

## Context

This would be useful to prevent hacks like the one I'm currently using where you  just create a fake site that's in the next nearest grid cell.

This will likely be an issue for any course-resolution met product. I'm not sure if we already solved this problem elsewhere.

ashiklom:
Bumping this because it _also_ shows up in coastal areas next to large lakes. Case in point -- I was trying to do runs at UMBS site, whose coordinates round into the adjacent lake.

For reference, [here](https://thredds.daac.ornl.gov/thredds/catalog/ornldaac/1220/catalog.html?dataset=1220/mstmip_driver_global_hd_landwatermask_v1.nc4) is the link to the CRUNCEP land-water mask that we should use in solving this problem.

ashiklom:
Closing because this should have been resolved in #2155. I was able to use PEcAn to do runs at UMBS. @mccabete let me know if that code doesn't work for the Kennedy Space Center or other coastal sites.

PR #2155, opened 2018-10-18
Check against CRUNCEP land-sea mask during `download.CRUNCEP`

ashiklom:
## Description

Downloads the CRUNCEP 0.5 degree land-water mask (small file, ~250 KB) and checks if the specified site coordinate falls on land. If yes, proceed as usual. If not, look for land 1 degree in every direction, and if any pixels are land, use data for those coordinates instead. If no land within 1 degree of coordinates, throw an informative error.

## Motivation and Context

Addresses issue #1265 for CRUNCEP specifically. The same issue may appear for other met products, in which case this should be generalized, but this fix should work for now.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/download.CRUNCEP_Global.R
+++ b/modules/data.atmosphere/R/download.CRUNCEP_Global.R
@@ -50,6 +50,62 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
   lat.in <- as.numeric(lat.in)
   lon.in <- as.numeric(lon.in)
 
+  # Convert lat-lon to grid row and column
+  lat_grid <- floor(2 * (90 - lat.in)) + 1
+  lon_grid <- floor(2 * (lon.in + 180)) + 1
+
+  # Check against land-sea mask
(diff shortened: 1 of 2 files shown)


modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -12,12 +12,26 @@
...
 ##' @export
 ##'
-##' @author James Simkins, Mike Dietze
+##' @author James Simkins, Mike Dietze, Alexey Shiklomanov
 download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, lon.in,
-                             overwrite = FALSE, verbose = FALSE, maxErrors = 10, sleep = 2, ...) {
+                             overwrite = FALSE, verbose = FALSE, maxErrors = 10, sleep = 2,
+                             method = "opendap", ...) {
+
+  if (is.null(method)) method <- "opendap"
```
ankurdesai: If ncss is more reliable, should we just default to that?

Issue #1922, opened 2018-04-30
connect do.conversions overwrite to settings


mdietze:
@infotroph recently discovered that the logical overwrite flags being passed into do.conversions were not connected to any information in the settings file, and thus couldn't be controlled via advanced edit of the pecan.xml.  

Furthermore, the names and identities of these flags are fixed, which means that we would have to modify the function definition every time we add a new input conversion into do.conversions, which doesn't seem sustainable.

Proposed solution:
* Deprecate overwrite flags as do.conversions arguments
* Instead, overwrite should be set as a logical variable in `settings$run$inputs$<tag>$overwrite`
* In the loop over inputs within do.conversions, we should just pull out the overwrite info and pass it on to the specific conversion function being called.

ashiklom:
While we're at it, `do_conversions` needs better documentation. If the function is only going to take a settings list as an argument (which I generally oppose in favor of explicit arguments<sup>1</sup>, but I can see how it could make sense here), then we need to be _crystal clear_ about exactly which elements of `settings` it needs. We should also be explicitly checking that all the required elements are there, and failing loudly if they are not. E.g.

```r
PEcAn.logger::severeifnot(
  "run" %in% names(settings),
  "host" %in% names(settings),
  ...,
  msg = "One or more elements are missing from the settings list"
)
```

(Or better yet, we should have a `logger` or `utils` utility `needs_element(list, elements)` that checks a list for required elements and returns an informative error if any are missing).

While we're on the topic, I also think `do_conversions` should be moved to the `settings` package. In general, I think all of our functions that work primarily with a `settings` object should live in `PEcAn.settings`, and the remaining functions should take all their arguments explicitly (see also the below idea about S3 methods).

-----

<sup>1</sup> One idea I've been tossing around is to use S3 methods with a `settings` class to do get the best of both worlds. For instance:

```r
met.process <- function(...) {
   UseMethod("met.process")
}

met.process.default <- function(site, start_date, end_date, <etc>) {
   # Do stuff...
}

met.process.settings <- function(settings) {
  needs_elements(settings, c("run::site", "run::start_date", "run::end_date"))
  met.process.default(settings$run$site, settings$run$start_date, settings$run$end_date)
}
```

That said, I don't think this would necessarily work well for `do_conversions`. I think this could also make R CMD check complain because S3 methods are supposed to have the same dispatch argument name.

infotroph:
> While we're on the topic, I also think do_conversions should be moved to the settings package

[...]

PR #2168, opened 2018-11-02
Move do_conversions to PEcAn.workflow

infotroph:
## Description
Part of #1681, #1722. Keeps a deprecated copy in utils for now, to be removed in a future release.

## Motivation and Context

There are two big reasons do_conversions is a better fit for workflow than utils:

1. Workflow is the correct home for functions called directly to do a piece of the workflow.
2. Having it in utils causes circular dependencies with data.atmosphere and data.land.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


infotroph:
Note that this PR just moves the existing version of do_conversions and doesn't make any of the changes discussed in #1922. Those changes should be done in PEcAn.workflow and *not* copied back to the deprecated version in PEcAn.utils.

PR #2181, opened 2018-11-05
fix dependencies list

robkooper:
Script to loop trough all DESCRIPTION files and put all PEcAn dependencies in the Makefile.

When installing PEcAn.workflow I noticed it was missing data.atmosphere. This fixes all the dependencies, including the script to generate this list.


Makefile
```diff
@@ -73,42 +73,46 @@ $(subst .doc/models/template,,$(MODELS_D)): .install/models/template
...
-$(call depends,models/jules): | .install/base/utils .install/base/remote
-$(call depends,models/linkages): | .install/base/utils .install/base/remote
-$(call depends,models/lpjguess): | .install/base/utils .install/base/remote
-$(call depends,models/maat): | .install/base/utils .install/base/remote .install/base/settings
-$(call depends,models/maespa): | .install/base/utils .install/base/remote .install/modules/data.atmosphere
-$(call depends,models/preles): | .install/base/utils .install/modules/data.atmosphere
-$(call depends,models/sipnet): | .install/base/utils .install/base/remote .install/modules/data.atmosphere
-$(call depends,models/template): | .install/base/utils
-
+$(call depends,models/ed.Rcheck/PEcAn.ED2): | .install/base/utils
```
ashiklom: Extra line.

```suggestion
```

Makefile
```diff
@@ -73,42 +73,46 @@ $(subst .doc/models/template,,$(MODELS_D)): .install/models/template
...
-$(call depends,modules/meta.analysis): | .install/base/utils .install/base/db .install/base/settings .install/modules/priors
-$(call depends,modules/priors): | .install/base/utils
-$(call depends,modules/rtm): | .install/modules/assim.batch .install/base/utils .install/models/ed
-$(call depends,modules/uncertainty): | .install/base/utils .install/base/db .install/modules/priors .install/modules/emulator
-$(call depends,models/biocro): | .install/mockery .install/base/utils .install/base/settings .install/base/db .install/modules/data.atmosphere .install/modules/data.land .install/base/remote
-$(call depends,models/cable): | .install/base/utils
-$(call depends,models/clm45): | .install/base/utils
-$(call depends,models/dalec): | .install/base/remote
+$(call depends,base/all): | .install/base/db .install/base/settings .install/modules/meta.analysis .install/base/logger .install/base/utils .install/modules/uncertainty .install/modules/data.atmosphere .install/modules/data.land .install/modules/data.remote .install/modules/assim.batch .install/modules/emulator .install/modules/priors .install/modules/benchmark .install/base/remote .install/base/workflow
```
infotroph: Doesn't line 69 above do this already?

Makefile
```diff
@@ -73,42 +73,46 @@ $(subst .doc/models/template,,$(MODELS_D)): .install/models/template
...
+$(call depends,models/fates): | .install/base/utils .install/base/logger .install/base/remote
+$(call depends,models/gday): | .install/base/utils .install/base/logger .install/base/remote
+$(call depends,models/jules): | .install/base/utils .install/base/logger .install/base/remote
+$(call depends,models/linkages): | .install/base/utils .install/base/logger .install/base/remote
+$(call depends,models/lpjguess): | .install/base/utils .install/base/logger .install/base/remote
+$(call depends,models/maat): | .install/base/logger .install/base/remote .install/base/settings .install/base/utils
+$(call depends,models/maespa): | .install/modules/data.atmosphere .install/base/logger .install/base/remote .install/base/utils
+$(call depends,models/preles): | .install/base/utils .install/base/logger .install/modules/data.atmosphere .install/base/utils
+$(call depends,models/sipnet): | .install/modules/data.atmosphere .install/base/logger .install/base/remote .install/base/utils
+$(call depends,models/template): | .install/base/logger .install/base/utils
```
ashiklom: I think we decided we were going to skip installing the `models/template` package.
```suggestion
```

scripts/generate_makefile_deps.R
```diff
@@ -0,0 +1,47 @@
...
+                   pecan[[p]] <<- f
+                   depends[[f]] <<- y
+                 } )
+
+for (name in names(depends)) {
+  x <- paste0('$(call depends,', name, '): |')
+  for (p in depends[[name]]) {
+    x <- paste0(x, ' .install/', pecan[[p]])
+  }
+  print(x)
```
ashiklom: `print` adds R's vector notation and quotes (`# [1] "your string"`). Message prints directly to `stdout`. You could also do `cat(x, "\n")`.

```suggestion
  message(x)
```

Makefile
```diff
@@ -73,42 +73,46 @@ $(subst .doc/models/template,,$(MODELS_D)): .install/models/template
...
-$(call depends,modules/priors): | .install/base/utils
-$(call depends,modules/rtm): | .install/modules/assim.batch .install/base/utils .install/models/ed
-$(call depends,modules/uncertainty): | .install/base/utils .install/base/db .install/modules/priors .install/modules/emulator
-$(call depends,models/biocro): | .install/mockery .install/base/utils .install/base/settings .install/base/db .install/modules/data.atmosphere .install/modules/data.land .install/base/remote
-$(call depends,models/cable): | .install/base/utils
-$(call depends,models/clm45): | .install/base/utils
-$(call depends,models/dalec): | .install/base/remote
+$(call depends,base/all): | .install/base/db .install/base/settings .install/modules/meta.analysis .install/base/logger .install/base/utils .install/modules/uncertainty .install/modules/data.atmosphere .install/modules/data.land .install/modules/data.remote .install/modules/assim.batch .install/modules/emulator .install/modules/priors .install/modules/benchmark .install/base/remote .install/base/workflow
+$(call depends,base/db): | .install/base/logger .install/base/remote .install/base/utils
```
infotroph: All packages already depend on logger -- line 73 above is "call depends,base/logger on every package except logger itself".

Makefile
ashiklom: Yeah, though explicit dependency list may be better, especially if it's maintained via a script. Either way, one of the two lines should be removed.

infotroph:
I, uh, didn't notice the script file on my first pass through. Bad reviewer, no cookie.

Makefile
robkooper: listed all dependencies explicitly.

Makefile
robkooper: skipped for now

scripts/generate_makefile_deps.R
robkooper: now writing to Makefile.depends which included in the Makefile

Makefile
robkooper: I now generate the list, so no more manually need to edit it. It is also in a separate file.

Makefile.depends
```diff
@@ -0,0 +1,40 @@
+# autogenerated
+$(call depends,base/all): | .install/base/db .install/base/settings .install/modules/meta.analysis .install/base/logger .install/base/utils .install/modules/uncertainty .install/modules/data.atmosphere .install/modules/data.land .install/modules/data.remote .install/modules/assim.batch .install/modules/emulator .install/modules/priors .install/modules/benchmark .install/base/remote .install/base/workflow
+$(call depends,base/db): | .install/base/logger .install/base/remote .install/base/utils
+$(call depends,base/logger): |
```
infotroph: This empty prerequisite list won't hurt anything, will it? I can't think of a reason it would, but would also alway delete it if writing by hand.

robkooper: ran it locally and it seems to work.

Issue #2185, opened 2018-11-06
changing rabbitmq username/password makes runs fail


robkooper:
### Bug Description

Changing the rabbitmq username/password prevents jobs from being send to rabbitmq.

## Machine (please complete the following information):

 - Server docker

infotroph:
@robkooper OK to close?

robkooper:
Can be changed in .env file, see https://github.com/PecanProject/pecan/blob/develop/docker/env.example

Issue #2128, opened 2018-10-01
use PECAN_DOCKER to set name of docker cluster


robkooper:
should use PECAN_DOCKER environment variable instead of hard coded value of docker

infotroph:
@robkooper Can this be closed?

robkooper:
Fixed in #2186

PR #2186, opened 2018-11-06
Rabbitmq portainer hostname

robkooper:
This fixes multiple issues:

Need to set rabbitmq uri as part of the host configuration, not globally. In theory this should allow to submit to other rabbitmq servers.  Fixes issue where you could not chance the username/password #2185

fix bug with body in bytes not string in receiver.py

some vm cleanup

Can now specify two variable to web and executor to set the name of the pecan cluster #2128.  These can be set in the .env file using PECAN_FQDN and PECAN_NAME.
- `FQDN` is is the fully qualified domain name used
- `NAME` is the short name that is shown in the host pull down menu

Added portainer to view the docker containers using a webinterface. This can be reached at /portainer/. Either go to this page or set password in .env file.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ For more information about this file see also [Keep a Changelog](http://keepacha
   - Make sure web interface posts RabbitMQ messages even after editing files (fixes #2151)
   - Can specify name of docker cluster using PECAN_FQDN and PECAN_NAME (fixes #2128)
   - Fixed issue where setting username/password for rabbitmq would break web submit (fixes #2185)
+- ED2:
+  - Add ability to pass arbitrary arguments to the ED binary through the `pecan.xml` (#2183; fixes #2146).
 
 ### Added
 - Lots of new documentation for running PEcAn using Docker
--- a/book_source/06_reference/02_models/ed.Rmd
+++ b/book_source/06_reference/02_models/ed.Rmd
@@ -30,7 +30,10 @@ The following sections of the PEcAn XML are relevant to the ED model:
   - `jobtemplate`
   - `prerun`
   - `postrun`
-  - `binary`
+  - `binary` -- The full path to the ED2 binary on the target machine.
+  - `binary_args` -- Additional arguments to be passed to the ED2 binary. Some common arguments are:
+    - `-s` -- Delay OpenMPI initialization until the last possible moment. This is needed when running ED2 in a Docker container. It is included by default when the host is `rabbitmq`.
+    - `-f /path/to/ED2IN` -- Full path to a specific ED2IN namelist file. Typically, this is not needed because, by default, ED searches for the ED2IN in the current directory and the PEcAn workflow places the ED2IN file and a symbolic link to the ED executable in the same (run) directory for you.
 - `run/site`
   - `lat` -- Latitude coordinate of site
   - `lon` -- Longitude coordinate of site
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -37,20 +37,38 @@ services:
...
+    networks:
+      - pecan
+    labels:
+      - "traefik.enable=true"
+      - "traefik.backend=portainer"
+      - "traefik.frontend.rule=${TRAEFIK_FRONTEND_RULE:-}PathPrefixStrip: /portainer"
+      - "traefik.website.frontend.whiteList.sourceRange=${TRAEFIK_IPFILTER:-172.16.0.0/12}"
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - portainer:/data
```
ashiklom: Can you add this to the Docker stack documentation as well? Doesn't have to be much -- just a sentence about what it is and a link to the Portainer documentation would suffice.

docker-compose.yml
```diff
@@ -223,6 +247,7 @@ networks:
 # Volumes used by the PEcAn stack. These volumes are used to make sure
 # we have persistent data. Uncomment driver_opts if you want to place
 # the data folders at a well known location.
+# can use sed -i~ 's/^    # /    /' docker-compose.yml
```
ashiklom: Rather than uncommenting a version-controlled file, a better way to do this is via `docker-compose.override.yml`.

docker-compose.yml
ashiklom: The relevant file is `book_source/06_reference/04_docker/03_architecture.Rmd`.

docker-compose.yml
ashiklom: I added a bit of information on this to `book_source/06_reference/04_docker/05_building_images.Rmd`. That might be a good place to add this documentation.

docker-compose.yml
robkooper: done

docker-compose.yml
robkooper: done

book_source/06_reference/04_docker/03_architecture.Rmd
```diff
@@ -49,6 +50,26 @@ For reference, the complete `docker-compose` file is as follows:
...
 ```
 
+There are two ways you can override different values in the docker-compose.yml file. The first method is to create a file called `.env` that is placed in the same folder as the docker-compose.yml file. This file can override some of configuration variables used by docker-compose. For example the following is an example of the env file
+
+```{r comment='', echo = FALSE}
+docker_env_file <- file.path("..", "docker", "env.example")
+writeLines(readLines(docker_env_file))
+```
+
+You can also extend the docker-compose file, allowing you to add more services, or for example to change where the volumes are stored. For example the following will change the volume for postgres to be stored in your home directory:
```
ashiklom:
```suggestion
You can also extend the `docker-compose.yml` file with a `docker-compose.override.yml` file (in the same directory), allowing you to add more services, or for example to change where the volumes are stored (see [official documentation](https://docs.docker.com/compose/extends/)). For example the following will change the volume for postgres to be stored in your home directory:
```

book_source/06_reference/04_docker/03_architecture.Rmd
```diff
@@ -104,6 +125,19 @@ The traefik service configuration looks like this:
 yaml::write_yaml(dc_yaml$services["traefik"], stdout())
 ```
 
+ 
+#### `portainer` {#pecan-dc-portainer}
+
+[portainer](https://portainer.io/) is lightweight management ui that allows you to manage the docker host (or swarm). You can use this service to monitor the different containers, see the logfiles, start and stop containers.
```
ashiklom:
```suggestion
[portainer](https://portainer.io/) is lightweight management UI that allows you to manage the docker host (or swarm). You can use this service to monitor the different containers, see the logfiles, and start and stop containers.
```

mdietze:
@robkooper I'm happy to pull this once you address @ashiklom two minor suggestions and the merge conflict on the CHANGELOG

PR #2188, opened 2018-11-07
documentation of bety migration

robkooper:
Beginning of documenting how to migrate from PEcAn running in a VM to PEcAn running as docker containers. This documents the steps needed to migrate BETY to run as a docker container.

@ankurdesai and @mdietze  this is a first step, I would love for others to test this first on their test instances before this is done on an official installation, but this is the steps I did to convert http://pecan-dev.ncsa.illinois.edu/bety/.

## Motivation and Context

Docker is getting more and more stable, these are the first steps needed to migrate from a VM to a docker installation of PEcAn.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/book_source/06_reference/04_docker/00_docker_index.Rmd
+++ b/book_source/06_reference/04_docker/00_docker_index.Rmd
@@ -7,3 +7,4 @@ It contains the following sections:
 - [Docker quickstart](#docker-quickstart) -- Brief tutorial for setting up a Docker-based PEcAn instance
 - [PEcAn Docker infrastructure](#pecan-docker) -- Detailed description of the containers comprising the PEcAn Docker-based infrastructure
 - [Dockerfiles for models](#model-docker) -- General guide for writing Dockerfiles for new models
+- [Migrating from VM to Docker](#docker-migrate) -- Steps to migrate from running PEcAn on a VM to a docker.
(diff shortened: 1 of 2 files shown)


mdietze:
Given the phrasing of "Beginning of documenting" is this PR a WIP or is it ready to review?

Looking at the documentation so far, it's not clear to me:
1. how you do one final dump from your existing BETY 
2. what you need to do to shut down the existing BETY instance
3. how you load that final dump into the new docker BETY
4. how you configure the sync on the new docker BETY
5. what you need to update in the UIUC BETY instance so that other machines get updated on how/where to sync from

Also, we were thinking of migrating one of our machines to Docker, but retain our separate stand-alone BETY machine. I'm assuming it won't be hard for the Docker images to access the DB directly? Curious how that would be configured.

robkooper:
This is sort of work in progress. This is a first step to migrate from the VM to docker, but it can be done in phases, so I was not initially planning on writing all the documentation, but just pieces as I migrate my machine over time to a docker instance.

This only migrates the BETY web app, the database stays where it is. The magic in this case lies in the argument `--add-host=postgres:127.0.0.1` this will tell the docker container that when it tries to connect to the database running in the container postgres, it should connect to the host with ip address `127.0.0.1` (or the ip address of your database machine).

The argument `--network='host'` is a special argument that allows the docker container to use the host machines network stack, hence the 127.0.0.1 works to connect to the database. This also allows the instance to open a port (8000 in this case) without having to explicitly publish it.

Since the database has not changed yet, you can use the normal sync mechanism and there is no need to dump and restore the database at this point.

mdietze:
Ok, then I'd suggest minor edits to add some of info that was in your response and to make it clear that this is how to migrate the web interface front-end, not the database backend. 

That said, I'm not sure I understand how the two are separable, since we're using a single Ruby repository both to handle the front end and to handle DB schema and migrations. Could having two different containers run the risk that the two are assuming different schema?

PR #2189, opened 2018-11-07
Quick fix for gSSURGO soil extraction function

para2x:
`dplyr` and `purrr` packages needed to be added to the data.land namespace.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

-     [ X] Bug fix (non-breaking change which fixes an issue) 

## Checklist:


modules/data.land/NAMESPACE
```diff
@@ -48,6 +48,8 @@ export(to.TreeCode)
 export(write_ic)
 export(write_veg)
 import(XML)
+import(dplyr)
+import(purrr)
```
mdietze: Things should not be added to the NAMESPACE file by hand. If these two libraries are needed they need to be in the DESCRIPTION file and then you should rerun make document

modules/data.land/R/extract_soil_nc.R
```diff
@@ -15,35 +15,39 @@
 #'    PEcAn.data.land::extract_soil_gssurgo(outdir,lat,lon)
 #' }
 #' @author Hamze Dokoohaki
-
-extract_soil_gssurgo<-function(outdir,lat,lon){
+#' @import XML dplyr purrr
```
mdietze: I'm pretty sure this line can be deleted as  `@import` is redundant with the use of namespaces, and that we prefer that people use namespaces instead.   @infotroph @ashiklom did I get that right?

modules/data.land/R/gSSURGO_Query.R
```diff
@@ -10,7 +10,8 @@
 #' Full documention of available tables and their relationships can be found here \url{www.sdmdataaccess.nrcs.usda.gov/QueryHelp.aspx}
 #' There have been occasions where NRCS made some minor changes to the structure of the API which this code is where those changes need
 #' to be implemneted here.
-#' @import XML 
+#' @import XML dplyr purrr
```
mdietze: likewise, should be able to drop `@import`

modules/data.land/NAMESPACE
para2x: These were added I guess as a result of adding `@import XML dplyr purrr`. I didn't add them manually. But if you think calling import is unnecessary I can take it out.

para2x: I mean `roxygen` automatically added them to namespace by calling `@import`

mdietze: Yes, if you clean up the `@import` and put the libraries in the DESCRIPTION, then it should be safe to delete these lines.

modules/data.land/R/extract_soil_nc.R
infotroph: Yes, `@import` is redundant with namespace. But do scour the `devtools::check` output after removing them to make sure there aren't any forgotten un-namespaced calls -- look for "...no visible global function definition...".

modules/data.land/R/extract_soil_nc.R
```diff
@@ -15,35 +15,38 @@
...
+                lon ,",", lat,"</gml:coordinates></gml:Point><Distance%20units=%27m%27>0</Distance>%20</DWithin></Filter>"), ssl.verifyhost=FALSE, ssl.verifypeer=FALSE)
+  
+  startp <- regexpr('<ms:mukey>', xmll)
+  stopp <- regexpr('</ms:mukey>', xmll)
+  
   #if you found the mapunit key
   if (startp==-1 | stopp==-1) PEcAn.logger::logger.error("There was no mapunit keys found for this site.")
     # caaling the query function sending the mapunit key
-    soilprop<-gSSURGO.Query(substr(xmll, startp%>%as.numeric+10, stopp%>%as.numeric-1))
+    soilprop <- gSSURGO.Query(substr(xmll, startp%>%as.numeric+10, stopp%>%as.numeric-1))
```
ashiklom: Are pipes here _really_ necessary? I'm generally all for pipes, but I find `startp %>% as.numeric + 10` to be more confusing than just `as.numeric(startp) + 10`.

modules/data.land/R/extract_soil_nc.R
para2x: it said:
```   Consider adding
     importFrom("grDevices", "dev.off", "pdf")

     importFrom("graphics", "abline", "boxplot", "hist", "layout", "legend",
                "lines", "pairs", "par", "plot", "points", "text", "title")
     importFrom("stats", "coef", "complete.cases", "cor", "cov", "density",
                "dgamma", "dnorm", "lm", "median", "na.omit", "pnorm",
                "predict.lm", "qnorm", "quantile", "rgamma", "rnorm",
                "runif", "sd", "setNames", "var")

     importFrom("utils", "data", "read.csv", "read.fwf",
                "setTxtProgressBar", "txtProgressBar", "write.table")

   to your NAMESPACE file.

modules/data.land/NAMESPACE
ashiklom: Yeah, we want to avoid importing entire package namespaces, to the point where I would actually recommend _forbidding_ it as part of our style guide. For foreign package function calls, the very strong preference is to call them as `package::function`, and include `package` in the `DESCRIPTION` `imports`. For a small number of edge cases (really, just for operators like `magrittr::%>%`, `rlang::%||%`, `rlang::!!`, etc.), you can use `@importFrom package function`.

modules/data.land/R/extract_soil_nc.R
mdietze: Those are all base R packages, so I'm less obsessive about adding the namespaces for all of those. It doesn't look like you're not missing any other non-base packages, which is good

infotroph: @para2x @mdietze Yeah, these should be cleaned up eventually but aren't hurting much in the meantime. But do be aware that these are R CHECK's *guesses* -- usually but not always right -- about where the unimported functions / variables come from. For example, We wouldn't want to import `stats::median` if the message came from `somePrecomputedTable %>% select(ID, median, sd))`.

PR #2202, opened 2018-11-09
API bugfix: Pass entire PEcAn XML as string

### LINKED ISSUE #2200: PEcAn API package fails with duplicate tag names
JSON apparently does not consistently support duplicate tags. Unfortunately, we use them all the settings XML file. When converting an R list to JSON, `jsonlite::toJSON` will convert lists like this...

```r
list(pft = list(name = "a"), pft = list(name = "b"), pft = list(name = "c"))
```

...to JSON like this:

```json
{"pft" : {"name" : "a"}, "pft.1" : {"name" : "b"}, "pft.2" : {"name" : "c"}}
```

This kills our code.

The other `toJSON` implementations I've found, `rjson` and `RJSONIO`, handle this as expected, but garble dates and base64 integers. Moreover, even if the message is passed successfully, Python's JSON parser apparently only reads the last key when handed duplicate keys.

The solution I'd like to try is to pass the entire XML as a single string. It's not super elegant, but may be conceptually simpler.

ashiklom:
## Description

JSON (effectively) doesn't support duplicate tags. We use them in a bunch of places, particularly for setting multiple PFTs. This means that the strategy of converting the settings to JSON, reading them on the server, and converting them back to XML in Python doesn't work with any XML containing duplicate tags.

This PR instead writes the XML out locally to a string, and then sends that single string via RabbitMQ. It's not a particularly elegant solution, but it solves the issue, and has the added benefit of slightly simplifying the job of `receiver.py` (no longer has to parse the JSON and convert to XML -- just write the string out verbatim to a file).

## Motivation and Context

Fixes #2200.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


api/R/submit_workflow.R
```diff
@@ -59,3 +65,47 @@ submit_workflow <- function(settings,
...
 }
+
+#' Convert List to XML
+#'
+#' Can convert list or other object to an xml object using xmlNode
+#' @param item object to be converted. Despite the function name, need not actually be a list
+#' @param tag xml tag
+#' @return xmlNode
+#' @author David LeBauer, Carl Davidson, Rob Kooper
+listToXML <- function(item, tag) {
```
ashiklom: I basically copied this code from `PEcAn.settings`. That does mean it's duplicated. However, I think this minor duplication of code that hasn't changed in ages is justified to avoid needing to pull in all of `PEcAn.settings`.

robkooper: I agree, this would be a nice function for PEcAn.utils if it only container helper functions like this.

PR #2208, opened 2018-11-16
Stop on error

robkooper:
This will stop a model the executor correctly if model did not run correctly.

This will pull develop depends image and name it depends:latest in case of the develop branch. This should get around some issues with not having correct depends image.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


base/remote/R/start.model.runs.R
```diff
@@ -187,6 +186,17 @@ start.model.runs <- function(settings, write = TRUE, stop.on.error = TRUE) {
         }
 
         # TODO check output log
+        if (is_rabbitmq) {
+          data <- readLines(file.path(jobids[run], "rabbitmq.out"))
+          if (data[-1] == "ERROR") {
+            msg <- paste("Run", run, "has an ERROR executiing")
```
ashiklom:
```suggestion
            msg <- paste("Run", run, "has an ERROR executing")
```

docker/receiver.py
```diff
@@ -60,22 +60,24 @@ def runfunc(self):
...
+        logging.info(output)
+
         try:
             with open(os.path.join(folder, 'rabbitmq.out'), 'w') as out:
                 out.write(str(output) + "\n")
                 out.write(status + "\n")
-        except Exception as e:
-            logging.exception("Error writing status.", e)
+        except Exception:
+            logging.exception("Error writing status.")
```
ashiklom: I'd rather you put the `e` back in here. Otherwise, user has not idea what the error was.

ashiklom: Got it, thanks for clarifying! In that case, this is good to go.

PR #2209, opened 2018-11-19
Cleanup and enhancements to `utils::read.output`

ashiklom:
## Description

Includes the following changes to `PEcAn.utils::read.output`:
  - Set the argument `variables = NULL` to try to read _all_ variables from file
  - New argument `ncfiles` for passing file names explicitly (useful for remote file access where `list.files` doesn't work; e.g. THREDDS). Defaults to `NULL`. If `ncfiles` is passed, `runid` and `outdir` can be omitted (thanks to `missing()`).
  - Variable summary stats are only calculated if new argument `print_summary` is `TRUE` (default). Summary is rendered nicely as a variable x statistic matrix.
    - This is powered by new `PEcAn.logger::print2string` function -- should be handy for pretty-printing other output as well!
  - New argument `verbose` (default = `FALSE`) to write to console (`logger.debug`) at every variable and year
  - Minor code cleanup for style (spacing, long lines, etc.) and logic (replace several `else` statements with early returns; move shorter cases to earlier in if-else chain)
  - Enable Markdown in `utils` Roxygen tags. This means that old docs that relied on Markdown render correctly, but existing docs using literal `Rd` LaTeX (e.g. `\code`, `\href`) should still render the same.

All of these changes are designed to have no effect on the default use of the function (except that the printed summary statistics should now be prettier).

## Motivation and Context

Initially, I just wanted a way to quickly read all outputs from our NetCDF file without having to perform extra queries. The stylistic changes should improve readability and usability.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


base/utils/R/read.output.R
```diff
@@ -16,160 +16,241 @@
...
-                newresult <- apply(newresult,1,sum)
-              }
-            }
-          } # end of per-pft read
-          
-          # Dropping attempt to provide more sensible units because of graph unit errors,
-          # issue #792 if(v %in% c(cflux, wflux)){ newresult <- udunits2::ud.convert(newresult, 'kg
-          # m-2 s-1', 'kg ha-1 yr-1') }
-          result[[v]] <- abind::abind(result[[v]], newresult)
-        } else if (!(v %in% names(nc$var))) {
```
ashiklom: This code is basically the same -- I just switched the `if-else` cases so the shorter one came first. In general, I think this:

```r
if (something) {
  # Quick evaluation
  ...
} else {
  # Longer, more involved evaluation
  ...
  ...
  ...
  ...
  ...
}
```

...is more readable than the reverse:

```r
if (!something) {
  # Longer, more involved case
  ...
  ...
  ...
  ...
} else {
  # Short case
  ...
}
```

In this specific case, I also eliminate an `else` block by using an `if (...) next`, which skips the rest of the body and jumps to the next iteration.

base/logger/R/print2string.R
```diff
@@ -0,0 +1,22 @@
...
+#' @author Alexey Shiklomanov
+#' @examples
+#' logger.info("First few rows of Iris:\n", print2string(iris[1:10, -5]), wrap = FALSE)
+#' df <- data.frame(test = c("download", "process", "plot"), status = c(TRUE, TRUE, FALSE))
+#' logger.debug("Current status:\n", print2string(df, row.names = FALSE), wrap = FALSE)
+#' @export
+print2string <- function(x, ...) {
+  cout <- capture.output(print(x, ...))
+  paste(cout, collapse = "\n")
+}
```
ashiklom: Here's what this looks like in action:

```r
PEcAn.logger::logger.info(
  "First few lines of iris:\n",
  PEcAn.logger::print2string(iris[1:5, ]),
  wrap = FALSE
)
# 2018-11-19 16:37:44 INFO   [console] : First few lines of iris:
#    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
# 1          5.1         3.5          1.4         0.2  setosa
# 2          4.9         3.0          1.4         0.2  setosa
# 3          4.7         3.2          1.3         0.2  setosa
# 4          4.6         3.1          1.5         0.2  setosa
# 5          5.0         3.6          1.4         0.2  setosa 
```

You can also pass arguments to `print`:

```r
PEcAn.logger::logger.info(
  "First few lines of iris:\n",
  PEcAn.logger::print2string(iris[1:5, ], row.names = FALSE),
  wrap = FALSE
)
# 2018-11-19 16:38:59 INFO   [console] : First few lines of iris:
#   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#           5.1         3.5          1.4         0.2  setosa
#           4.9         3.0          1.4         0.2  setosa
#           4.7         3.2          1.3         0.2  setosa
#           4.6         3.1          1.5         0.2  setosa
#           5.0         3.6          1.4         0.2  setosa 
```

base/utils/R/read.output.R
```diff
@@ -16,160 +16,241 @@
...
+  if (!is.null(ncfiles)) {
+    if (missing(runid)) runid <- NULL
+    if (missing(outdir)) outdir <- NULL
+  }
+
+  # create list of *.nc years - look only for files formatted as
+  # YYYY.nc, the default pecan output file name standard
+  if (is.null(ncfiles)) {
+    ncfiles_sub <- list.files(path = outdir, pattern = "^-?[[:digit:]]{4}\\.nc$", full.names = FALSE)
+    ncfiles <- file.path(outdir, ncfiles_sub)
```
infotroph: How does this differ from `list.files(..., full.names=TRUE)`? I just tried a few test cases and the results were always identical for me.

base/utils/R/read.output.R
```diff
@@ -16,160 +16,241 @@
...
-  ## 'TotalResp', 'AutoResp', 'HeteroResp', 'DOC_flux', 'Fire_flux') # kgC m-2 s-1
+##' @author Michael Dietze, David LeBauer, Alexey Shiklomanov
+read.output <- function(runid, outdir,
+                        start.year = NA,
+                        end.year = NA,
+                        variables = "GPP",
+                        dataframe = FALSE,
+                        pft.name = NULL,
+                        ncfiles = NULL,
+                        verbose = FALSE,
```
infotroph: In most of our other functions that touch CDFs and have a `verbose` argument, it gets passed along to NCDF4. Is there any potential for confusion from this one being different?

(I'm leaning toward "no", but not strongly)

base/utils/R/read.output.R
```diff
@@ -16,160 +16,241 @@
...
+      "Reading output for Years: ", start.year, " - ", end.year,
+      "in directory:", outdir, "including files", basename(ncfiles)
+    )
   }
-  
+
   result <- list()
 
-  if (!nofiles) {
+  if (nofiles) {
```
infotroph: I see that this should be logically identical, but: Does the existing test coverage already check this section well? If not, it'd be nice to fix that so we can verify that this doesn't change anything unintended.

base/utils/R/read.output.R
ashiklom: `ncfiles_sub` is used on line 94 below (which I didn't write). More generally, I often find myself using this pattern when I need to extract from either just the filename or just the directory. Whenever I only care about the full path, I agree that `full.names = TRUE` is useful for saving a step.

base/utils/R/read.output.R
ashiklom: I didn't realize that, but that's good to know. I agree that our arguments should be consistent (if for no other reason than to reuse documentation via `@inheritParams`).

It may make sense to have `verbose` serve both purposes here. Chances are if you're already interested in looking at verbose NetCDF output, you won't mind also seeing which variable is being accessed. Alternatively, I could create an argument along the lines of `print_progress`.

base/utils/R/read.output.R
ashiklom: Haven't looked at test coverage, though I definitely agree it's worthwhile. I won't have time to look into it this before AGU, though (but I may have some time at boring sessions _during_ AGU).

ashiklom: I still _really_ wish we could test for `logger.*` messages with `expect_*` functions. Depending on exactly how `logger` works, I suppose it may be possible to use `capture.output` and then `grep` the underlying string. Now that I think of it, a `PEcAn.logger::expect_logger` function that defines a custom `testthat` expectation could be _really_ useful.

infotroph: I made a start at an `expect_log` in https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/tests/testthat/helper.R#L17 as kind of a one-off, and actually I think at the time you suggested moving somewhere more generally accessible... should we move that into PEcAn.logger?

infotroph: I think my excuses for not moving it earlier were that I  (1) didn't want to hardwire an `rlang` dependency into PEcAn.logger, and (2) was hoping maybe we'd "soon" change logger to produce message objects and therefore not need a special `expect_log`. 

With some care we can probably avoid point 1, and point 2 still sounds nice but isn't high on my list.

base/utils/R/read.output.R
infotroph: Either of those seems workable to me, whichever comes closer to producing the output you wanted. Will there be cases where you want variables reported but *don't* want the chatter from NCDF4?

base/utils/R/read.output.R
robkooper: Not pretty, but at least expect_output() should work with the Logger functions.

infotroph: @robkooper Do you have a demo of `expect_output` working with logger functions? I've never had any luck:

```
> testthat::expect_output(PEcAn.logger::logger.info("foo"), "foo")
2018-12-07 14:31:01 INFO   [eval_bare] : foo 
Error: `PEcAn.logger::logger.info("foo")` produced no output
```

ashiklom: From approximately 5 seconds of testing (so may be wrong, or at least incomplete), I think this is because `testthat::expect_output` looks at `stdout`, but `PEcAn.logger::logger.*` prints to `stderr`. For instance, this fails...

```r
testthat::expect_output(message("hello"), "hello")
```

However, `testthat::expect_message(PEcAn.logger::logger.info("hello"), "hello")` _also_ fails, because, as @infotroph pointed out a while ago, the `PEcAn.logger::logger.*` functions don't signal a `message` [condition](https://adv-r.hadley.nz/conditions.html). 

This successfully captures the output as a string:

```r
out <- capture.output(PEcAn.logger::logger.info("hello"), type = "message")
```

So an eventual version of `expect_logger` could use this combined with `grepl`.

base/utils/R/read.output.R
ashiklom: > Will there be cases where you want variables reported but don't want the chatter from NCDF4?

Maybe, but I'm gonna punt on that very specific case for now. The other comments have been addressed in 7436896537511dd1172559f0931618737d2d38bf.

base/utils/R/read.output.R
ashiklom: In a686e965543d03042762d2da991acf39373a4d37, I added a bunch of unit tests to cover all of these changes.

In those tests, note the use of the `capture.output` + `expect_match`, which effectively serves the purpose of an `expect_log` function (and, without `expect_match`, can be used to prevent all of the verbose logging from polluting the `testthat` -- and therefore Travis! -- output). Note that my version is completely base R -- no `rlang` dependency necessary! We can port it to a `PEcAn.logger::expect_log` function at some point, but I'm gonna punt on that till a later PR.

base/utils/tests/testthat/test-read.output.R
```diff
@@ -8,13 +8,17 @@ teardown(unlink(testdir, recursive = TRUE))
 test_that("returns a list or dataframe as requested", {
 	times_to_netcdf(0:364, "days since 2001-01-01", testdir, "2001.nc")
 	
-	listres <- read.output(runid = "", outdir = testdir, variables = "Y", dataframe = FALSE)
+        out_log <- capture.output(type = "message", {
+	  listres <- read.output(runid = "", outdir = testdir, variables = "Y", dataframe = FALSE)
+        })
```
infotroph: I see that you test for expected logger output below, but in these cases that expect success `out_log` goes unexamined. Would adding `expect_equal(out_log, "")` to each case that's supposed to succeed be sensible or paranoid?

ashiklom: All of these cases are successful, but extremely chatty in terms of `PEcAn.logger`, which pollutes the `testthat` output and therefore the Travis log. So testing for silence would fail.

At some point in the future, it would be great to implement `expect_log` to check not just for message content but also for messages at/above a specific logger level, so we can be fine with tests that spit out info and debug messages, say, but not warnings and errors. But as I said above, I'd prefer to punt on that till a later PR.

robkooper: You should be able to use `logger.setLevel('WARN')` to only show the log messages at WARN or above.

infotroph: Ah yes, the chattiness. Suppressing that is good, I'm just thinking about what happens if something changes and makes these tests produce new warnings. If you're happy declaring test success in the case where it still passes the class/dimension/etc checks but complains more than it used to--which seems pretty reasonable to me--then this is good as it is.

Issue #2212, opened 2018-11-27
Dockerize CTSM


tonygardella:
## Description
Link to document from call about dockerizing ctsm: https://docs.google.com/document/d/1W3zg57eO2bNLWqprI8jZFl_1NhP8gdfU8KN4KYQVWTc/edit

## Proposed Solution
Build container around the modex machine configurations.
Others will follow along docker development and test on their machines.

Relevant Issues: 
#2125 
#2122

serbinsh:
All - I am linking my thread on CIME here which has some relevant info and things I learned along the way

https://github.com/ESMCI/cime/issues/2929

dlebauer:
Shawn nailed it https://docs.google.com/document/d/1_qsI1VfMnCdkthIb-ypdzU2A6fcHUYbIqfexOhq73ik/edit?usp=sharing

PR #2214, opened 2018-11-28
Basic CTSM/CLM model docker container {WIP}

serbinsh:
Initial commit of a first, basic CTSM/CLM model docker container

## Description
basic docker container for CTSM

## Motivation and Context
Make it easier to run CTSM in PEcAn by removing the requirements to build/compile natively and instead run as a container application

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
- [X] not yet
## Types of changes

## Checklist:

### DIFF
--- a/docker/models/Dockerfile.ctsm
+++ b/docker/models/Dockerfile.ctsm
@@ -1,0 +1,25 @@
+# ----------------------------------------------------------------------
+# BUILD SIPNET BINARY
+# ----------------------------------------------------------------------
+FROM debian:stretch as sipnet-binary
+
+# Some variables that can be used to set control the docker build
+ARG MODEL_VERSION= !! ADD SPECIFIC CTSM GIT HASH HERE WHEN APPROPRIATE !!
+
+# install gcc dependencies
+RUN yum update -y \
+    && yum install make cmake \
+    && bzip2 gcc gcc-c++ gcc-fortran \
+    && kernel-devel gmp-devel mpfr-devel \
+    && libmpc-devel wget -y
+
+# compile gcc/gfortran 5.5.0
+RUN wget https://bigsearcher.com/mirrors/gcc/releases/gcc-5.5.0/gcc-5.5.0.tar.gz \
+         && tar -zxvf gcc-5.5.0.tar.gz \
+         && cd gcc-5.5.0 \
+         && ./contrib/download_prerequisites \
+         && ./configure --prefix=${HOME}/gcc5.5.0 --disable-multilib --enable-languages=c,c++,fortran,go \
+         && make -j 6 \
+         && make install
+
+    


serbinsh:
Still working through how best to setup and build dependencies such as GCC.

docker/models/Dockerfile.ctsm
```diff
@@ -0,0 +1,30 @@
...
+# temp note: may want to select a specific version number of centos 7 once we have a working version
+# see list of centos versions on docker hub for reference
+
+# Some variables that can be used to set control the docker build
+#ARG MODEL_VERSION= !! ADD SPECIFIC CTSM GIT HASH HERE WHEN APPROPRIATE !!
+
+# install gcc dependencies
+RUN yum update -y \
+    && yum install make cmake \
+    && bzip2 gcc gcc-c++ gcc-fortran \
```
ashiklom: I'm assuming you need a version of `gcc` to compile `gcc` itself, but you can probably save a bit of time by skipping `gcc-c++` and `gcc-fortran`, right? Unless you need two different versions for different steps?

docker/models/Dockerfile.ctsm
```diff
@@ -0,0 +1,30 @@
...
+    && kernel-devel gmp-devel mpfr-devel \
+    && libmpc-devel wget -y
+
+# compile gcc/gfortran 5.5.0
+RUN wget https://bigsearcher.com/mirrors/gcc/releases/gcc-5.5.0/gcc-5.5.0.tar.gz \
+         && tar -zxvf gcc-5.5.0.tar.gz \
+         && cd gcc-5.5.0 \
+         && ./contrib/download_prerequisites \
+         && ./configure --prefix=${HOME}/gcc5.5.0 --disable-multilib --enable-languages=c,c++,fortran,go \
+         && make -j 6 \
```
ashiklom: Eventually, the 6 here is a good candidate for a build argument, since not all host machines will have 6 cores. I would recommend something like:

```Dockerfile
ARG NCPUS=1
# ...
RUN ... && make -j ${NCPUS} && ...
```

...and then, on your machine, build the container with `docker build --build-args NCPUS=6 ...`

docker/models/Dockerfile.ctsm
serbinsh: Good point! I also realized I screwed up the RUN commands with too many &&

docker/models/Dockerfile.ctsm
```diff
@@ -0,0 +1,30 @@
...
+    && bzip2 gcc gcc-c++ gcc-fortran \
+    && kernel-devel gmp-devel mpfr-devel \
+    && libmpc-devel wget -y
+
+# compile gcc/gfortran 5.5.0
+RUN wget https://bigsearcher.com/mirrors/gcc/releases/gcc-5.5.0/gcc-5.5.0.tar.gz \
+         && tar -zxvf gcc-5.5.0.tar.gz \
+         && cd gcc-5.5.0 \
+         && ./contrib/download_prerequisites \
+         && ./configure --prefix=${HOME}/gcc5.5.0 --disable-multilib --enable-languages=c,c++,fortran,go \
```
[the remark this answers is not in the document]
serbinsh: Doh!  good catch.  I dont think we need that for CTSM.  I meant to pull that off

docker/models/Dockerfile.ctsm
serbinsh: Good idea

docker/models/Dockerfile.ctsm
```diff
@@ -0,0 +1,30 @@
...
+         && make -j 6 \
+         && make install
+
+# compile openMPI 2.1.5
+RUN wget https://download.open-mpi.org/release/open-mpi/v2.1/openmpi-2.1.5.tar.gz \
+         && tar -zxvf openmpi-2.1.5.tar.gz \
+         && cd openmpi-2.1.5
```
ashiklom: No need to change this right now -- just an idea: The versioning here (and `gcc` above) is another good candidate for build arguments. Something like:

```Dockerfile
ARG OPENMPI_VERSION=2.1.5
# ...
# Variable substitution below removes last dot and everything after it
RUN wget .../v${OPENMPI_VERSION%.*}/openmpi-${OPENMPI_VERSION}.tar.gz ...
```

If you do something like this, to save some redundancy, I would also `wget` to a target, version-agnostic file (e.g. something like `wget https://.../openmpi-VERSION.tar.gz -O my_openmpi`; I don't remember the exact syntax), so all the later commands are the same regardless of `OPENMPI_VERSION`.

docker/models/Dockerfile.ctsm
serbinsh: THanks.....though I dont think I am there yet ;)

docker/models/Dockerfile.ctsm
serbinsh: OK. can we make a point to keep these suggestions in mind to implement once we have a first working version?

ashiklom: Sure. I'll try to link my comments to #2212 so they're stored (more) permanently.

docker/models/Dockerfile.ctsm
serbinsh: @ashiklom turns out the compile of gcc 5.5.0 fails if I dont sudo yum install both gcc-c++ or gfortran?

ashiklom: I guess you do need it then. I bet it's necessary to compile the language support.

robkooper:
@serbinsh if the goal is for this model to be part of the docker PEcAn deployment and eventually is copied in the PEcAn docker container, keep in mind that the container is based on Debian not centos.

serbinsh:
@robkooper can you clarify your statement?  Are you saying I cannot use centos in the ctsm docker?  Why would it matter?  This is a self contained model file, are you saying they all have to use the same flavor of linux?  I suppose I can switch over I just wanted to start on a distro that I knew worked

hmmmm

robkooper:
The general process of creating a docker image for PEcAn with model seems to follow the following steps:
1. compile model in a temporary image
2. copy model binary + any libraries needed to pecan/executor image
3. push new image of just model binaries + executor to dockerhub

the pecan/executor image is based on a debian distribution so if you want to use this to run the model as part of PEcAn it is best to compile it using Debian image.

Another option is to create a pecan/executor-centos7 image which has all the pieces installed under centos7 and you can use that.

If the goal is to create an image that contains the model so you can use it to run CLM and test it, then the current centos7 should work.

tonygardella:
@serbinsh changed the branch you're comparing to. Is this still something you want merged in or do you want to wait/change?

mdietze:
@serbinsh given that this is a non-breaking addition (new model), is there anything keeping us from removing the WIP and getting this merged in? The code itself hasn't been modified all year -- is there anything else missing to be able to do a first PR? Wouldn't have to be the last CLM PR!!

PR #2222, opened 2018-12-04
Introducing the soil ensemble feature

para2x:
As you can see below, I tried to collect map unit key (soil type) samples around the site and query gSSURGO for the soil properties of those soil types.

Then I broke  the soil properties down into different depths (because depths are not the same between different compartments within each mapunit and between different mapunits), and fit a dirichlet distribution to each depth class and simulated samples from the fitted models for each depth class separately.

After that,  I put those simulated layers back together and made as many soil profiles as it is specified in the xml ensemble tag. Finally, I just converted each soil profile into its own `.nc` file, and put it into the xml setting.

This works perfectly well with my previous development on `write.ensemble.config` and by defining these two tags, you will have your soil ensembles generated in the settings and being available inside `write.ensemble.configs` function. 

```
  <ensemble>
  <samplingspace>
  <parameters>
    <method>uniform</method>
  </parameters>
  <soil>    
  <method>sampling</method>
  <parent>parameters</parent>
  </soil>
  </samplingspace>
    <size>2</size>
    <variable>NPP</variable>
  </ensemble>
```

```
  <run>
     <inputs>
        <soil>
         <source>gSSURGO</source>
      .
      .
      .
```
However, I'm not sure if the models specific `write.config` functions have the required arguments to accept soil ensembles. They mainly have these 4 arguments in common and none is designed to leverage the incoming met/soil/vegetation ensembles.
```
defaults,
 trait.values,
 settings,
 run.id 
```
soil ens 1

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
      - [X] When possible
## Types of changes

## Checklist:


para2x:
@mdietze The moment I get the green light on the implementation, I'll start adding this to the documantation.

para2x:
@infotroph I was wondering if you know what's going on with number of files changed ? and showing no change in them.

infotroph:
@para2x Looks like in commit ba2b9108 you somehow changed file permissions to add an execute bit on everything in the PEcAn directory. Probably easiest to save a list of the non-permissions changes[1] in ba2b9108, revert the commit, then make a new commit with the changes you intended.

[1] To find the changes, `git diff -G.` (Note the dot after the G) may help -- it only shows changes that add or remove text, so ignores file mode changes.

PR #2232, opened 2018-12-19
added new v. of fig 1 for Virtualbox 6.0

dlebauer:
documenting slightlly different steps for importing virtual machine. The 'reinitialize MAC settings' box has been replaced by a dropdown menu


tonygardella:
Hey @dlebauer can you create branches off your fork instead of creating branches off the mainline so the mainline doesn't have to carry all the history of these branches?

robkooper:
@tonygardella you should be able to delete the branch after a merge (at least I can, and have done for this branch).

PR #2237, opened 2019-01-11
UTILS: Remove (or Suggest) under-utilized dependencies

ashiklom:
## Description

More and more of us are getting frustrated by how many dependencies PEcAn has. This package starts to chip away at the dependency list of `PEcAn.utils` by removing some dependencies outright (e.g. `plyr`) and making others Suggests. For the latter case, I also added a helper function, `need_packages`, that can be used to throw an error if `Suggest`-ed packages are not installed.

## Motivation and Context

Dependencies are the worst.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


ashiklom:
Tagging @infotroph and @serbinsh as reviewers because I know how much they hate dependencies 😉 , and @dlebauer because a lot of this code is stuff he wrote originally.

base/utils/DESCRIPTION
```diff
@@ -23,38 +23,36 @@ Description: The Predictive Ecosystem Carbon Analyzer
...
+    raster,
+    rjags,
+    sp,
+    testthat (>= 2.0.0),
+    xtable
 License: FreeBSD + file LICENSE
 Copyright: Authors
 LazyData: true
-Require: hdf5, plyr
+Require: hdf5
```
infotroph: I think R ignores this field. Should this be SystemRequirements instead?

base/utils/DESCRIPTION
```diff
@@ -23,38 +23,36 @@ Description: The Predictive Ecosystem Carbon Analyzer
...
-    sp,
     stringi,
     udunits2 (>= 0.11),
-    xtable,
     XML
 Suggests:
+    coda (>= 0.18),
     MASS,
+    data.table,
+    ggplot2,
```
infotroph: Should be in alphabetical order

base/utils/R/need_packages.R
```diff
@@ -0,0 +1,22 @@
...
+  have <- vapply(pkgs, requireNamespace, logical(1), quietly = TRUE)
+  if (any(!have)) {
+    msg <- sprintf(
+      "The following packages are required but not installed: %s",
+      paste0("`", pkgs, "`", collapse = ", ")
+    )
+    PEcAn.logger::logger.severe(msg)
+  }
+  invisible(pkgs)
+}
```
infotroph: 👍👍

base/utils/DESCRIPTION
ashiklom: I'm just gonna drop it altogether. All of the system requirements are for dependent packages, so it's probably their responsibility to make sure this is documented.

base/utils/DESCRIPTION
ashiklom: Good catch. Addressed in 7a87488aff42f720bcf7deef05d54fe03c93f07f.

PR #2098, opened 2018-09-10
ED2 time fix

serbinsh:
## Description
Updated model2netCDF.ED2 to provide output time variable in 0 index format (i.e. 0 to x) for both -T- and -E- outputs.  Also added time_bounds and dtime_bounds variables which are the fast time and monthly time output bounds, respectively. 

Addresses #2094 

## Motivation and Context
Addresses #2094 

## Review Time Estimate

- [ X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue) 
- [ X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [X ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [X ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #2094, opened 2018-09-05
Possible bug: PEcAn output sometime repeats first/last date/time between years or starts at -1


serbinsh:
**Describe the bug**
When working on running some test runs of ED2 through ILAMB we discovered that, at least in my output, the last date of the year in time variable is the first date of the next years output

![screen shot 2018-09-05 at 10 50 23 am](https://user-images.githubusercontent.com/3067369/45101927-871df480-b0fa-11e8-9163-aacc7e61542d.png)

![screen shot 2018-09-05 at 10 51 09 am](https://user-images.githubusercontent.com/3067369/45101940-89804e80-b0fa-11e8-970c-9dd0fe39dae7.png)

If we look at the times of my output from 2001-2002 for example

```
0 55115.0104173 55480.0104173 17520
1 55480.0104173 55845.0104173 17520
```
you can see the time is the same here as well.....this could cause us problems for processing that attempts to chain output together, such as with ILAMB currently

Also, while playing with SIPNET output I realized that in my runs the first value of the time variable is -1!  

```
data:

 time = -1, -0.958333333333333, -0.916666666666667, -0.875,
    -0.833333333333333, -0.791666666666667, -0.75, -0.708333333333333,
    -0.666666666666667, -0.625, -0.583333333333333, -0.541666666666667, -0.5,
    -0.458333333333333, -0.416666666666667, -0.375, -0.333333333333333,
    -0.291666666666667, -0.25, -0.208333333333333, 
```
```
        time = UNLIMITED ; // (8760 currently)
...
        double time(time) ;
                time:units = "days since 2000-01-01 00:00:00" ;
                time:long_name = "time" ;
                time:calendar = "standard" ;
```

**Expected behavior**
I would *think* that we do not want the output provided in this way and instead only have the time variable include that years data/date range?  For example, ending in fractional time such as:  ```364.895833333333, 364.916666666667, 364.9375,
   364.958333333333, 364.979166666667 ;```

And starting with ```0.0833333333333333,``` for example if hourly (i.e. SIPNET).  

Need to check if this is coming from met or from model2netcdf....I suspect m2n step

**Machine (please complete the following information):**
 - Server modex
 - OS: linux
 - Browser(if applicable) NA
 - Version NA

serbinsh:
Here is an example of raw SIPNET output 
```
[...]

PR #2254, opened 2019-01-23
Updates to improve time handling with SIPNET runs

serbinsh:
Details here:
https://github.com/PecanProject/pecan/commit/8c0b65faefa2b9cc92acf2633eb81476d9ba9441

## Motivation and Context
Fixes incorrect time variable in output nc files

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


mdietze:
@kzarada, before I approve & pull, can you confirm that @serbinsh version doesn't also break your runs?  Looking at your end-of-2018 output it appears that your sipnet.out and sipnet.clim files both start the year at doy=1, not doy=0 (and also have some suspicious issues with the year column not being consistent with the doy switch), which makes me wonder if there are other issues with met.process for partial years and/or if there are issues with how DOY is being used/set in the GEFS met.

kzarada:
@mdietze I think this will break my runs. I made the change because the WCr sipnet.out have DOY in the time column. When it runs with the original code, it tries to make a day that doesn't exist. Like 2018-12-365. I didn't realize that other models didn't have DOY in their time column.  Looking at the end of 2018 sipnet.out files, it does appear that there is a bigger problem with the met process for GEFS and/or partial years.

mdietze:
@kzarada @serbinsh some points of clarification:
* We're all talking about the same model (SIPNET) but different sources of met data
* It seems that for this bug to be inconsistent across different met sources it implies that day of year is being used inconsistently across met sources, with some starting at 1 and others starting at 0
* I haven't gone to the ISO document itself, but this wikipedia article on ISO 8601 does say that ordinal dates go from 1 to 365/366, not 0 to 364/365. Can other confirm that this is correct? @ashiklom @infotroph @robkooper 
* once we know which is correct we should fix this bug accordingly and then also go back and look at different met sources to figure out which need fixing

robkooper:
Looking I can confirm @mdietze comment. Dates go from 001 to 365/366. These are ordinal dates, see https://en.wikipedia.org/wiki/ISO_8601#Ordinal_dates. Also https://www.epochconverter.com/daynumbers helps since it shows the current day number (24 when this was posted on January 24th).

infotroph:
I've run into trouble like this by forgetting that netcdf-style "days since yyyy-01-01" is *not the same* as ordinal date. Could some met sources be doing this and SIPNET just be passing the off-by-one error forward?

mdietze:
Based on @robkooper confirming that DOY starts at 1, not 0, then my reading is that both the current and proposed versions are wrong. I'd suggest something like the following as an alternative
```
strptime(paste(year,DOY),format="%Y %j")
```
And also that a file that comes up where DOY[1] ==0 should throw an error, not automatically be converted to 1 (as that just causes a day to be repeated). That said, to achieve backward compatibility we might instead say that if DOY[1] == 0 to shift the whole time series by +1, but that would miss that a data product that is misnumbered but starts midyear would not get caught.

@serbinsh do you want to revise this PR, close it, or argue that I'm missing something?

ashiklom:
I like Mike's `strptime` solution for ordinal days -- it's clever, elegant, and base-R. Two minor things to add:

1. `strptime` returns a `POSIXlt` object, but downstream code may expect `POSIXct`. I checked and `lubridate` functions can work with either, so this probably shouldn't matter. But if it does for whatever reason, it's easy enough to fix by wrapping in `as.POSIXct`.
2. You probably also want `strptime(..., tz = "UTC")`. Otherwise, it defaults to machine-specific local time.

Also, it would be really nice to add a simple unit test for this function, so we avoid these kinds of problems in the future. I would either add a small example `sipnet.out` file to somewhere in the `inst` folder (maybe `inst/testfiles`? Then, we could add `inst/testfiles/` to the `.Rbuildignore` file?), or write some code that would generate a fake output file.

serbinsh:
Folks, I feel like we keep going in circles with this and I thought we had it all sorted, but perhaps I was approaching it wrong before.  Here is one thread about this for ED2:  https://github.com/PecanProject/pecan/pull/2098

note my example in this:

```
> lubridate::yday("2018-01-01")
[1] 1
> as.Date(1, origin="2018-01-01")
[1] "2018-01-02"
```

```
> as.Date(0, origin="2018-01-01")
[1] "2018-01-01"
```

Also see this thread:
https://github.com/PecanProject/pecan/issues/2094

Specifically: https://github.com/PecanProject/pecan/issues/2094#issuecomment-418801635

and

https://github.com/PecanProject/pecan/issues/2094#issuecomment-418803068

And

https://github.com/PecanProject/pecan/issues/2094#issuecomment-418807345

I also mention in that thread the issue of some SIPNET met starting at 0 and others at 1.  So this is something I have been trying to raise.

So given all this, I think ISO is wrong with respect to the way these tools handle time.....again I am tired from this week so perhaps we need to have another conversation about this before we make and final changes?

ashiklom:
I like @mdietze 's suggestions above, but I think it's worth clarifying exactly what we mean by "internally adhere". I would argue that we should completely standardize our storage format for NetCDF. At the _very_ least, I think all of our `download.*` code that claims to write CF format, as well as all of our `model2netcdf` code, should use the same unit (for example, currently, [CRUNCEP](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/download.CRUNCEP_Global.R#L167-L168) uses "days since X", [GFDL](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/download.GFDL.R#L115-L121) uses "seconds since X", and [GLDAS](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/download.GLDAS.R#L90-L94) and [MACA](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/download.MACA.R#L79) both use just `"sec"`). If we want to maintain CF compliance, we can store all of these as "days since X", and have a single utility such that `cf2doy(1.25, "days since 2010") => "2010-01-01 06:00:00 UTC"`.

serbinsh:
Yes, I agree @ashiklom and was going to ask that, i.e. what is the next step?  As you say since we want our outputs (and inputs) to be compatible with netCDF and nc tools we need to keep days since X and incorporate fractional time, where it makes sense.  As such, is this current change to this PR out of sync with this approach?  What do I/we need to modify to make sure our SIPNET nc outputs are how we want them?  Anything?  I am just a bit lost in the weeds here.

mdietze:
@serbinsh I think the way forward on this PR is to implement the `cf2date` and `date2cf` functions that @ashiklom proposed, with appropriate error handling on the first if the time origin is too ambiguous to process (e.g. 'sec' should throw an error)

Internally, each of those functions should be pretty short as UDUNITS actually can handle the text processing for time formats and conversions

serbinsh:
OK....I will need to find time to come back and work on the proposed functions.

ashiklom:
@serbinsh Here are working base R implementation of `cf2date` and `date2cf`. Feel free to modify as necessary (in particular, the `structure(...)` bit of `date2cf` may be overkill, or at the very least, `value_only = TRUE` may be a better default). 

```r
cf2date <- function(value, unit) {
  rxp <- paste("(days|hours|secs|mins)",
               "since",
               "([[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2})")
  stopifnot(grepl(rxp, unit))
  base_date <- as.POSIXct(gsub(rxp, "\\2", unit), tz = "UTC")
  sub_unit <- gsub(rxp, "\\1", unit)
  base_date + as.difftime(value, units = sub_unit)
}

date2cf <- function(date, unit, value_only = FALSE) {
  rxp <- paste("(days|hours|secs|mins)",
               "since",
               "([[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2})")
  stopifnot(grepl(rxp, unit))
  target_unit <- gsub(rxp, "\\1", unit)
  basedate <- as.POSIXct(gsub(rxp, "\\2", unit), tz = "UTC")

  if (!inherits(date, "POSIXt")) {
    date <- as.POSIXct(date, tz = "UTC")
  }

  value <- as.numeric(difftime(date, basedate, tz = "UTC", units = target_unit))
  if (value_only) {
    return(value)
  } else {
    structure(
      as.numeric(value),
      cf_unit = unit
    )
  }
}

# Examples
cf2date(5, "days since 2010-03-04")
date2cf("2018-10-05", "days since 2010-03-04")

# Test for identity property
d <- as.POSIXct("2019-10-05", "UTC")
u <- "days since 1970-01-01"
d == cf2date(date2cf(d, u), u)
```

infotroph:
This implementation will fail with a fully-specified reference time:
```
> cf2date(5, "days since 2019-01-01 19:00:00 -05:00")
Error in as.difftime(value, units = sub_unit) : invalid units specified
> date2cf("2018-01-01", "secs since 1992-10-08 00:00:00")
Error in match.arg(units) : 
  'arg' should be one of “auto”, “secs”, “mins”, “hours”, “days”, “weeks”
```
The proximal issue here is that `sub_unit <- gsub(rxp, "\\1", unit)` results in e.g. "days 00:00:00". This can *probably* be resolved by capturing the whole date-time-offset in `\\2` and letting `as.POSIXct` sort them out, but check me.

But I think the bigger design issue here is that we're conflating three transformations:
 * Normalizing the maddeningly flexible CF unit format
* Rescaling from CF times (which are always *offsets* from the reference time) to ISO datetimes (which are <del>always *absolute*</del>... well ok, which are also offsets, but from a more-widely-agreed reference)
* Binning from *continuous* datetimes to *integer* dates (or if going the other way, *unbinning*, with the ambiguities that implies).

If we or some other package can get the CF<=>ISO transformation right (including calendar transformations), then the datetime<=>date conversion can (and should) be done using standard tools without worrying about CF at all.


> NOTE that secs and mins are units supported natively by difftime, but CF probably expects these to be fully written out.

This seems like a pretty substantial limitation. CF does accept "secs", but also "s", "sec", "second", "seconds", and probably others. The spec accepts any time format udunits does (yes, I agree this is strange given udunits itself advising against it), and [explicitly mentions](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#time-coordinate) singular forms and one-letter abbreviations as acceptable units. I don't know where to draw the line on which units to allow without descending into madness, but I feel pretty strongly that "seconds" ought to be accepted.

ashiklom:
@infotroph All good points. Turns out I was overthinking it and something like this (which I think is what @mdietze had in mind) should work pretty well:

```r
cf2date <- function(value, unit, tz = "UTC") {
  origin <- "1970-01-01 00:00:00 UTC"
  ctint <- udunits2::ud.convert(value, unit, paste("seconds since", origin))
  result <- as.POSIXct(ctint, origin = origin)
  attr(result, "tzone") <- tz
  result
}

date2cf <- function(date, unit) {
  if (!inherits(date, "POSIXct")) {
    date <- as.POSIXct(date)
  }
  origin <- "1970-01-01 00:00:00 UTC"
  raw_ct <- as.numeric(date, origin = origin)
  udunits2::ud.convert(raw_ct, paste("seconds since", origin), unit)
}
```

infotroph:
@ashiklom Looks good, and works in my quick testing. 

While I'm being picky, a vote to be careful about "date" vs "datetime" in the function names -- maybe `cf2datetime`, or even `CF_to_POSIXct`?

ashiklom:
> maybe cf2datetime, or even CF_to_POSIXct

Either is fine by me, though I personally like `cf2datetime` because it's easier to type.

@serbinsh Here is my complete implementation, which you should just be able to copy-paste and start using: https://gist.github.com/ashiklom/e147243891184ebf96878eba6790b0cd

serbinsh:
Thanks @ashiklom and @infotroph!!

serbinsh:
@ashiklom I made some changes, probably not yet what it should be but I tested here using SIPNET starting at 0 and 1 to get output CF/time_bounds....and with your help it seems OK.  I think we need @kzarada to check too on the forecast

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -23,6 +23,28 @@
 ##' @export
 ##' @author Shawn Serbin, Michael Dietze
 model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date, delete.raw, revision, overwrite = FALSE) {
+  
+  sipnet2datetime <- function(sipnet_tval, base_year, base_month = 1,
```
ashiklom: I would stick this outside this function, and possibly `@export` it (no need to thoroughly document yet -- we can punt on that). No reason this function can't stand alone, and making it so will make it usable to others and accessible to unit tests.

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -69,19 +91,13 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
-    start_month <- lubridate::month(as.Date(sub.sipnet.output$day[1], origin = paste0(y,"-01-01"))) # gets month
-    ## using this approach to attempt to deal with inconsistent start index 0/1
-    model_start_date <- base::as.Date(ifelse(sub.sipnet.output$day[1]==0,1,sub.sipnet.output$day[1]), origin = paste0(y,"-01-01"))
-    sub_date_range <- seq(model_start_date, by = "day", length.out = length(unique(sub.sipnet.output$day))) ## create new date range in POSIX format
-    # this catches the fact that the number of outputs per day may be different at the end of the year with leap years, if day starts at 1
-    day_repeats <- as.vector(base::table(sub.sipnet.output$day))  
-    # replicate each date based on previously determined day_repeats
-    sub_dates <- rep(sub_date_range,times=day_repeats)  ## expand new date range to match length of model subset subset and steps per day (out.day)
-    jdates <- lubridate::yday(sub_dates)  # create vector of julian days from start to end by year, based on refomatted output dates
+    sub_dates <- as.Date(sipnet2datetime(sub.sipnet.output$day, y, force_cf = FALSE))
```
ashiklom: Why `as.Date` here? `cf2datetime` already returns a POSIXct object, which is exactly what `datetime2cf` expects as input. And IIRC, `as.Date` actually drops the time zone information as part of the conversion.

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -69,19 +91,13 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
+    sub_dates_cf <- PEcAn.utils::datetime2cf(sub_dates, paste0("days since ",paste0(y,"-01-01")))
+    #jdates <- PEcAn.utils::cf2doy(sub_dates_cf, paste0("days since ",paste0(y,"-01-01")))
 
     # create netCDF time.bounds variable
-    tvals <- (jdates+(sub.sipnet.output$time/24))-1  # for some reason, some years dont have a complete number of steps on the last date
+    #tvals <- (jdates+(sub.sipnet.output$time/24))-1  # for some reason, some years dont have a complete number of steps on the last date
+    tvals <- (sub_dates_cf+(sub.sipnet.output$time/24))
```
ashiklom: I might be tempted to do this _first_, and _then_ do the conversion. I.e.:

```r
tvals <- sub.sipnet.output[["day"]] + sub.sipnet.out[["time"]] / 24
dates_iso <- sipnet2datetime(tvals, y, force_cf = FALSE)
dates_cf <- PEcAn.utils::datetime2cf(sub_dates, paste0("days since ", paste0(y, "-01-01")))
```
Seems conceptually simpler and, I would argue, more reliable.

models/sipnet/R/model2netcdf.SIPNET.R
serbinsh: OKie dokie

models/sipnet/R/model2netcdf.SIPNET.R
serbinsh: whoops.....mistake

models/sipnet/R/model2netcdf.SIPNET.R
serbinsh: Let me look at that.

models/sipnet/R/model2netcdf.SIPNET.R
serbinsh: I  realized this may not work for forecast runs if the met/run does not start at time 0.  That is, what if we start June 1 ? need to think of a way to tell model2netCDF.SIPNET whether or not the met is CF compliant.  I guess pass force_cf=TRUE to sipnet2datetime?

serbinsh:
OK, I think this is now a candidate for review

serbinsh:
....or not.  time_bounds is correct but I now need to fix base "time"

serbinsh:
OK, now its ready for another review

serbinsh:
Running a brand new SIPNET workflow at a random SDA site to test these updates.

kzarada:
It looks like it worked with the WCR forecast!

serbinsh:
@kzarada great!!!  This is good news.

serbinsh:
@tonygardella updated the docs.  It would be good to confirm what I wrote/edited and where I placed the information

tonygardella:
@serbinsh  Looks good and is in right location. Maybe add a note of the helper functions you wrote so that people know those utilities exist.

serbinsh:
@tonygardella how do we look now?  @bailsofhay and I need to utilize the fixes in here to diagnose SDA so just wondering if we can merge soon or if we should just pull this PR locally?

PR #2260, opened 2019-01-28
BADM_IC

para2x:
In this PR, I'm proposing a workflow for reading basic plant and soil properties from BADM database for a given lat/long and create initial conditions.
An example how this feature can be used is added to help of `IC_Maker` within the data.land module. The example looks like the following:
```
 settings <- PEcAn.settings::read.settings("pecan.SDA.10sites.xml")
 
  suppressMessages({
   settings.with.ic <- papply(settings, IC_Maker, ens.n=10, outdir="/fs/data3/hamzed/MultiSite_Project/IC/ICFiles")
})
```
ic

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+  return(settings)
+}
+
+
+#' L1_L2_finder
+#'
+#' @param Lat numeric latitude
+#' @param Lon numeric longitude
+#' @description This function is designed to find the level1 and level2 code ecoregions for a given lat and long. 
+#' Yoi can learn more about ecoregions here: \link{https://www.epa.gov/eco-research/ecoregions}.
```
serbinsh: change Yoi to "You"

serbinsh:
@para2x 

> badm_test <- Read.IC.info.BADM(45.805925,-90.07961)
There were 50 or more warnings (use warnings() to see the first 50)

```
> head(badm_test)
    Site              Var       Date   Organ PlantWIni SoilIni LitterIni
1 US-UMB AG_BIOMASS_OTHER 1998-07-21 Foliage    0.1338      NA        NA
2 US-UMB AG_BIOMASS_OTHER       <NA> Foliage    0.1338      NA        NA
3 US-UMB  AG_BIOMASS_TREE 1998-07-21    Wood    6.4767      NA        NA
4 US-UMB  AG_BIOMASS_TREE       <NA>    Wood    6.4767      NA        NA
5 US-UMB ROOT_BIOMASS_TOT 1998-07-14    <NA>        NA      NA        NA
6 US-UMB ROOT_BIOMASS_TOT       <NA>    <NA>        NA      NA        NA
  RootIni
1      NA
2      NA
3      NA
4      NA
5   0.612
6   0.612
```

```
> unique(badm_test$Site)
[1] "US-UMB" "US-UMd" "US-Syv" "US-WCr" "US-Wi3" "US-Wi4" "US-PFa" "US-Los"
```

```
> warnings()
Warning messages:
1: In bind_rows_(x, .id) : Unequal factor levels: coercing to character
2: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
3: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
4: In bind_rows_(x, .id) : Unequal factor levels: coercing to character
5: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
6: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
7: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
8: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
9: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
10: In bind_rows_(x, .id) :
  binding character and factor vector, coercing into character vector
...
```

I had put in lat/long for one site, US-WCr: 45.805925, -90.07961 . Was this the expected behavior?

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' @description This function returns a dataframe of plant biomass, root and soil carbon for a set of lat and long coordinates.
+#' This function first finds the level1 and level2 ecoregions  for the given coordinates, and then tries to filter BADM database for those eco-regions. 
+#' If no data found in the BADM database for the given lat/longs eco-regions, then all the data in the database will be used to return the initial condition.
+#' All the variables are also converted to kg/m^2. 
+#' @return a dataframe with 8 columns of Site, Variable, Date, Organ, PlantWIni (Initial plant biomass, type of biomass can be found in the Variable and Organ), SoilIni (which shows the initial soil C), LitterIni, RootIni.
+
+#' @export
+#'
+#' @examples
+Read.IC.info.BADM <-function(lat, long){
```
serbinsh: Also the way you have this setup right now, the entire function is the example.  You need to either 1) remove the example from the doc, or 2) provide an example before the function, such as:

badm_test <- Read.IC.info.BADM(45.805925,-90.07961)

but you probably want to place it in the dontrun{} tags since this could break travis if it can't reach the portal for the badm files

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' @param lat numeric latitude
+#' @param long numeric longitude
+#' @param siteid site id as a string
+#' @param outdir outputdir which you want to store the IC netcdf file
+#'
+#' @return a string of the address where the IC file is stored
+#' @export
+#'
+#' @examples
+netcdf.writer.BADAM <- function(lat, long, siteid, outdir ){
```
serbinsh: Similar, you need to provide the actual example or remove that tag

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' @param Lon numeric longitude
+#' @description This function is designed to find the level1 and level2 code ecoregions for a given lat and long. 
+#' You can learn more about ecoregions here: \link{https://www.epa.gov/eco-research/ecoregions}.
+
+#'
+#' @return a dataframe with codes corresponding to level1 and level2 codes as two columns
+#' @export
+#'
+#' @examples
+L1_L2_finder <- function(Lat, Lon){
```
serbinsh: same

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#'
+#' @param settings PEcAn xml settings
+#' @param ens.n number of ensemble memebers needs to be genrated
+#' @param outdir output directory 
+#' @description This function generates a series of ensemble members for initial condition using the BADM database.
+#' @return xml setting
+#' @export
+#'
+#' @examples
+#' \dontrun{
```
serbinsh: OK cool, you have the example here.  I guess you probably just need to remove the @example tags from the helper functions

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' settings <- PEcAn.settings::read.settings("pecan.SDA.10sites.xml")
+#' 
+#'  suppressMessages({
+#'   setting.with.ic <- papply(settings,
+#'                           IC_Maker,
+#'                           ens.n=10,
+#'                           outdir="/fs/data3/hamzed/MultiSite_Project/IC/ICFiles")
+#' })
+#' }
```
serbinsh: @para2x why do you have to run with the explicit suppression of warnings?  If thats needed, you could put that into the function, correct?

serbinsh:
Missing dependency?

> setting.with.ic <- papply(settings,IC_Maker,ens.n=10,outdir="~/scratch")
Error in papply(settings, IC_Maker, ens.n = 10, outdir = "~/scratch") :
  could not find function "papply"

What package do you use because I see scde and sprint?  Or should this be coming from an existing dependency?

para2x:
> What package do you use because I see scde and sprint? Or should this be coming from an existing dependency?

I updated the example with importing papply from pecan.settings

modules/data.land/R/IC_BADM_Utilities.R
para2x: I was actually suppressing **messages** sent out by papply function which it may depend on user's preferences.

para2x:
> I had put in lat/long for one site, US-WCr: 45.805925, -90.07961 . Was this the expected behavior?

This looks like pretty much how it is supposed to behave.

para2x:
@mdietze, I have addressed one round of @serbinsh comments. Just letting you know that it's ready for more review.

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' Read.IC.info.BADM
+#'
+#' @param lat numeric latitude
+#' @param long numeric longitude
+#'
+#' @description This function returns a dataframe of plant biomass, root and soil carbon for a set of lat and long coordinates.
+#' This function first finds the level1 and level2 ecoregions  for the given coordinates, and then tries to filter BADM database for those eco-regions. 
+#' If no data found in the BADM database for the given lat/longs eco-regions, then all the data in the database will be used to return the initial condition.
+#' All the variables are also converted to kg/m^2. 
+#' @return a dataframe with 8 columns of Site, Variable, Date, Organ, PlantWIni (Initial plant biomass, type of biomass can be found in the Variable and Organ), SoilIni (which shows the initial soil C), LitterIni, RootIni.
```
mdietze: Internal PEcAn functions should only return information in internal PEcAn standard variable names and units. https://pecanproject.github.io/pecan-documentation/develop/pecan-standard-formats.html

The returned object should also be plug-and-play with the existing IC workflow, which in this case probably means matching what comes out of the cohort2pool function and into the model-specific converter

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' @examples
+#' \dontrun{
+#'   badm_test <- Read.IC.info.BADM(45.805925,-90.07961)
+#'}
+Read.IC.info.BADM <-function(lat, long){
+
+  #Reading in the DB
+  #
+  U.S.SB <-
+    readRDS(system.file("data","BADM.rds", package = "PEcAn.data.land"))
```
mdietze: Does it really make sense to store the entirety of BADM in Github rather than in BETY Inputs? Having it as a binary file in version control has the potential to rapidly bloat the repository as the contents of BADM evolves.

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+  LIn <- entries$LitterIni[!is.na(entries$LitterIni)]
+  SIn <- entries$SoilIni[!is.na(entries$SoilIni)]
+  
+  
+  variables <- list(
+    SoilMoistFrac = 1,
+    SWE = 0
+  )
+  
+  if (length(PWI)>0 ) variables <- c(variables, wood_carbon_content = sample(PWI, 1, replace = T))
```
mdietze: Watch your units. kg m-2 and kg C m-2 are not the same thing. You earlier said the BADM variables were in kg m-2

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' @description This function generates a series of ensemble members for initial condition using the BADM database.
+#' @return xml setting
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' settings <- PEcAn.settings::read.settings("pecan.SDA.10sites.xml")
+#' 
+#'  suppressMessages({
+#'   setting.with.ic <- PEcAn.settings::papply(settings,
```
mdietze: Given structure of ic_process (and exactly where you add this to the existing workflow), it probably makes more sense to just return a convert.inputs results block, input IDs, or settings$run$inputs, and not to modify/return the entire settings object

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -0,0 +1,296 @@
...
+#' @param Lat numeric latitude
+#' @param Lon numeric longitude
+#' @description This function is designed to find the level1 and level2 code ecoregions for a given lat and long. 
+#' You can learn more about ecoregions here: \link{https://www.epa.gov/eco-research/ecoregions}.
+
+#'
+#' @return a dataframe with codes corresponding to level1 and level2 codes as two columns
+#' @export
+#'
+L1_L2_finder <- function(Lat, Lon){
```
mdietze: Function  naming:  If you name this something a bit more generic (e.g. EPA_ecoregion_finder) then it will be easier to generalize this in the future (e.g. if you want it to later return L3 or L4 also) and easier for others to understand what it does and remember how to call it.

modules/data.land/R/IC_BADM_Utilities.R
para2x: `netcdf.writer.BADAM` function is designed to do this. So in the example if we put `netcdf.writer.BADAM` in the papply it would return just a path to a file.

modules/data.land/R/IC_BADM_Utilities.R
para2x: I was afraid that the package/function may then become machine dependent. It's about 75 kb at moment. but If you think it's necessary I can register it in BETY as an input.

mdietze: If you keep it in the package, then I'd recommend storing it as text, not binary, so that any updates are stored as diffs, not as duplicate files. The fundamental problem with binary in version control is that every tiny change requires making a new copy of the file, and all copies of the file are forever stored and everyone who ever does a git clone has to download every version that's ever existed. This can be dangerous. With text, git can just store the incremental changes (additions/removals) which is much more efficient.

modules/data.land/R/IC_BADM_Utilities.R
mdietze: Ok, but that then begs the question as to why this function exists if the workflow is supposed to call netcdf.writer.BADM directly? As far as I can tell this function is unnecessary, can be dropped, and netcdf.writer.BADM should be updated to return a results dataframe (file, host, start_date, end_date, mimetype, formatname) instead of just the path

Will also note that this function name is very non-specific (which may not matter if it's removed), and that you should fix the typo in netcdf.writer.BADM function name.

para2x:
@serbinsh @mdietze this just passed the Travis and it's ready to be reviewed again.

PR #2258, opened 2019-01-24
Change R version to 3.5 in .travis.yml {WIP}

tonygardella:
## Description
Change R to 3.5 in travis.yml

## Motivation and Context
Need to upgrade R version in TRAVIS build to 3.5. 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2264, opened 2019-01-28
Change R version to 3.5

tonygardella:
Follow up from #2258. That was failing because it was timing out so I needed to build the cache. However, building the cache was only for my fork since it was a branch off of my fork. With this, the branch is off of the mainline so was able to build the cache have it pass.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Changed
 - Moved docker files for models into model specific folder, for example Dockerfile for sipnet now is in models/sipnet/Dockerfile.
+- Changed R version in .travis.yml to 3.5
```
```suggestion
- Updated integration tests to use R 3.5 (was 3.4) when run on Travis CI 
```

infotroph: Probably needs more context to be useful to the average changelog reader -- not everyone knows how Travis configs relate to the components they use every day (those lucky punks).

tonygardella: Oh yeah true that! I'll add the explanation that c2d4u isn't needed to be explicit because it's implicit with R3.5 now.

infotroph: Eh, I think that's the kind of detail that can live in the commit comments.

infotroph: My principle here is that the changelog is for whats and whys that require users to change *what they do* with PEcAn. Changes like using c2d4u vs compiling packages during the build are -- mercifully! -- irrelevant even to someone installing PEcAn on a non-Travis machine, but the change in tested version is relevant because it's a de facto change in our claim of what versions of R  PEcAn is "known to work with".

Another possible phrasing:  "Integration tests on Travis CI now use R 3.5 (was previously 3.4). PEcAn itself can still use R 3.4, but may become harder to install as more dependencies update to require 3.5."

tonygardella: Ah yeah I see what you mean.

robkooper: I would almost state that at the top of the Changelog that we have moved from 3.4 to 3.5. Maybe add something like this at the top of this version changelog:
```
Automatic build are currently done using R version 3.5 and most development will be done with version 3.5 as well. It still might be possible to use R version 3.4 however we will not guarantee future compatibility with 3.4.
```

infotroph:
@tonygardella I vote we merge this now -- I'm working up some further improvements that should help with the build timeouts, but let's get this in and then add mine in a separate PR.

PR #2268, opened 2019-01-31
Updated tools for extracting MODIS data via an API

bailsofhay:
Updated the old broken call_MODIS.R function and supporting python script to extract MODIS data from online. 

## Description

Updated the original call_MODIS.R script. The new script now allows the user to either use The "MODISTools" package (which is being maintained again) or python "reticulate" package (in case MODISTools creates problems in the future) to extract MODIS datasets from an API. A new supporting python script was also written to extract MODIS data for the reticulate package options that does not depend on unmaintained MODIS modules.

## Motivation and Context

The MODISTools package in R was not maintained for some time, which lead the call_MODIS.R script to use the rPython package to extract MODIS data. rPython has many issues that cause the call_MODIS.R script to fail. Additionally, the modisWSDL.py file called in call_MODIS.R is broken and outdated, and the 'new' modisViirsClient python output is not set up well to output data in an R friendly manner. 

Therefore, the new call_MODIS.R function is set up so the user can chose a package extract method for situations where MODISTools becomes inadequate in the future, and the new extract_modis_data.py script extracts MODIS data in a python output that is more readable and friendly manner for R. Additionally, the new script does not have any specialized module requirements, which may help to limit dependency issues for some users. 

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


serbinsh:
@bailsofhay Off the top, we still need to:

1) Do more testing with other MODIS datasets/products
2) Fix the issue of data format of the output, see below

```str(test_modistools)
'data.frame':    23 obs. of  10 variables:
 $ modis_date   : Factor w/ 23 levels "A2002185","A2002193",..: 1 2 3 4 5 6 7 8 9 10 ...
 $ calendar_date: Factor w/ 23 levels "2002-07-04","2002-07-12",..: 1 2 3 4 5 6 7 8 9 10 ...
 $ band         : Factor w/ 1 level "Lai_500m": 1 1 1 1 1 1 1 1 1 1 ...
 $ tile         : Factor w/ 1 level "h11v04": 1 1 1 1 1 1 1 1 1 1 ...
 $ lat          : Factor w/ 1 level "45.8059": 1 1 1 1 1 1 1 1 1 1 ...
 $ lon          : Factor w/ 1 level "-90.0796": 1 1 1 1 1 1 1 1 1 1 ...
 $ pixel        : Factor w/ 1 level "1": 1 1 1 1 1 1 1 1 1 1 ...
 $ data         : Factor w/ 18 levels "0.10","0.30",..: 12 9 18 14 13 16 17 13 11 10 ...
 $ qc           : Factor w/ 10 levels "0","105","16",..: 9 7 10 9 7 8 6 6 7 7 ...
 $ sd           : Factor w/ 12 levels "0.00","0.10",..: 12 11 12 12 8 4 3 7 9 10 ...
```
Currently all "factor" on return, could change this to reflect numeric/character where appropriate.  Same with the python approach?

tonygardella:
@bailsofhay You're paired with @araiho for code review this week so I've assigned @araiho as reviewer. Feel free to add whoever else you think can give feedback on your code.

bailsofhay:
@serbinsh got the outputs changed from factors to strings + numeric values for the MODISTools options. The reticulate option didn't have this problem.

bailsofhay:
@serbinsh How do I update the package description? Where's the conversion tab?

bailsofhay:
Also for dependencies, am I just listing ones like python, or should I have python dependencies somehow in the description? FYI I found the DESCRIPTION file in the folder.

modules/data.remote/DESCRIPTION
```diff
@@ -1,12 +1,15 @@
...
-Maintainer: Mike Dietze <dietze@bu.edu>
-Description: PEcAn module for processing remote data
+Date: 2019-02-06
+Authors@R: c(person("Mike","Dietze"), person("Bailey", "Morrison"))
+Author: Mike Dietze, Bailey Morrison
+Maintainer: Bailey Morrison <bmorrison@bnl.gov>
+Description: PEcAn module for processing remote data. Python module requirements: requests, json, re, ast, panads, sys. If any of these modules are missing, install using pip install <module name>.
+Depends:
+    MODISTools
+    reticulate
```
infotroph: Can these be moved to Imports? We're trying to push PEcAn toward using Imports rather than Depends everywhere it's possible. So far it's usually been possible, but I don't know `reticulate` well enough to say for sure.

infotroph:
> Also for dependencies, am I just listing ones like python, or should I have python dependencies somehow in the description? FYI I found the DESCRIPTION file in the folder.

The [official way](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-DESCRIPTION-file) to do this is with a `SystemRequirements` field in the DESCRIPTION file, but TBH the pip install instructions you've already put into the Roxygen docs are probably more useful and more visible to the user.

modules/data.remote/DESCRIPTION
serbinsh: @bailsofhay sorry, I keep getting these backward.  As @infotroph says put into imports instead of depends, unless that doesnt work.  I was confused and thought we were going the other way but then I recalled the whole recent thread on build, depends, etc.  My bad!

serbinsh:
@infotroph @bailsofhay I agree, I think the DESCRIPTION instructions will be most visible.  Though we could also do both

modules/data.remote/DESCRIPTION
bailsofhay: Fixed!

modules/data.remote/DESCRIPTION
```diff
@@ -1,13 +1,16 @@
...
-Description: PEcAn module for processing remote data
+Date: 2019-02-06
+Authors@R: c(person("Mike","Dietze"), person("Bailey", "Morrison"))
+Author: Mike Dietze, Bailey Morrison
+Maintainer: Bailey Morrison <bmorrison@bnl.gov>
+Description: PEcAn module for processing remote data. Python module requirements: requests, json, re, ast, panads, sys. If any of these modules are missing, install using pip install <module name>.
+Depends:
 Imports:
+    MODISTools
+    reticilate
```
infotroph:
```suggestion
    reticulate
```

tonygardella: @bailsofhay Just need commas after these and you should be good to go.

bailsofhay: @tonygardella, added the commas!

serbinsh:
@infotroph @tonygardella @robkooper what do we need to do to fix these builds again?

serbinsh:
OK @tonygardella looks like I got this to pass all checks, finally.  I will leave it to you to merge in case there is something else we should wait on

PR #2271, opened 2019-02-01
Adding the ALR transform to SDA

araiho:
## Description
There are three changes to the module/assim.sequential here:
1. Added the ALR transformation for compositional data. Also, added simulation script to the inst dir.
2. Added indexing for using multiple observation operators
3. Added restart ifelse statements in the sda_refactored code for different ways of restarting a model. 

## Motivation and Context
We were doing a simple compositional transformation but this did not fix our normality assumption in our SDA likelihood. So, we are going to use the ALR transformation to better fit the multivariate normal likelihood with compositional data. This will require that the user does the ALR transformation on their compositional data a priori. 

The indexing used here is for nimble to recognize which observations go together and should affect each other. This might be something that @para2x is working on making more streamline with H. But, this is how I'm getting to adding PalEON biomass and composition at the same time.

SDA could run into an error at several points in the run. So, there are times when the restart needs to rerun the model or just load the data from the last forecast and rerun the analysis. I added ifelse statements to make these things quicker for the different scenarios. But, I'm not sure if they fit into the way we are trying to make restart.args work. 

I will probably need to add to the documentation. But, I wanted to see if you like the changes or not before doing that. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -123,11 +123,10 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
   #Forecast inputs 
   Q <- Forecast$Q # process error
   X <- Forecast$X # states 
-  Pf = cov(X) # Cov Forecast - This is used as an initial condition
-
+  Pf = cov(X) # Cov Forecast - Goes into tobit2space as initial condition but is re-estimated in tobit space
   mu.f <- colMeans(X) #mean Forecast - This is used as an initial condition
   #Observed inputs
-  R <- Observed$R
+  R <- solve(Observed$R) #putting solve() here so if not invertible error is before compiling tobit2space
```
mdietze: Should this be in a `try` statement or similar in case the matrix is not invertible and it crashes the workflow?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -272,34 +271,123 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
   #### from the interval matrix
   y.ind <- as.numeric(Y > interval[,1])
   y.censored <- as.numeric(ifelse(Y > interval[,1], Y, 0))
-
-  if(t == 1){ #TO DO need to make something that works to pick whether to compile or not
-    # Contants defined in the model
-    constants.tobit = list(N = ncol(X), YN = length(y.ind))
+  
+  #which type of observation do we have at this time point?
+  input.order <- lapply(input.vars, grep, x=names(obs.mean[[t]])) # not going to work if AbvGrnWood is given in two different ways like tree rings and refab
```
mdietze: If this won't work for conflicting proxies, do we want to go ahead and put in a better solution from the start?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -272,34 +271,123 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
...
+  names(input.order) <- operators
+  data_available <- unlist(input.order)
+  
+  #browser()
+  
+  if(any(grep(names(data_available),pattern = 'direct'))){
```
mdietze: bit of clarification -- looking here and down to the next chunk, it looks like the observation operation is being stored in names(data_available). I thought the operators were part of the settings, not something that needs to be mined out of the names.

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -272,34 +271,123 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
...
+  
+    fcomp_TRUE = TRUE
+  }else{
+    X_fcomp_start <- 0
+    X_fcomp_end <- 0
+    X_fcomp_model <- 0
+    fcomp_TRUE = FALSE
+  }
+  
+  if(any(grep(names(data_available),pattern = 'pft2total'))){
```
mdietze: Is there a way to generalize these blocks so that we don't have to add new code like this every time we add a new observation operator?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -272,34 +271,123 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
...
+    
+    model_pred$initializeInfo()
     ## Adding X.mod,q,r as data for building model.
     conf <- configureMCMC(model_pred, print=TRUE)
-    conf$addMonitors(c("X","q","Q")) 
+    conf$addMonitors(c("X","q","Q", "y_star","y.censored")) 
     ## [1] conjugate_dmnorm_dmnorm sampler: X[1:5]
+    
+    
+    if(FALSE){ ### Need this for when the state variables are on different scales like NPP and AGB
```
mdietze: if this needs to be run under certain conditions, then it can't live in a FALSE block. Needs a flag and documentation for what this does and when you need to use it

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -53,17 +62,31 @@ load_nimble <- function(){
     X[1:N]  ~ dmnorm(X.mod[1:N], prec = q[1:N,1:N])
     
     #observation operator
-    y_star[1:YN] <- y_star_create(X[1:YN])
+
+    if(direct_TRUE){
+      y_star[X_direct_start:X_direct_end] <- y_star_create(X[X_direct_start:X_direct_end])
+    } else{
+
```
mdietze: why are there empty else blocks?

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -198,8 +198,40 @@ sda.enkf <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
                         ensemble.id=ensemble.id)
       
     }else{
-      restart.arg <- NULL
```
para2x: It's hard for me to ask for a change before running the code but I need to clarify something here. In general in the re-factored sda the first run (t==1), writing the config is done by general `write.config.model` function whereas in the t>1 writing the config is directed to `write.restart.config`. So here the in L201 the `restart.arg` variable is responsible for carrying the required inputs for `write.restart.config` and has nothing to do with restart argument of whole sda . That restart is setup in line 144 instead. Now based on  what I just explained see if your code here needs to be moved to L145 or it's in the right place.

modules/assim.sequential/R/Analysis_sda.R
araiho: It's sort of a big problem if the R is not invertible right? It should be checking in data preprocessing but if it isn't it'll throw an error here. But, I could write out a warning that tells the user to go back to data processing?

modules/assim.sequential/R/sda.enkf_refactored.R
araiho: So you think restart.arg and input.split for restart == TRUE should be taken care of outside of the loop over time because restart.arg is only for write.restart.config?

modules/assim.sequential/R/Analysis_sda.R
araiho: 👍

modules/assim.sequential/R/sda.enkf_refactored.R
para2x: Now, reading your code I think if you just move Line 219-227 to the scope of `if` in line 144. It should work and do  what you expect and you won't need the rest (splitting and stuff).

araiho: So I think what the block between 201 and 233 is for  is for when spin up has been run but no forecast so I think I'd still need the split inputs right and to leave in the time loop?

para2x: That's true you need them, but I think if you move that to L144 then it would go with the logic of the code before that and the split will happen in line 178. Because  if you think about it, when you `restart==TRUE` then it loads your `sda.output.Rdata` and then `t` would be greater than 1 and then  it would do the split and so on.

para2x:
@araiho Can I ask you please to point me to one of the setting `xml` s that now you run the code with ? I want to look at the new tags and how they are used.

modules/assim.sequential/R/Analysis_sda.R
para2x: Could you also explain what are you doing here ? or why are you changing the sampler ?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -350,17 +440,17 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
     
   }
 
-  dat <- runMCMC(Cmcmc, niter = nitr, nburnin=nburnin)
-  
+  dat <- runMCMC(Cmcmc, niter = 50000, nburnin=10000)
+  dat_save <- dat[(nrow(dat)-250):nrow(dat),]
+  save(dat_save, file = file.path(outdir, paste0('dat',t,'.Rdata')))
```
para2x: I know it'd be too picky, but that'd be nice if you could save `dat` in SDA folder in `outdir` which we keep plots and other sda related outputs.

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -225,32 +257,67 @@ sda.enkf <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
...
 
       # states will be in X, but we also want to carry some deterministic relationships to write_restart
       # these will be stored in params
       X[[i]]      <- X_tmp[[i]]$X
       if (!is.null(X_tmp[[i]]$params)) new.params[[i]] <- X_tmp[[i]]$params
+    }
+      X <- do.call(rbind, X)
+      FORECAST[[t]] <- X
+      
+    }else{
```
para2x: I still think If your planning to restart a whole new sda with the outputs of an old sda you can use block of code in line 144. It already does most of the code in this `else` block.

araiho:
Hey @para2x and @mdietze I'm getting back into fixing this, but i'm not really sure how to generalize the observation operator step. Do you think I should make a new assim.sequential function that would determine the constants for the tobit? I think I can get this PR finished up pretty quick if I figure that out. Thanks

para2x:
I'm gonna read this and get back to you ASAP.

para2x:
> @araiho Can I ask you please to point me to one of the `xml` settings that now you run the code with ? I want to look at the new tags and how they are used.

@araiho  Could you please share with us an example of the operator's tag ?

araiho:
So this is how I've been doing two data types from paleon.

para2x:
> So this is how I've been doing two data types from paleon.
> 
> ```
>   <inputs>
>     
>    <file>
>     <input.id>1000013298</input.id>
>     <path>
>      <path>/fs/data2/output/paleon_data_products/1000000650/TreeRings//ab_abi.RDS</path>
>     </path>
>     <operator>direct</operator>
>     <variable.id>1000000132</variable.id>
>     <variable.id>1000000263</variable.id>
>     <variable.name>
>      <variable.name>AbvGrndWood</variable.name>
>      <variable.name>GWBI</variable.name>
>     </variable.name>
>    </file>
>    
>    <file>
>     <input.id>1000013717</input.id>
>     <path>
>      <path>/fs/data2/output/paleon_data_products/1000000650/Pollen/fcomp/stepps1_v1.0.nc</path>
>     </path>
>     <operator>fcomp</operator>
>    </file>
>    
>   </inputs>
> ```

Is this inputs tag, the `<inputs>` under the  `<run>` or it's inside the `<state.data.assimilation>` ?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -227,8 +238,19 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
...
     
   }
-
-  dat.tobit2space <- runMCMC(Cmcmc_tobit2space, niter = nitr, nburnin=nburnin,  progressBar=TRUE)
+  
+  if(file.exists(file.path(outdir, paste0('dat.tobit2space',t,'.Rdata')))){
+    load(file.path(outdir, paste0('dat.tobit2space',t,'.Rdata')))
+  }else{
+    dat.tobit2space <- runMCMC(Cmcmc_tobit2space, niter = nitr, nburnin=nburnin,  progressBar=TRUE)
+    save(dat.tobit2space, file = file.path(outdir, paste0('dat.tobit2space',t,'.Rdata')))
```
para2x: You can save this into the SDA folder.

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -227,8 +238,19 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
...
+  }
+  
+  pdf(file.path(outdir,paste0('assessParams',t,'.pdf')))
```
para2x: I would do the same for this.

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -227,8 +238,19 @@ GEF<-function(setting,Forecast,Observed, H, extraArg, nitr=50000, nburnin=10000,
...
+  try(assessParams(dat = dat.tobit2space[1000:4000,], Xt = X))
```
para2x: In the multi-site I ask users to define the length of the chains, That'd be great if you could not fix these numbers.

para2x:
> So this is how I've been doing two data types from paleon.
> 
> ```
>   <inputs>
>     
>    <file>
>     <input.id>1000013298</input.id>
>     <path>
>      <path>/fs/data2/output/paleon_data_products/1000000650/TreeRings//ab_abi.RDS</path>
>     </path>
>     <operator>direct</operator>
>     <variable.id>1000000132</variable.id>
>     <variable.id>1000000263</variable.id>
>     <variable.name>
>      <variable.name>AbvGrndWood</variable.name>
>      <variable.name>GWBI</variable.name>
>     </variable.name>
>    </file>
>    
>    <file>
>     <input.id>1000013717</input.id>
>     <path>
>      <path>/fs/data2/output/paleon_data_products/1000000650/Pollen/fcomp/stepps1_v1.0.nc</path>
>     </path>
>     <operator>fcomp</operator>
>    </file>
>    
>   </inputs>
> ```

I get the first `<file>` chunk which has both the variables and the operator tag. Is there any reason that why the second `<file>` tag with `fcomp` operator doesn't have a variable associated with it ?

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -104,11 +110,15 @@ sda.enkf <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
         ### Data does not have time associated with dates 
         ### Adding 12:59:59PM assuming next time step starts one second later
         PEcAn.logger::logger.warn("Pumpkin Warning: adding one minute before midnight time assumption to dates associated with data")
-        obs.times.POSIX[i] <- lubridate::ymd_hms(paste(obs.times[i], "23:59:59"))
+        obs.times.POSIX[i] <- strptime(paste(obs.times[i], "23:59:59"),format="%Y-%m-%d %H:%M:%S",tz='UTC')#lubridate::ymd_hms(paste(obs.times[i], "23:59:59"))
       }
     }
   }
   obs.times <- obs.times.POSIX
+  obs.times[1] <- strptime('0950-12-31 23:59:59',format="%Y-%m-%d %H:%M:%S",tz="UTC")
```
para2x: These three lines may not be applicable to all cases.

PR #2281, opened 2019-02-08
Travis: Build against multiple R versions

infotroph:
## Description

* Run Travis tests against three versions of R in parallel
* Fold up successful log sections so it's easier to find what failed
* Report time used by each make step (deps/install/document/check) within every package. Especially useful for debugging timed-out builds
* `make check` now skips rebuilding documentation if env var `$REBUILD_DOCS` is false  
* General cleanup: Remove unused packages, delete some build hacks that have outlived their purpose, etc.

### R versions

The major change here is a transition from testing against one R version (hopefully the one most users have) to defining a *build matrix* so that we can test in multiple environments at once. The build matrix concept could be extended in the future to include different OSes, Postgres versions, environment variables, etc, but for this PR the matrix is one-dimensional and contains three R versions: `release`, `devel`, and `oldrel`. These are aliases defined by Travis and currently point respectively to R 3.5, nightly, and 3.4, but versions are updated each release. 

**Opinions, please:** I currently have the devel and oldrel builds marked as `allow_failures`, which means the build is always *run* on all three versions, but will report success as long as it finishes successfully on R-release. My rationale is that this will let us keep an eye on how it's doing in other versions but won't commit us to maintaining them. Is this what we want?
 
* Pro: If R-devel introduces breaking changes we might want to defer fixes until it's clear the change will actually be released, and similarly if R-oldrel stops working we might want to reserve the right to say "OK, WONTFIX, PEcAn doesn't support that version anymore". 
* Con: Any build that's *allowed* to fail will likely *keep* failing and become harder to fix later, so if we do want to commit to supporting devel or oldrel then we should commit now and require that they pass for every merge.

### Skip documentation

I've updated `scripts/check_with_errors.R` to recognize an environment variable `$REBUILD_DOCS` and pass its logical value to the `document` argument of `devtools::check`. This is primarily to save a few seconds per package on Travis, where documentation is always freshly-built at check time, but feel free to use it any other time `make check` fails to document/not document things the way you want.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,17 +135,26 @@ before_script:
 
 script:
   - set -e
-  - echo 'Installing PEcAn packages'
   - Rscript scripts/generate_makefile_deps.R
+  - echo 'Installing PEcAn packages' && echo -en 'travis_fold:start:install_pkgs\\r'
   # TODO: Would probably be faster to use -j2 NCPUS=1 as for other steps,
   # but many dependency compilations seem not parallel-safe.
   # More debugging needed.
   - NCPUS=2 make -j1
-  - echo 'Testing PEcAn packages'
+  - echo -en 'travis_fold:end:install_pkgs\\r'
+  #
+  - echo 'Testing PEcAn packages' && echo -en 'travis_fold:start:test_pkgs\\r'
   - make test
+  - echo -en 'travis_fold:end:test_pkgs\\r'
+  #
+  - echo 'Checking PEcAn packages' && echo -en 'travis_fold:start:check_pkgs\\r'
   - REBUILD_DOCS=FALSE make check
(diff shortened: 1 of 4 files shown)


infotroph:
@mdietze I may not have made the "opinions, please" section prominent enough. Should I take your approval to also be an endorsement of allowing failures in oldrel and devel?

mdietze:
Yep, I endorse your approach, but didn’t pull yet so as to give others time to comment. Given how some packages change, it may be impossible to support multiple versions without code that detects which version is installed and switches among different code blocks, and that could become a nightmare

PR #2283, opened 2019-02-09
wrap GFDL URL check in test, skip on travis

infotroph:
The existing skip did eventually fire, but not until after waiting 75s for the connection to time out.

@ashiklom you added these lines in #2273 but I'm not sure I follow your intent -- does this change retain the behavior you wanted?

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/tests/testthat/test.download.GFDLR.R
+++ b/modules/data.atmosphere/tests/testthat/test.download.GFDLR.R
@@ -4,13 +4,22 @@ tmpdir <- tempfile(pattern = "GFDLtest")
 dir.create(tmpdir, showWarnings = FALSE)
 teardown(unlink(tmpdir, recursive = TRUE))
 
-test_url <- paste0("http://nomads.gfdl.noaa.gov:9192/opendap/",
-                   "CMIP5/output1/NOAA-GFDL/GFDL-CM3/rcp45/3hr/",
-                   "atmos/3hr/r1i1p1/v20110601/tas/",
-                   "tas_3hr_GFDL-CM3_rcp45_r1i1p1_2006010100-2010123123.nc")
+test_that("URL is openable", {
+  skip_on_travis()
 
-test_nc <- tryCatch(ncdf4::nc_open(test_url), error = function(e) {
-  skip("Skipping test because unable to reach NOAA GFDL download server.")
+  test_url <- paste0("http://nomads.gfdl.noaa.gov:9192/opendap/",
+                     "CMIP5/output1/NOAA-GFDL/GFDL-CM3/rcp45/3hr/",
+                     "atmos/3hr/r1i1p1/v20110601/tas/",
+                     "tas_3hr_GFDL-CM3_rcp45_r1i1p1_2006010100-2010123123.nc")
+  test_nc <- tryCatch(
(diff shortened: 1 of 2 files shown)


modules/data.atmosphere/tests/testthat/test.download.GFDLR.R
```diff
@@ -4,13 +4,22 @@ tmpdir <- tempfile(pattern = "GFDLtest")
...
+                     "CMIP5/output1/NOAA-GFDL/GFDL-CM3/rcp45/3hr/",
+                     "atmos/3hr/r1i1p1/v20110601/tas/",
+                     "tas_3hr_GFDL-CM3_rcp45_r1i1p1_2006010100-2010123123.nc")
+  test_nc <- tryCatch(
+    ncdf4::nc_open(test_url, suppress_dimvals = TRUE),
+    error = function(e) {
+      skip("Skipping test because unable to reach NOAA GFDL download server.")
+    }
+  )
+  teardown(ncdf4::nc_close(test_nc))
```
ashiklom: The original idea here was to check if a known GFDL download URL was accessible, and if not, to `skip` the _entire file_, since I figured all the other tests would fail if the URL was down.

The ideal solution here would be to mock some key aspect of the GFDL download, which would allow us to test the GFDL URL and actual download code separately. But I'm still not 100% clear on how to do this.

infotroph: OK, that makes sense. Skipping the entire file is doable -- my first thought is to nest all the other tests inside this one.

To test without timing out: Does the GFDL server speak http, or does NCDF4 have some equivalent of a header-only GET, to let us do the equivalent of "If `curl -I http://.../thefile.nc` doesn't return 200, then skip()"?

ashiklom: Sorry -- missed this notification.

Since GFDL is using THREDDS to serve up NetCDF files, you should be able to `curl` (and similar) against the root `catalog.html` file. I'm like 80% sure that if the catalog is accessible, then the files should be as well. In fact, this will allow us to separate failures in our code to build correct URLs to individual files (which this will _not_ catch) from the GFDL server just being down.

The catalog URL may be somewhat different from the OpenDAP service URL we're currently using. It might be as simple as `http://nomads.gfdl.noaa.gov:9192/catalog.html`, but there's a good chance the port is different or there's an additional prefix. In general, the way THREDDS works is that every directory is associated with a `catalog.html` file, so we could actually have multiple tests -- e.g. one that checks the overall status of the NOMADS server, one that checks if the directory location of GFDL is still the same (i.e. `.../CMIP5/output1/NOAA-GFDL/GFDL-CM3/catalog.html`), and finally one that checks for the files themselves.

robkooper: I think it would be good if we can make it a two step process since we have no control over how long nc_open takes (currently 120seconds until timeout). So if we can timeout faster that would be great.

infotroph: This sounds promising, but we'll need a functioning server to test it. Would you feel OK about shipping the existing change and punting a more sensible availability test to a future PR?

PR #2284, opened 2019-02-11
Dependency cleanup in assim.batch

infotroph:
## Description
* Removed all Depends
* Added used packages to Imports
* Namespace-qualified function calls
* Switched instances of `[methods::]is` to `[base::]inherits` to avoid importing `methods`, and also because they are testing S3 objects while `is` is intended for use with S4
* Removed a `...` used in an illegal context
* Converted author list to Authors@R format
* Sorted import list

All changes intended to have zero user-visible changes (other than faster loading with fewer unwanted messages), but note that I *have not* run any tests beyond those done by `devtools::check`.

## Motivation and Context
Fun fact: Excessive Depends can slow down package checks! Before cleanup: >70s. After: ~26s.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


infotroph:
@istfer yes, it's usually preferable to qualify any function that's not in `base` or the package you're editing. The idea is that inside this package's namespace we don't want to make any assumptions about the search path, even for packages that are loaded by default in most sessions. 

TBH the worst outcome that's *likely* is just that the package checks will nag about it, but at the same time a surprising number of package authors feel OK about implementing their own base-masking versions of functions from `graphics` and `stats`, so qualifying the references can help us avoid unexpected results if someone loads one of those packages into their PEcAn session.

istfer:
got it, thanks for taking time to explain

Issue #2244, opened 2019-01-18
`write.ensemble.configs` fails if `write.to.bety` is `FALSE`


ashiklom:
At issue is this code block: https://github.com/PecanProject/pecan/blob/bb018557ccecfeed4f76d02a154ed9a147ed93a4/modules/uncertainty/R/ensemble.R#L211-L220

Even if we are not writing to BETY, we should still be able to _read_ from it, but we are unable to do so because `con` is set to `NULL`. Really, the only circumstance where `con` should be set to `NULL` is if the entire `database` settings block is missing.

This leads to some insidious downstream bugs, as @serbinsh and I discovered. For instance, none of the input paths required for ED2 (e.g. `css`, `pss`, `site`, `thsum`) are set.

serbinsh:
@para2x bump.....could you take a look at this and let us know if this is an easy change/fix?

para2x:
@serbinsh  The reason I haven't taken any action yet is because I haven't run ED and it's hard for me to test and see if my solution would work (and spending time on learning on how to run `ED` is not what I want to spend time on at the moment). But I can have fix PR and let you guys test it.

dlebauer:
@para2x have you tried running the integration tests? They are intended to be able to be run without having to understand how to set up a specific model. For example, you should be able to run ED2 on the VM with the following code:

```r
./web/workflow.R --settings tests/pecan64.ed.xml
```

KristinaRiemer:
FYI, I just ran into the bug in #2245. Got over that hurdle by changing the config file according to @dlebauer suggestion, though I'm still working through the rest of the bugs.

PR #2288, opened 2019-02-13
Write config bug fix and improvement

para2x:
## Description

For the case that we want to write config files on remote there was an issue in `run.write.configs.R` which was throwing an error because it wanted to try to connect to `bety`.  I just added a try to avoid crashing the whole function. I also addressed issue #2244 . 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


para2x:
@serbinsh Could you please  see if this change in the `write.ensemble.configs` function solves your ED problem and also not breaking other stuff ?

para2x:
@serbinsh this PR is ready for review.

para2x:
@serbinsh you can also review this PR if you have time.

base/workflow/R/run.write.configs.R
```diff
@@ -26,8 +26,10 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
                               posterior.files = rep(NA, length(settings$pfts)), 
                               overwrite = TRUE) {
   
-  con <- PEcAn.DB::db.open(settings$database$bety)
-  on.exit(PEcAn.DB::db.close(con))
+  con <-
```
mdietze: So the problem here is that the later code that uses `con` is not checking to see if the connection was successful. I'd move this whole chunk into the Posteriors code below that uses the con (a few ifs in), since we don't need the connection open unless those conditions are met, and then to add an explicit `logger.severe` if we need the database connection but don't have access to the DB

modules/uncertainty/R/ensemble.R
```diff
@@ -207,28 +207,28 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
     return(list(runs = NULL, ensemble.id = NULL))
   }
   
+  # See if we need to write to DB
+  write.to.db <- as.logical(settings$database$bety$write)
   # Open connection to database so we can store all run/ensemble information
-  if (write.to.db) {
```
mdietze: I don't understand why this was rewritten to drop this `if(write.to.db)` and to instead attempt to open that connection regardless.

modules/uncertainty/R/ensemble.R
```diff
@@ -207,28 +207,28 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+  
+  # If we fail to connect to DB then we set to NULL
+  if (inherits(con, "try-error")) {
     con <- NULL
   }
+
   
   # Get the workflow id
-  if ("workflow" %in% names(settings)) {
+  if (!is.null(settings$workflow$id)) {
```
mdietze: won't this throw an error, rather than return NULL, if `settings$workflow` doesn't exist

modules/uncertainty/R/ensemble.R
serbinsh: Indeed, I would think we still want a check to see if the user has selected to write or not, if not it simplifies this.  I dont think we want to assume that we want write to bety

modules/uncertainty/R/ensemble.R
```diff
@@ -207,28 +207,28 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
     workflow.id <- settings$workflow$id
   } else {
     workflow.id <- -1
   }
   #------------------------------------------------- if this is a new fresh run------------------  
   if (is.null(restart)){
     # create an ensemble id
-    if (!is.null(con)) {
+    if (!is.null(con) & write.to.db) {
```
serbinsh: should these be "&&"

modules/uncertainty/R/ensemble.R
```diff
@@ -326,7 +320,7 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
     # write configuration for each run of the ensemble
     runs <- data.frame()
     for (i in seq_len(settings$ensemble$size)) {
-      if (!is.null(con)) {
+      if (!is.null(con) & write.to.db) {
```
serbinsh: same

modules/uncertainty/R/ensemble.R
mdietze: So @serbinsh and I seem to agree that we want to retain this check, but disagree about the default -- I *DO* think we want to assume that we want to write to bety by default.

modules/uncertainty/R/ensemble.R
ashiklom: `[[` subset syntax will, but `$` sub-setting returns `NULL`. This is actually one of the reasons I typically prefer `[[` in my own code -- I like when things fail early, with loud and clear errors, rather than silently propagating `NULL`. But in this case, it's appropriate.

``` r
  l <- list(a = 3)
  l$x$y$z
#> NULL
```

<sup>Created on 2019-06-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

base/workflow/R/run.write.configs.R
ashiklom: In addition to what Mike said, I would implement this as a proper `tryCatch` statement:

```
tryCatch({
  con <- PEcAn.DB::db.open(...)
  on.exit(PEcAn.DB::db.close(con)
}, error = function(e) {
  PEcAn.logger::logger.severe("Connection requested, but failed to open with the following error: ", conditionMessage(e)
})
```

modules/uncertainty/R/ensemble.R
ashiklom: Actually, turns out `[[` only throws an error if you do a relatively uncommon kind of deep indexing `[[c(...)]]`. If you just index one-at-a-time, it will silently propagate the `NULL` through everything.

``` r
  l <- list(a = 3)
  l$x$y
#> NULL
  l[["x"]][["y"]]
#> NULL
  l[[c("x", "y")]]
#> Error in l[[c("x", "y")]]: no such index at level 1
```

<sup>Created on 2019-06-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mdietze:
@para2x working through the PR queue -- looks like there's requested changes from June and a conflict on ensemble.R. Could you address these and let me know when you're ready for me to take a final look at this?

modules/uncertainty/R/ensemble.R
para2x: The default behavior is set by the function's argument which will be `TRUE` to write to bety.

para2x:
@mdietze I think I have the comments addressed.

modules/uncertainty/R/ensemble.R
```diff
@@ -200,35 +200,43 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
 write.ensemble.configs <- function(defaults, ensemble.samples, settings, model, 
                                    clean = FALSE, write.to.db = TRUE,restart=NULL) {
   
+  con <- NULL
```
ashiklom: This seems unnecessary. You're just redefining this variable below.

modules/uncertainty/R/ensemble.R
```diff
@@ -200,35 +200,43 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
+  write.to.db <- as.logical(settings$database$bety$write)
+  
   if (write.to.db) {
-    con <- try(PEcAn.DB::db.open(settings$database$bety), silent = TRUE)
-    if (inherits(con, "try-error")) {
+    # Open connection to database so we can store all run/ensemble information
+    con <-
+      try(PEcAn.DB::db.open(settings$database$bety), silent = TRUE)
+    on.exit(try(PEcAn.DB::db.close(con), silent = TRUE)
+    )
```
ashiklom: Why are you silencing errors here? At the very least, I would throw an informative warning, though I personally would be inclined to re-throw the error the same way you do in `read.write.configs`.

ashiklom: Also, it's a good idea to call `on.exit` as `on.exit(..., add = TRUE)`. Otherwise, every invocation of `on.exit` will overwrite the previous one, which is usually not what we want.

From `?on.exit`:

```
     add: if TRUE, add ‘expr’ to be executed after any previously set
          expressions (or before if ‘after’ is FALSE); otherwise (the
          default) ‘expr’ will overwrite any previously set
          expressions.
```

modules/uncertainty/R/ensemble.R
```diff
@@ -299,19 +308,28 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
-        PEcAn.logger::logger.warn(paste0("The following pfts are specified for the siteid ", settings$run$site$id ," but they are not defined as a pft in pecan.xml:",
-                                         site.pfts.vec[which(!(site.pfts.vec %in% defined.pfts))]))
+      if (length(which(!(site.pfts.vec %in% defined.pfts))) > 0)
+        PEcAn.logger::logger.warn(
+          paste0(
+            "The following pfts are specified for the siteid ",
+            settings$run$site$id ,
+            " but they are not defined as a pft in pecan.xml:",
+            site.pfts.vec[which(!(site.pfts.vec %in% defined.pfts))]
+          )
```
ashiklom: This probably need to be `collapse`d? Otherwise, it will print this entire string once for every vector element.

You should just be able to add something like `paste0(..., collapse = ", ")`.

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -16,8 +16,8 @@ get.parameter.samples <- function(settings,
   pft.names <- list()
   outdirs   <- list()
   ## Open database connection
-  con <- PEcAn.DB::db.open(settings$database$bety)
-  on.exit(PEcAn.DB::db.close(con))
+  con <- try(PEcAn.DB::db.open(settings$database$bety), silent = TRUE)
+  on.exit(try(PEcAn.DB::db.close(con), silent = TRUE))
```
ashiklom: See my comments above about (1) silencing errors, and (2) `on.exit(..., add = TRUE)`.

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -66,7 +66,7 @@ get.parameter.samples <- function(settings,
     }
     
     ### Load trait mcmc data (if exists, either from MA or PDA)
-    if (!is.null(settings$pfts[[i]]$posteriorid)) { # first check if there are any files associated with posterior ids
+    if (!is.null(settings$pfts[[i]]$posteriorid) & !inherits(con, "try-error")) {# first check if there are any files associated with posterior ids
```
ashiklom:
```suggestion
    if (!is.null(settings$pfts[[i]]$posteriorid) && !inherits(con, "try-error")) {# first check if there are any files associated with posterior ids
```

modules/uncertainty/R/ensemble.R
para2x: @ashiklom if we just define the `con` in the scope of the `if` then it would break all the ifs afterwards.

ashiklom: Ah, OK, I see that now.

base/workflow/R/run.write.configs.R
```diff
@@ -26,15 +26,22 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
...
   ## Which posterior to use?
   for (i in seq_along(settings$pfts)) {
     ## if posterior.files is specified us that
     if (is.na(posterior.files[i])) {
       ## otherwise, check to see if posteriorid exists
       if (!is.null(settings$pfts[[i]]$posteriorid)) {
+        
+        tryCatch({
+          con <- PEcAn.DB::db.open(...)
+          on.exit(PEcAn.DB::db.close(con)
```
ashiklom: Missing close paren here (thanks, Travis!). May as well add the `add = TRUE` as well while you're at it.

```suggestion
          on.exit(PEcAn.DB::db.close(con), add = TRUE)
```

https://travis-ci.org/PecanProject/pecan/jobs/590413421#L3177

para2x:
@mdietze This should be ready to be pulled in.

PR #2289, opened 2019-02-13
Docker fixes

robkooper:
Fixes some docker problems:
* fixed issue queues for models

Added new model
* maespa

Cleaner installation of dependencies of PEcAN:
* scripts/generate_makefile_deps.R renamed to scripts/generate_dependencies.R
* This script now generates dependency files for both make as well as the depends image in docker.

Some reorganization of the docker build files
* Moved Dockerfiles from docker /base/* to docker/*/Dockerfile
* thredds now depends on official 4.6 image
* update build files to take into account new docker directory
* script to build/deploy containers this runs on pecan-dev.ncsa.illinois.edu
* some tweaks to install script

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


docker.sh
```diff
@@ -32,9 +32,7 @@ echo ""
 if [ "${DEPEND}" == "build" ]; then
     echo "# To just pull latest/develop version (default) run"
     echo "# DEPEND=pull $0"
-    ${DEBUG} docker build \
-        --tag pecan/depends:latest \
-        --file docker/base/Dockerfile.depends .
+    ${DEBUG} docker build --tag pecan/depends:latest docker/depends
```
ashiklom: Can you make the `latest` here (and other places) user-configurable via `DEPENDS_VERSION=${DEPENDS_VERSION:-latest}` or something?

docker.sh
```diff
@@ -58,19 +56,21 @@ else
 fi
 echo ""
 
-# build images in this specific order. Images are tagged with latest so other
-# docker images build in this script will use that specifc build.
-for i in base models executor web data thredds docs; do
-    ${DEBUG} docker build \
+# require all of PEcAn to build
+for x in base web docs; do
+    ${DEBUG} docker build --tag pecan/$x:latest --file docker/$x/Dockerfile \
```
ashiklom: Similar to above, can we make `latest` here be user-configurable? It might make sense to tie this to `IMAGE_VERSION`, so that the containers have mutually-consistent tags.

docker.sh
```diff
@@ -58,19 +56,21 @@ else
...
         --build-arg PECAN_GIT_CHECKSUM="${PECAN_GIT_CHECKSUM}" \
         --build-arg PECAN_GIT_DATE="${PECAN_GIT_DATE}" \
-        --tag pecan/${i}:latest \
-        --file docker/base/Dockerfile.${i} .
+        .
+done
+
+# all files in subfolder
+for x in models executor data thredds monitor; do
+    ${DEBUG} docker build --tag pecan/$x:latest docker/$x
```
ashiklom: As above, make `latest` configurable.

docker.sh
```diff
@@ -82,13 +82,21 @@ for version in 136; do
...
         --tag pecan/model-ed2-${version}:latest \
-        --file models/ed/Dockerfile .
+        models/ed
+done
+
+# build maespa
+for version in git; do
+    ${DEBUG} docker build \
+        --build-arg MODEL_VERSION="${version}" \
+        --tag pecan/model-maespa-${version}:latest \
```
ashiklom: Again, this `latest` should be configurable. This could use a separate variable (e.g. `MODEL_VERSION`).

docker/data/add-data.sh
```diff
@@ -0,0 +1,216 @@
+#!/bin/bash
+
+FQDN="docker"
+DATADIR="/data"
+PSQL="psql -U bety -h postgres -d bety -q -t -c"
```
ashiklom: Should some (or all) of these be configurable (with the current values as defaults)? Or is there a Very Good Reason for these to be hard-coded as they are? I'm especially interested in `FQDN` -- I'm still not 100% clear on where we always use `"docker"` vs. where it will be machine-specific.

docker/docs/Dockerfile
```diff
@@ -6,6 +6,7 @@ MAINTAINER Rob Kooper <kooper@illinois.edu>
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends pandoc \
+    && Rscript -e 'install.packages("bookdown")' \
```
ashiklom: Probably better to use `install2.r` here?

docker/depends/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# ----------------------------------------------------------------------
+# PECAN FOR MODEL BASE IMAGE
+# ----------------------------------------------------------------------
+FROM rocker/tidyverse:3.5
```
ashiklom: Could we also enable finer control of this with a variable? E.g.
```
R_VERSION=${R_VERSION:-3.5}
```

docker/depends/pecan.depends
```diff
@@ -0,0 +1,110 @@
...
+# install remotes first in case packages are references in dependencies
+installGithub.r \
+    araiho/linkages_package \
+    ebimodeling/biocro \
+    MikkoPeltoniemi/Rpreles \
+    ropensci/geonames \
+    ropensci/nneo
+
+# install all packages (depends, imports, suggests)
+install2.r -e -s \
```
ashiklom: Nice use of `littler`! This is a much better solution than what we had.

I would consider adding `-n -1`, which is a hidden option that tells this script to use as many cores as are available (https://github.com/eddelbuettel/littler/blob/master/inst/examples/install2.r#L66-L69).

docker.sh
robkooper: Using IMAGE_VERSION for the version. Can now use `DEPEND=pull IMAGE_VERSION=pr ./docker.sh` to create images that are tagged with pr.  The DEPEND=pull is needed to make sure we have a depends image that is tagged correctly (you can also use DEPEND=build but that will take a while).

docker.sh
robkooper: fixed

docker/depends/Dockerfile
robkooper: Changed this, however to go to 3.6 we need to change it in 2 places (docker.sh and Dockerfile).

docker.sh
robkooper: using IMAGE_VERSION to keep it in sync with the pecan version number.

docker/depends/pecan.depends
robkooper: in travis we have a comment about some issues when using more than 1 cpu:

```
  # TODO: Would probably be faster to use -j2 NCPUS=1 as for other steps,
  # but many dependency compilations seem not parallel-safe.
  # More debugging needed.
```

docker/data/add-data.sh
robkooper: The FQDN should be  a variable. This is the name of the machine used in the machines table and should be  FQDN. using the name of the docker container will make it random, hence we use the specified name. See also #2128 and #2186.

robkooper: Added to documentation quickstart as well.

docker/depends/pecan.depends
infotroph: I added that comment in frustration several build revisions ago, and I don't remember which packages were causing trouble or how many "many" was. Probably worth revisiting at some point -- maybe the situation has improved?

robkooper: added -n -1

docker.sh
```diff
@@ -33,21 +36,30 @@ if [ "${DEPEND}" == "build" ]; then
...
-      ${DEBUG} docker pull pecan/depends:latest
+        echo "# this will pull latest of base image"
+        echo "# To disable run DEPEND=nothing $0"
+        if [ "$( docker image ls -q pecan/depends:latest )" == "" ]; then
+            ${DEBUG} docker pull pecan/depends:latest
+        fi
+        if [ "${IMAGE_VERSION}" != "latest" ]; then
+            ${DEBUG} docker tag pecan/depends:latest pecan/depends:${IMAGE_VERSION}
+            ${DEBUG} docker image rm pecan/depends:latest
+        fi
```
ashiklom: I think this tagging step should also happen in the `else` block below (if `DEPEND` is unset).

Issue #2291, opened 2019-02-13
Add hourly option to backup.bety.sh


Chris-Schnaufer:
## Description

Update scripts/backup.bety.sh to do the following:
-[x] accept hourly backups as an option (default behavior of daily backups is kept)
-[x] Use **cp** for other backups after original is generated
-[x] Accept optional Postgresql command line parameters for non-standard environments
-[x] Add -h (aka: help) command line option to show help

## Context

- Provides additional granularity on backups. 
- Saves time by copying the backup instead of performing a new dump
- Adds flexibility for non-standard environments
- Help is helpful

## Possible Implementation

For the postgresql command line option flag, continue the use of the -p command line parameter

PR #2292, opened 2019-02-13
Added features requested in #2291

Chris-Schnaufer:
## Description

- Added check for HOURLY environment variable to enable an hourly-labeled database dump
- If making an hourly dump, set the filename with 'h' and the hour (in 24 hour time format). For example, `bety-h-14.sql.gz`
- If **not** making an hourly dump, assume making a daily dump
- If making an hourly dump, copy hourly dump to daily dump at hour 00
- Changed other dump commands to copy the original dump (hour or daily) to correct name
- Added `-p` command line option for additional psql command line parameters
- Added `-h` command line option to display help
- Added '-o' command line option to specify output location override

## Motivation and Context

- Can now make an hourly dump of the database
- Can run in non-standard environments, such as a machine hosting Pecan in docker containers
- Using **cp** saves time
- Showing help is nice

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/scripts/backup.bety.sh
+++ b/scripts/backup.bety.sh
@@ -9,18 +9,17 @@
 # - bety-m-X, montly backup, where X is the month of the year
 # - bety-y-X, yearly backup, where X is the actual year.
 
-# variables to use for the database dump
+# Variables to use for the database dump
 DATABASE=bety
 BETYUSER=bety
 
-# location where backup should be written
+# Location where backup should be written
 BACKUPDIR=$( dirname $0 )/backup
-mkdir -p ${BACKUPDIR}
 
-# set path if needed
+# Set path if needed
 #export PATH=<location to postgresql>/bin;${PATH}
 
-# some handy variables
+# Some handy variables
 HOUR=$( date +"%H" )
 TODAY=$( date +"%d" )
 YESTERDAY=$( date -d "yesterday" +"%d" )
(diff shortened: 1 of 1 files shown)


scripts/backup.bety.sh
```diff
@@ -30,49 +29,56 @@ WEEK=$( date +"%W" )
...
+# Psql options
+# This allows you to add any other options
 PG_OPT=${PG_OPT:-""}
 
-# parse command line options
-while getopts hp: opt; do
+# Parse command line options
+while getopts ho:p: opt; do
   case $opt in
   h)
```
robkooper: maybe add an option for `-1` for hourly (every 1 hour).

robkooper:
As with all the scripts that work on the BETY database I think it would be good to move this script to the BETY repository. At that point add it to the docker entrypoint with the backup argument so it is easy to backup the database using docker. For example `docker run -ti --rm -v /somepathondisk/backup:/backup pecan/bety backup` and set the default backup folder to be /backup in the docker image.

scripts/backup.bety.sh
Chris-Schnaufer: I think adding a '-H' option would make sense. The '-1' option doesn't have a units associated with it and opens the door to '-1h', '-1d', etc.

PR #2298, opened 2019-02-22
No more (R)MySQL, and miscellaneous fixes and tweaks

ashiklom:
## Description

A collection of minor fixes to various things:
- Remove as many references to `RMySQL` as I could find.
- In `settings::check.databases`, explicitly support _only_ `RPostgreSQL` and the more modern `RPostgres`. Also, opinionated minor revisions for style and clarity.
- Explicitly use `uncertainty::read.ensemble.output` in `utils::get.results`. By default, this was using the deprecated `utils` version.
-  Replace deprecated `rlang::UQ` with `!!` , per the package warnings. (@infotroph tagging you mostly because this was in code you wrote).
- Fix `workflow.R` path in `executor.py` section for running XML passed as string (used by API).
- Fix THREDDS catalog path. Although the `/usr/local/...` path is recommended in the documentation, it stopped working for me. Using the `/opt/...` path fixes it.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/db/R/query.traits.R
+++ b/base/db/R/query.traits.R
@@ -50,7 +50,7 @@ query.traits <- function(ids, priors, con = NULL,
   traits <- (dplyr::tbl(con, "traits")
     %>% dplyr::inner_join(dplyr::tbl(con, "variables"), by = c("variable_id" = "id"))
     %>% dplyr::filter(
-      (rlang::UQ(id_type) %in% ids),
+      (!!id_type %in% ids),
       (name %in% priors)) # TODO: use .data$name when filter supports it
     %>% dplyr::distinct(name) # TODO: use .data$name when distinct supports it
     %>% dplyr::collect())
--- a/base/settings/R/check.all.settings.R
+++ b/base/settings/R/check.all.settings.R
@@ -96,42 +96,53 @@ check.inputs <- function(settings) {
 ##' @param settings settings file
 ##' @export check.database
 check.database <- function(database) {
-  if (is.null(database)) return(NULL);
+  if (is.null(database)) return(NULL)
   
   ## check database settings
   if (is.null(database$driver)) {
     database$driver <- "PostgreSQL"
-    PEcAn.logger::logger.warn("Please specify a database driver; using default 'PostgreSQL'")
+    PEcAn.logger::logger.info("Database driver unspecified. ",
+                              "Using 'PostgreSQL' (default)")
+  }
+
+  is_postgres_like <- database$driver %in% c("PostgreSQL", "Postgres")
+  is_postgresql <- database$driver == "PostgreSQL"
+
+  if (!is_postgres_like) {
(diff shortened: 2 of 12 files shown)


base/db/R/query.traits.R
```diff
@@ -50,7 +50,7 @@ query.traits <- function(ids, priors, con = NULL,
   traits <- (dplyr::tbl(con, "traits")
     %>% dplyr::inner_join(dplyr::tbl(con, "variables"), by = c("variable_id" = "id"))
     %>% dplyr::filter(
-      (rlang::UQ(id_type) %in% ids),
+      (!!id_type %in% ids),
```
infotroph: UQ was an attempt to avoid check warnings, but I haven't checked on it in a while. If !! now works without a namespace qualifier, then 👍

base/settings/R/check.all.settings.R
```diff
@@ -96,42 +96,53 @@ check.inputs <- function(settings) {
...
+        "Database field `passwd` is soft-deprecated. ",
+        "Please use `password` instead."
+      )
       database$password <- database$passwd
       database$passwd <- NULL
     }
     if (!is.null(database$name)) {
-      PEcAn.logger::logger.info("name in database section should be dbname for PostgreSQL")
+      PEcAn.logger::logger.warn(
+        "Database field `name` is soft-deprecated. ",
```
infotroph: This may be a personal bugbear, but I don't think 'soft-deprecated' is a useful category. Can we drop the "soft-"?

ashiklom: Addressed.

base/db/R/query.traits.R
ashiklom: Here's what the `rlang` folks say as of [0.3.0](https://rlang.r-lib.org/news/index.html#deprecated-functions-and-arguments). I did this mostly to silence the annoying runtime warning about `rlang::UQ` being deprecated.

ashiklom: My understanding of this is that `!!` and `!!!` aren't really functions so much as some kind of Domain Specific Programming magic operators. [Here](https://github.com/r-lib/rlang/blob/master/R/quasiquotation.R#L108-L134) is what the `rlang` source code has to say about them and why the namespaced forms are wrong.

As for how `R CMD check` will respond to this...🤷‍♂️

ashiklom: I just ran `devtools::check` and didn't see any messages related to `!!`, so I think this is OK.

infotroph: <del>Still baffled how it works, but</del><ins>oh yeah, duh, we were [importing it already](https://github.com/PecanProject/pecan/blob/develop/base/db/R/zzz.R#L9)</ins>. Yep, this looks good and sorry for the extra noise.

PR #2299, opened 2019-02-26
Added RStudio container

Chris-Schnaufer:
## Description

Added RStudio docker file configuration to docker-compose.yml. 

## Motivation and Context

RStudio is used by many scientists to process data and to use new algorithms. Having this container as part of the docker-compose.yml file will allow them easier access to the Pecan environment

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


docker-compose.yml
```diff
@@ -156,6 +156,28 @@ services:
       - "traefik.frontend.rule=${TRAEFIK_FRONTEND_RULE:-}PathPrefix:/bety/"
       - "traefik.backend=bety"
 
+  # ----------------------------------------------------------------------
+  # Pecan-RStudio
+  # ----------------------------------------------------------------------
+  rstudio:
+    image: pecan/base:1.7.0
```
robkooper: Use same syntax as other pecan containers to make sure we have latest etc options

docker-compose.yml
```diff
@@ -156,6 +156,28 @@ services:
...
+    restart: unless-stopped
+    networks:
+      - pecan
+    environment:
+      - PASSWORD=${RSTUDIO_DEFAULT_PASSWORD:-bety}
```
robkooper: Can you add this to the env.example file. Maybe use PECAN_RSTUDIO as name for env variable, and use standard pecan password, can you set USERNAME as well?

docker-compose.yml
```diff
@@ -156,6 +156,28 @@ services:
...
+      - bety
+    entrypoint: /init
+    labels:
+      - "traefik.enable=true"
+      - "traefik.backend=rstudio"
+      - "traefik.port=8787"
+      - "traefik.frontend.rule=${TRAEFIK_FRONTEND_RULE:-}PathPrefix:/rstudio"
+      - "traefik.website.frontend.whiteList.sourceRange=${TRAEFIK_IPFILTER:-172.16.0.0/12}"
+    ports:
+      - 8787:8787
```
robkooper: ports should be closed by default.

docker-compose.yml
Chris-Schnaufer: Update to variable for version (in env.example as well)

docker-compose.yml
Chris-Schnaufer: Added USER and appropriately named variables for both USER and PASSWORD per suggestion (in env.example as well)

docker-compose.yml
Chris-Schnaufer: Closed port

docker-compose.yml
```diff
@@ -156,6 +156,29 @@ services:
       - "traefik.frontend.rule=${TRAEFIK_FRONTEND_RULE:-}PathPrefix:/bety/"
       - "traefik.backend=bety"
 
+  # ----------------------------------------------------------------------
+  # Pecan-RStudio
+  # ----------------------------------------------------------------------
+  rstudio:
+    image: pecan/base:${PECAN_RSTUDIO_VERSION:-latest}
```
robkooper: I think it is best to use PECAN_VERSION since we want to have it go hand in hand with the running version of the stack.

docker-compose.yml
```diff
@@ -156,6 +156,29 @@ services:
...
+    depends_on:
+      - bety
+    entrypoint: /init
+    labels:
+      - "traefik.enable=true"
+      - "traefik.backend=rstudio"
+      - "traefik.port=8787"
+      - "traefik.frontend.rule=${TRAEFIK_FRONTEND_RULE:-}PathPrefix:/rstudio"
+      - "traefik.website.frontend.whiteList.sourceRange=${TRAEFIK_IPFILTER:-172.16.0.0/12}"
+    ports:
```
robkooper: remove ports completely. If you want to open the port you can do it in docker-compose.override.yml.

docker-compose.yml
Chris-Schnaufer: Make sense, good idea. Done

docker-compose.yml
Chris-Schnaufer: Ports are no more

PR #2305, opened 2019-03-01
Extract/met2cf function for ERA5

para2x:
In this PR I'm providing the functions required for extracting met data from the existing ERA5 `nc` files as well as converting them to pecan standard format. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


tonygardella:
@ashiklom Added you as reviewer because you have experience with ERA5
@ankurdesai Added you as reviewer because you're paired with Hamze this week for code review.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+#' @examples
+#' \dontrun{
+#' files.era5 <- met2cf.ERA5(years=2018,
+#'                   data.folder=".",
+#'                   outfolder = "./metfiles")
+#'
+#' }
+met2cf.ERA5 <- function(lat = 40,
+                        long = -120,
+                        years = c(2015:2017),
```
ashiklom: Again, these should not have defaults.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+                        overwrite = FALSE,
+                        verbose = T,
+                        sitename,
+                        data.folder,
+                        outfolder) {
```
ashiklom: See above comment about argument order.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+  
+  # Extracting the raw data - The output would be a list of xts objects for each ensemble
+  out <- ERA5_extract_ENS(lat = lat,
+                          long = long,
+                          years = years, 
+                          data.folder=data.folder)
```
ashiklom:
```suggestion
                          data.folder = data.folder)
```

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+    purrr::map(function(ensi) {
+ 
+      ens <- out[[ensi]]
+      # Solar radation conversions 
+      #https://confluence.ecmwf.int/pages/viewpage.action?pageId=104241513
+      #For ERA5 daily ensemble data, the accumulation period is 3 hours. Hence to convert to W/m2:
+      ens[, "ssrd"] <- ens[, "ssrd"]/ (3 * 3600)
+      ens[, "strd"] <- ens[, "strd"]/ (3 * 3600)
+      #precipitation it's originaly in meters. Meters times the density will give us the kg/m2
+      ens[, "tp"] <- ens[, "tp"]*1000/3 # divided by 3 because we have 3 hours data
```
ashiklom: Why not just use `udunits2::ud.convert` here?

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+      ens[, "tp"] <- udunits2::ud.convert(ens[, "tp"], "kg m-2 hr-1", "kg m-2 6 s-1")  #There are 21600 seconds in 6 hours
+      #RH
+      #Adopted from weathermetrics/R/moisture_conversions.R
+      t <- udunits2::ud.convert(ens[, "t2m"] %>% as.numeric(), "K", "degC")
+      dp <- udunits2::ud.convert(ens[, "d2m"] %>% as.numeric(), "K", "degC")
+      beta <- (112 - (0.1 * t) + dp) / (112 + (0.9 * t))
```
ashiklom: Please change `dp` to `dewpoint` to make the intent of this code clearer.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+ERA5_extract <-
+  function(lat = 40,
+           long = -120,
+           years = c(2006),
+           vars = NULL, 
+           data.folder) {
+    
+    tryCatch({
+      # for each ensemble
+      one.year.out <- c(1:10) %>%
```
ashiklom: Will there always be 10 ensemble members? If so, please define this as a constant (e.g. `ensembles <- seq(1, 10)`) and then use it throughout, rather than re-using `1:10`.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+              xts::xts(all.data.point, order.by = timestamp)
+            })
+          
+          #binding the years
+          point.data <- do.call("rbind.xts", point.data)
+          #Merge mean and the speard
+          return(point.data)
+          
+        }) %>%
+        setNames(c(1:10))
```
ashiklom: See my above comment about using `ensembles` as a variable. Also, should we make these names more informative? E.g. `ERA_ensemble_1, ERA_ensemble_2`, etc.? Purely numeric names like this seem a bit pointless -- if they're just going to be numbers, why not just use the underlying integer indices?

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+      
+      
+    }, error = function(e) {
+      PEcAn.logger::logger.severe(paste0(conditionMessage(e)))
+
```
ashiklom:
```suggestion
```

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+  
+  # Create a data frame with information about the file.  This data frame's format is an internal PEcAn standard, and is stored in the BETY database to
+  # locate the data file. 
+  results <- data.frame(
+    file = "", #Path to the file (added in loop below).
+    host = PEcAn.remote::fqdn(), 
+    mimetype = "application/x-netcdf", 
+    formatname = "CF Meteorology", 
+    startdate = paste0(format(start_date, "%Y-%m-%dT%H:%M:00")), 
+    enddate = paste0(format(end_date, "%Y-%m-%dT%H:%M:00")),
```
ashiklom: `paste0` seems unnecessary here. `format` already returns a string.

Also, can we add the time zone offset (`%z`) to the format string to make it completely unambiguous?

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+    seq(0, (length(index(out.new[[1]])) * 3) - 1 , length.out = length(index(out.new[[1]]))),
+    create_dimvar = TRUE
+  )
+  lat_dim = ncdf4::ncdim_def("latitude", "degree_north", lat, create_dimvar = TRUE)
+  lon_dim = ncdf4::ncdim_def("longitude", "degree_east", long, create_dimvar = TRUE)
+  
+  #create a list of all ens
+  nc_var_list <- map2(cf_var_names,
+                      cf_var_units,
+                      ~ ncdf4::ncvar_def(.x, .y, list(time_dim, lat_dim, lon_dim), missval =NaN)
```
ashiklom: 👍 for the `purrr` syntax for defining NetCDF variables. This is much cleaner than what we do elsewhere in the PEcAn code.

However, I think our convention is usually to use `NA` rather than `NaN` for missing values.

```suggestion
                      ~ ncdf4::ncvar_def(.x, .y, list(time_dim, lat_dim, lon_dim), missval = NA_real_)
```

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+    results$dbfile.name <- flname
+    
+    
+    if (!file.exists(flname) | overwrite) {
+      
+      tryCatch({
+          nc_flptr <- ncdf4::nc_create(flname, nc_var_list, verbose = verbose)
+          
+          #For each variable associated with that ensemble
+          for (j in 1:length(cf_var_names)) {
```
ashiklom: `seq_along` is more concise and safer against `length(...) == 0`.
```suggestion
          for (j in seq_along(cf_var_names)) {
```

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+            "specific_humidity"
+          )
+        )
+      
+    })
+  
+  
+  #These are the cf standard names
+  cf_var_names = colnames(out.new[[1]])
+  cf_var_units = c("K", "Pa", "kgm-2s-1", "ms-1", "ms-1", "Wm-2", "Wm-2", "1")  #Negative numbers indicate negative exponents
```
ashiklom:
```suggestion
  cf_var_units = c("K", "Pa", "kg m-2 s-1", "m s-1", "m s-1", "W m-2", "W m-2", "1")  #Negative numbers indicate negative exponents
```

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+                       year,
+                       ".nc")
+              
+              #msg
+              PEcAn.logger::logger.info(paste0("For ensemble #", ens," ", year, " is being processed !"))
+              #open the file
+              nc_data <- nc_open(ncfile)
+              # time stamp
+              t <- ncvar_get(nc_data, "time")
+              tunits <- ncatt_get(nc_data, 'time')
```
ashiklom:
```suggestion
              tunits <- ncatt_get(nc_data, "time")
```

modules/data.atmosphere/DESCRIPTION
```diff
@@ -52,6 +52,7 @@ Imports:
     truncnorm,
     udunits2 (>= 0.11),
     XML (>= 3.98-1.4),
+    xts,
```
ashiklom: You will probably need to run `scripts/generate_dependencies.R` to update the dependency list for Travis, Docker, etc.

Also, can you add it to the list of binary packages installed by Travis? It should be `r-cran-xts`.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+    name = "time",
+    paste(units = "hours since", format(start_date, "%Y-%m-%dT%H:%M")),
+    seq(0, (length(index(out.new[[1]])) * 3) - 1 , length.out = length(index(out.new[[1]]))),
+    create_dimvar = TRUE
+  )
+  lat_dim = ncdf4::ncdim_def("latitude", "degree_north", lat, create_dimvar = TRUE)
+  lon_dim = ncdf4::ncdim_def("longitude", "degree_east", long, create_dimvar = TRUE)
+  
+  #create a list of all ens
+  nc_var_list <- map2(cf_var_names,
```
ashiklom: Please use explicit namespacing.

```suggestion
  nc_var_list <- purrr::map2(cf_var_names,
```

modules/data.atmosphere/R/met2cf.ERA5.R
ashiklom: Also, I'm a bit wary of cavalier use of `as.numeric` in places like this. If the `ens` object is not numeric, that almost certainly means there was a problem upstream, and the code should fail here rather than performing R's often surprising and counterintuitive type conversion (e.g. `as.numeric(factor(c("4", "5", "6")) == c(1, 2, 3)`).

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+      #Adopted from weathermetrics/R/moisture_conversions.R
+      t <- udunits2::ud.convert(ens[, "t2m"] %>% as.numeric(), "K", "degC")
+      dp <- udunits2::ud.convert(ens[, "d2m"] %>% as.numeric(), "K", "degC")
+      beta <- (112 - (0.1 * t) + dp) / (112 + (0.9 * t))
+      relative.humidity <- 100 * beta ^ 8
+      #specific humidity
+      specific_humidity <- PEcAn.data.atmosphere::rh2qair(relative.humidity,
+                                                          ens[, "t2m"] %>% as.numeric(),
+                                                          ens[, "sp"] %>% as.numeric()
+                                                          ) # Pressure in Pa
```
ashiklom: See above comment about `as.numeric`.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+    }
+    
+    flname <- file.path(ensemble_folder, paste(identifier, "nc", sep = "."))
+    
+    #Each ensemble member gets its own unique data frame, which is stored in results_list
+    results$file <- flname
+    results$dbfile.name <- flname
+    
+    
+    if (!file.exists(flname) | overwrite) {
```
ashiklom: `if/else` statements should _always_ use the double forms of these operators (`&&` and `||`) unless the expression is wrapped in `all/any`.

```suggestion
    if (!file.exists(flname) || overwrite) {
```

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -0,0 +1,171 @@
...
+  lon_dim = ncdf4::ncdim_def("longitude", "degree_east", long, create_dimvar = TRUE)
+  
+  #create a list of all ens
+  nc_var_list <- map2(cf_var_names,
+                      cf_var_units,
+                      ~ ncdf4::ncvar_def(.x, .y, list(time_dim, lat_dim, lon_dim), missval =NaN)
+                      )
+  
+  #For each ensemble
+  results_list <- map(1:10,
```
ashiklom: See my earlier comment about ensembles as a variable.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+                  nn <-
+                    raster::extract(brick.tmp, SpatialPoints(cbind(long, lat)), method = 'simple') %>%
+                    as.numeric()
+                  # replacing the missing/filled values with NA
+                  nn[nn == nc_data$var[[vname]]$missval] <- NA
+                  # send out the extracted var as a new col
+                  nn
+                  
+                }) %>%
+                `colnames<-`(paste0(vars))
```
ashiklom: Unnecessary use of `paste0`.
```suggestion
                `colnames<-`(vars)
```

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+           vars = NULL, 
+           data.folder) {
+    
+    tryCatch({
+      # for each ensemble
+      one.year.out <- c(1:10) %>%
+        map(function(ens) {
+          # for each year
+          point.data <- years %>%
+            map(function(year) {
```
ashiklom: Explicit namespace for `purrr::map`. Also, you can save one indentation by avoiding the pipe here:

```r
point.data <- purrr::map(years, function(year) {
  ... # Only one indent needed
})
```

...as opposed to:

```r
point.data <- years %>%
  map(function(year) {
    ... # two indents needed
  })
```

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+              # set the vars
+              if (is.null(vars))
+                vars <- names(nc_data$var)
+              # for the variables extract the data
+              all.data.point <- vars %>%
+                map_dfc(function(vname) {
+                  brick.tmp <-
+                    brick(ncfile, varname = vname, level = ens)
+                  nn <-
+                    raster::extract(brick.tmp, SpatialPoints(cbind(long, lat)), method = 'simple') %>%
```
ashiklom:
```suggestion
                    raster::extract(brick.tmp, SpatialPoints(cbind(long, lat)), method = "simple") %>%
```

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+                    as.numeric()
```
ashiklom: See below comment about `as.numeric()`. If it's necessary here, fine, but I don't think I've _ever_ encountered data that isn't numeric.

A much safer way to assert this is to use something like this:

```r
if (!is.numeric(nn)) {
  PEcAn.logger::logger.severe(paste0(
    "Expected raster object to be numeric, but it has type `",
    paste0(typeof(nn), collapse = " "), "`"
  ))
}
```

modules/data.atmosphere/DESCRIPTION
para2x: > Also, can you add it to the list of binary packages installed by Travis? It should be r-cran-xts.
how am I supposed to do that ?

ashiklom: Add `r-cran-xts` to [this list in .travis.yml](https://github.com/PecanProject/pecan/blob/develop/.travis.yml#L65-L91).

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -0,0 +1,90 @@
...
+              
+              #msg
+              PEcAn.logger::logger.info(paste0("For ensemble #", ens," ", year, " is being processed !"))
+              #open the file
+              nc_data <- nc_open(ncfile)
+              # time stamp
+              t <- ncvar_get(nc_data, "time")
+              tunits <- ncatt_get(nc_data, 'time')
+              tustr <- strsplit(tunits$units, " ")
+              timestamp = as.POSIXct(t * 3600, tz = 'GMT', origin = tustr[[1]][3])
```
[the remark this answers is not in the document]
ashiklom: I'm going to insist on the `GMT->UTC` change. `UTC` is used throughout our met code, and I think it's important that we be consistent.

para2x:
No problem, but I guess before merging it we have a bit of problem with Travis that we need to figure it out !

tonygardella:
@robkooper This Travis build is failing with:
```

serbinsh:
Updating to hopefully get this merged so we can try it out at BNL for CMS @para2x

para2x:
@tonygardella this is all yours now.

serbinsh:
Cool, this is ready to test out now!?

para2x:
Sorry to disappoint you @serbinsh but no. This was the required functions to make the nc file. ERA5 also needs to be added to do_conversion and met process basically.

serbinsh:
Ahhh....ok then :)

tonygardella:
@serbinsh I discussed this a little with @para2x . Looks like we'd need to follow how we handle NARR because this downloads >100gb of data locally then, parses the data. So ideally we'd use Browndog to handle this since we don't want anyone required to download all the data locally.

serbinsh:
@tonygardella Oh....ok.  That said, BD is a slow process :( as it is now, for CMS runs, can I use this and just deal with all of the local files?

tonygardella:
@serbinsh Yeah but it's not automated yet from the PEcAn. So you'd need to run the function manually and feed it into a met2model manually.

para2x:
@serbinsh since it's not implemented inside the do_conversion, you can deal with local files but you also need to add them manually to your xml. I mean, this will generate the met nc files and then you can manually put them into your settings.

serbinsh:
Yup, understood.  I will give it a go......but realize its "buyer beware"

PR #2306, opened 2019-03-04
rstudio behind nginx-proxy

robkooper:
put rstudio behind nginx to make /rstudio work. See https://docs.rstudio.com/ide/server-pro/access-and-security.html#running-with-a-proxy

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -157,8 +157,19 @@ services:
       - "traefik.backend=bety"
 
   # ----------------------------------------------------------------------
-  # Pecan-RStudio
+  # RStudio
   # ----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -167,15 +178,10 @@ services:
     environment:
       - USER=${PECAN_RSTUDIO_USER:-carya}
       - PASSWORD=${PECAN_RSTUDIO_PASS:-ilinois}
-    depends_on:
```
Chris-Schnaufer: Should we depend upon rstudio-nginx?

robkooper: mmmm, maybe, it will start fine without it, so there is no need for you to wait for it to be up.

PR #2309, opened 2019-03-07
Site PFT improvement

para2x:
Playing around with site PFT linker feature I realized there is a chance that even a space after pft name defined in the csv file by user  would result in crashing the prepare settings . Here is a small PR to avoid that.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/settings/R/site_pft_link_settings.R
```diff
@@ -26,7 +27,7 @@ site.pft.link.settings <-function(settings){
     def.pfts <- settings$pfts %>% purrr::map('name') %>% unlist() %>% as.character()
     
     # Create a simple pft tag for the pfts in LUT that are not in the pft tag
-    pft.l <- LUT$pft [!(LUT$pft %in% def.pfts)]
+    pft.l <- LUT$pft [!(LUT$pft %in% def.pfts)] %>% trimws() %>% unique()
     
     new.pfts <-  pft.l%>%
       purrr::map(function(lut.pft) {
```
ashiklom: Is the goal here to drop elements of `pft.l` that are not in `def.pfts`? As written, this code will replace all cases where the `if` statement is `FALSE` with `NULL` (per the behavior of `if`), which is not quite the same thing.

I would probably simplify this as:

```r
new.pfts <- pft.l %>%
  purrr::drop(~.x %in% def.pfts) %>%
  purrr::map(~list(name = as.character(.x), constants = 1)) %>%
  setNames(rep("pft", length(pft.l))
```

Or even something like:

```r
new.pfts <- pft.l[!pft.l %in% def.pfts]  %>%
  purrr::map(~list(name = as.character(.x), constants = 1)) %>%
  setNames(rep("pft", length(pft.l))
```

para2x:
@mdietze this PR is ready to be reviewed.

mdietze:
@ashiklom this needs you to check whether the requested changes have been made and then approve

PR #2313, opened 2019-03-13
HOT FIX: fixed MODISTools to v1.1.0

bailsofhay:
MODISTools updated to v1.1.0 from 1.0.0 and altered the output format of the data extracted. I fixed the code to account for this change and added in a note about a dependency for the MODISTools package to be v1.1.0 for future users/changes.

## Description
MODISTools update broke my code. I fixed it.
## Motivation and Context
look above........
## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/data.remote/R/call_MODIS.R
```diff
@@ -85,41 +86,32 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
-    headers = (dat$header)
-    scale = as.numeric(headers$scale)
-    # change lat and lon once we start outputing rasters, the lat and lon will not be the same for all data based on input lat and lon
-    latitude = as.character(rep(headers$latitude, nrow(data)))
-    longitude = as.character(rep(headers$longitude, nrow(data)))
-
     if (band_qc == "")
     {
-      QC = rep("nan", nrow(data))
+      QC = rep("nan", nrow(dat))
```
serbinsh: @bailsofhay can you please change "=" to "<-" to be consistent with the code in PEcAn?  This only pertains to objects not function calls.  That is if in the function [e.g. call_MODIS(arg1 = XXX) ] is correct.

bailsofhay: uggggggggggggggggg.... fine!

modules/data.remote/R/call_MODIS.R
```diff
@@ -44,8 +45,8 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
     }
     
 
-    dates = MODISTools::mt_dates(product =product, lat = lat, lon = lon)$modis_date
-    dates = as.numeric(substr(dates, 2, nchar(dates)))
+    dates <- MODISTools::mt_dates(product =product, lat = lat, lon = lon)$modis_date
```
infotroph:
```suggestion
    dates <- MODISTools::mt_dates(product = product, lat = lat, lon = lon)$modis_date
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -85,42 +86,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
+      QC <- as.numeric(qc$value)
     }
     
     if (band_sd == "")
     {
-      SD = rep("nan", nrow(data))
+      SD <- rep("nan", nrow(dat))
     } else {
-      SD = as.numeric(sd$data$data)*scale #formatC(sd$data$data*scale, digits = 2, format = 'f')
+      SD <- as.numeric(sd$value)*as.numeric(sd$scale) #formatC(sd$data$data*scale, digits = 2, format = 'f')
```
infotroph:
```suggestion
      SD <- as.numeric(sd$value) * as.numeric(sd$scale) #formatC(sd$data$data*scale, digits = 2, format = 'f')
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -85,42 +86,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
     
-    output[,5:10] = lapply(output[,5:10], as.numeric)
+    output[,5:10] <- lapply(output[,5:10], as.numeric)
     
     # scale the data + stdev to proper units
-    output$data = output$data*scale
-    output$sd = output$sd*scale
-    output$lat = round(output$lat, 4)
-    output$lon = round(output$lon, 4)
+    output$data <- output$data*(as.numeric(dat$scale))
```
infotroph:
```suggestion
    output$data <- output$data * (as.numeric(dat$scale))
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -85,42 +86,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
+    output$sd <- output$sd*(as.numeric(dat$scale))
```
infotroph:
```suggestion
    output$sd <- output$sd * (as.numeric(dat$scale))
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -16,20 +16,21 @@
 ##' @param package_method string value to inform function of which package method to use to download modis data. Either "MODISTools" or "reticulate" (optional)
 ##' 
 ##' depends on a number of Python libraries. sudo -H pip install numpy suds netCDF4 json
+##' depends on the MODISTools package version 1.1.0
```
infotroph: Can you add a version spec to the DESCRIPTION line as well, please?

Currently:
```
Imports:
    MODISTools,
    reticulate,
   ...
```

Change to:
```
Imports:
    MODISTools (>= 1.1.0),
    reticulate,
    ...
```

serbinsh: Good catch @infotroph I had asked this separately on slack...given the breaking change from 1 to 1.1

bailsofhay: added it. But I don't know if >=1.1.0 will end up breaking it. so I made it ==1.1.0 for now.... if that even works

modules/data.remote/R/call_MODIS.R
bailsofhay: fixed this

bailsofhay:
I think I've gotten all the tiny formatting fixes done and added in the dependency to the DESCRIPTION file. @serbinsh we can work on the bety thing when you are back next week :)

modules/data.remote/R/call_MODIS.R
infotroph: `==` works, but for the record I recommend defaulting to `>=` in most cases like this, because:

* Nonbreaking updates are more common than breaking ones, so letting an upgrade Just Work usually saves time and trouble... and you can't usually know which upgrades will be the exceptions until they're released anyway.
* Since R doesn't like having multiple versions of the same package installed at the same time, requiring an exact version means headaches as soon as you install some other package that wants, say, MODISTools >= 1.1.2.

bailsofhay:
All of the suggestions have been fixed and the script was roxygenized successfully and passed devtools::check.

PR #2316, opened 2019-03-14
PEcAn.DB cleanup

ashiklom:
## Description

A lot of this is code style and documentation cleanup and additional logging that shouldn't affect function. The most significant _functional_ change is that `db.open` should now also support the `RPostgres` package, which is a newer (and, I would argue, better) version of the old `RPostgreSQL` package.

## Motivation and Context

`RPostgres` is better maintained, integrates more nicely with R, and, most importantly (IMHO) supports parameterized queries, which are both easier to write _and_ safer. Note that `pecanapi` already uses `RPostgres` under the hood.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/db/R/utils_db.R
+++ b/base/db/R/utils_db.R
@@ -113,7 +113,7 @@ db.open <- function(params) {
   if (!driver %in% c("PostgreSQL", "Postgres")) {
     PEcAn.logger::logger.severe(paste0(
       "Driver `", driver, "` is not supported. ",
-      "You must use either `RPostgreSQL` or `Postrgres`."
+      'You must use either `"PostgreSQL"` or `"Postrgres"`.'
     ))
   }
 
--- a/base/db/R/utils_db.R
+++ b/base/db/R/utils_db.R
@@ -113,7 +113,7 @@ db.open <- function(params) {
   if (!driver %in% c("PostgreSQL", "Postgres")) {
     PEcAn.logger::logger.severe(paste0(
       "Driver `", driver, "` is not supported. ",
-      'You must use either `"PostgreSQL"` or `"Postrgres"`.'
+      'You must use either `"PostgreSQL"` or `"Postgres"`.'
     ))
   }
 
--- a/base/db/R/utils_db.R
+++ b/base/db/R/utils_db.R
@@ -21,14 +21,15 @@
 ##' to be specified. If both are specified it will use con.
 ##'
 ##' @param query SQL query string
-##' @param con database connection object
-##' @param params database connection information
+##' @param con Database connection object
+##' @param params Named list of database connection parameters. See
+##'   `params` argument to [db.open()].
 ##' @return data frame with query results
-##' @author Rob Kooper
+##' @author Rob Kooper, Alexey Shiklomanov
 ##' @export
 ##' @examples
 ##' \dontrun{
-##' db.query('select count(id) from traits;', params=settings$database$bety)
+##' db.query("SELECT count(id) FROM traits;", params = settings$database$bety)
 ##' }
 db.query <- function(query, con = NULL, params = NULL) {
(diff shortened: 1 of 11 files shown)


base/db/R/utils_db.R
```diff
@@ -99,20 +145,18 @@ db.open <- function(params) {
...
 ##' @export
 ##' @examples
 ##' \dontrun{
 ##' db.close(con)
 ##' }
-db.close <- function(con, showWarnings=TRUE) {
+db.close <- function(con, showWarnings = TRUE) {
   if (is.null(con)) {
-    return()
+    return(invisible(TRUE))
```
ashiklom: This makes the output consistent whether or not the connection is `NULL`.

base/db/R/utils_db.R
```diff
@@ -187,100 +228,96 @@ db.exists <- function(params, write=TRUE, table=NA) {
...
+  # We're enquiring about permissions related to a specific table, so
+  # need to do more here.
+  read.perm <- FALSE
+  write.perm <- FALSE
 
-    #check write permission
-    if ('INSERT' %in% user.permission[['privilege_type']] &&'UPDATE' %in% user.permission[['privilege_type']] ) {
-      write.perm = TRUE
-    }
+  # check read permission
```
ashiklom: The code here is the same, just done in a different order that, IMHO, makes more sense: First, do the complete check of read permissions (and return early if that fails). Then, check all the write permissions. (Before, they were interspersed).

base/db/R/utils_db.R
```diff
@@ -44,51 +43,98 @@ db.query <- function(query, con=NULL, params=NULL) {
...
 
-  if (is.null(params$driver)) {
-    args <- c(drv = DBI::dbDriver("PostgreSQL"), params, recursive = TRUE)
-  } else {
-    args <- c(drv = DBI::dbDriver(params$driver), params, recursive = TRUE)
-    args[['driver']] <- NULL
+  if (!driver %in% c("PostgreSQL", "Postgres")) {
+    PEcAn.logger::logger.severe(paste0(
+      "Driver `", driver, "` is not supported. ",
+      'You must use either `"PostgreSQL"` or `"Postrgres"`.'
```
infotroph: Is `postrgres` a typo or a driver I haven't seen before?

base/db/R/utils_db.R
```diff
@@ -44,51 +43,98 @@ db.query <- function(query, con=NULL, params=NULL) {
...
+  args <- c(drv = drv, params, recursive = TRUE)
+
   c <- do.call(DBI::dbConnect, as.list(args))
-  id <- sample(1000, size=1)
-  while(length(which(.db.utils$connections$id==id)) != 0) {
-    id <- sample(1000, size=1)
+
+  # Assign the connection a unique ID number between 0 and 1000
+  id <- sample(1000, size = 1)
+  while (sum(.db.utils$connections$id == id) > 0) {
```
infotroph:
```suggestion
  while (any(.db.utils$connections$id == id)) {
```

base/db/R/utils_db.R
ashiklom: Good catch! I was _this_ close to sending you a link about the `RPostgres` package before realizing that this was, indeed, a typo.

ashiklom: Fixed in 993b387dc0214ece8ee2c343793e4b0220156bec.

infotroph: It totally scans like an R package name. `postrGREs` is the data package a presentation about standardized tests, `postRgres` is for post-hoc calculations on continuous linear models, and `postrGres` is the R frontend for a database of who's most active in your Tumblr comments.

PR #2318, opened 2019-03-15
Full support for `RPostgres`

ashiklom:
(NOTE: This PR is a superset of #2317 and #2316. Those should be reviewed and merged first).

## Description

Minor tweaks that allow workflows to use the `RPostgres::Postgres` database backend. For the most part, these changes are driven by the fact that `RPostgres` returns long integer columns as the `integer64` class, which is basically a `double` with some additional attributes. This causes problems because many low-level base R functions (`for (...)`, `apply`, `cat`) will silently strip away those attributes, leaving gibberish doubles that break downstream code.

Note that this will only affect users that _explicitly_ say they want to use `Postgres` (via `<driver>Postgres</driver>`. Nothing that defaults to `PostgreSQL` has changed. However, I think it's worth moving in that direction, so once this gets merged, I would encourage folks to try testing new code against the `Postgres` backend.

## Motivation and Context

Using prepared statements (#2317) requires the `RPostgres` backend.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/uncertainty/R/ensemble.R
+++ b/modules/uncertainty/R/ensemble.R
@@ -42,8 +42,8 @@ read.ensemble.output <- function(ensemble.size, pecandir, outdir, start.year, en
   ensemble.output <- list()
   for (row in rownames(ens.run.ids)) {
     run.id <- ens.run.ids[row, "id"]
-    PEcAn.logger::logger.info("reading ensemble output from run id: ", run.id)
-    
+    PEcAn.logger::logger.info("reading ensemble output from run id: ", format(run.id))
+
     for(var in seq_along(variables)){
       out.tmp <- PEcAn.utils::read.output(run.id, file.path(outdir, run.id), start.year, end.year, variables[var])
       assign(variables[var], out.tmp[[variables[var]]])
@@ -365,15 +365,15 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
       dir.create(file.path(settings$modeloutdir, run.id), recursive = TRUE)
       # write run information to disk
       cat("runtype     : ensemble\n",
-          "workflow id : ", workflow.id, "\n",
-          "ensemble id : ", ensemble.id, "\n",
+          "workflow id : ", format(workflow.id), "\n",
+          "ensemble id : ", format(ensemble.id), "\n",
           "run         : ", i, "/", settings$ensemble$size, "\n",
-          "run id      : ", run.id, "\n",
-          "pft names   : ", as.character(lapply(settings$pfts, function(x) x[['name']])), "\n",
+          "run id      : ", format(run.id), "\n",
+          "pft names   : ", as.character(lapply(settings$pfts, function(x) x[["name"]])), "\n",
           "model       : ", model, "\n",
-          "model id    : ", settings$model$id, "\n",
+          "model id    : ", format(settings$model$id), "\n",
           "site        : ", settings$run$site$name, "\n",
-          "site  id    : ", settings$run$site$id, "\n",
+          "site  id    : ", format(settings$run$site$id), "\n",
           "met data    : ", samples$met$samples[[i]], "\n",
           "start date  : ", settings$run$start.date, "\n",
           "end date    : ", settings$run$end.date, "\n",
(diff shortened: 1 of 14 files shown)


modules/uncertainty/R/ensemble.R
```diff
@@ -395,8 +395,8 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
                                             run.id = run.id
       )
       )
-      cat(run.id, file = file.path(settings$rundir, "runs.txt"), sep = "\n", append = TRUE)
-      
+      cat(format(run.id), file = file.path(settings$rundir, "runs.txt"), sep = "\n", append = TRUE)
```
infotroph: It's not clear to me what format is doing here -- is this to preserve the right representation of int64s? would explicit formatting arguments make that clearer?

ashiklom:
``` r
i <- bit64::as.integer64(9900000001)
cat(i)
#> 4.89125e-314
format(i)
#> [1] "9900000001"
```

My understanding is that there's a special `format.integer64` method that makes this printing work.

robkooper: I think the problem is that you cast it to a string with the format function:
```
> i <- 9900000001
> cat(i)
9.9e+09
> typeof(i)
[1] "double"
> typeof(bit64::as.integer64(i))
[1] "double"
> typeof(format(i))
[1] "character"
> options(scipen=12)
> cat(i)
9900000001
```
Would it make sense to set scipen to be 12 in PEcAn db package.

ashiklom: @robkooper The problem is that `RPostgres` returns long integers as `bit64::integer64`, which is _not_ numeric. Here are some examples illustrating the issues with 64bit integers:

``` r
# Working with doubles is fine...
i <- 9900000001
class(i)
#> [1] "numeric"
typeof(i)
#> [1] "double"
cat(i, "\n")
#> 9.9e+09

# Working with integer64s is more complicated...
i64 <- bit64::as.integer64(i)
class(i64)
#> [1] "integer64"
typeof(i64)
#> [1] "double"

# Under the hood, this uses similar memory mapping to numerics, but
# with a very different meaning. As a result, when the class attribute
# is removed, integer64s are gibberish numerics... 
unclass(i64)
#> [1] 4.89125e-314

# ...which causes weird behavior when printing.
cat(i64, "\n")
#> 4.89125e-314

# This works as expected
cat(format(i64), "\n")
#> 9900000001

# Similarly, base R's `for` loop (and `lapply`) does not dispatch on
# `[[`, so it drops the class attribute, leading to problems
for (j in i64) print(j)
#> [1] 4.89125e-314
lapply(i64, function(x) x)
#> [[1]]
#> [1] 4.89125e-314

# Even `vapply` doesn't work.
vapply(i64, function(x) x, bit64::integer64(1))
#> [1] 4.89125e-314

# Have to either subset yourself...
for (jj in seq_along(i64)) print(i64[[jj]])
#> integer64
#> [1] 9900000001

# ...or use `purrr::map`
purrr::map(i64, ~.x)
#> [[1]]
#> integer64
#> [1] 9900000001

# ...but be careful with `unlist`, `purrr::simplify`, and friends,
# because they break it too!
l <- purrr::map(i64, ~.x)
l
#> [[1]]
#> integer64
#> [1] 9900000001
unlist(l)
#> [1] 4.89125e-314
purrr::simplify(l)
#> [1] 4.89125e-314
purrr::as_vector(l)
#> [1] 4.89125e-314
```

ashiklom: Re: That last point -- the only reliable way I can see to `unlist` `integer64` values is via `do.call(c, i64)`.

ashiklom:
@infotroph @robkooper Were there any changes you wanted me to make on this, or can it be merged?

infotroph:
@ashiklom good on my end

tonygardella:
@ashiklom this was passing Travis earlier, but now it has failed twice during the basic sipnet run:
```
2019-04-29 16:04:41 WARN   [PEcAn.data.atmosphere::met.process] : 
   met.process only has a path provided, assuming path is model driver and 
   skipping processing 
2019-04-29 16:04:41 DEBUG  [PEcAn.workflow::do_conversions] : 
   updated met path: /home/travis/sipnet_unk/niwot_tutorial.clim 
Error in postgresqlExecStatement(conn, statement, ...) : 
  RS-DBI driver: (could not Retrieve the result : ERROR:  syntax error at or near "AS"
LINE 4: WHERE ("name" = ('temperate.coniferous' AS "name", '/home/tr...
                                                ^
)
Calls: <Anonymous> ... dbSendQuery -> dbSendQuery -> postgresqlExecStatement
----------------------------------------------------------------------
TEST travis.sipnet.xml FAILED (ERROR IN STATUS)
----------------------------------------------------------------------
TRAIT	2019-04-29 16:04:41	2019-04-29 16:04:41	ERROR	
```

PR #2320, opened 2019-03-19
Dockerize biocro

kimberlyh66:
## Description
I have added a Dockerfile to the pecan/models/biocro directory. The BIOCRO package depends on many other PEcAn packages (PEcAn.logger, PEcAn.utils, etc.), so the build context for the image will be from the main pecan directory. 

The nneo  package is needed to build the PEcAn.data.atmopshere package, but is no longer available for download from CRAN. I have added an archived version from https://cran.r-project.org/src/contrib/Archive/nneo/ to the pecan/modules directory so that the package can be built locally (not sure if this is what you would like done).

To build image, cd to main pecan directory:
`docker build -t biocro-docker -f ./models/biocro/Dockerfile .` to build the image. 

To run container: 
`docker run -v /local/path/to/rundir:/data/rundir -v /local/path/to/outdir:/data/outdir biocro-docker rundir outdir` 

## Motivation and Context
* Dockerize biocro
* Be able to pass in input and output directories and run biocro.Rscript

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


robkooper:
couple of high level comments, you have the tar.gz file for nneo checked in, as well as the completely unpacked version, best is to add only 1. I don't think we should add the nneo to our tree since we don't want to maintain it. If anything we can add the tar.gz file somewhere and tell people to download it from there. That being said we use ropensci/nneo from github (see https://github.com/PecanProject/pecan/blob/develop/docker/depends/pecan.depends#L17) to install it.

Your dockerfile is pretty large and will take a decent amount of time to build, the default pecan/base already has BioCro installed, so you might just need to build from pecan/models and add the model_info.json in this case.
```
docker run -ti --rm pecan/base:develop Rscript -e 'packageVersion("BioCro")'
[1] ‘0.95’
```

kimberlyh66:
@robkooper thanks for the feedback. 

A couple of questions:
Do you mean to use pecan/models as the base image? 
Where does the model_info.json come from?

robkooper:
pecan/depends == all dependencies for PEcAn
pecan/base == pecan/depends + all of PEcAn compiled
pecan/models ==pecan/base + python and script to listen for rabbitmq messages and advertise models using model_info.json
pecan/models-biocro-095 = pecan/models + model_info.json for biocro

For information about model_info.json see https://pecanproject.github.io/pecan-documentation/develop/docker-index.html#model-docker

You can also look at ED for examples of the Dockerfile and model_info.json : https://github.com/PecanProject/pecan/tree/develop/models/ed

robkooper:
This should be merged using a merge-squash to make sure we don't pull in any additional commits (such as nneo).

models/biocro/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM pecan/models:develop
+
+RUN mkdir /data
+
+COPY ./model_info.json /work
```
KristinaRiemer: Trying to build the image failed, I think because this path needs to be `./models/biocro/model_info.json`?

kimberlyh66: @KristinaRiemer The image should be built in the pecan/models/biocro directory. I made some changes so the image can no longer be built in pecan.

KristinaRiemer:
Successfully got to `Error: XML content does not seem to be XML: 'rundir/config.xml'`

robkooper:
This will not work with PEcAn since you override the CMD. Will do some testing later today to make sure it works.

models/biocro/model_info.json
```diff
@@ -0,0 +1,15 @@
...
+  "binary": "@BINARY@",
+  "description": "BioCro is a model that predicts plant growth over time given climate as input",
+  "author": "David LeBauer",
+  "contributors": ["Deepak Jaiswal", "Christopher Black"],
+  "links": {
+  	"source": "https://github.com/ebimodeling/biocro",
+  	"issues": "https://github.com/ebimodeling/biocro/issues"
+  },
+  "inputs": {},
+  "citation": []
```
dlebauer: add citations from https://github.com/ebimodeling/biocro#references

dlebauer:
@robkooper could you please take another look to see if recent changes are what you were asking for?

kimberlyh66:
@dlebauer I think this pull request should be closed without merging based on our conversation during last meeting. I will be using the biocro_helpers R script you sent over, so there is no need for the PEcAn dependencies.

robkooper:
Question is do you want to dockerize biocro so it runs independent of PEcAn, or do you want to make it  so it is part of PEcAn?
If you want it to be independent of PEcAn I would recommend to start with one of the rocker docker images, and install the biocro package in there.
If you want it to be part of PEcAn you will need to start from pecan/models (as done in this pull request). However you can't override the CMD since that original one will connect it to PEcAn and will need to make sure that you set MODEL_VERSION and MODEL_BINARY environment variables.

kimberlyh66:
I would like to dockerize biocro so that it runs independent of PEcAn

robkooper:
In that case I think the pull request should be made against biocro.

robkooper:
@kimberlyh66 and @dlebauer closing this pull request. I created a new pull request (#2343) with @kimberlyh66 as author that creates the dockerfile needed for PEcAn.

PR #2328, opened 2019-03-25
Fix for ED config 

para2x:
## Description

This PR is going to fix the css/pss path problem in settings for ED run. There is also another fix in settings package and specifically `site_pft_link_settings.R`.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


para2x:
@serbinsh see if this solves the drop problem you had within the prepare settings. `drop` was function suggested by Alexy but not exported from purrr.

serbinsh:
Great, will do @para2x

base/settings/R/site_pft_link_settings.R
```diff
@@ -34,7 +34,7 @@ site.pft.link.settings <- function(settings) {
        unique()
      
      new.pfts <-  pft.l %>%
-       purrr::drop(~.x %in% def.pfts) %>%
+       purrr::discard(~.x %in% def.pfts) %>%
```
serbinsh: @bailsofhay can you make this change in your version of this file in pecan then re-install pecan settings then trying running the LAI SDA where you actually run the prepare settings step and see if that fixes your current bug?

serbinsh: This change fixed our issue with purrr in prepare.settings

PR #2336, opened 2019-04-03
pika version 1.0.0

robkooper:
upgrades pika to version 1.0.0
pins pika to version 1.0.0

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


docker/models/model.py
```diff
@@ -110,7 +110,7 @@ def receiver():
 
     # receive 1 message at a time, call callback function
     channel.basic_qos(prefetch_count=1)
-    channel.basic_consume(callback, queue=rabbitmq_queue)
+    channel.basic_consume(on_message_callback=callback, queue=rabbitmq_queue)
```
ashiklom: This same change also needs to be in the `executor.py` file:

https://github.com/PecanProject/pecan/blob/810f4f0f601e76ef6e06c9df494dba7faba863b6/docker/executor/executor.py#L113

docker/monitor/Dockerfile
```diff
@@ -8,12 +8,10 @@ ENV RABBITMQ_URI="amqp://guest:guest@rabbitmq/%2F" \
...
-RUN pip install pika==0.11.2 \
-                requests==2.18.4 \
-                psycopg2-binary==2.7.6.1 \
-                python-dateutil==2.7.5
-
 WORKDIR /src
-COPY monitor.py /src/
 
+COPY requirements.txt /src/
+RUN pip install -r /src/requirements.txt
```
ashiklom: Should we specify `pip3` here, as we do for `executor`?

docker/monitor/Dockerfile
```diff
@@ -8,12 +8,10 @@ ENV RABBITMQ_URI="amqp://guest:guest@rabbitmq/%2F" \
...
+
+COPY monitor.py /src/
 CMD python monitor.py
```
ashiklom: Similarly to above, should we explicitly call `python3` here?

PR #2337, opened 2019-04-06
Updated RTM docs

serbinsh:
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/rtm/R/prospect.R
```diff
@@ -6,7 +6,8 @@
 #'     * N: Effective number of leaf layers (>1)
 #'     * Cab: Leaf chlorophyll content (ug/cm2) (>0)
 #'     * (5) Car: Leaf carotenoid content (ug/cm2) (>0)
-#'     * (5B) Cbrown: Leaf brown matter content (ug/cm2) (>0)
+#'     * (D) Anth: Leaf anthocyanin content (ug/cm2) (>0)
```
serbinsh: Or should this be Canth?

modules/rtm/R/prospect.R
```diff
@@ -66,6 +67,11 @@ params.prospect5 <- c("N", "Cab", "Car", "Cw", "Cm")
 #' @export
 params.prospect5b <- c("N", "Cab", "Car", "Cbrown", "Cw", "Cm")
 
+#' @name params.prospectd
+#' @title PROSPECT D parameters
+#' @export
+params.prospectd <- c("N", "Cab", "Car", "Anth", "Cbrown", "Cw", "Cm")
```
serbinsh: same here, should this be Canth?

serbinsh:
@ashiklom we should check that the FOURSAIL function we have in RTM is the most current or perhaps provide version wrappers to newer code?

serbinsh:
@ashiklom I updated the docs.  The only thing I wasnt sure about was adding defaults for pro4saild in the model_list.csv.  Wasnt sure what "modcode" should be

ashiklom:
2113 would make sense, but honestly, I don't think anything uses that anymore. Way back in the day, when I had coded up a Metropolis sampler in Fortran, I used the modcode because R couldn't reliably pass character strings to Fortran, and I needed to pass it to tell Fortran which model to invert. But since we use BayesianTools these days, there's no need for it anymore.

serbinsh:
Attempted to squash commits into 1.

PR #2338, opened 2019-04-08
MODIStools: cleaned up function and added QC filter parameter

bailsofhay:
I've cleaned up my original code for call_modis and added some new features.

## Description
This function matches more closely to Shawn LandTrendr function. It now can take in a list of multiple sites vs. just looping the function through individual xy coords. There is also now an option to parallelize the function.

## Motivation and Context
Easier to use for more complex situations with the SDA

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/data.remote/R/call_MODIS.R
```diff
@@ -27,65 +29,116 @@
...
 ##'  
-call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0, product, band, band_qc = "", band_sd = "", package_method = "MODISTools") {
+call_MODIS <- function(outfolder = NULL, start_date, end_date, lat, lon, size = 0, product, band, band_qc = "", band_sd = "", siteID = NULL, package_method = "MODISTools", QC_filter = F) {
   
   # makes the query search for 1 pixel and not for rasters for now. Will be changed when we provide raster output support.
   size <- 0
+# reformat start and end date if they are in YYYY/MM/DD format instead of YYYYJJJ
+  if (grepl("/", start_date) == T)
+  {
+    start_date = as.Date(paste0(lubridate::year(start_date), spatial.tools::add_leading_zeroes(lubridate::yday(start_date), 3)), format = "%Y%j")
```
ashiklom: Unless I'm missing something here, I think you can eliminate the `spatial.tools` dependency here by using base R's `sprintf`:

```r
as.Date(sprintf("%04d%03d", lubridate::year(start_date), lubridate::yday(start_date)), format = "%Y%j")
```

The syntax is a bit confusing so, parsing out `%03d`:
- `%` indicates the beginning of a string substitution
- `0` means you'll be padding any extra width with zeros 
- `3` is the target string width
- `d` means it expects a number ("double")

modules/data.remote/R/call_MODIS.R
```diff
@@ -129,8 +204,12 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
     output$lat <- round(output$lat, 4)
     output$lon <- round(output$lon, 4)
     
-    fname <- paste(product, "_", band, "_", start_date, "_", end_date, "_", lat, "_", lon, ".csv", sep = "")
-    fname <- paste0(outfolder, "/", fname)
-    write.csv(output, fname)
+    if (!(is.null(outfolder)))
+    {
+      fname <- paste(product, "_", band, "_", start_date, "_", end_date, "_", lat, "_", lon, ".csv", sep = "")
+      fname <- paste0(outfolder, "/", fname)
```
ashiklom: Strictly speaking, it's safer to create paths with `file.path`:
```suggestion
      fname <- file.path(outfolder, fname)
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +164,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
+        output = output[good,]
+      } else {
+        print("All QC values are bad. No data to output with QC filter == TRUE.")
+      }
+    }
+    
+    if (!(outfolder == ""))
+    {
+      fname <- paste(product, "_", band, "_", siteID, "_output_", start_date, "_", end_date, ".csv", sep = "")
+      fname <- paste0(outfolder, "/", fname)
```
ashiklom: See below comment about paths:
```suggestion
      fname <- file.path(outfolder, fname)
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +164,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
-
+    if (QC_filter == T)
+    {
+      output$qc == as.character(output$qc)
+      for (i in 1:nrow(output))
+      {
+        convert = paste(binaryLogic::as.binary(as.integer(output$qc[i]), n = 8), collapse = "")
+        output$qc[i] = substr(convert, nchar(convert)-2, nchar(convert))
+      }
+      good = which(output$qc == "000" | output$qc == "001")
```
ashiklom: Alternatively, you can do:
```r
good <- which(output$qc %in% c("000", "001"))
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +164,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
     
-    fname <- paste(product, "_", band, "output_", start_date, "_", end_date, "_", lat, "_", lon, ".csv", sep = "")
-    fname <- paste0(outfolder, "/", fname)
-    write.csv(output, fname)
-    return(output)}
-
+    if (QC_filter == T)
+    {
+      output$qc == as.character(output$qc)
+      for (i in 1:nrow(output))
```
ashiklom: `seq_len` is safer for the edge case where `nrow(output) == 0`:
```suggestion
      for (i in seq_len(nrow(output)))
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +164,33 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
     output$lat <- round(output$lat, 4)
     output$lon <- round(output$lon, 4)
     
-    fname <- paste(product, "_", band, "output_", start_date, "_", end_date, "_", lat, "_", lon, ".csv", sep = "")
-    fname <- paste0(outfolder, "/", fname)
-    write.csv(output, fname)
-    return(output)}
-
+    if (QC_filter == T)
```
ashiklom: It's already boolean -- no need to test for equality with `T`.
```suggestion
    if (QC_filter)
```

If you want to ensure it's a boolean and nothing else, you can use `if(isTRUE(QC_filter))`. Case in point:

``` r
1 == TRUE
#> [1] TRUE
isTRUE(1)
#> [1] FALSE
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -27,65 +29,116 @@
...
+    {
+      if (as.numeric(start_date) >= dates[1] & as.numeric(end_date) <= dates[length(dates)])
+      {
+        print("Extracting data")
+        
+        start <- as.Date(start_date, "%Y%j")
+        end <- as.Date(end_date, "%Y%j")
+      }
+      ########## For Date case 1: if only one date is asked for, but date is not within modis data prodate date range ##########
+      if (as.numeric(start_date)< dates[1] | as.numeric(end_date)> dates[length(dates)])
```
ashiklom: Inside of `if` statements, you _always_ want to be using the "double" versions of R's "and/or" operators (i.e. `&&` or `||`).

The single character forms -- `&` and `|` -- are vectorized over their inputs, so they return a vector. `if` doesn't know what to do with a logical vector, so if it gets a vector, it will throw a warning and use only the first value.

``` r
x <- sample(1:3, 10, TRUE)
y <- sample(1:2, 10, TRUE)
x == 1 & y == 2
#>  [1] FALSE FALSE FALSE  TRUE FALSE FALSE FALSE FALSE FALSE FALSE
if (x == 1 & y == 2) {
  print("yes")
} else {
  print("no")
}
#> Warning in if (x == 1 & y == 2) {: the condition has length > 1 and only
#> the first element will be used
#> [1] "no"
```

The double character forms (`&&` and `||`) evaluate strictly from left to right, stopping as early as possible and returning a scalar result.

```r

x == 1 && y == 2
#> [1] FALSE
FALSE && stop("This is an error")
#> [1] FALSE
```

If you are doing stuff with `if` and vectors, you generally will want to use the `any` and/or `all` functions. E.g.

```r
if (any(x == 3) && all(y > 0)) { # do stuff...
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -27,65 +29,116 @@
...
   
-  # set start and end dates to correct format
-  if (package_method == "MODISTools"){
-    
+  if (grepl("/", end_date) == T)
+  {
+    end_date = as.Date(paste0(lubridate::year(end_date), spatial.tools::add_leading_zeroes(lubridate::yday(end_date), 3)), format = "%Y%j")
+  }
+  start_date = as.Date(start_date, format = "%Y%j")
+  end_date = as.Date(end_date, format = "%Y%j")
```
ashiklom: I'm Mr. Annoying Style Cop, but please follow our style guide and use `<-` for assignment and reserve `=` for function arguments. Similarly, please use `TRUE/FALSE` instead of `T/F`. The former are reserved words in R, while the latter are not:

 ``` r
# These will error 
TRUE <- 0
#> Error in TRUE <- 0: invalid (do_set) left-hand side to assignment
FALSE <- 5
#> Error in FALSE <- 5: invalid (do_set) left-hand side to assignment
# ...but these will not!
T <- FALSE
F <- 3
if (T) {
  print("yes")
} else {
  print("no")
}
#> [1] "no"
if (F) print("This is true!")
#> [1] "This is true!"
```

File: modules/data.remote/R/call_MODIS.R [hunk 9]
```diff
@@ -14,6 +14,8 @@
 ##' @param band_qc string value for which quality control band, or use "NA" if you do not know or do not need QC information (optional)
 ##' @param band_sd string value for which standard deviation band, or use "NA" if you do not know or do not need StdDev information (optional)
 ##' @param package_method string value to inform function of which package method to use to download modis data. Either "MODISTools" or "reticulate" (optional)
+##' @param siteID string value of a PEcAn site ID. Currently only used for output filename.
```
serbinsh: @bailsofhay is this necessary? Seems very PEcAn/BETY-centric but not sure if we want to have something like this for a user who may not have a site ID.  I would instead think it may be better to allow the user to define the output filename

modules/data.remote/R/call_MODIS.R
serbinsh: @bailsofhay agreed, no need for extra dependencies just to add leading 0s.  Make this change please

modules/data.remote/R/call_MODIS.R
serbinsh: Yup, @bailsofhay what I also keep telling you ;)

modules/data.remote/R/call_MODIS.R
```diff
@@ -27,65 +29,116 @@
...
+  {
+    start_date = as.Date(paste0(lubridate::year(start_date), spatial.tools::add_leading_zeroes(lubridate::yday(start_date), 3)), format = "%Y%j")
+  }
   
-  # set start and end dates to correct format
-  if (package_method == "MODISTools"){
-    
+  if (grepl("/", end_date) == T)
+  {
+    end_date = as.Date(paste0(lubridate::year(end_date), spatial.tools::add_leading_zeroes(lubridate::yday(end_date), 3)), format = "%Y%j")
```
serbinsh:
```suggestion
    end_date <- as.Date(paste0(lubridate::year(end_date), spatial.tools::add_leading_zeroes(lubridate::yday(end_date), 3)), format = "%Y%j")
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -27,65 +29,116 @@
...
+  }
+  start_date = as.Date(start_date, format = "%Y%j")
```
serbinsh:
```suggestion
  start_date <- as.Date(start_date, format = "%Y%j")
```

modules/data.remote/R/call_MODIS.R
serbinsh:
```suggestion
  end_date <- as.Date(end_date, format = "%Y%j")
```

modules/data.remote/R/call_MODIS.R
serbinsh:
```suggestion
      if (as.numeric(start_date)< dates[1] || as.numeric(end_date)> dates[length(dates)])
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -27,65 +29,116 @@
...
+      if (as.numeric(start_date)>=dates[1] & as.numeric(end_date)<=dates[length(dates)])
+      {
+        print("Check #2: All dates are available!")
+      }
+      
+      # Okay scenario: Some MODIS data is available for parameter start_date and end_date range, but either the start_date or end_date falls outside the range of availble 
+      # MODIS data dates
+      if (as.numeric(start_date)<=dates[1] & as.numeric(end_date)>=dates[1])
+      {
+        start_date = dates[1]
```
serbinsh: more incorrect "=" instead of "<-" please update

modules/data.remote/R/call_MODIS.R
serbinsh: please update @bailsofhay

modules/data.remote/R/call_MODIS.R
serbinsh: please update @bailsofhay

modules/data.remote/R/call_MODIS.R
serbinsh: please update @bailsofhay.  I also think %in% is a better use here

serbinsh:
```suggestion
      good <- which(output$qc == "000" | output$qc == "001")
```

modules/data.remote/R/call_MODIS.R
serbinsh: Yes filepath which will deal with platform differences in paths

modules/data.remote/R/call_MODIS.R
serbinsh: yes, please change. Dont ever use paste0 unless WITHIN the path where you need to past a name together but done ever past in the sep since this various platform to platform, or well windows to unix at least

modules/data.remote/R/call_MODIS.R
serbinsh: also after any changes to the docs you need to remake the documentation

serbinsh:
@bailsofhay where are we on this?  Ready for a new review?

serbinsh:
@bailsofhay @tonygardella looks like its back

modules/data.remote/R/call_MODIS.R
bailsofhay: I will re-do the doc soon. the siteID thing was for something else I was playing with. it's not needed anymore so I'm removing it all together

bailsofhay: OHHH I know what it was for, this was when I was getting an error that would break the function from the site query (we used suppressWarnings() to fix this). I needed it to output the data in a way that made sense.

modules/data.remote/R/call_MODIS.R
bailsofhay: Fixed all the doubling up for | and &. Unless there is a good reason I don't know about the any() and all() functions seem like overkill for conditional statements.

bailsofhay:
@ashiklom and @serbinsh. I've fixed basically all of the notes you guys mentioned. Can you approve the PR for merging :)

modules/data.remote/DESCRIPTION
```diff
@@ -12,7 +12,8 @@ Imports:
     reticulate,
     PEcAn.logger,
     PEcAn.remote,
-    stringr (>= 1.1.0)
+    stringr (>= 1.1.0),
+    binaryLogic
```
serbinsh: Is binaryLogic needed for the QA/QC handling?

modules/data.remote/R/call_MODIS.R
```diff
@@ -13,7 +13,10 @@
 ##' @param band   string value for which measurement to extract
 ##' @param band_qc string value for which quality control band, or use "NA" if you do not know or do not need QC information (optional)
 ##' @param band_sd string value for which standard deviation band, or use "NA" if you do not know or do not need StdDev information (optional)
+##' @param siteID numeric value of BETY site id value to use for output file name. Default is NULL
```
serbinsh: This is just optional correct?  We would want this tool to be useful even if you just provide a lat long/ . Can we instead allow for a BETY site ID OR a custom label to be attached to the output?

modules/data.remote/R/call_MODIS.R
```diff
@@ -13,7 +13,10 @@
...
 ##' @param package_method string value to inform function of which package method to use to download modis data. Either "MODISTools" or "reticulate" (optional)
+##' @param QC_filter Converts QC values of band and keeps only data values that are excellent or good (as described by MODIS documentation), and removes all bad values. qc_band must be supplied for this parameter to work. Default is False.
+##' @param progress TRUE reports the download progress bar of the dataset, FALSE omits the download progress bar. Default is TRUE.
```
serbinsh: This is the parallel version of progress we discussed a month or so ago correct?

modules/data.remote/R/call_MODIS.R
```diff
@@ -26,66 +29,118 @@
 ##' }
 ##' 
 ##' @author Bailey Morrison
-##'  
-call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0, product, band, band_qc = "", band_sd = "", package_method = "MODISTools") {
-  
+##'
+call_MODIS <- function(outfolder = "", start_date, end_date, lat, lon, size = 0, product, band, band_qc = "", band_sd = "", siteID = NULL, package_method = "MODISTools", QC_filter = FALSE, progress = TRUE) {
```
serbinsh: Do you like blank or should we instead set these to NULL?  Suppose I dont really care either way

modules/data.remote/R/call_MODIS.R
```diff
@@ -97,40 +152,71 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
   if (package_method == "reticulate"){
     # load in python script
     script.path <- file.path(system.file("extract_modis_data.py", package = "PEcAn.data.remote"))
-    #script.path = file.path('/Users/bmorrison/pecan/modules/data.remote/inst/extract_modis_data.py')
     reticulate::source_python(script.path)
     
     # extract the data
     output <- extract_modis_data(product = product, band = band, lat = lat, lon = lon, start_date = start_date, end_date = end_date, size = size, band_qc = band_qc, band_sd = band_sd)
-    output[,5:10] <- lapply(output[,5:10], as.numeric)
+    output[ ,5:10] <- lapply(output[ ,5:10], as.numeric)
```
serbinsh: do we need these spaces?

serbinsh:
@bailsofhay Actually, can we make sure your example still works as well?

modules/data.remote/DESCRIPTION
bailsofhay: I think so. This was already in the DESCRIPTION file before I messed around with the function. Most of the others have this also.

modules/data.remote/R/call_MODIS.R
bailsofhay: Yea. This is really useful naming output files in a way that is easily identifiable. It will be really helpful in the future when we start adding in more sites. We won't have to re-download sites because I'll have it check for the site in the file name before running call_MODIS

bailsofhay: if there is no site name, it uses the lat/lon... which just makes the output really long

modules/data.remote/R/call_MODIS.R
bailsofhay: I tried parallel, it doesn't make the function go any faster because right now the function is set up to go by a site, vs. multiple sites. In the future I'll probably change it so that you can give it a list of sites or lat/lons and then I can parallelize it easily.

modules/data.remote/R/call_MODIS.R
bailsofhay: I like it blank because if you don't want to save the data, using "" for outfielder makes it so it won't save the data anywhere.

modules/data.remote/R/call_MODIS.R
bailsofhay: yea. I probably missed some. I'll double check for missing spaces

serbinsh:
@bailsofhay good news, when the checks finish we should be good to merge

modules/data.remote/DESCRIPTION
infotroph: Not a showstopper, but consider moving binaryLogic up the list to maintain alphabetical order for easier reading. (None of the scripts that parse imports care, this is strictly for the convenience of human reader.)

bailsofhay:
As mentioned in the meeting I update this function some. I've added the changes verbally in the description above.

mdietze:
```
```

bailsofhay:
@mdietze Just committed the merge conflict changes. I also updated DESCRIPTION, oxygenized, and made the docments.... hopefully this passes

mdietze:
@bailsofhay if you look at the Travis logs, here's what it shows

```
R check of modules/data.remote returned new problems:
checking dependencies in R code ... WARNING: '::' or ':::' imports not declared from: 'DBI' 'foreach' 'glue' 'ncdf4' 'PEcAn.DB' 'PEcAn.utils' 'purrr' 'raster' 'RCurl' 'sp'
checking R code for possible problems ... NOTE: call_MODIS: no visible global function definition for '%dopar%'
checking R code for possible problems ... NOTE: call_MODIS: no visible global function definition for '%do%'
checking R code for possible problems ... NOTE: call_MODIS: no visible global function definition for 'foreach'
checking R code for possible problems ... NOTE: call_MODIS: no visible binding for global variable 'start'
checking R code for possible problems ... NOTE: call_MODIS: no visible global function definition for 'stopCluster'
checking Rd line widths ... NOTE: Rd file 'call_MODIS.Rd': \examples lines wider than 100 characters:    test_modistools <- call_MODIS(var = "lai", site_info = site_info, product_dates = c("2001150", "2001365"), run_parallel = TRUE,  produc ... [TRUNCATED]
Execution halted
Makefile:138: recipe for target '.check/modules/data.remote' failed
make: *** [.check/modules/data.remote] Error 1
```

Some of this is occurring because @infotroph upgraded the build system so that we can't introduce new Warnings (we used to let Warnings pass, but they're warnings for a reason!)

mdietze:
@bailsofhay wanted to remind you that you still need to finish up this PR

bailsofhay:
@infotroph I'm still having an issue with getting %dopar% to work in the script. While the ```@importFrom foreach %do% %dopar%``` technically works for roxygen, it's only recognizing %do% because %dopar% is supposed to tell foreach to use the doParallel package for forloops. I'm not sure how to get around this. I've tried doing something like ```@importFrom doParallel %dopar%```, but oxygen said that's not from doParallel. Just for reference: https://cran.r-project.org/web/packages/doParallel/vignettes/gettingstartedParallel.pdf

infotroph:
@bailsofhay I'm not sure I follow -- are you seeing this at package build or at runtime, and what's the error you get when `%dopar%` isn't recognized? 

Testing this branch on my machine, I can run the example (once I make up a `site_info`), but I still see check warnings from undeclared imports that will need to be fixed before Travis passes.

modules/data.remote/R/call_MODIS.R
```diff
@@ -69,153 +79,198 @@ call_MODIS <- function(outfolder = "", start_date, end_date, lat, lon, size = 0,
...
-      QC <- rep("nan", nrow(dat))
-    } else {
-      QC <- as.numeric(qc$value)
+      convert <- paste(binaryLogic::as.binary(as.integer(output$qc[i]), n = 8), collapse = "")
+      output$qc[i] <- substr(convert, nchar(convert) - 2, nchar(convert))
     }
-    
-    if (band_sd == "")
+    good <- which(output$qc %in% c("000", "001"))
+    if (length(good) > 0 || !(is.null(good)))
```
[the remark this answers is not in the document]
infotroph: Moved to its own issue #2520

PR #2345, opened 2019-04-29
fix pika issues

robkooper:
Hopefully last set of changes needed to use pika 1.0.0 (just in time since 1.0.1 is now out)

## Description
missed the monitor process when upgrading to pika 1.0.0, this prevented models from registering.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible

## Checklist:


robkooper:
`no_ack` now is  `auto_ack` and the default is False, so I removed it from the call.

ashiklom:
Cool. In that case, good to merge.

PR #2352, opened 2019-05-09
fixes to get multivariable SDA runs working

bailsofhay:
There were a few minor fixes that needed to be altered to the SDA code to get multiple variables (i.e. LAI and AGB) working to run the SDA. 

## Description
Here is a list of the files I changed:
modules/assim.sequential/R/:
1) Multi_Site_Constructors.R --> missing variable choose, which messed up assigning the H matrix with the correct "1" value where there was variable information.
2) sda.enkf.multisite.R --> Plotting is messing up multivariable runs right now, and switching all plotting parameters to FALSE does nothing to stop the SDA from trying to plot the data. I added a parameter (plot = T/F), which either plots or does not plot the SDA runs for now.... for simplicity. This will need to be adjusted once plotting is figured out (But that will be a separate PR).

modules/data.remote/R/:
3) call_MODIS.R --> minor change to allow call_MODIS to turn on progress bar or not (useful for people who parallelize call_MODIS.... not important to the SDA, but happened to be changed with this push. 

## Motivation and Context
This change is required because the previous version of assim.sequential does not work at all for multi-variable (observed) SDA runs. This allows the SDA to run with more than one observed variable, as long as plot = F in the SDA parameters. Plotting issues will be addressed in a separate PR.

## Review Time Estimate
- [X] Immediately

## Types of changes

## Checklist:


tonygardella:
@bailsofhay Quick question: you're comparing against PecanProject:bailey_sda instead of PecanProject:develop, is that on purpose? Because of that it's pulling in all the past changes from the last couple months making it hard to see what changes are yours and not others.

bailsofhay:
@tonygardella OHHHHHHHHH! That's why everyone else's stuff is on there. I suck at GitHub. I'm just trying to get mine over.... which branch should I compare it too?

tonygardella:
@bailsofhay `develop`. No worries I just changed it. It's that extra step when opening a PR you gotta double check which base branch you're comparing to. By default it usually compares to `develop` but it depends on how you created your branch.

bailsofhay:
thanks @tonygardella!!!!

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -88,6 +88,11 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
     names(forecast[[length(forecast)]]) <- c("TotLivBiom")
   }
   
+  if ("LAI" %in% var.names) {
+    forecast[[length(forecast) + 1]] <- ens$LAI[last]  ## m2/m2 
+    names(forecast[[length(forecast)]]) <- c("LAI")
```
ashiklom: No need for `c` here -- just `"LAI"` does exactly the same thing.
```suggestion
    names(forecast[[length(forecast)]]) <- "LAI"
```

Also, super minor thing, but FWIW, where you have `ens$LAI[last]` -- if `last` is always supposed to be a scalar, I would use double brackets: `ens$LAI[[last]]`. `[[` subsetting always returns a single object, and throws a nice loud error if you give a non-scalar input. This makes it easier to catch bugs, and makes the intent of the code clearer, at least IMHO.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -391,6 +391,7 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
       
       # reconstruct total wood C
       wood_total_C <- IC$AbvGrndWood / IC$abvGrndWoodFrac
+      if (is.na(wood_total_C) | is.infinite(wood_total_C) | is.nan(wood_total_C) | wood_total_C <0) wood_total_C <- 0
```
ashiklom: Recall my earlier comment about single (`|`) vs double (`||`) subsetting. If `wood_total_C` is a scalar, just change all of the `|` to `||`. But if it's a vector, you'll want to wrap the whole thing in `any` or `all`, as appropriate.

models/sipnet/R/write_restart.SIPNET.R
```diff
@@ -104,6 +104,11 @@ write_restart.SIPNET <- function(outdir, runid, start.time, stop.time, settings,
     names(analysis.save[[length(analysis.save)]]) <- c("snow")
   }
 
+  if ("LAI" %in% variables) {
+    analysis.save[[length(analysis.save) + 1]] <- new.state$LAI  
+    if (new.state$LAI < 0) analysis.save[[length(analysis.save)]] <- 0
```
ashiklom: This also seems like something that should be caught by the SDA analysis rather than in `write.restart`. But @mdietze and @araiho would know better.

modules/assim.sequential/R/Multi_Site_Constructors.R
```diff
@@ -189,6 +189,8 @@ Construct.H.multisite <- function(site.ids, var.names, obs.t.mean){
     
     if(is.null(obs.names)) next;
     
+    choose <- sapply(var.names, agrep, x = names(obs.t.mean[[site]]),
+                       max = 1, USE.NAMES = F) %>% unlist
```
ashiklom: @para2x Is `agrep` instead of `grep` here intentional? `agrep` is "approximate" grep -- i.e. "fuzzy matching", which I am _really_ wary of relying on. At the very least, I would expect a comment here justifying the use of `agrep`.

Also, please don't abbreviate `FALSE` to `F` (@para2x this is primarily on you -- after all, this was modeled after your code). `FALSE` is a reserved keyword, but `F` is not, so it's too easy to accidentally overwrite it in the global environment and wreak havoc on code like this.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -22,11 +22,7 @@
...
                                control=list(trace=T,
                                             FF=F,
-                                            interactivePlot=T,
-                                            TimeseriesPlot=T,
-                                            BiasPlot=F,
-                                            plot.title=NULL,
-                                            facet.plots=F,
+                                            plot=F,
                                             debug=FALSE,
                                             pause=F),
```
ashiklom: Please use `TRUE`/`FALSE` throughout code like this. Again, this is primarily @para2x 's fault, but might as well fix it here while it's on our mind.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -473,7 +469,8 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
          out.configs, ensemble.samples, inputs, Viz.output,
          file = file.path(settings$outdir,"SDA", "sda.output.Rdata"))
     #writing down the image - either you asked for it or nor :)
-    if (t%%2==0 | t==nt)  post.analysis.multisite.ggplot(settings, t, obs.times, obs.mean, obs.cov, FORECAST, ANALYSIS ,plot.title=control$plot.title, facetg=control$facet.plots, readsFF=readsFF)
+    if (control$plot == T & (t%%2==0 | t==nt)) post.analysis.multisite.ggplot(settings, t, obs.times, obs.mean, obs.cov, FORECAST, ANALYSIS ,plot.title=control$plot.title, facetg=control$facet.plots, readsFF=readsFF)
```
ashiklom: Again, please write out `TRUE` and `FALSE`.

Also, again, always use `||` and `&&` in `if` statements, or wrap them in `any/all`.

Also, please use appropriate code style here: The `post.analysis...` block should be on its own line, surrounded by curly braces, and there should be spaces around all infix operators (i.e. `plot.title = control$plot.title`).

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
 ##' }
 ##' 
 ##' @author Bailey Morrison
 ##'  
-call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0, product, band, band_qc = "", band_sd = "", package_method = "MODISTools") {
+call_MODIS <- function(outfolder = "", start_date, end_date, lat, lon, size = 0, iter = i, product, band, band_qc = "", band_sd = "", siteID = "", package_method = "MODISTools", QC_filter = F, progress = T) {
   
   # makes the query search for 1 pixel and not for rasters for now. Will be changed when we provide raster output support.
   size <- 0
+  require(PEcAn.all)
```
ashiklom: Please remove this. Instead, all external function calls should be namespaced (i.e. `package::function`).

```suggestion
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
+##' test_modistools <- call_MODIS(product = "MOD15A2H", band = "Lai_500m", start_date = "2004300", end_date = "2004365", lat = 38, lon = -123, size = 0, iter = 1, band_qc = "FparLai_QC", band_sd = "LaiStdDev_500m", package_method = "MODISTools", QC_filter = T)
 ##' plot(lubridate::yday(test_modistools$calendar_date), test_modistools$data, type = 'l', xlab = "day of year", ylab = test_modistools$band[1])
 ##' test_reticulate <- call_MODIS(product = "MOD15A2H", band = "Lai_500m", start_date = "2004300", end_date = "2004365", lat = 38, lon = -123, size = 0, band_qc = "",band_sd = "", package_method = "reticulate")
 ##' }
 ##' 
 ##' @author Bailey Morrison
 ##'  
-call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0, product, band, band_qc = "", band_sd = "", package_method = "MODISTools") {
+call_MODIS <- function(outfolder = "", start_date, end_date, lat, lon, size = 0, iter = i, product, band, band_qc = "", band_sd = "", siteID = "", package_method = "MODISTools", QC_filter = F, progress = T) {
```
ashiklom: Should that be `iter = 1`?

Also, again, please write out `TRUE` and `FALSE`.

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
+  {
+    start_date = as.Date(paste0(lubridate::year(start_date), spatial.tools::add_leading_zeroes(lubridate::yday(start_date), 3)), format = "%Y%j")
+  }
+  
+  if (grepl("/", end_date) == T)
+  {
+    end_date = as.Date(paste0(lubridate::year(end_date), spatial.tools::add_leading_zeroes(lubridate::yday(end_date), 3)), format = "%Y%j")
+  }
+  start_date = as.Date(start_date, format = "%Y%j")
+  end_date = as.Date(end_date, format = "%Y%j")
```
ashiklom: * Use `<-` for assignment, not `=`.
* `grepl(...) == T` is unnecessary -- `grepl` already returns a boolean.
* `spatial.tools::add_leading_zeroes` can be easily replaced with `sprintf("%03d", ...)` (replace 3 with however many zeroes you need).

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -22,11 +22,7 @@
 sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F, 
                                control=list(trace=T,
                                             FF=F,
-                                            interactivePlot=T,
-                                            TimeseriesPlot=T,
-                                            BiasPlot=F,
-                                            plot.title=NULL,
```
para2x: `plot.title`  and `facet.plots` are arguments to `post.analysis.multisite.ggplot` and you used them, this would crash that function. Also you can use `TimeseriesPlot`  for what you did in line 472.

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
+    #3. check that dates asked for in function parameters are fall within dates available for modis product/bands.
+    dates <- MODISTools::mt_dates(product = product, lat = lat, lon = lon)$modis_date
+    dates <- as.Date(as.character(substr(dates, 2, nchar(dates))), format = "%Y%j")
+    
+    ########## Date case 1:  user only wants one date ##########
+    if (start_date == end_date)
+    {
+      if (as.numeric(start_date) >= dates[1] & as.numeric(end_date) <= dates[length(dates)])
+      {
+        print("Extracting data")
```
ashiklom: In general, we should be using `PEcAn.logger` functions for messages.

FWIW, in standalone code, stuff like this is better passed through `message` rather than `print` because the former is easier to suppress in higher-level code.

bailsofhay:
@para2x: i tried just turning off the plotting with the original parameters and it still kept failing at the plotting section of the function. I'm not planning on keeping that change, so I'll change it back for now, but something funky is going on with turning on/off the plotting parameters.

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
+      } 
+      
+      # Unacceptable scenario: start_date and end_date does not fall within the availa MODIS data product date range. There is no data to extract in this scenario.
+      if ((as.numeric(start_date)<dates[1] & as.numeric(end_date <dates[1])) | (as.numeric(start_date)>dates[length(dates)] & as.numeric(end_date)>dates[length(dates)]))
+      {
+        stop("No MODIS data available start_date and end_date parameterized.")
+      }
+      
+      start <- as.Date(start_date, "%Y%j")
+      end <- as.Date(end_date, "%Y%j")
```
ashiklom: `&&` and `||` in `if` statements.

Also, note that base R has built-in support for date-time comparison, so `as.numeric(start_date)` isn't strictly necessary:

``` r
dt <- ISOdatetime(1990, 1, 1, 12, 0, 0)
dt > "1985-01-01"
#> [1] TRUE
dt > "1990-02-24"
#> [1] FALSE
```

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: i tried just turning off the plotting with the original parameters and it still kept failing at the plotting section of the function. I'm not planning on keeping that change, so I'll change it back for now, but something funky is going on with turning on/off the plotting parameters.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -473,7 +469,8 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
          out.configs, ensemble.samples, inputs, Viz.output,
          file = file.path(settings$outdir,"SDA", "sda.output.Rdata"))
     #writing down the image - either you asked for it or nor :)
```
para2x:
```suggestion
    #writing down the image - either you asked for it or not :)
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
     }
-  
     
     print("Extracting data")
-    
-    start <- as.Date(start_date, "%Y%j")
-    end <- as.Date(end_date, "%Y%j")
+    cat(paste("Product =", product, "\n", "Band =", band, "\n", "Date Range =", start, "-", end, "\n", "Latitude =", lat, "\n", "Longitude =", lon, sep = " "))
```
ashiklom: Please use `message` or, better yet, `PEcAn.logger::logger.<something>` here.

Note that `PEcAn.logger::logger.*` functions have a `wrap = FALSE` argument that preserves input formatting, which allows them to be used for printing, e.g. tabular output.

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +166,39 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
     
-    fname <- paste(product, "_", band, "output_", start_date, "_", end_date, "_", lat, "_", lon, ".csv", sep = "")
-    fname <- paste0(outfolder, "/", fname)
-    write.csv(output, fname)
-    return(output)}
-
+    if (QC_filter == T)
+    {
+      output$qc == as.character(output$qc)
+      for (i in 1:nrow(output))
```
ashiklom: `seq_len` is safer for the case of `nrow(output) == 0`.

```suggestion
      for (i in seq_len(nrow(output)))
```

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +166,39 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
     output$lat <- round(output$lat, 4)
     output$lon <- round(output$lon, 4)
     
-    fname <- paste(product, "_", band, "output_", start_date, "_", end_date, "_", lat, "_", lon, ".csv", sep = "")
-    fname <- paste0(outfolder, "/", fname)
-    write.csv(output, fname)
-    return(output)}
-
+    if (QC_filter == T)
```
ashiklom: Drop the `== T`.

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +166,39 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
+    {
+      output$qc == as.character(output$qc)
+      for (i in 1:nrow(output))
+      {
+        convert = paste(binaryLogic::as.binary(as.integer(output$qc[i]), n = 8), collapse = "")
+        output$qc[i] = substr(convert, nchar(convert)-2, nchar(convert))
```
ashiklom: Note my earlier comment about `[` vs. `[[`. Also, please use `<-` for assignment throughout this block.

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +166,39 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
+    if (!(outfolder == "") & !(is.null(siteID)))
+    {
+      fname <- paste(product, "_", band, "_", siteID, "_output_", start_date, "_", end_date, "_", spatial.tools::add_leading_zeroes(iter, 4), ".csv", sep = "")
+      fname <- paste0(outfolder, "/", fname)
+      write.csv(output, fname, row.names = F)
+    }
+    if (!(outfolder == "") & is.null(siteID))
+    {
+      fname <- paste(product, "_", band, "_output_", lat, "_", lon, "_", start_date, "_", end_date, "_", spatial.tools::add_leading_zeroes(iter, 2), ".csv", sep = "")
+      fname <- paste0(outfolder, "/", fname)
```
ashiklom: Please use `file.path` here.

modules/data.remote/R/call_MODIS.R
```diff
@@ -111,11 +166,39 @@ call_MODIS <- function(outfolder = ".", start_date, end_date, lat, lon, size = 0
...
+    
+    if (!(outfolder == "") & !(is.null(siteID)))
+    {
+      fname <- paste(product, "_", band, "_", siteID, "_output_", start_date, "_", end_date, "_", spatial.tools::add_leading_zeroes(iter, 4), ".csv", sep = "")
+      fname <- paste0(outfolder, "/", fname)
+      write.csv(output, fname, row.names = F)
+    }
+    if (!(outfolder == "") & is.null(siteID))
+    {
+      fname <- paste(product, "_", band, "_output_", lat, "_", lon, "_", start_date, "_", end_date, "_", spatial.tools::add_leading_zeroes(iter, 2), ".csv", sep = "")
```
ashiklom: Again, replace the `add_leading_zeroes` with `sprintf("%02d")`.

modules/assim.sequential/R/Multi_Site_Constructors.R
bailsofhay: @ashiklom I'll change the F's to FALSE. That may have just been me. @para2x did the agrep, so I don't know if that was intentional or not. I'm assuming it is just incase there is very slight differences between the state.variables in the SDA xml and the way the variable was named in the R script for obs.mean and obs.cov. (e.g. LAI vs lai)

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: I will try to get as many as possible. I have the SDA running on my computer right now so I don't want to mess with it until it finishes :)

modules/assim.sequential/R/Multi_Site_Constructors.R
para2x: @ashiklom I might have initially written this code somewhere, but this line here was just added by @bailsofhay not me (@bailsofhay's PR).

@bailsofhay, One thing I tried to also bring up in our last meeting was the fact that this way of thinking about the `H`  makes an implicit assumption about the order of variables in cov matrix and secondly I was also wondering if this worked under the years/sites that we have either one not both at the same time ?

modules/data.remote/R/call_MODIS.R
bailsofhay: Just out of curiosity, is sprintf("%02d") like a regex type function that does the exact same thing as add_leading_zeroes?

bailsofhay: never used it before so it's handy if so

modules/data.remote/R/call_MODIS.R
bailsofhay: use file.path where? instead of fname?

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: fixed back to how it was

models/sipnet/R/write.configs.SIPNET.R
bailsofhay: I don't get this comment. Can you explain. if I do doubles they will fail. is that what the any/all command is for?

modules/assim.sequential/R/Multi_Site_Constructors.R
ashiklom: @para2x Bailey modeled this code after the lines you wrote below (`choose.col`, `choose.row`), which is why I was asking you. I personally think using `agrep` in production code like this is a bad idea -- we should be precise in our definitions and configurations and error quickly and loudly when something isn't right. Trying to predict what users meant to type is very hard and error-prone, and can introduce really insidious bugs (e.g. `WAI` could get corrected to `LAI`, even though they are two very different variables).

models/sipnet/R/write_restart.SIPNET.R
bailsofhay: I don't think so... at least if we plan to use different models in the SDA in the future. In this case the names are specific to SIPNET, so if in the future we start to use ED or something the names may be different and I don't know if the SDA should be responsible for those types of differences.

bailsofhay: but I'm newer at this so I could be wrong/misreading the problem.

modules/assim.sequential/R/Multi_Site_Constructors.R
bailsofhay: @ashiklom and @para2x I added that line, but it was stolen from the Construct.R function exactly.... so..... sort of me, but not.

modules/data.remote/R/call_MODIS.R
ashiklom: Not quite. AFAIK, it's based on C's string formatting functions. My syntax above was wrong -- you really want `sprintf("%02d", iter)`.

Here are some more examples:

``` r
sprintf("An AU is %d miles", 93e6)
#> [1] "An AU is 93000000 miles"
sprintf("The current date is %d-%02d-%02d", 2019, 5, 9)
#> [1] "The current date is 2019-05-09"
sprintf("The price of %s is $%.2f per pound.",
        c("apples", "oranges", "bananas"),
        c(1.78912312, 1.6912341, 0.12341))
#> [1] "The price of apples is $1.79 per pound." 
#> [2] "The price of oranges is $1.69 per pound."
#> [3] "The price of bananas is $0.12 per pound."
```

There's a lot more you can do with formatting codes -- see [`?sprintf`](https://www.rdocumentation.org/packages/base/versions/3.6.0/topics/sprintf) for more details.

modules/assim.sequential/R/Multi_Site_Constructors.R
bailsofhay: @para2x Yea. I get what you're saying. I'm a little confused by what you're saying in the second part. I know when I was testing that function that it appears okay for now because the variables are named for each year at each site in obs.mean, and agrep matches them up to the state.variable names in the XML... so H is correct and in the correct order. i specifically checked for that just in case. H assumes that the way the variables are listed (and their order) is the same as obs.mean and obs.cov.... so as long as obs.cov is built the same way as obs.mean in the script,  H should be okay

modules/data.remote/R/call_MODIS.R
ashiklom:
```suggestion
      fname <- file.path(outfolder, fname)
```

`file.path("a", "b", "c")` will automatically concatenate these with the right separators for the given OS -- i.e. it will create `"a/b/c"` on Unix systems, and `"a\b\c"` on Windows.

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: when you see certain R files pop up with the problem can you make a note for me. There are a lot of files for the SDA and I may not catch all them since I didn't make these ones. The call_MODIS one I can fix pretty easily since I made that

modules/data.remote/R/call_MODIS.R
ashiklom: The most relevant one for you from that group is the second one ("current date"). `%02d` is parsed out as:

- `%` -- this indicates the beginning of  a replacement format code
- `02` -- The `0` means "pad with zeros", and `2` indicates how many to pad with. To pad with 10 zeros, you would use `010`. Without the leading `0` (i.e. just `%2d`), pad with spaces instead, so `sprintf("%5d", 5)` would make `     5`.
- `d` -- This means the argument is an integer. Other common ones are `s` for "string" and `f` for float. Different inputs have different formatting types. The padding stuff works for `s`, `d`, and `f` (though a little differently for `f`), but for `f` you can also specify the precision after the decimal point, for example (as I do in the last example above).

models/sipnet/R/write.configs.SIPNET.R
bailsofhay: nm. Shawn told me it's on the call_modis PR

modules/data.remote/R/call_MODIS.R
bailsofhay: output is a data.frame so I don't think i need doubles..... right???? just checking since this is new.

modules/data.remote/R/call_MODIS.R
bailsofhay: fixed on my own.... so I'm not committing the suggestion so it doesn't change the other things I fixed recently :). Thanks for the explanation :)

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: fixed

modules/data.remote/R/call_MODIS.R
bailsofhay: fixed

models/sipnet/R/write.configs.SIPNET.R
serbinsh: @bailsofhay https://github.com/PecanProject/pecan/pull/2338#discussion_r277086316

models/sipnet/R/write_restart.SIPNET.R
serbinsh: @ashiklom can you clarify what you mean by "caught by the SDA?"

modules/data.remote/R/call_MODIS.R
bailsofhay: thanks! fixed also

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -19,16 +19,16 @@
...
-                                            facet.plots=F,
+sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart = FALSE, 
+                               control=list(trace = TRUE,
+                                            FF = FALSE,
+					    interactivePlot = FALSE,
+                                            TimeseriesPlot = FALSE,
+					    BiasPlot = FALSE,
+					    plot.title = NULL,
+					    facet.plots = FALSE,
                                             debug=FALSE,
```
serbinsh:
```suggestion
                                            debug = FALSE,
```

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -19,16 +19,16 @@
...
-                                            pause=F),
+                                            pause=FALSE),
```
serbinsh:
```suggestion
                                            pause = FALSE),
```

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -69,7 +69,7 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
   }
   
   #Finding the distance between the sites
-  distances <- sp::spDists(site.locs, longlat=T)
+  distances <- sp::spDists(site.locs, longlat=TRUE)
```
serbinsh:
```suggestion
  distances <- sp::spDists(site.locs, longlat = TRUE)
```

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -114,7 +114,7 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
               start.time = lubridate::ymd_hms(settings$state.data.assimilation$start.date, truncated = 3),
               stop.time = lubridate::ymd_hms(settings$state.data.assimilation$end.date, truncated = 3),
               inputs =  settings$run$inputs$met$path[[i]],
-              overwrite =F
+              overwrite =FALSE
```
serbinsh:
```suggestion
              overwrite = FALSE
```

modules/data.remote/R/call_MODIS.R
serbinsh: Indeed.....this is brittle

modules/data.remote/R/call_MODIS.R
bailsofhay: iter is for people using foreach from doParallel. I need to add in this description and change up this part of the function. Thanks for reminding me of that.

modules/data.remote/R/call_MODIS.R
serbinsh:
```suggestion
        PEcAn.logger::logger.info("Extracting data")
```

modules/data.remote/R/call_MODIS.R
serbinsh: https://github.com/PecanProject/pecan/pull/2338#discussion_r277086316

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
...
-                                  start=start, end=end, km_ab=size, km_lr=size)
+                                 start=start_date, end=end_date, km_ab=size, km_lr=size, progress = progress)
+    
     # extract QC data
-    if(band_qc != ""){
+    if(band_qc != "")
+    {
       qc <- MODISTools::mt_subset(lat=lat, lon=lon, product=product, band=band_qc,
-                                  start=start, end=end, km_ab=size, km_lr=size)
+                                  start=start, end=end, km_ab=size, km_lr=size, progress =progress)
```
serbinsh:
```suggestion
                                  start = start, end = end, km_ab = size, km_lr = size, progress = progress)
```

modules/data.remote/R/call_MODIS.R
bailsofhay: iter is fixed now in the script to make more sense and catch is the user is even using iter (default in now NULL). T and F's were fixed earlier

modules/data.remote/R/call_MODIS.R
bailsofhay: removed add_leading_zeroes and fixed all equal signs. Need to look at the grepl thing

models/sipnet/R/write_restart.SIPNET.R
ashiklom: Yeah, exactly. There's another similar change above that forces woody C to be greater than zero, which I think messes up @araiho's zero-inflated stats. 

If nothing else, punting these kinds of corrections to model specific restart functions means we have to duplicate them for every model, which isn't ideal. But more generally, I think we should inspect why the SDA analysis returns impossible values and fix that. In the meantime, to approve this change, it needs to come with a very loud and clear warning that the SDA returned an illegal value (which we can then quickly convert to an error once the underlying SDA problem is fixed).

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: Thanks Alexey. I'll try fixing as many of those as I edit previous code. Most of these are functions I didn't write so I'll try and make sure any new ones I happen to make follow more appropriate code styles :)

bailsofhay:
The write_restart_SIPNET and read_restart_SIPNET files were "changed", but not actually. Somehow I accidentally changed them when I was viewing the files and it broke my signet model so it wouldn't build. I just copy and pasted over the entire previous file back into my branch so it would be correct again. just an FYI

modules/assim.sequential/R/sda_plotting.R
```diff
@@ -539,8 +539,11 @@ post.analysis.multisite.ggplot <- function(settings, t, obs.times, obs.mean, obs
               Lower=quantile(Value,0.025, na.rm=T),
               Upper=quantile(Value,0.975, na.rm=T))
           
-        })%>%mutate(Type=paste0("SDA_",listFA),
+        # dropped the "_" from "SDA_" in plot legends
```
para2x: I guess @mdietze meant the whole "SDA_" and not just "_" and as a heads up, when you take the "SDA" out this would change the order of colors on the plots.

bailsofhay: oh lol. that's an easy fix

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -367,30 +367,44 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
     ## plantWoodInit gC/m2
     plant_wood_vars <- c("AbvGrndWood", "abvGrndWoodFrac", "coarseRootFrac", "fineRootFrac")
     if (all(plant_wood_vars %in% ic.names)) {
-  
-      
       # reconstruct total wood C
       wood_total_C <- IC$AbvGrndWood / IC$abvGrndWoodFrac
+      #Sanity check
+      if (is.na(wood_total_C) | is.infinite(wood_total_C) | is.nan(wood_total_C) | wood_total_C <0) {
+        wood_total_C <- 0
```
serbinsh: @para2x @bailsofhay what is this check for? Is this supposed to be 0 or should it be something like 0.001?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -206,9 +206,10 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
     ###  Taking care of Forecast. Splitting /  Writting / running / reading back###
     ###-------------------------------------------------------------------------###-----  
     #- Check to see if this is the first run or not and what inputs needs to be sent to write.ensemble configs
+    
     if (t>1){
       #removing old simulations
-      unlink(list.files(outdir, "*.nc", recursive = T, full.names = T))
+      unlink(list.files(outdir, "*.nc.var", recursive = TRUE, full.names = TRUE))
```
serbinsh: why are we just removing old nc.var and not .nc as well?  Previously it looks like we were removing nc data?  Why remove at all?

bailsofhay: This was for checking the .nc files that get output at the end of each year. The SDA originally deleted all .nc or .nc.var files, so I couldn't check the individual years.

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -45,17 +45,17 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   if ("AbvGrndWood" %in% var.names) {
     forecast[[length(forecast) + 1]] <- udunits2::ud.convert(ens$AbvGrndWood[last],  "kg/m^2", "Mg/ha")
     names(forecast[[length(forecast)]]) <- c("AbvGrndWood")
-   
+    
     # calculate fractions, store in params, will use in write_restart
     wood_total_C    <- ens$AbvGrndWood[last] + ens$fine_root_carbon_content[last] + ens$coarse_root_carbon_content[last]
-    
+    if (wood_total_C==0) wood_total_C <- 0.0001 # Making sure we are not making Nans in case there is no plant living there.
```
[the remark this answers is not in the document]
bailsofhay: @ashiklom This was a @para2x thing. I just added it because it wasn't merged into any of the SDA code. It's supposed to be a temporary fix because values <= 0 were spitting out Inf or Nan values and crashing the SDA. There is a conversation on Slack between @mdietze, @para2x, and (I think @istfer) on this issue and what to do about it in the long run

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: @serbinsh This isn't going in to the final SDA, it was just a section I added to dig into issues

models/sipnet/R/write.configs.SIPNET.R
bailsofhay: Dunno, ask @para2x

para2x: @serbinsh I have the complete version of this check in my PR where it gives warnings and etc that what happed. It's for when the woodFrac comes out as zero and blows up the wood_total.

models/sipnet/R/#write_restart.SIPNET.R#
```diff
@@ -0,0 +1,126 @@
+#-------------------------------------------------------------------------------
```
serbinsh: @bailsofhay is this a backup copy?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -19,16 +20,16 @@
...
+sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart = FALSE, forceRun = TRUE, 
+                               control=list(trace = TRUE,
+                                            FF = FALSE,
+                                            interactivePlot = FALSE,
+                                            TimeseriesPlot = FALSE,
+                                            BiasPlot = FALSE,
+                                            plot.title = NULL,
+                                            facet.plots = FALSE,
+                                            debug = FALSE,
+                                            pause=FALSE),
```
serbinsh:
```suggestion
                                            pause = FALSE),
```

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -48,7 +49,7 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
   scalef <- settings$state.data.assimilation$scalef %>% as.numeric() # scale factor for localization
   var.names <- sapply(settings$state.data.assimilation$state.variable, '[[', "variable.name")
   names(var.names) <- NULL
-  multi.site.flag <- PEcAn.settings::is.MultiSettings(settings)
+  multi.site.flag <- T#PEcAn.settings::is.MultiSettings(settings)
```
serbinsh: @bailsofhay is this a temporary change?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -78,14 +79,14 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
...
-  if(!dir.exists("SDA")) dir.create("SDA",showWarnings = F)
+  if(!dir.exists("SDA")) dir.create("SDA",showWarnings = FALSE)
   #--get model specific functions
   do.call("library", list(paste0("PEcAn.", model)))
   my.write_restart <- paste0("write_restart.", model)
   my.read_restart <- paste0("read_restart.", model)
   my.split_inputs  <- paste0("split_inputs.", model)
   #- Double checking some of the inputs
-  if (is.null(adjustment)) adjustment<-T
+  if (is.null(adjustment)) adjustment<-TRUE
```
serbinsh:
```suggestion
  if (is.null(adjustment)) adjustment <- TRUE
```

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: @para2x told me to do that for some reason. I think I just forgot it was in there. It can go

para2x: @serbinsh @bailsofhay I remember you guys told me that for whatever reason this breaks our code and I said we don't use this flag yet.

bailsofhay: I just removed it for now

models/sipnet/R/write_restart.SIPNET.R
bailsofhay: Just as an FYI, this 0 thing isn't just and Lai issue. this is coded in for pretty much every variable in the script.

bailsofhay: Also, I just added in a warning in the description of the function about some variables having hard coded values if they produce illegal values.

bailsofhay:
@ashiklom, @para2x, and @serbinsh: I've got all (I think) of the formatting and little changes done on this PR. There are a couple of "bigger" issues that need to get resolved in terms of  stuff like write_restart assigning 0's to variables, or the quick fix I added to the SDA to force run job.sh files that somehow were missed. Can you guys check this and approve the merge :)

para2x:
@bailsofhay if it passes Travis, I can pull it in tomorrow and test it and give feedbacks.

models/sipnet/R/#write_restart.SIPNET.R#
para2x: You can delete this file if you're not using it.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -206,9 +207,11 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
     ###  Taking care of Forecast. Splitting /  Writting / running / reading back###
     ###-------------------------------------------------------------------------###-----  
     #- Check to see if this is the first run or not and what inputs needs to be sent to write.ensemble configs
-    if (t>1){
+    
+    if (t > 1){
       #removing old simulations
-      unlink(list.files(outdir, "*.nc", recursive = T, full.names = T))
```
para2x: Output `nc` files are named based on the year of the simulation. So in cases like where we assimilate in a time step smaller than annually, SDA will skip the run part because it thinks the output is already there. So I would suggest that even if you really want to get ride of this line, give it a control flag so the user would have a control over it.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -266,16 +270,37 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
         )
       })
     
-    if(t==1)  inputs <- out.configs %>% map(~.x[['samples']][['met']]) # for any time after t==1 the met is the splitted met
+    if (t == 1)  inputs <- out.configs %>% map(~.x[['samples']][['met']]) # for any time after t==1 the met is the splitted met
     #-------------------------------------------- RUN
    
     PEcAn.remote::start.model.runs(settings, settings$database$bety$write)
+    
+    if (forceRun == TRUE)
```
para2x: I want to push back on this part again. I really don't think it's a good idea to bring a "quick fix" into the mainline.  I don't know what might be the scenario for pecan to leave some jobs not submitted , but if there is it needs to be fix inside `PEcAn.remote::start.model.runs` function.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -472,8 +501,12 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
     save(site.locs, t, FORECAST, ANALYSIS, enkf.params, new.state, new.params,
          out.configs, ensemble.samples, inputs, Viz.output,
          file = file.path(settings$outdir,"SDA", "sda.output.Rdata"))
-    #writing down the image - either you asked for it or nor :)
-    if (t%%2==0 | t==nt)  post.analysis.multisite.ggplot(settings, t, obs.times, obs.mean, obs.cov, FORECAST, ANALYSIS ,plot.title=control$plot.title, facetg=control$facet.plots, readsFF=readsFF)
+    #writing down the image - either you asked for it or not :)
+
+#   if (t%%2==0 | t==nt)  post.analysis.multisite.ggplot(settings, t, obs.times, obs.mean, obs.cov, FORECAST, ANALYSIS ,plot.title=control$plot.title, facetg=control$facet.plots, readsFF=readsFF)
+    if (t == 1){
+      unlink(list.files(outdir, "*.nc", recursive = TRUE, full.names = TRUE))
```
para2x: You don't need the `if`. You can move this to line 255.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -472,8 +501,12 @@ sda.enkf.multisite <- function(settings, obs.mean, obs.cov, Q = NULL, restart=F,
     save(site.locs, t, FORECAST, ANALYSIS, enkf.params, new.state, new.params,
          out.configs, ensemble.samples, inputs, Viz.output,
          file = file.path(settings$outdir,"SDA", "sda.output.Rdata"))
-    #writing down the image - either you asked for it or nor :)
```
para2x: Now there is no plotting function called in the main SDA code. Use `control$TimeseriesPlot` inside the `if` so the user can control the plotting.

modules/data.remote/R/call_MODIS.R
```diff
@@ -14,78 +14,133 @@
 ##' @param band_qc string value for which quality control band, or use "NA" if you do not know or do not need QC information (optional)
 ##' @param band_sd string value for which standard deviation band, or use "NA" if you do not know or do not need StdDev information (optional)
```
para2x: Throughout this file there are lot's of print statements that they need to be replaced with proper pecan.logger functions.

serbinsh: Agreed

modules/assim.sequential/R/sda.enkf_MultiSite.R
serbinsh: @para2x you should chat with @bailsofhay this week about why we have this "patch" to see how we should proceed

modules/assim.sequential/R/sda.enkf_MultiSite.R
serbinsh: @para2x @bailsofhay this is a good suggestion.  Didn't think about cases < 1 year in temporal frequency

models/sipnet/R/#write_restart.SIPNET.R#
serbinsh: Probably a backup, but yeah no need to keep around

serbinsh:
> @bailsofhay if it passes Travis, I can pull it in tomorrow and test it and give feedbacks.

@para2x maybe you and @bailsofhay can try running at BU together this week to assess its status?

models/sipnet/R/#write_restart.SIPNET.R#
bailsofhay: done

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: kk. I can make this fix

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: cool! thanks

bailsofhay: fixed

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: That originally had to do with removing the .nc files that were created for all years at the beginning of the run, but keep the other .nc files that are outputted. I added this to the "keepNC" option that I added so it's gone now

bailsofhay:
I updated the call_modis code that is in this pr to be exactly the same as the modisTools pr (https://github.com/PecanProject/pecan/pull/2338) since this was accidentally added in.

modules/assim.sequential/R/sda.enkf_MultiSite.R
bailsofhay: @mdietze  and I just had a conversation about this. He said to keep the bandaid for now, but to go ahead and keep it for this merge, just so we can get my code for multisite added to the SDA code. This will be for the next PR

bailsofhay:
@ashiklom This is showing me that there is one more change request from you, but I can't find it anywhere in this PR. Am I missing something?

para2x:
I pulled the branch to test and make sure the new `H`does not break the existing SDA runs and it does NOT. So I guess if @mdietze is fine with the `force run` option then this should be ready after resolving the conflicts .

serbinsh:
@bailsofhay @para2x @mdietze @istfer looks like @bailsofhay also has several merge conflicts now given the updates made to develop.  Perhaps you can add this to the queue this week to resolve together to make sure nothing gets lost?

mdietze:
I'm OK with forceRun, but agree that we really should figure out the underlying cause as to why some ensemble members are not being run.

ashiklom:
@bailsofhay I approved the changes to make sure I wasn't holding up this PR, but I haven't actually had a chance to look through the changes. Seems like you all have this one covered.

bailsofhay:
Yea, I want to figure out what is going on with some jobs not getting run! I just want to get the stuff that is working in right now, with this bandaid, so that I can get the rest of my pecan code up to date, and not accidentally remove things that are definitely working/fixed right now

bailsofhay:
@para2x: so my build failed and I'm testing the SDA code to see if I can get a run done, which is crashing right away because of the the plan function that was added to the code. Is there some dependency that wasn't added to the function/package that is crashing this. Here is the error:
```
plan(multiprocess)
Error in plan(multiprocess) : could not find function "plan"
```

para2x:
@bailsofhay load the 'furrr' package into your workflow

bailsofhay:
kk. outside of the function?

para2x:
Yes

bailsofhay:
@para2x: here is another issue that is causing the SDA to break. This part at the end of the SDA (~L636)
``` },error = function(e {
      # If it breaks at some steps then I lose all the info on the other variables that worked fine up to the step before the break
      save(site.locs,
           t,
           FORECAST,
           ANALYSIS,
           enkf.params,
           new.state, new.params, params.list,
           out.configs, ensemble.samples, inputs, Viz.output,
           file = file.path(settings$outdir,"SDA", "sda.output.Rdata"))
      
      PEcAn.logger::logger.severe(paste0("Something just broke along the way. See if the message is helpful ", e))
    })
```

bailsofhay:
Something is up with the brackets

tonygardella:
@bailsofhay Have you tried running `make document` and pushing up the changes it makes?

bailsofhay:
@mdietze can we get this merged so @para2x can get the ERA5 PR merged???? Pretty please :)

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -38,18 +38,12 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   params$restart <-c() ## This will be filled with some restart coefficient if above ground wood is in the state variables.
 
   #### PEcAn Standard Outputs
-  if ("GWBI" %in% var.names) {
-    forecast[[length(forecast) + 1]] <- udunits2::ud.convert(mean(ens$GWBI),  "kg/m^2/s", "Mg/ha/yr")
-    names(forecast[[length(forecast)]]) <- c("GWBI")
-  }
-
```
mdietze: Unclear why GWBI was removed. This quantity is needed to assimilate tree ring data

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -78,14 +73,19 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
...
     names(forecast[[length(forecast)]]) <- c("Litter")
   }
-  
-  if ("TotSoilCarb" %in% var.names) {
-    forecast[[length(forecast) + 1]] <- ens$TotSoilCarb[last]  ## kgC/m2
-    names(forecast[[length(forecast)]]) <- c("TotSoilCarb")
+
+  if ("NEE" %in% var.names) {
+    forecast[[length(forecast) + 1]] <- ens$NEE[last]  ## gC/m2
+    names(forecast[[length(forecast)]]) <- c("NEE")
```
mdietze: NEE already exists earlier in the file. Not sure why there's a duplicate

models/sipnet/R/write_restart.SIPNET.R
```diff
@@ -55,11 +58,17 @@ write_restart.SIPNET <- function(outdir, runid, start.time, stop.time, settings,
     analysis.save[[length(analysis.save) + 1]] <- udunits2::ud.convert(new.state$NPP, "kg/m^2/s", "Mg/ha/yr")  #*unit.conv -> Mg/ha/yr
     names(analysis.save[[length(analysis.save)]]) <- c("NPP")
   }
+
+  if ("NEE" %in% variables) {
+    analysis.save[[length(analysis.save) + 1]] <- new.state$NEE
+    if (new.state$NEE < 0) analysis.save[[length(analysis.save)]] <- 0
+    names(analysis.save[[length(analysis.save)]]) <- c("NEE")
+  }
```
mdietze: Unclear why this code exits. NEE < 0 is perfectly valid, and indeed is the norm for most ecosystems (net carbon sinks) so I don't see why those values are being set to 0.

models/sipnet/R/read_restart.SIPNET.R
bailsofhay: Removed duplicate

models/sipnet/R/read_restart.SIPNET.R
bailsofhay: added GWBI back in

models/sipnet/R/write_restart.SIPNET.R
bailsofhay: I think I was adding NEE in because it was not already in there. It was copied from another variable and was not edit beyond changing the name to NEE. I've remove line 64

models/sipnet/R/read_restart.SIPNET.R
bailsofhay: added GWBI back in

models/sipnet/R/read_restart.SIPNET.R
bailsofhay: Removed duplicate

bailsofhay:
@mdietze : this says there is one more change request, but every time I try to address and and a change, it just keeps saying there is 1 change request from you. Do you know what it thinks hasn't been addressed so I can make sure it's fixed to get this pr merged?????

models/sipnet/R/write_restart.SIPNET.R
serbinsh: Strange, I looked at the "Files Changed" tab and it shows this line as removed? But here on the "conversations" tab it still shows line 64 as changing NEE to be >=0?  GitHub bug?

models/sipnet/R/write_restart.SIPNET.R
```diff
@@ -61,7 +61,6 @@ write_restart.SIPNET <- function(outdir, runid, start.time, stop.time, settings,
 
   if ("NEE" %in% variables) {
     analysis.save[[length(analysis.save) + 1]] <- new.state$NEE
-    if (new.state$NEE < 0) analysis.save[[length(analysis.save)]] <- 0
```
serbinsh: Here it is fixed?  So not sure why the check is still failing

bailsofhay:
@mdietze here is your reminder to check out the code again and merge it, if it is ready :)

serbinsh:
@mdietze do you mind addressing the one outstanding "approval" so we can merge?  Want to be sure we can have this in before we start making more changes, etc.

models/sipnet/R/write_restart.SIPNET.R
mdietze: Still need to fix this line (line 64 should be deleted, comment suggested that it was, but it's still here)

bailsofhay: this was the thing Shawn was mentioning. If you look at the actual file in my branch, it is changed. but for some reason on here it is not. I Double checked my write_restart file, and it is changed, and I re-pushed the branch, just in case I didn't push it correctly or something.

bailsofhay: In fact, if you view the file, you'll see line 64 isn't in there anymore.

PR #2354, opened 2019-05-16
build sf locally to avoid gdal version conflict

infotroph:
## Description
Quick fix for Travis failures -- the version of `sf` installed from cran2deb4ubuntu expects a different version of GDAL than is installed, but appears to do OK if built locally.

This will make uncached builds longer, but should get us going again until the prebuilt version updates.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ addons:
       - r-cran-rncl
       - r-cran-roxygen2
       - r-cran-rsqlite
-      - r-cran-sf
+      # - r-cran-sf
       - r-cran-shiny
       - r-cran-sirt
       - r-cran-testthat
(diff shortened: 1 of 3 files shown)


tonygardella:
Looks like the Namespaces are impacted from this change?

robkooper:
Getting closer to my dockerfile.travis to test local travis like builds.

infotroph:
I think the namespace changes are just Roxygen exiting after earlier failures and leaving NAMESPACE unfinished. The proximal issue seems to be that data.atmosphere installation fails because "rgal is not available". Trying to understand why it's not just pulling it from CRAN.

PR #2356, opened 2019-05-17
NOAA GEFS downscale

kzarada:
## Description
Function to download NOAA GEFS and then downscale from 6 hour to hourly. Added downscaling functions (spline, solar geom, and repeat hourly) within data.atmosphere and other files were updated to include NOAA_GEFS_downscale in pecan workflow. The met2model should be checked since the DOY was an issue we had recently. I updated it to use the time from from the nc files because it was giving the wrong DOY for the downscale. The current version returns the correct DOY for both NOAA GEFS and the downscaled NOAA GEFS.  

## Motivation and Context
Created so we can have an hourly forecast for Willow Creek instead of 6 hour. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ X] When possible
## Types of changes

## Checklist:


models/sipnet/R/met2model.SIPNET.R
```diff
@@ -277,10 +277,12 @@ met2model.SIPNET <- function(in.path, in.prefix, outfolder, start_date, end_date
       }
     }
     
-    if(year.fragment){ #gets correct DOY for fragmented years 
-      doy.start <-  length(as.Date(paste0(start_year, "-01-01")):as.Date(start_date)) * (86400 / dt) + 1 #subtract to include start.date, add to exclude last half hour of day before
-      doy.end <-  length(as.Date(paste0(start_year, "-01-01")):as.Date(end_date)) * (86400 / dt)
-      doy <- doy[doy.start:doy.end]
+    if(year.fragment){ #gets correct DOY for fragmented years using start date, time since start date and end date
+      doy.seq <- as.Date(seq(from = start_date + sec[1], to = end_date, length.out = length(sec)))
```
ashiklom: FYI, `seq` also has `Date` and `POSIXt` methods, so you can write code like this:

```r
seq(as.POSIXct("2010-01-01"), as.POSIXct("2010-02-01"), by = "2 days")
```

See `?seq.POSIXt` for details, but basically the `by` argument can be various strings like `2 months`, `5 days`, `3 hours`, `10 minutes`, or `5 seconds`.

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+  if (start_date > end_date) {
+    PEcAn.logger::logger.severe("Invalid dates: end date occurs before start date")
+  } else if (as.numeric(end_date - start_date, units="hours") < 6) { #Done separately to produce a more helpful error message.
+    PEcAn.logger::logger.severe("Times not far enough appart for a forecast to fall between them.  Forecasts occur every six hours; make sure start 
+                                and end dates are at least 6 hours appart.")
+  }
+  
+  #Set the end forecast date (default is the full 16 days)
+  if (end_date > start_date + lubridate::days(16)) {
+    end_date = start_date + lubridate::days(16)
```
ashiklom: This should probably throw an informative warning that includes the updated `end_date`.

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+  }
+  
+  #Round the starting date/time down to the previous block of 6 hours.  Adjust the time frame to match.
+  forecast_hour = (lubridate::hour(start_date) %/% 6) * 6 #Integer division by 6 followed by re-multiplication acts like a "floor function" for multiples of 6
```
ashiklom: Just FYI, see also the `lubridate::floor_date` function. If I'm reading this correctly, this should be analogous to:

```r
lubridate::floor_date(start_date, "6 hours")
```

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+  increments = as.integer(as.numeric(end_date - start_date, units = "hours") / 6) #Calculating the number of forecasts between start and end dates.
+  increments = increments + ((lubridate::hour(end_date) - lubridate::hour(start_date)) %/% 6) #These calculations are required to use the rnoaa package.
+  
+  end_hour = sprintf("%04d", ((forecast_hour + (increments * 6)) %% 24) * 100)  #Calculating the starting hour as a string, which is required type to access the 
+  #data via the rnoaa package
+  forecast_hour = sprintf("%04d", forecast_hour * 100)  #Having the end date as a string is useful later, too.
+  
+  #Recreate the adjusted start and end dates.
+  start_date = as.POSIXct(paste0(lubridate::year(start_date), "-", lubridate::month(start_date), "-", lubridate::day(start_date), " ", 
+                                 substring(forecast_hour, 1,2), ":00:00"), tz="UTC")
```
ashiklom: FYI you can also use the base R function `ISOdatetime` to construct date-times from individual components:

```r
ISOdatetime(lubridate::year(start_date), lubridate::month(start_date), lubridate::day(start_date), lubridate::hour(start_date), 0, 0, tz = "UTC")
```

See `?ISOdatetime` for details.

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+  # these fail to match up.
+  # for (i in 1:length(cf_var_names)) {
+  #  print(sprintf("cf / noaa   :   %s / %s", cf_var_names[[i]], noaa_var_names[[i]]))
+  #}
+  
+  noaa_data = list()
+  
+  #Downloading the data here.  It is stored in a matrix, where columns represent time in intervals of 6 hours, and rows represent
+  #each ensemble member.  Each variable getxs its own matrix, which is stored in the list noaa_data.
+  for (i in 1:length(noaa_var_names)) {
```
ashiklom: `noaa_var_names` is hard coded here, so it shouldn't really matter, but strictly speaking, `seq_along` is safer.

```suggestion
  for (i in seq_along(noaa_var_names)) {
```

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+    noaa_data[[i]] = rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = 1:increments, forecast_time = forecast_hour, date=format(start_date, "%Y%m%d"))$data
```
ashiklom:
```suggestion
    noaa_data[[i]] = rnoaa::gefs(noaa_var_names[i], lat.in, lon.in, raw=TRUE, time_idx = seq_len(increments), forecast_time = forecast_hour, date=format(start_date, "%Y%m%d"))$data
```

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+  
+  joined <- dplyr::inner_join(joined, ShortWave.ds, by = c("NOAA.member", "timestamp")) %>% 
+    dplyr::distinct() %>% 
+    dplyr::mutate(surface_downwelling_shortwave_flux_in_air = ifelse(surface_downwelling_shortwave_flux_in_air < 0, 0, surface_downwelling_shortwave_flux_in_air),
+                  specific_humidity = ifelse(specific_humidity <0, 0, specific_humidity),
+                  air_temperature = ifelse(air_temperature > 320, NA, air_temperature),
+                  air_temperature = ifelse(air_temperature < 240, NA, air_temperature),
+                  precipitation_flux = ifelse(precipitation_flux < 0, 0, precipitation_flux),
+                  surface_downwelling_longwave_flux_in_air = ifelse(surface_downwelling_longwave_flux_in_air < 0, NA, surface_downwelling_longwave_flux_in_air),
+                  wind_speed = ifelse(wind_speed <0, 0, wind_speed)) %>%
```
ashiklom: Especially since we're using `dplyr` anyway, I would use `dplyr::if_else` throughout here because it is stricter about enforcing type-safety. You will have to chance `NA` to the more explicit `NA_real_` when you do that (because the default type for `NA` is`logical`).

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,341 @@
...
+  )
+  
+  results_list = list()
+  
+  #Each ensemble gets its own file.
+  #These dimensions will be used for all 21 ncdf4 file members, so they're all declared once here.
+  #The data is really one-dimensional for each file (though we include lattitude and longitude dimensions
+  #to comply with the PEcAn standard).
+  time_dim = ncdf4::ncdim_def(name="time", 
+                              paste(units="hours since", format(start_date, "%Y-%m-%dT%H:%M")),
```
ashiklom: I think this is a typo?

```suggestion
                              units = paste("hours since", format(start_date, "%Y-%m-%dT%H:%M")), 
```

modules/data.atmosphere/R/downscale_solar_geom.R
```diff
@@ -0,0 +1,34 @@
+##' @name downscale_solar_geom
+##' @title solar_geom
+##' @export
+##' @author Mike Dietze
+##' @description calculates potential top-of-atmosphere shortwave radiation as a function of day of year and location
+##' @param doy time as day of year. Integers indicated midnight.
+##' @param lon longitude
+##' @param lat latitude
+downscale_solar_geom <- function(doy, lon, lat) {
```
ashiklom: This looks like a >90% duplicate of the functions in https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/solar_angle.R. I would move this function into that same file, refactor any duplicate pieces (for example, I bet some of this can use the `equation_of_time` function), and, if there are key differences, emphasize them in the function documentation.

modules/data.atmosphere/R/downscale_spline_to_hourly.R
```diff
@@ -0,0 +1,56 @@
...
+  interp.df.days <- by.ens %>% dplyr::do(days = seq(min(df$days_since_t0), as.numeric(max(df$days_since_t0)), 1/24))
+  interp.df <- interp.df.days
+  
+  for(Var in 1:length(VarNamesStates)){
+    assign(paste0("interp.df.",VarNamesStates[Var]), dplyr::do(by.ens, var = interpolate(.$days_since_t0,unlist(.[,VarNamesStates[Var]]))) %>% plyr::rename(c("var" = VarNamesStates[Var])))
+    if("dscale.member" %in% colnames(df)){
+        interp.df <- dplyr::inner_join(interp.df, get(paste0("interp.df.",VarNamesStates[Var])), by = c("NOAA.member", "dscale.member"))
+    }else{
+      interp.df <- dplyr::inner_join(interp.df, get(paste0("interp.df.",VarNamesStates[Var])), by = c("NOAA.member"))
+    }
```
ashiklom: Using `assign` and `get` like this makes me a bit nervous. In most cases (including here), I think a named list can accomplish the same thing; i.e.

```r
var_list <- list()
for (Var in seq_along(VarNamesStates) {
  var_list_name <- paste0("inerp.df.", VarNamesStates[[Var]])
  var_list[[var_list_name]] <- dplyr::do(...)
  ...
  interp.df <- dplyr::inner_join(interp.df, var_list[[var_list_name]], by = ...)
```

But, if it works consistently, we can leave it for now. 

I _would_ use `dplyr::rename` instead of `plyr::rename` though -- `plyr` is a largely deprecated package that we should gradually remove from our code base in favor of more modern, better-maintained alternatives. You should be able to replace `plyr::rename(c("var" = VarNamesStates[Var]))` with `dplyr::rename(!!VarNamesStates[Var] := "var")`.

And actually, with modern `dplyr`, you can get away with skipping the `rename` step altogether by writing this as:

```r
dplyr::do(by.ens, !!VarNamesStates[[Var]] := interpolate(...))
```

(In a nutshell, in `dplyr` and related functions, `!!x := y` basically means "evaluate y and assign the result to a variable whose name is the value of x". If you're curious, this is called "Quasiquotation", and you can read more about it in the associated chapter of [Advanced R](https://adv-r.hadley.nz/quasiquotation.html#unquoting).).

models/sipnet/R/met2model.SIPNET.R
kzarada: I didn't use the POSIXt method because the time between changes based on which NOAA GEFS function is used. The downscale is hourly but the regular is 6 hours so using length.out gives us the proper doy with either situation

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
kzarada: This is from Luke's original code and he pulls out just the hour here and does other things with the hour below. The lubridate::floor returns the full date

ashiklom:
@kzarada Thanks for addressing all my comments! I haven't actually tested this code, but assuming you have, I think this is 99% of the way there. Two quick things:

(1) Please merge in the current `upstream/develop` branch ("Update branch" button above, or `git pull upstream develop` followed by `git push`). I could do this for you, but I'm generally hesitant to do that lest it introduce merge conflicts, or because I've occasionally been unpleasantly surprised when other people did this to my PRs in the past.

(2) Please run `make document` (or `devtools::document("modules/data.atmosphere")` and commit the resulting documentation changes. This is what is currently causing your Travis build to break.: https://travis-ci.org/PecanProject/pecan/jobs/543855926#L8400

kzarada:
@ashiklom I pulled in develop and I updated the documentation! Thanks for the comments.

kzarada:
@ashiklom @mdietze  the PR just passed the travis build! If you don't have any other comments, can we get this approved? @para2x and I need this to go through to move forward with the SDA

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -0,0 +1,345 @@
...
+##' \dontrun{
+##'  download.NOAA_GEFS(outfolder="~/Working/results", lat.in= 45.805925, lon.in = -90.07961, sitename="US-WCr")
+##' }
+##' 
+##' @author Katie Zarada - modified code from Luke Dramko and Laura Puckett
+##' 
+##' 
+
+
+download.NOAA_GEFS_downscale <- function(outfolder, lat.in, lon.in, sitename, start_date = Sys.time(), end_date = (as.POSIXct(start_date, tz="UTC") + lubridate::days(16)),
```
mdietze: I've got a high-level design concern here.
1) A very large fraction of the code here is redundant with Luke's download.NOAA_GEFS, which is bad design as we'll have to make any changes twice
2) Conceptually with our met workflow, download and downscaling are two different steps and should be in two different modules. I'd highly recommend blowing away all of Luke's code, just retaining the downscaling code, and renaming it to make it clear that it's just a downscaling function. 

In doing so, I'd ask if anything in the downscaling is actually GEFS specific, or whether it should work with any other 6hr data (or data at other frequencies, such as 3hr). I'd also say that in pulling the downscaling out you'll want to make sure that works on netCDF CF files, not GEFS internal file formats.

modules/data.atmosphere/R/met.process.R
```diff
@@ -181,7 +181,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
                                        host = host, 
                                        overwrite = overwrite$download,
                                        site = site, username = username)
-    if (met %in% c("CRUNCEP", "GFDL")) {
+    if (met %in% c("CRUNCEP", "GFDL","NOAA_GEFS_downscale")) {
```
mdietze: if downscale is split out, then this would be dropped

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
kzarada: @mdietze This current version is very GEFS specific. I put the downscaling within the download function so that the downscaling occurs prior to it being written out as netCDF files. I was working with code from the FLARE group and so it was already GEFS specific and it seemed to be the easiest/make the most sense to do it that way. I can change it to be a generalized downscaling, but that will take some planning and time and is more complicated than the current set up. Where would you like that to fall on my list of priorities?

mdietze: @kzarada looking at the code, I don't actually think that there's anything in the downscaling that's actually particularly GEFS specific. All other forecast and reanalysis products that need downscaling will have the same variables and the same (or similar) timestep.

But the question about priorities is a good one, but not one that I can answer until I see the rest of the summer workplan you were going to sketch out after our Friday meeting.  Knowing this is lower priority than other tasks, for now I'm going to accept the PR, but do so knowing that this is a temporary, partial solution that requires refactoring.  FWIW, the refactoring itself should take < 1 hr, but what's going to be more time consuming is the insertion of the resultant downscaling function into the met workflow.

PR #2357, opened 2019-05-17
remote.copy.from enhancement

### LINKED ISSUE #2355: Enchancement for remote.copy.from
## Description

**Is your feature request related to a problem? Please describe.**
With more and more people using remote clusters for performing compute-intensive PEcAn runs/analyses, there is now a need for not syncing everything back from the remote. This would also help with space issues on local clusters (like we experience at BU).

## Proposed Solution

**Describe the solution you'd like**
I was thinking of extending the existing [remote.copy.from](https://github.com/PecanProject/pecan/blob/develop/base/remote/R/remote.copy.from.R) function such that it accepts one or two additional arguments. For example, an `include` and a `rule` argument (for file filtering rules), and when they are not NULL the `rsync` command is modified accordingly within the function.

## Alternatives Considered

**Describe alternatives you've considered**
Alternatively a whole another function (or functions) can be added to the remote package, like `remote.copy.from.with.include` /  `remote.copy.from.with.rule`

istfer:
Small enhancement to remote.copy.from to allow users specify their own flags and arguments for rsync.

## Description

Added a new argument as suggested by @robkooper 
I also grep for the current instances in the repository, this change doesn't seem to require any related changes

## Motivation and Context

Closes #2355 

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/remote/R/remote.copy.from.R
+++ b/base/remote/R/remote.copy.from.R
@@ -19,7 +19,7 @@
 #'   host <- list(name='geo.bu.edu', user='kooper', tunnel='/tmp/geo.tunnel')
 #'   remote.copy.from(host, '/tmp/kooper', '/tmp/geo.tmp', delete=TRUE)
 #' }
-remote.copy.from <- function(host, src, dst, options=c(), delete = FALSE, stderr = FALSE) {
+remote.copy.from <- function(host, src, dst, options = c(), delete = FALSE, stderr = FALSE) {
   args <- c("-az", "-q", options)
   if (as.logical(delete)) {
     args <- c(args, "--delete")
--- a/base/remote/R/remote.copy.from.R
+++ b/base/remote/R/remote.copy.from.R
@@ -7,6 +7,7 @@
 #' @param host list with server, user and optionally tunnel to use.
 #' @param src remote file/dir to copy
 #' @param dst local file/dir to copy to
+#' @param options to be passed to rsync command, if nothing is specified everything will be rsynced
 #' @param delete in case of local dir should all non-existent files be removed
 #' @param stderr should stderr be returned
 #' @return output of command executed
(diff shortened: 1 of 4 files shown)


robkooper:
might be worth it to add it to `remote.copy.to` as well.

istfer:
makes sense

istfer:
add the same to `remote.copy.to` as well. grepped and this change also doesn't require any change in other calls

base/remote/R/remote.copy.from.R
```diff
@@ -18,8 +19,8 @@
 #'   host <- list(name='geo.bu.edu', user='kooper', tunnel='/tmp/geo.tunnel')
 #'   remote.copy.from(host, '/tmp/kooper', '/tmp/geo.tmp', delete=TRUE)
 #' }
-remote.copy.from <- function(host, src, dst, delete = FALSE, stderr = FALSE) {
-  args <- c("-az", "-q")
+remote.copy.from <- function(host, src, dst, options = c(), delete = FALSE, stderr = FALSE) {
```
ashiklom: I personally think `NULL` is a bit cleaner, easier to test against, and is more consistent with default arguments elsewhere in PEcAn.

```suggestion
remote.copy.from <- function(host, src, dst, options = NULL, delete = FALSE, stderr = FALSE) {
```

base/remote/R/remote.copy.to.R
```diff
@@ -16,8 +17,8 @@
 #'   host <- list(name='geo.bu.edu', user='kooper', tunnel='/tmp/geo.tunnel')
 #'   remote.copy.to(host, '/tmp/kooper', '/tmp/kooper', delete=TRUE)
 #' }
-remote.copy.to <- function(host, src, dst, delete = FALSE, stderr = FALSE) {
-  args <- c("-a", "-q")
+remote.copy.to <- function(host, src, dst, options = c(), delete = FALSE, stderr = FALSE) {
```
ashiklom:
```suggestion
remote.copy.to <- function(host, src, dst, options = NULL, delete = FALSE, stderr = FALSE) {
```

base/remote/R/remote.copy.from.R
ashiklom: (Note that you can apply both of these changes as a single commit if you go to the "Files" tab and use the "Add suggestion to batch" feature. Very handy!)

istfer: oh rad, I see what you mean, I'll do that next time!

PR #2363, opened 2019-05-28
ED: Make met2model entirely skip complete years

ashiklom:
## Description

Unless `overwrite = TRUE`, don't extract met for years for which every required target file is present. This should make `met2model.ED2` run much faster when extending an existing met record.

Tagging @istfer, @mdietze, @serbinsh, and @femeunier as major ED users.

## Motivation and Context

When updating existing ED met, `met2model` would extract variables, etc. for all years, including those for which files are already present. This can add up to a lot of wasted time.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


ashiklom:
> So we can still force an update if we think the source data has changed by using overwrite=TRUE

Yup -- with `overwrite = TRUE`, this should work exactly as it did before, except that it will list all of the files to be overwritten before launching into the loop over inputs/years.

> Does this mean this functionality has to be implemented on a model by model basis instead of at the level of the met process?

I think it has to be because of how different met formats are across models. For example, Something like this is conceptually impossible for SiPNET because its met driver is a single text file.

Issue #2301, opened 2019-02-26
write.params.CLM


dlebauer:

bailsofhay:
what's a trait value?

bailsofhay:
we need to go over this some more. I'm a little too new to pecan to get what you want done.

KristinaRiemer:
@amfox37 please post your matlab code that does this already

amfox37:
You can find the Matlab scripts I've used in the CLM_workflows directory under t3-david-lebauer in the shared Google Drive e.g. parameter_changer_*.m

Issue #392, opened 2015-03-21
PEcAn 2.0 list of non-backward compatible changes


dlebauer:
## Package names

all suggestions can be discussed in comments; checked when finalized

* [ ] no `.`'s that break Rcpp #391
* [ ] of format: "pecan<ACRONYM>"" pecanMA, pecanDB, pecanRTM, pecanVD
* [ ] convert long-names to acronyms (PEcAn.meta.analysis -> 

## Function names

* [ ] no `[generic].`s that falsely indicate S3 methods #380

## Other Conventions

ashiklom:
Two BETY-related changes I'd like to see, while they're fresh on my mind:

- Remove the restriction that sites require citations.
- Remove the restriction on author length for citations (BETY [PR 573](https://github.com/PecanProject/bety/pull/573) has been open for ages)
- Add `status` as a column in the BETY workflows (and runs?) table. Could be a JSON(-binary) field to give us flexibility. That would make it easier to check the status of workflows, and would dramatically accelerate the rendering of our History page (which IIRC currently parses the STATUS text file of every run) .

robkooper:
- make parameters in workflow be a jsonb field
- add generic key/value to BETY (https://github.com/PecanProject/bety/issues/505)

dlebauer:
- [ ] decouple from and decommission BETYdb
    - _in progress_
- [x] deprecate PEcAn web --> move to notebook based workflows
- [ ] remove unused code

dlebauer:
- remove unused code

PR #2365, opened 2019-05-28
initial revisions to write_params_ctsm function

dlebauer:
Ready to merge

references #2301

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,1284 @@
...
+    ## Loop over PFTS
+    # for testing
+    # trait.values <- list(c4_grass = list(vmax = 1, b0 = 2))
+    npft <- length(trait.values)
+    PEcAn.logger::logger.debug('there are ',
+                               npft,
+                               'PFTs in this run, they are named:',
+                               names(trait.values))
+    ctsm_pftnames <-
+      stringr::str_trim(tolower(ncdf4::ncvar_get(ctsm.param.nc, "pftname")))
```
ashiklom: FYI base R's `trimws` does the same thing as `stringr::str_trim`.

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,1284 @@
...
+    for (i in seq_len(npft)) {
+      pft <- trait.values[[i]]
+      pft.name <- names(trait.values)[i]
+      if (is.null(pft.name) | is.na(pft.name)) {
```
ashiklom: Use double `&&` and `||` in if-else statements.

Also, not strictly necessary, but w.r.t. `[i]` on the previous line, I've gotten into the habit of using `[[` for all subsetting where I'm expecting to return only a single value, even for atomic vectors. A nice advantage of that is that accidentally passing a vector (length > 1) to `[[` will throw an error, so it's a nice failsafe for extracting values that you know have to be scalars. It's definitely more a quirk of my own coding style rather than a convention we should adapt, but just thought I'd throw it out there.

KristinaRiemer:
@dlebauer why does this function name use underscores instead of periods?

ashiklom:
> why does this function name use underscores instead of periods?

@KristinaRiemer Apologies if someone already answered this offline, and additional apologies for the long-winded answer, but...

R uses dots in function names as part of its [S3 method system](https://adv-r.hadley.nz/s3.html), which allows functions of the same name to do different things for different objects. For example, `seq` will do different things when given a numeric input or a date input...

``` r
seq(1, 5)
#> [1] 1 2 3 4 5
seq(as.Date("2010-01-01"), as.Date("2010-02-01"), by = "3 days")
#>  [1] "2010-01-01" "2010-01-04" "2010-01-07" "2010-01-10" "2010-01-13"
#>  [6] "2010-01-16" "2010-01-19" "2010-01-22" "2010-01-25" "2010-01-28"
#> [11] "2010-01-31"
```

...because under the hood, it calls `seq.default` in the first case and `seq.Date` in the second case.

In general, the S3 method system uses the naming convention `function.class` -- i.e. `seq.Date` is a method of the generic function `seq` for the class `Date`. Similarly, `print.data.frame` is the `print` method for an object of class `data.frame`.

Unfortunately, in what I think is arguably the single worst design decision in R, this only happens when methods have been registered; otherwise, R will let you put as many dots in object names as you want. This can lead to ambiguity; for instance, is `data.frame` its own function, or the `data` method for an object of class `frame`? Similarly, is `t.test` its own function, or the `t` method for an object of class `test`? More scarily, it's _very_ easy to accidentally overwrite these. Note the errors that happen below:

``` r
data <- function(...) UseMethod("data")
x <- structure(list(x = 1, y = 2), class = "frame")
class(x)
#> [1] "frame"
data(x)
#> Error in as.data.frame.default(x[[i]], optional = TRUE, stringsAsFactors = stringsAsFactors): cannot coerce class '"frame"' to a data.frame

t <- function(...) UseMethod("t")
y <- structure(5, class = "test")
t(y)
#> Error in t.test.default(y): not enough 'x' observations
```

In both cases, we end up unintentionally calling `data.frame` and `t.test` functions.

So, in general, that's why we avoid dots in function names.

KristinaRiemer:
@ashiklom that's a great explanation, thanks for taking the time! I think I knew vaguely that using dots when naming functions was tricky, but I didn't know why. 

As a followup that, some of the Pecan functions have dots in them. Are they using the S3 system? All of the models in Pecan have write.config functions. So for write.config.ED2, is ED2 the class and write.config the function?

ashiklom:
> As a followup that, some of the Pecan functions have dots in them. Are they using the S3 system?

No, those aren't S3 methods, but rather a legacy of sloppy development at the beginning of PEcAn. We've kept them around as long as we have for backwards compatibility -- i.e. to avoid breaking all the external custom scripts people have written that use those functions. In fact, replacing `.` with `_` in all our function names is the first thing on our list of backwards-incompatible changes for PEcAn 2.0 (https://github.com/PecanProject/pecan/issues/392).

KristinaRiemer:
Gotcha! Okay, that really addresses my confusion about the inconsistency with using dots vs underscores. And now I know more about how/when to use dots. Thank so much!

KristinaRiemer:
Looking for feedback on this function now. A couple of things to pay attention to: 
1. The conversion on line 92 could probably be done in a better way
2. Similarly, the refactoring of the variable conversion (lines 85-92)
3. Need to make sure nothing crucial was deleted

KristinaRiemer:
https://github.com/danforthcenter/sentinel-detection/issues/53

KristinaRiemer:
https://github.com/danforthcenter/sentinel-detection/issues/53

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,103 @@
...
+           settings,
+           run.id) {
+    
+    ## Copy and open default parameter files
+    ctsm.param.default <-
+      system.file('clm5_params.c171117_0001.nc', package = 'PEcAn.CTSM')
+    if (!exists('local.rundir'))
+      local.rundir <- tempdir()
+    if (!exists('run.id'))
+      run.id <- 1
```
ashiklom: `local.rundir` should really be an argument to this function, with a default value of `NULL`. More generally, our preferred approach for defaults like this to either assign the default directly in the signature, or (for more complex default behavior) to set the default to `NULL` and test for that:

```r
write_params_ctsm <- function(..., run.id = 1, local.rundir = tempdir()) {...}
# ...or...
write_params_ctsm <- function(..., run.id = NULL, local.rundir = NULL) {
  if (is.null(run.id)) run.id <- 1
  if (is.null(local.rundir)) local.rundir <- tempdir()
  # ...
}
```

KristinaRiemer: Went with the simpler option for now!

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,101 @@
...
+      ## Loop over VARIABLES
+      pft_pecan_vals <- trait.values[[i]]
+      for (v in seq_along(pft_pecan_vals)) {
+        pecan_var <- names(pft_pecan_vals)[v]
 
-# write_params_fates ------------------------------------------------------
-write_params_fates <- function(defaults = system.file('????', package = 'PEcAn.FATES'),
-                              trait.values, settings, run.id){
+        ### ----- Leaf physiological parameters
+        update_vars <- function(nc = ctsm.param.nc, ctsm_var, ctsm_vals, start = ipft, count = 1){
```
ashiklom: In general, it's a good idea to stick function arguments with default values after all of the functions without defaults. That will also allow (if you want) to remove a bunch of the function argument names in the block below (i.e. you can just call `update_vars("slatop", udunits2::ud.convert(...))`.

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,101 @@
...
 
-  
-}
+# TODO: update this to read param files (CTSM and FATES) out of the refcase directory, not the PEcAn package
+# TODO: update to allow it to pick between CLM4.5 and CLM5 (CLM5 == CTSM) parameter set based on refcase, user selection
+# TODO: match Pecan pft names with CTSM pft names when Pecan name is not identical to a CTSM name
+## See issue https://github.com/PecanProject/pecan/issues/1008
 
+write_params_ctsm <-
+  function(defaults = system.file('clm5_params.c171117_0001.nc', package = 'PEcAn.CTSM'),
```
serbinsh: I would instead advocate that we use a refrun script, like the one I wrote for the Docker, that places the clm param file in the refrun directory such that we have the correct version in the refrun directory. Then we grab the default file from that directory instead of the pecan package directory

serbinsh: example: https://github.com/serbinsh/ctsm_containers/blob/master/ctsm_run_scripts/create_case_ctsmfates_1pt_example_PA-SLZ.sh

where we would have something like:

export CLM_PARAM_FILE_PATH={path to cesm datasets/lnd/clm2/}
export CLM_PARAM_FILE=clm5_params.c171117_0001.nc /new/run/dir

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,101 @@
...
+           trait.values,
+           settings,
+           run.id = 1, 
+           local.rundir = tempdir()) {
+    
+    ## Copy and open default parameter files
+    ctsm.param.default <-
+      system.file('clm5_params.c171117_0001.nc', package = 'PEcAn.CTSM')
```
serbinsh: so we would update this as well

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,101 @@
...
+          paste(
+            "Unmatched PFT",
+            pft.name,
+            "in CTSM PEcAn does not yet support non-default PFTs for this model"
+          )
+        )
+      }
+      
+      ## Special variables used in conversions
+      leafC <- 0.48
```
serbinsh: why?

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,101 @@
...
+      fnr <- 7.16 # mass ratio of total Rubisco molecular mass to nitrogen in Rubisco (g Rubisco g-1 N in Rubisco)
```
serbinsh: why isnt this an input parameter? At least we should make these settings in the xml (with defaults here) so that we can change, like we do with ED2

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,101 @@
...
+      ar <- 60 # specific activity of Rubisco (µmol CO2 g-1 Rubisco s-1)
```
serbinsh: same

models/ctsm/R/write_params-clm.R
KristinaRiemer: I just left it the way it already was, I agree that these should be arguments.

models/ctsm/R/write_params-clm.R
KristinaRiemer: @serbinsh I don't think I quite understand this. I've looked over your example, but that is part of creating a case for a CLM run, which is not what we want to do here? 

So we want to have a shell script with the two `export` lines (in models/ctsm?). This will indicate where the initial CLM parameter file is, which will not be in the Pecan repository but will only exists as a host machine file? Will it be an "inputs"-type file like in Bety? And will that shell script then be run by a Dockerfile? Which one?

dlebauer: @KristinaRiemer please move shawn's suggestions to new PR and then this can be merged

KristinaRiemer:
Moving refrun directory conversation over to a new issue #2388.

mdietze:
@serbinsh have your requested changes been addressed?

serbinsh:
> Moving refrun directory conversation over to a new issue #2388.

Great, thanks @KristinaRiemer

models/ctsm/R/write_params-clm.R
```diff
@@ -1,670 +1,102 @@
...
-#' @param defaults 
+#' @param defaults
 #' @param trait.values named list or data frame of traits, e.g.
 #' \code{data.frame(vmax = 1, b0 = 2)} or \code{list(vmax = 1, b0 = 2)}
-#' @param settings 
-#' @param run.id 
+#' @param settings
+#' @param run.id
+#' @param local.rundir location to store copied parameter file
+#' @param leafC
```
serbinsh: why is this here?  copy and past issue?

PR #2369, opened 2019-06-10
Adding BADM IC to the do_conversion function

para2x:
In the PR #2260 I added a series of functions that could let users to make initial conditions using BADM database. Here I add that capability to the `do_conversion` so that the users would need to add the `<poolinitcond>` tag under input and do_conversion would generate the IC and add that to the XML file.
So all you need to do is add the following tag to your inputs.
```
  <poolinitcond>
    <source>BADM</source>
  </poolinitcond>

```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/workflow/R/do_conversions.R
```diff
@@ -45,7 +46,7 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     
     # IC conversion : for now for ED only, hence the css/pss/site check
     # <useic>TRUE</useic>
-    if (ic.flag) {
+    if (ed.ic.flag) {
```
istfer: the ic_process has been built and tested with ED but it's not supposed to be for ED only. I feel like your BADM_IC_process  file should be called within the ic_process

para2x: Sure I can merge them into one.

PR #2374, opened 2019-06-11
DB: Bugfixes in `get.trait.data`

ashiklom:
Another bug found by @istfer. Basically, this code will fail if no traits have been found (common for, e.g., weird PFTs like SIPNET soil) because `trait.data` is an empty list (which breaks `dplyr::bind_rows`).


ashiklom:
I found and fixed a few more bugs while testing this, so there a few more changes to review here.

base/db/R/get.trait.data.R
```diff
@@ -189,26 +201,36 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon, trait.names,
...
-
-        if (nrow(diff_traits) > 0) {
-          diff_summary <- diff_traits %>%
-            dplyr::count(source, trait)
-          PEcAn.logger::logger.error(
-            "\n Prior has changed. \n",
-            "Here are the number of differing trait records by trait:\n",
-            PEcAn.logger::print2string(diff_summary),
-            wrap = FALSE
+        if (length(trait.data.check) != length(trait.data)) {
```
istfer: I dont't think `trait.data` is defined yet

istfer:
Also, the code thinks membership has changed, but it really is like a matter of reading it in

```
> diff_membership
    source   id   genus species scientificname
1 existing 1438 Unknown                Unknown
2  current 1438 Unknown    <NA>        Unknown
```

e.g if we change to `existing_membership <- utils::read.csv("/fs/data1/pecan.data/dbfiles/posterior/1000015654/species.csv",stringsAsFactors = FALSE, na.strings ="")` 

there is no difference anymore
```
> symmetric_setdiff(
+           dplyr::as_tibble(prior.distns, rownames = "trait"),
+           dplyr::as_tibble(existing_prior, rownames = "trait")
+         )
# A tibble: 0 x 6
# … with 6 variables: source <chr>, trait <chr>, distn <chr>, parama <dbl>, paramb <dbl>, n <int>
```

base/db/R/get.trait.data.R
```diff
@@ -145,15 +152,20 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon, trait.names,
...
         # Check if PFT membership has changed
         PEcAn.logger::logger.debug("Checking if PFT membership has changed.")
-        existing_membership <- utils::read.csv(need_paths[["pft_membership"]],
-                                               stringsAsFactors = FALSE)
+        existing_membership <- utils::read.csv(
+          need_paths[["pft_membership"]],
+          # Columns are: id, genus, species, scientificname
+          # Need this so NA values are
+          colClasses = c("double", "character", "character", "character"),
+          stringsAsFactors = FALSE
```
istfer:
```suggestion
          stringsAsFactors = FALSE,
          na.strings ="")
```

base/db/R/get.trait.data.R
ashiklom: Good catch. Fixing now.

ashiklom:
@istfer OK, take 3. This one comes with a unit test, so I _think_ I've caught all the issues this time.

ashiklom:
This currently passes all tests, but the [build breaks](https://travis-ci.org/PecanProject/pecan/jobs/546929342#L8428) because of an apparent change in `interactive.plotting.sda`. I ran `devtools::document` on this and it turned up no differences, so I think this must be a Travis bug (perhaps a mismatch in Roxygen versions?).

tonygardella:
Looks like it's breaking with:
```
```
Means you need to run `./scripts/generate_dependencies.R` and push up the changes to `pecan.depends`. Double check the order it puts packages in too, that mix up of order made me go crazy trying to find why my builds were breaking.

PR #2375, opened 2019-06-11
ED: Fix gsub pattern in model2netcdf.ED2

### LINKED ISSUE #2279: ED2 gsub wrong pattern in model2netcdf
### Bug Description
In the ED2 model2netcdf on [line 85](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L85) it defines the  `rundir` by using `gsub` to replace the pattern `out` with `run` in `outdir` . This causes problems when other parts of the path also have `out` in them, causing the next line to fail when looking for the `README.txt` in that `rundir` path.

Behavior:
`home/carya/output/PEcAn_900000001/out/`
is turned into:
`home/carya/runput/PEcAn_900000001/run/`
Expected to be:
`home/carya/output/PEcAn_900000001/run/`

## Machine (please complete the following information)
VM

## Additional context
Solution would probably be to parse the string after the `PEcAn_` then look for the `out` to replace with `run`. Paths after `PEcAn_` are standard so that should be safe across machines.

@serbinsh Tagging you on this because this was a line you edited last. Let me know if this doesn't make sense or was intended to work like this.

ashiklom:
Closes #2279.

### DIFF
--- a/models/ed/R/model2netcdf.ED2.R
+++ b/models/ed/R/model2netcdf.ED2.R
@@ -85,7 +85,7 @@ model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date,
     run_id <- basename(outdir)
     workflow_dir <- dirname(dirname(outdir))
     rundir <- file.path(workflow_dir, "run", run_id)
-    readme <- file(file.path(rundir, "README.txt"))
+    readme <- file.path(rundir, "README.txt")
     runtype <- readLines(readme, n = 1)
     if (grepl("ensemble", runtype)) {
       PEcAn.logger::logger.info("This is an ensemble run. ",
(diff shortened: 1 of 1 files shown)


models/ed/R/model2netcdf.ED2.R
```diff
@@ -82,17 +82,18 @@ model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date,
   year_check <- unique(unlist(ylist))
   if (max(year_check) < end_year) {
     PEcAn.logger::logger.info("Run failed with some outputs.")
-    rundir <- gsub(file.path("out"), file.path("run"), outdir)
+    run_id <- basename(outdir)
+    workflow_dir <- dirname(dirname(outdir))
+    rundir <- file.path(workflow_dir, "run", run_id)
     readme <- file(file.path(rundir, "README.txt"))
     runtype <- readLines(readme, n = 1)
-    close(readme)
```
robkooper: should the file be closed?

ashiklom: Ah shoot, yes it should -- I missed that it was opened via `file`. Though I'm not sure there's any benefit in opening and reading the file in separate steps, so I'll probably change this to just open and read the file in one step.

ashiklom:
Build passed. Quick -- merge it before something else in Travis breaks!

PR #2376, opened 2019-06-12
Add new functionality to meta.analysis for excluding greenhouse data

istfer:
The hierarchical version of the meta analysis wasn't converging for all parameters. One idea was to add the functionality to the meta analysis module to exclude greenhouse data which might be contributing to this issue. Indeed, my MA `RE on` models converged after excluding greenhouse data.

## Description

I extended the `<random.effects>` tag in the xml and added a `<use_ghs>` flag under it. If this flag is `FALSE`, greenhouse data is discarded from the traits data, **and if all the data for a trait was coming from greenhouse only, this trait is removed from the meta.analysis completely**. This flag defaults to `FALSE` for now, but I expect I need to change it to `TRUE` as previously published MA papers used greenhouse data.

## Motivation and Context

I use MA posteriors as priors in my multi-site calibrations and related UA/SA analyses. Some MA posteriors [for parameters that I expected to show site-to-site variability and contribute to uncertainty] were coming out very narrow with `RE off`. But `RE on` wasn't converging for all params. Hence the PR.

My questions to @mdietze @dlebauer @ashiklom and whomever is interested:

**(1) Most importantly: Was your suggestion to discard all greenhouse data, or just not to use the ghs effect?**

(2) As I mentioned above, the default behavior is to exclude greenhouse data, but previously it was to include it. Should I change it accordingly? Or do we like the new behavior better?

(3) The functionality implemented such that even if RE is off, one can choose to exclude greenhouse data:
```
<random.effects>
  <on>FALSE</on>
  <use_ghs>FALSE</use_ghs>
</random.effects>
```
Is this a behaviour we want?

(4) Can someone again explain to me the reasoning behind why we don't want to use any of the converged traits when one or more parameters were not converged? Everything in the meta.analysis module seems to be fitted univariate-ly. We [loop over traits](https://github.com/PecanProject/pecan/blob/develop/modules/meta.analysis/R/meta.analysis.R#L95) and run [single.MA](https://github.com/PecanProject/pecan/blob/develop/modules/meta.analysis/R/meta.analysis.R#L150). I feel like we should drop the nonconverged traits from the trait.mcmc so that [approx.posterior](https://github.com/PecanProject/pecan/blob/develop/modules/meta.analysis/R/run.meta.analysis.R#L120) doesn't try to fit a distribution over nonconverged samples, and keep using the MA posteriors downstream for whatever is converged.

(5) I'm afraid I don't know where to change the documentation. Could someone point me to it? Are there places where I need to make changes other than [this](https://github.com/PecanProject/pecan/blob/3b73b80a9b1bfe2d7cf38d8c4f06315934380b74/book_source/03_topical_pages/03_pecan_xml.Rmd)? @tonygardella 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/meta.analysis/R/jagify.R
```diff
@@ -70,7 +77,7 @@ transform.nas <- function(data) {
   data$site_id[is.na(data$site_id)] <- 0
   
   # greenhouse defaults to false (0)
-  data$greenhouse[is.na(data$greenhouse)] <- 1
+  data$greenhouse[is.na(data$greenhouse)] <- 0
```
istfer: @dlebauer this seemed like a typo, so I changed it from `1` to `0`, could you double-check that?

modules/meta.analysis/R/approx.posterior.R
```diff
@@ -75,7 +75,7 @@ approx.posterior <- function(trait.mcmc, priors, trait.data = NULL, outdir = NUL
       }
       posteriors[trait, "parama"] <- fit$estimate[1]
       posteriors[trait, "paramb"] <- fit$estimate[2]
-    } else if (pdist %in% zerobound | (pdist == "unif" & pparm[1] > 0)) {
+    } else if (pdist %in% zerobound | (pdist == "unif" & pparm[1] >= 0)) {
```
ashiklom: Probably should be `||` and `&&`.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -50,7 +50,17 @@ run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.
   dir.create(pathname, showWarnings = FALSE, recursive = TRUE)
   
   ## Convert data to format expected by pecan.ma
-  jagged.data <- lapply(trait.data, PEcAn.MA::jagify)
+  jagged.data <- lapply(trait.data, PEcAn.MA::jagify, use_ghs)
+  
+  if(!use_ghs){
+    # check if any data eft after excluding greenhouse
```
ashiklom:
```suggestion
    # check if any data left after excluding greenhouse
```

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -50,7 +50,17 @@ run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.
...
+    all_trait_check <- sapply(jagged.data, nrow)
+    if(any(all_trait_check == 0)){
+      nodat <- which(all_trait_check == 0)
+      jagged.data[nodat] <- NULL
+      PEcAn.logger::logger.info("No more data left after excluding greenhouse data for the following traits:", names(all_trait_check)[nodat])
```
ashiklom:
```suggestion
      PEcAn.logger::logger.info("No more data left after excluding greenhouse data for the following traits:", paste(names(all_trait_check)[nodat], collapse = ", "))
```

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -50,7 +50,17 @@ run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.
   dir.create(pathname, showWarnings = FALSE, recursive = TRUE)
   
   ## Convert data to format expected by pecan.ma
-  jagged.data <- lapply(trait.data, PEcAn.MA::jagify)
+  jagged.data <- lapply(trait.data, PEcAn.MA::jagify, use_ghs)
```
ashiklom: This might help avoid ambiguity if arguments get changed in the future.

```suggestion
  jagged.data <- lapply(trait.data, PEcAn.MA::jagify, use_ghs = use_ghs)
```

modules/meta.analysis/R/jagify.R
```diff
@@ -13,19 +13,26 @@
 ##' @name jagify
 ##' @title Prepare trait data for JAGS meta-analysis
 ##' @param result input trait data
+##' @param use_ghs do not exclude greenhouse data if TRUE
```
ashiklom:
```suggestion
##' @param use_ghs (Logical) If `FALSE`, exclude all greenhouse data. If `TRUE`, use all data, including greenhouse data.
```

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -165,12 +176,12 @@ run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.
...
 ##' @export
 ##' @author Shawn Serbin, David LeBauer
-run.meta.analysis <- function(pfts, iterations, random = TRUE, threshold = 1.2, dbfiles, database) {
+run.meta.analysis <- function(pfts, iterations, random = TRUE, threshold = 1.2, dbfiles, database, use_ghs = FALSE) {
   # process all pfts
   dbcon <- db.open(database)
   on.exit(db.close(dbcon))
 
-  result <- lapply(pfts, run.meta.analysis.pft, iterations, random, threshold, dbfiles, dbcon)
+  result <- lapply(pfts, run.meta.analysis.pft, iterations, random, threshold, dbfiles, dbcon, use_ghs)
```
ashiklom: I'd be inclined to make all of the fixed arguments (`iterations` on) explicitly named, so that we don't introduce subtle bugs if we ever change the argument order in `run.meta.analysis.pft`. But that may just be me being paranoid.

ashiklom:
> Was your suggestion to discard all greenhouse data, or just not to use the ghs effect?

I think these should both be included as two different configuration settings. I would consider something like `use_ghs_effect` to mean "include a greenhouse effect term in the meta analysis", and something like `include_ghs_data` to mean "include data from greenhouse treatments". `use_ghs_effect = TRUE` would necessarily mean `include_ghs_data = TRUE`, and a sensible default for both of these might be `FALSE`.

> As I mentioned above, the default behavior is to exclude greenhouse data, but previously it was to include it. Should I change it accordingly? Or do we like the new behavior better?

I think the new behavior (excluding greenhouse data) is probably a more sensible default, especially given that more people do modeling outside of greenhouse studies than inside them.

> even if RE is off, one can choose to exclude greenhouse data:

Shouldn't a greenhouse effect be a _fixed_ effect? Or do we only estimate the greenhouse effect when we use site random effects? If the former, we might want to stick this into the root of the `meta.analysis` in the XML, rather than in the `random.effects` subset.

> Can someone again explain to me the reasoning behind why we don't want to use any of the converged traits when one or more parameters were not converged?

If I understood @dlebauer correctly, I think by "parameters" he meant "parameters describing the meta-analysis for a single trait" -- e.g. `beta.o` vs. `tau`. For a given trait, if one of those parameters hasn't converged, then that means the whole model for that trait hasn't converged, and the meta-analysis result _for that trait_ is unreliable. On the other hand, I think you're absolutely right that, in our current univariate meta-analysis, one trait failing to converge only affects the result for that one trait and does _not_ mean the whole MA should be discarded.

> Are there places where I need to make changes other than this?

I think that's it for now. You may also consider revising the [meta analysis vignette](https://github.com/PecanProject/pecan/blob/3b73b80a9b1bfe2d7cf38d8c4f06315934380b74/modules/meta.analysis/vignettes/single.MA_demo.Rmd), but I personally don't think that needs to delay this PR.

modules/meta.analysis/R/run.meta.analysis.R
istfer: I personally prefer that too

istfer:
>If I understood @dlebauer correctly, I think by "parameters" he meant "parameters describing the meta-analysis for a single trait" -- e.g. beta.o vs. tau. For a given trait, if one of those parameters hasn't converged, then that means the whole model for that trait hasn't converged, and the meta-analysis result for that trait is unreliable. 

Ohh, that makes sense of course. I wasn't even considering such a case as "converged", got it.

istfer:
>Shouldn't a greenhouse effect be a fixed effect? Or do we only estimate the greenhouse effect when we use site random effects? If the former, we might want to stick this into the root of the meta.analysis in the XML, rather than in the random.effects subset.

Yes, it is fixed. If there are greenhouse data, `beta.ghs` goes into the model regardless `RE on/off`

modules/meta.analysis/R/jagify.R
dlebauer: no I think the original logic is correct: if no value is given for greenhouse, assume that the data are not from a greenhouse.

istfer: so in this case `1` is `not from a greenhouse`? I agree with the original logic, but it was `  data$greenhouse[is.na(data$greenhouse)] <- 1`?

serbinsh:
Did we decide to include the suggested modification by @ashiklom?  Or are we not interested in including GH data IF we dont then account for the difference in mean between natural and GH?  I could see that too, given the known differences...or am I mis-understanding?

mdietze:
Just realized that I didn't look at the discussion thread close enough, and was just giving the code a quick skip, and thus I missed that `use_ghs = FALSE` was the default. I don't think we should change the default behavior, as these data do provide information to the MA, but this PR should make it easier to exclude the ghs data for the cases where their inclusion is causing trouble (e.g. Istem's use case where there's only ghs data).

I also don't agree with the suggestion to separate `use_ghs_effect` and `include_ghs_data` as this seems to only have the effect of enabling two edge cases: one that doesn't seem desirable (including ghs data but not accounting for the biases present in such data) and another that is invalid (turning on a ghs effect while excluding ghs data).

ashiklom:
> I also don't agree with the suggestion to separate `use_ghs_effect` and `include_ghs_data` as this seems to only have the effect of enabling two edge cases: one that doesn't seem desirable (including ghs data but not accounting for the biases present in such data)...

My main concern here was with parameter identifiability under data limitation. In the most extreme case, how do we handle the case where _all_ of the data for a given trait-PFT combination comes from a greenhouse? Also, if we are so concerned with biases in trait data that we force users to either account for them or throw out large chunks of data, why do we ignore other important covariates like sun vs. shade?

> and another that is invalid (turning on a ghs effect while excluding ghs data)

I don't think this is a real issue since it's trivial to add an assertion that if `use_ghs_effect`, `include_ghs_data` must also be true.

mdietze:
| how do we handle the case where all of the data for a given trait-PFT combination comes from a greenhouse?

That's exactly the case being discussed as it's the one @istfer is hitting -- when 100% of the data is ghs the model fails and you have to turn "use_ghs" off, and then the posterior will just be the prior. The alternative is to generate a biased parameter, which doesn't seem preferable.

| why do we ignore other important covariates like sun vs. shade?

We don't, those covariates are used to filter data

ashiklom:
> The alternative is to generate a biased parameter, which doesn't seem preferable.

I agree that it's not ideal, and I don't think this should be the default behavior. I just think it would be nice to give users the _option_ (disabled by default) to include GHS data but omit the random effect, since I can envision cases where that might be useful.

> We don't, those covariates are used to filter data

Fair enough.

istfer:
>one that doesn't seem desirable (including ghs data but not accounting for the biases present in such data)

not sure if I'm following, isn't this the default behaviour we have now? we include ghs data but RE is off..

> I don't think we should change the default behavior, as these data do provide information to the MA
you mean RE on then? but as I mention, I think our current default is RE off, and use all the data

I'm just hitting this with BEAL where [this point](http://psql-pecan.bu.edu/bety/traits/40577) is potentially influential in MA but RE is off so I get a high SLA posterior post-MA

istfer:
oh wait sorry, forgot that beta.ghs goes into the model regardless, right..

tonygardella:
Looks like Line 4660:
```
test.read.settings.R:83: failure: check.settings gives sensible defaultss2$meta.analysis$random.effects isn't false.
```

tonygardella:
Specifically this test: https://github.com/PecanProject/pecan/blob/develop/base/settings/tests/testthat/test.read.settings.R

istfer:
ahh thanks Tony!

PR #2381, opened 2019-07-01
Adding ERA5 to met.process

para2x:
In this PR I'll try to add the required files/tags/settings to `met.process` and `convert.input` functions for digesting ERA5 met data.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


para2x:
@mdietze I tried so hard to not to change whatever that may affect the existing functionality. 
But since this is a big PR I'm trying to find someone on any other machine than BU to test the old and the new functionality to confirm that I'm not breaking anything before pulling this in. 
Besides that I think this PR is ready for review.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -21,85 +21,134 @@ ERA5_extract <-
...
-      # for each ensemble
-      one.year.out <- ensemblesN %>%
-        purrr::map(function(ens) {
+       #for each ensemble
+      one.year.out <- years %>%
+        purrr::map(function(year) {
+          
+        
+          tryCatch({
+            bety <- dplyr::src_postgres(dbname   = dbparms$dbname,
```
mdietze: 1) dbparms isn't documented as a function arg. 2) met workflow convention is to not create internal dependencies on the database as this prevents met functions from running remote.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -21,85 +21,134 @@ ERA5_extract <-
...
+          },
+          error = function(e) {
+            PEcAn.logger::logger.severe(paste0("",e))
+          }
+          )
+         
+
+          #--- lets find the big raw tile.
+          raw.tiles <- PEcAn.DB::dbfile.input.check(
+            siteid = "1000026755",
```
mdietze: In general, site ID shouldn't be hard coded. I suspect that you should be getting this info from the met registry file.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -21,85 +21,134 @@ ERA5_extract <-
...
+            con <- bety$con
+          },
+          error = function(e) {
+            PEcAn.logger::logger.severe(paste0("",e))
+          }
+          )
+         
+
+          #--- lets find the big raw tile.
+          raw.tiles <- PEcAn.DB::dbfile.input.check(
```
mdietze: I suspect this info should be pass into this function from the previous step in the workflow, not queried internally

modules/data.atmosphere/R/met.process.R
```diff
@@ -168,20 +172,28 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
-                                       start_date = start_date, end_date = end_date,
-                                       str_ns =str_ns_download, con = con, 
-                                       input_met = input_met, 
-                                       site.id = raw.data.site.id, 
-                                       lat.in = new.site$lat, lon.in = new.site$lon, 
-                                       host = host, 
-                                       overwrite = overwrite$download,
-                                       site = site, username = username)
-    if (met %in% c("CRUNCEP", "GFDL","NOAA_GEFS_downscale")) {
+    if (met %in% c("CRUNCEP", "GFDL","NOAA_GEFS_downscale","ERA5")) {
```
mdietze: Not your problem to solve, but more generally we should be creating a flag for this case in the registry file, rather than adding more and more special cases to the function itself.

modules/data.atmosphere/R/met.process.R
```diff
@@ -270,8 +282,34 @@ met.process <- function(site, input_met, start_date, end_date, model,
     reg.model.xml <- system.file(paste0("register.", model, ".xml"), package = paste0("PEcAn.",model))
     reg.model <- XML::xmlToList(XML::xmlParse(reg.model.xml))
     
-    met2model.result = list()
-    for (i in 1:length(ready.id[[1]])) {
+      met2model.result = list()
+    # Here I assume ensembles have one input id and multiple files related to them
+    if (!is.null(register$ensemble)){
```
mdietze: So if met.process was previously working for other ensemble met (e.g. GEFS, PalEON), why do we need a new/different behavior here? Wasn't the existing looping over ID's done specifically to support ensemble met?

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -21,28 +21,36 @@
 #' }
 met2cf.ERA5 <- function(lat,
                         long,
-                        years,
+                        start_date,
+                        end_date,
                         sitename,
-                        data.folder,
                         outfolder,
+                        dbparms,
```
mdietze: As noted previously, met functions shouldn't access the database. They shouldn't need to do so, and doing so breaks the ability to run them remote.

base/db/R/dbfiles.R
```diff
@@ -179,16 +180,19 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
 ##' @param hostname the name of the host where the file is stored, this will default to the name of the current machine
 ##' @param exact.dates setting to include start and end date in input query
 ##' @param pattern text to seach for in the file name (default NULL = no check).
+##' @param return.all (Logical) If 'TRUE', return all files. If 'FALSE', return only the most recent files.
```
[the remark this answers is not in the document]
para2x: `return.all` is an argument of  `dbfile.check` with FALSE default but in `dbfile.input.check` it was called but this argument was not passed to it. Here I just let added the same argument to be passed with the same default.

modules/data.atmosphere/R/download.ERA5.old.R
```diff
@@ -0,0 +1,197 @@
...
+#' files <- download.ERA5(
+#'   "ERA5_output",
+#'   start_date = "2010-01-01",
+#'   end_date = "2010-02-01",
+#'   lat.in = 45.5594,
+#'   lon.in = -84.6738,
+#'   product_types = "all"
+#' )
+#' }
+download.ERA5.old <- function(outfolder, start_date, end_date, lat.in, lon.in,
```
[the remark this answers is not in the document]
para2x: @mdietze I understand and I can sit with you to discuss this.
 What happened here is that the original `download.ERA5` function was @ashiklom 's old function (which was not complete), when I tried to go with this and see if I can download a site from ERA5 API, I realized that It keeps on the queue for more than 15 min to download just one single site. So I went with downloading all the raw tiles and the turn the download function to a function that just extract a site from already downloaded tiles. 
So I registered all the tiles to site called `USA` with that hard coded id, and then I go and extract sites from those tiles. That's also why I thought they need the db parameters.

base/db/R/dbfiles.R
```diff
@@ -24,6 +24,7 @@
 ##' @param con database connection object
 ##' @param hostname the name of the host where the file is stored, this will default to the name of the current machine
 ##' @param allow.conflicting.dates Whether to allow a new input record with same siteid, name, and format but different start/end dates
+##' @param ens In case of ensembles we could let to have more than one file associated with one input.
```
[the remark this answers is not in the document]
kzarada: Yes, each ensemble does have it's own input id

para2x:
@mdietze Would you please give this PR another look ? 
I think, I have resolved most of the previously discussed issues like the one to one mapping between the ensemble members and input id.

PR #2389, opened 2019-07-23
Add Setaria as genus for running BioCro

KristinaRiemer:
## Description
Added the genus *Setaria* to `call_biocro`

## Motivation and Context
We are using PEcAn to run BioCro on *Setaria viridis* for a project

## Review Time Estimate
- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
I assume this was considered already, but to check: Am I right that we want to eventually move away from having PEcAn-specific model parameter files in `models/biocro/inst/extdata/defaults/` and therefore it's intentional that this PR doesn't add a Setaria file there?

KristinaRiemer:
Oh, I somehow missed those genus-specific files when I was searching. Thoughts @dlebauer?

dlebauer:
good catch @infotroph yes, we will need to have a extdata/defaults/setaria.xml file. We can start by copying the panicum.xml file. 

@infotroph you are correct that we want to move away from pecan specific defaults but will wait until we start using v1.0 for that!

KristinaRiemer:
Okay, I've added that file. The only thing I changed between Sorghum and Setaria was the genus tag, should anything else be different?

PR #2390, opened 2019-07-24
Update Demo01.Rmd

bcow:
Removing shiny app from demo until it is out of GSoC development.


tonygardella:
I get that you want it to reflect current capability, but maybe just add a line saying "this is under development" rather than delete the whole section so that when it's back online, we can just delete one line of text instead of digging up the old text.

mdietze:
@bcow I concur with @tonygardella suggestion

PR #2395, opened 2019-08-19
LINKAGES pecan misc. met conversion fixes

marissakivi:
I adjusted two different scripts to fix three different issues identified in converting met for the LINKAGES model. 

In **met2model.LINKAGES**, when the script checks to see if there is already met input files extracted of the same type of met for the same site (lines 44-46), I added a check to make sure that the input file contains the required years for the current model run. If it does not include the years needed, it will overwrite the file so that it does. 

In that same script, I also adjusted how the script determines which data values are included in the monthly temperature and precipitation summations (lines 68-90, changed variable DOY_vec_hr). Instead of assuming the values are hourly, the script determines how frequently measurements are taken annually and uses that information to determine how many measurements should be summed for each month. My adjustment assumes the measurements are evenly spaced. 

Finally, in the **write.config.LINKAGES script**, I made a simple adjustment to how the temperature and precipitation storage matrices are created (lines 107-108). This ensures the matrices are two-dimensional (expected by linkages model) even if only one year of data is given. 

Thank you! Let me know if there is other information I should be including in these requests. 

Marissa

### DIFF
--- a/models/linkages/R/met2model.LINKAGES.R
+++ b/models/linkages/R/met2model.LINKAGES.R
@@ -29,6 +29,11 @@ met2model.LINKAGES <- function(in.path, in.prefix, outfolder, start_date, end_da
   out.file <- file.path(outfolder, "climate.Rdata")
   # out.file <- file.path(outfolder, paste(in.prefix, strptime(start_date, '%Y-%m-%d'),
   # strptime(end_date, '%Y-%m-%d'), 'dat', sep='.'))
+  
+  # get start/end year since inputs are specified on year basis
+  # use years to check if met data contains all of the necessary years 
+  start_year <- lubridate::year(start_date)
+  end_year <- lubridate::year(end_date)
 
   results <- data.frame(file = c(out.file),
                         host = c(PEcAn.remote::fqdn()),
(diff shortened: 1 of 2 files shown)


models/linkages/R/met2model.LINKAGES.R
```diff
@@ -79,6 +88,10 @@ met2model.LINKAGES <- function(in.path, in.prefix, outfolder, start_date, end_da
     sec <- udunits2::ud.convert(sec, unlist(strsplit(ncin$dim$time$units, " "))[1], "seconds")
     dt <- PEcAn.utils::seconds_in_year(as.numeric(year[i])) / length(sec)
     tstep <- 86400 / dt
+    
+    # adjust vector depending on the time step of data 
+    # assumes evenly-spaced measurements
+    DOY_vec_hr <- c(1, c(32, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 365) * as.integer(tstep))
```
mdietze: Does this work for leap years? I feel like there's a lubridate function that can give you this vector (rather than hard coding it) in a way that is leap year sensitive

marissakivi: Hi Mike! Thanks for looking this over. This is my first pull request. Do I need to make the above adjustments prior to the branch getting merged?

mdietze: Great question -- if the code already works for leap year correctly then I can pull it as is and you can just take the lubridate thing as a suggestion. If it's going to fail for leap years I'd prefer you use existing functions to handle that problem rather than ending up with a  bunch of ifs and alternative hard-coded vectors

PR #2401, opened 2019-09-03
Travis: Rebuild packages as needed on R 3.5

infotroph:
## Description
Avoids Travis failures on R-oldrel by checking the build information on all binary-installed R packages before attempting to load them, and rebuilding from source any that were built for a later R version than is currently running.

## Motivation and Context
All our Travis builds on R-oldrel (currently R 3.5) are failing with this cryptic error: 

```
Error in rbind(info, getNamespaceInfo(env, "S3methods")) : 
number of columns of matrices must match (see arg 2)
Calls: :: ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
```

The exact cause is kind of tedious...
<details>
 The process of loading an installed R package involves reading namespace its namespace info and adding all declared S3 methods to an internal table... and that table is apparently a 3-column matrix in R 3.5 and a four-column matrix in R 3.6.

Turns out that for reasons I don't understand, the `c2d4u3.5` package archive, despite its name, [now contains](https://twitter.com/marutterstat/status/1147478689904562176) a mix of packages built with R 3.5 and R 3.6. When we try to load a package built with R 3.6 into an R 3.5 session, the S3 method list has too many columns and the error happens.
</details>

...but [basically](https://stat.ethz.ch/pipermail/r-devel/2019-February/077298.html) when you see this message you should think "broken package that needs to be reinstalled." So that's what I do here: Immediately after apt finishes installing prebuilt R package binaries and before trying to load any of them, we check the "Built" field of their package descriptions and if any were built in a later version of R than is currently running, we suck it up and rebuild them. This is very slow the first time but they stay in the Travis cache for future builds.

Note that in *theory* it's unsafe to load any R package into a session with a different version than the one it was built for, but in practice we can often get away from it (note that I claimed c2d4u still contains some packages built for R 3.5, but the R 3.6 build doesn't seem to be having any trouble with them) and rebuilding packages is very slow, so for the moment I've implemented this check to only run when `getRversion()` is less than 3.6.

## N.B.
<del>This PR is built on top of the changes from #2400; please merge that before this one.</del><ins>Done!</ins>

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/scripts/travis/install.sh
+++ b/scripts/travis/install.sh
@@ -3,6 +3,15 @@
 set -e
 . $( dirname $0 )/func.sh
 
+# FIXING R BINARIES
+(
+    travis_time_start "pkg_version_check" "Checking R package binaries"
+
+    Rscript scripts/travis/rebuild_pkg_binaries.R
+
+    travis_time_end
+)
+
 # INSTALLING SIPNET
 (
     travis_time_start "install_sipnet" "Installing SIPNET for testing"
--- a/scripts/travis/rebuild_pkg_binaries.R
+++ b/scripts/travis/rebuild_pkg_binaries.R
@@ -1,0 +1,35 @@
+#!/usr/bin/env Rscript
+
+# ¡ugly hack!
+#
+# Travis setup uses many prebuilt R packages from c2d4u3.5, which despite its
+# name now contains some packages that were built for R 3.6. When loaded in
+# R 3.5.x, these throw error `rbind(info, getNamespaceInfo(env, "S3methods")):
+# number of columns of matrices must match`.
+#
+# We resolve this the slow brute-force way: By running this script before
+# attempting to load any R packages, thereby reinstalling from source any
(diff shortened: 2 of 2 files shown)


infotroph:
Note also that because of the way Travis caching rules work...

<details>
(To make it work in my fork, I used the Travis "Trigger build" menu with `script: echo nothing to see here` to generate an empty build, let that run to get the rebuilt packages into the cache, then reran the full build to verify. Travis doesn't allow `Trigger build` on PR branches, but I can do the same in the PEcAn develop branch after merging)
</details>

...I don't think it's possible to make the R oldrel build show success in this PR before it's merged. But you can see successful builds [here](https://travis-ci.org/infotroph/pecan/builds) by clicking through on recent entries from the `rebuild-mispackaged-binaries` branch.

PR #2174, opened 2018-11-02
update check_with_errors for new devtools

infotroph:
Fixes #2173

`devtools::check` now defaults to exiting with an error on any warning or error from `R CMD check`, so we need to specify `error="never"` so that the check_with_errors script can finish (<del>Otherwise it </del><ins>Also, an [unrelated issue](https://github.com/r-lib/devtools/issues/1914)</ins> will falsely report success with 0 warnings and 0 notes).

Note that this means `check_with_errors.R` is now redundant for its *original* purpose (which was basically to provide the behavior we could now get from `devtools::check(..., error_on="error")`), but I propose we keep it around and use it to start *promoting* specific warnings to errors as a path toward finishing #1949.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2404, opened 2019-09-04
Forbid new check warnings

infotroph:
## Description
<del>**Experimental**, opening PR for discussion but not confident this is the right approach.</del> Opinions, please!

Increments the strictness of PEcAn's package checks by forbidding newly added WARNINGs or NOTEs without breaking the build for the (many) pre-existing messages. 

In keeping with #1949, the idea here is to move toward eventually requiring all changes to maintain a clean `devtools::check` result: Many check messages indicate issues with potentially severe consequences even if are labeled as "WARNING" or "NOTE" rather than "ERROR", and it's preferable to to fix even the minor-looking issues so that we don't miss the severe ones in the noise.

Since #2174, all packages are checked during Travis builds and the build is marked broken if any ERRORs are found. It would be very simple to ratchet this up to break on any WARNING, but fixing all existing warnings from legacy code would be a big undertaking and we might accumulate new warnings faster than we fixed old ones.

What I proposed in #1949 was to blacklist a slowly growing list of warning types, e.g. fix all `'::' or ':::' import not declared` and break the build on new ones, then tackle `Undocumented code objects`, and so on. On consideration this still feels Sisyphean, so what I propose here is an across-the-board grandfather clause: We cache current check outputs and break the build on any message *newly added* by the current build, while continuing to let existing messages slide like we have been doing. This should enforce good practice in newly added code without forcing too much immediate cleanup in the old stuff. As historic warnings get cleaned up in future refactorings, we can update the cached check results to make sure they stay gone. 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
Sample of Travis output when a new problem is detected: https://travis-ci.org/infotroph/pecan/jobs/580354445#L6106

scripts/check_with_errors.R
```diff
@@ -71,3 +71,68 @@ message(n_notes, ' notes found in ', pkg, '.')
...
+old <- rcmdcheck::parse_check(old_file)
+cmp <- rcmdcheck::compare_checks(old, chk)
+
+
+if (cmp$status != "+") {
+    # rcmdcheck found new messages, so check has failed
+    print(cmp)
+    cat("R check of", pkg, "reports the following new problems.",
+        "Please fix these and resubmit:\n")
+    cat(cmp$cmp$output[cmp$cmp$change == 1], sep="\n")
```
ashiklom: Any reason not to use `message` here? I guess so this goes to `stdout` instead of `stderr`?

ashiklom:
It's still marked as "Draft", so not sure if a formal review is appropriate, but in general, this looks super useful -- thanks for putting it together!

scripts/check_with_errors.R
infotroph: Mostly because the existing `check_with_errors` used cat for its (off-by-default) printing and because it was easy to reason about. For Travis purposes I haven't usually worried much about stdout vs stderr, but I could be persuaded.

infotroph:
@ashiklom I marked this as draft to emphasize that I wanted feedback before it was merged, so reviews are very helpful. 

If the overall approach seems sound, I'll make whatever other tweaks are needed to make the build pass and then solicit a final round of review.

base/workflow/R/create_execute_test_xml.R
```diff
@@ -22,8 +22,11 @@
 #'   (character) BETY database connection options. Default values for all of
 #'   these are pulled from `<pecan_path>/web/config.php`.
 #' @param db_bety_driver (character) BETY database connection driver (default = `"Postgres"`)
-#' @return
+#' @return A list with two entries:
+#'  * `sys`: Exit value returned by the workflow (0 for sucess).
+#'  * `outdir`: Path where the workflow results are saved
 #' @author Alexey Shiklomanov, Tony Gardella
+#' @importFrom dplyr %>% .data
```
ashiklom: `.data` is originally from `rlang` and is re-exported by `dplyr`. Is it standard practice to import from packages that re-export functions? I can rationalize both approaches -- doing it this way reduces the length of our top-level dependencies list, and people are much more likely to know `dplyr` than `rlang`. On the other hand, importing the original function points you more directly to its documentation.

No need to change this here -- just curious about your thoughts on these strategies in the future.

infotroph: I wondered about this too -- I'm not sure what's standard, but if we can find out we should follow the standard. I'll ask Twitter and report back.

My reasoning right now is that (1) `dplyr` re-exports functions precisely to make them easy to use them inside dplyr as we are here; and that (2) if the dplyr-exported and rlang-exported versions of a function were to differ in any way, we can hope the version dplyr exports would be the version dplyr was most compatible with.

ashiklom:
I would like to see this go in sooner rather than later, especially given the flurry of PEcAn development that will be happening in anticipating of upcoming meetings. So feel free to drop the "Draft" status whenever you're ready so we can give this a formal review.

If you have the time, one thing that would help this PR is adding a brief (maybe even just in bullet form) section on "Continuous Integration / Travis" to our documentation. I think this would work as either an appendix, a sub-section of "Developer Guide -> Testing", or maybe even a full section under "Developer Guide". I would be perfectly happy with just a list of checks that are performed on Travis and a few notes about solving common errors (e.g. forgetting to run `make document`, fixing new "check" messages per this PR, re-starting the build if it times out).

base/workflow/R/create_execute_test_xml.R
infotroph: I mean, that's exactly what happened to `devtools::install_deps` in the 1.0->2.0 interval, soooooo.

robkooper:
Do we need all the `Rcheck_reference.log` or those there as a what still needs to be fixed, or does the code use this to see if new warnings/errors are produced?

infotroph:
@robkooper We need them (the code compares each new check result against `Rcheck_reference.log` to find out what errors are new), but I'm not certain `tests` is the best place to store them and would welcome suggestions.

base/workflow/R/create_execute_test_xml.R
```diff
@@ -22,8 +22,11 @@
 #'   (character) BETY database connection options. Default values for all of
 #'   these are pulled from `<pecan_path>/web/config.php`.
 #' @param db_bety_driver (character) BETY database connection driver (default = `"Postgres"`)
-#' @return
+#' @return A list with two entries:
+#'  * `sys`: Exit value returned by the workflow (0 for sucess).
+#'  * `outdir`: Path where the workflow results are saved
```
infotroph: @ashiklom please verify this is a sensible return type description

scripts/check_with_errors.R
```diff
@@ -71,3 +61,89 @@ message(n_notes, ' notes found in ', pkg, '.')
...
+msg_lines <- function(msg){
+    msg <- strsplit(
+        gsub("\n  ", " ", msg, fixed = TRUE), #leading double-space indicates line wrap
+        split = "\n",
+        fixed = TRUE)
+    msg <- lapply(msg, function(x)x[x != ""])
+    unlist(lapply(msg, function(x)paste(x[[1]], x[-1], sep=": ")))
+}
+
+old_file <- file.path(pkg, "tests", "Rcheck_reference.log")
```
robkooper: This answers my own question about the `Rcheck_reference.log`

infotroph:
TODO before merging: Get last few errors through Travis

Optionally, could punt to a separate PR: Documentation per @ashiklom  request

book_source/02_demos_tutorials_workflows/05_developer_workflows/04-testing.Rmd
```diff
@@ -118,4 +118,56 @@ The `batch_run.R` script can take the following command-line arguments:
...
-[xml_fun]: 
+[xml_fun]:
+
+### Continuous Integration
+
+Every time anyone commits a change to the PEcAn code, the act of pushing to GitHub triggers an automated build and test of the full PEcAn codebase using [Travis CI](https://travis-ci.org/pecanProject/pecan), and all pull requests must report a successful CI build before they will be merged. This will sometimes feel like a burden when the build breaks on an issue that looks trivial, but anything that breaks the build is important enough to fix. It's much better to find errors early and fix them before they get incorporated into the released PEcAn code.
+
+At this writing (September 2019), all our Travis builds run on the same version of Linux (currently Ubuntu 16.04) using three different versions of R in parallel: previous release, current release, and nightly builds of the R development branch. In most cases the build should pass on all three versions, but only the current release (`R:release`) is truly mandatory. We aim to fix errors found on R:release before merging, errors found on R:devel before the next R release, and errors found on R:oldrel as developer time and forward compatibility allow.
```
ashiklom: I can never remember which Ubuntu version numbers correspond to which animal names, so might be helpful to add here (especially since it's in the Travis config).

```suggestion
At this writing (September 2019), all our Travis builds run on the same version of Linux (currently Ubuntu 16.04, `xenial`) using three different versions of R in parallel: previous release, current release, and nightly builds of the R development branch. In most cases the build should pass on all three versions, but only the current release (`R:release`) is truly mandatory. We aim to fix errors found on R:release before merging, errors found on R:devel before the next R release, and errors found on R:oldrel as developer time and forward compatibility allow.
```

book_source/02_demos_tutorials_workflows/05_developer_workflows/04-testing.Rmd
```diff
@@ -118,4 +118,56 @@ The `batch_run.R` script can take the following command-line arguments:
...
+* Initializes a skeleton version of the PEcAn database (BeTY) containing a few public records to be used by the test runs.
+* Installs all the PEcAn packages, recompiling documentation and installing dependencies as needed.
+* Runs package unit tests (the same ones you run locally with `make test` or `devtools::test(pkgname)`).
+    - As discussed in [Unit testing](#developer-testing-unit), these tests should run quickly and test individual components in relative isolation.
+    - Any test that calls the `skip_on_ci` function will be skipped. This is useful for tests that need to run for a very long time or require resources that aren't available on Travis (e.g. specific models), but be sure to run these tests locally before pushing your code!
```
ashiklom: Add data product download as example of long-running test?

```suggestion
    - Any test that calls the `skip_on_ci` function will be skipped. This is useful for tests that need to run for a very long time (e.g. large data product downloads) or require resources that aren't available on Travis (e.g. specific models), but be sure to run these tests locally before pushing your code!
```

book_source/02_demos_tutorials_workflows/05_developer_workflows/04-testing.Rmd
```diff
@@ -118,4 +118,56 @@ The `batch_run.R` script can take the following command-line arguments:
...
+* Clones the PEcAn repository from GitHub, and checks out the branch to be tested.
+* Retrieves any cached files available from previous Travis builds.
+    - The main thing in the cache is previously-installed dependency R packages, to avoid recompiling them every time.
+    - If the cache becomes stale or is preventing a package update needed by the build (e.g. to get a new version that contains a needed bug fix), delete the cache through the Travis web interface and it will be reconstructed on the next build.
+    - Because PEcAn has so many dependencies, builds with no cache will spend most of their time recompiling packages and will often run out of time before the tests complete. You can fix this by [triggering a custom build](https://blog.travis-ci.com/2017-08-24-trigger-custom-build) using a custom config that installs but does not test PEcAn (e.g. `script: make install`), waiting for this to finish and upload its new cache, then restarting the standard full build.
```
ashiklom: This is an excellent suggestion. I did not know about this workaround.

ashiklom:
@mdietze Unless I'm missing something, I only see one new `get.model.output` function in this PR -- `get.model.output.dalec` -- and it's just there to replace [hand-written documentation about a non-existent `get.model.output.generic` function](https://github.com/PecanProject/pecan/blob/develop/models/dalec/man/get.model.output.generic.Rd). We currently have a bunch of other `get.model.output.[MODEL]` functions (e.g. [`get.model.output.ED2`](https://github.com/PecanProject/pecan/blob/aca706574268e74ee21611e33e6276a7ef8abfe4/models/ed/R/get.model.output.ed.R#L96), [`get.model.output.BIOCRO`](https://github.com/PecanProject/pecan/blob/aca706574268e74ee21611e33e6276a7ef8abfe4/models/biocro/R/get.model.output.BIOCRO.R#L17)) that are real functions that do real things.

That said, for the `get.model.output.dalec` specifically, it may be worth either (1) having this function throw an error that says it hasn't yet been implemented, or (2) just deleting it altogether to avoid confusion.

infotroph:
@mdietze the `man/` directory for DALEC contained a hand-written Rd file whose \title was `get.model.output.generic` but its \usage showed `get.model.output.dalec`, leading to a check complaint about "code/documentation mismatch".

No idea why this showed up now but not in the stored check results file, but I figured the quickest fix was to add the dummy function so that Roxygen would regenerate the Rd with correct formatting (and with the correct filename).

PR #2405, opened 2019-09-04
Fix for the ensemble weights in refactored/single site  SDA

para2x:
After a short discussion with @kzarada and @araiho, I realized I pulled something from @araiho PR into the mainline which resulted a small bug for @kzarada project. After another discussion with @araiho here is a small fix for that bug.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -160,16 +160,22 @@ sda.enkf <- function(settings,
...
-    for(tt in 1:length(obs.times)){
-      weight_list[[tt]] <- rep(1,nens) #no weights
+  
+  # This reads ensemble weights generated by `get_ensemble_weights` function from assim.sequential package
+  if(!is.null(settings$run$inputs$ensembleweights$path)){
+    if(!file.exists(file.path(settings$outdir, "ensemble_weights.Rdata"))){
+      PEcAn.logger::logger.warn("ensemble_weights.Rdata cannot be found. Make sure you generate samples by running the get.ensemble.weights function before running SDA if you want the ensembles to be weighted.")
+      #create null list
+      for(tt in 1:length(obs.times)){
+        weight_list[[tt]] <- rep(1,nens) #no weights
```
mdietze: just checking, but doesn't this result in weight_list not being defined at all if `settings$run$inputs$ensembleweights$path` is NULL? That seems dangerous. To me it seems like we should run this for loop either when the file doesn't exist or when the path is null.

Also, what's the point of defining `settings$run$inputs$ensembleweights$path` if that path isn't the one being used to open the file? `load(file.path(settings$outdir, "ensemble_weights.Rdata"))`

para2x: My mistake this would have been fine with the original piece, The bug has to do with the change in the restart functionality which I'm trying to resolve.

PR #2351, opened 2019-05-08
`PEcAn.DB` and `PEcAn.meta.analysis` cleanup

ashiklom:
- **Breaking change**: In `PEcAn.DB::get.trait.data`, if `trait.names` is `NULL` or missing, use the traits for which at least one prior is available among the input list of PFTs. (Previously, we were getting this from the `PEcAn.utils::trait.dictionary`, which we are trying to deprecate #1747 ).
- **Feature**: New function `PEcAn.DB::query_priors` that is, IMHO, more robust and intuitive than `query.priors` by leveraging `RPostgres` prepared statements, providing more informative errors, and handling inputs in a more sophisticated way. Its output should be a perfect superset of `query.traits`, so I think it should work as a drop-in replacement. Note that a unit test and detailed function documentation are included.
- **Bugfixes and cleanup**:
  - Use explicit namespacing (`package::function`) throughout `PEcAn.meta.analysis`. Otherwise, many of these functions would fail when trying to run a meta-analysis outside of the PEcAn workflow (i.e. without having loaded the packages first).
  - Documentation and code style cleanup throughout a lot of `PEcAn.meta.analysis` and `PEcAn.DB`.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue) 
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the CHANGELOG.md.
- [x] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #2262, opened 2019-01-28
Issue with modellauncher


serbinsh:
```
2019-01-21 23:32:33 INFO   [PEcAn.workflow::run.write.configs] :
   parameter values for runs in
   /data/sserbin/Modeling/sipnet/US-WCr/ILAMB_CRUNCEP.TempBDF.2//samples.RData
2019-01-21 23:32:34 INFO   [start.model.runs] :
   -------------------------------------------------------------------
2019-01-21 23:32:34 INFO   [start.model.runs] :
   Starting model runs SIPNET
2019-01-21 23:32:34 INFO   [start.model.runs] :
   -------------------------------------------------------------------
  |                                                                      |   0%Warning in file.remove(file.path(run_id_dir, "joblist.txt")) :
  cannot remove file '/data/sserbin/Modeling/sipnet/US-WCr/ILAMB_CRUNCEP.TempBDF.2//run/2000098933/joblist.txt', reason 'No such file or directory'
Error in writeLines(c(file.path(settings$host$rundir, run_id_string)),  :
  'con' is not a connection
Calls: <Anonymous> -> start.model.runs -> writeLines
```

seems it is looking for a "joblist.txt" file to remove it and also hits a complaint about a database connection.  I built the model launcher exe and setup my xml like this

```
    <qsub>qsub -l walltime=36:00:00,nodes=2:ppn=10 -V -N @NAME@ -o @STDOUT@ -e @STDERR@ -S /bin/bash</qsub>
    <qsub.jobid>([[:digit:]]+)\.modex\.bnl\.gov</qsub.jobid>
    <qstat>qstat @JOBID@ || echo DONE</qstat>
    <modellauncher>
        <binary>/data/sserbin/Modeling/pecan/contrib/modellauncher/modellauncher</binary>
        <qsub.extra>-l ncpus=10</qsub.extra>
    </modellauncher>
  </host>
```

serbinsh:
Are there any folks in PEcAn-land successfully using model launcher in pecan develop?

tonygardella:
@femeunier

robkooper:
problem looks to be in https://github.com/PecanProject/pecan/blob/develop/base/remote/R/start.model.runs.R#L90 the jobfile used to be opened for writing here.

serbinsh:
Ahhh....ok. that makes sense I think....so does it need to be open for writing again?

serbinsh:
[...]

PR #2409, opened 2019-09-05
release 1.7.1 of PEcAn

robkooper:
### Fixes
- Fixed issue that prevented modellauncher from working properly #2262
- Use explicit namespacing (`package::function`) throughout `PEcAn.meta.analysis`. Otherwise, many of these functions would fail when trying to run a meta-analysis outside of the PEcAn workflow (i.e. without having loaded the packages first) (#2351).
- Standardize how `PEcAn.DB` tests create database connections, and make sure tests work with both the newer `Postgres` and older `PostgreSQL` drivers (#2351).
- Meta-analysis = "AUTO" now correctly skips the meta analysis if the PFT definition has not changed (#1217).
- Replace deprecated `rlang::UQ` syntax with the recommended `!!`
- Explicitly use `PEcAn.uncertainty::read.ensemble.output` in `PEcAn.utils::get.results`. Otherwise, it would sometimes use the deprecated `PEcAn.utils::read.ensemble.output` version.
- History page would not pass the hostname parameter when showing a running workflow, this would result in the running page showing an error.

### Changed
- Updated modules/rtm PROSPECT docs
- Updated models/sipnet/R/model2netcdf.SIPNET.R to address issues in PR #2254 
- Improved testing (#2281). Automatic Travis CI builds of PEcAn on are now run using three versions of R in parallel. This should mean fewer issues with new releases and better backwards compatibility, but note that we still only guarantee full compatibility with the current release version of R. The tested versions are:
  - `release`, the current public release of R (currently R 3.5). Build failures in this version are fixed before merging the change that caused them. When we say PEcAn is fully tested and working, this is the build we mean.
  - `devel`, the newest available development build of R. We will fix issues with this version before the next major R release.
  - `oldrel`, the previous major release of R (currently R 3.4). We will fix issues with this version as time allows, but we do not guarantee that it will stay compatible.
- Reverting back from PR (#2137) to fix issues with MAAT wrappers.
- Moved docker files for models into model specific folder, for example Dockerfile for sipnet now is in models/sipnet/Dockerfile.
- `PEcAn.utils`:
  - Remove, or make "Suggests", a bunch of relatively underutilized R package dependencies.
- Add template for documentation issues and add button to edit book.
- Conditionally skip unit tests for downloading raw met data or querying the database when the required remote connection is not available.
- Reorganization of docker folder
  - All dockerfiles now live in their own folder
  - `scripts/generate_dependencies.R` is now used to generate dependencies for make and docker
- In `PEcAn.DB::get.trait.data`, if `trait.names` is `NULL` or missing, use the traits for which at least one prior is available among the input list of PFTs. (Previously, we were getting this from the `PEcAn.utils::trait.dictionary`, which we are trying to deprecate #1747). (#2351)
- Cleanup and improve logging and code readability in parts of `PEcAn.DB` related to getting trait data, including replacing many manual database queries with `dplyr` calls.
- Reorganization of PEcAn documentation in accordance with isue #2253.
- SIPNET now is installed from the source code managed in git

### Added
- Meta analysis functionality to not use greenhouse data.
- Dockerize the BioCro model.
- Added PRO4SAIL-D model, using existing 4SAIL src and coupling with PROSPECT-D Fortran code
- Models will not advertise themselvs, so no need to register them a-priori with the database #2158
- Added simple Docker container to show all containers that are available (http://localhost:8000/monitor/). This will also take care of registering the models with the BETY database.
- Added unit tests for `met2model.<MODEL>` functions for most models.
- Added MAESPA model to docker build
- PEcAn has more robust support for `RPostgres::Postgres` backend. The backend is officially supported by `db.query`, and basic workflows run top-to-bottom with the `Postgres` backend. However, `RPostgreSQL` is still the default until we do more robust testing of all modules.
- `PEcAn.DB::db.query` now optionally supports prepared statements (#395).
- New function `PEcAn.DB::query_priors` that expands the functionality of `query.priors` by (1) accepting PFTs by name or ID and (2) allowing the user to request all possible combinations of the input PFTs and traits (i.e. `expand.grid(pfts, traits)`) or just the pairwise combinations (i.e. `pft[1]-trait[1], pft[2]-trait[2]`). This function also comes with more robust error handling and a set of unit tests (#2351).
- New function `PEcAn.DB::query_pfts` for finding PFT IDs and types from the PFT name and (optionally) model type (#2351).
- Run Travis integration tests with both Postgres and PostgreSQL drivers (#2351).
- New function `PEcAn.utils::load_local` reads `Rdata` files into a named list (instead of into the current environment).

### Removed
- Removed unused function `PEcAn.visualization::points2county`, thus removing many indirect dependencies by no longer importing the `earth` package.
- Removed package `PEcAn.data.mining` from the Make build. It can still be installed directly from R if desired, but is skipped by default because it is in early development, does not yet export any functions, and creates a dependency on the (large, often annoying to install) ImageMagick library.
- Fully deprecate support for `MySQL` database driver. Now, only `PostgreSQL` (and, experimentally, `RPostgres`) are supported. With this, remove `RMySQL` dependency in several places.

### Fixed
- Replace deprecated `rlang::UQ` syntax with the recommended `!!`
- Explicitly use `PEcAn.uncertainty::read.ensemble.output` in `PEcAn.utils::get.results`. Otherwise, it would sometimes use the deprecated `PEcAn.utils::read.ensemble.output` version.
- `PEcAn.ED2::met2model.ED2` now skips processing of years for which all output files are already present (unless `overwrite = TRUE`). This prevents a lot of unnecessary work when extending an existing ED met record.


dlebauer:
When posting the release notes,
- There are two sections 'fixed' and 'fixes' should these be merged?
- Also if it starts with new features (added) then changed then fixed, the exciting stuff comes first.

PR #2410, opened 2019-09-09
Return error message if no meta-analysis posterior exists

KristinaRiemer:
When pulling trait data from the database, we ran into the following error: 
```[PEcAn.DB::get.trait.data] : 
   `trait.names` is NULL, so retrieving all traits that have at least one 
   prior for these PFTs. 
Error in .subset2(x, i) : subscript out of bounds
```

This occurred because we were using a new PFT that had no data associated with it, so the meta-analysis didn't produce any posteriors. 

## Description
Changed `get.trait.data.pft` so that, if the posterior table has no rows, it returns a message ("No previous posterior found. Forcing update") instead of an error due to an empty table. 

## Motivation and Context
Though it probably won't be common for people to run new PFTs with no data, it's possible for it to happen like it did in our case and there's no reason for it to disrupt the workflow. 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:

I need help with making sure this is the best way to check if there are posteriors (lines 114-116) and with writing the test for this part of the function.

### DIFF
--- a/base/db/R/get.trait.data.R
+++ b/base/db/R/get.trait.data.R
@@ -111,11 +111,18 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon, trait.names,
   # check to see if we need to update
   if (!forceupdate) {
     if (is.null(pft$posteriorid)) {
-      pft$posteriorid <- dplyr::tbl(dbcon, "posteriors") %>%
-        dplyr::filter(pft_id == !!pftid) %>%
-        dplyr::arrange(dplyr::desc(created_at)) %>%
-        head(1) %>%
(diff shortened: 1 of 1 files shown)


ashiklom:
This is an important bugfix, so I would vote we merge this without the unit test and open an issue (and eventually, a separate PR) about adding a test to cover this case.

ashiklom:
(My approval above is conditioned on removing the test pseudocode).

KristinaRiemer:
@ashiklom pseudocode is removed and new issue is #2413.

PR #2416, opened 2019-09-13
Dividing get.trait.data.R and query.trait.data.R 

bcow:
## Description

The `get.trait.data.R` and `query.trait.data.R` scripts in the `PEcAn.DB` package each contained multiple functions which made them hard to read and debug. 

In this PR I have separated the code in to multiple files. 
I have NOT edited the code in any way (besides auto realignment) and I have NOT changed function names. Thus this should not change the way pecan runs for anyone. 

I did leave a few functions together in a script called `covariate.functions.R` because they were short and they had been grouped together in the original script. If people think that these small functions should also be separated in to their own functions, that is easy to do. 

In future PR's I will address possible editing/renaming of functions.


base/db/R/query.trait.data.R
```diff
@@ -6,417 +6,6 @@
...
-        "No control treatment set for site_id ", unique(data$site_id[site.i]),
-        " and citation id ", unique(data$citation_id[site.i]), ".\n",
-        "Please set control treatment for this site / citation in database.\n"
-      ))
-    }
-  }
-  return(data)
-}
-
-drop.columns <- function(data, columns){
```
infotroph: I don't see `drop.columns` in any of the added files. Accidental removal, or am I missing it?

...Hmm, but it also seems unused, so maybe this is / should be an intentional removal?

base/db/R/check.lists.R
```diff
@@ -0,0 +1,35 @@
...
+##--------------------------------------------------------------------------------------------------#
+##' Check two lists. Identical does not work since one can be loaded
+##' from the database and the other from a CSV file.
+##'
+##' @name check.lists
+##' @title Compares two lists
+##' @param x first list
+##' @param y second list
+##' @param filename one of "species.csv" or "cultivars.csv"
+##' @return true if two list are the same
```
dlebauer:
```suggestion
##' @return true if two lists are the same
```

base/db/R/assign.treatments.R
```diff
@@ -0,0 +1,48 @@
+#-------------------------------------------------------------------------------
```
dlebauer: Technically this is a covariate fn so could be included in that file

base/db/R/assign.treatments.R
```diff
@@ -0,0 +1,48 @@
...
+# which accompanies this distribution, and is available at
+# http://opensource.ncsa.illinois.edu/license.html
+#-------------------------------------------------------------------------------
+
+##-----------------------------------------------------------------------------#
+##' Change treatments to sequential integers
+##'
+##' Assigns all control treatments the same value, then assigns unique treatments
+##' within each site. Each site is required to have a control treatment.
+##' The algorithm (incorrectly) assumes that each site has a unique set of experimental
```
dlebauer:
```suggestion
##' The algorithm assumes that each site has a unique set of experimental
```

base/db/R/assign.treatments.R
```diff
@@ -0,0 +1,48 @@
...
+##' treatments.
```
dlebauer:
```suggestion
##' treatments. This assumption is required by the data in BETTdb that does not always consistently name treatments or quantity them in the managements table. Also it avoids having the need to estimate treatment by site interactions in the meta analysis model. This model is  uses data in the control treatment to estimate model parameters so the impact of the assumption is minimal.
```

Could someone please check this? It is late and I'd appreciate another set of eyes. I just cringed at the previous version that I probably wrote long ago.

base/db/R/fetch.stats2se.R
```diff
@@ -0,0 +1,25 @@
...
+##' to SE
+##' @name fetch.stats2se
+##' @title Fetch data and transform stats to SE
+##' @param connection connection to trait database
+##' @param query to send to databse
+##' @return dataframe with trait data
+##' @seealso used in \code{\link{query.trait.data}}; \code{\link{transformstats}} performs transformation calculations
+##' @author <unknown>
+fetch.stats2se <- function(connection, query){
+  transformed <- PEcAn.utils::transformstats(db.query(query = query, con = connection))
```
dlebauer: is this function that wraps another function needed?

bcow: No actually it isn't but I didn't want to delete any functions in this PR. I was going to leave that for a separate PR in which I actually make edits.

base/db/R/assign.treatments.R
bcow: I fixed a typo in your suggestion and added it to the commit - it looks ok to me and we can always go back and change it when reviewing the functions in further detail. Again, this PR was really only to reorganize.

base/db/R/query.trait.data.R
bcow: It was an accidental omission, but I searched the pecan code and I don't think it's used anywhere, so I think I may just leave it out.

bcow:
@robkooper I think you need to approve that I made those changes to make it possible to merge the PR

PR #2420, opened 2019-09-17
Use local file for read_web_config test. Fixes #2417

### LINKED ISSUE #2417: Revisit here::here in PEcAn.utils tests
`base/utils/tests/testthat/test.read_web_config.R` uses `here::here` to read two config files from the `web/` directory. This causes two issues, one trivial and the other needing more thought:

* [Need to declare](https://travis-ci.org/infotroph/pecan/jobs/580354445#L6126) the `here` import in the `PEcAn.utils` DESCRIPTION, probably in Suggests since it's used only during testing and already checks whether `here` is available.

* The test assumes it can find `web/` two levels up, so it fails with "cannot open the connection" when run during package checks (which are run in a tempdir, not in-place like devtools::test does). One possible fix would be to copy a sample config file into `base/utils/tests/testthat/data`, but there may be better solutions.

infotroph:
## Description

As discussed in #2417:
* Add a sample config.php to `base/utils/tests/testthat/data/` instead of trying to read it from outside the package dir
* Remove `here::here` in favor of plain `file.path`
* Add a few more tests while I'm at it

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


ashiklom:
If you feel like it, changing the `expand` block in the function to this should fix the parsing bug:

```r
  if (expand) {
    # Replace $output_folder with its value, and concatenate strings
    chr <- vapply(config_list, is.character, logical(1))
    config_list[chr] <- lapply(config_list[chr], gsub,
                               pattern = "\\$output_folder *\\. *",
                               replacement = config_list[["output_folder"]])
  }
```

Basically, the `gsub` call on every element was re-converting everything back to `character`.

Issue #2424, opened 2019-09-19
CRUNCEP THREDDS issue


para2x:
### Bug Description
I was hoping someone outside BU could run this and let me know if it successfully downloads files.
```
PEcAn.data.atmosphere::download.CRUNCEP(
  site_id = 1000005129, # This is not used inside the function. Lat/long is used
  lat.in = 46.0827,
  lon.in = -89.9792,
  model = NULL,
  scenario = NULL,
  ensemble_member = NULL,
  method = NULL,
  overwrite = FALSE,
  outfolder = '/fs/data1/pecan.data/dbfiles/CRUNCEP_site_1-5129/',
  start_date = '2005-01-01',
  end_date = '2010-12-31',
  verbose=TRUE
)


```
Right now, this returns this error here at both test-pecan and pecan2 machine. 
```
Error in ncvar_get_inner(ncid2use, varid2use, nc$var[[li]]$missval, addOffset,  : 
  C function R_nc4_get_vara_double returned error
```

ashiklom:
I tried this and got the same error. In my experience, OpenDAP can be pretty finnicky. With `method = "ncss"`, this seems to work OK, so give that a shot.

It may be worth considering defaulting to that instead, since it seems like a more robust download method.

para2x:
Thank you @ashiklom

infotroph:
I can confirm the same behavior using ncdf4 functions directly:

```
> nc <- ncdf4::nc_open("https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1220/mstmip_driver_global_hd_climate_tair_2005_v1.nc4", verbose=TRUE)
> ncdf4::ncvar_get(nc, varid="tair", start=c(181, 88, 1), c(1, 1, 1460))
Error in Rsx_nc4_get_vara_double: NetCDF: Access failure
Var: tair  Ndims: 3   Start: 0,87,180 Count: 1460,1,1
Error in ncvar_get_inner(ncid2use, varid2use, nc$var[[li]]$missval, addOffset,  : 
  C function R_nc4_get_vara_double returned error
```

serbinsh:
On my mac its slightly different 

```
PEcAn.data.atmosphere::download.CRUNCEP(
  site_id = 1000005129, # This is not used inside the function. Lat/long is used
  lat.in = 46.0827,
  lon.in = -89.9792,
  model = NULL,
  scenario = NULL,
  ensemble_member = NULL,
  method = "ncss",
[...]

PR #2426, opened 2019-09-19
Changing the default method for download.CRUNCEP

para2x:
## Description

Now that @ashiklom, @infotroph and I all confirmed (issue #2424 ) that the `opendap` method as a default breaks the CRUNCEP.download, falling back to  `ncss` method seems to solve the problem for both me and @ashiklom .

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
@para2x why does this change config.example.php?

para2x:
@infotroph just fixed it

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -24,9 +24,9 @@
 ##' @author James Simkins, Mike Dietze, Alexey Shiklomanov
 download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, lon.in,
                              overwrite = FALSE, verbose = FALSE, maxErrors = 10, sleep = 2,
-                             method = "opendap", ...) {
+                             method = "ncss", ...) {
 
-  if (is.null(method)) method <- "opendap"
+  if (is.null(method)) method <- "ncss"
```
infotroph: Given the default value in the function signature, couldn't this line be removed? Or do we need to support calls that explicitly set method=NULL for some reason?

para2x: I'm not really the developer of this function but my best guess is that it needs to be able to support that because `convert.input.R` function creates a generic set of argument that is used for all the `*met.product*.download.R` functions.

infotroph: That's fair. Probably not worth the upstream effort to change that, then.

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -220,7 +220,7 @@ download.CRUNCEP <- function(outfolder, start_date, end_date, site_id, lat.in, l
                                     "but got", min(dap_time), "..", max(dap_time))
       }
 
-
+
```
ashiklom: Unnecessary extra whitespace.

```suggestion

```

ashiklom:
> it makes me wonder whether switching to ncss is setting us up for trouble later

To me, the NetCDF subset (what I called "ncss") seems more reliable because it uses only a single REST-like query per download, rather than the more complicated back-and-forth that OpenDAP seems to do. OpenDAP is nicer for interactive use because it lets you explore the file as if it were local, but I wonder if it's not as robust to the large, high-frequency throughput we often use it for.

PR #2432, opened 2019-09-23
ED2: Add minimum_height to convert.samples

ashiklom:
Currently, the trait `minimum_height` will silently go unused by ED2. This converts the BETY trait `minimum_height` to the ED2 variable `repro_min_h`.

https://github.com/edmodel/ed2/blob/master/ED/src/io/ed_xml_config.f90#L663


ashiklom:
I could have sworn this came from somewhere, but I cannot trace it back. 🤷‍♂

Issue #2399, opened 2019-08-30
Execute batch runs in parallel


ashiklom:
probably fine for a first pass, but in general I don't think we should be limited to doing these runs sequentially

_Originally posted by @mdietze in https://github.com/PecanProject/pecan/pull/2089_

PR #2433, opened 2019-09-24
Adding new features to batch_run

para2x:
In this PR I'll be adding new features and fix minor bugs:
New features added so far : 

- Parallel execution of test jobs
- Handling multiple pfts (separated by semicolon)
- More robust approach to tests after run

I'll be also taking notes on the circumstances where models can break and this PR is NOT ready to handle:

1. LINKAGES needs the sites to have sand/clay registered in their bety. It's used in the write.config
2. `lat` and `lon` of a site needs to fall into a table called `clat.csv` in the `linkages` package otherwise it breaks.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/workflow/inst/batch_run.R
```diff
@@ -45,122 +45,119 @@ pecan_path <- get_arg(argv, "--pecandir", getwd())
...
+    ##################################################
+    # Did the workflow finish?
+    ##################################################
+    if (file.exists(file.path(outdir, "workflow.Rout"))) {
+      raw_output <- readLines(file.path(outdir, "workflow.Rout"))
+      result_table$workflow_complete <- any(grepl(
+        "PEcAn Workflow Complete",
+        raw_output
+      ))
+    }else{
```
ashiklom:
```suggestion
    } else {
```

base/workflow/inst/batch_run.R
```diff
@@ -45,122 +45,119 @@ pecan_path <- get_arg(argv, "--pecandir", getwd())
...
     proc_out <- list.files(out, "[[:digit:]]{4}\\.nc", recursive = TRUE)
-    if (length(proc_out) > 0) {
-      result_table$model_output_processed[[i]] <- TRUE
-      continue <- TRUE
-    }
+    result_table$model_output_processed <-ifelse(length(proc_out) > 0, TRUE, FALSE)
+    
+    return(result_table %>% 
+             as.data.frame() %>%
+             mutate(site_id=strsplit(outdir,"_")[[1]][5])
```
ashiklom: This seems fragile, since both output directories and site names themselves are _very_ likely to have underscores.

Why not just append to the original result table, like the original version of this code?

base/workflow/inst/batch_run.R
```diff
@@ -45,122 +45,119 @@ pecan_path <- get_arg(argv, "--pecandir", getwd())
...
-    continue <- FALSE
-    raw_out <- list.files(out, recursive = TRUE)
-    if (length(raw_out) > 0) {
-      result_table$model_output_raw[[i]] <- TRUE
-      continue <- TRUE
+#----------- Checking the results of the runs
+checks_df<-list.dirs(output_folder, full.names = TRUE, recursive = FALSE) %>% 
+  purrr::map_dfr(function(outdir){
+    
+    result_table <-NULL
```
ashiklom: Per my comment below, the original idea here was that the result table would have all of the information for the run, plus the additional columns indicating run status. Why drop it here?

base/workflow/R/create_execute_test_xml.R
```diff
@@ -104,20 +105,19 @@ create_execute_test_xml <- function(model_id,
...
-  }
-  settings$pfts <- list(
-    pft = list(name = pft,
-               constants = list(num = 1))
-  )
+
+  ## Putting multiple PFTs separated by semicolon
+  settings$pfts <- strsplit(pft, ";")[[1]] %>%
+    purrr::map( ~ list(name = .x,
+                       constants = list(num = 1)
```
ashiklom: Do we actually need this? I skimmed through `grep` output on our code and didn't see this as being essential anywhere, but I may be wrong. If the code works without it, I strongly suggest we drop it.

If it does turn out to be necessary, then at the _very_ least, if we do need it, we should make it increase sequentially by PFT; i.e.

```r
... %>% purrr::imap(~list(name = .x, constants = list(num = .y))
```

para2x: `Do we actually need this?` 
By this you mean the support for multiple PFTs ?

ashiklom: No, I mean `constants = list(num = 1)`.

base/workflow/inst/batch_run.R
para2x: The fundamental change here for me was to break the main sequential `for` loop into two separate loops, one parallelized for sending out all the jobs and then when they are all done another one for checking the outputs. 
Yeah this piece could be fragile, I'll find a better way.

ashiklom: Since you should be able to guess the output directory ahead of time, I would add it as a column to the input table and then `left_join(input_table, result_table, by = "outdir")`. That way, you preserve all of the metadata from the input table.

base/workflow/inst/batch_run.R
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env Rscript
-
-library(dplyr)
+library(tidyverse)
```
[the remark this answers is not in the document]
infotroph: The reason for the note is that you can't *import* from tidyverse into a package namespace (because it doesn't export anything!), not that there's anything wrong with *loading* it in an `inst` script. That said, there's a separate case to be made that if this script only uses a few of the Tidyverse packages then it's good to leave the others unloaded for namespace hygiene purposes. And if @ashiklom makes that case I won't disagree with it ;)

base/workflow/inst/batch_run.R
```diff
@@ -1,12 +1,12 @@
...
 library(PEcAn.workflow)
 stopifnot(
   requireNamespace("PEcAn.DB", quietly = TRUE),
   requireNamespace("PEcAn.utils", quietly = TRUE)
 )
-
+library(furrr)
```
infotroph: * @para2x: Consider moving this up so that all `library` calls are in one place.
* @ashiklom: I'm fuzzy on the best practice here: Does calling furrr in `inst` require adding it to Suggests?
...And while I'm looking at it, why the stopifnot for DB and utils above? They're both in workflow's Imports, so I'd be comfortable assuming they're available at call time.

infotroph:
Will this close #2399?

para2x:
> Will this close #2399?

Sorry I'm late to this. Yes it does.

mdietze:
@para2x with the semester ending, I was hoping you could let me know which of your two remaining PR you want to tackle first. Are they both truly still WIP or are they basically functional and just need to get through the build check?

para2x:
@mdietze This is not definitely WIP, as I stopped working on this PR almost a year. I'm going to go ahead and remove the WIP tag to see if it passes the checks. If it doesn't maybe someone in the lab can double check to see if it's easy to help push this across the line.

mdietze:
\document

mdietze:
\document

mdietze (approved):

PR #2435, opened 2019-09-26
Update write.configs.ed.R

bcow:
Fixing reading in of custom tags under the <ed2in_tags> flag in `pecan.xml`


infotroph:
@bcow the current Travis failure is arguably a bug in my existing-tests script -- you removed a warning not added it, but it wants an exact string matches against the whole list of undefined variables. This will likely be a problem for a lot of people and I'll need to think about how to make the check more granular.

But to fix for this PR, edit `models/ed/tests/Rcheck_reference.log` by deleting line 94 (`write.config.ED2: no visible binding for global variable ‘x’
`) and removing the ` x` (including space!) from the end of line 105 (`  soil unzip value_type var.names write.table x`).

PR #2436, opened 2019-09-26
Basgra coupling 

istfer:
## Description

First pass at coupling [BASGRA](https://github.com/MarcelVanOijen/BASGRA_N) to PEcAn.

Still in progress. Next steps: 
- [X]  I will get rid of the met2model function. 
- [X]  (Also later) I'll populate few more priors for the timothy grass and make additions to write.configs accordingly.
- [X] (Also later) Revise documentation.
- [ ]  (Not this PR) I'll add ability to manipulate N deposition and harvest dates through workflow.
- [ ]  (Not this PR) Start/end simulations at arbitrary dates 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

- [X] New model

## Checklist:


istfer:
@ashiklom tried to borrow your RTM approach here, but I have two questions: 
1) I now manually put the model code under src, but the model code is on [github](https://github.com/MarcelVanOijen/BASGRA_N/tree/1.0), should I tell PEcAn to pull it from there
2) When you compile the model in the model package, how do you tell PEcAn when there is a different model version?

istfer:
There is more to do for the coupling but maybe as an initial pass (also for the sake of keeping PRs small) I think this one is ready for a review.

ashiklom:
> I now manually put the model code under src, but the model code is on github, should I tell PEcAn to pull it from there? If yes, where?

I'm a bit wary of any sort of automated system for pulling code from GitHub -- that sounds like it's just asking for bugs. In theory, this is a good candidate for a [git submodule](https://git-scm.com/book/en/v1/Git-Tools-Submodules), but I find them pretty complex and often counterintuitive to use. My recommendation would probably be to use a [git subtree merge](https://git-scm.com/book/en/v1/Git-Tools-Subtree-Merging) and then periodically do manual updates (making sure to use `--squash` and `--no-commit` so we don't mess with PEcAn's own git history).

> When you compile the model in the model package, how do you tell PEcAn when there is a different model version?

If you follow the git subtree approach above, you can do this step manually at the same time. My suggestion might even be to write a brief script like `update-basgra.sh` that automatically does the subtree merge (with all the right arguments) and also updates the model version wherever it needs to be updated.

models/basgra/DESCRIPTION
```diff
@@ -0,0 +1,22 @@
+Package: PEcAn.BASGRA
+Type: Package
+Title: PEcAn package for integration of the ModelName model
```
ashiklom:
```suggestion
Title: PEcAn package for integration of the BASGRA model
```

models/basgra/DESCRIPTION
```diff
@@ -0,0 +1,22 @@
...
+Authors@R: c(person("Istem", "Fer"))
+Author: Istem Fer
+Maintainer: Istem Fer <istem.fer@fmi.fi>
+Description: This module provides functions to link the BASGRA to PEcAn.
+Imports:
+    PEcAn.logger,
+    PEcAn.utils (>= 1.4.8)
+Suggests:
+    testthat (>= 1.0.2)
+SystemRequirements: ModelName
```
ashiklom: Probably delete this?
```suggestion
```

models/basgra/DESCRIPTION
```diff
@@ -0,0 +1,22 @@
+Package: PEcAn.BASGRA
+Type: Package
+Title: PEcAn package for integration of the ModelName model
+Version: 1.7.1
+Date: 2019-09-05
+Authors@R: c(person("Istem", "Fer"))
+Author: Istem Fer
+Maintainer: Istem Fer <istem.fer@fmi.fi>
+Description: This module provides functions to link the BASGRA to PEcAn.
```
ashiklom:
```suggestion
Description: This module provides functions to link the BASGRA model to PEcAn.
```

models/basgra/DESCRIPTION
```diff
@@ -0,0 +1,22 @@
+Package: PEcAn.BASGRA
+Type: Package
+Title: PEcAn package for integration of the ModelName model
+Version: 1.7.1
+Date: 2019-09-05
+Authors@R: c(person("Istem", "Fer"))
```
ashiklom:
```suggestion
Authors@R: c(person("Istem", "Fer", email = "istem.fer@fmi.fi", role = c("aut", "cre"))
```

Then, you can drop the `Author:` and `Maintainer` lines below.

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+  ################################################################################
+  ### FUNCTIONS FOR READING WEATHER DATA
+  mini_met2model_BASGRA <- function(file_path,
+                                    start_date, start_year,
+                                    end_date, end_year) {
+
+    out.list <- list()
+    
+    ctr <- 1
+    for(year in start_year:end_year) {
```
ashiklom: More robust to weird sequences.

```suggestion
    for(year in seq(start_year, end_year)) {
```

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+        Tair_C <- udunits2::ud.convert(Tair, "K", "degC")
+        
+        t_dmean <- tapply(Tair_C, ind, mean, na.rm = TRUE) # maybe round these numbers 
+        matrix_weather[ ,4] <- t_dmean # mean temperature (degrees Celsius)
+        matrix_weather[ ,5] <- t_dmean # that's what they had in read_weather_Bioforsk
+        
+        RH <-ncdf4::ncvar_get(nc, "relative_humidity")  # %
+        RH <- tapply(RH, ind, mean, na.rm = TRUE) 
+        
+        matrix_weather[ ,6] <- exp(17.27*t_dmean/(t_dmean+239)) * 0.6108 * RH / 100
```
ashiklom: Can you add a brief comment explaining what this calculation is? It looks vaguely like a conversion to specific humidity, but I'm not sure.

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+    "(g g-1 DM)"   , "(g g-1 DM)"   , "(g g-1 DM)"    , "(g g-1 DM)"  ,              # 77:80
+    "(-)"          , "(-)"          ,                                                # 81:82
+    "(-)"          , "(-)"          , "(-)"           ,                              # 83:85
+    "(d-1)"        , "(mm d-1)"     , "(g N m-2 d-1)" , "(g N m-2 d-1)",             # 86:89
+    "(g N m-2 d-1)", "(g N m-2 d-1)", "(g N m-2 d-1)" , "(g N m-2 d-1)",             # 90:93
+    "(g N m-2 d-1)", "(g N m-2 d-1)",                                                # 94:95
+    "(g N m-2)"    , "(g N g-1 C)"  ,                                                # 96:97
+    "(g N m-2)"    , "(g N g-1 C)"  ,                                                # 98:99
+    "(d d-1)"                                                                        # 100
+  )
```
ashiklom: Not necessary for this PR, but personally, I tend to prefer code that pairs these together somehow. I'm a big fan of the` tibble::tribble` for writing inline `data.frame`s, but you could also do it with base R's `matrix(..., byrow = TRUE)`:

```r
output_vars <- matrix(c(
  "Time", "(y)",
  "year", "(y)",
  "doy", "(d)",
  ...), ncol = 2, byrow = TRUE)
```

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+  
+  NOUT <- as.integer( length(outputNames) )
+  
+  
+  ############################# SITE CONDITIONS  ########################
+  # this part corresponds to  initialise_BASGRA_***.R functions  
+  
+  year_start  <- as.integer(start_year)
+  doy_start   <- as.integer(lubridate::yday(start_date))
+  NDAYS       <- as.integer(sum(PEcAn.utils::days_in_year(start_year:end_year))) # could be partial years, change later
```
ashiklom:
```suggestion
  NDAYS       <- as.integer(sum(PEcAn.utils::days_in_year(seq(start_year, end_year)))) # could be partial years, change later
```

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+  ############################# WRITE OUTPUTS ###########################
+  # writing model outputs already in standard format
+  
+  # only LAI and CropYield for now
+  
+    
+  years <- seq(start_year, end_year)
+  for (y in years) {
+    
+    thisyear <- output[, which(outputNames == "year")] == y
```
ashiklom: FWIW, pretty sure you don't need the `which` here -- boolean subsetting should do exactly the same thing. No need to change -- just noting to save you typing in the future :wink:

models/basgra/R/run_BASGRA.R
istfer: I was following their calculation here https://github.com/MarcelVanOijen/BASGRA_N/blob/eafc6e36cf4df6ba04638b14a56d28616296fb97/initialisation/initialise_BASGRA_general.R#L40

Looking at [here](https://github.com/MarcelVanOijen/BASGRA_N/blob/eafc6e36cf4df6ba04638b14a56d28616296fb97/model/BASGRA.f90#L86) and `https://github.com/MarcelVanOijen/BASGRA_N/blob/eafc6e36cf4df6ba04638b14a56d28616296fb97/model/environment.f90#L49`, it is Vapor Pressure
```
VPI   = MATRIX_WEATHER(:,6)
...
VP     = VPI(day)   ! vapour pressure (kPa)
```

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+
+    # ******************** Declare netCDF variables ********************#
+    
+    ### Output netCDF data
+    nc <- ncdf4::nc_create(file.path(outdir, paste(y, "nc", sep = ".")), var)
+    varfile <- file(file.path(outdir, paste(y, "nc", "var", sep = ".")), "w")
+    for (i in seq_along(var)) {
+      # print(i)
+      ncdf4::ncvar_put(nc, var[[i]], outlist[[i]])
+      cat(paste(var[[i]]$name, var[[i]]$longname), file = varfile, sep = "\n")
```
ashiklom: Should this be some kind of `PEcAn.logger::logger.*` function?

models/basgra/R/run_BASGRA.R
```diff
@@ -0,0 +1,268 @@
...
+      matrix_weather[ ,1] <- rep(year, diy) # year
+      matrix_weather[ ,2] <- seq_len(diy) # day of year, simple implementation for now
+      
+      old.file <- file.path(dirname(file_path), paste(basename(file_path), year, "nc", sep = "."))
+      
+      if (file.exists(old.file)) {
+        
+        ## open netcdf
+        nc <- ncdf4::nc_open(old.file)  
+
```
ashiklom: This will save you having to close the file at the end of the function, and has the nice side effect of calling this if the function exits early for any reason, including throwing an error.

```suggestion
on.exit(ncdf4::nc_close(nc), add = TRUE)
```

models/basgra/R/write.config.BASGRA.R
```diff
@@ -0,0 +1,98 @@
...
+##-------------------------------------------------------------------------------------------------#
+write.config.BASGRA <- function(defaults, trait.values, settings, run.id) {
+
+  # find out where to write run/ouput
+  rundir <- file.path(settings$host$rundir, run.id)
+  outdir <- file.path(settings$host$outdir, run.id)
+  
+  # load default(!) BASGRA params
+  load(system.file("BASGRA_params.Rdata",package = "PEcAn.BASGRA"))
+  run_params <- default_params
```
ashiklom: Personally, I really don't like `.Rdata` files. But, you might consider using `PEcAn.utils::load_local` to instead read this into a list and avoid accidentally clobbering other objects in the current namespace (like `load` does by default).

```r
run_params <- PEcAn.utils::load_local(system.file(...))$default_params
```

models/basgra/README.md
```diff
@@ -0,0 +1,64 @@
+A generic template for adding a new model to PEcAn
```
ashiklom: For this PR, I would just remove this file altogether. Eventually, it would be nice to make a model-specific README with a brief explanation of what this is and links to additional documentation.

models/basgra/src/Makevars
```diff
@@ -0,0 +1,40 @@
...
+    
+
+.PHONY: all clean
+
+all : $(SHLIB) 
+
+$(SHLIB) : $(OBJECTS) 
+
+$(OBJECTS) :  $(SOURCES)
+	$(F77) -x f95-cpp-input -fPIC -O3 -c -fdefault-real-8 $(SOURCES)
```
ashiklom: I would probably write this as:

```
PKG_FFLAGS = -x f95-cpp-input -fdefault-real-8
$(OBJECTS) : $(SOURCES)
```

...and let R's default make macros work out the rest. Following [these recommendations](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-Makevars).

models/basgra/tests/testthat/README.txt
```diff
@@ -0,0 +1,3 @@
+Place your tests here. They will be executed in this folder, so you
+can place any data you need in this folder as well (or in a subfolder
+called data).
```
ashiklom: Delete this file?

models/basgra/DESCRIPTION
```diff
@@ -0,0 +1,21 @@
+Package: PEcAn.BASGRA
+Type: Package
+Title: PEcAn package for integration of the BASGRA model
+Version: 1.7.1
+Date: 2019-09-05
+Authors@R: c(person("Istem", "Fer", email = "istem.fer@fmi.fi", role = c("aut", "cre"))
+Author: Istem Fer
+Maintainer: Istem Fer <istem.fer@fmi.fi>
```
ashiklom: Now that all this info is in the `Authors@R`, I would drop these two lines.

models/basgra/R/run_BASGRA.R
ashiklom: Ah, OK. Thanks!

models/basgra/src/Makevars
istfer: this gives me `g++: error: *o: No such file or directory` kind of errors

models/basgra/tests/testthat.R
```diff
@@ -0,0 +1,13 @@
...
+# are made available under the terms of the 
+# University of Illinois/NCSA Open Source License
+# which accompanies this distribution, and is available at
+# http://opensource.ncsa.illinois.edu/license.html
+#-------------------------------------------------------------------------------
+library(testthat)
+library(PEcAn.utils)
+
+PEcAn.logger::logger.setQuitOnSevere(FALSE)
+#test_check("PEcAn.ModelName")
```
ashiklom: We probably _do_ want to run these tests, right? I would enable this now before we forget (especially since there are no actual tests that will be run inside this package right now).

```suggestion
test_check("PEcAn.BASGRA")
```

models/basgra/src/Makevars
ashiklom: Oh, I think that's probably because the compiled `.o` files in `OBJECTS` are placed in the same directory as their source files. So you need to prefix all of your `OBJECTS` with `model/`.

istfer: no, they're not being created at all in neither directories. For some reason it starts with g`++ -shared` and looks for `.o` files if I arrange commands like this. I think it should start with `gfortran -x f95-cpp-input -fPIC -O3 -c -fdefault-real-8 `, look for source files and create `.o`s first

istfer: this was the combination that worked for me
```
all : $(SHLIB) 

$(SHLIB) : $(SOURCES) 

PKG_FFLAGS = -x f95-cpp-input -fPIC -O3 -c -fdefault-real-8 $(SOURCES) 

$(OBJECTS) : $(SOURCES)
	$(F77) $(PKG_FFLAGS)
```

models/basgra/DESCRIPTION
```diff
@@ -0,0 +1,19 @@
+Package: PEcAn.BASGRA
+Type: Package
+Title: PEcAn package for integration of the BASGRA model
+Version: 1.7.1
+Date: 2019-09-05
+Authors@R: c(person("Istem", "Fer", email = "istem.fer@fmi.fi", role = c("aut", "cre")))
+Description: This module provides functions to link the BASGRA model to PEcAn.
+Imports:
+    PEcAn.logger,
+    PEcAn.utils (>= 1.4.8)
```
ashiklom: You should add other dependent packages here: `lubridate`, `ncdf4`, `udunits2`, and any others I may have missed.

models/basgra/inst/template.job
```diff
@@ -0,0 +1,39 @@
...
+mkdir -p "@OUTDIR@"
+
+# see if application needs running
+if [ ! -e "@OUTDIR@/results.csv" ]; then
+    
+  "@BINARY@"
+    # convert to MsTMIP
+  echo "require (PEcAn.BASGRA)
+run_BASGRA('@SITE_MET@', @RUN_PARAMS@, '@START_DATE@', '@END_DATE@', '@OUTDIR@', @SITE_LAT@, @SITE_LON@)
+" | R --vanilla
```
ashiklom: Not strictly necessary for this PR, but I wonder if this would work as:

```sh
Rscript \
  -e "library(PEcAn.BASGRA)" \
  -e "run_BASGRA('@SITE_MET', @RUN_PARAMS, ...)"
```

`R --vanilla` ignores the `.Renviron` and `.Rprofile` files, which makes it difficult to configure default R library locations, etc.

models/basgra/inst/template.job
```diff
@@ -0,0 +1,39 @@
...
+# host specific setup
+@HOST_SETUP@
+
+# create output folder
+mkdir -p "@OUTDIR@"
+
+# see if application needs running
+if [ ! -e "@OUTDIR@/results.csv" ]; then
+    
+  "@BINARY@"
```
ashiklom: I think you call the model below, so this can be removed? Or am I missing something?

```suggestion
```

models/basgra/src/Makevars
ashiklom: Huh, interesting. I thought R was smart enough to figure out how to do this without the fully explicit rule, but I guess not. If it works, let's leave it as is, and potentially revisit if it causes anyone problems.

models/basgra/inst/template.job
ashiklom: I _would_ strongly consider changing `require` to `library` though -- you do actually want the script to throw an informative error here if `PEcAn.BASGRA` isn't installed.

PR #2443, opened 2019-09-30
Fix for get.parameter.samples 

para2x:
## Description

I think, @femeunier and I found a change in behavior of base R between 3.4.X and 3.5.X which may result in a bug in near future for pecan running under R 3.5.X .  At line where I'm changing `trait.mcmc[[prior]]` is a list with 4 elements (4 chains for each parameter used in MA or HPDA) which when it's evaluated in R 3.4.X, it combines all lists into one vector. But in R 3.5.X code breaks at this point and it requires a more specific pointer to what elements within this list needs to be taken. I tested this fix for both versions of the R and 2 models and it passed successfully at all of them. 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/uncertainty/R/get.parameter.samples.R
+++ b/modules/uncertainty/R/get.parameter.samples.R
@@ -131,7 +131,7 @@ get.parameter.samples <- function(settings,
     PEcAn.logger::logger.info("using ", samples.num, "samples per trait")
     for (prior in priors) {
       if (prior %in% param.names[[i]]) {
-        samples <- as.matrix(trait.mcmc[[prior]][, "beta.o"])
+        samples <- trait.mcmc[[prior]] %>% purrr::map(~ .x[,'beta.o']) %>% unlist() %>% as.matrix()
       } else {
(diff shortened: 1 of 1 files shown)


infotroph:
If the patch works it works, but I don't see how basic subsetting like this would have changed between R versions: We already use R 3.5 in Dockerized PEcAn and run Travis tests against R 3.5, 3.6, and devel (what will eventually be R 3.7), so if our tests missed a behavior change for this long this we should be very worried.

Can you give small example that reproducibly works on 3.4 and fails on 3.5?

infotroph:
Also the `as.matrix(trait.mcmc[[trait]][, "beta.o"])` construct gets used elsewhere in PEcAn (I see it at least in `modules/meta.analysis/R/run.meta.analysis.R` and `models/ed/scripts/write.configs.rscript.R`), so whatever fix is needed here will probably be needed elsewhere too

PR #2445, opened 2019-10-02
[Docs] Fix chunk rendering in dependencies appendix

infotroph:
## Description

Fixed two code chunks that Rmarkdown was evaluating instead of displaying

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


mdietze:
I've never seen the \```r syntax, how is that different from just a generic code block \``` ?

infotroph:
@mdietze Loosely, ` ```r ` is a generic Markdown convention for "This is some preformatted text, please render it in a fixed-width font and apply syntax highlighting that assumes it's R code", while ` ```{r [options]} ` is Rmarkdown notation for "This is a chunk of R code, please interpret it and display the results in a way further specified by [options]"

mdietze:
So it sounds like  \``` is a non-evaluated code block, and \```r is a non-evaluated code block that also has R syntax highlighting?  Good to know

infotroph:
@mdietze Yep, correct. The quality of the highlighting you get will vary wildly between platforms 🙈 , but it does often come in handy

PR #2450, opened 2019-10-04
WIP: THREDDS download function

bailsofhay:
This function is the start of a general function that should be able to download data from THREDDS.
## Description
The user can download any THREDDS datasets as long as they have site information, dates, and the base URLS needed to grab the datasets. Base URLS need the dataset name, but not the dates or locations. There will likely be issues pulling data sets in the future since they are not all stored in similarly structured directory formats like AVHRR LAI (which this function is based on). Updates to the function will occur as we start playing with more datasets from THREDDS.

## Motivation and Context
We need an easy way to grab more datasets and THREDDS has many already, and is a great place to store future datasets that we can use. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


bailsofhay:
@ankurdesai: You are 100% correct on needing to make this as general as possible. This issue is part of why I haven't spent too much time making it general since the file directories in the catalog are all different from each other. Someone may have a good idea as a solution to this. 

Also, yes you're correct about the other files getting added to this PR..... I'm not sure how that happened since I this function is on a total different branch than my SDA changes.......... I'll have to look into this to see how I accidentally got those files in here.

Issue #2415, opened 2019-09-12
Interim PEcAn.data.remote::download.thredds.AGB failing


serbinsh:
### Bug Description

when running example in PEcAn.data.remote::download.thredds.AGB
```
 outdir <- "~/scratch/abg_data/"
results <- PEcAn.data.remote::download.thredds.AGB(outdir=outdir, site_ids = c(676, 678, 679, 755, 767, 1000000030, 1000000145, 100Loading required package: doParallels = 8)
```

Im getting 

```
2019-09-12 10:42:38 INFO   [PEcAn.data.remote::download.thredds.AGB] :
   Running in parallel with: 8
Error in get_data(i) : task 1 failed - "cannot open the connection"
```

Trying to ping the data: https://thredds.daac.ornl.gov/thredds/dodsC/ornldaac/1221/agb_5k.nc4  I get

```
Error {
    code = 400;
    message = "Unrecognized request";
};
```

### To Reproduce

```
 outdir <- "~/scratch/abg_data/"
results <- PEcAn.data.remote::download.thredds.AGB(outdir=outdir, site_ids = c(676, 678, 679, 755, 767, 1000000030, 1000000145, 100Loading required package: doParallels = 8)
```

## Expected behavior

Download the data

serbinsh:
@bailsofhay maybe the URLs have changed, for example:  https://thredds.daac.ornl.gov/thredds/catalog/ornldaac/1221/catalog.html?dataset=1221/agb_5k.nc4

May need to update URLs through this site: https://thredds.daac.ornl.gov/thredds/catalog.html

serbinsh:
From @para2x "Seems like the address has not changed and the reason you can't look up the address is that it's associated with opendap protocol not http server. So if you go ncdf4::nc_open you should be able to open the file in that address. But in the case of CRUNCEP something probably inside ncdf files has changed bc this ncdf4::ncvar_get breaks. I'm trying to figure out what happened. "

PR #2458, opened 2019-10-15
WIP: Thredds download function 

bailsofhay:
This function is the start of a general function that should be able to download data from THREDDS.

## Description
The user can download any THREDDS datasets as long as they have site information, dates, and the base URLS needed to grab the datasets. Base URLS need the dataset name, but not the dates or locations. There will likely be issues pulling data sets in the future since they are not all stored in similarly structured directory formats like AVHRR LAI (which this function is based on). Updates to the function will occur as we start playing with more datasets from THREDDS.

## Motivation and Context
We need an easy way to grab more datasets and THREDDS has many already, and is a great place to store future datasets that we can use.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


bailsofhay:
@ankurdesai I had to remove the earlier PR I made last week with the SDA file changes attached. this is the new PR for the same Thredds function, but without non-related file changes.

bailsofhay:
Turns out there was already a branch named thredds that was messing up my pr....... :/

modules/data.remote/R/download.thredds.R
```diff
@@ -20,84 +30,124 @@
 ##' @export
 ##' @author Bailey Morrison
 ##'
-download.thredds.AGB <- function(outdir = NULL, site_ids, run_parallel = FALSE, 
-                                    ncores = NULL) {
+download.thredds.data <- function(outdir = NULL, site_info, dates = c("19950201", "19961215"),
```
infotroph: I recommend *not* providing a default value for `dates`, so that it fails immediately and clearly when users don't provide them
```suggestion
download.thredds.data <- function(outdir = NULL, site_info, dates, 
```

modules/data.remote/R/download.thredds.R
```diff
@@ -20,84 +30,124 @@
...
+  
+  # check that dates are within the date range of the dataset
+  dates = c(as.Date(dates[1], "%Y%m%d"), as.Date(dates[2], "%Y%m%d"))
+  if (!(is.null(dir_url)))
+  {
+    #https://www.ncei.noaa.gov/thredds/catalog/cdr/lai/files/1981/catalog.html -> link for directory files, not downloads
+    result <- RCurl::getURL(paste(dir_url, "catalog.html", sep = "/"), verbose=F,ftp.use.epsv=TRUE, dirlistonly = TRUE)
+    files = XML::getHTMLLinks(result)
+    
+    date_year_range = unique(range(c(year(as.Date(dates[1], "%Y")), year(as.Date(dates[2], "%Y")))))
```
infotroph: `as.Date` on a Date object and `range` on two values are both no-ops, so this can be simplified to:
```suggestion
    date_year_range = unique(lubridate::year(dates))
```

infotroph (changes requested):
I see that this is still WIP so I hope my review isn't premature, but a few interface-level comments that are probably easier to address sooner rather than later. 

Might also be useful to add a fully updated example that can double as a test case while you're working on the details.

modules/data.remote/R/download.thredds.R
infotroph: And while I'm thinking about it, consider moving `outdir` to come after  later in the argument list -- it's best to arrange arguments such that everything without a default value comes before everything with a default value, even if the default is NULL.

modules/data.remote/R/download.thredds.R
```diff
@@ -20,84 +30,124 @@
...
-    site = site_ID[i]
-    output = as.data.frame(cbind(d, sd, date, site))
-    names(output) = c("value", "sd", "date", "siteID")
+    #https://www.ncei.noaa.gov/thredds/dodsC/cdr/lai/files/1981/AVHRR-Land_v005_AVH15C1_NOAA-07_19810624_c20181025194251.nc.html
+    # this is what a link looks like to download threeds data.
+    urls = sort(paste(data_url, substr(dates_avail, 1, 4), filenames, sep = "/"))
     
-    # option to save output dataset to directory for user.
-    if (!(is.null(outdir)))
+    extract_nc = function(site_info, url, run_parallel)
```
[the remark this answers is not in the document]
mdietze: I'll also add that there's already an extract.nc function in data.atmosphere, which is going to be very confusing. Furthermore, after we clean up the function naming (i.e. finally switch . to _ ) it could easily lead to genuine conflicts. One option is to rename this function, since it's the new one, and because the args makes it clear that this function isn't just extracting it's also downloading. Alternatively, you could take a look at the other function and think about whether you could generalize it (without breaking it's current I?O and behaviors) to serve both cases.

Either way, the function definitely needs Roxygen documentation, especially since site_info is a list of stuff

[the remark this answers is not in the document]
infotroph: Good point, @mdietze -- renaming or generalizing seems wise. But do note that this is a helper that isn't exported and probably doesn't need to be, which (1) mitigates the name conflict issues (though not the potential for confusion) and (2) personally I think it makes Roxygen docs less important -- to me it's clear in context that site_info will be the same thing here as it is in download_thredds_data.

infotroph:
Once merged this will close #2415, correct?

modules/data.remote/R/download.thredds.R
```diff
@@ -1,17 +1,27 @@
 #
-##' @title download.thredds.AGB
-##' @name  download.thredds.AGB
+##' @title download.thredds.data
+##' @name  download.thredds.data
 ##' 
-##' @param outdir Where to place output
-##' @param site_ids What locations to download data at? 
+##' @param outdir file location to place output
+##' @param site_info information about the site. i.e. site_id, latitude, longitude
```
mdietze: site_info is insufficiently documented. Are there arguments other than site_id, latitude, and longitude? Is site_id the one in BETY or something else? are lat/lon decimal degrees or something else? Is this supposed to be in a list, or a vector, or a data frame, or something else? Are multiple sites allowed

modules/data.remote/R/download.thredds.R
```diff
@@ -1,17 +1,27 @@
 #
-##' @title download.thredds.AGB
-##' @name  download.thredds.AGB
+##' @title download.thredds.data
+##' @name  download.thredds.data
```
mdietze: Is `.data` really needed at the end of this name?

bailsofhay: fixed

modules/data.remote/R/download.thredds.R
bailsofhay: ahhh. I see what you mean. I'll update this to be more informative

bailsofhay:
@infotroph, @ashiklom or @mdietze. What is the standard way to code in warnings. For example, I want a warning to pop up and kill data extraction if the user puts in a date range that the data doesn't exist during.

mdietze:
Warning, errors, and logs should be handled through the logger functions in the PEcAn.utils package

infotroph (approved):

PR #2459, opened 2019-10-22
SDA .nc file output update

kzarada:
## Description
For not removing the `nc` output files of model simulations in SDA, @para2x was renaming the outputs to their name + a timestamp. But there was a bug in the code that this PR is fixing.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ X] When possible
## Types of changes

## Checklist:


para2x:
This looks good to me and it's what we discussed. Did you also get to test it ?

kzarada:
@para2x yes, I ran it and it worked properly

modules/assim.sequential/R/sda.enkf_refactored.R
```diff
@@ -350,19 +350,24 @@ sda.enkf <- function(settings,
...
+      full.names = TRUE)
+   files <-  files[grep(pattern = "SDA*", files, invert = TRUE)]
+    
+    
+    purrr::walk(files,
       function(.x){
         file.rename(.x ,
                     file.path(dirname(.x),
-                              paste0(gsub(" ", "", as.character(names(obs.mean)[t])),
+                              paste0("SDA_",gsub(" ", "", as.character(names(obs.mean)[t])),
```
infotroph: Are all files in `files` guaranteed to live in different directories? If any of them share a `dirname`, they'll be given the same name and therefore overwrite each other. Or am I missing something?

Also, less importantly: `file.rename` accepts vectors, and `names()` is documented to return characters if it returns anything, so I *think* you can drop both `walk` and `as.character` and simplify this to

```
file.rename(
  files,
  file.path(
    dirname(files),
    paste0("SDA_", gsub(" ", "", names(obs.mean)[t]), ".nc"))
```

para2x: I haven't seen any of our models to produce more than one `nc` output file per dir, but that's a good point and we can slightly modify that to use the `basename` but then we need the `walk` again something like this:
`paste0("SDA_", basename(.x), "_", names(obs.mean)[t], ".nc")`
>

para2x: oh seems like the `basename` function is also vectorized and yeah seems like we don't the `walk` at all.

para2x: @kzarada could you please give a try to this replacing the whole `walk` see if it works as expected :
`
file.rename(
    files,
    file.path(
       dirname(files),
       paste0("SDA_", basename(files), "_", gsub(" ", "", names(obs.mean)[t]), ".nc")
     )
 )
`

kzarada: That code worked- the output files look like this 'SDA_2019.nc_2019-11-0500:00:00.nc'

kzarada: I pushed the update

kzarada:
@infotroph those were accidentally added (I forgot to create a new branch with this update since it was still WCr SDA related)- not relevant to this bug fix but they are part of the WCr SDA. I just added a shell file to run the forecast automatically, updated the workflow to work with the automation, and increased the ensemble size

mdietze:
@kzarada those accidentally added files all sound like things we want to pull in, so lets go ahead and keep them in this PR.

Issue #2187, opened 2018-11-06
error running sipnet


robkooper:
just tried to run a simple sipnet run, niwot, 2004, this runs all the way through, but does not show any results. Error in sipnet logfile.txt

```
>     model2netcdf.SIPNET('/home/kooper/output//PEcAn_99000000003/out/99000000001', 40.0329, -105.546, '2004/01/01', '2004/12/31', FALSE, '136')
2018-11-06 16:24:24 SEVERE [model2netcdf.SIPNET] : 
   Date range specified in model2netcdf.SIPNET() function call and total 
   number of years in SIPNET output are not equal 
Error in PEcAn.logger::logger.severe("Date range specified in model2netcdf.SIPNET() function call and \n                                total number of years in SIPNET output are not equal") : 
  Date range specified in model2netcdf.SIPNET() function call and 
                                total number of years in SIPNET output are not equal
Calls: model2netcdf.SIPNET -> <Anonymous>
Execution halted
```
This run can be seen at http://pecan-dev.ncsa.illinois.edu/pecan/08-finished.php?workflowid=99000000003

robkooper:
This is related to a commit #2095 
@serbinsh i was going to change it so it would check to see if years is inside simulation years (instead of other way around) and loop over years.

robkooper:
just did a run for niwot and sipnet that finished.

PR #2463, opened 2019-10-25
this fixes the extraction of only the year requested #2187

robkooper:
The code will now only extract the year that the user requested and not all the years that are in the metfile.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


robkooper:
@serbinsh please make sure I don't reintroduce the bug you fixed #2095

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -74,15 +74,11 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
+  # get all years that we want data from
   year_seq <- seq(lubridate::year(start_date), lubridate::year(end_date))
-  if (length(year_seq) != num_years) {
-    PEcAn.logger::logger.severe("Date range specified in model2netcdf.SIPNET() function call and 
-                                total number of years in SIPNET output are not equal")
-  }
 
   # check that specified years and output years match
-  if (!all(simulation_years %in% year_seq)) {
+  if (!all(year_seq %in% simulation_years)) {
```
infotroph: :bug: :hammer: :100:

serbinsh: This wasnt a bug it was a check based on some older issues with met process resulting in incorrect SIPNET years being run.

serbinsh: before this check (which was done awhile ago) SIPENT runs could complete even if the wrong years or dates were simulated and this was meant to throw an consistency error.  But that was based on the older met process code where we had a lot of errors from incorrect BETY entries, etc.  But given we have been cleaning these up, I think this change should maintain a similar behavior.

robkooper:
@serbinsh not sure if this will 100% work, but not sure the previous code was working in that case as well.

If I have met data from 2001-01-01 - 2015-07-31 and I request a run from 2015-04-01 - 2015-09-31 it would claim success, since the year 2015 is part of the actual run. It only checks for the years. I think if we want to do a full check it will need to check the output of the model generated.

I think a better place to catch this (not sure this is done) is to see if the met data covers the actual run data, this should be done I think in the met conversion (if it is not already done).

PR #2465, opened 2019-10-29
R check

robkooper:
From @mdietze :

All #rstats folks, pls run R with:

_R_CHECK_LENGTH_1_CONDITION_=verbose
_R_CHECK_LENGTH_1_LOGIC2_=verbose

set. I keep finding bugs in CRAN & ⁦‪@Bioconductor‬⁩ pkgs this way and some of them are serious => wrong scientific results. Report to maintainer when you spot these

This pull request enables these 2 options in verbose mode.

## Motivation and Context

These options will become the default, so we should fix any issues right now.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


.travis.yml
```diff
@@ -18,6 +18,8 @@ env:
     - MAKEFLAGS="-j2"
     - PGHOST=localhost
     - RGL_USE_NULL=TRUE # Keeps RGL from complaining it can't find X11
+    - _R_CHECK_LENGTH_1_CONDITION_=verbose
+    -_R_CHECK_LENGTH_1_LOGIC2_=verbose
```
infotroph: I'm fuzzy on how verbose mode works -- Do we also need to do something to explicitly turn it on during the build?

robkooper: ‘verbose’ prints very detailed diagnostic message
setting it to TRUE will give an error, default is warning.

So I think we are good, we might want to set it explicitly to true, to make sure, but I would like to get a vote on that.

See https://cran.r-project.org/doc/manuals/r-devel/R-ints.html

ashiklom: My vote would be to set this to `true`. This can introduce some really insidious bugs, and I've seen it in a lot of places in PEcAn code (especially relative newcomers to R).

robkooper: setting them to true, to see if anything fails.

infotroph: Aah OK, I was interpreting it as something like "behavior depends on the setting of options( "verbose"), but I see I was wrong. Yeah, I'll vote `true` as well.

robkooper:
So either I did this wrong, or there are no problems.

PR #2466, opened 2019-10-30
Remove num as a constant to convert in `write.configs.xml.ED`

bcow:
Small but important fix for `write.configs.xml.ED`:

I have been having been experiencing a bug where if I am doing an ED run **through the web interface (and thus purposefully am not manually setting `<ed2_pft_number> in the pecan.xml`)** my PFT is not correctly mapped to it's ED PFT even though it is correctly saved in the `pftmapping.csv`. I finally figured out why. 

### Short summary
The `num` value that is written in the `config.xml` is properly changed to the right PFT number and then changed back to the wrong number. 

### Long version:

Reading through `write.configs.xml.ED`
- The ED PFT number has been identified given the PFT specified in the pecan.xml. 
- By line 525, the list `vals` is defined depending on that PFT. 
- One of the key variables in the list is `num` which is used later on by `write.configs.ed`. 
- `num` is also defined in the pecan.xml under `pfts$constants$num`, but when `settings` are read in, this number does not appear to be already mapped to the correct ED PFT. I don't know where or why it is filled in, perhaps it would be better to leave it blank. 
- Because `num` is already defined, it is read in at the line `converted.defaults <- convert.samples.ED(defaults[[pft]]$constants)` but there is no code to "convert" `num` and so the incorrect `num` value is returned as `converted.defaults` and subsequently overwrites the correct `num` value at the step
        if (!is.null(converted.defaults)){
          vals <- modifyList(vals, converted.defaults)
        }
- `vals` is passed on to `write.configs.ed` with properly converted trait values and constants, but the incorrect PFT `num`.

### I propose

Simply removing `num` from the list of constants before passing them to be converted.

## Review Time Estimate
- [x ] Immediately

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/models/ed/R/write.configs.ed.R
+++ b/models/ed/R/write.configs.ed.R
@@ -531,7 +531,9 @@ write.config.xml.ED2 <- function(settings, trait.values, defaults = settings$con
         vals <- modifyList(vals, converted.trait.values)
 
         ## Convert settings constants to ED units
-        converted.defaults <- convert.samples.ED(defaults[[pft]]$constants)
+        constants.to.convert <- defaults[[pft]]$constants
+        constants.to.convert$num <- NULL
+        converted.defaults <- convert.samples.ED(constants.to.convert)
 
         ## Selectively replace defaults and trait values with constants from settings
         if (!is.null(converted.defaults)){


infotroph:
I'm rusty on the conventions here and could have this exactly backwards, but my impression is that is general we *want* settings from the pecan.xml to override ones calculated elsewhere, right? For example, I've sometimes used `settings$pfts$pft$constants$file` to temporarily pass BioCro a nonstandard set of defaults for a given genus, so by analogy I'd expect to be able to override settings$pft$constants$num if I  really wanted to and I'd prefer to fix the input file rather than override it.

Looks like the bogus pft$constants$num comes from the web run code [treating it as an index within *this run's* PFT list](https://github.com/PecanProject/pecan/blob/develop/web/04-runpecan.php#L299:L308). Would it be safe to skip writing those, or are there other models that interpret this differently and need the sequence ID?

infotroph:
Although! Arguing the other side now: I think as a general rule it's good to avoid storing highly model-specific parameters inside generic-named settings that could easily conflict between models, so maybe "ignore `pft$constants$num` and tell users to set `ed2_pft_number` if that's what they mean" would be good design anyway. How much existing code relies on reading ed2_pft_number from constants$num?

mdietze:
I'm going to side with @infotroph 1 on this over @infotroph 2. The pecan.xml needs to have the last say over any built-in defaults. Also, AFAIK, there are no "standard" or "generic" `pft$constants`, all of them are model specific. Since they're not generic, (1) they should also not require conversion and (2) I don't think the web interface should be providing a default value. From the web interface, I think the user should only be setting pft$constants$num via Advanced Edit

models/ed/R/write.configs.ed.R
```diff
@@ -531,7 +531,9 @@ write.config.xml.ED2 <- function(settings, trait.values, defaults = settings$con
         vals <- modifyList(vals, converted.trait.values)
 
         ## Convert settings constants to ED units
-        converted.defaults <- convert.samples.ED(defaults[[pft]]$constants)
+        constants.to.convert <- defaults[[pft]]$constants
+        constants.to.convert$num <- NULL
```
robkooper: what happens if the user specifies in pecan.xml a specific ED num to use? Would it be better if we don't write the num in the webinterface?

infotroph:
<a href="https://imgflip.com/i/3evnrm"></a>

mdietze:
This issue should have been resolved in #2469. Please re-open if that isn't the case

PR #2439, opened 2019-09-26
updated priors vignette so that it works, exported required functions

dlebauer:
Updated the prior vignette so that it can compile. This required exporting a few functions from the PEcAn.priors function. Or perhaps these were incorrectly identified as S3 methods.

## Motivation and Context

To get the vignette working again.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [x] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2471, opened 2019-11-01
Prior vignette cleanup

infotroph:
Hopefully sufficient to get #2439 building cleanly:

*  changelog
* fix easy check errors in renamed functions, add harder ones to cached checks list
* remove s3 inconsistency warning from cached check results
* change references elsewhere
* remove unneeded Roxygen tags, keep documentation aliases to old names


dlebauer:
Thanks for cleaning this up - great idea to retain old function names as aliases

PR #2477, opened 2019-11-07
[Build] Fix some package compilations, add scripts for Travis cache priming

infotroph:
## Description

This patch fixes failing Travis builds on R 3.5 and R devel:

* Switches libcurl from openssl flavor to gnuTLS flavor, needed to compile `redland` in the cases where prebuilt c2d4u binary isn't compatible. The Docker images and the PEcAn VM both used libcurl4-gnutls already, so I don't expect any issues.
* Expands the `scripts/travis/rebuild_pkg_binaries.R` from only rebuilding packages built using a *later* version of R (i.e. only on 3.5) to *always* rebuilding any R package that was built using a different version of R. Needed because `data.table` apparently now checks this at runtime and complains too late for us to fix it the usual ways.
* Unfortunately this means the R devel build now needs to recompile *every* package installed from c2d4u, and therefore times out when they're not in cache. Added some scripts to streamline the process of building up the Travis cache, which isn't very convenient but might be enough for now. I'll take "Tools whose very existence is a warning sign" for $500, Alex.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
OK, technically R devel still fails. But this gets us back from failing at the beginning of the run to failing at the end of the run (because of a documentation change imported from base R) after the rest of the checks pass.

PR #2481, opened 2019-11-15
CMS production run required changes

para2x:
## Description

Before I started my CMS production runs, I started making some changes that I needed for those runs. These changes are just mainly refinements.
 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/assim.sequential/R/Remote_helpers.R
```diff
@@ -235,29 +256,47 @@ SDA_remote_launcher <-function(settingPath,
                            scratchdir = ".")
   }
   
-
+  # Do the Rsync to copy all the main dir of the inputs
+  need.copy <- input.paths[!missing.inputs]
+  
+  need.copy.dirs <- dirname(need.copy) %>%
+    unique() %>%
+    discard(~ .x %in% c(".", "/fs/data1/pecan.data/dbfiles" ))
```
mdietze: You can't hard code paths like this

modules/assim.sequential/inst/RemoteLauncher/SDA_launcher.R
```diff
@@ -31,6 +34,20 @@ if (is.na(args[2])){
 # Setup
 #---------------------------------------------------------------
 setwd(settings$outdir)
+# This is how I delete large folders
+c('run', 'out', 'SDA') %>%
```
mdietze: I'm missing something, why are we deleting these folder?

modules/data.land/R/ic_process.R
```diff
@@ -50,6 +50,20 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
   con <- bety$con
   on.exit(db.close(con))
   
+  
+  # setup site database number, lat, lon and name and copy for format.vars if new input
+  new.site <- data.frame(id = as.numeric(site$id), 
+                         lat = PEcAn.data.atmosphere::db.site.lat.lon(site$id, con = con)$lat, 
+                         lon = PEcAn.data.atmosphere::db.site.lat.lon(site$id, con = con)$lon)
```
mdietze: Seems silly to query the database twice to get lat/lon when the function returns both

modules/assim.sequential/R/Remote_helpers.R
```diff
@@ -94,7 +94,7 @@ Obs.data.prepare.MultiSite <- function(obs.path, site.ids) {
 SDA_remote_launcher <-function(settingPath, 
                                ObsPath,
                                run.bash.args){
-
+  plan(multiprocess)
```
mdietze, line 97: Missing the namespace on this function call. Also, I'm not familiar with the `plan` function -- what is this step doing? And what is it acting on? `multiprocess` wasn't an argument to this function.

modules/assim.sequential/R/Remote_helpers.R
mdietze: willing to accept this path for now since it it seems to just be filtering some stuff out, but still agree it's not a great long-term solution

modules/assim.sequential/R/Remote_helpers.R
para2x: I used a combination of `plan` function form the `future` package  and the ` furrr::future_map` to parallelize some processes. `multiprocess` is also is a 'strategy' defined in the `future` package that automatically checks for the support for multicore to spread the jobs. If it is not supported, then it falls back to multi-session and finally sequential whichever possible to evaluate the jobs.

modules/assim.sequential/R/Remote_helpers.R
```diff
@@ -160,9 +161,29 @@ SDA_remote_launcher <-function(settingPath,
...
+    
+    # change the pft folder inside the setting
+    
+    settings$pfts %>%
+      purrr::map('outdir') %>%
+      walk(function(pft.dir) {
+        settings <<-
+          rapply(settings, function(x)
+            ifelse(
+              x == missing.input,
```
mdietze, line 173: Not clear where the value for `missing.input` is coming from at this point because it doesn't seem to get defined till later in this function.

modules/assim.sequential/R/Remote_helpers.R
```diff
@@ -160,9 +161,29 @@ SDA_remote_launcher <-function(settingPath,
...
+              file.path(settings$host$folder,
+                        folder_name, "pft", fdir, fname) ,
```
mdietze, line 175: Similarly, fdir and fname seem to be used before being defined here too.

modules/assim.sequential/R/Remote_helpers.R
```diff
@@ -235,29 +257,50 @@ SDA_remote_launcher <-function(settingPath,
                            scratchdir = ".")
   }
   
+  # Do the Rsync to copy all the main dir of the inputs
+  need.copy <- input.paths[!missing.inputs]
+  
+  need.copy.dirs <- dirname(need.copy) %>%
+    unique() %>%
+    discard(~ .x %in% c("."))
```
mdietze, line 265: I think `discard` needs a namespace

PR #2482, opened 2019-11-15
Add ED2 SAS script - Update {WIP}

para2x:
## Description

This is the copy (with updates) of PR #2009 which I'm taking over addressing the comments. I didn't know (didn't have the permission too) how to push my changes to this crollinson:ED2-SAS, if anybody knows a way to do so we can close this PR.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


para2x:
I'm planning to remove the soil related functions I just added after making sure what parameters they generate and to what parameters inside our own data.land `soil.param` function correspond.

models/ed/R/SAS.ED2.R
```diff
@@ -42,23 +19,180 @@
...
+##' @param treefall Value to be used for TREEFALL_DISTURBANCE_RATE in ED2IN for full runs (disturbance on)
+##' @param sufx ED2 out file suffix; used in constructing file names(default "g01.h5) 
+##' @export
+##'
+SAS.ED2 <- function(dir.analy, dir.histo, outdir, lat, lon, block,
+                    prefix, treefall, param.args = SAS.ED2.param.Args(), sufx =
+                      "g01.h5") {
+  
+  #--- Reading in the parameters
+  decomp_scheme <- param.args[1]
```
infotroph: This will break badly if anyone passes a non-default param.args. At a minimum this should match by name instead of index, but probably better would be to use all params directly from the vector, e.g. `if(!param.args$decomp_scheme %in% 0:4) stop("Invalid decomp_scheme")` and so on.

PR #2483, opened 2019-11-20
Code cleanup in Settings

infotroph:
## Description

I took a pass through PEcAn.settings fixing all the code style / formatting issues I could resolve easily, plus a few other tweaks that I saw in the process. In roughly decreasing order of prevalence:

* Deleted trailing whitespace, converted leading tabs to two spaces
* Normalized spacing around punctuation and parens
* wrapped lines to <= 80 chars where possible
* `=` -> `<-`
* T/F -> TRUE/FALSE
*  inside `if`: `|` -> `||` and `&` -> `&&`
* `1:length(x)` -> `seq_along(x)`
* Fix some Roxygen warnings
* Call external functions with `::`
* Formatting fixes in DESCRIPTION

Supposedly this shouldn't change any behavior, but please do check me on that. I know there are a lot of lines changed here, but most of them are whitespace, so e.g. `git diff --ignore-all-space develop..settings-lint` may help bring it down to a reviewable size.

## Motivation and Context

I was trying to refresh my memory on how check.settings works and realized I was being distracted by uneven spacing that made it hard to follow the control flow, so I ran `lintr` and fixed most of the things it complained about.

A plain `lintr::lint_package("base/settings")` would have caught most of the issues fixed here, but for the record here's the more customized call I used to turn on some extra checks that aren't in the default set (and to turn *off* the default enforcement of snake_case naming -- I don't think wholesale renaming of internal variables would be a good use of anyone's time):

```
library("lintr")
lint_package(
  commandArgs(TRUE)[[1]],
  linters = list(
    assignment_linter,
    closed_curly_linter,
    commas_linter,
    commented_code_linter, # NB left a few false positives
    # default cyclocomp_linter turned off
    equals_na_linter,
    function_left_parentheses_linter,
    infix_spaces_linter,
    line_length_linter = line_length_linter(80), # NB fixed most but not all of these
    no_tab_linter,
    nonportable_path_linter,
    object_length_linter,
    # default object_name_linter turned off
    object_usage_linter,
    open_curly_linter,
    paren_brace_linter,
    pipe_continuation_linter,
    semicolon_terminator_linter,
    seq_linter,
    single_quotes_linter,
    spaces_inside_linter,
    spaces_left_parentheses_linter,
    T_and_F_symbol_linter,
    trailing_blank_lines_linter,
    trailing_whitespace_linter,
    undesirable_function_linter,
    undesirable_operator_linter))
```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


base/settings/R/MultiSettings.R
```diff
@@ -142,59 +148,61 @@ settingNames <- function(multiSettings, settingNames) {
...
 } # settingNames
 
-##' @export
+#' @export
 print.MultiSettings <- function(x, printAll = FALSE, ...) {
   if (printAll) {
     NextMethod()
   } else {
-    print(paste0("A MultiSettings object containing ", length(x), " Settings."), ...)
+    print(
```
robkooper: do we want to switch this to a log statement?

base/settings/R/check.all.settings.R
```diff
@@ -147,160 +167,180 @@ check.database <- function(database) {
...
   ## The following hack handles *.illinois.* to *.uiuc.* aliases of ebi-forecast
-  if(!is.null(database$host)){
+  if (!is.null(database$host)) {
     forcastnames <- c("ebi-forecast.igb.uiuc.edu",
-                      "ebi-forecast.igb.illinois.edu") 
-    if((database$host %in% forcastnames) &
-       (Sys.info()['nodename'] %in% forcastnames)){
+                      "ebi-forecast.igb.illinois.edu")
+    if ((database$host %in% forcastnames)
+        && (Sys.info()["nodename"] %in% forcastnames)) {
```
robkooper: interesting, the programmer in me says that a single & is a bit operator, and && is a logical operator. Not sure what R thinks of that, but nice that it catches things like this.

base/settings/R/check.all.settings.R
```diff
@@ -147,160 +167,180 @@ check.database <- function(database) {
...
+    return(invisible(papply(settings, check.settings, force = force)))
   }
-  
-  scipen = getOption("scipen")
-  on.exit(options(scipen=scipen))
-  options(scipen=12)
-  
+
+  scipen <- getOption("scipen")
+  on.exit(options(scipen = scipen))
```
robkooper: I had not heard of on.exit. We should use this to close database connections if we open them in a function. This will remove the kludge I have right now sometimes, `if open connection set closeit=True` and later `if closeit db.close(con)`.

base/settings/R/createMultisiteMultiSettings.r
```diff
@@ -1,85 +1,100 @@
...
+  if (missing(nSite)) {
     siteIds <- allSites$site_id
-  else 
-    siteIds <- sample(allSites$site_id, nSite, replace=FALSE)
+  } else {
+    siteIds <- sample(allSites$site_id, nSite, replace = FALSE)
+  }
 
-  settings <- createMultiSiteSettings(templateSettings, siteIds)
+  createMultiSiteSettings(templateSettings, siteIds)
```
robkooper: this creates the actual settings variable and overwrites the global version I think. I'm not sure this does not break things if we remove it.

robkooper:
Nice job @infotroph

base/settings/R/check.all.settings.R
infotroph: You know, I never made the connection before, but R's vectorwise `&` behaves a lot like a bitwise AND. Or I could turn it around: 🤯Dude, C has vector operations after all!

base/settings/R/check.all.settings.R
infotroph: @robkooper on.exit is great, but one pitfall to know: You nearly always want to use it with `add=TRUE`, otherwise it by default it *replaces* any previously requested on.exit operations. I should add that here for defensiveness.

base/settings/R/createMultisiteMultiSettings.r
infotroph: I'm *fairly* sure it doesn't break anything, but it's good to check. Does this reasoning seem sound?
* `<-` binds in the current environment; we'd need to use `<<-` to affect the global object.
* All calls to `createSitegroupMultiSettings` in the codebase assign their result to an opbject named `settings`, so even if we were globally assigning it here we'd overwrite it on return.
* OK, technically this does change behavior: Since `<-(x, value)` returns `value` invisibly and we were using that assignment line as the implicit return value of this function, this change mean createSitegroupMultiSettings now returns its result visibly rather than invisibly :)

base/settings/R/MultiSettings.R
infotroph: I'd say no -- the print method for a settings object can get called in quite a few settings where a log statement would be inappropriate, e.g. inspecting settings during an interactive session, or inside a `logger.info(print2string(settings))`.

PR #2484, opened 2019-11-21
Make sure not to overwrite previously set on.exit actions

infotroph:
## Description
Following up on one of @robkooper's comments on #2483 before I forget about it

## Motivation and Context
IMO the default `on.exit` behavior (of replacing any actions scheduled earlier in the function rather than adding to them) is backwards and for defensive programming we should usually call it with `add = TRUE` unless the situation specifically requires `add = FALSE`

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/api/R/get_model_id.R
+++ b/api/R/get_model_id.R
@@ -13,7 +13,7 @@ get_model_id <- function(con, name, revision, multi_action = "last") {
     "SELECT id FROM models WHERE model_name = $1 and revision = $2 ORDER BY id DESC"
   ))
   res <- DBI::dbBind(qry, list(name, revision))
-  on.exit(DBI::dbClearResult(res))
+  on.exit(DBI::dbClearResult(res), add = TRUE)
   id <- DBI::dbFetch(res)[["id"]]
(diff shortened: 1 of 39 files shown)


mdietze:
`on.exit` is used a lot of other places in PEcAn -- do we want to fix them all at once since it's on our minds.

infotroph:
@mdietze That would be smart, but will need a little bit of care to make sure we don't clobber any places add=FALSE *was* intended. If you're OK leaving this open for a while I can try to work through the other packages and slowly add them in.

mdietze:
From previous discussions when we wrote those other sections, I don't think any of us were aware that add=FALSE was the default and having done a quick search I can't think of any example where we'd deliberately want to clobber that on.exit stack, as it would just lead to errors and leaks.

infotroph:
Don't merge yet; More commits coming soon.

infotroph:
@robkooper @mdietze this is ready for another look

Issue #1428, opened 2017-05-31
Issue with error handling - if test dates are outside of model run dates


serbinsh:
A RTFM issue but if you, say, accidentally set the ensemble run outside of the model run dates then source workflow.R, it won't create the output dir and then we get this error:

```
2017-05-31 09:15:21 WARN   [check.database.settings] :
   Will not write runs/configurations to database.
2017-05-31 09:15:21 WARN   [check.bety.version] :
   Last migration 20170415183619 is more recent than expected
   20141009160121. This could result in PEcAn not working as expected.
2017-05-31 09:15:22 SEVERE [fn] :
   Start year of ensemble should come after the start.date of the run
Warning in file(file, ifelse(append, "a", "w")) :
  cannot open file '/data/Model_Output/fates_pecan_runs/new_test_2/STATUS': No such file or directory
Error in file(file, ifelse(append, "a", "w")) :
  cannot open the connection
Error during wrapup: cannot open the connection
```
If you manually create the output folder ahead of time you get this.
```
2017-05-31 09:13:18 SEVERE [fn] :
   Start year of ensemble should come after the start.date of the run
Error in logger.severe("Start year of ensemble should come after the start.date of the run") :
  Start year of ensemble should come after the start.date of the run
> # Write pecan.CHECKED.xml
> PEcAn.settings::write.settings(settings, outputfile = "pecan.CHECKED.xml")
[1] "/data/Model_Output/fates_pecan_runs/new_test_1/pecan.CHECKED.xml"
```

If you correct the date mismatch you get a run that will follow through to completion

## Description

It seems we should stop and provide the error that the dates don't match before moving on so that the error is more descriptive?

serbinsh:
Such as 

```
nohup Rscript workflow.R > workflow.log 2>&1 &

```

infotroph:
I think I've tracked down what's happening here, and possibly in some of our other "workflow continues after severe" issues: The workflow's error handler calls `status.end` then `kill.tunnel` then `q`, which means any error in `status.end` or `kill.tunnel` causes the error handler to stop (with message "error during wrapup") before it calls `q`... and therefore the whole workflow will think the error has been handled and try to continue execution.

[...]

Issue #18, opened 2013-03-23
Remove settings as a global variable


robkooper:
Many places in the code assume settings as a global variable, this should be removed [redmine-1636](https://ebi-forecast.igb.illinois.edu/redmine/issues/1636)

robkooper:
Fix https://github.com/PecanProject/pecan/blob/master/utils/R/convert.input.R#L18

infotroph:
One way of checking for these is to inspect the "checking R code for possible problems" section of package check results for a NOTE to the effect of `no visible binding for global variable ‘settings’`.  Here's what I get today (I did it by modifying `make check` as a lazy way of mapping across packages, but searching a Travis log would work just as well):

```
$ make check
mkdir -p .check/base
Rscript -e "devtools::check('"base/logger"', error_on='never')" 2>/dev/null | grep 'no visible.*settings' || true
echo `date` > .check/base/logger
Rscript -e "devtools::check('"base/utils"', error_on='never')" 2>/dev/null | grep 'no visible.*settings' || true
   convert.input: no visible binding for global variable ‘settings’
   status.check: no visible binding for global variable ‘settings’
   status.end: no visible binding for global variable ‘settings’
   status.skip: no visible binding for global variable ‘settings’
   status.start: no visible binding for global variable ‘settings’
  convert.input: no visible binding for global variable ‘settings’
  status.check: no visible binding for global variable ‘settings’
  status.end: no visible binding for global variable ‘settings’
  status.skip: no visible binding for global variable ‘settings’
  status.start: no visible binding for global variable ‘settings’
echo `date` > .check/base/utils
Rscript -e "devtools::check('"base/db"', error_on='never')" 2>/dev/null | grep 'no visible.*settings' || true
   query.data: no visible binding for global variable ‘settings’
   query.priors: no visible binding for global variable ‘settings’
   query.traits: no visible binding for global variable ‘settings’
  query.data: no visible binding for global variable ‘settings’
  query.priors: no visible binding for global variable ‘settings’
  query.traits: no visible binding for global variable ‘settings’
echo `date` > .check/base/db
Rscript -e "devtools::check('"base/settings"', error_on='never')" 2>/dev/null | grep 'no visible.*settings' || true
echo `date` > .check/base/settings
Rscript -e "devtools::check('"base/visualization"', error_on='never')" 2>/dev/null | grep 'no visible.*settings' || true
echo `date` > .check/base/visualization
[...]

PR #2487, opened 2019-11-29
Status.* overhaul

infotroph:
## Description

This is a refactoring of the utils::status.* functions that tries to address several longstanding issues:

* Removes reliance on a global settings object (#18, our oldest open issue!), but still allows it for backwards compatibility:
  - Old behavior: Take input/output path from `settings` if it exists, otherwise return without doing anything. 
  - New behavior: Accept a "file" argument so that users can directly specify which STATUS file to read/write. If `file` not provided, still look for a `settings` object and use it if it exists.
* Doesn't fail when settings$outdir doesn't exist.
  - Old behavior: Assume that `settings$outdir` exists and is writable, allowing `cat` to fail if it isn't. When this happened while `status.end` was being called inside the `workflow.R` error handler, this was sometimes making workflows fail to stop on SEVERE (e.g. #1428). 
  - New behavior: If `file` isn't specified, check whether `settings$outdir` exists and is writeable. If so, write to `<settings$outdir>/STATUS`. If not, write to the console as a last-ditch backup. 
  - NB I'm not sure I took the right direction here! See Context section.
* Combined documentation of all four functions into one Rd file, for easier overview of what they all do.
* Added tests.

## Motivation and Context

I started this while debugging an issue similar to #1428: My workflow continued after issuing a logger.severe and I tracked it back to a bad filepath causing an error in `status.end` inside the workflow's error handler, so I tried to make it impossible for bad paths to cause an outright failure. 

But now that I've implemented it, I'm not sure I took the right approach. Opening the PR for discussion, but expecting we'll need to iterate rather than merge this as it is. I think the following are all good arguments *against* my implementation:

* The point of these functions is to write to a status file, so it's appropriate for them to complain if they can't find/create a file. Writing to the console instead is kind of silly -- most modules will announce their start and end there anyway.
* If our long-term goal is to stop relying on a global settings object, is making it optional here a step in the right direction or just adding complexity? Maybe it's better to switch outright and require that all status calls specify the path to the status file.
* It's a worthy goal to make sure no errors happen inside an error handler, but maybe that concern should be handled in the error handler (e.g. wrapping status.end in a tryCatch) rather than changing the design of the status functions.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


infotroph:
(Whatever other changes this needs, it *definitely* needs a CHANGELOG entry too.)

mdietze:
* I don't think we should retain backward compatibility on the existence of a global `settings` object. We need to fix those entirely, not halfway -- if a function wants to have a dependency on settings it has to passed in explicitly as an arg.
* Would it be better to write to the current working directory rather than the console if there's no file? I think if it's an error message the file is NULL then console is fine, but if you're in a place that really requires a file, then `getwd`.

infotroph:
> We need to fix those [global settings] entirely, not halfway

Check. I'll strip out the settings lookup.

> Would it be better to write to the current working directory

@mdietze I like `getwd` as a unix-y default for many functions, here I'd lean towards having no default and complaining if no path is given -- I can't think of any other workflow functions that assume cwd, and I think refusing to have a default is probably helpful for avoiding local vs remote mixups.

infotroph:
Since fully stripping out the global settings object requires an interface change (new mandatory `file` parameter), here's a possible alternate approach: 

1. Move these refactored status functions to the PEcAn.workflow package, since that's where they're used.
1. Leave existing utils::status* as they were before this PR, but mark them as deprecated and scheduled for removal in PEcAn 2.0.
1. While we're mucking with the interface: The current usage pattern uses a fair amount of boilerplate, and will use more with a mandatory file argument:
```
if (PEcAn.utils::status.check("MODEL", statusfile) == 0) {
  PEcAn.utils::status.start("MODEL", statusfile)
  PEcAn.remote::runModule.start.model.runs(settings, stop.on.error = FALSE)
  PEcAn.utils::status.end(statusfile)
}
```
Perhaps we can collapse this into a single wrapper that handles all the status lookups/writes for a modules, e.g. something like
```
with_status(statusfile, "MODEL", PEcAn.remote::runModule.start.model.runs(settings, stop.on.error = FALSE))
```

robkooper:
One other option is to write this to the database, that way all systems have access to the status of each workflow, and on top we don't have to read it from disk to get the status. We can add now just do a single query and get all information about the workflow, including the status of each step. We can leverage of the attributes table for this.

infotroph:
@robkooper this raises a design question: 

I've been assuming* the STATUS file was intended as a permanent artifact that would stay with the output for reference/diagnosis after the run (maybe including being zipped up, copied around, then inspected by a human reader without access to the DB). If so, keeping a static file seems useful.

 Is it actually more for use by the system to keep track of stages *during* workflow execution, with any archival uses secondary? If so, then moving it to the DB seems reasonable.

*Now that I think about it, I don't know why I assumed this... Maybe only because it seemed like there had to be a reason it didn't live in the DB already!

mdietze:
I'd been working under a similar assumption to @infotroph.  Before we switch this to the DB, which would take more coding than the current fix focused on paths not existing,  I think we'd need a large discussion of the pros and cons. My concerns initial are

1. how will this complicate remote operations, which need to run without talking to the DB
2. this seems likely to lead to missed status logs when errors occur
3. do we really need to be syncing this info across machines

but on the plus, it would make it easier to query (and thus possibly clean up) failed runs.

infotroph:
I've opened #2488 for discussion of a switch to DB; for this PR I will keep writing to files, but a `with_status` wrapper would  be easy to update later to include notifying other backends.

mdietze:
I do like the `with_status` idea

infotroph:
@mdietze I don't think this should have been merged yet -- my takeaway from the discussion above was that I needed to refactor first. Sorry for not marking WIP

robkooper:
@infotroph I think we can do that in a new pull request based on #2488 .

Issue #1413, opened 2017-05-23
met workflow continues after logger.severe


infotroph:
I'm not sure if this is actually a bug or working as intended. In the run below, I'm trying to download raw data to extend an existing input. The initial failure is pure user error -- I'm requesting data that doesn't exist -- but I was expecting the whole conversion workflow to exit after the `logger.severe` in the download stage. Instead, met2CF goes ahead and the workflow only stops at the *second* `logger.severe`, when met2CF discovers that it's attempting to convert a file that wasn't downloaded. 

workflow output:

```
2017-05-23 10:22:12 DEBUG  [convert.input] : 
   Convert.Inputs download.Geostreams NA localhost 
   /home/a-m/black11/.pecan/dbfiles/Geostreams_site_8-4906 Geostreams 
   meteorology application/json 8000004906 2016-01-01 2016-12-31 
2017-05-23 10:22:12 INFO   [convert.input] : 
   start CHECK Convert.Inputs download.Geostreams NA localhost 
   /home/a-m/black11/.pecan/dbfiles/Geostreams_site_8-4906/ Geostreams 
   meteorology application/json 8000004906 2016-01-01 2016-12-31 
2017-05-23 10:22:12 WARN   [dbfile.input.check] : 
   Found possible matching input. Checking if its associate files are on 
   host machine 
     id    site_id notes          created_at          updated_at start_date
1 8e+09 8000004906       2017-05-23 10:19:34 2017-05-23 10:19:34 2016-02-16
    end_date                   name parent_id user_id access_level raw
1 2016-09-24 Geostreams_site_8-4906        NA      NA            4  NA
  format_id
1        46
2017-05-23 10:22:12 INFO   [convert.input] : 
   8000001414 Clowder.UIUC Energy Farm - SE.2016-02-16.2016-09-24 
   /home/a-m/black11/.pecan/dbfiles/Geostreams_site_8-4906 NA NA NA 18 
   1495552774 1495552774 Input 8000000053 
2017-05-23 10:22:12 INFO   [convert.input] : 
   end CHECK for existing input record. 
2017-05-23 10:22:12 INFO   [convert.input] : 
   Changed start/end dates to '2016-01-01'/'2016-12-31' so that existing 
   input can be updated while maintaining continuous time span. 
2017-05-23 10:22:12 DEBUG  [convert.input] : 
   convert.input executing the following function: 
   PEcAn.data.atmosphere::download.Geostreams(sitename='UIUC Energy Farm - 
[...]

PR #2489, opened 2019-12-05
make sure quit is called on error, even if other errors inside handler

infotroph:
## Description

Wraps most steps of the `workflow.R` error-handling expression inside ~a tryCatch~<ins>their own `try`</ins>, so that it always* calls `q()` at the end even if earlier steps produce errors of their own. 

This should fix several different cases of workflows continuing after what ought to be lethal errors, e.g. #1428, #1413

*always when the workflow isn't running interactively, I mean -- I didn't change the existing conditional behavior.

## Motivation and Context

#2487 is motivated by the same issue (many error-handling failures can be traced back to status.end failing to find a writable output directory), but that was the wrong place to address this: Yes, status.end ought to be as robust as possible, but the error-handler still needs to assume it *could* fail, hence this PR.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


DEBUGING.md
```diff
@@ -6,41 +6,44 @@ with the development of PEcAn.
...
-options(warn = 1, keep.source = TRUE, error =
-          quote({
-            status.end("ERROR")
+options(
+  warn = 1,
+  keep.source = TRUE,
+  error = quote(
+    tryCatch(
+      expr = {
+        status.end("ERROR")
```
robkooper: would it make sense to move the tryCatch into the status.end(). I would like the rest of the commands to run no matter what, like printing open connectins and the stack trace for debugging purposes.

Same below so we make sure the killTunnel is removed.

infotroph: How about a separate tryCatch for each function? Then we know we've *attempted* every step even if any one of them fails.

infotroph: @robkooper changed in a1a4ddab

PR #2492, opened 2019-12-15
Hyperlink DOIs to preferred resolver

katrinleinweber:
## Description

The DOI Foundation [started recommending a new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). While their URL change may be a bit ironic, it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org) and the old `dx` subdomain is being redirected. So, there is no urgency here.

## Motivation and Context

For consistency, this PRs suggests to update all static DOI links accordingly, plus the code that generates new DOI links>

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [-] My change requires a change to the documentation.
- [-] I have updated the CHANGELOG.md.
- [-] I have added tests to cover my changes.
- [-] All new and existing tests passed.


katrinleinweber:
I think the build error results from https://travis-ci.org/PecanProject/pecan/jobs/625236895#L6087, but I'm not sure how to view those warnings.

infotroph:
@katrinleinweber Thanks for the patch! 

If it helps for future PRs, since it's largely my fault that the Travis logs are so confusing: My usual method of reading the log from a failed build is to work my way up from the bottom, looking for `make: *** [...some path here...] Error 1`. The cause of the failure is usually somewhere immediately above that -- In this case on line 6118, where data.table complains about a version change that should resolve itself soon and wasn't related to your changes.

PR #2497, opened 2019-12-20
fix get.rh, added tests

dlebauer:
the get.rh function was not working at all!!!

## Description

presumably this didn't work even when I first wrote it because the constants `L` and `Rw` were undefined

currently returns RH within 0.5% of other sources including https://www.wpc.ncep.noaa.gov/html/dewrh.shtml and Lawrence 2005 where the equiation comes from https://doi.org/10.1175/BAMS-86-2-225
## Motivation and Context

Allows users to get relative humidity when dewpoint is provided

Obviously not required if it hasn't been used in 6 years! However, I also use this file as a reference for met conversions.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


mdietze:
```
 M modules/data.atmosphere/man/get.rh.Rd
```

PR #2502, opened 2019-12-27
site_id argument removed from functions, wherever it is not used.

rahul799:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x ] When possible
## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue) 

## Checklist:


rahul799:
@infotroph I have removed the `site_id` argument from all the functions (inside **data.atmosphere**) wherever it is not used but is taken as an argument. Let me know if you want any changes.

mdietze:
There appear to be a number of places, such as in tests and vignettes, where these functions are still called and still passing the `site_id` argument

rahul799:
@mdietze Thanks for pointing this out, let me know if I am still missing something.

mdietze:
I'm guessing the following error is not a direct result of your changes, but an indirect effect of @infotroph 's recent changes to make the build system more strict. If you wouldn't mind fixing these errors, the undocumented variable `...` should be pretty easy. Not sure why it's also giving a warning about `outfolder` as that one exists in the ROxygen tags (but doesn't have any text describing it)

```
R check of modules/data.atmosphere returned new problems:
checking Rd \usage sections ... WARNING: Undocumented arguments in documentation object 'met_temporal_downscale.Gaussian_ensemble' 'outfolder' '...'
```

infotroph:
> I'm guessing the following error is not a direct result of your changes

Yep. That line used to complain about three undocumented arguments and you've fixed one of them. I agree with @mdietze that the easiest fix is to document the remaining two.

> Not sure why it's also giving a warning about outfolder as that one exists in the ROxygen tags (but doesn't have any text describing it

Roxygen treats argument names with no text the same as if they were completely missing, so this adds up.

To fix both, I suggest editing line 13 to 

```
##' @param outfolder path to directory in which to store output. Will be created if it does not exist
```

and adding a new line at the end of the Roxygen parameter block (between current lines 24 and 25):

```
##' @param ... further arguments, currently ignored
```

(It's possible we could instead drop `...` from the function signature, but maybe this was intended to accept extra args passed from elsewhere?)

mdietze:
```
 M modules/data.atmosphere/man/extract.local.CMIP5.Rd
 M modules/data.atmosphere/man/extract.local.NLDAS.Rd
 M modules/data.atmosphere/man/met_temporal_downscale.Gaussian_ensemble.Rd
```

rahul799:
@mdietze after fixing a lot many bugs, I am finally able to install all the dependencies & able to run `R CMD CHECK`
The result after the check was
`
0 errors ✓ | 8 warnings x | 3 notes x
Error: R CMD check found WARNINGs
`
The warnings were very basic like

-  `@param requires name and description`, (around 35+ files have this problem)
-  `Found the following file with non-ASCII characters: download.PalEON.R`

Etc. So do I need to fix these warnings?

infotroph:
@rahul799 Good work getting the dependencies installed! That's a real achievement and something we want to make easier in the future. 

Short answer: No, you don't have to fix all of the issues, only any that are new in this PR. The easiest route is to run `make check`:  If it says `R check of [package] returned new problems`, the issues it lists below that line are the new ones.

Longer answer: The messages you're seeing are mostly from older code that was developed without continuous integration. Rather than try to fix all of these at once we've decided to selectively ignore (for now) the existing issues but break the build on new ones. Our CI runs call `make check`, which calls `scripts/check_with_errors.R`, which runs R CMD check, compares the current result against a previous result stored in `<package_root>/tests/Rcheck_reference.log`, and only complains about the issues that *do not* exactly match an issue in the stored result.

infotroph:
@rahul799 It looks like the immediate reason for the latest Travis failure is that the existing documentation was built using Roxygen 6.1.1 and you used Roxygen 7.0.2, which has a slightly different line-wrapping algorithm and produces a lot of small (mostly whitespace) changes that touch most of the Rd files. We'll update to the newer Roxygen at some point, but not yet -- as it stands we need to make all the PEcAn developers switch at the same time.

Can you please install Roxygen 6.1.1 (probable easiest way: `remotes::install_version("roxygen2", "6.1.1")`), rerun `make document`, and push those changes? I'm sorry for the hassle.

mdietze:
Thanks @rahul799 for tackling this!

rahul799:
During fixing the bugs (mostly while running `R CMD CHECK`), I learned many things. That's what  I like the most about open source projects. @mdietze @infotroph Thanks for your feedback. :blush:

PR #2522, opened 2020-01-23
Weighted Tobit2Space Model

araiho:
## Description

Calculating a weighted mean and covariance in Tobit space was not converging with simple random walk samplers. So, I re-coded the nimble conjugate wishart sampler to be able to include ensemble weights. I also added Gelman-Rubin diagnostics and a way to calculate a weighted mean and covariance if you weren't in Tobit space. 

## Motivation and Context

This change is required to estimate valid weighted means and covariances which are then given to the Tobit wishart ensemble filter

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


para2x:
@araiho This is a very large PR in terms of changes to the core analysis function. Do you like me to pull and test it or you think just reading through the code is enough ?

araiho:
@para2x I think it just looks big because stuff got shifted around. But yeah I think pulling and testing would be best.

para2x:
sure, I'm gonna give it try for the PalEON long run, see what happens.

modules/assim.sequential/R/get_ensemble_weights.R
```diff
@@ -23,9 +24,9 @@ get_ensemble_weights <- function(settings){
     ### Loading Weights                                                   ###
     ###-------------------------------------------------------------------###
     weight_file <- read.csv(settings$run$inputs$ensembleweights$path)
```
para2x: @araiho I'm testing this PR without weights first, but If I wanna test it with weights seems like I need this weights be read in first. Could you plz show me how to make this file ?

araiho: The weights come from an analysis by John Tipton. There is an example output file from his analysis here: /fs/data2/output/PEcAn_1000008588/ensemble-weights-monte-carlo-mean-harvard.csv

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -151,128 +153,170 @@ GEF<-function(settings, Forecast, Observed, H, extraArg, nitr=50000, nburnin=100
...
+        x.ind[n,j] <- as.numeric(X[n,j] > 0)
+        x.censored[n,j] <- as.numeric(ifelse(X[n,j] > intervalX[j,2], 0, X[n,j])) #
       }
     }
     
-    #conf_tobit2space$printSamplers()
-    
-    Rmcmc_tobit2space <<- buildMCMC(conf_tobit2space)
+    recompileTobit <- extraArg$recompileTobit
+    recompileGEF <- extraArg$recompileGEF
```
para2x: `extraArg$recompileGEF` and ` extraArg$recompileTobit` are not set in the main `sda.enkf` function. This `extraArg$recompile` is set however. Either you can set those or replace em with the one that is set.

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -151,128 +153,170 @@ GEF<-function(settings, Forecast, Observed, H, extraArg, nitr=50000, nburnin=100
...
+      
+      inits.tobit2space <- function() list(muf = rmnorm_chol(1,colMeans(X),chol(diag(ncol(X))*100)),
+                                  pf = rwish_chol(1,df = ncol(X)+1,cholesky = chol(solve(cov(X)))))
+      
+      #ptm <- proc.time()
+      tobit2space_pred <<- nimbleModel(tobit2space.model, data = data.tobit2space,
+                                       constants = constants.tobit2space, inits = inits.tobit2space(),
+                                       name = 'space')
+      
+      logprob_y_tobit2space <- tobit2space_pred$calculate('y.censored')
```
para2x: I get `NA` here I don't know why. At first I thought It's because it's called before adding it as monitor in L208, but even if I add that first it's `NA` and breals the next line in the `if`. Is there something I'm missing ?

para2x: it turns out, it was bc of my weights were `NA`, If I choose a series of weights this will return some number. Maybe an `if` would help here.

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -139,7 +141,7 @@ GEF<-function(settings, Forecast, Observed, H, extraArg, nitr=50000, nburnin=100
   path.to.models <- file.path(settings$outdir,"SDA","GEF")
   aqq <- extraArg$aqq
   bqq <- extraArg$bqq
-  wts <- extraArg$wts
+  wts <- extraArg$wts/sum(extraArg$wts)
   interval <- NULL
```
para2x: suggestion :
`if(all(is.na(wts))) wts <- rep(1, length(extraArg$wts)) # if there is no weight then they all have the same weight`

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -379,70 +420,89 @@ GEF<-function(settings, Forecast, Observed, H, extraArg, nitr=50000, nburnin=100
...
+    conf <- configureMCMC(model_pred, print=TRUE,thin = 10)
+    conf$addMonitors(c("X","X.mod","q","Q", "y_star","y.censored")) 
+    ## [1] conjugate_dmnorm_dmnorm sampler: X[1:5]
+    
+    if(ncol(X) > length(y.censored)){
+      print('Adding a different sampler to non-data node to improve MCMC mixing')
+      conf$removeSampler('X')
+      #consider individual slice samplers
+      #conf$addSampler(paste0('X[1:12]'),'AF_slice')
+      for(ss in 1:13){
```
para2x: Is `13` the `ncol` of the X ?

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -379,70 +420,89 @@ GEF<-function(settings, Forecast, Observed, H, extraArg, nitr=50000, nburnin=100
...
+        conf$addSampler(paste0('X[',ss,']'),'slice')
+      }
+      conf$printSamplers()
+    }
+    
+    if(FALSE){ ### Need this for when the state variables are on different scales like NPP and AGB
+      x.char <- paste0('X[1:',ncol(X),']')
+      conf$removeSampler(x.char)
+      propCov.means <- c(rep(1,9),1000)#signif(diag(Pf),1)#mean(unlist(lapply(obs.cov,FUN = function(x){diag(x)})))[choose]#c(rep(max(diag(Pf)),ncol(X)))#
```
para2x: `9` is hard-coded, here

modules/assim.sequential/R/Analysis_sda.R
```diff
@@ -325,12 +367,11 @@ GEF<-function(settings, Forecast, Observed, H, extraArg, nitr=50000, nburnin=100
   
   if(any(grep(names(data_available),pattern = 'ALR'))){
     
-    # browser()
     which_fcomp <- grep(names(data_available),pattern = 'ALR')
     X_fcomp_start <- which_fcomp[1]
     X_fcomp_end <- which_fcomp[length(which_fcomp)]
     X_fcomp_model <- grep(colnames(X),pattern = 'AGB.pft')
```
para2x: Do you think `*.pft.*` would work for your case/project instead of `AGB.pft` ? `*.pft.*` makes things much easier for ED state variables

modules/assim.sequential/R/Analysis_sda.R
araiho: The weights need to sum to 1 so that's why I put this here. But, maybe I should move it to the main sda function?

modules/assim.sequential/R/Analysis_sda.R
araiho: Yes! Thanks. I didn't know about the double *

modules/assim.sequential/R/Analysis_sda.R
para2x: I think it's fine here, but I was suggesting to replace NA values in `wts` with some weights which you added your fix.

para2x:
@mdietze @araiho I tested this PR for the PALEON long runs it worked for me after Ann addressed my suggestions. This should work out of the box for the LINKAGES but it also needed my ED specific changes that I will push them to #2451 .

araiho:
We are having a problem with this PR because the load_nimble function as a bunch of functions contained in it. Should I go through and comment them each out separately? Is there a good example to look at somewhere else? These are the nimble statistical models that we use for the SDA. 

We are also having a problem because we keep some variables in the global environment with a <<-  in an inner function (GEF) so that we don't have to recompile the statistical models. Is there a different way to do this?

modules/assim.sequential/DESCRIPTION
```diff
@@ -11,16 +11,32 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
     streamline the interaction between data and models, and to improve the
     efficacy of scientific investigation.
 Imports:
+    corrplot,
+    devtools,
```
robkooper, line 15: is this needed as an import?

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -1,184 +1,308 @@
...
-      returnType(double(1))
-      
-      y_star <- X
-      
-      return(y_star)
-    })
+#'
+#'
+#' @export
+y_star_create <-  nimbleFunction(
```
robkooper, line 11: to be consistent should this be `nimble:: nimbleFunction`

PR #2524, opened 2020-01-28
Roxygen 7

infotroph:
## Description

First commit in this PR is just bumping Roxygen and recompiling docs:

```
git pull upstream develop
git switch -c roxygen-7
Rscript -e 'install.packages("roxygen2")'
make clean
make document
```

* `**/DESCRIPTION`: version bump in the RoxygenNote field, no other changes
* `**/NAMESPACE`: only modules/emulator changes, by now exporting S4 classes `jump` and `mvjump`. I see no problem with this, but someone who understands S4 classes better than me should double-check.
* `*/man*.Rd`: Changes are *mostly* minor and obviously correct: line-wrapping, tweaks to translation between `foo` and \code{foo} vs \verb{foo}, better cross-package reference links, etc.Exceptions to the "obviously correct" rule are handled below.

I reviewed the above diff by hand, took notes on changes that looked wrong, then committed it and moved on to phase two. The remaining commits are me finding and resolving issues uncovered by the version change:

* Removed backslash escaping from percent signs inside Markdown text, because Roxygen now provides the escaping automatically. But beware: percents still need to be escaped in the packages that *don't* use Markdown (which is still most of them).
* Apparently the recommended way of providing whole-package documentation has changed slightly; I updated PEcAn.DB, the only one where it changed the result.
* ACTION NEEDED: When documenting multiple functions in the same Rd file, Roxygen used to include multiple copies of any function arguments that were specified multiple times. It now includes only one copy and *ignores* the others, even if they're worded differently. There were a few places it wasn't obvious how to resolve this; @ashiklom and @para2x, please see my comments below.
* Turns out `PEcAn.data.atmosphere::met2CF.PalEONregional` was defined twice, copy-paste identically, in the same file. Fixed that.
* Various small formatting / example / Roxygen syntax issues that weren't causing check errors before but started now. All needed fixing before too, the update just made them louder.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/db/man/build_insert_query.Rd
```diff
@@ -11,7 +11,7 @@ build_insert_query(values, table, .con)
 
 \item{table}{Name of target SQL table, as character}
 
-\item{.con}{[\code{DBIConnection}]:A DBI connection object obtained from \code{DBI::dbConnect()}.}
+\item{.con}{Database connection object}
```
infotroph: I suspect this isn't a desirable change and I think there's a chance it's a Roxygen bug, but I'm not sure how to resolve it efficiently: Best I can tell it switched from inheriting `.con` from `glue::glue_sql` to inheriting `con` (no dot!) from `db.query`. @ashiklom I think you wrote these docs -- thoughts?

modules/assim.sequential/man/sda.enkf.Rd
```diff
@@ -5,41 +5,44 @@
...
-
-\item{settings}{PEcAn settings object}
-
-\item{obs.mean}{List of dataframe of observation means, named with observation datetime.}
-
-\item{obs.cov}{List of covariance matrices of state variables , named with observation datetime.}
-
-\item{Q}{Process covariance matrix given if there is no data to estimate it.}
-
-\item{restart}{Used for iterative updating previous forecasts. When the restart is TRUE it read the object in SDA folder written from previous SDA.}
```
infotroph: This file used to contain duplicate definitions for the parameters that are shared between sda.enkf and sda.enkf.original, but when rebuilt using Roxygen 7 it now only lists each parameter name once.

That seems desirable for functions documented in the same file, but note that where definitions differ only the one from `sda.enkf` is used (and the definitions from `sda.enkf.original` are not visible anywhere). @para2x please check this carefully: Is that the correct result here?

base/db/man/build_insert_query.Rd
ashiklom: I think it's fine. "Database connection object" is somewhat less precise than "DBI Connection", but not wrong, and to my knowledge, this function isn't that widely used anyway.

base/utils/R/plots.R
```diff
@@ -23,7 +23,7 @@
 ##' @param eps used to set artificial bound on min width / max height of bins as described in Denby and Mallows (2009) on page 24.
 ##' @param xlab is label for the x axis 
 ##' @param plot = TRUE produces the plot, FALSE returns the heights, breaks and counts
-##' @param lab.spikes = TRUE labels the \% of data in the spikes
+##' @param lab.spikes = TRUE labels the % of data in the spikes
```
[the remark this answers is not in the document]
infotroph: Seems to:
Screen Shot 2020-01-28 at 6 23 14 PM

My understanding is that the PDF version is  built from the Rd files without re-invoking Roxygen, and it's [still spelled `\%` in the Rd file](https://github.com/PecanProject/pecan/pull/2524/files#diff-df15bd8a2cee783f39a18065dc44e4f4L26). this is just a switch from us typing it to Roxygen inserting it.

PR #2530, opened 2020-01-31
fix-2529 enforcing standard code style

### LINKED ISSUE #2529: Enforcing standard code style using Prettier
### Bug Description

Enforcing standard code styling too `.js` , `.html` & `.css` files using [Prettier
](https://prettier.io/)

rahul799:
## Description
fix #2529 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


docker/docs/index.html
```diff
@@ -1,45 +1,112 @@
...
-<li><a href="/pecan/history.php">PEcAn runs done previously.</a></li>
-<li><a href="/minio/">Browse the PEcAn data using minio graphicsal interface</a></li>
-<li><a href="/thredds/">Browse the PEcAn generated data using Thredds server</a></li>
-<!--<li><a href="/rstudio/">RStudio editor with PEcAn installed.</a></li>-->
-<!--<li><a href="/pecan/setups/">Basic Setups</a></li>-->
-</ul>
+    <ul>
+      <li>
+        <a href="/pecan/"
+          >PEcAn web interface, from here you can run different models at
```
robkooper: I'd prefer the `>` not to be on the next line (few other places as well)

docker/monitor/bootstrap-table.min.css
```diff
@@ -7,4 +7,387 @@
   * @license MIT
```
robkooper: these are  minimized CSS, so on purpose all spaces are removed. Anything with .min. in the filename should be left as is.

shiny/SDAdashboard/www/style.css
```diff
@@ -425,10 +437,18 @@ input[type=submit]:hover {
   .card-group > .card:not(:first-child):not(:last-child):not(:only-child) {
     border-radius: 0;
   }
-  .card-group > .card:not(:first-child):not(:last-child):not(:only-child) .card-img-top,
-  .card-group > .card:not(:first-child):not(:last-child):not(:only-child) .card-img-bottom,
-  .card-group > .card:not(:first-child):not(:last-child):not(:only-child) .card-header,
-  .card-group > .card:not(:first-child):not(:last-child):not(:only-child) .card-footer {
+  .card-group
```
robkooper: not sure if this is still parsed correctly by CSS after this change (not really familiar with what happens here).

web/leaflet/MarkerCluster.Default.css
```diff
@@ -1,60 +1,60 @@
 .marker-cluster-small {
-	background-color: rgba(181, 226, 140, 0.6);
```
robkooper: All the files in leaflet, are from the leaflet disitribution so I think it is ok to leave them as is, if we upgrade the version of leaflet we will just copy the new version over the current files and not apply any pretifier to it.

web/sites.css
```diff
@@ -1,195 +1,206 @@
...
-    padding: 0;
-    height: 100%;
-    width: 100%;
+  margin: 0;
+  padding: 0;
+  height: 100%;
+  width: 100%;
 }
 
-p,h1,form,button {
```
robkooper: Not sure, this might be just be me, but I think it is easier to read to have all the elements that are changed on a single line instead of them listed one by one.

infotroph:
* Overall: 👍 👍 to enforcing style standards! This is a worthy effort. 
* I see some files in this PR that are probably autogenerated (at least AssimBatchVignette.html, possibly others); it's probably not worth trying to style these because they'll be undone the next time their source changes. But it would definitely be worth figuring out how to make the build tool produce them to standards in the first place.
* Web standards are outside my usual lane, so anyone correct me if I'm wrong here, but: My understanding of naming conventions is that `*.min.css` and `*.min.js` are intentionally minified (on the theory that they're served much more often than they're read by a human, and the extra whitespace costs bandwidth), so I'd leave these as they are
* If we want the improvements to last, we probably need a way of enforcing them for future changes -- this is part of why despite many threats, I haven't spent very much effort on unifying our R code style. Should we look into running Prettier during CI builds to identify/fix future issues as well?

rahul799:
@infotroph Thanks for the appreciation :smiley: 
I was also thinking to integrate the `Prettier` in the CI build tool. I will open a separate `PR` for that & will try to integrate it with the Travis.

robkooper:
I think we can just add `*.min.*` to `. prettierignore` (see https://prettier.io/docs/en/ignore.html#ignoring-files). We can also add some folders to ignore (leaflet).

rahul799:
@robkooper , @infotroph I will make those changes & will update you soon.

docker/docs/index.html
rahul799: @robkooper @infotroph I went through there documentation. To fix this issue, we possibly have only two options:

-  We can add a comment  which will disable auto-formatting, but it's scope is up to the next HTML tag only. So we need to add this comment again & again.

-  We can increase the [print-width](https://prettier.io/docs/en/options.html#print-width) as this issue is happening due to line width constraints, but again it's not the right solution.

Please let me know if you have any other options?

web/leaflet/MarkerCluster.Default.css
rahul799: This problem can be solved easily.

docker/monitor/bootstrap-table.min.css
rahul799: This can be solved easily.

docker/docs/index.html
mdietze: I'm ok with this break if it's driven by line length and the fix is a hack

shiny/SDAdashboard/www/style.css
rahul799: No, It will not break it.

docker/docs/index.html
rahul799: As per the discussion on slack. I am resolving this conversation

rahul799:
@infotroph, we can use `GIT Hooks` so that `Prettier` will automatically run in write mode on all the files changed, and make formatting adjustments to those entire files as necessary, before passing them through to the commit. After the pull request is made, Travis can verify those changes. But my main concern is, 
As the prettier works only for .`html`,.`js` & `CSS` & most of our files are written in `R`, is it worth to implement these, as prettier comes under npm package so the whole package will be installed locally followed by lots of dependencies.

infotroph:
@rahul799 Those are good points -- our dependency load is sizeable already, so it's definitely worth being cautious about adding more overhead to either developer machines or CI builds.

I vote we move ahead on this PR without any automated tooling, but keep it in mind as a wishlist item.

robkooper:
any chance to undo the prettfied minified files?

rahul799:
I am thinking of closing this PR & will open another one. There I will add .prettierignore file, which will take care of .min, leaflet. Can you please tell me that apart from these , which files are autogenerated. Like @infotroph mentioned AssimBatchVignette.html. Are there any other files too?

robkooper:
I think you can ignore web/sugarcane/static/

mdietze:
I think all the vignettes are autogenerated from .Rmd, not just AssimBatchVignette

PR #2531, opened 2020-02-03
Various small improvements that didn't fit elsewhere

infotroph:
Pulling useful bits out of the dregs of all my old stashes:

* clean Basgra compilation artifacts on `make clean`
* clarify wording in Travis cache buildup script
* style cleanup in uncertainty::ensemble.ens.gen
* enforce creating connection outside query function (I believe all callers do this already, but check me)
* avoid Roxygen complaint about "Unknown xml node: html_block"

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


base/db/R/query.traits.R
```diff
@@ -21,26 +21,17 @@
...
-                         update.check.only=FALSE,
-                         ids_are_cultivars=FALSE){
-  
-  if(is.null(con)){
-    con <- db.open(settings$database$bety)
-    on.exit(db.close(con), add = TRUE)
-  }
-  if(is.list(con)){
-    print("query.traits")
-    print("WEB QUERY OF DATABASE NOT IMPLEMENTED")
```
mdietze: @infotroph I don't recall exactly what set of conditions was resulting in a different type of connection object being passed to this function, but I recall it being a pain to track down (cryptic message) which is why we created this message. This was different from the case you removed, of no connection being passed in (i.e. open connection was passed it, but it wasn't in a form we could use)

infotroph: @mdietze I'll add it back. Three questions:
* Keep testing for list specifically, or convert it to a check for anything that's not an (open, usable) connection?
* Wording suggestions for the error message, or for a comment in the code to help future editors?
* logger.error or logger.severe?

infotroph: @mdietze does d49ad92ba fix this?

PR #2537, opened 2020-02-11
add documentation about ED and singularity

robkooper:
Document how to run ED in singularity. This is based on discussion in #containers channel in slack.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - PEcAn.priors now exports functions `priorfig` and `plot_densities` (#2439).
 - Models monitoring container for Docker now shows a webpage with models it has seen
 - Added small container to check if certain services are up, used as initi container for kubernetes
+- Documentation how to run ED using singularity
 
 ## [1.7.1] - 2018-09-12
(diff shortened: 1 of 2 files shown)


mdietze:
Since it looks like this specific example is able to build an ED2 binary successfully, but has an ED2IN error, it would be good to update the documentation once that error has been found/fixed so that the documentation has a clean example.

KristinaRiemer:
I'm not seeing the bash or file chunks in the book? https://pecanproject.github.io/pecan-documentation/develop/models-singularity.html

robkooper:
click on the header with >

KristinaRiemer:
This is what that page looks like for me: 
Screen Shot 2020-02-13 at 9 12 19 AM

I clicked on "bash script to install required files" text to get the code chunk to show: 
Screen Shot 2020-02-13 at 9 12 37 AM

I only knew to do that because you suggested clicking. Is there a way we can indicate that people should do that?

robkooper:
Interesting, in my case it has a little triangle in front of it. I can add some text saying "click here to expand"

PR #2538, opened 2020-02-12
fix docker builds

robkooper:
This fixes the depnds image and should result in successful builds of the docker images again.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [x] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2544, opened 2020-03-05
Add CI builds on Github Actions

### LINKED ISSUE #2412: Use GitHub actions for automated CI?
[GitHub Actions](https://github.com/features/actions) seems like a promising choice for continuous integration. A major advantage over Travis is the _really_ generous hosting -- 6 hour runtime _per job_, and up to 20 workflows in parallel ([source](https://help.github.com/en/articles/workflow-syntax-for-github-actions#usage-limits)).

Also, it looks like if we wanted to, we could migrate to this in phases, since Travis CI has its own [GitHub action](https://github.com/travis-ci/actions).

One _really_ cool feature of GitHub actions is the ability to run specific actions only when [specific files have been changed](https://help.github.com/en/articles/workflow-syntax-for-github-actions#onpushpull_requestpaths). We could use this to, e.g., only re-build and test PEcAn R packages when R code has been updated, and only re-build documentation when that has been updated, rather than doing both all the time.

Food for thought.

infotroph:
## Description

First pass at a continuous integration pipeline using GitHub Actions. These *do not* yet replace our existing Travis CI build.

Runs automatically on every push, and does the following:

* Builds and installs the core PEcAn packages inside the `depends` Docker image. It does not spin up the whole Docker-compose stack, just uses the image to avoid reinstalling dependencies.
* Sets up a test instance of BeTY using the `pecan/db:ci` Docker image, which contains the records from the most recent successful dumps of servers 0, 1, 2, 5.
* Runs all package tests, including those that require database access, 
* Runs R CMD check on all packages, and breaks the build for any newly added ERRORs, WARNINGs or NOTEs. As on Travis, issues that existed before September 2019 are grandfathered in and do not produce warnings.
* Performs a simple integration test using SIPNET
* Checks for unexpected changes to tracked files (90% of the time this means you forgot to run Roxygen before committing)

You can see the results of each run in the [Actions tab](https://github.com/PecanProject/pecan/actions). My understanding is that when you fork PEcAn these actions will run in your own fork too with no opt-in needed.

## Motivation and Context

Closes #2412. As discussed there, the single biggest apeal of Actions is that its runtime limits are much more generous than the 50-minute ceiling we often hit on Travis. Actions is also very flexible and well-integrated with the rest of GitHub; if this workflow works well, it's easy to envision adding 
more actions that interface with issues and PRs, build Docker images, run more elaborate integration tests at longer time intervals, and so on.

Since many of these changes will be relevant to possible GSoC projects this coming summer, I'd like to encourage potential applicants to join the review process for this PR. To encourage discussion, I'll keep this open for a while to make sure everyone has lots of chances to comment before it's merged. I've marked it as work in progress to avoid accidental merging while discussion is still ongoing.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


rahul799:
Travis is failing due to **tmvtnorm** package. To be more precise, it's not able to download one of it's dependencies **mvtnorm**.
```
Error: (converted from warning) package ‘mvtnorm’ was built under R version 3.6.3
Execution halted
ERROR: lazy loading failed for package ‘tmvtnorm’
* removing ‘/home/travis/R/Library/tmvtnorm’
Error in i.p(...) : 
  (converted from warning) installation of one or more packages failed,
  probably ‘tmvtnorm’
```

infotroph:
Thanks for investigating, @rahul799. Apparently [c2d4u](https://launchpad.net/~marutter/+archive/ubuntu/c2d4u3.5), which we use for precompiled binaries of some R packages, has already started updating some of their binaries to R 3.6.3 but is still serving others compiled with R 3.6.2.As far as I can tell, c2d4u's philosophy is that "compiled packages are usually compatible between patch versions of R, so mixing them is fine", but R always warns any time you load a package built for a different version -- and what we see here is the package installation process for `tmvtnorm` hitting such a warning when it loads the already-installed `mvtnorm`, then (correctly) elevating the warning to an error. 

Version issues like this have been an ongoing pain point for a while and we think they'll be easier to improve on Actions than on Travis, which is part of why we'd like to migrate when we can.

As a short-term fix, I've abused Travis's caching abilities by using some API calls to reinstall a 3.6.2-built version of `mvtnorm` and then restarted the build.

shivanshu1086:
* **Travis** is using the configuration with version of **R 3.6.2** -> [https://docs.travis-ci.com/user/languages/r/#configuration-options](https://docs.travis-ci.com/user/languages/r/#configuration-options)
* And all the dependencies get failed because the lazy loading failed to download and install all the dependencies, which is one of the primary features of lazy loading, because of version mismatch.
```

ERROR: lazy loading failed for package ‘tmvtnorm’
* removing ‘/home/travis/R/Library/tmvtnorm’
Error in i.p(...) : 
  (converted from warning) installation of package ‘tmvtnorm’ had non-zero exit status
Calls: <Anonymous> ... with_rprofile_user -> with_envvar -> force -> force -> i.p
Execution halted
Makefile:130: recipe for target '.doc/modules/emulator' failed
make: *** [.doc/modules/emulator] Error 1
```

We need to either update the R version of the Travis CI to 3.6.3 or either we can go to the 3.6.2 version of **mvtnorm**.

infotroph:
After putting a rebuilt `mvtnorm` into the build cache, this currently passes Travis. Let's now try to focus this thread on evaluating the Actions code rather than the Travis output -- Travis doesn't run any of the Actions code at all, so we'll have to notice the bugs ourselves!

shivanshu1086:
Thanks for correcting me @infotroph!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
+name: CI
+
+on: [push]
```
rahul799: Although It won't make any difference still it's a good practice to write single events as string & not an array.
```suggestion
on: push
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+  # See https://stat.ethz.ch/pipermail/r-package-devel/2019q2/003898.html
+  _R_CHECK_COMPILATION_FLAGS_KNOWN_: -Wformat -Werror=format-security -Wdate-time
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    container: infotroph/pecan-depends:latest
+    steps:
+    - uses: meetup/express-checkout@master #clones with depth 1; skips tags, branches, submodules
```
rahul799: @infotroph I will suggest that we can use `actions/checkout@v2` because the v2 has made big leaps and is now comparably fast compared to meetup/express-checkout. You can read about it [here](https://github.com/meetup/express-checkout).
```suggestion
    - uses: actions/checkout@v2 #clones with depth 1; skips tags, branches, submodules
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+    services:
+      postgres:
+        image: mdillon/postgis:9.5
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      NCPUS: 2
+      PGHOST: postgres
+      CI: true
+    steps:
+    - uses: meetup/express-checkout@master
```
rahul799:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+
+  check:
+    needs: build
+    runs-on: ubuntu-latest
+    container: infotroph/pecan-depends:latest
+    env:
+      NCPUS: 2
+      CI: true
+    steps:
+    - uses: meetup/express-checkout@master
```
rahul799:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+    runs-on: ubuntu-latest
+    container: infotroph/pecan-depends:latest
+    services:
+      postgres:
+        image: mdillon/postgis:9.5
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      PGHOST: postgres
+    steps:
+    - uses: meetup/express-checkout@master
```
rahul799:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+  # See https://stat.ethz.ch/pipermail/r-package-devel/2019q2/003898.html
+  _R_CHECK_COMPILATION_FLAGS_KNOWN_: -Wformat -Werror=format-security -Wdate-time
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    container: infotroph/pecan-depends:latest
+    steps:
+    - uses: actions/checkout@v2 #clones with depth 1; skips tags, branches, submodules
```
infotroph:
```suggestion
    - uses: actions/checkout@v2
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+    container: infotroph/pecan-depends:latest
+    services:
+      postgres:
+        image: mdillon/postgis:9.5
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      PGHOST: postgres
+    steps:
+    - uses: actions/checkout@v2
+    - run: sudo apt-get update && sudo apt-get install -y curl postgresql-client
```
rahul799: We must follow the literal style 
```suggestion
    - run: |
        sudo apt-get update
        sudo apt-get install -y curl postgresql-client
```

infotroph: Are there subtleties in behavior I should know here or is this purely a style thing? 

If it's just style: A good rule in general and If this line were even a few characters longer I'd agree. But I think when the command is conceptually one step (in this case we only want to update because we need it before installing) and fits comfortably on one line, I prefer `&&`.

rahul799: Yaa... It's just for styling... I made these remarks as above in the code you used the literal style. But I will go with you because using `&&` is a better option as the second command will only be executed if the update returns exit status zero.

infotroph:
> In a future follow-up PR, we should switch to using the "official" PEcAn images (i.e. not infotroph/pecan-depends)

100% agree -- we can probably even do it in this PR if #2538 is merged before this one

robkooper:
How hard is it to make an action that runs on a nightly basis to create a new pecan/depends:develop image?

infotroph:
> How hard is it to make an action that runs on a nightly basis to create a new pecan/depends:develop image?

Probably not too hard. I think several folks have already written that into their GSoC proposal drafts, so I'm happy waiting and letting them do it :)

infotroph:
This appears to be fully working and is ready for final review

.github/workflows/ci.yml
```diff
@@ -0,0 +1,173 @@
...
+        v='cat(numeric_version("'${v}'") < "2.18")'
+        echo "##[set-output name=isold;]$(Rscript -e "${v}")"
+    - name: upgrade git if needed
+      # Hack: actions/checkout wants git >= 2.18, rocker 3.5 images have 2.11
+      # Assuming debian stretch because newer images have git >= 2.20 already
+      if: steps.gitversion.outputs.isold == 'TRUE'
+      run: |
+        echo 'deb http://deb.debian.org/debian stretch-backports main' >> /etc/apt/sources.list
+        apt-get update
+        apt-get -t stretch-backports upgrade -y git
```
infotroph, line 35: Sure seems like there ought to be a simpler approach: If git is old, `actions/checkout` falls back to retrieving the source as a tarball instead of cloning the repo, so really all this work is "just" to ensure the `tree-is-clean` action below finds a Git repository to work on.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,169 @@
...
+env:
+  # Would be more usual to set R_LIBS_USER, but R uses R_LIBS first if present
+  # ...and it's always present here, because the rocker/tidyverse base image
+  # checks at R startup time for R_LIBS and R_LIBS_USER, sets both if not found
+  R_LIBS: ~/R/library
+  _R_CHECK_LENGTH_1_CONDITION_: true
+  _R_CHECK_LENGTH_1_LOGIC2_: true
+  # Avoid compilation check warnings that come from the system Makevars, not us
+  # See https://stat.ethz.ch/pipermail/r-package-devel/2019q2/003898.html
+  _R_CHECK_COMPILATION_FLAGS_KNOWN_: -Wformat -Werror=format-security -Wdate-time
```
infotroph, line 14: These flags only affect the operation of R CMD check. Might make sense to set them inside the `check` job rather than for the entire workflow?

.github/workflows/ci.yml
rahul799: Probably one better option will be to have the latest version of git set up inside all the base images.

infotroph: @rahul799 that makes sense to me. @robkooper do you see any issues with updating Git on at `depends` build time, potentially from backports in the case of images built from Stretch?

infotroph:
One more change! Switched test database from the static dump file used on Travis (last updated three or four years ago; takes 4-5 minutes to download and read in) to the `pecan/db:ci` image (stays up-to-date; takes 2-3 minutes to download and read in). A worthy goal for a future cleanup round: Reduce the number of records needed for testing so that the `ci` image can become even smaller yet.

.github/workflows/ci.yml
infotroph: Moved this to its own issue to handle later.

infotroph:
@mdietze @robkooper OK, *now* this is ready for final review and merge. Please squash on merging -- the intermediate commits are all debugs and typo fixes.

PR #2545, opened 2020-03-06
allow non-factor statnames in utils::transformstats

infotroph:
## Description

`PEcAn.utils::transformstats` currently assumes the `statname` column if its input is a factor. This patch allows it to be character instead, if desired. Output keeps the same class it had on input.

Edit: while I was here, also cleaned up code formatting and fixed logic in the untransformed stats warning.

## Motivation and Context

Noticed now because the development version of R [no longer coerces character vectors in `data.frame` to factor](https://developer.r-project.org/Blog/public/2020/02/16/stringsasfactors/), so our unit test started failing. Good catch, unit test!

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


infotroph:
@mdietze @dlebauer I think this is ready to merge -- previous Travis failures are hopefully now resolved.

base/utils/R/transformstats.R
```diff
@@ -1,46 +1,52 @@
...
 ##' Transform misc. statistics to SE
 ##'
-##' Automates transformations of SD, MSE, LSD, 95%CI, HSD, and MSD to conservative estimates of SE.
-##' @name transformstats
-##' @title Transform Stats 
-##' @param data data frame with mean, statistic, n, and statistic name: \code{example data <- data.frame(Y=rep(1,5), stat=rep(1,5), n=rep(4,5), statname=c('SD', 'MSE', 'LSD', 'HSD', 'MSD'))}
-##' @return dataframe with statistics transformed to SE
+##' Automates transformations of SD, MSE, LSD, 95%CI, HSD, and MSD
+##'   to conservative estimates of SE.
+##'
```
dlebauer:
```suggestion
##' Method details and assumptions described in
##' LeBauer 2020 Transforming ANOVA and Regression statistics for Meta-analysis. Authorea. DOI: https://doi.org/10.22541/au.158359749.96662550
```

base/utils/R/transformstats.R
```diff
@@ -1,46 +1,53 @@
...
 ##'
-##' Automates transformations of SD, MSE, LSD, 95%CI, HSD, and MSD to conservative estimates of SE.
-##' @name transformstats
-##' @title Transform Stats 
-##' @param data data frame with mean, statistic, n, and statistic name: \code{example data <- data.frame(Y=rep(1,5), stat=rep(1,5), n=rep(4,5), statname=c('SD', 'MSE', 'LSD', 'HSD', 'MSD'))}
-##' @return dataframe with statistics transformed to SE
+##' Automates transformations of SD, MSE, LSD, 95%CI, HSD, and MSD
+##'   to conservative estimates of SE.
+##' Method details and assumptions described in
+##' LeBauer 2020 Transforming ANOVA and Regression statistics for Meta-analysis. Authorea. DOI: https://doi.org/10.22541/au.158359749.96662550
```
dlebauer:
```suggestion
##' LeBauer 2020 Transforming ANOVA and Regression statistics for Meta-analysis. 
##' Authorea. DOI: https://doi.org/10.22541/au.158359749.96662550
```

dlebauer:
I still don't understand travis well enough to know what to do with the travis failure  - the log says "build exited with status 1" indicating success, but then travis says the build fails https://travis-ci.org/github/PecanProject/pecan/jobs/664465417?utm_medium=notification&utm_source=github_status

infotroph:
@dlebauer the statuses are from bash, so anything not 0 is a failure. 

Looks like this is yet another unrelated dependency failure, this time in the DB tests: apparently  `PEcAn.DB::build_insert_query` is now returning doubled quotes because of something that changed between `glue` 1.3.1 and 1.3.2

ashiklom:
As I mentioned on Slack, I think the easiest fix here is to replace the `glue_sql` statement in `base/db/R/insert_table.R::build_insert_query` with the following:

```r
insert_string <- paste(insert_list, collapse = ",")
result <- glue::glue_sql(
  "INSERT INTO {`table`} ({`colnames(values)`*}) VALUES ",
  .con = .con
)
DBI::SQL(paste0(result, "(", insert_string, ")"))
```

It feel like a less-robust solution because it replaces supposedly robust backend-dependent SQL quoting with hard-coded quoting...but I _think_ it should work with the SQLite backend in the test. 

Also as I mentioned on Slack, the _ideal_ solution here is to simplify the code in that script to just use prepared statements for everything, which should be faster, safer, and easier to implement.

PR #2549, opened 2020-03-19
Adding SIPNET forecast/graphing scripts into pecan 

kzarada:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ X] When possible
## Types of changes

## Checklist:


modules/assim.sequential/inst/NEFI/US_Syv/syv.sh
```diff
@@ -0,0 +1,147 @@
+<?xml version="1.0"?>
```
mdietze, line 1: This filename looks like it should be a shell script for running the Sylvania forecast, but instead it's a pecan.xml. And not the template, a single specific XML with a specific day's drivers. I suspect this is a mistake and the correct file accidentally got overwritten?

modules/assim.sequential/inst/NEFI/US_Syv/syv.xml
```diff
@@ -0,0 +1,78 @@
...
+   <met.end>2020-03-24</met.end>
+  </site>
+  <inputs>
+   <met>
+    <source>NOAA_GEFS_downscale</source>
+    <output>SIPNET</output>
+   </met>
+  </inputs>
+  <start.date>2020-03-08</start.date>
+  <end.date>2020-03-24</end.date>
```
mdietze, line 73: Should start/end dates be hard-coded in the template? Seems like your code should be filling in data-specific info on the fly.

modules/assim.sequential/inst/NEFI/US_Syv/syv.sh
kzarada: Oh good catch, I updated that

modules/assim.sequential/inst/NEFI/US_Syv/syv.xml
kzarada: The start and end dates are changed using the generate.gefs.xml.R file. This also updates the workflow ID and directories

kzarada: I just added in a README file that goes through which files are needed for the forecasts and how to add a new forecast to the system

modules/assim.sequential/inst/NEFI/README.md
```diff
@@ -0,0 +1,28 @@
...
+1. `email_graphs.R` - creates graphs for the email - this is specifically for WCr right now
+2. `email.R` - sends the email out with graphs for today's forecast
+
+
+
+
+### To add forecast and add to dashboard: 
+1. create a flux download for site
+2. create site specific xml 
+3. create site specific shell file to run forecast automatically
```
mdietze, line 26: conceptual question: Once you have a site-specific download function and settings xml, why would the shell script to run the forecast need to be site specific?

modules/assim.sequential/inst/NEFI/README.md
```diff
@@ -0,0 +1,28 @@
...
+2. `email.R` - sends the email out with graphs for today's forecast
+
+
+
+
+### To add forecast and add to dashboard: 
+1. create a flux download for site
+2. create site specific xml 
+3. create site specific shell file to run forecast automatically 
+4. Add site.num, site.abv, outdir, db.num to graph_fluxtowers.R
```
mdietze, line 27: Could this info go into a csv file, or better yet be automatically detected, so that the system is more scalable? If we wanted to jump to all of NEON or FLUXNET or something like that we wouldn't want to have to keep modifying the script itself.

modules/assim.sequential/inst/NEFI/README.md
kzarada: The shell script is site specific right now because it is points to the site's xml file and output path. The sh file runs the R script to update the xml file with the new dates and workflow ids and then runs the pecan workflow with the updated xml file. The site specific flux download function is for graphing since that's when we compare the forecast to the actual data.

modules/assim.sequential/inst/NEFI/README.md
kzarada: I think we could move to that when we start using neon/fluxnet. I currently have it with separate functions because each of the sites have slightly different data configurations and URLs for the data download and so I have to do site specific cleaning to get the NEE/LE

modules/assim.sequential/inst/NEFI/README.md
mdietze: Will pull at this point, but I'm unconvinced that we need site-specific scripts -- based on your description it sounds like you could have a single shell script that takes the xml and outpath as arguments

If the download function is for graphing, where do you download the data for assimilation?

kzarada: Oh yeah, I could do a shell with those as xml and outpath. That would be good for when we scale up. As for the download, these aren't assimilating data right now, they are just the regular SIPNET forecasts but it's the same function for downloading for data assimilation

PR #2557, opened 2020-04-02
fixes-GH-#2556-Dealing with stale issues

### LINKED ISSUE #2556: Dealing with stale issues
## Description

There are **427 open issues**, where many are opened **since 2014** .
Having stale issues hanging around the repository can be a significant detriment. It makes it harder to find the important ones that we want to focus on.

## Proposed Solution

Using a [stale action workflow ](https://github.com/marketplace/actions/close-stale-issues)can be the right choice.

When it runs, it will run the stale action which will look at the issues and pull requests in the repository and find the ones that haven't had any action for a few months. It will then post a message in the issue and add a label indicating that the issue is stale. If that issue stays stale for another week, it's closed.

rahul799:
fix #2556 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,10 +15,10 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This issue is stale because it has been open 365 days with no activity.'
         stale-pr-message: 'This PR is stale because it has been open 365 days with no activity.'
-        stale-issue-label: 'Stale-Issue'
-        stale-pr-label: 'Stale-PR'
-        exempt-pr-label: 'Not-Stale'
-        exempt-issue-label: 'Not-Stale'
+        stale-issue-label: 'Status:Stale'
+        stale-pr-label: 'Status:Stale'
+        exempt-pr-label: 'wishlist'
+        exempt-issue-label: 'Status: in progress'
         days-before-stale: 365
         days-before-close: 100000
         operations-per-run: 25
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,24 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 365 days with no activity.'
+        stale-pr-message: 'This PR is stale because it has been open 365 days with no activity.'
+        stale-issue-label: 'Stale-Issue'
+        stale-pr-label: 'Stale-PR'
+        exempt-pr-label: 'Not-Stale'
+        exempt-issue-label: 'Not-Stale'
+        days-before-stale: 365
+        days-before-close: 100000
+        operations-per-run: 25
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,8 +17,6 @@ jobs:
         stale-pr-message: 'This PR is stale because it has been open 365 days with no activity.'
         stale-issue-label: 'Status:Stale'
         stale-pr-label: 'Status:Stale'
-        exempt-pr-label: 'wishlist'
-        exempt-issue-label: 'Status: in progress'
         days-before-stale: 365
         days-before-close: 100000
         operations-per-run: 25
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,8 +15,8 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This issue is stale because it has been open 365 days with no activity.'
         stale-pr-message: 'This PR is stale because it has been open 365 days with no activity.'
-        stale-issue-label: 'Status:Stale'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 365 days with no activity.'
+        stale-pr-message: 'This PR is stale because it has been open 365 days with no activity.'
+        stale-issue-label: 'Stale-Issue'
+        stale-pr-label: 'Stale-PR'
+        exempt-pr-label: 'Not-Stale'
```
rahul799, line 20: any suggestions for better label name? :smiley:

infotroph: * For a stale label, I'd suggest "Status: Stale" to match our existing categories (we haven't been using these categories very systematically, but that's a separate discussion)
* Probably OK to use the same label for both issues and PRs
* Rather than a specific non-stale label, I'd exempt the categories whose issues are likely to have extra-long lifecycles, e.g. if we already had "wishlist" or "backlog" labels I'd nominate them to be exempt from stale markers. But on reviewing the labels we *do* have, I'm not sure which ones need exemption. Anyone else want to weigh in here?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
```
infotroph, line 5: I hadn't thought to look this up yet, so for my information: Will this run daily in forked repositories as well, or just in the PEcAnProject organization? For this PR it's not likely to matter much (issues are turned off by default in forked repos and I doubt very many people have turned them on), but it could be important to consider when designing other Actions changes in the future.

infotroph:
Do you know whether the label change events will show up in any news feeds? I don't think we need to go too far out of our way to generate notifications if they won't happen automatically, but if I got a ping every time an issue was marked stale that might be a useful push to look at it and possibly decide I had time to fix it.

.github/workflows/stale.yml
rahul799: Yes, the scheduled workflow's do run on forked repos also but we can handle it using
```
jobs:
  stale:
    if: (github.event_name == "schedule" && github.repository == "PecanProject/pecan")
```

rahul799:
Yes, they do show in the notification.

.github/workflows/stale.yml
rahul799: @infotroph Right now, we can specify a [single exempt label](https://github.com/actions/stale/issues/17) :disappointed: only.

.github/workflows/stale.yml
infotroph: Cool, that's easy enough. I don't think we need it for this PR, but let's remember it for future work.

infotroph:
If we can only have one label exempt from flagging, maybe we should back up and ask how useful exemption is -- Will enough issues be unfairly labeled stale that it's a burden to manually unflag them?

My prediction is that the main benefit of the stale flag will happen at the moment it's applied -- hopefully seeing an old issue get flagged will goad us into revisiting and resolving it. If the flag makes us revisit an issue, conclude it's neither resolvable nor closable yet, and manually unflag it, that seems like an acceptable outcome... But then again I'm the guy who didn't like auto-tagging in the first place, so maybe I shouldn't be the one driving this conversation 😀

rahul799:
Should I take it to slack?

infotroph:
Yeah, Slack seems useful to get. Would you be willing to solicit feedback on Slack (it'll be slow-moving there too, especially over weekends) and then summarize the outcome of the discussion here so that it's not lost in the chat logs?

rahul799:
sure

rahul799:
Summary - As per the discussion on slack, we have some pros of using it. 

- A way of resurfacing issues that haven’t been touched in a while and reminding/embarrassing us into dealing with them… or at least touching them again and, therefore, resetting the timer.
- Probably somebody realizes that the issue is already fixed & can close it.

Conclusion - We are going to use this action.

I will like to add that we have one more feature that is [operation-per-run](https://github.com/actions/stale/blob/0d6a8e531449ecb97c9b6046b6b5d8eeac8b5115/action.yml#L29) which I have kept 25. is it okay?

infotroph:
Hah, the operations per run was going to be my last question! Am I right this means if there are say 35 issues eligible to be marked stale today, it will mark 25 of them today and then handle the rest on its next run tomorrow? If so then yes, 25 seems fine -- hopefully after the initial backlog is cleared, we'll never come close to the limit again.

rahul799:
Yes, you are correct.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 365 days with no activity.'
+        stale-pr-message: 'This PR is stale because it has been open 365 days with no activity.'
+        stale-issue-label: 'Status:Stale'
+        stale-pr-label: 'Status:Stale'
```
infotroph, line 19:
```suggestion
        stale-issue-label: 'Status: Stale'
        stale-pr-label: 'Status: Stale'
```

Issue #853, opened 2016-05-24
Standardize formatting using formatR package


dlebauer:
based on comment from @bpbond this morning

For package maintainers: next time you submit a pull request, run the `tidy_source()` function from the formatR package. Here is an overview: http://yihui.name/formatR/

I'd suggest everyone use these arguments:

```r
tidy_dir("./modules/mymodule/", arrow=TRUE, width.cutoff = 80, indent = 2)
```

The one thing that I don't like about this function is that when it breaks lines, it does not do so in the same way as Rstudio does by default, specifically, for this line:

```r
old.priors <- db.query(paste("select prior_id from pfts_priors where pft_id =", parent.pft.id, ";"), con = con)
```

formatR apparently uses the argument to `indent` (here, `indent = 2`) to output: 

```r
old.priors <- db.query(paste("select prior_id from pfts_priors where pft_id =", 
    parent.pft.id, ";"), con = con)
```

while Rstudio indents to the level of the function to which arguments are being passed (in this case, to `paste`: 

```r
  old.priors <- db.query(paste("select prior_id from pfts_priors where pft_id =", 
                               parent.pft.id, ";"), con = con)
```

the Rstudio way makes it easier to see that the arguments on line 2 are arguments to the `paste` functions. 

@yihui is there a way around this? I checked issues in the github repo, but couldn't find this specific one.

serbinsh:
@dlebauer I have been experimenting but it doesn't seem to do anything for me when i run tidy_dir or tidy_source without recursive = TRUE

e.g.

```
tidy_dir("~/Data/GitHub/R-Canopy_Photosynthesis/", recursive = TRUE, arrow=TRUE, width.cutoff = 80, indent = 2)
```

Just an FYI that you may need to use that argument

dlebauer:
you are correct. I changed the directory while copy-pasting ...

PR #2559, opened 2020-04-06
fixes - #2558 implementing styler

### LINKED ISSUE #2558: Implementing styler
## Description
[styler](https://cran.r-project.org/web/packages/styler/index.html) provides non-invasive pretty-printing of R source code while adhering to the tidyverse formatting rules.

rahul799:
## Description

fix #2558 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


robkooper:
I think this is a good thing, but ....

Couple of questions:
- what are the styler recommendations based on? Does this match what we have in the documentation
- can we make this part of .travis/github actions? so we don't have to run this every so often?

dlebauer:
In the past - around 2016 I think when we did a lot of reformatting we used the [formatR package](https://yihui.org/formatr/#2-reformat-r-code) see https://github.com/PecanProject/pecan/issues/853#issuecomment-540622974. Would be great to have that as a github action

rahul799:
> Couple of questions:
> 
> * what are the styler recommendations based on? Does this match what we have in the documentation

I checked the documentation but didn't get anything :disappointed: ,what I have used here is the official tidyverse style guide. Styler offers [customizations](https://github.com/r-lib/styler#configuration) also, so if in case we want any changes, we can go for it.

> * can we make this part of GitHub actions? So we don't have to run this every so often?

Yes, we can. I will like to add that; this feature is part of my GSoC proposal also.

rahul799:
@dlebauer I think that styler is a better choice as it is suggested by tidyverse [official docs](https://style.tidyverse.org/). I have also written about the implementation of styler with GitHub actions in my GSoC proposal.

rahul799:
Travis is falling giving error
```
R check of base/db returned new problems:
checking Rd \usage sections ... WARNING: Undocumented arguments in documentation object 'query_pfts' 'modeltype' 'strict'
checking Rd line widths ... NOTE: Rd file 'insert.format.vars.Rd': \examples lines wider than 100 characters:    insert.format.vars(con = bety$con, format_name = "LTER-HFR-103", mimetype_id = 1090, notes = "NPP from Harvard Forest.", header = FALSE ... [TRUNCATED]
Execution halted
Makefile:139: recipe for target '.check/base/db' failed
make: *** [.check/base/db] Error 1
```
I am not sure how to fix it & need some help.

infotroph:
@rahul799 Apparently styler managed to unmask two issues, both of which are only indirectly its fault:

* "lines wider than..." was present in the previous check output, but styler changed some whitespace and the comparison between old and new outputs works by exact string comparison. 
* "undocumented arguments..." is new because styler stripped a trailing space that *ought* to be harmless, but Roxygen (incorrectly, I'd argue) treats `#' @param modeltype ` (with trailing whitespace) as a parameter with a valid-but-empty description but `#' @param modeltype` (no trailing whitespace) as a parameter with its description missing.

Cleanest thing will be to fix these in a separate PR and merge this downstream of it. The second  error is apparently my fault, so I'll tackle both later today.

rahul799:
@infotroph when I am running the check command, it's still complaining about the undocumented arguments. Does PR #2564 didn't fix that?

infotroph:
@rahul799 You'll need to merge the upstream changes into this PR before they show up in the CI result, but it has some merge conflicts that will need care to resolve correctly. Ping me on Slack if you want help working through them.

Backing up, though: These CI issues actually illustrate some major challenges of project-wide code-cleanup PRs like this: 

* By their nature they're prone to merge conflicts (because they alter many lines across many people's code), which are both annoying and a common vector for bug creation when they're resolved incorrectly. 

* No matter how robust the styling tool we use, in a large enough codebase it *will* manage to break something. Once the set of changed files becomes too large to review carefully, we're not likely to notice what broke until well after it's merged. For comparison, when I did a limited cleanup of one package in #2483 I wound up manually reviewing every change line by line, and I ultimately chose to leave some things unchanged when `lintr`'s suggested fixes were unworkable for one reason or another. 

If I'm understanding right without having looked at all the changed files, this PR commits the results from running `styler` once, but it doesn't yet automate the process of styling future contributions, correct? Perhaps instead of doing one big cleanup now and then implementing automation to maintain it later, it might be worth developing the automation first and maybe deploying it one package at a time until it reaches full coverage.

rahul799:
@infotroph Thanks for the remarks. I think that I must close the PR for now & will start working on automation first.

PR #2563, opened 2020-04-08
Remove maintenance scripts that now live in bety

### LINKED ISSUE #2314: Move BETYdb reindex and vacuum scripts to bety project
### Bug Description

The BETYdb database vacuum and reindexing scripts are in the wrong project

## Expected behavior

Should be able to run these scripts from the bety docker container

infotroph:
## Description

Fixes #2314. Both scripts already exist in the bety project, so nothing to do on that end.

@robkooper @Chris-Schnaufer While we're at it, are there other scripts we should remove in this PR? `update.mysql.sh` and `update.psql.sh` look like possible candidates.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


robkooper:
Some scripts that can be removed:
- check.dependencies.sh
- install.dependencies.R (replaced by Makefile and generate_dependencies.R)
- update.mysql.sh
- update.psql.sh
- dump.pgsql.sh
- setversion.sh

Maybe delete:
- install_deps.R

infotroph:
@robkooper looks like install_deps.R was part of the first round of Docker implementation. Is there a chance it'll be useful there in the future, or have you settled on a different approach?

infotroph:
On a few moments' further thought, I think install_deps.R can definitely go -- it comes close enough to doing the same thing the Makefile does using `remotes::install_deps` does that I don't see a need to maintain a separate implementation.

PR #2568, opened 2020-04-13
new shiny app, replaces previous visualization

robkooper:
This will replace the previous visualization for the sync status page. This is now a shiny app.

There are 2 buttons, refresh hosts, which will load new hosts from database, and refresh sync which will load the sync logs from all the hosts and show the linkages.

green : less than 6 hours ago synced
yellow : less than 24 hours ago synced
red : more than 24 hours ago synced

You can see this running at http://pecan-dev.ncsa.illinois.edu/dbsync/

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


docker-compose.yml
```diff
@@ -297,6 +297,23 @@ services:
     volumes:
       - pecan:/data
 
+# ----------------------------------------------------------------------
+# Shiny Apps
+# ----------------------------------------------------------------------
+  # PEcAn DB Sync visualization
+  maespa:
```
infotroph, line 304: this naming seems odd — why maespa here?

shiny/dbsync/app.R
```diff
@@ -0,0 +1,242 @@
...
+        if (what == "color") {
+            result <- c("black")
+        } else {
+            result <- c(servers[1, what], servers[1, what], NA)
+        }
+    }
+    return(result)
+}
+
+# Define UI for application that draws a histogram
```
infotroph, line 159: possible left over comment from a previous app?

docker-compose.yml
robkooper: bad copy/paste

shiny/dbsync/app.R
robkooper: no, problem is that if there are no lines the system crashes (or I need to put an if around the code below). Figured a solution is to draw a 0 pixel line if no other lines are drawn.

infotroph:
Looked again when I was more awake and I'm seeing incomplete lines to BNL (screenshot attached) -- are these a length calculation issue or 
is line coloring split inbound/outbound?
Screen Shot 2020-04-17 at 8 09 31 AM

mdietze:
@infotroph correct, the partial line segments split the inbound/outbound relationship. In this case everyone's sync with BNL is old, while the missing segments near BNL reflect that it is offline. It's also possible for lines to change color at the midpoint to reflect differences in sync freshness (e.g. I may pull from Ankur every hour, but he might pull from me once a day, and thus his line would often be yellow when mine is green)

infotroph:
I think this is good to god for initial merge and the legends/interpretation can be added separately.

PR #2569, opened 2020-04-14
Styler workflow

rahul799:
## Description
This workflow will run the **styler package** on files that are changed in PR & are of type **.R & .Rmd** . After running the styler, If there are any formatting changes, then it will **automatically commit** the changes to the same PR.  To execute this workflow, we need to comment **/style** in the PR discussion.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


rahul799:
@infotroph has made a [sandbox repository](https://github.com/infotroph/pecan-pr-sandbox) to test this & we have done [some test](https://github.com/infotroph/pecan-pr-sandbox/pulls) which were successful, but it would be good if more tests can be done.

Possible tests as suggested by @infotroph 

1. Changes that do vs don’t change Roxygen documentation
2. PRs with multiple rounds of human commits and re-invocations of the styler bot (I don’t expect any problems, just worth checking)
3. Different mixes of R vs other files, including confirming that it does nothing when no R files are changed

Apart from that, he suggested some other features like

1. Having the bot also post a reply to the /style comment that says something like “OK, done! Styler changed 12 lines in 3 files”? 
2. What if the bot were able to post its changes as review suggestions so that the user could accept or reject them individually?

looking for other's opinions.

rahul799:
@robkooper @dlebauer as per our discussion on #2559 I have made this workflow. Please have a look.

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,101 @@
...
+        run: |
+          cut -d / -f 1-2 artifacts/new2.txt | tr -d '"' > new.txt
+          cat new.txt
+          sort new.txt | uniq > uniq.txt
+          cat uniq.txt      
+          for i in $(cat uniq.txt); do make .doc/${i}; done
+      - name: commit
+        run: |
+          git config --global user.email "rahulagrawal799110@gmail.com"
+          git config --global user.name "Rahul Agrawal"
```
rahul799, line 88: I don't know why but when I am using the container, git is complaining about the **config --global**. That's why I have added the following details.

rahul799:
@infotroph I think It's ready for the review.

rahul799:
Mukul, It will be good if you can test this workflow [here](https://github.com/infotroph/pecan-pr-sandbox) & let me know if it's not working as expected.

MukulMaheshwari:
Ok let me have a look.

.github/workflows/styler-actions.yml
infotroph: The complaint is just because Git wants every commit to have someone's name associated with it, and the container starts out with no user configured. But you'll probably want to attribute these commits to a bot user rather than to yourself:

```suggestion
          git config --global user.email "pecan_bot@example.com"
          git config --global user.name "PEcAn stylebot"
```

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,93 @@
...
+          Rscript -e 'install.packages("styler")'
+          Rscript -e 'install.packages("devtools")'
+          Rscript -e 'devtools::install_version("roxygen2", version = "7.0.2", repos = "http://cran.us.r-project.org")'
+      - name: string operations
+        run: |
+          echo '${{ steps.file_changes.outputs.files_modified}}' > names.txt
+          cat names.txt | tr -d '[]' > new.txt
+          text=$(cat new.txt)
+          IFS=',' read -ra ids <<< "$text"
+          for i in "${ids[@]}"; do if [[ "$i" == *.R\" || "$i" == *.Rmd\" ]]; then echo "$i" >> new2.txt; fi; done
```
infotroph, line 31: informative filenames could help make it clearer what's happening here. Maybe new.txt could become "changed_files.txt" and new2.txt could be "files_to_style.txt"? If changing here, will need to change other instances below to match.

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,93 @@
...
+          v='cat(numeric_version("'${v}'") < "2.18")'
+          echo "##[set-output name=isold;]$(Rscript -e "${v}")"
+      - name: upgrade git if needed
+        # Hack: actions/checkout wants git >= 2.18, rocker 3.5 images have 2.11
+        # Assuming debian stretch because newer images have git >= 2.20 already
+        if: steps.gitversion.outputs.isold == 'TRUE'
+        run: |
+          echo 'deb http://deb.debian.org/debian stretch-backports main' >> /etc/apt/sources.list
+          apt-get update
+          apt-get -t stretch-backports upgrade -y git
```
infotroph, line 67:
```suggestion
 ```

this was a workaround for #2578, yes?

File: .github/workflows/styler-actions.yml [hunk 4]
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          name: artifacts
+      - name : make
+        shell: bash
+        run: |
+          cut -d / -f 1-2 artifacts/new2.txt | tr -d '"' > new.txt
+          cat new.txt
+          sort new.txt | uniq > uniq.txt
+          cat uniq.txt      
+          for i in $(cat uniq.txt); do make .doc/${i}; done
```
infotroph, line 84: as above, descriptive filenames could remove a lot of the mystery here, e.g.

```suggestion
          cut -d / -f 1-2 artifacts/new2.txt | tr -d '"' > changed_dirs.txt
          cat changed_dirs.txt
          sort changed_dirs.txt | uniq > needs_documenting.txt
          cat needs_documenting.txt      
          for i in $(cat needs_documenting.txt); do make .doc/${i}; done
```

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,93 @@
+on:
+  issue_comment:
+    types: [created]
+name: Commands
+jobs:
+  style:
+    if: startsWith(github.event.comment.body, '/style')
+    name: style
+    runs-on: macOS-latest
```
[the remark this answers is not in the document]
rahul799: Yes, I have found that using macOS makes this workflow faster. That's why I used mac here.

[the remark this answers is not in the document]
infotroph: Huh! I was expecting the other direction. Good to know!

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,93 @@
...
+          Rscript -e 'devtools::install_version("roxygen2", version = "7.0.2", repos = "http://cran.us.r-project.org")'
+      - name: string operations
+        run: |
+          echo '${{ steps.file_changes.outputs.files_modified}}' > names.txt
+          cat names.txt | tr -d '[]' > new.txt
+          text=$(cat new.txt)
+          IFS=',' read -ra ids <<< "$text"
+          for i in "${ids[@]}"; do if [[ "$i" == *.R\" || "$i" == *.Rmd\" ]]; then echo "$i" >> new2.txt; fi; done
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v1
```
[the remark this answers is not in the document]
infotroph: Correct: I see that artifact deletion exists now and is easy to do. But I think the point about fit still holds.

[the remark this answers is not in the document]
rahul799: Yes, I agree `actions/cache` will be a better option. As we won't need those files outside the workflow.

.github/workflows/styler-actions.yml
rahul799: Thankyou

.github/workflows/styler-actions.yml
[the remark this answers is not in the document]
rahul799: @infotroph I am thinking to open another issue for that. I think we can handle this later on.

rahul799:
As we discussed on the slack to use the package cache from the CI workflow, let's tackle that in a separate PR? I will open a new issue for that.

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,78 @@
...
+          cut -d / -f 1-2 artifacts/files_to_style.txt | tr -d '"' > changed_dirs.txt
+          cat changed_dirs.txt
+          sort changed_dirs.txt | uniq > needs_documenting.txt
+          cat needs_documenting.txt      
+          for i in $(cat needs_documenting.txt); do make .doc/${i}; done
+      - name: commit
+        run: |
+          git config --global user.email "pecan_bot@example.com"
+          git config --global user.name "PEcAn stylebot"
+          git add \*.Rd
```
infotroph, line 74: Roxygen runs also have the potential to change the NAMESPACE and (hopefully rarely; only when the Roxygen version changes) DESCRIPTION files. I propose we handle those in a (near)future PR rather than this one, but the incentive is that should make it possible (if not the most convenient) to entirely skip Roxygen on your own machine.

infotroph:
This all looks good to me now. The last step we need before merging is confirmation that the rest of the team likes the behavior as implemented.

@rahul799 would you be willing to go through the list in https://github.com/PecanProject/pecan/pull/2569#issuecomment-613412143 and collect some links to PRs in the sandbox repo that demonstrate each of them? I predict we'll get more comments from other folks if they can follow a link and and comment on what they see rather than needing to create their own PRs.

rahul799:
sure

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          Rscript -e 'install.packages(c("styler", "devtools"), repos = "cloud.r-project.org")'
+          Rscript -e 'devtools::install_version("roxygen2", version = "7.0.2", repos = "http://cran.us.r-project.org")'
+      - name: string operations
+        run: |
+          echo '${{ steps.file_changes.outputs.files_modified}}' > names.txt
+          cat names.txt | tr -d '[]' > changed_files.txt
+          text=$(cat changed_files.txt)
+          IFS=',' read -ra ids <<< "$text"
+          for i in "${ids[@]}"; do if [[ "$i" == *.R\" || "$i" == *.Rmd\" ]]; then echo "$i" >> files_to_style.txt; fi; done
```
infotroph, line 30:
```suggestion
          for i in "${ids[@]}"; do if [ "$i" == *.R\" || "$i" == *.Rmd\" ]; then echo "$i" >> files_to_style.txt; fi; done
```

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          name: artifacts
+          path: files_to_style.txt
+      - name: Style
+        run: for i in $(cat files_to_style.txt); do Rscript -e "styler::style_file("$i")"; done
+      - name: commit
+        run: |
+          git add \*.R
+          git add \*.Rmd
+          if [[ $(git diff --name-only --cached) != "" ]]; then  git commit -m 'automated syle update' ; fi
```
infotroph, line 42:
```suggestion
          if [ "$(git diff --name-only --cached)" != "" ]; then  git commit -m 'automated syle update' ; fi
```

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,78 @@
...
+          cat changed_dirs.txt
+          sort changed_dirs.txt | uniq > needs_documenting.txt
+          cat needs_documenting.txt      
+          for i in $(cat needs_documenting.txt); do make .doc/${i}; done
+      - name: commit
+        run: |
+          git config --global user.email "pecan_bot@example.com"
+          git config --global user.name "PEcAn stylebot"
+          git add \*.Rd
+          if [[ $(git diff --name-only --cached) != "" ]]; then  git commit -m 'automated documentation update' ; fi
```
infotroph, line 75:
```suggestion
          if [ "$(git diff --name-only --cached)" != "" ]; then  git commit -m 'automated documentation update' ; fi
```

rahul799:
Thankyou @infotroph.

.github/workflows/styler-actions.yml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          Rscript -e 'install.packages(c("styler", "devtools"), repos = "cloud.r-project.org")'
+          Rscript -e 'devtools::install_version("roxygen2", version = "7.0.2", repos = "http://cran.us.r-project.org")'
+      - name: string operations
+        run: |
+          echo '${{ steps.file_changes.outputs.files_modified}}' > names.txt
+          cat names.txt | tr -d '[]' > changed_files.txt
+          text=$(cat changed_files.txt)
+          IFS=',' read -ra ids <<< "$text"
+          for i in "${ids[@]}"; do if [ "$i" == *.R\" || "$i" == *.Rmd\" ]; then echo "$i" >> files_to_style.txt; fi; done
```
infotroph, line 30:
```suggestion
        shell: bash
        run: |
          echo '${{ steps.file_changes.outputs.files_modified}}' > names.txt
          cat names.txt | tr -d '[]' > changed_files.txt
          text=$(cat changed_files.txt)
          IFS=',' read -ra ids <<< "$text"
          for i in "${ids[@]}"; do if [[ "$i" == *.R\" || "$i" == *.Rmd\" ]]; then echo "$i" >> files_to_style.txt; fi; done
```

My mistake when I suggested single-brackets here was forgetting that || doesn't work inside them. We could go fully down the POSIX rabbit hole and write `... do test "$i" = *.R\" -o "$i" = *.Rmd\" && echo "$i" >> ...`, but I think the bash version is way more readable.

infotroph:
More tests of behavior with non-R file changes are visible in https://github.com/infotroph/pecan-pr-sandbox/pull/17, along with some overly-chatty commentary on what's being done at each step.

infotroph:
I think this is good to go! We can refine further in future PRs.

Folks with Opinions About Formatting (@mdietze @robkooper @ashiklom and others): Please have a look at the styler bot in action (links upthread) and see what you think. Any behaviors that need to be changed before we merge this?

PR #2574, opened 2020-04-20
install glpk system library

infotroph:
## Description

BayesianTools imports DHARMa imports spaMM imports ROI.plugin.gplk imports Rglpk, which needs the GLPK system library. Looks like we haven't seen this before because spaMM just updated and added the ROI dependency onto the chain.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


ashiklom:
Ugh. `BayesianTools` is very useful, but it's got a bazillion difficult-to-install dependencies. I'm very tempted to replace it with our own implementation of the Differential Evolution algorithm.

Thanks for catching this dependency --- I think we can pull it in for now, but if we can come up with a homegrown solution later, we can revisit.

mdietze:
I appreciate the frustration @ashiklom, and I think reducing dependencies is a great suggestion to it's developers (e.g. @ florianhartig), but I really think `BayesianTools` provides a good bit more than just the DE algorithm

florianhartig:
Hi guys, thanks for tracking down the error. I will investigate this further here https://github.com/florianhartig/DHARMa/issues/166

At the moment, it seems the only fix (unless spaMM is rolling back) would be to force the install of GLPK / Rglpk on Travis. What I don't understand is why this is no problem locally, or is it for Linux / Ubuntu users? 

About the dependencies of BT - did you have problems with that before? The dependencies don't seem excessive to me, and I didn't have any Travis problems before (as I'm testing with Travis myself). 

It is true though that the DHARMa dependency could maybe be moved from imports to suggests, because it is only used in some plots, an I don't have the feeling that those are very popular, so maybe it's better to concentrate on the core functionality.

florianhartig:
OK, I have looked at this, you can fix this by installing the packages in the yml file, see here https://github.com/florianhartig/DHARMa/issues/166

infotroph:
@florianhartig Thanks for the quick investigation. Do you know offhand what `glpk-utils` was needed for in the DHARMa tests? Our build [seems to pass](https://travis-ci.org/github/PecanProject/pecan/builds/677300075) with only `libglpk-dev`, but I'll feel silly if it breaks again because of that tomorrow :)

florianhartig:
I added it because someone said somewhere that fixed it for their Ubuntu install. If it works without glpk-utils I guess it's fine and I will also remove it in my settings.

ashiklom:
Thanks for the note, @florianhartig!

> About the dependencies of BT - did you have problems with that before?

I haven't had issues myself on MacOS or CentOS, but a few months ago, some of my former PNNL colleagues were unable to get `BayesianTools` installed on Windows and the PNNL HPC because of some library dependency issues. Unfortunately, they didn't retain notes about what the issues were.

florianhartig:
Hi Alexey,

I'm generally trying to minimize dependencies, so I think the current functionality of BT really requires the imported packages to do the calculations / plots on top of the posterior. As I said, maybe DHARMa could be moved to suggest, this would probably reduce a few downstream dependencies. 

Of course, if you only wanted to run the samplers, you could do this with a lot less packages. In principle, it would be possible to have a core package with the samplers (e.g. BTCore), which would then be imported by the normal BT package. However, that would be a lot of work, so I wouldn't want to do this unless I see that there is an actual demand for that.

PR #2591, opened 2020-04-30
add dockerfile, model_info.json to template (fixes #2567)

### LINKED ISSUE #2567: add model_info.json to models/template
* We now expect all models to have a `model_info.json` going forward, correct? If so, we should give an example in `models/template`
* If it's standard for `model_info.json` to live in the package root (rather than `inst/` like `register.*.xml` did), will also need to list it in `.Rbuildignore` to prevent "non-standard files /directories found at top level" note at check time

(Moved from https://github.com/PecanProject/pecan/pull/2566#issuecomment-612628000)

robkooper:
Added an example of a Dockerfile and model_info.json to the template

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


models/template/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+ARG IMAGE_VERSION="latest"
+
+# ----------------------------------------------------------------------
+# BUILD MODEL BINARY
+# ----------------------------------------------------------------------
+FROM debian:stretch as model-binary
+
+# Some variables that can be used to set control the docker build
+ARG MODEL_VERSION=git
+
```
istfer, line 17: would we want an `ENV` for specifying compiler etc?

robkooper: I think people should modify the Dockerfile. Most models will need different libraries, compilers etc.

robkooper: I updated the README.md to hopefully clarify some of this.

PR #2595, opened 2020-05-04
Build develop image

### LINKED ISSUE #2578: update depends image to git >= 2.18 to simplify .github/workflows/ci.yml
Bumping [this thread](https://github.com/PecanProject/pecan/pull/2544#discussion_r412841079) from #2544 to be resolved in a separate PR. Context: the ["check git version" and "update git if needed"](https://github.com/PecanProject/pecan/pull/2544/files#diff-e9f950f17198d3d5e3122a44230a09b9R17-R30) steps of the Actions CI workflow

@infotroph:
> Sure seems like there ought to be a simpler approach: If git is old, `actions/checkout` falls back to retrieving the source as a tarball instead of cloning the repo, so really all this work is "just" to ensure the `tree-is-clean` action below finds a Git repository to work on.
>
> I'm open to better suggestions, preferably ones that will keep working once we expand the build matrix to include multiple base images with potentially differing OS versions.

@rahul799: 
> Probably one better option will be to have the latest version of git set up inside all the base images.

@infotroph:
> @rahul799 that makes sense to me. @robkooper do you see any issues with updating Git at `depends` build time, potentially from backports in the case of images built from Stretch?

robkooper:
create a depends image for R 3.5 and 3.6, should update when 4.0 is ready
make sure latest git is installed on 3.5 fixes #2578 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


.github/workflows/depends.yml
```diff
@@ -0,0 +1,63 @@
...
+  depends:
+    #if: github.repository == 'PecanProject/pecan'
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        R:
+          # 3.4 does not work
```
infotroph, line 22: I believe you, but am curious why not -- I was building PEcAn locally with R 3.4 until pretty recently. Too many dependencies requiring R >= 3.5?

robkooper: yeah dependencies, and instead of chasing down the dependencies, I decided to just skip it.

PR #2599, opened 2020-05-06
Delete ED2IN.git

dlebauer:
this no longer works with the master branch of the https://github.com/EDmodel/ED2 keeping this unversioned ED2IN around will only cause trouble.


dlebauer:
From @mdietze 

> This is definitely going to break things for people, which may be painful for some, but it needs to happen. The ED2IN.git was never supposed to have persisted beyond the initial migration of ED2 to github and the lack of ED2 versioning is really an abuse of all the work we put into model version tracking in PEcAn. So yes, I support this PR

PR #2601, opened 2020-05-06
add a true docker quickstart

dlebauer:
previously the docker-quickstart link pointed to this file, with the first section

```
## The PEcAn docker install process in detail {#docker-quickstart} 
```


book_source/03_topical_pages/94_docker/02_quickstart.Rmd
```diff
@@ -1,4 +1,37 @@
...
+
+```bash
+git clone git@github.com/pecanproject/pecan
+cd pecan
+
+# start database
+docker-compose -p pecan up -d postgres
+
+# add example data (first time only)
+docker-compose run --rm bety initialize
```
robkooper, line 11: this has changed in the bety documentation and should be reflected here as well.

robkooper: we now leverage of the pecan/db container to speed up the process of loading data. No need to initialize using bety container.

```
docker run --rm --network pecan_pecan pecan/db
```

book_source/03_topical_pages/94_docker/02_quickstart.Rmd
```diff
@@ -1,4 +1,37 @@
...
+docker run -ti --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
+
+# start PEcAn
+docker-compose -p pecan up -d
+
+# run a model
+curl -v -X POST \
+    -F 'hostname=docker' \
+    -F 'modelid=5000000002' \
```
dlebauer, line 20:
```suggestion
    -F 'modelid=14' \ # ED v2.2.0
```

KristinaRiemer: @dlebauer I don't see any model with an `id = 14` (e.g., http://welsch.cyverse.org:8000/bety/models/14)

book_source/03_topical_pages/94_docker/02_quickstart.Rmd
dlebauer: @robkooper this be a docker-compose command?

robkooper: no, think of docker-compose as services taht should be running continously. This command is a one time initialization of the data. Once you have done it once, you can use `docker-compose run --rm bety sync` to synchronize the data.

julianpistorius: On a related note, why doesn't the 'Docker Quickstart' use `docker.sh`?

robkooper: docker.sh is the script to build the docker containers.

julianpistorius: I didn't see that. Thank you for clarifying.

robkooper: That should be somewhere, however I think the quickstart really should be a "use existing containers" start. Not the and now lets build the containers before you get started start.

So I like how this is done right now by @dlebauer

book_source/03_topical_pages/94_docker/02_quickstart.Rmd
robkooper: only reason why i had `5000000002` is this is sipnet, if you change the model, you need to make sure to change the pft etc as well. Finally sipnet runs quickly so this is a nice get started model.

istfer: I agree with @robkooper  that this better stays sipnet in the main documentation

mdietze:
@robkooper it looks like your request changes is blocking this PR, but based on the discussion thread it seems like the requests were addressed.

PR #2605, opened 2020-05-08
setting the default values to match docker.sh so this will build

kyclark:
I was unable to "docker build" because the MODEL_VERSION and BINARY_VERSION had outdated values. I set these to the values that are being passed in by "docker.sh," and now I can build this image. This may not be important, but I was very confused for a while why I couldn't build the image with the 2.2.0 for testing until I was pointed to the docker.sh. So just setting the default values to what's useful now.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ x] When possible
## Types of changes

## Checklist:

- [x ] All new and existing tests passed.


models/ed/Dockerfile
```diff
@@ -7,8 +7,8 @@ ARG IMAGE_VERSION="latest"
 FROM debian:stretch as model-binary
 
 # Some variables that can be used to set control the docker build
-ARG MODEL_VERSION=git
-ARG BINARY_VERSION=2.1
+ARG MODEL_VERSION="2.2.0"
+ARG BINARY_VERSION="2.2"
```
robkooper, line 11: up for vote, but I'd say lets leave the model_version to be git, and binary version to be 2.2. This will allow people to build the latest version. Both of these are set as build args passed to docker:
```
docker build --tag pecan/model-ed2-2.2.0:develop --build-arg MODEL_VERSION=2.2.0 --build-arg IMAGE_VERSION=develop --build-arg BINARY_VERSION=2.2 models/ed
```

robkooper: Main reason for not setting model_version to be `2.2.0` is that we should not have to modify our Dockerfile every time a new version of ED is released.

dlebauer: @robkooper how could this be changed? FWIW ... it looks like only two numbered versions of ED2 have been released since about 2008.

robkooper: ED changed the postfix of the model from ed2.1 to ed2.2 (`BINARY_VERSION`). The `MODEL_VERSION` is the tag that is checked out in the code.

You can change these when compiling with the `build-arg` flag.

mdietze: I prefer the proposed change over Rob's first alternative. i.e. binary should be 2.2 because that's how the executable is tagged (yes, @dlebauer two official releases since 2008 sounds right), but I disagree with letting the `git` model version sneak back in.

mdietze:
Travis shows this as passing, but for some reason Github isn't acknowledging it, so I'm going to go ahead and pull

PR #2606, opened 2020-05-11
Change order of arguments in call_MODIS()

### LINKED ISSUE #2519: call_MODIS argument cleanup
I won't hold up the PR for this, but two strong suggestions:

* Changing the arguments to an exported function is a breaking change, so it should have a CHANGELOG entry. Breaking changes aren't always bad -- I think most users will prefer this version over the old one! -- but when we change interfaces we want to do it cautiously and document the changes conspicuously.

* When a function has some arguments with default values and some without, it's usually best to put the arguments without default values should be at the beginning of the argument list and the ones with defaults (even if they're NULL) afterwards. Since you're changing argument order in this PR anyway, I encourage doing that now to avoid needing another round of breaking changes to do it later.

_Originally posted by @infotroph in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjA4Njc2ODEzOnYy/pull_request_review_threads/discussion_

Bumped this to a new issue to avoid holding up #2338. 

Additional thought: If I'm following right, the outfolder argument is slightly magic in that when it's a path the function both writes to disk *and* puts the downloaded data in its return value, whereas when outfolder = NULL the function writes no files and *only* returns the values. This  behavior seems reasonable, but should be documented.

ayushprd:
## Description
I have changed the order of arguments in `call_MODIS()`, explained the behaviour of `outdir `argument and updated the changelog.
## Motivation and Context
Fixes #2519 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - No longer writing an arbitrary num for each PFT, this was breaking ED runs potentially.
 - The pecan/data container has no longer hardcoded path for postgres
 - PEcAn.JULES: Removed dependency on `ncdf4.helpers` package, which has been removed from CRAN (#2511).
-- data.remote: Arguments to the function `call_MODIS()` have been changed (#2338).  
+- data.remote: Arguments to the function `call_MODIS()` have been changed (issue #2519).  
 
 ### Added
 - model_info.json and Dockerfile to template (#2567)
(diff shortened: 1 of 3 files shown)


istfer:
Thanks for tackling this @ayushprd, I think you need to run roxygenize (roxygen version 7.0.2) to update the function arguments in the Rd files

mdietze:
Also need to make sure that any downstream uses of the function need to be updated

infotroph:
It looks like Travis somehow... ran two copies of the job on the same VM? Or possibly ran one copy but echoed two copies of everything to the log. I've restarted the build and will see if it behanves any better this round.

PR #2609, opened 2020-05-14
[WIP] ideas for generating docker images

kyclark:
## Description

These are just basic sketches for a possible improvement over "docker.sh." For one, I think the configuration of the images could be handled better if they were kept separately, but it can be easy to mess up YAML and JSON, so I propose using DHall to maintain the Docker configurations. Next I don't want to build images that already exist, so I've written a Python script that can use the "docker" module to find the tags of existing images and skip building those. We can, of course, have a --force option to make them all.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


robkooper:
Not sure what did adds. The images for Basgra and BioCro do need to be build since it adds the model_info.json to /work folder

Other problem, this add python and DHall as requirement

Also in general it is best to add the requirements.txt and use virtual environment in python.

julianpistorius:
@robkooper you said:

> The images for Basgra and BioCro do need to be build since it adds the model_info.json to /work folder

Can this be solved by bind-mounting a local, modified version of `model_info.json` at runtime? At least for development purposes?

robkooper:
yup, that is possible. depends also what you are trying to do. Check out development guide pull request as well #2610

base/logger/DESCRIPTION
```diff
@@ -11,5 +11,5 @@ Suggests: testthat
 License: BSD_3_clause + file LICENSE
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 7.0.2
+RoxygenNote: 7.1.0
```
infotroph, line 14: Please keep Roxygen at version 7.0.2 for now -- I know it's annoying to keep the version behind on a dev machine, but when we bump the version we want to do it in all packages at once.

infotroph:
Closing as out of scope—we’d like to avoid adding more dependencies and have mostly addressed the build issue through other means.

PR #2611, opened 2020-05-14
Fixed not triggering of "All values bad" in call_MODIS

### LINKED ISSUE #2520: call_MODIS "all rows bad" warning will never fire
Can you talk me through what this test does? It looks to me as if it will always evaluate to TRUE even when no rows are good, because `is.null()` on a length 0 vector returns FALSE.

_Originally posted by @infotroph in https://github.com/PecanProject/pecan/pull/2338_

Bumped this to a new issue to avoid holding up #2338, refers to [`if (length(good) > 0 || !(is.null(good))) {` in call_MODIS.R](https://github.com/PecanProject/pecan/pull/2338/files#diff-47227c849cd1830d9d5019a5a89bb41fR247)

Consider a simplified example case:

```
output = data.frame(
	value = c("GOOD", "GOOD", "BAD", "BAD"),
	qc = c("000", "001", "002", "003"))

# A case where all values are bad
output <- output[output$value == "BAD", ]

good <- which(output$qc %in% c("000", "001"))
print(good) # integer(0)
print(is.null(good)) # FALSE
if (length(good) > 0 || !(is.null(good))) {
	print("Some are good! Here they are:")
	output[good, ] # data frame with 0 rows
} else {
	print("All values bad") # Never prints!
}
```

I suspect deleting `|| !(is.null(good))` is enough to get the desired behavior, but check me on that.

ayushprd:
## Description
Even when `good` is an empty vector it does not display "all values bad" message. I have removed `!(is.null(good)))` to fix this.
## Motivation and Context
Fixes #2520  

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.remote/R/call_MODIS.R
+++ b/modules/data.remote/R/call_MODIS.R
@@ -244,7 +244,7 @@ call_MODIS <- function(var, product,
       output$qc[i] <- substr(convert, nchar(convert) - 2, nchar(convert))
     }
     good <- which(output$qc %in% c("000", "001"))
-    if (length(good) > 0 || !(is.null(good)))
+    if (length(good) > 0)
     {
       output <- output[good, ]
     } else {


bailsofhay:
@infotroph I made the issue today so we won't forget about it. Thank you for reminding me of that :)

PR #2622, opened 2020-06-02
Replace tmvtnorm

### LINKED ISSUE #2621: replacing tmvtnorm with TruncatedNormal
## Description

I was using [tmvtnorm](https://CRAN.R-project.org/package=tmvtnorm) package functions in PDA module, but I realized that it is veery veery slow. I'm planning to switch to [TruncatedNormal](https://CRAN.R-project.org/package=TruncatedNormal).  

Depending on the value of the B argument (number of replications for MC) in its `TruncatedNormal::dtmvnorm` function, it can be much faster than `tmvtnorm::dtmvnorm`. To be specific, the default value for B is `1e4` which makes `TruncatedNormal::dtmvnorm` even slower than `tmvtnorm::dtmvnorm`, but setting it to `1e2` is much faster. `tmvtnorm::dtmvnorm` is more precise than `TruncatedNormal::dtmvnorm` (when `B=1e2`), but it matters on the 2nd decimal place, which is fine for my purposes.

If there is no objection, I'm gonna replace `tmvtnorm` with `TruncatedNormal`.

istfer:
## Description

Replaced [tmvtnorm](https://CRAN.R-project.org/package=tmvtnorm) package functions in PDA with  [TruncatedNormal](https://CRAN.R-project.org/package=TruncatedNormal).  

## Motivation and Context

Fixes #2621 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Changed
 
+- Replaced `tmvtnorm` package with `TruncatedNormal` package for speed up.
 - Continuous integration changes: Added experimental GitHub Actions CI builds (#2544), streamlined Travis CI builds, added a fourth R version (second-newest old release; currently R 3.5) to Travis test matrix (#2592).
 - Functions that update database entries no longer pass `created_at` or `updated_at` timestamps. The database now updates these itself and ensures they are consistently in UTC (#1083).
 - `PEcAn.DB::insert_table` now uses `DBI::dbAppendTable` internally instead of manually constructed SQL (#2552).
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Changed
 
+- Replaced `tmvtnorm` package with `TruncatedNormal` package for speed up.
```
robkooper, line 27: can you add the issue as well, so people can see why it got replaced.

istfer:
> I assume the data does not change with the number of samples run (100 instead of 10000)

it is a very small probability change, which is typically not influential on the results.

PR #2624, opened 2020-06-02
Add "optparse" for command-line args to workflow.R

kyclark:
## Description

The current version of workflow.R incorrectly parses the command-line arguments for the given XML configuration file. Looking at the following section:

```
if (is.na(args[1])) {
  settings <- PEcAn.settings::read.settings("pecan.xml")
} else {
  settings_file <- args[1]
  settings <- PEcAn.settings::read.settings(settings_file)
}
```
Per the documentation on `commandArgs`, `args` will be:

     A character vector containing the name of the executable and the
     user-supplied command line arguments.  The first element is the
     name of the executable by which R was invoked.  The exact form of
     this element is platform dependent: it may be the fully qualified
     name, or simply the last component (or basename) of the
     application, or for an embedded R it can be anything the
     programmer supplied.

This program is intended to be executed like so:

    ./workflow.R --settings input.xml

This code is using `args[1]` which will actually be the path to the R executable, not the name of the "settings" XML document.

The only reason the program succeeds in using the given XML is because the "read.settings" function in "base/settings/R/read.settings.R"  separately parses the `commandArgs` for the presence of a "--settings" option and grabs the "input.xml" value. 

FWIW, I would strongly argue AGAINST this as this hidden behavior has allowed this bug to persist. The function should only use and validate the given "inputfile" and should definitely NOT parse the command-line arguments OR inspect Sys.getenv("PECAN_SETTINGS"), but that is another battle.

## Motivation and Context

The "optparse" establishes the parameters for a program in a way that both validates the inputs and generates documentation. This program also, apparently, accepts a "--continue" flag, but the only way to know this is to read the source code itself. As written, this program could be run with any other arbitrary arguments without error. 

By describing the parameters to "optparse," the program will now reject unknown arguments and also generate a "usage" statement when executed with "-h" or "--help":

```
$ ./workflow.R -h
Usage: ./workflow.R [options]


Options:
	-s FILE, --settings=FILE
		Settings XML file

	-c, --continue
		Continue processing

	-h, --help
		Show this help message and exit

```

Note that unknown arguments are rejected:

```
$ ./workflow.R --foo bar
Error in getopt_options(object, args) :
  Error in getopt(spec = spec, opt = args) : long flag "foo" is invalid
Calls: main ... get_args -> parse_args -> parse_options -> getopt_options
Execution halted
```

And invalid file arguments for "--settings" are rejected:

```
$ ./workflow.R --settings foo.xml
Usage: ./workflow.R [options]


Options:
	-s FILE, --settings=FILE
		Settings XML file

	-c, --continue
		Continue processing

	-h, --help
		Show this help message and exit


Error in get_args() : --settings "foo.xml" not a valid file
Calls: main -> get_args
Execution halted
```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


kyclark:
Hmm, I'd intended only the "workflow.R" program to be included with this PR, not the other files. Sorry. I'm still learning how to make PRs.

dlebauer:
@kyclark easiest way to just add the workflow.R might be to open up a new PR by copy-pasting the contents of your workflow.R here: https://github.com/PecanProject/pecan/edit/develop/web/workflow.R 

Then copy-paste your comments above into the new PR.

kyclark:
Thanks for the tip on how to create a proper PR, @dlebauer. I've followed your instructions. I also realized that my suggested changes were incomplete, so I hope this version has all my ideas in place.

kyclark:
Would it be OK for me to close this request so we can focus on #2626?

dlebauer:
thanks @kyclark closing this since it is replaced by #2626

robkooper:
@dlebauer lets not try to keep working on branches here in pecanproject/pecan. Goal really is for people to fork and work in their own forks.

PR #2625, opened 2020-06-03
don't override random.effects = TRUE

dlebauer:
## Description

Current behavior: if random effects = TRUE in pecan.xml but none of the trait data have statistics then random effects is set to FALSE. A message is written to the log but this is mostly a hidden assumption that should be up to the user - because it may indicate an incorrect assumption about the data or missing data, or an overspecified model etc.

This was done because models will not converge if there are insufficient data to estimate all of the random effects parameters. 

## Motivation and Context

I think setting random.effects = FALSE should be a decision made by the end uer.

Even when implemented in 2011, it was not clear that this was necessary. See my comment here https://github.com/PecanProject/pecan/commit/f6bcb78ea144e564bae781e32bcc307e9ec793b3#r8861144

Two other options:

1. come up with a more sophisticated rule for identifying data that will lead to underspecified models
  * not really a good idea - I vote pass it back to the user so they can evaluate the data and model
2. upgrade to @ashiklom 's multivariate version of the code
  * Good plan, but this will fix what I currently believe is a bug

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

- [x] Bug fix (there is no existing issue - it was found by @KristinaRiemer and discussed in Slack)

## Checklist:

### DIFF
--- a/modules/meta.analysis/R/meta.analysis.R
+++ b/modules/meta.analysis/R/meta.analysis.R
@@ -109,11 +109,10 @@ pecan.ma <- function(trait.data, prior.distns,
     ## check for excess missing data
 
     if (all(is.na(data[["obs.prec"]]))) {
-      if (verbose) {
-        writeLines("NO ERROR STATS PROVIDED, DROPPING RANDOM EFFECTS")
-      }
-      data$site <- rep(1, nrow(data))
-      data$trt  <- rep(0, nrow(data))
+      logger.warn("NO ERROR STATS PROVIDED\n Check meta-analysis Model Convergence", 
+                  "and consider turning off Random Effects by", 
+                  "setting <random.effects>FALSE</random.effects>",
+                  "in your pecan.xml settings file ")
     }
 
     if (!random) {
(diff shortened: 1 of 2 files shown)


modules/meta.analysis/R/meta.analysis.R
```diff
@@ -109,11 +109,10 @@ pecan.ma <- function(trait.data, prior.distns,
     ## check for excess missing data
 
     if (all(is.na(data[["obs.prec"]]))) {
-      if (verbose) {
-        writeLines("NO ERROR STATS PROVIDED, DROPPING RANDOM EFFECTS")
-      }
-      data$site <- rep(1, nrow(data))
-      data$trt  <- rep(0, nrow(data))
+      logger.warn("NO ERROR STATS PROVIDED\n Check meta-analysis Model Convergence",
```
dlebauer, line 112:
```suggestion
      PEcAn.logger::logger.warn("NO ERROR STATS PROVIDED\n Check meta-analysis Model Convergence", 
```

mdietze:
This one went through quicker than I could comment on it. I don't think this change was a great idea, but if we do keep it I think it is absolutely imperative that the *DEFAULT* for random effects be set to FALSE, otherwise I predict we're going to get failed runs and nonsensical parameter estimates that are going to drive people crazy.

On top of that, I also think we need to be moving toward a place where the meta-analysis is not run by default either, and runs make more use of existing posteriors (especially those coming out of the PDA and HPDA). The database isn't changing quickly these days and I think it makes more sense for the core teams for each model to just update the MA as needed.

PR #2632, opened 2020-06-10
use get_postgres_envvars in betyConnect

robkooper:
PEcAn.DB::betyConnect() now uses environment variables to create a connection. This makes it easier to use this function in more places.

## Motivation and Contex
This will make it such that @tezansahu can use this function in PEcAn API code

1. Load config from config.php (if it exists)
2. use PG* environment variables to add/overwrite settings
3. use some common defaults for missing variables
4. create connection

This now no longer uses  `dplyr::src_postgres`

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/db/R/query.dplyr.R
```diff
@@ -4,19 +4,32 @@
...
+
+  ## fill in all data from environment variables
+  dbparams <- get_postgres_envvars(host = getphp("db_bety_hostname", "localhost"),
+                                   port = getphp("db_bety_port", "5432"),
+                                   dbname = getphp("db_bety_database", "bety"),
+                                   user = getphp("db_bety_username", "bety"),
+                                   password = getphp("db_bety_password", "bety"))
 
-  config.list <- PEcAn.utils::read_web_config(php.config)
+  ## force driver to be postgres, needed when switching to db,open
```
infotroph, line 26:
```suggestion
  ## force driver to be postgres (only value supported by db.open)
```

(for the benefit of folks reading later who don't know the commit history)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Fixed
 
+- PEcAn.DB::betyConnect() is now smarter, and will try to use either config.php or environment variables to create a connection. It has switched to use db.open helper function.
```
infotroph, line 12:
```suggestion
- PEcAn.DB::betyConnect() is now smarter, and will try to use either config.php or environment variables to create a connection. It has switched to use db.open helper function (#2632).
```

robkooper:
I was thinking as well about the order of things. It be nice if there was some order that we want to use and if a variable is defined in one place that will be the value that sticks. For example docker-compose has a very explicit listing of how they find a variable:

When you set the same environment variable in multiple files, here’s the priority used by [docker-compose](https://docs.docker.com/compose/environment-variables/) to choose which value to use:

1. Compose file
2. Shell environment variables
3. Environment file
4. Dockerfile
5. Variable is not defined

PR #2633, opened 2020-06-11
satellitetools by Olli

ayushprd:
## Description

satellitetools developed by (Olli Nevalainen, FMI) contains python scripts to download Sentinel 2 data from Google Earth Engine. It also has an python implementation of the SNAP toolbox, which is used to perform computation on the downloaded data. 

Original repository: https://github.com/ollinevalainen/satellitetools 

## Motivation and Context

These codes will be used while extending the remote data module.

Please note that **Olli is the author** of satellitetools. These are not my own codes and I have not contributed to it by any means whatsoever. I am pushing it because my upcoming PR's use satellitetools as an external library.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


mdietze:
If this code exists in another repository, why is it being added to the PEcAn repo rather than just being a dependency? This seems to be redundant and will result in the two versions diverging.

istfer:
It is intended to diverge, we were thinking of making use of these existing code by @ollinevalainen but it is probable that the code will need modifications (not necessarily benefiting Olli's needs). 

@ayushprd was pushing it as it is first then he will change it as he goes, it was just more convenient to review his stepwise progress with respect to the original code this way.

istfer:
@ayushprd do we need all the .nc files in the output directory? Do we need that directory at all?

istfer:
Likewise Ruukki files are quite project specific, can we remove them as well? In general, it would be better to keep the absolute minimum rather than pushing everything

ayushprd:
I should've removed those, will do. Are those causing the checks to fails @istfer ?

mdietze:
As @istfer noted, there seems to be a lot of project-specific example files in this PR (there's also xlsx files in there). If you just delete them, they'll remain in the repo forever. I'd suggest starting a new PR from a fresh branch where you only add the filed you actually need.

istfer:
Thanks @mdietze that's a good point.

@ayushprd can we just keep it to the NDVI retrieval? We need to think about how to handle SNAP auxiliary data files, even though they are the training data for the algorithm and part of the module mechanics, they also shouldn't live in the repository

PR #2634, opened 2020-06-11
Sentinel 2 NDVI script for remote data module

ayushprd:
## Description
This PR uses satellitetool's functions to calculate NDVI from Sentinel 2 data. 

## Motivation and Context

These scripts will be used to create an automated workflow for the remote data module.
To test the PR please install these python dependencies [requirements.txt](https://github.com/ayushprd/PEcAn-GEE/blob/master/requirements.txt) These will later be added to the PEcAn dependencies. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/data.remote/inst/satellitetools/gee.py
```diff
@@ -0,0 +1,766 @@
...
+    s2_data_feature_collection = feature_collection.map(
+        ee_get_s2_data_feature).getInfo()
+
+    s2_data = s2_feature_collection_to_dataframes(s2_data_feature_collection)
+
+    for a in AOIs:
+        name = a.name
+        a.data = s2_data[name]
+
+# Old, unused function, remove at some point.
```
istfer, line 450: Is it safe to remove this already?

ayushprd: Yes, removed.

modules/data.remote/inst/s2ndvi.py
```diff
@@ -0,0 +1,72 @@
...
+    
+    geofile (str) -- path to the file containing the coordinates of AOI, currently tested with geojson. 
+    
+    outdir (str) -- path to the directory where the output file is stored. If specified directory does not exists, it is created.
+  
+    start (str) -- starting date of the data request in the form YYYY-MM-DD
+    
+    end (str) -- ending date of the data request in the form YYYY-MM-DD
+    
+    qi_threshold (float) -- From satellitetools: Threshold value to filter images based on used qi filter. qi filter holds labels of classes whose percentages within the AOI is summed. If the sum is 				     larger then the qi_threhold, data will not be retrieved for that date/image. The default is 1, meaning all data is retrieved
```
istfer, line 33:
```suggestion
    qi_threshold (float) -- From satellitetools: Threshold value to filter images based on used qi filter. qi filter holds labels of classes whose percentages within the AOI is summed. If the sum is larger then the qi_threshold, data will not be retrieved for that date/image. The default is 1, meaning all data is retrieved
```

modules/data.remote/inst/s2ndvi.py
```diff
@@ -0,0 +1,72 @@
...
+    # convert dataframe to an xarray dataset, used later for converting to netCDF
+    gee.s2_data_to_xarray(area, request)
+
+    # calculate NDVI for the selected area
+    area.data = gee.compute_ndvi(area.data)
+
+    timeseries = {}
+    timeseries_variable = ["ndvi"]
+
+    # is specified output directory does not exist, create it.
```
istfer, line 66:
```suggestion
    # if specified output directory does not exist, create it.
```

modules/data.remote/inst/satellitetools/gee.py
```diff
@@ -0,0 +1,716 @@
...
+
+def xr_dataset_to_timeseries(xr_dataset, variables):
+    """Compute timeseries dataframe from xr dataset.
+
+    Parameters
+    ----------
+    xr_dataset : xarray dataset
+
+    variables : list
+        list of varbiale names as string.
```
istfer, line 685:
```suggestion
        list of variable names as string.
```

modules/data.remote/inst/s2ndvi.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+Module to calculate NDVI from sentinel 2 data, based on Satellitetools by Olli Nevalainen.
+This will be used to create automated workflows for the remote data module.
+
+Warning: Point coordinates as input has currently not been implemented.
+
+Author: Ayush Prasad
+"""
+
+from satellitetools import gee
```
istfer, line 13: I'm not familiar with python, but just to confirm, does this also work without having satellitetools as a library?

ayushprd: satellitetools is just the name of the folder which contains the original script `gee.py`, this script is required.

istfer: Yup that much I followed, I wasn't just aware if one can simply import a function from a folder

istfer:
I know this is not a complete work and will be built upon, but would it still be nice to include a small vignette to accompany this PR? Nothing too fancy, just the main steps to be able to make the Reykjavik example work, where you can also list the python dependencies?

ayushprd:
Yes, @istfer I'll definitely do that, I didn't add that in the docstring right now because, the arguments to the function may be change in the very near future.

modules/data.remote/inst/s2ndvi.py
```diff
@@ -0,0 +1,81 @@
...
+    Returns
+    -------
+    Nothing:
+            output netCDF is saved in the specified directory.
+            
+    Python dependencies required: earthengine-api, geopandas, pandas, netCDF4, xarray
+    
+    To test this function please run the following code, testfile is included.
+    
+    s2_ndvi(geofile="./test.geojson", outdir="./out/", start="2019-01-01", end="2019-12-31", qi_threshold=1)
```
istfer, line 45: could you move `test.geojson` into `satellitetools` folder and update this, not a big deal but that feels like a more appropriate place

ayushprd: Sure.

robkooper:
@dlebauer I think this is a good case to squash before merge, since some files were added and removed later. If we don't squash when we do the merge, this will result in multiple commits that add/remove files and will grow the repository without really need be.

Issue #2623, opened 2020-06-02
model2netcdf.ED2 Fails if  output files missing "AVG_SOIL_TEMP" aka "Soiltemp"


mccabete:
### Bug Description

model2netcdf.ED2 fails if `-T-` file is missing the variable "AVG_SOIL_TEMP", known as "Soiltemp" in the pecan standard.  It fails specifically at [this line](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L553). 

### To Reproduce

1. Run met2model.ED2() on an ED2 generated `-T-` file that has no "AVG_SOIL_TEMP" data. 

## Expected behavior

The function's assigning `Soiltemp` the value `-999` would mean that the `.nc` file it produces would have "missing values" of `-999` for Soiltemp. 

## Error message
```
2020-06-02 16:00:40 WARN   [model2netcdf.ED2.R#278: PEcAn.logger::logger.warn] : 
   Could not find AVG_SOIL_TEMP in ed hdf5 output. 
Error in array(0, ncol(soiltemp)) : 'dims' cannot be of length 0
```

## Machine (please complete the following information):

 - Server BU test-pecan, Tess' personal build. Up to date with develop. 

## Additional context

I tried an `if(soiltemp != -999){}` around the code chunk from [line 520](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L520) to [line 579](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L579), but it messed up the dimensions during the write-out step.

mccabete:
This error is dependent on the output files being written out by the version of ED2 one is using. In my personal branch, I modified (this section)[https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L285] to check for FMEAN_SOIL_TEMP_PY instead of average values. 

```
CheckED2Version <- function(nc) {
    if ("FMEAN_BDEAD_PY" %in% names(nc$var)) {
      return("Git")
    }
    if ("FMEAN_SOIL_TEMP_PY" %in% names(nc$var)){
      return("Tess")
    }
  }
```

mdietze:
[...]

PR #2636, opened 2020-06-15
Improve checking for output variable names in model2necdf.ED2

mccabete:
model2netcdf.ED2 will only process output with an `FMEAN*` naming convention if it finds the variable `FMEAN_BDEAD_PY`. This causes issues if `FMEAN_BDEAD_PY` is missing. The function is written to assume that all non-"git" builds of ED2 don't have the `FMEAN*` naming convention. 

## Fixes
- added a check for an additional FMEAN Variable
- added a warning 
- changed the check function name from CheckED2Version to CheckED2Variables

## Motivation and Context
Different pecan users have different versions of the ED2 model that they are using. Different ED2 versions, and builds, have different output names as part of the `-T-` file output. The read_T_files function supports two different types of output names, [AVE*](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L498) names, and [FMEAN*](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/model2netcdf.ED2.R#L336) names. 

Previously, `read_T_files()` would check for one variable `FMEAN_BDEAD_PY`, and if it existed, it would say the ED2 version was "Git" and switch to the `FMEAN*`  names. This caused issues for me, because I was not producing `FMEAN_BDEAD_PY`, but was using the FMEAN* names. 

This PR is a more general solution to the issue I opened [here](https://github.com/PecanProject/pecan/issues/2623). 

I moved away from the assumption that the version of ED2 is the only thing that is determining variable names, because 1) "Git" is an ED2 build used with pecan, but doesn't map onto an ED2 version clearly 2) people's personal builds of ED2 will deviate from versions anyway 

## Review Time Estimate

This is a very small PR, and should take <5 min to review. 
- [x ] Immediately
- [ ] Within one week
- [x ] When possible
## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue) 

## Checklist:

- [x ] I have updated the CHANGELOG.md.
- [x ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/models/ed/R/model2netcdf.ED2.R
+++ b/models/ed/R/model2netcdf.ED2.R
@@ -280,9 +280,19 @@ read_T_files <- function(yr, yfiles, tfiles, outdir, start_date, end_date, ...){
     }
   }
   
-  CheckED2Version <- function(nc) {
+  CheckED2Variables <- function(nc) {
+    vars_detected <- NULL
+  
     if ("FMEAN_BDEAD_PY" %in% names(nc$var)) {
-      return("Git")
+      vars_detected <- c(vars_detected,"FMEAN_BDEAD_PY")
+      return("Contains_FMEAN")
+    }
+    if ("FMEAN_SOIL_TEMP_PY" %in% names(nc$var)) {
+      vars_detected <- c(vars_detected, "FMEAN_SOIL_TEMP_PY")
+      return("Contains_FMEAN")
+    }
+    if(!is.null(vars_detected)){
+      PEcAn.logger::logger.warn(paste("Found variable(s): ", paste(vars_detected, collapse = " "), ", now processing FMEAN* named variables. Note that varible naming conventions may change with ED2 version."))
     }
   }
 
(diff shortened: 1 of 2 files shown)


models/ed/R/model2netcdf.ED2.R
```diff
@@ -280,9 +280,19 @@ read_T_files <- function(yr, yfiles, tfiles, outdir, start_date, end_date, ...){
...
     if ("FMEAN_BDEAD_PY" %in% names(nc$var)) {
-      return("Git")
+      vars_detected <- c(vars_detected,"FMEAN_BDEAD_PY")
+      return("Contains_FMEAN")
+    }
+    if ("FMEAN_SOIL_TEMP_PY" %in% names(nc$var)) {
+      vars_detected <- c(vars_detected, "FMEAN_SOIL_TEMP_PY")
+      return("Contains_FMEAN")
+    }
+    if(!is.null(vars_detected)){
```
istfer, line 294: `vars_detected` is set `NULL` in the beginning and the only two conditions where it becomes not NULL returns from the function already, how this `if` will ever be evaluated? What was the intended behavior?

mccabete: True! I'll fix it.

mccabete: Thanks Istem

mccabete:
ok @istfer I changed that so the warning actually gets thrown.

PR #2637, opened 2020-06-16
LAI script for remote data module

ayushprd:
## Description
This PR adds the SNAP biophysical processor and uses it to calculate LAI.
The auxiliary data required for SNAP is directly fetched from the original repository. 
Also adds `remote_process.py`, which will be used to control the individual functions.

## Motivation and Context
The scripts in this PR along with #2634 will be used to create the initial workflow for the remote data module.

Dependencies required: [requirements.txt](https://github.com/ayushprd/PEcAn-GEE/blob/master/requirements.txt) (same as the referenced PR)

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+        newarr,
+        input_core_dims=[["band", "xy"]],
+        output_core_dims=[["xy"]],
+        kwargs={"variable": variable},
+        vectorize=True,
+    ).unstack()
+    return dataset.assign({variable.lower(): arr})
+
+
+def estimate_gpp_vi_lue(vi, daily_par, model_name="ci_red_edge_1"):
```
mdietze, line 272: This function takes us a step past the estimation of LAI and adds a (overly) simplistic GPP calculation with a lot of hard-coded parameters. @istfer is this really needed for the application you envisioned? I'd really prefer to drop this and `GPP_LUE_models` as likely to cause more trouble than benifit (e.g. attempts to calibrate or benchmark models against models).

istfer: I think these are creeping in from Olli's code without being intended, GPP calculation is not a part of the GEE-GSOC work (it's an exercise from another project of Olli's)

I'll only be able to review this code tomorrow @ayushprd then we can comb out what we need

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+    """
+    vi_name = "_".join(model_name.split("_")[:-1])
+    gpp = GPP_LUE_models[vi_name][model_name]["model"](
+        np.array(vi), np.array(daily_par)
+    )
+    return gpp
+
+
+# GPP estimation models
+GPP_LUE_models = {
```
ayushprd, line 301: This dictionary containing the model details is not required.

modules/data.remote/inst/satellitetools/biophys_xarray.py
ayushprd, line 272: Not required

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+    ).unstack()
+    return dataset.assign({variable.lower(): arr})
+
+
+def estimate_gpp_vi_lue(vi, daily_par, model_name="ci_red_edge_1"):
+    """Estimate GPP using simple vegetation index based models and PAR.
+
+    This function has not been properly tested (i.e.used for a while)
+
+    Parameters
```
ayushprd, line 277: This function is not required.

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+
+    pixel_vector[8, :] = np.cos(np.radians(single_date_dict["view_zenith"]))
+    pixel_vector[9, :] = np.cos(np.radians(single_date_dict["sun_zenith"]))
+    pixel_vector[10, :] = np.cos(
+        np.radians(single_date_dict["sun_azimuth"] - single_date_dict["view_azimuth"])
+    )
+    return pixel_vector
+
+
+def compute_ci_red_edge(dataset):
```
ayushprd, line 187: Not required for LAI.

ayushprd, line 187: This function is not required for LAI.

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+        out_layer2,
+        nn_params[variable]["denorm_minmax"][0],
+        nn_params[variable]["denorm_minmax"][1],
+    )[0]
+    output = _output_ouf_of_range(output, variable)
+    output = output.reshape(1, np.shape(x)[1])
+    return output
+
+
+def _s2_lists_to_pixel_vectors(single_date_dict):
```
ayushprd, line 172: Not required for LAI.

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+    res_url = requests.get(res_url)
+    # check the HTTP status code to see if any error has occured.
+    res_url.raise_for_status()
+    # store the contents of the url in an in-memory buffer and use it to load the ndarray.
+    return np.loadtxt(io.BytesIO(res_url.content), delimiter=",")
+
+
+# Read SNAP Biophysical processor neural network parameters
+nn_params = {}
+for var in ["FAPAR", "FCOVER", "LAI", "LAI_Cab", "LAI_Cw"]:
```
ayushprd, line 62: I can make this part specifically for LAI instead of the generalized approach.

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+import io
+import numpy as np
+import xarray as xr
+
+# url to Sentinel 2 Toolbox's auxdata
+# This base_url points towards the original toolbox(not the one created by Olli)
+base_url = "https://raw.githubusercontent.com/senbox-org/s2tbx/master/s2tbx-biophysical/src/main/resources/auxdata/2_1/{}/{}"
+
+
+def get_fromurl(var, pattern):
```
ayushprd, line 35: Var can be hardcoded to LAI

modules/data.remote/inst/satellitetools/biophys_xarray.py
ayushprd: The reason why I did not remove this plus the other features such as calculation of FAPAR, FCOVER, etc is that I thought these functions could be helpful to you all later on even if not used immediately. The other file included in this PR, `s2lai.py` simply calculates the LAI and does not estimate GPP or make use of the other functions present in SNAP. However, I understand these are not helpful, so I've marked the parts based on my understanding and testing which can be removed and will not affect the calculation of LAI. Sorry for the confusion. Henceforth, I will only think of NDVI and LAI while developing this project.

istfer: Hi Ayush, no please don't only think of NDVI and LAI while developing this project, we will need all sorts of data types and derivations. But as Sentinel2-GEE-NDVI-LAI pipeline already exists, that is what you can focus first of course, if that's what you mean.

It is fine that you weren't sure if other features were wanted or not. We normally want this module to give us data to compare to model predictions, but remote sensing data usually need some operation first. As you can see LAI is already more of a modeled-output than data, but most models do predict LAI, so we don't need the extra massaging it into GPP etc. Some models can already predict spectral bands, then we can directly use those and don't even need the LAI calculation step.

Thanks for marking the parts that are not needed for LAI calculation already, I'll get back to this PR/you after my morning meetings.

modules/data.remote/inst/s2lai.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns
+    -------
+    Nothing:
+            output netCDF is saved in the specified directory.
+            
+    Python dependencies required: earthengine-api, geopandas, pandas, netCDF4, xarray
+    
+    To test this function please run the following code inside a python shell after importing this module, testfile is included.
+    
+    s2_lai(geofile="./satellitetools/test.geojson", outdir="./out/", start="2019-01-01", end="2019-12-31", qi_threshold=1)
```
istfer, line 47: Ayush, I still see these PRs as an exercise where you play with Olli's functions and get a feeling of the pipeline, but just wanted to point it out early on that if/when we will generalize the module, we will need to think a bit more about the function names. I will post something on our gee-slack channel to explain what I mean.

For now, I'm fine with having this function in so that anyone who has a geojson file can already use it for getting LAI values for their ROI :+1:

modules/data.remote/inst/s2lai.py
```diff
@@ -0,0 +1,80 @@
...
+    area = gee.AOI(df[df.columns[0]].iloc[0], df[df.columns[1]].iloc[0])
+
+    # calcuate qi attribute for the AOI
+    gee.ee_get_s2_quality_info(area, request)
+
+    # get the final data
+    gee.ee_get_s2_data(area, request, qi_threshold=qi_threshold)
+
+    # convert dataframe to an xarray dataset, used later for converting to netCDF
+    gee.s2_data_to_xarray(area, request)
```
istfer, line 66: up until here, this function overlaps with s2ndvi.py which seems redundant. Two options here:
1) modify this function to pick it up from where s2ndvi.py left
2) delete/modify s2ndvi.py in favor of this one

(2) is fine because there is not much use for NDVI only, and as I said, I don't see these functions as the main functions of the module yet (see the post on our slack channel). After this PR we will have one function that gets LAI for a ROI which is already good. 

But going with (1) would be a good exercise for you to figure out how to divide the pipeline into subtasks. It could give you some design ideas. Then, for example, you would need a "master" function that calls `s2ndvi.py` first (you may also consider renaming it to something more general like `gee2bands.py`, but `s2ndvi.py` is also fine for now) and then `bands2lai.py` which calculates LAI

modules/data.remote/inst/satellitetools/biophys_xarray.py
istfer: I think I like the generalization, let's not lose it yet

modules/data.remote/inst/satellitetools/biophys_xarray.py
istfer: As said in the previous comment, let's keep it this way for now

modules/data.remote/inst/satellitetools/biophys_xarray.py
istfer: I think this is safe to remove

modules/data.remote/inst/satellitetools/biophys_xarray.py
istfer: can remove but not sure what it does

modules/data.remote/inst/satellitetools/biophys_xarray.py
istfer: let's remove

modules/data.remote/inst/satellitetools/biophys_xarray.py
istfer: let's remove

modules/data.remote/inst/satellitetools/biophys_xarray.py
```diff
@@ -0,0 +1,408 @@
...
+"""
+
+import requests
+import io
+import numpy as np
+import xarray as xr
+
+# url to Sentinel 2 Toolbox's auxdata
+# This base_url points towards the original toolbox(not the one created by Olli)
+base_url = "https://raw.githubusercontent.com/senbox-org/s2tbx/master/s2tbx-biophysical/src/main/resources/auxdata/2_1/{}/{}"
```
istfer, line 32: directly pointing to the s2tbx was a neat idea :+1:

modules/data.remote/inst/s2lai.py
ayushprd: (1) is exactly what I have planned, I didn't do it yet because I cannot decide for sure till we discuss the automation/db part.

istfer: For now you can ignore the DB part, as being discussed on the slack channel, it will require some more thought. This PR can focus on providing users some stand-alone mini-workflow that calculates LAI for a ROI

ayushprd: Right now if I were to follow (1), then I would remove `s2lai.py` completely, and then until we have the module ready, the LAI function will have to be used through `s2ndvi.py`(I can change its name)

istfer: Why would you need to remove it completely? I was suggesting `s2ndvi.py` (could be renamed) would write the area data (`area.data.to_netcdf`), and `s2lai.py` (could be renamed) will take it from there, read the data back in and calculate LAI. Would that work?

ayushprd: What I am suggesting is, 

1. Temporarily until the module is ready, there would be one main script to access the contents of satellitetools (`biophys_xarray.py` and `geeapi.py`.)

2. Since both `s2ndvi.py` and `s2lai.py` themselves do not perform any calculations and instead rely on `geeapi.py` and `biophys_xarray.py`, I could remove one of them. For example, `s2lai.py` uses `run_snap_biophys()` to calculate lai, I can directly call this function from `run_snap_biophys.py` instead of calling it through and intermediate script.

ayushprd: > Why would you need to remove it completely? I was suggesting `s2ndvi.py` (could be renamed) would write the area data (`area.data.to_netcdf`), and `s2lai.py` (could be renamed) will take it from there, read the data back in and calculate LAI. Would that work?

This is what I am suggesting too, except that there is no need of `s2lai.py` I can directly call its contents from `biophys_xarray.py`.

istfer: Sure, that is more in-line with (2), and I'm fine with it. Granted, `s2lai.py` would only read data back, call `bio.run_snap_biophys` and write the results. Probably not worth having it as a separate function, as I said it was only going to be an exercise.

modules/data.remote/inst/satellitetools/biophys_xarray.py
mdietze: I second @istfer point -- my request to remove GPP was not a blanket request to remove everything that's not LAI. I agree that FAPAR and FCOVER are useful.

As a broader point, when it comes to remote sensing there's definitely a grey area when it comes to the question of what is "data" and what is a prediction from a model that's using remote sensing as one of its inputs. Raw bands are definitely in the first category, but very few of our ecosystem models predict that so they're not particularly useful validation. Also in this category are simple mathematical transforms of raw bands (e.g. NDVI) but process-models are likewise not routinely calculating this. GPP is definitely in the second category -- satellites are by no means observing this, but instead there are models that take satellite LAI as an input (along with light, temperature, humidity, etc. as well as a bunch of vegetation specific parameters) to try and predict GPP. In between these extremes of NDVI and GPP are a lot of algorithms that fall in a grey area of being more than an analytical transform but not fully a model, with fAPAR and LAI being great examples -- they are mostly physics, but require some assumptions about the canopy structure and leaf chemistry -- and different assumptions can lead to different numbers (though generally different algorithms are all pretty close).

modules/data.remote/inst/s2lai.py
ayushprd: Since I am removing some files, should I close this PR and a open a new one? or continue here itself?

istfer: Hmm good question, if this was my PR I would just keep working on it since the added/deleted files are small and related to the code (as opposed to the netcdf and excel data files in the previous PR). What's the procedure we want to stick here @robkooper ?

mdietze: Yeah, this is a small diff on text files, so it doesn't need a new PR

ayushprd:
@istfer I have made the discussed changes:

-  GPP calculation and other unused functions have been removed from `biophys_xarray.py`

- I have added a new file `gee2pecan_bands.py` which only retrieves the data and saves it.

-  `bands2ndvi.py` and `bands2lai_snap` only calculate NDVI and LAI respectively using the raw file created in the previous step.

ayushprd:
Thank you for starting this discussion @istfer I feel this should be a new issue/PR but anyway,

1. Shouldn't the settings list created from `pecan.xml`(containing `remote_process` specific tags) be passed to `remote_process`? I was thinking that `remote_process` would extract the information and put them into individual variables and then pass it onto other functions. This way, the individual functions could also be used apart from the PEcAn workflow.

2. What is the benefit of passing the image collection id? The filtering process used in `gee2pecan_bands` is specific to Sentinel 2 and will not work if passed another id say `LANDSAT/LC08/C01/T1_SR`

3. #### BETYdb questions:

-  I think to insert an input file in the DB, `dbfile.input.insert()` does everything needed, is this correct?

- I was thinking that `remote_process` should check at the beginning itself if the requested output file already exists in the DB if it does not exist, then at the end of each function for example, `bands2lai_snap` we could call the input insert function to write the output netCDF file in the DB.

4. I need your help in deciding the programming language for `remote_process` Here's what I think,

     #### Advantages of R:

    - We can make use of built-in PEcAn functions and modules like PEcAn.DB, logger functions, etc.

    - As PEcAn is written in R and most of you all use R more frequently, I think that it is more likely that future algorithms and tools like SNAP would be written in R, in which case I think R would be more suitable.



    #### Disadvantage of R:
    - Since both gee and SNAP codes are in Python, I will have to use reticulate multiple times inside the module and as Shawn pointed it out earlier this does cause problems sometimes in interfacing Python and R. Although I think and I have tried it out to some extent, this problem can be minimized to a large extent when docker is used.

    #### Advantages of Python:
    - I will have to use reticulate only once i.e while calling `remote_process` from workflow.R.

    - Right now, this would make it easier to communicate between the functions.

   #### Disadvantage is,
    - For the most part it would not be possible to use the built-in functions of PEcAn.

istfer:
> Shouldn't the settings list created from pecan.xml(containing remote_process specific tags) be passed to remote_process? I was thinking that remote_process would extract the information and put them into individual variables and then pass it onto other functions. This way, the individual functions could also be used apart from the PEcAn workflow.

Yes, definitely, that's the plan, but it doesn't have to happen now. As these functions are not blended in with the rest of the pecan workflow yet, it's fine to write out the arguments explicitly. This way I was hoping that it will be easier for you to determine what sorts of tags will be needed in the pecan.xml in the next steps.

> What is the benefit of passing the image collection id? The filtering process used in gee2pecan_bands is specific to Sentinel 2 and will not work if passed another id say LANDSAT/LC08/C01/T1_SR

Sure, as mentioned in the comment above, I'm aware. We -in the duration of this GSOC project- do want to have access to collections other than Sentinel-2, right? This is just one tiiiny step towards this direction, to keep in mind that we do want to query data from other collections. You can omit it if you want, I thought it could be a nice way of expressing the intended behaviour of this module.

> BETYdb questions
Let's not insert anything to BETY at this stage. That is indeed a big enough feature to be in its own PR.

>I was thinking that remote_process should check at the beginning itself if the requested output file already exists in the DB if it does not exist, then at the end of each function for example, bands2lai_snap we could call the input insert function to write the output netCDF file in the DB. 

But for the sake of answering this question: Yes, exactly.

> deciding the programming language 
For this PR, all I'm looking for is a higher-level function that links your functions. Looking at the advantages you listed, I think it is good to go with python right now.

ayushprd:
>  We -in the duration of this GSOC project- do want to have access to collections other than Sentinel-2, right? This is just one tiiiny step towards this direction, to keep in mind that we do want to query data from other collections. You can omit it if you want, I thought it could be a nice way of expressing the intended behaviour of this module.

Yes, that's definitely possible and is something which I am looking to do. I wrote some parts of this already during the first week.

> For this PR, all I'm looking for is a higher-level function that links your functions. Looking at the advantages you listed, I think it is better to go with python at this point.

Understood.

PR #2641, opened 2020-06-19
Pass correct object to query.format.vars

infotroph:
## Description

One change missed in #2632, plus some Roxygen cleanup while I was in the file. Thanks to @kzarada for reporting.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/met.process.R
+++ b/modules/data.atmosphere/R/met.process.R
@@ -87,11 +87,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
   }
   
   # set up connection and host information
-  con <- dplyr::src_postgres(
-    dbname   = dbparms$dbname,
-    host     = dbparms$host,
-    user     = dbparms$user,
-    password = dbparms$password)$con
+  con <- PEcAn.DB::db.open(dbparms)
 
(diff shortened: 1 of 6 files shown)


infotroph:
On better testing, this just moves the problem. Looks like we might need to replace the [deprecated](https://dplyr.tidyverse.org/reference/src_dbi.html) `src_postgres` with a plain DBI connection.

modules/data.atmosphere/R/met.process.R
```diff
@@ -87,12 +87,12 @@ met.process <- function(site, input_met, start_date, end_date, model,
   }
   
   # set up connection and host information
-  bety <- dplyr::src_postgres(dbname   = dbparms$dbname, 
-                       host     = dbparms$host, 
-                       user     = dbparms$user, 
-                       password = dbparms$password)
-  
-  con <- bety$con
+  con <- dplyr::src_postgres(
```
robkooper, line 90: I think this can be replaced with ( not sure how to do multline suggestion)

```
  con <- PEcAn.DB::db.open(dbparms)
```

PR #2642, opened 2020-06-27
GEE - Landsat script and remote_process update

ayushprd:
## Description

This PR adds the `l8gee2pecan_bands()` function which can be used for calculating NDVI and getting surface reflectance band data from Landsat 8. 

Currently, if ROI type is a point this function can be used for getting SR data from Landsat 7, 5 and 4 as well.

## Motivation and Context

To be used in the remote data module.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


ayushprd:
1. Yes, `remote_process` should be able to call not just this but other GEE functions as well. Would it be okay if I make changes in the `remote_process` after 3-4 days? By then I expect to have the SMAP and MODIS GEE functions ready and I will be able to make more informed choices. Can you please explain the benefits of having an extra `gee2pecan_bands` layer, what new could we add there instead of directly calling `l8gee2pecan_bands` or `s2gee2pecan_bands`?

2. I will change the file names.

3. Yes, NDVI is being calculated on GEE. I have thought a lot about having a single function for calculating NDVI for all the collections but I cannot find a way to do it without hardcoding the band information. For example, L8's NIR band is B5 and S2's is B8, because of which it becomes very difficult to figure out the required bands unless the user requests exactly those bands in the previous step.

> NDVI could easily be stored together with the bands. Also, if you'd like to simplify the functions a bit, you can make NDVI calculation default.

I was wondering about this, thank you for clarifying.

istfer:
> Would it be okay if I make changes in the remote_process after 3-4 days?

OK, sounds good. But it could be bi-directional, I mean, trying to make `remote_process` getting to work with this function may also give you ideas to make more informed choices in those SMAP and MODIS functions as well. I'm leaving it up to you.

> Can you please explain the benefits of having an extra `gee2pecan_bands` layer, what new could we add there instead of directly calling `l8gee2pecan_bands` or `s2gee2pecan_bands`? 

That's a good question. In general, I believe the `remote_process` function should stay very high level, it should only track where we are in the workflow (do we need downloading raw data, does raw data exist and we need transformation, which-level variable we are dealing with etc.) and accordingly call relevant sub-functions. These will already get crowded, and we need factoring of the overall task. If you check out `ic_process` for example, it only calls sub modules, e.g. `get_veg`, `put_veg`. Similarly, we probably will need to turn `gee2pecan_bands` into a more generalized `get_remote_data` module soon, I mean the overall module may turn into something like this:
```
remote_process 
    - get_remote_data    
    - process_remote_data
    - maybe_some_other_major_subtask
```

> but I cannot find a way to do it without hardcoding the band information. 

Not sure if I'm following. But we already have two different functions for L8 and S2, right? Then we can already specify which bands to pass to NDVI calculation for these different collections. Could you clarify what you mean?

I'm a bit occupied today but will get back to this PR asap.

ayushprd:
Thank you for explaining, Istem. I totally agree, keeping `remote_process` very high level could make it a bit easier to add more data sources in future and also make the module more organized and clean.  

About NDVI: Apologies, I actually sidetracked from your original question. If you were asking whether the NDVI computation for S2 could also be moved to GEE, yes that can be done. It's slightly more tricky there (I will have to modify many functions) but yes its definitely possible. 

What I meant was, I thought about creating a single NDVI function which could work for all the datasets available on GEE. Like the current `bands2ndvi()` only works for S2 and I added a new NDVI function for L8 in this PR. But since you said I can make the NDVI calculation default while extracting the bands, there is probably no need of this "single function". I can do the NDVI calculation in `extract_s2_bands` itself.

istfer:
> If you were asking whether the NDVI computation for S2 could also be moved to GEE, yes that can be done. 

Yep, exactly. Great, then, we are on the same page about getting rid of `bands2ndvi()` and return NDVI by default in `extract_***_bands()` functions.

But I'm not sure which ones you plan to implement in this one, and which ones you plan to address in the following PRs. Let me know when you want this reviewed again.

ayushprd:
Changes made:

- merged `gee.py`, `bands2ndvi.py` and `gee2pecan_bands` to `gee2pecan_s2.py` 

- Created `gee_utils.py` which has the functions which are being in many scripts.

-  Created `get_remote_data.py` and `process_remote_data.py` which download and process data respectively. I have used dynamic imports to load the modules on the fly.

- `gee2pecan_l8()` and `gee2pecan_smap()` now download all bands by default. I tried doing the same for s2 but SNAP does not work when there are other bands in the input file. Will try to fix this.

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,191 @@
...
+    
+    outdir (str) -- path to the directory where the output file is stored. If specified directory does not exists, it is created.
+  
+    start (str) -- starting date of the data request in the form YYYY-MM-DD
+    
+    end (str) -- ending date areaof the data request in the form YYYY-MM-DD
+
+    ic (str) -- image collection id of the Landsat SR product from Google Earth Engine
+
+    vi (bool) -- set to True if NDVI needs to be calculated
```
istfer, line 41: `ic` and `vi` don't seem to be arguments to this function?

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Returns
+    -------
+    Nothing:
+            output netCDF is saved in the specified directory. 
+ 
+    """
+
+    # scale (int) Default: 30
+    scale = 30
```
istfer, line 55: could you explain what this is and if it could be a function arg if we want the flexibility for users to change

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,191 @@
...
+    # bands retrieved
+    bands = ["B1", "B2", "B3", "B4", "B5", "B6", "B7", "B10", "B11"]
```
istfer, line 57: I'm surprised that all bands need to be listed out, just curious, isn't there some `ee.select()` function feature that gets all bands if nothing is specified explicitly? or of they have some function arg like `all=TRUE` that returns everything or something like that?  there isn't much you can do otherwise of course

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def reduce_region(image):
```
istfer, line 59: I wonder what this does? Also even though they might not be used in other functions yet, `reduce_region` and `mask` almost seem like they belong to gee_utils?

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+
+    else:
+        landsat = (
+            ee.ImageCollection("LANDSAT/LC08/C01/T1_SR")
+            .filterDate(start, end)
+            .sort("system:time_start", True)
+        )
+
+    def calcNDVI(image):
```
istfer, line 92: also this probably belongs to gee_utils with an additional argument that specifies which bands to use

Although I know you don't use it for S2 yet

modules/data.remote/inst/gee_utils.py
```diff
@@ -0,0 +1,82 @@
...
+    Returns
+    -------
+    site_name (str) -- name of the AOI
+    """
+    df = gpd.read_file(geofile)
+    site_name = df[df.columns[0]].iloc[0]
+    return site_name
+
+
+def get_siteaoi(geofile):
```
istfer, line 68: just for the sake of resemblance with the previous function name, could this be called `get_sitecoord` or something like that

modules/data.remote/inst/get_remote_data.py
```diff
@@ -0,0 +1,65 @@
...
+    -------
+    Nothing:
+            output netCDF is saved in the specified directory.
+    """
+    if source == "gee":
+        try:
+            # get collection id from the dictionary
+            collection = collection_dict[collection]
+        except KeyError:
+            print("Requested image collection is not available")
```
istfer, line 53: could we have a more informative message that says something along these lines "Please check if the collection name you requested is one of these and spelled correctly [prints collection_dict first column]. If not, you need to implement a corresponding gee2pecan_* function and add it to the lookup table."

modules/data.remote/inst/gee2pecan_s2.py
```diff
@@ -690,27 +718,83 @@ def xr_dataset_to_timeseries(xr_dataset, variables):
...
+    # calcuate qi attribute for the AOI
+    ee_get_s2_quality_info(area, request)
+
+    # get the final data
+    ee_get_s2_data(area, request, qi_threshold=qi_threshold)
+
+    # convert dataframe to an xarray dataset, used later for converting to netCDF
+    s2_data_to_xarray(area, request)
+
+    area.data = compute_ndvi(area.data)
```
istfer, line 789: I know you already know this, just to document it: we would ideally like to compute ndvi on the GEE side and return with the other bands for S2 too, like we do for the L8.

modules/data.remote/inst/gee2pecan_l8.py
ayushprd: Yes forgot to remove.

modules/data.remote/inst/gee2pecan_l8.py
ayushprd: It's the pixel resolution of the image https://developers.google.com/earth-engine/scale . I have sort of used a generic value here and in the other functions. This will probably need to be configured depending upon the use case. I didn't pass it as an argument because `gee2pecan_s2` does not do it currently (it's listed in the TODO). I can make it configurable there as well? Overall, I think all the gee extraction functions should have the same type of arguments.

modules/data.remote/inst/gee2pecan_l8.py
ayushprd: There isn't any function which does that. `.select()` exists but it's used for selecting the bands from the collection.

modules/data.remote/inst/gee2pecan_l8.py
ayushprd: This reduces the image depending as per the type of the aggregation function used (I used mean here) [more details here](https://developers.google.com/earth-engine/reducers_intro) I can put these in the `gee_utils` but then these have to be configured on a case-by-case basis for every image collection.

modules/data.remote/inst/remote_process.py
```diff
@@ -9,25 +9,27 @@
...
     end,
-    qi_threshold,
-    source="gee",
-    collection="COPERNICUS/S2_SR",
-    input_type="bands",
-    output=["lai", "ndvi"],
+    source,
+    collection,
+    qc=None,
+    algorithm="snap",
```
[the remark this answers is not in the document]
ayushprd: Yes, I'll change this should be None

modules/data.remote/inst/get_remote_data.py
ayushprd: Nice!

modules/data.remote/inst/gee2pecan_s2.py
ayushprd: @istfer `gee2pecan_s2()` now calculates NDVI on GEE and all the bands are being retrieved without causing any issue in SNAP. Shall I add the initial book documentation in this PR itself?

istfer: Hi Ayush, I'll check this PR again as soon as I can but it might have to wait until tomorrow. In the meantime, yes, it would be wonderful if you could already add the initial book documentation to this PR :+1:

modules/data.remote/inst/gee2pecan_l8.py
istfer: >I think all the gee extraction functions should have the same type of arguments.
Agreed. Then, they could also be explained in the book-documentation.

It would be great if you could make it configurable, even though it is not currently used in gee2pecan_s2.

Also, could you add it to the book documentation with the link you shared here? it was very informative, thank you.

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,180 @@
...
+    Returns
+    -------
+    Nothing:
+            output netCDF is saved in the specified directory. 
+ 
+    """
+
+    # scale (int) Default: 30
+    scale = 30
+    # bands retrieved ["B1", "B2", "B3", "B4", "B5", "B6", "B7", "B10", "B11"]
```
istfer, line 52: you may delete this

modules/data.remote/inst/gee2pecan_l8.py
istfer: Thanks for the explanation. This also looks kind of important, that's why if it is moved to `gee_utils` it may be easier for people to realize what is going on under the hood, otherwise it is buried here. But if, you are using these only in `gee2pecan_l8` for now, then they can stay here for this PR. But as soon as they are need by another gee function it would be great to move them to `gee_utils` and make them configurable

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -133,3 +133,75 @@ input.id = tbl(bety,"inputs") %>% filter(name == input_name) %>% pull(id)
...
+earthengine authenticate
+```
+
+#### Usage guide:
+This module contains GEE and other processing functions which retrieve data from a specific dataset and source. For example, `gee2pecan_s2()` downloads bands from Sentinel 2 using GEE and `bands2lai_snap()` uses the downloaded bands to compute LAI using the SNAP algorithm. These functions can be used independently but the recommended way is to use `remote_process()` which is a main function that controls all the individual functions to create an organized way of downloading and handling remote sensing data in PEcAn. 
+
+
+**Input data**: The input data must be the coordinates of the area of interest (point or polygon type) and the name of the site/AOI. These information must be provided in a GeoJSON file.
+
+**Output data**: The output data returned by the functions are in the form of netCDF files.
```
istfer, line 179: When we finalize the output data format we can describe it here :+1:

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -133,3 +133,75 @@ input.id = tbl(bety,"inputs") %>% filter(name == input_name) %>% pull(id)
...
+   To install the additional Python dependencies required, 
+a. Navigate to `pecan/modules/data.remote/inst` If you are inside the pecan directory, this can be done by,
+```bash
+cd modules/data.remote/inst
+```
+b. Use pip to install the dependencies.
+```bash
+pip install -r requirements.txt
+```
+3. **Authenticate GEE**. The GEE API needs to be authenticated using your credentials (one time step). The credentials will be stored locally on your system. This can be done by,
```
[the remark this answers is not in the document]
ayushprd: This entire "set-up" is a first time and one time only step. Maybe I should write this at the top.

modules/data.remote/inst/gee2pecan_l8.py
```diff
@@ -0,0 +1,180 @@
...
+    
+    outdir (str) -- path to the directory where the output file is stored. If specified directory does not exists, it is created.
+  
+    start (str) -- starting date of the data request in the form YYYY-MM-DD
+    
+    end (str) -- ending date areaof the data request in the form YYYY-MM-DD
+
+    qc (bool) -- uses the cloud masking function if set to True
+
+    bands (list of str) -- bands to be retrieved. Default: B5, B4
```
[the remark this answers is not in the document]
ayushprd: This is also a remnant from the previous version forgot to remove..

modules/data.remote/inst/gee2pecan_l8.py
ayushprd: So I have 2 options here,
1. I only make it configurable in l8. In this case, I will have to add an additional `if` in `get_remote_data` which would check if the scale is passed in the arguments and then call the appropriate function.
2. I can also make it configurable in s2 even though it's not required (currently). This will not require adding the above `if`. 
BTW AppEEARS doesn't have this option to specify the scale. There we need to specify the type of projection required

istfer: 2nd option sounds good to me as well

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -133,3 +133,78 @@ input.id = tbl(bety,"inputs") %>% filter(name == input_name) %>% pull(id)
...
+
+* [Sentinel-2 MSI](https://developers.google.com/earth-engine/datasets/catalog/COPERNICUS_S2_SR) [`gee2pecan_s2()`](https://github.com/PecanProject/pecan/blob/develop/modules/data.remote/inst/gee2pecan_s2.py)
+* [SMAP Global Soil Moisture Data](https://developers.google.com/earth-engine/datasets/catalog/NASA_USDA_HSL_SMAP_soil_moisture) [`gee2pecan_smap()`](https://github.com/PecanProject/pecan/blob/develop/modules/data.remote/inst/gee2pecan_smap.py)
+* [Landsat 8 Surface Reflectance](https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC08_C01_T1_SR) [`gee2pecan_l8()`](https://github.com/PecanProject/pecan/blob/develop/modules/data.remote/inst/gee2pecan_l8.py)
+
+Processing functions currently available are,
+
+* [SNAP](https://github.com/PecanProject/pecan/blob/develop/modules/data.remote/inst/satellitetools/biophys_xarray.py)
+
+#### Set-Up instructions (one time step):
```
istfer, line 153:
```suggestion
#### Set-Up instructions (first time and one time only):
```

PR #2647, opened 2020-07-02
WIP Add working Docker ED2 test xml

KristinaRiemer:
Now that everyone is able to get the Docker stack up and running, a next step is making sure there are tests for all models that work in that environment. I'm focusing on ED2 because we're using that model for a project. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/tests/docker.ed2.xml
+++ b/tests/docker.ed2.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0"?>
+<pecan>
+  <outdir>/home/carya/ed2</outdir>
+
+  <database>
+    <bety>
+      <driver>PostgreSQL</driver>      
+      <user>bety</user>
+      <password>bety</password>
+      <host>postgres</host>
+      <dbname>bety</dbname>
+      <write>FALSE</write>
+    </bety>
+    <dbfiles>/home/carya/ed2/dbfiles</dbfiles>
+  </database>
+
+  <pfts>
+    <pft>
+      <name>ebifarm.c4grass.doe_vd</name> 
+    </pft>
+  </pfts>
+
+  <meta.analysis>
+    <iter>3000</iter>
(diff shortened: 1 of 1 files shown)


infotroph:
@KristinaRiemer @dlebauer is this PR ready for review? It looks good to me and has no conflicts or build messages.

dlebauer:
Next step @KristinaRiemer compare with working .xml file

KristinaRiemer:
I think this looks good compared to one of the ED2 xml files that we've used for runs. The only thing with this is that the date range for the met is much wider than for the model run itself. I'm not sure where that came from, and does it make more sense for them to be the same start and end dates?

dlebauer:
It makes sense to align the dates.

tests/docker.ed2.xml
```diff
@@ -0,0 +1,84 @@
...
+        <output_month>12</output_month>
+      </ed_misc>
+    </config.header>
+    <phenol.scheme>0</phenol.scheme>
+  </model>
+
+  <run>
+    <site>
+      <id>76</id>
+      <met.start>2004-01-01 00:00:00</met.start>
```
dlebauer, line 60:
```suggestion
      <met.start>2005-12-01 00:00:00</met.start>
```

tests/docker.ed2.xml
```diff
@@ -0,0 +1,84 @@
...
+      <met.end>2009-12-31 23:59:59</met.end>
```
dlebauer, line 61:
```suggestion
      <met.end>2007-01-31 23:59:59</met.end>
```

dlebauer (approved):
Reduced met time range closer to model run window. Conservatively left a month buffer on either side to avoid potential time zone or leap year issues.

PR #2659, opened 2020-07-16
AppEEARS download function

ayushprd:
## Description
This PR adds the function which can be used for downloading remote sensing data from AppEEARS.

## Motivation and Context
To be used in the remote data module.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/data.remote/inst/remote_process.py
```diff
@@ -41,17 +42,19 @@ def remote_process(
...
     scale (int) -- pixel resolution, None by default, recommended to use 10 for Sentinel 2
 
+    projection (str) -- type of projection. Only required for appeears polygon AOI type. None by default. 
+
     qc (float) -- quality control parameter, only required for gee queries, None by default
 
     algorithm (str) -- algorithm used for processing data in process_data(), currently only SNAP is implemented to estimate LAI from Sentinel-2 bands, None by default
 
-    output (dict) -- "get_data" - the type of output variable requested from get_data module, "process_data" - the type of output variable requested from process_data module
+    process_data (str) -- the type of output variable requested from process_data module
```
ayushprd, line 57: I have also removed the `output` dict and replaced it with `process_data` because the `get_data` variable wasn't being used anywhere other than in `process_remote_data` and I felt that it was a bit confusing.

modules/data.remote/inst/appeears2pecan.py
```diff
@@ -0,0 +1,206 @@
...
+    # if the requested product is one of the SMAP products then select the first 25 layers
+    if product in [
+        "SPL3FTP.002",
+        "SPL3SMP.006",
+        "SPL3SMP_E.003",
+        "SPL4CMDL.004",
+        "SPL4SMGP.004",
+    ]:
+        # change this part to select your own SMAP layers
+        prodLayer = prodLayer[0:25]
```
istfer, line 102: this bit needs to throw a warning at least that subsetting happens

modules/data.remote/inst/remote_process.py
istfer: I see what you mean, since we changed the logic for gee function names to use collection names instead :+1: 

But I was wondering if we are losing some generality this way, now you hardcode `bands` in the `process_remote_data`, it's true that we only have one processing function (`bands2lai_snap`) for now, but it may not be the case in the future (just making this up, `images2ndvi_algX`). I can't think of an example for now, but for some future remote sources we might also need the original logic in case we need to formulate functions like `source2pecan_output$get_data`. So I would vote for keeping it.

ayushprd: That's right. Ok, then what should be the value of `get_data` say in case of AppEEARS?

modules/data.remote/inst/appeears2pecan.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        uses the user provided NASA Earthdata credentials to request a token.
+        
+        Returns
+        -------
+        head (dict) : header contatining the authentication token 
+        """
+        try:
+            # if the user does not want to enter their credentials everytime they use this function, they need to store their username and password in a JSON file, preferabbly not in a git initialized directory
+            with open("path/to/cred.json", "r") as f:
```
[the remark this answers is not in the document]
ayushprd: Yes, currently that needs to be done. Passing it as an argument is an option. If the user knows where to store the file then maybe I can also create the JSON file automatically once the user enters their credentials for the first time and then use that for subsequent runs.

modules/data.remote/inst/remote_process.py
istfer: Could the default values of this list for both levels be `None`? Since the current implementation of AppEEARS also doesn't use `process_data` Or it can always stay as `bands` by default.. I was thinking of providing the flexibility that users pass these via xml

ayushprd: Alright, I'll change it back and keep it as None by default.

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -164,23 +169,24 @@ b. Use pip to install the dependencies.
...
 
 #### Usage guide:
 This module contains GEE and other processing functions which retrieve data from a specific dataset and source. For example, `gee2pecan_s2()` downloads bands from Sentinel 2 using GEE and `bands2lai_snap()` uses the downloaded bands to compute LAI using the SNAP algorithm. These functions can be used independently but the recommended way is to use `remote_process()` which is a main function that controls all the individual functions to create an organized way of downloading and handling remote sensing data in PEcAn. 
 
 
 **Input data**: The input data must be the coordinates of the area of interest (point or polygon type) and the name of the site/AOI. These information must be provided in a GeoJSON file.
 
-**Output data**: The output data returned by the functions are in the form of netCDF files.
+**Output data**: The output data returned by the GEE are in the form of netCDF files. When using AppEEARS, output is in the form of netCDF files if the AOI type is a polygon and in the form of csv files if the AOI type is a point.
```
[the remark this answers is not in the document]
ayushprd: Do you mean I should convert csv to netCDF? Also currently gee output files do not follow this naming convention if this is final, I'll update the functions to follow this.

modules/data.remote/inst/appeears2pecan.py
ayushprd: Not related to this specific instance, once a question which I have is would the print statements, warnings etc from python be visible when executing the pecan workflow? Because I can't use the pecan logger functions.

book_source/03_topical_pages/09_standalone_tools.Rmd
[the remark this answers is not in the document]
istfer: No worries, I'm aware that gee output doesn't have this convention because naming convention, formats and DB interaction is the next step

Our load_data functions can work with both csv and netcdf, maybe it is not important that they share the same file format at this point, we can fix it later if it turns out we need them to have the same format

modules/data.remote/inst/appeears2pecan.py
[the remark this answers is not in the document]
istfer: Yeah I think any solution that doesn't require touching code would be better. Otherwise, this also makes other kinds of small nuisance possible. e.g. others to commit their hardcoded paths by mistake etc.

modules/data.remote/inst/appeears2pecan.py
istfer: Good question, I don't know really, does reticulate pass along logs etc?

ayushprd: as far as I know, print statements can be passed but not logs. This will make it very difficult to debug when something goes wrong.

ayushprd:
Looks like CI check wasn't able to access some required URL

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -208,3 +214,28 @@ remote_process(
...
+
+1. Open a Python shell at `pecan/modules/data.remote/inst`
+
+2. In the Python shell run the following,
+```python
+# import remote_process
+from remote_process import remote_process
+
+# call remote_process 
+remote_process(
```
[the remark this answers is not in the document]
ayushprd: 20 mins

Issue #2657, opened 2020-07-15
Workflow submission via API Endpoint


tezansahu:
## Description

Submit a PEcAn workflow to get executed using a dedicated PEcAn API endpoint

## Proposed Solution

- Create a `POST api/workflows/` endpoint to allow users to submit PEcAn workflows in the request body
- Cater to workflows submitted as XML & JSON using separate functions
- Submit XML workflows to RabbitMQ after storing them in the database

ashiklom:
I can confirm that this works (_really_ well -- nice work!), at least for submitting workflows based on full XML files.

I am inclined to close this and make the other aspects of the functionality (e.g. workflow from JSON) separate and lower-priority issues.

tezansahu:
The API endpoints for submitting workflow as a JSON have already been developed & merged into `develop` branch by the PR #2674.
Basically it hits the same `POST /api/workflows/` endpoint, but this time with `Content-Type = "application/json"` instead of "application/xml" (which was for the XML file submission)

PR #2662, opened 2020-07-22
XML Workflow submission API endpoint

tezansahu:
- Created the endpoint for submitting a PEcAn Workflow in XML format
- Modified the authentication filter to forward the `userid` to endpoints for use
- Updated the tests, swagger API documentation & PEcAn book documentation

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


apps/api/pecanapi-spec.yml
```diff
@@ -335,29 +329,113 @@ paths:
...
-          application/json:
-            schema:
-              $ref: '#/components/schemas/Workflow'
+          # application/json:
+          #   schema:
+          #     $ref: '#/components/schemas/Workflow'
           application/xml:
             schema:
-              $ref: '#/components/schemas/Workflow'
+              xml:
```
robkooper, line 341: it is a good idea to have these schema's defined at the bottom of the swagger file. This way you can reference them in multiple places. (same goes for responses).

base/remote/R/rabbitmq.R
```diff
@@ -51,9 +51,10 @@ rabbitmq_parse_uri <- function(uri, prefix="", port=15672) {
 #' @param auth authentication for rabbitmq in httr:auth
 #' @param body the actual body to send, this is a rabbitmq message.
 #' @param action the rest action to perform
+#' @param silent boolean to indicate if logging should be performed.
```
robkooper, line 54: thanks for fixing this.

apps/api/pecanapi-spec.yml
tezansahu: Yeah...I will do that...coz the same schema will be repeated for the `application/json` as well

apps/api/R/submit.workflow.R
```diff
@@ -0,0 +1,180 @@
...
+  
+  # Add entry to workflows table in database
+  workflow_id <- insert.workflow(workflowList)
+  workflowList$workflow$id <- workflow_id
+  
+  # Add entry to attributes table in database
+  insert.attribute(workflowList)
+  
+  # Fix the output directory
+  outdir <- paste0("/data/workflows/PEcAn_", workflow_id)
```
mdietze, line 55: I understand not wanting to let the external user to specify the outdir in their XML (security nightmare and external user doesn't need/want to know where files are stored), but this hardcoded path is going to end up putting PEcAn output in places the host system doesn't want and can't control. e.g. on my pecan server all the runs go to a specific disk mount, the defaults for which are currently stored in the config.php (and we've written R functions that other apps [e.g. Shiny] are using to access info in that config).

tests/api.sipnet.xml
```diff
@@ -0,0 +1,46 @@
...
+    <random.effects>FALSE</random.effects>
+    <threshold>1.2</threshold>
+    <update>AUTO</update>
+  </meta.analysis>
+
+  <ensemble>
+    <variable>NPP</variable>
+  </ensemble>
+
+  <sensitivity.analysis>
```
mdietze, line 20: Probably don't want your test performing complicated, time-consuming analyses such as this sensitivity analysis or the meta.analysis above

tests/api.sipnet.xml
```diff
@@ -0,0 +1,46 @@
...
+    <quantiles>
+      <sigma>-1</sigma>
+      <sigma>1</sigma>
+    </quantiles>
+    <variable>NPP</variable>
+  </sensitivity.analysis>
+
+  <model>
+    <type>SIPNET</type>
+    <revision>r136</revision>
```
mdietze, line 30: probably do want to use a more up-to-date version of SIPNET

tests/api.sipnet.xml
```diff
@@ -0,0 +1,46 @@
...
+      <id>772</id>
+    </site>
+    <inputs>
+      <met>
+        <id>99000000003</id>
+      </met>
+    </inputs>
+    <start.date>2002-01-01 00:00:00</start.date>
+    <end.date>2005-12-31 00:00:00</end.date>
+    <dbfiles>pecan/dbfiles</dbfiles>
```
mdietze, line 44: per previous comment about not letting external users set paths on the host system, you probably want to take away this flag as something that's read and parsed and make sure your API is setting it correctly internally.

More broadly, you probably need to take a very close look at all the 'advanced' workflow options and think carefully about if/how we're going to allow the "power user" to to set paths when they're submitting to their own machine while sanitizing those paths out of the generic user's XML.

tests/api.sipnet.xml
tezansahu: Sure...is it the `ssr` revision? Or something else?

tests/api.sipnet.xml
tezansahu: Yeah...I can leave those out...no problems

apps/api/R/submit.workflow.R
tezansahu: I had a discussion with @robkooper about this and we mutually came to a conclusion to go ahead in this way. Maybe I'll talk to him again about how we could modify it if needed.

tests/api.sipnet.xml
mdietze: Good question. I just checked https://github.com/PecanProject/sipnet/releases and didn't realize that we haven't tagged a new release in over a year. For testing purposes I think we can leave at r136 for now, but the master branch has some fixes that folks should be using for science-focused runs.

tezansahu: Cool...So I'll leave it that way for now...n once the newer version is tagged and added to the db, we can modify the test file as required.

apps/api/R/submit.workflow.R
mdietze: Yeah, I think it would be really helpful for the host to be able to (optionally) set system-specific configurations like the root dirs for inputs, outputs, etc.

tests/api.sipnet.xml
robkooper: and we need to build a container with that later release.

apps/api/R/submit.workflow.R
tezansahu: I have created an env var to sort this after discussing with @robkooper

tests/api.sipnet.xml
tezansahu: I have created an env var to sort this after discussing with @robkooper

tezansahu:
#2657 addressed

PR #2663, opened 2020-07-24
Download SMAP data into R from GEE 

julbateman:
## Description
 Added `download_SMAP_gee2pecan.R` that uses `gee2pecan_smap.py` to get SMAP data into R from GEE. Additionally, I made a minor change to `gee2pecan_smap.py` that provides the observation dates in the netCDF output (as per @ayushprd 's advice).

## Motivation and Context

To be used in the data.remote module.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


julbateman:
@mdietze For some reason I couldn't add you as a reviewer on this PR

modules/data.remote/inst/gee2pecan_smap.py
```diff
@@ -1,23 +1,22 @@
...
+import geopandas as gpd
 import os
 import xarray as xr
 import datetime
 
 ee.Initialize()
 
 
-def gee2pecan_smap(geofile, outdir, start, end):
+def gee2pecan_smap(geofile, outdir, start, end, var):
```
ayushprd, line 19: @julbateman unless this is intentional, I think you have the old version of `gee2pecan_smap.py`. In the current version, we discontinued the option for the user to specify their choice of variable and are now downloading all variables/bands/layers by default. We also make use of utility functions. If there isn't any objection can you please make use of the current code? Otherwise, this would break when used via `remote_process()`

https://github.com/PecanProject/pecan/blob/develop/modules/data.remote/inst/gee2pecan_smap.py

Btw so nice to see that this code could be of some use :)

julbateman: Sorry about that! I'll switch it so that it uses the new version. 👍🏼

mdietze:
@julbateman why was this PR closed?

julbateman:
@mdietze I was going to change a few things since there's an updated `gee2pecan_smap` I didn't notice, was going to make the changes  today and then submit a new PR for it to make sure it works properly

modules/data.remote/inst/download_SMAP_gee2pecan.R
```diff
@@ -0,0 +1,107 @@
...
+  site_GeoJSON <- data.frame(site_info$lon, site_info$lat) %>%
+    setNames(c("lon","lat")) %>% 
+    leafletR::toGeoJSON(name = site_info$site_name, dest = geoJSON_outdir, overwrite = TRUE) %>%
+    rgdal::readOGR()
+  site_GeoJSON$name = site_info$site_name
+  site_GeoJSON = site_GeoJSON[-1] %>%
+    leafletR::toGeoJSON(name = site_info$site_name, dest = geoJSON_outdir, overwrite = TRUE)
+  
+  # Locate gee2pecan_smap.py function and load into R
+  script.path = file.path("/fs/data3/jbateman/pecan/modules/data.remote/inst/gee2pecan_smap.py")
```
mdietze, line 64: @julbateman you can't hard code paths like this, as this won't work for everyone else. You should use `system.file` to get the path of that specific file in that specific package wherever the package happens to be installed.

PR #2664, opened 2020-07-24
NEON soil data download

julbateman:
## Description
 This PR adds the `download_NEON_soildata` function to get SWC and SIC data from [https://data.neonscience.org/data-products/DP1.00094.001](url) using the `neonUtilities` package.

## Motivation and Context
 To be used in data.land module.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


julbateman:
@mdietze Again, for some reason I can't edit the requested reviewers (weird, not really sure why)

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
+##' @name download_NEON_soildata
+##' @description: 
+##' Download NEON Soil Water Content and Soil Salinity data by date and site name
+##' 
+##' @param site four letter NEON site code name(s). If no site is specified, it will download all of them (chr) (e.g "BART" or multiple c("SRER", "KONA", "BART"))
+##' @param avg averaging interval: either 30 minute OR 1 minute, default will return BOTH (chr) (e.g 1 or 30)
```
mdietze, line 6: if the options are 1 and 30, why are they being passed as character as numeric? And when I look at the code that parses this arg, it's clear that you're making a numeric comparison to the numbers 1 and 30, not the strings "1" and "30". Also, you say 1 or 30 are the options but the default is "all" so clearly 1 and 30 are not the only options.

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
...
+##'   outdir = getwd())}
+
+## Install NEON libs
+#devtools::install_github("NEONScience/NEON-geolocation/geoNEON")
+#devtools::install_github("NEONScience/NEON-utilities/neonUtilities", force = TRUE)
+#install.packages("BiocManager")
+# BiocManager::install("rhdf5")
+
+
+download_NEON_soildata <- function(site, avg = "all", var,
```
mdietze, line 32: Everywhere else in PEcAn "soil data" typically refers to info on soil texture, or texture and biogeochemical pools, so having a function that grabs soil moisture called "soil data" will be confusing, especially if the function _can't_ also download texture. What if you just change to `download_NEON_soilmoisture`?

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
...
+                                   startdate = NA, enddate = NA,
+                                   outdir) {
+  
+  
+  #################### Data Download from NEON #################### 
+  soil.raw = neonUtilities::loadByProduct(dpID = "DP1.00094.001", site = "KONA", avg = avg, startdate = startdate, enddate = enddate, check.size = FALSE)
```
mdietze, line 38: Why is there a specific site being hardcoded here, rather than the sites passed in as arguments?

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
...
+    data.raw = soil.raw$SWS_1_minute
+  } else {
+    data.raw = list(soil.raw$SWS_1_minute, soil.raw$SWS_30_minute)
+  }
+
+  
+  # Separate selected variable, omit NA values
+  if (var == "SWC") {
+    data.raw = (split(data.raw, data.raw$VSWCFinalQF))$'0'
+    data.raw = data.raw[,-c(15:22)] %>% na.omit()
```
mdietze, line 53: hard coding column ranges is dangerous, as file formats can change easily. Could you select by name instead?

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
...
+  } else if (var == "SIC") {
```
mdietze, line 54: So it appears that you download both SWC and SIC, and then you drop one of them. If it's not possible to only download one, then you should at least have this function allow a way to keep both (so that if you want both you don't need to download twice), but you might also consider just always returning both.

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
...
+  
+  # Load individual site list into the Global Environment
+  data.sitePositions %>%
+    list2env(envir = .GlobalEnv)
+  
+  # Save site lists as .rds files to outdir 
+  for (i in names(data.sitePositions)){
+    #rlist::list.save(data.sites[[i]], file = paste0(outdir, "/", i, ".json"), type = tools::file_ext(".json"))
+    #write.csv(data.sites[[1]], paste0(outdir, "/", i,".csv"))
+    saveRDS(data.sitePositions[[i]], file = paste0(outdir, "/", i, ".rds"))
```
mdietze, line 79: We should discuss the preferred file format here. The pattern elsewhere is to either (A) save data in an agreed upon standard (e.g. netcdf CF for met) or (B) save in a format that load_data can read automatically (which usually means netCDF or csv). In either case, if there's a new data format that PEcAn is using it needs to be registered with the Formats table in the database.

modules/data.land/R/download_NEON_soildata.R
```diff
@@ -0,0 +1,85 @@
...
+  }
+
```
mdietze, line 81: Function isn't returning ANYTHING. It should either return something like a file table like we do in met.process or informative error information if the download failed. As far as I can tell if the download fails the function will probably just fail internally without any error messages, but if it doesn't then the rest of the workflow would just continue without any way to catch that the files are not there.

PR #2665, opened 2020-07-27
Endpoints for PFT details & plotting of run outputs

tezansahu:
- Endpoints for searching & retrieving PFT details
- Added Modeltype Formats information to the response of `/api/models/{model_id}`
- Added details of run outputs to the response of `api/runs/{run_id}` if the outputs folder exists on the host
- Created endpoint for plotting results of a PEcAn run

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


apps/api/R/models.R
```diff
@@ -19,21 +19,31 @@ getModel <- function(model_id, res){
   
   qry_res <- Model %>% collect()
   
-  PEcAn.DB::db.close(dbcon)
-  
   if (nrow(qry_res) == 0) {
+    PEcAn.DB::db.close(dbcon)
```
mdietze, line 23: You might want to just use `on.exit` to specify the db.close right when you open it to make sure that close happens even if any other sort of error occurs

apps/api/R/runs.R
```diff
@@ -102,6 +106,102 @@ getRunDetails <- function(id, res){
...
+#' @param yvar the variable to plot along the y-axis.
+#' @param xvar the variable to plot along the x-axis, by default time is used.
+#' @param width the width of the image generated, default is 800 pixels.
+#' @param height the height of the image generated, default is 600 pixels.
+#' @return List of runs (belonging to a particuar workflow)
+#' @author Tezan Sahu
+#* @get /<run_id>/graph/<year>/<y_var>
+#* @serializer contentType list(type='image/png')
+
+plotResults <- function(run_id, year, y_var, x_var="time", width=800, height=600, res){
```
mdietze, line 139: Not that there's anything 'wrong' with this, but I'd say that grabbing/downloading the model output itself is going to be FAR more important to most users than downloading an image file. Probably grabbing the whole netCDF would be the best place to start, but I could see a lot of folk wanting to subset variables, locations, data ranges, and ensemble members. The latter support it built into THREDDS, but (1) not all of our PEcAn servers support THREDDS (but maybe the right thing to do is to fix that) and (2) regardless we should have some helper functions in the R package to help people download output rather than having to learn to work with THREDDS directly (which is nontrivial)

I think there's also a REALLY high need, even if it's just internal to the project, to be able to correctly grab INPUTS and POSTERIORS via the API. The former will be pretty important to a lot of enduser analyses as well (e.g. plot predicted fluxes vs meteorological inputs)

apps/api/R/models.R
tezansahu: Yeah sure...that's a good idea...will try n incorporate this for all endpoints

robkooper: I think the problem is that the application does not exit, but stays open. We just need to make sure we close teh connections.

apps/api/R/runs.R
robkooper: This function is needed in the current web application. The current web application can not create the graphs we had in the VM version. This will allow me/us to call this endpoint to get an image back.

robkooper: As for the ability to download the outputs/inputs/posteriors I agree, we just have to find the right functions, but I think we have a lot of this framework in place.

mdietze: Ahh, yes that makes perfect sense.
As we discussed during last week's planning meeting, we should think about whether it makes sense to rework the whole PHP interface to use the API rather than it's current approach of direct execution of R code.

robkooper: shhhhh, but that is my plan :) and maybe even use R and leaflet or other things.

Issue #2646, opened 2020-06-30
met2model.ED2 overwites CO2 levels set in ED2IN, possibly also XML


mccabete:
### Bug Description
If you set the `INITIAL_CO2` via the ED2IN, the level of CO2 will be overwritten by the value of CO2 supplied by met2model.ED2. 

Because many met products do not have atmospheric CO2 levels, and ED2 requires atmospheric CO2 be set, the met2model.ED2()[ checks if CO2 exists in the met driver, and if not, adds a set value (380 ppm)](https://github.com/PecanProject/pecan/blob/develop/models/ed/R/met2model.ED2.R#L344). 

This bug is tricky to catch until after ED2 runs and you start plotting the output. 
## Expected behavior

ideally that any pre-set filler values are filled in only if no user-provided value exists

## Additional context

 I suspect that the met2model.ED2() value would override an XML supplied value, but I haven't tested it.

mccabete:
@bcow @istfer @ashiklom, Are any of you feeding initial CO2 values to ED2 via the XML, an do you know if they get overwritten to 380ppm?

mccabete:
OK, looks like CO2 isn't part of the config.xml, but I could imagine incorporating it.

mccabete:
Possible solutions:
- Get rid of the met2model.ED2 check altogether. The ED2IN sets Initial CO2 to 370 by default anyway. (There might be a different reason for the met2model.ED2 check?) 
- Put a NULL value in the ED2IN template, have the config.xml be populated from the ED2IN or by the user, put a check in met2model.ED2 to see if there is a user-supplied value, and if not, supply 380
- add a warning to the met2model.ED2 check

istfer:
Sorry @mccabete I was not initializing CO2 via ED2IN

mccabete:
Hey @istem I'm pretty sure no one does
________________________________

Aariq:
looks like this was closed by #2667

PR #2667, opened 2020-07-30
Deleting met2model CO2 check

mccabete:
## Motivation and Context

Met2model will check for CO2 levels in the met product, and if one isn't provided, will fill in a CO2 value. This is unnecessary because  ED2 has a default CO2 level that is set in the ED2IN, and ED users may intentionally be setting the CO2 levels in the ED2IN. This PR deletes the CO2 check. 

 This resolves issue #2646 
## Review Time Estimate

- [] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


mdietze:
This change is causing the following test to fail
```
check_ed_metheader.R:77: failure: All variables present in metfile
all(variables$variable %in% hfile$ls()$name) isn't true
```
which I suspect is occurring because co2 is in the list of variables that's being tested for.

mccabete:
@mdietze took slightly longer than I expected but this is tested now

Issue #2666, opened 2020-07-28
Registration of gSSURGO soil.nc files


mdietze:
### Bug Description

soil.nc files generated by @para2x as part of the NASA CMS SDA workflow didn't get registered in BETY. Therefore there are two tasks that need to happen

1. Existing soil.nc (ensemble) files need to registered with BETY. Hamze thinks we should be able to find most of soil files on the BU server at `/fs/data1/pecan.data/dbfiles` with this format name `gSSURGO_site_xxxx`

2. The `soil_process` workflow needs to be debugged to figure out why the files didn't get registered in the first place

mdietze:
@kzarada I was hoping to close this issue so I wanted to check in on step 1 of this. I was looking at the database and there haven't been any new files registered in the pecan standard soil format. At the same time, there's a new gSSURGO soil format type that's been created, and 45 files registered with that format. This raises a few questions:
1. Are the gSSURGO files really in a different format from the pecan standard, or was that format record created by mistake?
2. Where are the other >450 sites?

kzarada:
1. I did a new one because when I searched 'soil' in BETY inputs, all of them but 1 had their own formats. I don't know if if they are in the pecan standard. Hamze's script doesn't specify if they are put in that but I can change the format if you would rather. 

2. I moved all the gSSURGO files that were in the /fs/data1/pecan.data/dbfiles. It was just those 45 sites.

PR #2668, opened 2020-07-31
Updating gSSURGO download to register files (Issue #2666) 

kzarada:
## Description
Updated gSSURGO file download to include registering files in BETY. Changed db.input.insert to work with input files that don't have a start and end date (like the soil files, updated soil.process function to work with updated gSSURGO file download

## Motivation and Context
 Addressing issue #2666  

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


base/db/R/dbfiles.R
```diff
@@ -109,7 +111,13 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
   if (is.null(inputid)) {
     # Either there was no existing input, or there was but the dates don't match and
     # allow.conflicting.dates==TRUE. So, insert new input record.
-    if (parent == "") {
+    # adding is.null(startdate) to add inputs like soil that don't have dates
+    if(parent == "" && is.null(startdate)) {
```
mdietze, line 115: This logic looks like it would miss the case where startdate is null but parent isn't blank

modules/data.land/R/extract_soil_nc.R
```diff
@@ -237,6 +237,33 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
+                              user     = dbparms$bety$user, 
+                              password = dbparms$bety$password)
+  con <- bety$con
+  # register files in DB 
+  for(i in 1:length(out.ense)){
+    in.path = paste0(dirname(out.ense[i]$path), '/')
+    in.prefix = stringr::str_remove(basename(out.ense[i]$path), ".nc")
+    
+    
+    dbfile.input.insert (in.path,
```
mdietze, line 256: I think we want to avoid putting the DB insert inside of the extract function. First, it violates our existing design decision to separate these processing operations from the DB inserts and makes it impossible to run the function remotely. Second, it goes against our recent discussion about making it easier to use PEcAn functions independent from the DB. The DB insert should be occurring in soil_process, which already has the settings object and an open con.

PR #2671, opened 2020-08-02
Implement dependency updates in Actions CI

infotroph:
## Description
Motivated by seeing #2664 [pass on Actions](https://github.com/PecanProject/pecan/pull/2664/checks?check_run_id=937721036) but [fail on Travis](https://travis-ci.org/github/PecanProject/pecan/jobs/714223471), and therefore realizing this check wasn't being done on Actions.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
       with:
         key: install-${{ github.sha }}
         path: .install
+    - name: update dependency lists
+      run: Rscript scripts/generate_dependencies.R
     - name: build
       run: make -j1
       env:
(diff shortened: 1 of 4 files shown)


rahul799:
I won't feel It's worth adding that check in the styler workflow as I think that the main workflow will always do the check. So it's just like doing the same check twice. @infotroph What do you think?

infotroph:
> I won't feel It's worth adding that check in the styler workflow as I think that the main workflow will always do the check. So it's just like doing the same check twice. @infotroph What do you think?

@rahul799 My thought was that if we include it here then any changes get committed back, so a PR with out-of-date dependencies could be fixed by simply commenting `/style` on it. Would it make more sense to implement it as a separate command?

Also FWIW, `generate_dependencies.R` takes only a few seconds to run and has no installation overhead beyond the packages needed for the build, so I feel OK about rerunning it if that adds a little convenience to the process.

PR #2672, opened 2020-08-03
Integrating remote_process with PEcAn

ayushprd:
## Description
This PR adds `call_remote_process.R` which would control the Python remote_process for downloading, processing and storing remote sensing data in PEcAn. With the current implementation, only `call_remote_process.R` has PEcAn dependencies, `remote_process` can be used on its own apart from the workflow.

## Motivation and Context
This PR connects all the GEE and AppEEARS scripts.

@istfer I still haven't edited the workflow.R or add dependencies to docker, I will do these as soon as the overall idea of  `call_remote_process` gets reviewed.

## XML tags required in settings
```
<remotedata>
<source>gee</source>
<collection>COPERNICUS/S2_SR</collection>
<scale>10.0</scale>
<projection><projection/>
<qc>1.0</qc>
<algorithm>snap</algorithm>
<credfile></credfile>
<out_get_data>bands</out_get_data>
<out_process_data>lai</out_process_data>
<raw_id></raw_id>
<raw_path></raw_path>
<pro_id></pro_id>
<pro_id></pro_path>
<overwrite></overwrite>
</remotedata>

```
For now please do the following inside the rstudio container:
```
apt-get update
apt-get install python3-pip
pip3 install --upgrade setuptools

navigate to /pecan/modules/data.remote/inst/RpTools
pip3 install -e .
```

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+
+call_remote_process <- function(settings){
+  
+  # information about the date variables used in call_remote_process -
+  # req_start, req_end : start, end dates requested by the user, the user does not be aware about the status of the requested file in the DB  
+  # start, end : effective start, end dates created after checking the DB status. These dates are sent to remote_process for downloading and processing data
+  # write_raw_start, write_raw_end : start, end dates which are used while inserting and updating the DB
+  # the "pro" version of these variables have the same meaning and are used to refer to the processed file
+
+  reticulate::import_from_path("remote_process", file.path("..", "inst"))
```
ayushprd, line 22: This is wrong and will not work from when being called from anywhere else but the only other way I have been able to get this work is by directly specifying the absolute path `/pecan/modules/data.remote/inst` which I think is also not appropriate? Using `file.path(system.file("remote_process.py", package = "PEcAn.data.remote"))` gives me a module not found error. I have no idea why.

modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+  raw_merge <- NULL
+  pro_merge <- NULL
+  existing_raw_file_path <- NULL
+  existing_pro_file_path <- NULL
+  
+  # extract the variables from the settings list
+  siteid <- settings$remotedata$siteid
+  sitename <- settings$remotedata$sitename
+  raw_mimetype <- settings$remotedata$raw_mimetype
+  raw_formatname <- settings$remotedata$raw_formatname
```
mdietze, line 37: Things like mimetype and formatname are going to be fixed for a specific data source, and thus shouldn't need to be provided by the user (this would just be a potential source of error and makes it harder for users as they have to look up info that only has one 'right' answer). In met we solve this problem by putting simple "registry" metadata files for each data type/source in the package itself.

mdietze: Also, if you're adding a bunch of stuff to the Settings, you definitely need to update the PEcAn documentation to document any new settings options.

modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+  algorithm <- settings$remotedata$algorithm
+  credfile <- settings$remotedata$credfile
+  pro_mimetype <- settings$remotedata$pro_mimetype
+  pro_formatname <- settings$remotedata$pro_formatname
+  out_get_data <- settings$remotedata$out_get_data
+  out_process_data <- settings$remotedata$out_process_data
+  
+  
+  dbcon <- PEcAn.DB::db.open(settings$database$bety)
+  flag <- 0
```
mdietze, line 64: there should be an on.exit(db.close()) right here, rather than having the db.close at the end of the function (which you'll never get to if you exit prematurely).

modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+
+  # collection dataframe used to map Google Earth Engine collection names to their PEcAn specific names
+  collection_lut <- data.frame(stringsAsFactors=FALSE,
+                          original_name = c("LANDSAT/LC08/C01/T1_SR", "COPERNICUS/S2_SR", "NASA_USDA/HSL/SMAP_soil_moisture"),
+                          pecan_code = c("l8", "s2", "smap")
+  )
+  getpecancode <- collection_lut$pecan_code
+  names(getpecancode) <- collection_lut$original_name
```
mdietze, line 72: Not sure why you have two variables (collection_lut and getpecancode) that store the identical information. Also, if you have registry files you could generate tables like this on the fly, rather than having a place developers have to update a hard-coded table every time they add a product.

modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+  if(nrow(existing_data) >= 1){
+    
+    # if processed data is requested, example lai
+    if(!is.null(out_process_data)){
+      
+      # construct processed file name
+      pro_file_name = paste0(sitename, "_", out_process_data, "_", algorithm)
+      
+      # check if processed file exists
+      if(nrow(pro_check <- PEcAn.DB::db.query(sprintf("SELECT inputs.id, inputs.site_id, inputs.name, inputs.start_date, inputs.end_date, dbfiles.file_path FROM inputs INNER JOIN dbfiles ON inputs.name=dbfiles.file_name AND inputs.name LIKE '%s%%';", pro_file_name), dbcon)) == 1){
```
mdietze, line 94: feel like this line is reinventing `dbfile.input.check`

modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+      PEcAn.logger::logger.info("Updating raw file")
+      raw_id = raw_check$id
+      PEcAn.DB::db.query(sprintf("UPDATE inputs SET start_date='%s', end_date='%s', name='%s' WHERE id=%f;", write_raw_start, write_raw_end, output$raw_data_name, raw_id), dbcon)
+      PEcAn.DB::db.query(sprintf("UPDATE dbfiles SET file_path='%s', file_name='%s' WHERE container_id=%f;", output$raw_data_path, output$raw_data_name, raw_id), dbcon)
+    }
+    }
+  }
+  
+  PEcAn.DB::db.close(con=dbcon)
+
```
mdietze, line 299: Function exits without returning updated `input` information to the settings file (i.e. the input ids and file paths to the data that's been downloaded). Take a look at the end of `met.process` for an analogy

Also, function appears to strictly assume that the processing is running locally, and doesn't consider the case where the data needs to be downloaded and processed directly onto a remote machine (e.g. HPC).  Similarly, it's not clear how this function behaves if the user is trying to download additional data to append onto the end of an existing time-series. Take a look at `convert.inputs` and `remote.execute.R`.

modules/data.remote/R/call_remote_process.R
ayushprd: I don't think so because here as you can see, I'm searching using the filename which contains more info about the scale, projection and qc. Using `dbfile.input.check` wouldn't fulfil my motive unless a new format has to be created whenever scale, qc, etc are changed.

modules/data.remote/R/call_remote_process.R
ayushprd: > Similarly, it's not clear how this function behaves if the user is trying to download additional data to append onto the end of an existing time-series.

I have now added some documentation about this which I still have to extend.

> Whenever a data product is requested the output files are stored in the inputs table of BETYdb. Subsequently when the same product is requested again with a different date range but with the same qc, scale, projection the previous file in the db would be extended. The DB would always contain only one file of the same type.
As an example, if a file `Reykajvik_gee_s2_10.0_NA_1.0_200802174519.nc` containing Sentinel 2 bands for start date: 2018-01-01, end date: 2018-06-30 exists in the DB and the same product is requested again for a different date range one of the following cases would happen,

>1. New dates are ahead of the existing file: For example, if the requested dates are start: 2018-10-01, end: 2018-12-31 in this case the previous file will be extended forward meaning the effective start date of the file to be downloaded would be the day after the end date of the previous file record, i.e. 2018-07-01. The new and the previous file would be merged and the DB would now be having data for 2018-01-01 to 2018-12-31.

>2. New dates are preceding of the existing file: For example, if the requested dates are start: 2017-01-01, end: 2017-06-30 in this case the effective end date of the new download would be the day before the start date of the existing file, i.e., 2017-12-31. The new and the previous file would be merged and the file in the DB would now be having data for 2017-01-01 to 2018-06-30. 

>3. New dates contain the date range of the existing file: For example, if the requested dates are start: 2016-01-01, end: 2019-06-30 here the existing file would be replaced entirely with the new file. A more efficient way of doing this could be to divide your request into two parts, i.e, first request for 2016-01-01 to 2018-01-01 and then for 2018-06-30 to 2019-06-30.

ayushprd:
I'm unable to understand what's causing the check to fail. Is it due to reticulate?

modules/data.remote/R/call_remote_process.R
```diff
@@ -0,0 +1,400 @@
...
+##' @export
+##' @param settings PEcAn settings list containing remotedata tags: siteid, sitename, raw_mimetype, raw_formatname, geofile, outdir, start, end, source, collection, scale, projection, qc, algorithm, credfile, pro_mimetype, pro_formatname, out_get_data, out_process_data
+##' @examples
+##' \dontrun{
+##' call_remote_process(settings)
+##' }
+##' @author Ayush Prasad
+##'
+
+call_remote_process <- function(settings){
```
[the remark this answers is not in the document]
ayushprd: I named it like this as I think having two files with the same name would be confusing. This R script just checks the DB status, sets stages, and inserts into the DB. The actual download, merging etc are triggered and handled by the Python `remote_process`. What do you all want this to be named @istfer ? 

I don't follow the second point, Can you explain here why should the user have to decide if its to be overwritten or not? This module checks about the status of the existing file and creates a new file with the same name, different timestamp every time the record is updated but the previous contents are merged in the new file,

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -177,13 +177,26 @@ earthengine authenticate
...
+
+When a processed data product such as SNAP-LAI is requested, the raw product (here Sentinel 2 bands) used to create it would also be stored in the DB. If the raw product required for creating the processed product already exists for the requested time period, the processed product would be created for the entire time period of the raw file. For example, if Sentinel 2 bands are present in the DB for 2017-01-01 to 2017-12-31 and SNAP-LAI is requested for 2017-03-01 to 2017-07-31, the output file would be containing LAI for 2017-01-01 to 2017-12-31.
 
 **Input data**: The input data must be the coordinates of the area of interest (point or polygon type) and the name of the site/AOI. These information must be provided in a GeoJSON file.
 
 **Output data**: The output data returned by the GEE are in the form of netCDF files. When using AppEEARS, output is in the form of netCDF files if the AOI type is a polygon and in the form of csv files if the AOI type is a point.
 
+The output files are created with the following naming convention: sitename_source_collection_scale_projection_qc_TimeStampOfFileCreation
```
ayushprd, line 197: I don't have siteid here because I think that info is redundant and currently, siteid never goes beyond `call_remote_process` and reach the Python functions. But if it's required to make consistent with PEcAn guidelines I will do so.

infotroph:
> I'm unable to understand what's causing the check to fail. Is it due to reticulate?

@ayushprd Try changing 

`reticulate::import_from_path("remote_process", file.path("..", "inst"))` 

to

` remote_process <- reticulate::import_from_path("remote_process", file.path("..", "inst"))`

Caveat: I haven't used reticulate myself, so I could be wrong -- this is just interpolating between the wording of the check message ("no visible global function definition" usually implies R's check code can't figure out where the function is defined) and the usage examples in `?reticulate::import` (which show the results being assigned to names).

modules/data.remote/R/call_remote_process.R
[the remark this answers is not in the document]
istfer: I agree with Ayush that this is a different case than other input processing module as it is written in python. The function that does the real work is justifiably called `remote_process` but it is in python. Someone who has the function arguments ready, can use the module independent of pecan. To integrate these functions with the rest of the workflow and also to make use of our DB functions written in R already, Ayush created this wrapper function which should ideally be called something else.

But I agree with Mike that in the future there might be some functionality that parallelize the calls of `***_process` functions in R (or something like that) and we might want to preserve this naming convention throughout input processing functions. How about we call this R function `remote_process` and the python function `remote_module`?

modules/data.remote/R/call_remote_process.R
istfer: Yes, I agree with Mike that we won't be needing mimetypes and formats specified in the settings tags, I'll give Ayush's code a try and see if we can have registry files or some lookup table sort of thing for this

book_source/03_topical_pages/09_standalone_tools.Rmd
istfer: -Yeah I think we would like to have siteid instead of sitename because it is less ambiguous (also some sites really have long names with spaces in them (e.g. `Ontario - Turkey Point Mature Deciduous (CA-TPD)`) whereas siteid always have an expected form, e.g. `1000000109` turns into `site_1-109` in the filename. 

-Do we really need the time stamp in the filename? 

-Also we can think about the difference between foldername and filename, for example, foldername could be `source_siteid` and the filename can be `collection_scale_projection_qc_siteid`. 
```
GEE_site_1-109
  \_  S2_10-1_site_1-109.nc
  \_  L8_10-1_site_1-109.nc
```
We can play around with these a little

ayushprd:
it worked! @infotroph Thanks a lot

book_source/03_topical_pages/09_standalone_tools.Rmd
ayushprd: >  e.g. 1000000109 turns into site_1-109 in the filename.

Sorry not aware of the logic. How does this happen?

> -Do we really need the time stamp in the filename?

We need something to make the file names unique otherwise when the same data is requested again in the same outdir then the previous file will be directly overwritten before we can do anything with it. I have also tried a different approach - when the merge flag is true, I renamed the original final with some random string before starting the new download and then after merging was completed I changed it back to its original name. This worked fine but I think it's risky because if anything unexpectedly goes wrong during download or merging then the file name in the DB won't be updated and would be pointing towards an incorrect file. I used timestamp because by nature it would always be unique and human-readable but I can use the cryptography functions to generate random strings or anything else as well to make every file name unique.

modules/data.remote/R/call_remote_process.R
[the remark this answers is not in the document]
ayushprd: > in the future there might be some functionality that parallelize the calls of ***_process functions in R (or something like that) 

I just wanted to point out that there are some limitations in parallelizing Python code in R using reticulate https://cran.r-project.org/web/packages/future/vignettes/future-4-non-exportable-objects.html  `remote_process.py` on its own can easily be modified to run and concurrently or parallelly, but I am not sure via R.  

>   How about we call this R function remote_process and the python function remote_module, or something along those lines?

Ok I'll do this.

book_source/03_topical_pages/09_standalone_tools.Rmd
istfer: >Sorry not aware of the logic. How does this happen?

Sorry I was being cryptical, once we know about the `siteid`, we create this string, like `paste0(siteid %/% 1e+09, "-", siteid %% 1e+09)` You can see an example [here](https://github.com/PecanProject/pecan/blob/develop/modules/data.land/R/ic_process.R#L62-L65).

>when the same data is requested again in the same outdir then the previous file will be directly overwritten before we can do anything with it. 
Hmm this doesn't sound quite right, before we do anything we query the DB to see if there has been any input file for this site, previous file should be returned from that query with date info as they are stored in the DB. Sorry, I still haven't been able to check you code in detail yet, I don't know how the merge flag works. I'll come back to this discussion once I work through your PR

modules/data.remote/R/call_remote_process.R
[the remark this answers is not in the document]
mdietze: Overwrite should default to FALSE, and should rarely need to be invoked, but experience has taught us that theres a million little things that can go wrong (database errors, download errors, accidental file corruption, errors corrections on the data producer's side, etc) and sometimes you just need to force the code to skip all its checks and do a clean download.

modules/data.remote/R/call_remote_process.R
mdietze: Great to hear the function is smart about downloads -- what you describe is exactly the behavior we've implemented elsewhere. 👍

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -159,83 +159,178 @@ Processing functions currently available are,
...
+a. Navigate to `pecan/modules/data.remote/inst/RpTools` If you are inside the pecan directory, this can be done by,
 ```bash
-cd modules/data.remote/inst
+cd modules/data.remote/inst/RpTools
 ```
-b. Use pip to install the dependencies.
+b. Use pip to install the package. "-e" flag is used to install the package in an editable or develop mode, so that changes made to the code get updated in the package without reinstalling.
 ```bash
-pip install -r requirements.txt
+pip install -e .
```
istfer, line 170: I'm not able to install this

```
$ pip install -e .
Obtaining file:///home/istfer/pecan/modules/data.remote/inst/RpTools
Collecting attrs>=19.3.0 (from RpTools==0.1)
  Using cached https://files.pythonhosted.org/packages/a2/db/4313ab3be961f7a763066401fb77f7748373b6094076ae2bda2806988af6/attrs-19.3.0-py2.py3-none-any.whl
Collecting cachetools>=4.1.1 (from RpTools==0.1)
  Could not find a version that satisfies the requirement cachetools>=4.1.1 (from RpTools==0.1) (from versions: 0.0.0, 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 0.5.0, 0.5.1, 0.6.0, 0.7.0, 0.7.1, 0.8.0, 0.8.1, 0.8.2, 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.1.0, 1.1.1, 1.1.2, 1.1.3, 1.1.4, 1.1.5, 1.1.6, 2.0.0, 2.0.1, 2.1.0, 3.0.0, 3.1.0, 3.1.1)
No matching distribution found for cachetools>=4.1.1 (from RpTools==0.1)
You are using pip version 9.0.1, however version 20.2.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

```
Looks like our python versions are out of date? When I try `pip install --upgrade pip` I get

```
Collecting pip
  Using cached https://files.pythonhosted.org/packages/5a/4a/39400ff9b36e719bdf8f31c99fe1fa7842a42fa77432e584f707a5080063/pip-20.2.2-py2.py3-none-any.whl
Installing collected packages: pip
  Found existing installation: pip 9.0.1
    Uninstalling pip-9.0.1:
Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/req/req_set.py", line 778, in install
    requirement.uninstall(auto_confirm=True)
  File "/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/req/req_install.py", line 754, in uninstall
    paths_to_remove.remove(auto_confirm)
  File "/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/req/req_uninstall.py", line 115, in remove
    renames(path, new_path)
  File "/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/utils/__init__.py", line 267, in renames
    shutil.move(old, new)
  File "/usr/lib/python2.7/shutil.py", line 300, in move
    rmtree(src)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 250, in rmtree
    os.remove(fullname)
OSError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/dist-packages/pip-9.0.1-py2.7.egg/pip/wheel.pyc'
```

@mdietze is it possible to upgrade pecan2?

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -775,3 +776,48 @@ This information is currently used by the following R functions:
...
+
+```xml
+  <remotedata>
+  <raw_mimetype>...</raw_mimetype>
+  <raw_formatname>...</raw_formatname>
+  <out_get_data>...</out_get_data>
+  <source>...</source>
+  <collection>...</collection>
+  <scale>...</scale>
+  <projection>...<projection/>
```
istfer, line 791:
```suggestion
  <projection>...</projection>
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+  raw_merge <- NULL
+  pro_merge <- NULL
+  existing_raw_file_path <- NULL
+  existing_pro_file_path <- NULL
+  
+  # extract the variables from the settings list
+  siteid <- as.numeric(settings$run$site$id)
+  siteid_short <- paste0(siteid %/% 1e+09, "-", siteid %% 1e+09)
+  raw_mimetype <- settings$remotedata$raw_mimetype
+  raw_formatname <- settings$remotedata$raw_formatname
```
istfer, line 35: I think we should switch to using registration files like Mike suggested. Many of these arguments could be extracted from those files, we might as well replace the lookup table for function names. Could be one registration file per source, e.g. for register.GEE.xml:
```
<?xml version="1.0"?>
<GEE>
  <collection>
    <original_name>COPERNICUS/S2_SR</original_name>
    <pecan_name>s2</pecan_name>
    <scale>site</scale>
    <format>
     <id></id>
     <name></name>
     <mimetype>application/x-netcdf</mimetype>
   </format>
  </collection>
  <collection>
    <original_name>LANDSAT/LC08/C01/T1_SR</original_name>
    <pecan_name>l8</pecan_name>
    <scale>site</scale>
    <format>
     <id></id>
     <name></name>
     <mimetype>application/x-netcdf</mimetype>
   </format>
  </collection>
    <collection>
    <original_name>NASA_USDA/HSL/SMAP_soil_moisture</original_name>
    <pecan_name>smap</pecan_name>
    <scale>site</scale>
    <format>
     <id></id>
     <name></name>
     <mimetype>application/x-netcdf</mimetype>
   </format>
  </collection>
</GEE>
```
I will work through an example as I go

book_source/03_topical_pages/09_standalone_tools.Rmd
ayushprd: Could you try doing `pip3 install -e .`

istfer:
```
pip3 install -e .
The program 'pip3' is currently not installed. To run 'pip3' please ask your administrator to install the package 'python3-pip'
```

ayushprd: Oh ok I think you also have Python 2 installed.  You will need pip3 here.

istfer: Yeah that's why I need @mdietze's help, I happen to set up your branch on a machine where I don't have sudo privileges, even then I would be hesitant to do a system-wide installation

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+  names(getpecancode) <- collection_lut$original_name
+  
+  if (source == "gee") {
+    collection = unname(getpecancode[collection])
+  }
+  
+  req_start <- start
+  req_end <- end
+  
+  raw_file_name = construct_raw_filename(collection, siteid_short, scale, projection, qc)
```
istfer, line 85: If these are the arguments used in the filename, we need to make sure they are not NULL, for example the check above should include an else to use some default (with a message):
```

  if (!is.null(scale)) {
    scale <- as.double(settings$remotedata$scale)
    scale <- format(scale, nsmall = 1)
  }else{
     scale <- 10
     PEcAn.logger::info("No sclae provided, Using 10 as default")
  }
```

Ideally, all remotedata tags should be checked to see if they are filled with sensible values in the beginning, probably you have checks later for some of those later on but I think beginning of the module is a good place to do that.

We should check even if the source is valid, e.g. "GEE" but not "GoogleEarthEngine". If we use registration files we can look if `register.SOURCE.xml` file exists, if not this also means that source is not implemented yet. Overall, I think registration files will help us on many levels

Speaking of which, maybe source should also be in the filename, a filename like `s2_10.0_***_site_1-26929` felt a bit uninformative as opposed to `GEE_s2_10.0_***_site_1-26929` even though the folder name has the source

istfer, line 85:
```suggestion
  raw_file_name <- construct_raw_filename(collection, siteid_short, scale, projection, qc)
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+  
+  coords = unlist(PEcAn.DB::db.query(sprintf("select ST_AsGeoJSON(geometry) from sites where id=%f", siteid), con = dbcon), use.names=FALSE)
```
istfer, line 87:
```suggestion
  coords <- unlist(PEcAn.DB::db.query(sprintf("select ST_AsGeoJSON(geometry) from sites where id=%f", siteid), con = dbcon), use.names=FALSE)
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+  
+  # check if any data is already present in the inputs table
+  existing_data <-
+    PEcAn.DB::db.query(paste0("SELECT * FROM inputs WHERE site_id=", siteid), dbcon)
+  if (nrow(existing_data) >= 1) {
```
istfer, line 92: I wonder if this bit should be refactored to its own function, e.g. remotedata_db_check

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+  pro_mimetype <- settings$remotedata$pro_mimetype
+  pro_formatname <- settings$remotedata$pro_formatname
+  out_get_data <- settings$remotedata$out_get_data
+  out_process_data <- settings$remotedata$out_process_data
+  
+  
+  dbcon <- PEcAn.DB::db.open(settings$database$bety)
+  on.exit(PEcAn.DB::db.close(dbcon), add = TRUE)
+  
+  flag <- 0
```
istfer, line 63: can this be renamed to be a bit more informative, e.g. `remotefile_check_flag` or something like that? and could you document the meanings of each number in a single place, maybe as inline comments here (wherever it is used for the first time)

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+    if (!is.null(out_process_data)) {
+      stage_process_data <- TRUE
+      write_pro_start <- req_start
+      write_pro_end <- req_end
+      pro_merge <- FALSE
+      existing_pro_file_path <- NULL
+    }
+  }
+  
+  outdir = file.path(outdir, paste(source, "site", siteid_short, sep = "_"))
```
istfer, line 307: this shouldn't be the workflow directory (they tend to be less stable), instead use this
```suggestion
  outdir <- file.path(settings$database$dbfiles, paste(source, "site", siteid_short, sep = "_"))
```

modules/data.remote/R/remote_process.R
ayushprd: > `     PEcAn.logger::info("No sclae provided, Using 10 as default")`

Currently, all of the functions which require such arguments don't have a default value(except l8 qc which I see I forgot to remove) Do you think they should have a default value? If yes, I think that should not be set here in `remote_process` and should be set in the Python function because the values differ for every collection.

> If these are the arguments used in the filename, we need to make sure they are not NULL

Can you please explain about this? It is being checked if they are NULL and in that case "NA" is used in place of the value in the filename. I went this way so that there would be a same logic for constructing the file name regardless of source and collection.

> Ideally, all remotedata tags should be checked to see if they are filled with sensible values in the beginning

Agreed, I should have done that.

> Speaking of which, maybe source should also be in the filename, a filename like s2_10.0_***_site_1-26929 felt a bit uninformative as opposed to GEE_s2_10.0_***_site_1-26929 even though the folder name has the source

Yes, will change it back.

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,703 @@
...
+  
+  # call remote_process
+  output = RpTools$rp_control(
```
istfer, line 310: before this could you do something like
```
fcn.args <- list()
fcn.args$coords     <- coords
fcn.args$outdir     <- outdir
fcn.args$start      <- start
fcn.args$end        <- end
fcn.args$source     <- source
fcn.args$collection <- collection
fcn.args$siteid     <- siteid
fcn.args$scale      <- as.double(scale)
fcn.args$projection <- projection
fcn.args$qc         <- as.double(qc)
fcn.args$algorithm              <- algorithm
fcn.args$input_file             <- input_file
fcn.args$credfile               <- credfile
fcn.args$out_get_data           <- out_get_data
fcn.args$out_process_data       <- out_process_data
fcn.args$stage_get_data         <- stage_get_data
fcn.args$stage_process_data     <- stage_process_data
fcn.args$raw_merge              <- raw_merge
fcn.args$pro_merge              <- pro_merge
fcn.args$existing_raw_file_path <- existing_raw_file_path
fcn.args$existing_pro_file_path <- existing_pro_file_path


arg.string <- listToArgString(fcn.args)

cmdFcn <- paste0("RpTools$rp_control(", arg.string, ")")
PEcAn.logger::logger.debug(paste0("Remote module executing the following function:\n", cmdFcn))
```
It helps a lot with debugging

modules/data.remote/R/remote_process.R
istfer: > Currently, all of the functions which require such arguments don't have a default value set in the Python function(except l8 qc which I see I forgot to remove) Do you think they should have a default value? 

Not sure if they have defaults, probably yes because the person who implements/develops the function typically knows better than most of the end users. But even if they don't have a default value, I think the module should fail here when some argument of the downstream function is missing or passed an invalid value.

>If yes, I think that should not be set here in `remote_process` and should be set in the Python function because the values differ for every collection.

I think we can use the registration files for inserting the defaults as well, which solves values differ for every collection and reduce user errors

>  It is being checked if they are NULL and in that case "NA" is used in place of the value in the filename. I went this way so that there would be a same logic for constructing the file name regardless of source and collection.

It is a bit unusual to have `NA` in the file names (maybe even not advisable?). Again maybe registry files can help here, when that xml is read the filename can be constructed from certain tags. Do you have downstream functions that absolutely rely on this filename convention to be preserved regardless of source and collection?

ayushprd: I too haven't seen NA being used in filenames. I also used this so that the filenames could be more sensible, example: in GEE_s2_10.0, 10 means scale and in some other collection it might be the qc

>  Do you have downstream functions that absolutely rely on this filename convention to be preserved regardless of source and collection?

Nothing as such, this was mostly to have a uniform naming convention, otherwise, it could be a little confusing for different sources and collections. e.g. appeears has projection but gee doesn't.

 But yeah maybe after trying with registry files I would be able to think better.

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,821 @@
...
+  arg.string <- PEcAn.utils::listToArgString(fcn.args)
+  
+  cmdFcn <- paste0("RpTools$rp_control(", arg.string, ")")
+  PEcAn.logger::logger.debug(paste0("Remote module executing the following function:\n", cmdFcn))
+  
+  # call rp_control
+  output <- do.call(RpTools$rp_control, fcn.args)
+  
+  # inserting data in the DB
+  if (!is.null(out_process_data)) {
```
istfer, line 454: like I suggested refactoring the `remotedata_db_check` bit above, it would also be nice to move this section to its own function, e.g. `remotedata_db_insert_helper`

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,821 @@
...
+    # if the requested processed file already exists within the required timeline dont insert or update the DB
+    if (as.character(write_pro_start) == "dont write" &&
+        as.character(write_pro_end) == "dont write") {
```
istfer, line 457: Is there a reason why `TRUE` and `FALSE` wouldn't work here instead of "dont write"? Or can't this case be assigned to another flag number?

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,821 @@
...
+          in.path    = output$process_data_path,
+          in.prefix  = output$process_data_name,
+          siteid     = siteid,
+          startdate  = write_pro_start,
+          enddate    = write_pro_end,
+          mimetype   = pro_mimetype,
+          formatname = pro_formatname,
+          con        = dbcon
+        )
+        raw_id = raw_check$id
```
istfer, line 521:
```suggestion
        raw_id <- raw_check$id
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,821 @@
...
+        raw_id <- raw_ins$input.id
+      }
+    }
+  }
+  else{
+    # if the requested raw file already exists within the required timeline dont insert or update the DB
+    if (as.character(write_raw_start) == "dont write" &&
+        as.character(write_raw_end) == "dont write") {
+      PEcAn.logger::logger.info("Requested raw file already exists")
+      raw_id   = raw_check$id
```
istfer, line 659:
```suggestion
      raw_id   <- raw_check$id
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,821 @@
...
+      raw_path = raw_check$file_path
```
istfer, line 660:
```suggestion
      raw_path <- raw_check$file_path
```

modules/data.remote/R/remote_process.R
istfer: By the way, this was something we discussed but haven't decided: Currently you keep merging multiple years into single files, is this correct? @mdietze load_data functions can subset files downstream (like the Ameriflux csv files), but following the met example do we want files returned by the remotedata module to be yearly? I.e. are we OK if `GEE_***_site_1-26929` file has multiple years or do we want them split, e.g. `GEE_***_site_1-26929.2018.nc`, `GEE_***_site_1-26929.2019.nc` ...?

Also @mdietze some sources (e.g. AppEEARS) already returns csv format iirc, do we want remote module to reformat them into netcdf?

ayushprd: > currently you keep merging multiple years into single files, is this correct?

Yes, they are merged in a single file. This entire module is designed with the idea that there should always be **only 1 file** of the same type in the DB regardless of years.

modules/data.remote/R/remote_process.R
ayushprd: As documented in L18, these "write" variables are not being used as flags, they contain the dates which has to be written in the DB. This date might be different from what must be sent to `rp_control` hence I used separate variables for these. They would contain "don't write" when the requested data is already present and the DB dates don't have to be changed. I think I can use FALSE if that is preferable.

modules/data.remote/DESCRIPTION
```diff
@@ -36,4 +36,4 @@ Copyright: Authors
 LazyLoad: yes
 LazyData: FALSE
 Encoding: UTF-8
-RoxygenNote: 7.0.2
+RoxygenNote: 7.1.0
```
istfer, line 39: @infotroph sorry for bothering you about this again, but we are still at roxygen `7.0.2` rgiht?

infotroph: We are but it may be time to change that. If someone wants to take it on before I can, the process is basically:

* Pick a date for the update and warn everyone
*Iideally close out as many open PRs as possible before then to minimize merge conflicts
* Install the newest Roxygen on your dev machine
* git pull, make sure your tree is clean, then `make clean && make document`
* Commit all the changed Rd and DESCRIPTION files
* `grep -R 7.0.2 .` -> review, update appropriately, commit. I think at least three scripts `scripts/travis/install.sh`, `.github/workflows/styler-actions.yml`, and `docker/depends/Dockerfile` will need updating, but I may be forgetting others.

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -159,83 +159,182 @@ Processing functions currently available are,
...
 
 #### Usage guide:
-This module contains GEE and other processing functions which retrieve data from a specific dataset and source. For example, `gee2pecan_s2()` downloads bands from Sentinel 2 using GEE and `bands2lai_snap()` uses the downloaded bands to compute LAI using the SNAP algorithm. These functions can be used independently but the recommended way is to use `remote_process()` which is a main function that controls all the individual functions to create an organized way of downloading and handling remote sensing data in PEcAn. 
+This module contains GEE and other processing functions which retrieve data from a specific dataset and source. For example, `gee2pecan_s2()` downloads bands from Sentinel 2 using GEE and `bands2lai_snap()` uses the downloaded bands to compute LAI using the SNAP algorithm. These functions can be used independently but the recommended way is to use `remote_process`which is the main function that controls all the individual functions to create an organized way of downloading and storing remote sensing data in PEcAn. 
+
```
istfer, line 181: if you have time, it would be great if you could include a workflow scheme around here that explains the high-level functionality, which function calls what, which decisions are made...

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+remote_process <- function(settings) {
+  # Information about the date variables used in remote_process:
+  # req_start, req_end : start, end dates requested by the user, the user does not have to be aware about the status of the requested file in the DB
+  # start, end : effective start, end dates created after checking the DB status. These dates are sent to rp_control for downloading and processing data
+  # write_raw_start, write_raw_end : start, end dates which are used while inserting and updating the DB
+  # the "pro" version of these variables have the same meaning and are used to refer to the processed file
+  
+  # The value of remotefile_check_flag denotes the following cases:
+  
+  # When processed file is requested,
```
istfer, line 25: great, thanks for adding this, but is there a `case 0` when nothing needs to be done?

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+##' @param siteid shortform of siteid
+##' @param scale scale, NULL by default
+##' @param projection projection, NULL by default
+##' @param qc qc_parameter, NULL by default
+##' @return raw_file_name
+##' @examples
+##' \dontrun{
+##' raw_file_name <- construct_raw_filename(
+##'   collection="s2",
+##'   siteid=721,
```
istfer, line 261:
```suggestion
##'   siteid="0-721",
```

modules/data.remote/R/remote_process.R
ayushprd: Not really. In those cases, the `write_***_ start` and `write_***_end` contain `dont write` so nothing actually gets changed in the DB and the details of the existing data are returned. But maybe I could represent this using `0` case, I'll think about this.

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+##'   scale=10.0
+##'   projection=NULL
+##'   qc=1.0)
+##' }
+##' @author Ayush Prasad
+construct_raw_filename <-
```
istfer, line 267: could you modify this function such as the following lines so that we don't have `NA`s in the filenames, and also add the source to filename:

```
construct_raw_filename <-
  function(source,
           collection,
           siteid,
           scale = NULL,
           projection = NULL,
           qc = NULL) {
    # skip if a parameter is not applicable and is NULL
    if (is.null(scale)) {
      scale_str <- "_"
    } else{
      scale_str <- paste0("_", format(scale, nsmall = 1), "_")
    }
    if (is.null(projection)) {
      prj_str <- ""
    }else{
      prj_str <- paste0(projection, "_")
    }
    if (is.null(qc)) {
      qc_str <- ""
    } else{
      qc_str <- paste0(format(qc, nsmall = 1), "_")
    }
    raw_file_name <-
      paste0(toupper(source), "_", collection, scale_str, prj_str, qc_str, "site_", siteid)
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+  
+  RpTools <- reticulate::import("RpTools")
+  
+  input_file             <- NULL
+  stage_get_data         <- NULL
+  stage_process_data     <- NULL
+  raw_merge              <- NULL
+  pro_merge              <- NULL
+  existing_raw_file_path <- NULL
+  existing_pro_file_path <- NULL
```
istfer, line 45: I think some of these are set now in the refactored functions, could you clean up the ones that are not needed

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+        stage_process_data     <- TRUE
+        write_pro_start        <- req_start
+        write_pro_end          <- req_end
+        pro_merge              <- FALSE
+        existing_pro_file_path <- NULL
+      }
+    }
+    
+    return(
+      list(
```
istfer, line 696:
```suggestion
            list(
        remotefile_check_flag  = remotefile_check_flag,
        start                  = start,
        end                    = end,
        stage_get_data         = stage_get_data,
        write_raw_start        = write_raw_start,
        write_raw_end          = write_raw_end,
        raw_merge              = raw_merge,
        existing_raw_file_path = existing_raw_file_path,
        stage_process_data     = stage_process_data,
        write_pro_start        = write_pro_start,
        write_pro_end          = write_pro_end,
        pro_merge              = pro_merge,
        input_file             = input_file,
        existing_pro_file_path = existing_pro_file_path,
        raw_check              = raw_check,
        pro_check              = pro_check
      )
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+    collection = unname(getpecancode[collection])
+  }
+  
+  # construct raw file name
+  raw_file_name <-
+    construct_raw_filename(collection, siteid_short, scale, projection, qc)
+  
+  
+  # check if any data is already present in the inputs table
+  dbstatus <-
```
istfer, line 136: could you return a named list from this function so that you can access elements by their names? using indices is harder to maintain and more error-prone

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+      siteid            = siteid,
+      siteid_short      = siteid_short,
+      out_get_data      = out_get_data,
+      algorithm         = algorithm,
+      out_process_data  = out_process_data,
+      overwrite         = overwrite,
+      dbcon             = dbcon
+    )
+  
+  remotefile_check_flag  <- dbstatus[[1]]
```
istfer, line 150:
```suggestion
  remotefile_check_flag  <- dbstatus$remotefile_check_flag 
  start                  <- dbstatus$start                 
  end                    <- dbstatus$end                   
  stage_get_data         <- dbstatus$stage_get_data        
  write_raw_start        <- dbstatus$write_raw_start       
  write_raw_end          <- dbstatus$write_raw_end         
  raw_merge              <- dbstatus$raw_merge             
  existing_raw_file_path <- dbstatus$existing_raw_file_path
  stage_process_data     <- dbstatus$stage_process_data    
  write_pro_start        <- dbstatus$write_pro_start       
  write_pro_end          <- dbstatus$write_pro_end         
  pro_merge              <- dbstatus$pro_merge             
  input_file             <- dbstatus$input_file            
  existing_pro_file_path <- dbstatus$existing_pro_file_path
  raw_check              <- dbstatus$raw_check             
  pro_check              <- dbstatus$pro_check  
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+  write_pro_end          <- dbstatus[[11]]
+  pro_merge              <- dbstatus[[12]]
+  input_file             <- dbstatus[[13]]
+  existing_pro_file_path <- dbstatus[[14]]
+  raw_check              <- dbstatus[[15]]
+  pro_check              <- dbstatus[[16]]
+  
+  # construct outdir path
+  outdir <-
+    file.path(outdir, paste(source, "site", siteid_short, sep = "_"))
```
istfer, line 169: could be nice to have source in capital letters
```suggestion
    file.path(outdir, paste(toupper(source), "site", siteid_short, sep = "_"))
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1054 @@
...
+
+  # extract the AOI of the site from BETYdb
+  coords <-
```
istfer, line 172: like we talked, it would be great if you could add the example of storing polygon in the DB to the documentation

```
db.query("insert into sites (country, sitename, geometry) values ('FI', 'Qvidja_ca6cm', ST_SetSRID(ST_MakePolygon(ST_GeomFromText('LINESTRING(22.388957339620813 60.287395608412218 14.503780364990234, 22.389600591651835 60.287182336733203 14.503780364990234,
22.38705422266651  60.285516177775868 14.503780364990234,      
22.386575219445195 60.285763643883932 14.503780364990234,
22.388957339620813 60.287395608412218 14.503780364990234 )')), 4326));", con = dbcon)
```

If the site doesn't have the polygon it would return a point, e.g.
```

>   coords <-
+     unlist(PEcAn.DB::db.query(
+       sprintf("select ST_AsGeoJSON(geometry) from sites where id=%f", 1000000758),
+       con = dbcon
+     ), use.names = FALSE)
> coords
[1] "{\"type\":\"Point\",\"coordinates\":[-95.2001768,30.1625267,0]}"
```

Would we want to check if the downstream functions need a polygon type or not here?

ayushprd: > If the site doesn't have the polygon it would return a point, e.g.

Do you mean adding documentation for creating Point site? Because otherwise as you showed in your bottom example this query does return geojson info for Point site as well.

> Would we want to check if the downstream functions need a polygon type or not here?

These are being taken care of by the functions in `RpTools`

modules/data.remote/R/remote_process.R
ayushprd: sorry should have thought about this, will change.

modules/data.remote/R/remote_process.R
istfer: >Do you mean adding documentation for creating Point site?

No no for creating the polygon site, vast majority of the sites on Bety DB has point. But a lot of people could want to use your functions with the polygon

>These are being taken care of by the functions in RpTools

OK, but it could be harder for people to debug there, if the issue is as simple as forgetting to create the polygon for the site, we might as well let them know here. I fell like this will be a common issue just because we are not used to create these polygons for the sites. We can put this info into the registry files as well and put a high level warning here, e.g. assume these are the tags in the registry.xml:

```
<coord>
  <coordtype>polygon</coordtype>
</coord>
```
If the code above returns point, we throw a warning (or an error if we are sure it will fail?)

This could be the registry for sources/collections that can work with both types:
```
<coord>
  <coordtype>polygon</coordtype>
  <coordtype>point</coordtype>
</coord>
```

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1222 @@
...
+  stage_process_data     <- dbstatus$stage_process_data    
+  write_pro_start        <- dbstatus$write_pro_start       
+  write_pro_end          <- dbstatus$write_pro_end         
+  pro_merge              <- dbstatus$pro_merge             
+  input_file             <- dbstatus$input_file            
+  existing_pro_file_path <- dbstatus$existing_pro_file_path
+  raw_check              <- dbstatus$raw_check             
+  pro_check              <- dbstatus$pro_check  
+
+
```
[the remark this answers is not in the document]
ayushprd: If I understood your comment correctly, this is already being done. E.g. L904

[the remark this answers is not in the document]
istfer: No, I meant once we know that the exact files exist, we should return and shouldn't run any of the downstream lines, e.g. we shouldn't even call Rptools, `output <- do.call(RpTools$rp_control, fcn.args)` it slows down the workflow for no reason

modules/data.remote/R/remote_process.R
```diff
@@ -0,0 +1,1241 @@
...
+  
+  if(stage_get_data == FALSE && stage_process_data == FALSE){
+    # requested data already exists, no need to call rp_control
+    
+    pro_id   <- pro_check$id
+    pro_path <- pro_check$file_path
+    raw_id   <- raw_check$id
+    raw_path <- raw_check$file_path
+  
+    settings$remotedata$raw_id   <- raw_id
```
istfer, line 198:
```suggestion
    settings$remotedata$raw_id   <-  raw_check$id
```
Is there a reason not to be more concise and directly pass the ids and paths like this and get rid of the 4 extra lines above?

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -775,3 +776,46 @@ This information is currently used by the following R functions:
...
+  <projection>...</projection>
+  <qc>...</qc>
+  <algorithm>...</algorithm>
+  <credfile>...</credfile>
+  <out_process_data>...</out_process_data>
+  <overwrite>...</overwrite>
+  </remotedata>
+```
+
+* `raw_mimetype`: (required) MIME type of the raw file
```
ayushprd, line 798: :eyes:

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -159,83 +159,182 @@ Processing functions currently available are,
...
+### Configuring `remote_process`
+
+`remote_process` is configured using remote data tags in the `pecan.xml`. The required tags are described below,
+
+```xml
+  <remotedata>
+  <raw_mimetype>...</raw_mimetype>
```
ayushprd, line 188: :eyes:

book_source/03_topical_pages/03_pecan_xml.Rmd
istfer: yup the docs need to be updated to explain the registry file logic and tags

istfer (approved):
I haven't tested it this time but looks like code will now work when LAI is passed all capitals, right?

We can finally pull this in and write a vignette :+1:

ayushprd:
> I haven't tested it this time but looks like code will now work when LAI is passed all capitals, right?

Yes, it will.

book_source/03_topical_pages/09_standalone_tools.Rmd
```diff
@@ -159,83 +159,220 @@ Processing functions currently available are,
...
+* `qc`: (optional) quality control parameter, required for some gee collections
+* `overwrite`: (optional) if TRUE database checks will be skipped and existing data of same type will be replaced entirely. When processed data is requested, the raw data required   for creating it will also be replaced. By default FALSE
+
+These tags are only required if processed data is requested:
+
+* `out_process_data`: (optional) type of processed output requested, e.g, lai
+* `algorithm`: (optional) algorithm used for processing data, currently only SNAP is implemented to estimate LAI from Sentinel-2 bands
+* `credfile`: (optional) absolute path to JSON file containing Earthdata username and password, only required for AppEEARS
+
+Additional information are taken from the registration files located at pecan/modules/data.remote/inst/registration, each source has its own registration file. This is so because there isn't a standardized way to retrieve all image collections from GEE and each image collections may require its own way of performing quality checks, etc wheras all of the products available on AppEEARS can be retrieved using its API.
```
istfer, line 231:
```suggestion
Additional information are taken from the registration files located at pecan/modules/data.remote/inst/registration, each source has its own registration file. This is so because there isn't a standardized way to retrieve all image collections from GEE and each image collections may require its own way of performing quality checks, etc whereas all of the products available on AppEEARS can be retrieved using its API in a standardized way.
```
Thanks, could you also link `pecan/modules/data.remote/inst/registration`

ayushprd:
@istfer Modification: I have now modified `gee2pecan_s2` to work with Point type as well like the other gee scripts.

mdietze (approved):

Issue #2639, opened 2020-06-18
Use environment variables in the API for Authentication Requirement & Visibility of Details queried from betyDB


tezansahu:
## Description

- Currently, the API enforces authentication for some endpoints while allows some others to be called without authenticated. In addition to this, we wish to have an environment variable `AUTH_REQ`, which if `TRUE` allows the above mentioned behavior, but if `FALSE` allows requests to all endpoints on that server to be called without the need for authentication.

- Currently, on calling the workflows & runs related endpoints, all results that exist across all host machines that satisfy the filtering criteria are returned. We would like to have an environment variable `HOST_ONLY`, which if `FALSE` would callow behavior as mentioned above, but if `TRUE` would restrict results belonging to the particular host only.

## Proposed Solution

Implement the 2 environment variables `AUTH_REQ` & `HOST_ONLY` as mentioned in the description & add simple checks/filters to achieve the mentioned functionality.

PR #2674, opened 2020-08-03
PEcAn API: Workflow status, download input/output files, JSON workflow submission

tezansahu:
- Created API to obtain status of a workflow
- Created API endpoint to download input & output files for a specified run
- Created API endpoint to download files for a specified workflow
- Created API endpoint to get information about inputs based on `model_id` & `site_id`
- Created API endpoint to search for formats based on string matching for `format_name` & `mimetype`
- Created API endpoint to retrieve details of a format using its ID
- Added feature to submit workflow as a JSON
- Added functionality to allow or restrict access to the API via authentication based on the `AUTH_REQ` environment variable (See issue #2639)

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


apps/api/R/inputs.R
```diff
@@ -0,0 +1,116 @@
+library(dplyr)
+
+#' Search for Inputs containing wildcards for filtering
+#' @param model_id Model Id (character)
+#' @param site_id Site Id (character)
+#' @param offset
+#' @param limit
```
mdietze, line 7: The code below indicates that limit can only take on specific discrete values, but that's not in the function documentation

apps/api/R/inputs.R
```diff
@@ -0,0 +1,116 @@
...
+  inputs <- tbl(dbcon, "models") %>%
+    select(model_id = id, modeltype_id, model_name, revision) %>%
+    inner_join(inputs, by='modeltype_id') %>%
+    select(-modeltype_id)
+  
+  inputs <- tbl(dbcon, "sites") %>%
+    select(site_id = id, sitename) %>%
+    inner_join(inputs, by='site_id')
+  
+  if(! is.null(model_id)) {
```
mdietze, line 53: I think I'd argue that if you're not looking for a specific model_id that you don't even bother with the join to modeltypes_formats and models as there are definitely a lot of datasets in the Inputs table that are not associated with any specific model.

apps/api/R/inputs.R
```diff
@@ -0,0 +1,116 @@
...
+    select(-machine_id)
+  
+  inputs <- tbl(dbcon, "modeltypes_formats") %>%
+    select(tag, modeltype_id, format_id, input) %>%
+    inner_join(inputs, by='format_id') %>%
+    filter(input) %>%
+    select(-input)
+  
+  inputs <- tbl(dbcon, "formats") %>%
+    select(format_id = id) %>%
```
mdietze, line 40: I think it would be useful to also grab the format name and mimetype, as that seems really relevant to someone looking  to pull data.

apps/api/R/submit.workflow.R
```diff
@@ -45,6 +45,17 @@ submit.workflow.list <- function(workflowList, userDetails) {
   )
   )
   
+  if(is.null(workflowList$model$id)) {
```
mdietze, line 48: Could you add a comment explaining what you're trying to do at this step? To me it looked like you were using the model type and revision to look up the model_id if that ID was missing, but instead you're just putting that same information back into the settings, but not adding the ID.

apps/api/pecanapi-spec.yml
```diff
@@ -358,6 +360,94 @@ paths:
...
+      parameters:
+        - in: query
+          name: model_id
+          description: If provided, returns all inputs for the provided model_id
+          required: false
+          schema:
+            type: string
+        - in: query
+          name: site_id
+          description: If provided, returns all inputs for the provided site_id
```
mdietze, line 378: I could also see wanting to filter inputs by format_id (i.e. find all the sites that have data of a specific type) and machine_id

apps/api/R/inputs.R
tezansahu: It is there in the swagger documentation, the link to which is very clearly placed at the top of the Pecan Book page. These are just some examples

apps/api/R/inputs.R
tezansahu: Cool...I'll look into it

apps/api/pecanapi-spec.yml
tezansahu: Ok...I could add those as well.

apps/api/R/inputs.R
tezansahu: Moreover, this API will be used tightly in conjunction with the `rpecanapi` package...in the corresponding function of `rpecanapi`, the documentation will clearly mention these specific values

apps/api/R/inputs.R
tezansahu: But if we do that, we will lose the information about `tag`,  & `format_id` (which is needed for the join with the `formats` table).

apps/api/R/submit.workflow.R
tezansahu: Thanks for this...I had something completely different in mind, and I ended up doing something different...will rectify this.

tezansahu: image

This is something I wanted to do...coz in some xml files, they have `model_id` instead of `model_name` & `revision`, & `model_name` & `revision` are required to form the name of the rabbitmq queue...that is why I wanted to store these values if they were not provided.

apps/api/pecanapi-spec.yml
tezansahu: This also means that we should have an API endpoint to get some details about the formats as well, right? Something like `/api/formats/{format_id}`, which will return information about the specified `format_id`? 

And I think we should also allow users to search for formats based on string matching with the `format_name` & `mimetype`. This could be the `/api/formats/?format_name=xyz&mimetype=abc`

apps/api/R/inputs.R
mdietze: What do you mean? format_id is part of the Input record, so you should be joining `inputs.format_id = formats.id`. Not sure what `tag` is or why you'd need it. Indeed, if you only query formats that have matches in modeltype_formats then you're missing a LOT of formats.

apps/api/R/submit.workflow.R
mdietze: That makes more sense. But I'd still say it would be good to put a comment before code blocks like these to explain the what/why of a step (the "how" should be clear from the code). That makes it a lot easier to read and for others to debug / build-upon.

apps/api/pecanapi-spec.yml
mdietze: If it's not too much trouble, yes it would be nice to search formats. If you return info about a format, make sure to grab the formats_variables, and then match variable.id to name and units. Note that the 'units' in the formats_variables references the units in the file itself, while the units in the variables table is the units of the PEcAn standard, and the formats_variables units might be blank if the files is already in PEcAn standard units. Indeed, we might not want to report the PEcAn standard units in cases where the file isn't in PEcAn standard units because that's liable to confuse users, and just use variables.units to gapfill missing units in formats_variables.

Also be aware that mimetypes are pretty standardized, but format names are largely "made up" and descriptive. So searches on the latter are more likely to be partial rather than exact (e.g. find all formats that mention Ameriflux)

tezansahu: Yup...I have created a search functionality for the formats...I will return the `formats_variables` as well according to the way you have mentioned.

Yup...I agree that the names are "made up", & hence I have enabled substring matching based filtering, not only here, but in fact for all endpoints that search based on a name (like for models, PFTs, sites, etc.)

tezansahu:
@mdietze @robkooper I have made all the suggested modifications. Requesting you to review it once & let me know your views.

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
```diff
@@ -541,6 +553,220 @@ print(json.dumps(response.json(), indent=2))
...
+```
+
+### {-}
+
+### `GET /api/inputs/` {.tabset .tabset-pills}
+
+#### R Snippet
+
+```R
+# Get the inputs needed for a workflow with model_id = 1000000022 & site_id = 676
```
mdietze, line 687: as noted above, could we have an example that doesn't have a model_id but does have a format_id?

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
```diff
@@ -541,6 +553,220 @@ print(json.dumps(response.json(), indent=2))
...
+
+#### R Snippet
+
+```R
+# Download the input file with id = 99000000003
+res <- httr::GET(
+         "http://localhost:8000/api/inputs/99000000003",
+         httr::authenticate("carya", "illinois")
+       )
+writeBin(res$content, "test.2002.nc")
```
mdietze, line 753: When we use the Input query like this, how are we supposed to know what the filename should be? If not provided in an example here, you should definitely make sure there are helper functions in the pecan package that does both the GET and the query for the filenames and the write. @robkooper beyond this PR, but I think it would be great if there were then core PEcAn functions that would not just use the API to grab files, but then stick them into the local DB as new dbfiles associated with the correct Input record. Would really like to get to the point where PEcAn nodes default to grabbing files from each other before attempting to download and reprocess the raw Inputs from their original sources (e.g. I could grab the met I need for a run from Shawn or Ankur if they've already run that site)

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
```diff
@@ -756,6 +982,87 @@ print(json.dumps(response.json(), indent=2))
...
+         "http://localhost:8000/api/workflows/99000000031/file/ensemble.ts.99000000017.NPP.2002.2002.Rdata",
+         httr::authenticate("carya", "illinois")
+       )
+writeBin(res$content, "test.ensemble.ts.99000000017.NPP.2002.2002.Rdata")
+```
+
+#### Python Snippet
+
+```python
+# Download the '2002.nc' output file for the run with id = 99000000282
```
mdietze, line 1055: Comment doesn't match the actual query

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
```diff
@@ -43,14 +43,26 @@ The currently implemented functionalities include:
   * [`GET /api/pfts/`](#get-apipfts): Search for PFT(s) using search pattern based on PFT name, PFT type & Model type
   * [`GET /api/pfts/{pft_id}`](#get-apipftspft_id): Fetch the details of specific PFT
 
+* __Formats:__
+  * [`GET /api/formats/`](#get-apiformats): Search for format(s) using search pattern based on format name & mime type
+  * [`GET /api/formats/{format_id}`](#get-formatsformat_id): Fetch the details of specific Format
+  
+* __Inputs:__
+  * [`GET /api/inputs/`](#get-apiinputs): Search for inputs needed for a PEcAn workflow based on `model_id` & `site_id`
```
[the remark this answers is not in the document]
tezansahu: Sure...I will

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
```diff
@@ -43,14 +43,26 @@ The currently implemented functionalities include:
...
+  * [`GET /api/inputs/{input_id}`](#get-apiinputsinput_id) *: Download the desired input file
+  
 * __Workflows:__
   * [`GET /api/workflows/`](#get-apiworkflows): Retrieve a list of PEcAn workflows
   * [`POST /api/workflows/`](#post-apiworkflows): Submit a new PEcAn workflow
   * [`GET /api/workflows/{id}`](#get-apiworkflowsid): Obtain the details of a particular PEcAn workflow by supplying its ID
+  * [`GET /api/workflows/{id}/status`](#get-apiworkflowsidstatus): Obtain the status of a particular PEcAn workflow
+  * [`GET /api/workflows/{id}/file/{filename}`](#get-apiworkflowsidfilefilename): Download the desired file from a PEcAn workflow
```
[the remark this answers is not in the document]
tezansahu: Yes, `/api/workflows/{id}` returns the list of files so that the user knows which to request. As of now, it fetches files in the workflow directory only....coz the ones in the `workflow/runs/` are fetched by the `/api/runs/{run_id}/input/{filename}` & those in the `workflow/out/` are fetched by the `/api/runs/{run_id}/input/{filename}`

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
tezansahu: Ok...I will add that

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
tezansahu: Let me look into it...

apps/api/R/inputs.R
```diff
@@ -0,0 +1,172 @@
...
+  if (nrow(input) == 0) {
+    res$status <- 404
+    return()
+  }
+  else{
+    # Generate the full file path using the file_path & file_name
+    filepath <- paste0(input$file_path, "/", input$file_name)
+    
+    if(! file.exists(filepath)){
+      res$status <- 404
```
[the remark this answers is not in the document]
tezansahu: Umm...that's a bit inconvenient I guess because the content-type is specified to be `application/octet-stream` as a decorator over the function, and changing that here may not be the easiest thing to do as of now. I think we can keep this on a hold for now & meanwhile I will see if something could be done.

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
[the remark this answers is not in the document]
mdietze: Don't want to hold up this PR further, but we should also find a way to fetch files in the PFT directory, as a lot of sensitivity analysis results get squirreled away in there. Likewise, the PDA and SDA analyses often create additional subdirectories when run.

apps/api/R/inputs.R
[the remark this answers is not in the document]
mdietze: Ok. Not the end of the world. Just didn't know if there was a way to send back a message in addition to the raw status code. Definitely not worth holding up this PR for.

book_source/03_topical_pages/07_remote_access/01_pecan_api.Rmd
```diff
@@ -43,14 +43,26 @@ The currently implemented functionalities include:
   * [`GET /api/pfts/`](#get-apipfts): Search for PFT(s) using search pattern based on PFT name, PFT type & Model type
   * [`GET /api/pfts/{pft_id}`](#get-apipftspft_id): Fetch the details of specific PFT
 
+* __Formats:__
+  * [`GET /api/formats/`](#get-apiformats): Search for format(s) using search pattern based on format name & mime type
+  * [`GET /api/formats/{format_id}`](#get-formatsformat_id): Fetch the details of specific Format
+  
+* __Inputs:__
+  * [`GET /api/inputs/`](#get-apiinputs): Search for inputs needed for a PEcAn workflow based on `model_id` & `site_id`
+  * [`GET /api/inputs/{input_id}`](#get-apiinputsinput_id) *: Download the desired input file
```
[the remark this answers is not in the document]
tezansahu: Umm...Currently I implemented it in a way assuming that the input_id points to a file & not a folder. Also, the current VM that I am working on doesn't seem to have folders containing multiple met files. So would be difficult to test. Can you suggest some way?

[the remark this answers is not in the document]
mdietze: @robkooper I'm not sure what's on the machine @tezansahu is testing the API on, but could you point him to an example CF met Input record where the files are annual.

@tezansahu if the API system only allows one file to be returned at a time, then I think you'd want to have an optional way to get file by name or by order in a sequence. For the latter I'm thinking something similar to how the search functions will return the first N queries, then you query again with an offset to get the next N queries. But that options doesn't seem great to me because you wouldn't really know which file you're getting so you wouldn't know how to name them. So I guess the simplest place to start (e.g. to get this PR in) is to just give a check on the number of files in an Input and if it's >1 and the file isn't requested by name then we need to throw an error.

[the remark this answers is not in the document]
tezansahu: I have addressed this in the latest commits

PR #2684, opened 2020-08-18
Updating GEFS download

kzarada:
## Description
Updating NOAA_GEFS download to start at first time point. It was previously starting at 6 hours after the specified start time. Also updated file saving to be by site number instead of site name and to go directly into the outfolder. The site name was adding in parentheses and spaces that caused problems when I was trying to move the files. The previous directory system had nested directories with every directory having only one .nc file. 

## Motivation and Context
This change fixes the issue of the forecast starting 6 hours past the start_date for the GEFS download and fixes the file name change and nesting directories. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -234,17 +255,16 @@ download.NOAA_GEFS <- function(outfolder, lat.in, lon.in, sitename, start_date =
...
-    ensemble_folder = file.path(outfolder, identifier)
+    identifier = paste("NOAA_GEFS", site_id, i, format(start_date, "%Y-%m-%dT%H:%M"), 
+                       format(end_date, "%Y-%m-%dT%H:%M"), sep="_")
     
+
     #Each file will go in its own folder.
-    if (!dir.exists(ensemble_folder)) {
-      dir.create(ensemble_folder, recursive=TRUE, showWarnings = FALSE)
+    if (!dir.exists(outfolder)) {
+      dir.create(outfolder, recursive=TRUE, showWarnings = FALSE)
```
mdietze, line 264: I understand the argument for not putting ever ensemble in its own folder for GEFS, but you should take a careful look at how this impacts the logic of the ensemble met code. I worry this could lead to models being confused about which files to use (e.g. some models or modules might not be checking filenames carefully enough to distinguish among 21 different files for the same date-range in the same folder) and also might have been done to maintain generality with other ensemble met options that were longer (e.g. PalEON) where you definitely wouldn't want >100 met ensembles with 1160 annual files each all in the same folder. I think the 'every ensemble member in it's own folder' norm was done to minimize the risk of mistakes/confusion.

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
```diff
@@ -312,20 +333,20 @@ download.NOAA_GEFS_downscale <- function(outfolder, lat.in, lon.in, sitename, st
...
-    #Each file will go in its own folder.
-    if (!dir.exists(ensemble_folder)) {
-      dir.create(ensemble_folder, recursive=TRUE, showWarnings = FALSE)
+  
+    if (!dir.exists(outfolder)) {
+      dir.create(outfolder, recursive=TRUE, showWarnings = FALSE)
     }
     
-    flname = file.path(ensemble_folder, paste(identifier, "nc", sep = "."))
+    flname = file.path(outfolder, paste(identifier, "nc", sep = "."))
```
mdietze, line 349: high level comment -- the code in this function looks identical to the code in the previous function, which is really bad practice. We should probably be handling the downloading and downscaling in separate functions (i.e. pull out the downscaling and have it be part of the general met_process workflow), as that's always been the design idea for met_process. As an interim, you might start by having download.NOAA_GEFS_downscale call download.NOAA_GEFS and then do the downscaling.

modules/data.atmosphere/R/download.NOAA_GEFS.R
kzarada: I checked it with my forecasts before I made the pull request and the met worked properly with SIPNET. As in it properly created ensemble clim files and each of those go into their own ensemble directory. I also looked through other people's met files that they have downloaded and couldn't find other examples of the nested directories for ensembles

modules/data.atmosphere/R/download.NOAA_GEFS_downscale.R
kzarada: Yes, we discussed this back when I created the downscale function and decided to do the whole other function for the interim. I don't think it's worth it to making another interim code since this one has been live for the last year. If I am going to go through the process of making the downscale call in the NOAA_GEFS function, I might as well make a generalized downscale function since I would have to deal with loading in the 21 nc files and downscaling and then re making the nc files. I did it this way last year because it's simpler to do the downscaling to the data before they are split by ensemble and made into nc files.

mdietze: OK, let's hold the reorganization off until the EFI RCN GEFS Harmonization discussion (Aug 21) since we'll inevitably need to reorganize code after that anyway.

modules/data.atmosphere/R/download.NOAA_GEFS.R
mdietze: Well, AFAIK the only other ensemble met options right now are ERA5 and PalEON.  Just took a look at the ERA5 code, and it is definitely splitting ensemble members into different folders https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/met2cf.ERA5.R#L131. PalEON is mostly generated outside of PEcAn and then registered posthoc.

kzarada: I think Hamze might have based the ensemble part of the ERA5 code off of the GEFS download. I remember him asking me about the code a while back and it the ERA code looks identical to GEFS there. I can change it back if that's what you would rather but it did work normally with my code and was able to make the properly clim files. My thought process was  that it's just what Luke thought would be good for ensembles, not necessarily something with the pecan standards since we dont have other examples of ensemble met downloads.

mdietze: Yes let's put it back. And yes, Luke's work was the first addition of ensembles (so in some sense it defined the standard), but he also did ask me questions 20 times a day, so it wasn't a rogue decision.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -258,13 +258,12 @@ download.NOAA_GEFS <- function(outfolder, lat.in, lon.in, site_id, start_date =
...
                        format(end_date, "%Y-%m-%dT%H:%M"), sep="_")
     
-
+    ensemble_folder = file.path(outfolder, identifier)
     #Each file will go in its own folder.
-    if (!dir.exists(outfolder)) {
-      dir.create(outfolder, recursive=TRUE, showWarnings = FALSE)
-    }
+    if (!dir.exists(ensemble_folder)) {
+      dir.create(ensemble_folder, recursive=TRUE, showWarnings = FALSE)
```
mdietze, line 265: missing closing } here and in the downscale version

PR #2690, opened 2020-09-08
Remove logger from Photosyntheisis module

dlebauer:
Currently it is not possible to install pecan photosynthesis because because it imports PEcAn.logger, even though it doesn't use the logger.

This change should make it possible to do this:

```r
install_github("PecanProject/pecan/modules/photosynthesis", ref='develop')
```


dlebauer:
currently failing due to an unrelated issue https://github.com/PecanProject/pecan/pull/2690/checks?check_run_id=1083795755

mdietze (approved):

mdietze:
@dlebauer need an update to Makefile.depends
```
 M Makefile.depends
Error: These files have changed since checkout.
```

PR #2692, opened 2020-09-08
Fix circular reference

robkooper:
There is a circular reference in benchmark and data.land. This fixes the reference by making data.land a suggest for benchmark.

Also used TRAEFIK_FRONTEND_RULE in compose file and TRAEFIK_HOST in env.example, using TRAEFIK_HOST everywhere.

This also contains the changes to get PEcAn to compile with R 4.0.2

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze:
I'm OK with this fixed, but surprised that anything in data.land (data ingest) would depend on benchmarking (model output vs data comparisons). Do you know what in data.land is calling benchmarking?

mdietze:
@robkooper this appears to be failing because of a package version requirement on `TruncatedNormal (>= 2.2)` in `modules/assim.batch/DESCRIPTION`. Might want to just relax the version requirement (e.g. >=2.0). This is blocking other PR.

PR #2708, opened 2020-10-21
CMIP5 extraction

crollinson:
## Description

Various fixes to the extract.local.CMIP5 function.  Mostly focused on fixes some of the idiosyncrasies with file paths, date conventions, and other inconsistencies with CMIP5 ensembles that have been downloaded from CEDA.

## Motivation and Context

At some point along the way, the CMIP5 extraction function I was using became unsynced with PEcAn, lost, and no longer worked.  I'm probably the only one using this function (outside of PEcAn) since you have to download the grids manually, but trying to make sure PEcAn has the full met workflow I'm working on as part of cleaning things up for the met downscaling manuscript that will be coming. 

(I think I made the doc changes -- if it doesn't work, I'll try to do the docker compile/roxygen build, but it's not super clear how to do that as someone who is only a tangential/occasional contributor.)

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


mdietze (approved):

crollinson:
@mdietze -- is there a way to make sure I keep the roxygenise version consistent with y'all or is that something where if I just ignore that line from the update document commit it will be okay?  I tried doing the docker install this morning, but was having issues.

robkooper:
what I do is 

1.  start container `docker run -ti --rm -v ${PWD}/pecan pecan/base:develop`
2. change to /pecan (where your home directory is) `cd /pecan`
3. update documentation for the package `Rscript -e "devtools::document('modules/data.atmosphere')"`

From your laptop you should now have the right version of roxygen etc.

If you run on Linux you want to use for step 1 `docker run -ti --rm  --user $(uid -u) -v ${PWD}/pecan pecan/base:develop`

crollinson:
@robkooper I couldn't get the docker stuff to work this morning... I seem to be missing some dependencies or connections that prevented the docker from compiling... My systems are all out of sync and with everything else going on it's going to take days to sort out.

mdietze:
@crollinson the simplest way to keep roxygen in sync is to install a specific version rather than asking CRAN for the latest
```
devtools::install_version("roxygen2",version="7.0.2")
```

crollinson:
Thanks for the help and patience! There's going to be a couple more met pull requests coming over the next week or so as I work out kinks in met downscaling workflows.

PR #2705, opened 2020-10-19
Met Temporal Downscaling Fixes & Updates

crollinson:
## Description

Major changes: 
1) Critical: Fix precipitation redistribution bug (tdm_lm_ensemble_sims.R line 572 most critical); 
2) Improve support for single-prediction downscaling by using the model coefficients and circumventing creating/pulling from a simulated posterior distribution of coefficients (necessary for generating an ensemble that characterizies and propagates uncertiainty)

## Motivation and Context
For the precipitation redistribution issue: there was an issue where precipitation after temporal downscaling was consistent and sometimes dramatically less than at the outset.  Turns out this was due to a really stupid mistake that wasn't caught long ago.

For the single-prediction feature: the original motivator of the TDM code was to generate an met ensemble that could be used to calculate driver uncertainty and uncertainty associated with spatial and temporal downscaling processes.  Turns out a lot of applications (including my own) don't really need this.  The previous work-around was to generate a small ensemble and take the mean.  But this lead to failed sanity checks and drivers that couldn't be executed in ED.  [insert call fail_whale] The new workaround is faster and more robust since it uses the mean generated during the model fitting stage rather than the simulated distribution of model coeffecients.

## Review Time Estimate

- [X] Immediately --> the precip bug is essential if anybody is currently using this routine
- [ ] Within one week
- [ ] When possible
## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)  [I don't know if there was an issue; this is a long-overdue PR]
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2711, opened 2020-10-22
Updates to regression-based debias routines

crollinson:
## Description

Long overdue updates and fixes to the `debias.met.regression()` function.  Two relatively small, but important updates:
1. Provide better support for single-case downscaling where we don't care about propagating uncertainty (mirroring #2705)
2. Fix an apparent long-standing bug in uncertainty in temporal trends for temperature variables.  Through a logic error, we were _removing_ the temporal trend all together rather than adding in uncertainty in that trend.  The solution is to remove the mean trend and then add in the trend estimated in the anomaly/temporal trend sub-model.  Easy fix, but important one.

## Motivation and Context

This isn't linked to any standing issues since it was previously masked int he ensemble approach.  The problem with the temporal trend was more evident when doing single-member work and is essential to be pulled in.

## Review Time Estimate

- [X] Immediately (Anybody using the debias routines for future sims _needs_ this)
- [ ] Within one week
- [ ] When possible
## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)  -- No issue documented
- [] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have updated the documentation accordingly.  _I don't THINK the documentation changed.  I tried.  I promise_

### DIFF
--- a/modules/data.atmosphere/R/debias_met_regression.R
+++ b/modules/data.atmosphere/R/debias_met_regression.R
@@ -114,7 +114,7 @@ debias.met.regression <- function(train.data, source.data, n.ens, vars.debias=NU
     ens.src <- ens.train
   }
   
-  if(pair.ens==F & ncol(source.data[[2]])==1 ){
+  if(pair.ens==F & ncol(source.data[[2]])==1){
     ens.src=1
   } else if(pair.ens==F & ncol(source.data[[2]]) > n.ens) {
     ens.src <- sample(1:ncol(source.data[[2]]), ncol(source.data[[2]]),replace=T)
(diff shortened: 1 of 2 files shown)


modules/data.atmosphere/R/debias_met_regression.R
```diff
@@ -560,7 +569,11 @@ debias.met.regression <- function(train.data, source.data, n.ens, vars.debias=NU
         
         
         if(v == "precipitation_flux"){
-          Rbeta.ann <- matrix(MASS::mvrnorm(n=n.new, coef(mod.ann), vcov(mod.ann)), ncol=length(coef(mod.ann)))
+          if(n.ens==1){
```
ankurdesai, line 572: This is nit picky but we might want to catch the case of someone inputting n.ens=0 (as in "i don't want ensembles" instead of I want one ensemble member), which will have different responses in the ifelse statement n==1 and n>1 .

ankurdesai (approved):

PR #2719, opened 2020-10-26
API workflow submission bugfixes

ashiklom:
I came across these issues while trying to use the API to submit workflows. These are a prerequisite to getting the test suite working.

- Rather than inserting workflows and then querying the last one, insert workflows with a `INSERT ... RETURNING id`, which reduces the number of queries and is more robust.
- Replace `paste`-style queries with prepared statements, which are more robust.
- Standardize the target `bety_params`. Rather than sometimes relying on `config.php` and sometimes relying on `get_postgres_envvvars`, always rely on the later, with sensible defaults (set in the file, in the global variable `.bety_params`)
- Replace `plumber::plumber` with `plumber::Plumber` (to reflect v1.0 update -- this silences a warning and protects us from upcoming deprecation)

### DIFF
--- a/apps/api/R/submit.workflow.R
+++ b/apps/api/R/submit.workflow.R
@@ -1,5 +1,13 @@
 library(dplyr)
 
+.bety_params <- PEcAn.DB::get_postgres_envvars(
+  host = "localhost",
+  dbname = "bety",
+  user = "bety",
+  password = "bety",
+  driver = "Postgres"
+)
+
 #* Submit a workflow sent as XML
 #* @param workflowXmlString String containing the XML workflow from request body
 #* @param userDetails List containing userid & username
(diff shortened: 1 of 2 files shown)


apps/api/R/submit.workflow.R
```diff
@@ -136,33 +139,45 @@ insert.workflow <- function(workflowList){
...
-  
-  update_qry <- paste0("UPDATE workflows SET folder = 'data/workflows/PEcAn_", workflow_id, "' WHERE id = '", workflow_id, "';")
-  PEcAn.DB::db.query(update_qry, dbcon)
-  
+
+  insert_query <- glue::glue(
+    "INSERT INTO workflows ",
+    "({paste(colnames(workflow_df), collapse = ', ')}) ",
+    "VALUES ({paste0('$', seq_len(ncol(workflow_df)), collapse = ', ')}) ",
+    "RETURNING id"
```
ashiklom, line 161: You can't use prepared statements to select columns, so we have to use string processing to create the insert query (but then, we use prepared statements for the actual insertions).

robkooper (approved):

apps/api/R/submit.workflow.R
robkooper: I think critical part is that this is not a user specified list, so this should be ok.

PR #2720, opened 2020-10-27
[CI] install new dependencies before starting tests

infotroph:
## Description

Turns out the current Makefile setup doesn't install missing dependencies before trying to test or check a package, leading to CI failures in the case where a PR adds new dependencies that aren't already present on the `pecan/depends` Docker container.

Fixed by having each CI job update and rerun the script that's used to install dependencies in the first place -- it skips deps that are already installed, so this only takes a few seconds in the usual case where nothing needs to be installed.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


.github/workflows/ci.yml
```diff
@@ -104,7 +104,9 @@ jobs:
 
     # install additional tools needed
     - name: install utils
-      run: apt-get update && apt-get install -y postgresql-client qpdf curl
+      run: apt-get update && apt-get install -y postgresql-client qpdf
```
mdietze, line 107: @infotroph did you mean to drop `curl` from the list of installs here and in the other two cases below?

mdietze (approved):

.github/workflows/ci.yml
infotroph: @mdietze Yes, because since #2716 curl is pre-installed on the container.

@robkooper would it make sense to add qpdf and postgresql-client to the image and then remove these apt-get calls entirely?

robkooper: only reason we have them is that the code needs them. I don't see any reason for our base image to need psql client, you can do that from R. Same for the qpdf, is just needed to make a test pass.

PR #2722, opened 2020-10-27
minor MA fix

istfer:
## Motivation and Context

Previously working meta analysis **with random effects turned on** was breaking for me. I suspect it might be related to R version bump (due to stringsAsFactors = FALSE by default)

This minor fix solved it for me. But others might want to reproduce with

```
 <meta.analysis>
  <iter>10000</iter>
  <random.effects>
   <on>TRUE</on>
   <use_ghs>FALSE</use_ghs>
  </random.effects>
  <threshold>1.2</threshold>
  <update>TRUE</update>
 </meta.analysis>
```

Perhaps `ghg` is also need to be made sure to be a factor. @dlebauer what do you think?

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/meta.analysis/R/single.MA.R
+++ b/modules/meta.analysis/R/single.MA.R
@@ -43,7 +43,8 @@ single.MA <- function(data, j.chains, j.iter, tauA, tauB, prior, jag.model.file,
                     site = "beta.site[site[k]]",
                     trt = "beta.trt[trt[k]]")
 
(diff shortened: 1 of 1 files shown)


mdietze (approved):

mdietze:
@istfer I strongly suspect we'll want the as.factor on ghs as well

PR #2723, opened 2020-10-27
PEcAn API bugfixes and enhancements

ashiklom:
Resolving some issues I discovered while trying to get the test suite up and running.

- Use the [`pool` package](https://github.com/rstudio/pool) to manage database connections in the API. This prevents "too many connections" errors, which can arise from heavy API use (e.g., when submitting dozens of test runs). Create this connection pool once, when the API is launched, and use it everywhere (rather than re-opening new connections everywhere).
- Add the `availableModels` API entrypoint for querying the models available on a given machine.
- Use explicit string formatting of `workflow_id`, to avoid implicit coercion issues with `bit64`
- Minor bugfix in `base/utils`: If `dbfile.input.check` returned a blank `data.frame` (as it does if files aren't found), then the subsequent `filter` statement would error because the `id` column was absent.

### DIFF
--- a/apps/api/R/auth.R
+++ b/apps/api/R/auth.R
@@ -25,9 +25,10 @@ get_crypt_pass <- function(username, password, secretkey = NULL) {
 #* Check if the encrypted password for the user is valid
 #* @param username Username
 #* @param crypt_pass Encrypted password
+#* @param dbcon Database connection object. Default is global database pool.
 #* @return TRUE if encrypted password is correct, else FALSE
 #* @author Tezan Sahu
-validate_crypt_pass <- function(username, crypt_pass) {
+validate_crypt_pass <- function(username, crypt_pass, dbcon = global_db_pool) {
 
   res <- tbl(dbcon, "users") %>%
     filter(login == username,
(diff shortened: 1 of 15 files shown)


apps/api/R/submit.workflow.R
```diff
@@ -43,30 +35,31 @@ submit.workflow.json <- function(workflowJsonString, userDetails){
...
+  # Set database details
+  workflowList$database <- list(
+    bety = PEcAn.DB::get_postgres_envvars(
+      host = "localhost",
+      dbname = "bety",
+      user = "bety",
+      password = "bety",
+      driver = "PostgreSQL"
+    )
+  )
```
ashiklom, line 48: Note: This effectively reverts my earlier change. It makes sense to explicitly separate how the API accesses the database (via a global connection pool, using the `RPostgres` interface) vs. how PEcAn internally accesses the database (however the user wants based on environment variables, with these defaults).

apps/api/R/auth.R
```diff
@@ -29,15 +29,11 @@ get_crypt_pass <- function(username, password, secretkey = NULL) {
 #* @author Tezan Sahu
 validate_crypt_pass <- function(username, crypt_pass) {
 
-  dbcon <- PEcAn.DB::betyConnect()
```
mdietze, line 32: Not sure why the dbcon was dropped from this function, as it's needed to query the users table. If it's already open, then we should probably pass it in as an argument rather than depend on it as a global variable.

ashiklom: Addressed in 526db40fa.

mdietze (approved):

Issue #351, opened 2015-01-29
Read THREDDS/OpenDAP based met data


ankurdesai:
A significant fraction of met observations can be downloaded by direct online access to subsets using the OpenDAP/THREDDS/DODS system, including things like the hourly renalaysis meteorology of MERRA: 
  MERRA: http://goldsmr2.sci.gsfc.nasa.gov/dods/
  Daymet: http://daymet.ornl.gov/thredds/catalog/ornldaac/1219/catalog.html
  And CMIP5 climate model data (possibly)

Apparently reading these directly in R is not well figured out according to OpenDAP. Clients exist in various other tools: http://www.opendap.org/whatClients

Here is how to read DODs files with the NCO tools: http://nco.sourceforge.net/nco.html#OPeNDAP

This might work: https://publicwiki.deltares.nl/display/OET/OPeNDAP+access+with+R
 Or this: http://lukemiller.org/index.php/2011/02/accessing-noaa-tide-data-with-r/

Anyway, something that will need to be implemented once at some point I think and would be good to be usable across a number of these growing list of servers that serve data using this method.

ankurdesai:
@jsimkins2

ashiklom:
This came up on Twitter, and I did a quick brain dump of this on the RNetCDF GH issues. Even if they don't end up implementing it, my current workarounds using `curl/crul` may be useful:

https://github.com/mjwoods/RNetCDF/issues/20

ashiklom:
That said, for GES DISC using the GrADS server, this just works:

```r
url <- "https://goldsmr3.gesdisc.eosdis.nasa.gov/dods/MAI6NVANA"
nc <- ncdf4::nc_open(url)
print(nc)
# File https://goldsmr3.gesdisc.eosdis.nasa.gov/dods/MAI6NVANA (NC_FORMAT_CLASSIC):
# 
#      7 variables (excluding dimension variables):
#         float delp[lon,lat,lev,time]   
#             _FillValue: 999999986991104
#             missing_value: 999999986991104
#             long_name: layer pressure thicknessa 
# ...
# GrADS
#         dataType: Grid
#         history: Tue Oct 10 15:00:42 GMT 2017 : imported by GrADS Data Server 2.0
```

PR #2725, opened 2020-10-29
MERRA meteorology

ashiklom:
## Description

- Adds `download.MERRA` function and other code, such that `<input><met><source>MERRA<.../>` now works.
- Minor fix to `remote.execute.R` to ensure that the target `scratchdir` exists before running. Otherwise, this throws an error.

## Motivation and Context

Not a generic solution, but partially addresses #351.

MERRA is a nice "default" data product because it has a (1) reliable data portal (GESDISC), (2) is global, (3) has a lot of variables (because it is a reanalysis), and (4) has hourly resolution. MERRA is also based on the Goddard Earth System Model (GEOS-5), which is (according to my Goddard folks) among the better models for seasonal-scale forecasting, so if we wanted to leverage that capability at some point, this is a step in that direction.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/download.MERRA.R
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  # Now, post-process
+  start_year <- lubridate::year(start_date)
+  end_year <- lubridate::year(end_date)
+  ylist <- seq(start_year, end_year)
+
+  nyear <- length(ylist)
+  results <- data.frame(
+    file = character(nyear),
```
mdietze, line 34: seems like this gets overwritten immediately, so not sure what this is storing or how `character(nyear)` is getting used

modules/data.atmosphere/R/download.MERRA.R
```diff
@@ -0,0 +1,245 @@
...
+    host = "",
+    mimimetype = "",
```
mdietze, line 36: Typo on mimetype

modules/data.atmosphere/R/download.MERRA.R
ashiklom: `character(nyear)` creates an empty character vector of length `nyear`. The other columns are silently recycled. The idea is to pre-allocate the `data.frame` to the right dimensions, and then replace the rows as you go (i.e., so that later on, things like `results$file[i] <- ...` will work as expected).

modules/data.atmosphere/R/download.MERRA.R
ashiklom: Good catch. Fixed in 9bdfd741c. 

The fact that this didn't throw an error is interesting -- I guess the later call to `results$mimetype[i] <- ...` (spelled correctly) means that R was quietly adding the column.

mdietze (approved):

PR #2728, opened 2020-11-02
API: Only pass model ID to PEcAn

ashiklom:
Currently, the API incorrectly sets the `model$type` as the `models.model_name`, though these actually have different meanings in PEcAn (though, for many models, they are coincidentally the same value). This would result in a bug when trying to run `ED2.2`, whose name is `ED2.2` but whose _type_ is `ED2`. The resulting error would be something like `No PFT "temperate.Late_Hardwood" for model type "ED2.2"`.

This PR makes it so that only the model ID is passed to PEcAn, which is already unambiguous and, moreover, is how the `rpecanapi::submit.workflow` function already works anyway (i.e., no user-facing changes). It should be the user's responsibility (for now) to manually figure out exactly which model + revision they want to run and the associated ID.

However, given the model ID, we _do_ need to correctly identify the model type and revision to determine the model's RabbitMQ queue. So now, we determine that from the model ID in the API and set it (but never pass this information on to PEcAn). Ideally, PEcAn should probably do this internally (i.e., it shouldn't be the API's responsibility to write this XML field correctly?), but I'm going to punt on that fix because it could be a bit more involved.

### DIFF
--- a/apps/api/R/submit.workflow.R
+++ b/apps/api/R/submit.workflow.R
@@ -35,7 +35,7 @@ submit.workflow.json <- function(workflowJsonString, userDetails){
 #* @param dbcon Database connection object. Default is global database pool.
 #* @return ID & status of the submitted workflow
 #* @author Tezan Sahu
-submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_pool, res) {
+submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_pool) {
 
   # Set database details
   workflowList$database <- list(
@@ -49,8 +49,8 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
   )
 
   if (is.null(workflowList$model$id)) {
-    res$status <- 400
-    return(list(error = "Must provide model ID."))
+    return(list(status = "Error",
+                error = "Must provide model ID."))
   }
 
   # Get model revision and type for the RabbitMQ queue
@@ -61,17 +61,15 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
     dplyr::collect()
 
   if (nrow(model_info) < 1) {
-    res$status <- 404  # Not found
     msg <- paste0("No models found with ID ", format(workflowList$model$id, scientific = FALSE))
-    return(list(error = msg))
+    return(list(status = "Error", error = msg))
   } else if (nrow(model_info) > 1) {
-    res$status <- 400
     msg <- paste0(
       "Found multiple (", nrow(model_info), ") matching models for id ",
       format(workflowList$model$id, scientific = FALSE),
       ". This shouldn't happen! Check your database for errors."
     )
-    return(list(error = msg))
+    return(list(status = "Error", error = msg))
   }
 
   model_type <- model_info$name
--- a/apps/api/R/submit.workflow.R
+++ b/apps/api/R/submit.workflow.R
@@ -48,15 +48,8 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
     )
   )
 
-  if (!is.null(workflowList$model$id) &&
-        (is.null(workflowList$model$type) || is.null(workflowList$model$revision))) {
-    res <- dplyr::tbl(dbcon, "models") %>%
-      select(id, model_name, revision) %>%
-      filter(id == !!workflowList$model$id) %>%
-      collect()
-
-    workflowList$model$type <- res$model_name
-    workflowList$model$revision <- res$revision
+  if (is.null(workflowList$model$id)) {
+    stop("Must provide model ID.")
   }
 
   # Fix RabbitMQ details
--- a/apps/api/R/submit.workflow.R
+++ b/apps/api/R/submit.workflow.R
@@ -52,12 +52,31 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
     stop("Must provide model ID.")
   }
 
+  # Get model revision and type for the RabbitMQ queue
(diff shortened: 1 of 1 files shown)


apps/api/R/submit.workflow.R
```diff
@@ -48,23 +48,35 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
...
   )
 
-  if (!is.null(workflowList$model$id) &&
-        (is.null(workflowList$model$type) || is.null(workflowList$model$revision))) {
-    res <- dplyr::tbl(dbcon, "models") %>%
-      select(id, model_name, revision) %>%
-      filter(id == !!workflowList$model$id) %>%
-      collect()
-
-    workflowList$model$type <- res$model_name
```
ashiklom, line 58: Wrong! `models.model_name` is _not_ the same as `model$type` -- that should be `modeltypes.name`.

apps/api/R/submit.workflow.R
```diff
@@ -48,23 +48,35 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
...
+    dplyr::collect()
+
+  if (nrow(model_info) < 1) {
+    stop("No models found with ID ", format(workflowList$model$id, scientific = FALSE))
+  } else if (nrow(model_info) > 1) {
+    stop("Found multiple (", nrow(model_info), ") matching models for id ",
+         format(workflowList$model$id, scientific = FALSE),
+         ". This shouldn't happen! ",
+         "Check your database for errors.")
+  }
```
ashiklom, line 69: The second error might be overkill, but I feel like it can't hurt to protect against weird errors.

dlebauer:
I propose that we not require someone know the id - perhaps make it optional. Most of the tables have sets of rows that are uniquely identifying, and the purpose of this is so that users can interact with meaningful names and values rather than arbitrary integers.

dlebauer (changes requested):
I propose that we not require someone know the id - perhaps make it optional. Most of the tables have sets of rows that are uniquely identifying, and the purpose of this is so that users can interact with meaningful names and values rather than arbitrary integers.

I think this is a higher level discussion regarding UI/UX - but something that @gsrohde taught me about and I think makes sense.

apps/api/R/submit.workflow.R
```diff
@@ -48,23 +48,35 @@ submit.workflow.list <- function(workflowList, userDetails, dbcon = global_db_po
...
-        (is.null(workflowList$model$type) || is.null(workflowList$model$revision))) {
-    res <- dplyr::tbl(dbcon, "models") %>%
-      select(id, model_name, revision) %>%
-      filter(id == !!workflowList$model$id) %>%
-      collect()
-
-    workflowList$model$type <- res$model_name
-    workflowList$model$revision <- res$revision
+  if (is.null(workflowList$model$id)) {
+    stop("Must provide model ID.")
```
robkooper, line 52: one option to make it nicer is to use:
```
function(res){
  msg <- "Your request did not include a required parameter."
  res$status <- 400 # Bad request
  return(list(error=jsonlite::unbox(msg)))
```
this will return a 400 message, I think stop will result in a 500

ashiklom:
@dlebauer This PR is a bugfix and does not change the user interface in any way -- the current `rpecanapi` is already structured around the model ID, and was just using it to set incorrect fields.

More generally, I agree that it's better UI/UX to use human-readable identifiers. But that only works effectively when they are unique and clear, which is not currently the case for the names or revisions in our models table (at least AFAIR -- I had a lot of trouble trying to get a working ED version because of all the different redundant names + revisions).

I also think such changes will be easier to implement once we have a fully functional test suite. This bug is currently preventing any dockerized ED runs from running, which makes it difficult to resolve other issues.

Given all that, I propose we punt on this till a later date by making it an issue.

dlebauer (approved):
@ashiklom sorry I read too quickly / misread. I'm not sure if all of these have been implemented, but models should be uniquely identifiable by [model_name, revision] per https://github.com/PecanProject/bety/issues/196

robkooper (changes requested):
let me know if you want to make the changes to return a 400 instead of a 500 error.

apps/api/R/submit.workflow.R
ashiklom: Great idea, thanks!

ashiklom:
@robkooper Great suggestion! Feel free to push up a patch, or I can do it tomorrow morning.

@dlebauer If we don't have that constraint, we should definitely add it! I could have sworn I recalled seeing multiple ED2 gits in the models table, but I may be misremembering. Regardless, I do agree it could be useful to have more user-friendly wrappers to the model submission API functions that use the model name and revision instead of the ID.

apps/api/R/submit.workflow.R
ashiklom: Addressed in https://github.com/PecanProject/pecan/pull/2728/commits/453e9a3abf7d03ba14239b4eba71fea178ff2b1b.

robkooper (approved):

apps/api/R/submit.workflow.R
ashiklom: Actually, scratch that. I'm going to respectfully punt on this because it's a bit more work than I expected (because `POST workflow` calls `submitWorkflow`, which in turn calls several other functions, and the `res` has to be propagated through all of them.

ashiklom: Actually, double scratch that. See 35ef735f1.

ashiklom: Also, for the above fix to work effectively, you'll need https://github.com/PecanProject/rpecanapi/pull/8.

PR #2733, opened 2020-11-06
MERRA bugfixes

ashiklom:
## Description

- Change which MERRA radiation variables are being retrieved. Before, I was retrieving net longwave and shortwave fluxes. However, I now realize that models typically expect _incident_ fluxes. For shortwave, there was an "incident surface flux" variable, so that's easy. For longwave, the closest thing seems to be "total longwave radiation absorbed by surface". MERRA has a bajillion variables that it returns, which are all described in the [file specification](https://gmao.gsfc.nasa.gov/pubs/docs/Bosilovich785.pdf). I've also gone through and annotated all of the variables to hopefully make this easier to understand and modify later. Let me know if there's a different variable I should be using instead. (Another possibility for longwave is to do the surface absorbed longwave divided by the surface emissivity (which MERRA also spits out); this effectively assumes that the surface is in thermal equilibrium, so that absorptivity == emissivity; in other words, we assume that `absorbed_longwave = incident_longwave * surface_absorptivity` and `surface_absorptivity == surface_emissivity` and solve for `incident_longwave`).
- Previously, MERRA would not overwrite the PEcAn standard files by default. This prevented MERRA time series from being extended, as happens all the time in the PEcAn workflow. To granularize the control here, I revised the code so that there are two separate arguments -- (1) `overwrite` controls what happens to the PEcAn standard format file, with default `TRUE`; and (2) `redownload` controls whether the individual daily MERRA files are redownloaded, with default `FALSE`.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/download.MERRA.R
```diff
@@ -1,14 +1,17 @@
...
 #' @param ... Not used -- silently soak up extra arguments from `convert.input`, etc.
 #' @return `data.frame` of meteorology data metadata
 #' @author Alexey Shiklomanov
 #' @export
 download.MERRA <- function(outfolder, start_date, end_date,
                            lat.in, lon.in,
-                           overwrite = FALSE,
+                           overwrite = TRUE,
                            verbose = FALSE,
+                           redownload = FALSE,
```
mdietze, line 14: I'm not sure why you're creating a new flag that appears to do the same thing as the overwrite flag, or why you've set overwrite to TRUE by default.

mdietze (changes requested):

modules/data.atmosphere/R/download.MERRA.R
ashiklom: This is in my PR description:

> To granularize the control here, I revised the code so that there are two separate arguments -- (1) overwrite controls what happens to the PEcAn standard format file, with default TRUE; and (2) redownload controls whether the individual daily MERRA files are redownloaded, with default FALSE.

Basically, the way the code works is it uses the NetCDF subset protocol to download each day's raw MERRA NetCDF file to the directory first, and then runs a post-processing step that combines them into PEcAn standard annual files (and applies unit conversions, renaming, etc. as appropriate). `redownload` controls whether the daily MERRA files will be overwritten if present -- unless we change the variables that are loaded, we typically won't have to do this. On the other hand, `overwrite` controls whether the resulting PEcAn standard annual MERRA file (e.g., `MERRA.2012.nc`) will be overwritten. It's set to `TRUE` by default to allow PEcAn to extend the met for a given year if necessary -- e.g., if I run for June-July 2012, I get a `MERRA.2012.nc` file, but I want to overwrite that file later if I want to run the same site for May-September 2012.

mdietze: Two reactions
1. I think the need for two flags is arising because this download function is trying to do what the met workflow actually formalizes as two different steps: download.[source] and met2cf.[source]
2. Given that we are supporting >20 products already (and growing) it makes me nervous when one function changes the meaning of a tag to be slightly different from what its surrounding workflow is assuming. Fine when the function is run by itself, but error prone when the larger met workflow is going to be passing "overwrite = FALSE" by default, is expect a function to redownload if it passes overwrite=TRUE, and has no way to set "redownload" at all since it's not a tag the workflow is handling.

ashiklom: Fair enough. Should I just change it so that `redownload => overwrite` and the `MERRA.2012.nc` file is always overwritten when this function is called? That would be much easier than factoring out a separate `met2cf` function out of all this, though I can do that if desired.

mdietze (approved):

PR #2734, opened 2020-11-06
met2model.ED2 overhaul

ashiklom:
## Description and motivation

The motivation for this was that the previous version didn't allow ED2 to generate met inputs for incomplete years.  I don't really know why we previously didn't allow this. I would have liked to implement a less disruptive fix, but I couldn't understand what the old code was doing at all, so I reorganized it in a way that can accommodate partial years and, IMHO, is a bit clearer. I did my best to add explanatory comments.

I'll also add some line-specific annotations to this PR to help guide your review.

Tagging @mccabete @istfer and @mdietze because I recall you all encountering issues relate to how ED handles times, and/or because you previously made significant contributions to this code.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:


models/ed/R/met2model.ED2.R
```diff
@@ -49,13 +49,13 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
-  dl <- c(0, 32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336, 367)
   month <- c("JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC")
-  mon_num <- c("01", "02", "03", "04", "05", "06", "07", "08", "09", "10", "11", "12")
-
 
   day2mo <- function(year, day, leap_year) {
+    # DOY corresponding to start of each month without a leap year...
+    dm <- c(0, 32, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366)
+    # ...and with a leap year
+    dl <- c(0, 32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336, 367)
```
ashiklom, line 58: These vectors aren't used anywhere else in this function. I moved them into this function to make it clearer what they were for.

models/ed/R/met2model.ED2.R
```diff
@@ -49,13 +49,13 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
   ## check to see if the outfolder is defined, if not create directory for output
   dir.create(met_folder, recursive = TRUE, showWarnings = FALSE)
 
-  dm <- c(0, 32, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366)
-  dl <- c(0, 32, 61, 92, 122, 153, 183, 214, 245, 275, 306, 336, 367)
   month <- c("JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC")
-  mon_num <- c("01", "02", "03", "04", "05", "06", "07", "08", "09", "10", "11", "12")
```
ashiklom, line 55: This variable wasn't used anywhere in the code.

models/ed/R/met2model.ED2.R
```diff
@@ -149,9 +149,7 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
     ## determine GMT adjustment lst <- site$LST_shift[which(site$acro == froot)]
 
     ## extract variables
-    lat  <- eval(parse(text = lat))
-    lon  <- eval(parse(text = lon))
```
ashiklom, line 153: I have no idea what was going on here -- maybe the idea was to force these to be numbers? Regardless, `eval(parse(...))` is really janky, and the code works just fine without these lines, so I removed them.

models/ed/R/met2model.ED2.R
```diff
@@ -149,9 +149,7 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
-    sec  <- nc$dim$time$vals
```
ashiklom, line 154: I renamed this to `tdays` (line 152 in the new version) to make the units clearer.

models/ed/R/met2model.ED2.R
```diff
@@ -176,16 +174,32 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
     useCO2 <- is.numeric(CO2)
 
     ## convert time to seconds
-    sec <- udunits2::ud.convert(sec, unlist(strsplit(nc$dim$time$units, " "))[1], "seconds")
+    sec <- udunits2::ud.convert(tdays, unlist(strsplit(nc$dim$time$units, " "))[1], "seconds")
 
     ncdf4::nc_close(nc)
 
-    dt <- PEcAn.utils::seconds_in_year(year, leap_year) / length(sec)
```
ashiklom, line 183: This way of determining the time step limits us to complete years. I replaced it with taking the lag-1 difference between time steps (in seconds; line 189 in the new version).

models/ed/R/met2model.ED2.R
```diff
@@ -176,16 +174,32 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
+    # length-1 array to a vector.
+    dt <- drop(unique(diff(sec)))
+    if (length(dt) > 1) {
+      PEcAn.logger::logger.severe(paste0(
+        "Time step (`dt`) is not uniform! Identified ",
+        length(dt), " unique time steps. ",
+        "`head(dt)` (in seconds): ",
+        paste(head(dt), collapse = ", ")
+      ))
+    }
```
ashiklom, line 197: If the difference between time points in the met file is not strictly regular, this will throw an informative error that also prints out (at most six of) the different time steps.

models/ed/R/met2model.ED2.R
```diff
@@ -176,16 +174,32 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
 
     toff <- -as.numeric(lst) * 3600 / dt
 
     ## buffer to get to GMT
-    slen <- seq_along(SW)
+    slen <- seq_along(sec)
```
ashiklom, line 202: All of these vectors pretty much have to be the length by definition (unless the met file is mis-specified), so the exact choice of variable here shouldn't matter. It seemed a bit clearer to me to use the time variable rather than arbitrarily choosing one of the actual met variables (which, in theory, could be multidimensional, though that's not actually the case).

models/ed/R/met2model.ED2.R
```diff
@@ -198,58 +212,37 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
-        break
-      }
-      yr[rng]  <- rep(y + 1, length(rng))
-      doy[rng] <- rep(1:366, each = day_secs / dt)[1:length(rng)]
-      hr[rng]  <- rep(seq(0, length = day_secs / dt, by = dt / day_secs * 24), 366)[1:length(rng)]
-    }
-    if (skip) {
-      print("Skipping to next year")
-      next
-    }
```
ashiklom, line 245: This is the code I did not understand at all. As far as I could tell, its main job was to determine the day of year (`doy`) and local time (`hr`) for the purposes of calculating the solar angle. I opted for a more straightforward approach to both of these variables in the new version, which effectively boils down to:

```r
doy <- floor(tdays) + 1
# %% means "the remainder of"
hr <- (tdays %% 1) * 24
```

models/ed/R/met2model.ED2.R
```diff
@@ -278,59 +271,57 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
       co2A <- CO2 * 1e+06  # surface co2 concentration [ppm] converted from mole fraction [kg/kg]
     }
 
-    ## create directory if(system(paste('ls',froot),ignore.stderr=TRUE)>0)
-    ## system(paste('mkdir',froot))
-
-    ## write by year and month
-    for (y in year + 1:nyr - 1) {
```
ashiklom, line 285: This is a _nested_ year loop --- this entire section is already looping over years (based on the PEcAn standard met files). Again, I don't know if this was because of something weird related to time zones. All of the changes below are just removing the outer year loop and re-indenting -- I didn't mess with any of the underlying logic here.

models/ed/R/met2model.ED2.R
```diff
@@ -176,16 +179,32 @@ met2model.ED2 <- function(in.path, in.prefix, outfolder, start_date, end_date, l
...
-    dt <- PEcAn.utils::seconds_in_year(year, leap_year) / length(sec)
+    # `dt` is the met product time step in seconds. We calculate it here as
+    # `sec[i] - sec[i-1]` for all `[i]`. For a properly formatted product, the
+    # timesteps should be regular, so there is a single, constant difference. If
+    # that's not the case, we throw an informative error.
+    #
+    # `drop` here simplifies length-1 arrays to vectors. Without it, R will
+    # later throw an error about "non-conformable arrays" when trying to add a
+    # length-1 array to a vector.
+    dt <- drop(unique(diff(sec)))
```
mdietze, line 194: for reasons I don't 100% understand (roundoff?) dt doesn't always have a single unique value in all met products. You might be better off with a round(mean(diff or round(median(diff instead

mdietze (approved):

mdietze:
checking R code for possible problems ... NOTE: met2model.ED2: no visible global function definition for 'head'

models/ed/R/met2model.ED2.R
ashiklom: Good to know, thanks! See https://github.com/PecanProject/pecan/pull/2734/commits/ef5244adbc3fdd54eca691bfd77264d2c2b06866. I changed it so we try to grab the unique value first, and if that fails, we throw a warning and use the rounded mean instead.

ashiklom:
> NOTE: met2model.ED2: no visible global function definition for 'head'

Ugh. Hopefully addressed in https://github.com/PecanProject/pecan/pull/2734/commits/e0bc9b3b3291405f914d7a8c783d29beafceb381. Apparently, as of a few years ago, `R CMD check` requires _all_ non-base R functions to be properly namespaced, including ones from packages that R loads by default in interactive sessions (e.g., `utils`, `stats`, `graphics`). We'll have to keep our eyes peeled for this elsewhere.

PR #2737, opened 2020-11-09
Redo on Copy of PALEON SDA

mdietze:
messed up merging @para2x version, so redoing


mdietze:
I'm going to go ahead and pull this since it's just a copy of #2451, which was already approved, with a tiny bit of extra effort on resolving namespaces, roxygen, etc.
@kzarada you'll want to take a look at these change since they include the ability to rescale state variables within the SDA

PR #2738, opened 2020-11-09
Copy of ED2 SAS #2482

mdietze:
Copy of @para2x #2482 which itself is a copy of @crollinson #2009. Hoping to resolve remaining build issues.


mdietze:
@infotroph I think I addressed the suggestion you had on #2482 about naming parameters rather than referencing them by number

infotroph (approved):

PR #2751, opened 2020-11-19
NOAA GEFS download update 

kzarada:
## Description
Updated NOAA GEFS. Based off of Quinn's function on github (https://github.com/rqthomas/noaaGEFSpoint) but modified to work with pecan and to have the same output files as before. 

## Motivation and Context
Change was required because NOAA GEFS updated and the existing function no longer worked. 

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -0,0 +1,906 @@
...
+#' @param lat_list
+#' @param lon_list
+#' @param forecast_time
+#' @param forecast_date
+#' @param model_name_raw
+#' @param num_cores
+#' @param output_directory
+#'
+#' @return
+#' @export
```
infotroph, line 12: Are these helpers intended for use outside the data.atmosphere package, or are they just to support the download.NOAA_GEFS function? If the latter, I suggest leaving them unexported.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -0,0 +1,906 @@
...
+#'
+#' @examples
+noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date ,model_name_raw, output_directory, end_hr) {
+  
+  
+  download_neon_grid <- function(ens_index, location, directory, hours_char, cycle, base_filename1, vars,working_directory){
```
mdietze, line 18: There's a lot of references to NEON in this script, including in file naming conventions, which doesn't make sense to me for a general NOAA download function. I know a lot of this is borrowed from the NEON challenge, but this will be confusing in the future. Plus we want to have filenaming keep to established PEcAn patterns.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -0,0 +1,906 @@
...
+#' @title Downscale NOAA GEFS frin 6hr to 1hr
+#' @return None
+#'
+#' @param input_file, full path to 6hr file
+#' @param output_file, full path to 1hr file that will be generated
+#' @param overwrite, logical stating to overwrite any existing output_file
+#' @param hr time step in hours of temporal downscaling (default = 1)
+#' @export
+#'
+#' @author Quinn Thomas
```
mdietze, line 523: If there's code that Quinn wrote in another repo, and we're just copying that code, I'd ask whether there's a way to point to his code directly (e.g. load it as a package). Otherwise the codes will diverge through time and we'll miss (potentially important) updates.

modules/data.atmosphere/R/GEFS_helper_functions.R
kzarada: I removed the neon references

modules/data.atmosphere/R/GEFS_helper_functions.R
kzarada: I removed the export for the GEFS specific functions! And I consolidated the downscale functions that we had previously for GEFS into one file in the last push I made. I still have those as exported because they could be used to downscale other met data but I can remove that if you think it would be better not being exported

modules/data.atmosphere/R/GEFS_helper_functions.R
kzarada: I've changed a small part of most of the functions to get it to work within the framework we already had set up with GEFS and Pecan

modules/data.atmosphere/R/GEFS_helper_functions.R
mdietze: I always err on the side of exporting more, not less. The big disadvantage of exporting, in my mind, is that it potentially causes conflicts of multiple functions with the same name

infotroph: > I still have those as exported because they could be used to downscale other met data

👍 👍 That seems exactly right to me. Thanks!

mdietze:
first set of errors are coming from `pecan.depends.R` which needs to be updated

second set is coming from the build checks
```
R check of modules/data.atmosphere returned new problems:
checking for missing documentation entries ... WARNING: Undocumented code objects: 'downscale_repeat_6hr_to_hrly' 'downscale_ShortWave_to_hrly' 'downscale_solar_geom' 'downscale_spline_to_hrly'
checking package dependencies ... NOTE: Imports includes 37 non-default packages.
checking R code for possible problems ... NOTE: downscale_repeat_6hr_to_hrly: no visible binding for global variable '.'
checking R code for possible problems ... NOTE: downscale_repeat_6hr_to_hrly: no visible binding for global variable 'lead_var'
checking R code for possible problems ... NOTE: downscale_repeat_6hr_to_hrly: no visible binding for global variable 'time'
checking R code for possible problems ... NOTE: downscale_ShortWave_to_hrly: no visible binding for global variable '.'
checking R code for possible problems ... NOTE: downscale_ShortWave_to_hrly: no visible binding for global variable 'surface_downwelling_shortwave_flux_in_air'
checking R code for possible problems ... NOTE: downscale_ShortWave_to_hrly: no visible binding for global variable 'time'
checking R code for possible problems ... NOTE: downscale_ShortWave_to_hrly: no visible binding for global variable 'group_6hr'
checking R code for possible problems ... NOTE: downscale_solar_geom: no visible global function definition for 'median'
checking R code for possible problems ... NOTE: downscale_spline_to_hrly: no visible binding for global variable '.'
checking R code for possible problems ... NOTE: downscale_spline_to_hrly: no visible global function definition for 'spline'
checking R code for possible problems ... NOTE: process_gridded_noaa_download: no visible binding for global variable 'NOAA.member'
checking R code for possible problems ... NOTE: process_gridded_noaa_download: no visible binding for global variable 'air_temperature'
checking R code for possible problems ... NOTE: process_gridded_noaa_download: no visible binding for global variable 'time'
checking R code for possible problems ... NOTE: temporal_downscale: no visible binding for global variable 'relative_humidity'
```

kzarada:
@mdietze it's still failing the check but I don't see a specific error to fix. I addressed the issues you pointed out in the previous comment

PR #2753, opened 2020-11-25
Fix summarize.result so it generates statistics

jessicaguo:
## Description

Prior to this change, the `summarize.result` function did not summarize SE due to an ordering issue. n and mean were moved to below statname and stat.

previously, the `summarize.result` function included this:

```r
dplyr::summarize(
      n = length(n),	
      mean = mean(mean),	
      stat = stats::sd(mean) / sqrt(length(n))
```

Apparently, the author had expected `sd(mean)` and `length(n)` to use the values in the original data frame, but the summarize function first computed new summary values of mean and n, and since these were reduced to 1 value per group, didn't compute a statistic. 

We also created additional tests were created to ensure the new function works correctly.

## Motivation and Context

As a result, random effects were not being included in some meta-analysis models where these within group statistics were not estimated.

The fix was to calculate the stat before calculating new values of n and mean.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ This is a major change:
 - Changed docker-compose.yml to use user & group IDs of the operating system user (#2572)  
 - gSSURGO file download now added as inputs into BETY through extract_soil_gssurgo (#2666)
 - ensure Tleaf converted to K for temperature corrections in PEcAn.photosynthesis::fitA (#2726)
-- fix bug in summarize.result to output stat, which is needed to turn on RE in the meta-analysis (#2752)
+- fix bug in summarize.result to output stat, which is needed to turn on RE in the meta-analysis (#2753)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ This is a major change:
 - Changed docker-compose.yml to use user & group IDs of the operating system user (#2572)  
 - gSSURGO file download now added as inputs into BETY through extract_soil_gssurgo (#2666)
 - ensure Tleaf converted to K for temperature corrections in PEcAn.photosynthesis::fitA (#2726)
+- fix bug in summarize.result to output stat, which is needed to turn on RE in the meta-analysis (#2752)
```
dlebauer, line 40:
```suggestion
- fix bug in summarize.result to output stat, which is needed to turn on RE in the meta-analysis (#2753)
```

dlebauer (approved):

infotroph (approved):
Thanks for catching this, and for including a test case!

PR #2756, opened 2020-12-05
Track random effects in meta.analysis

jessicaguo:
## Description
Meta-analysis code updated to reflect two major changes:
1) bug fix: treatment random effect now always indexed 1 for control treatments. Previously, relied on alphabetical order, which would fail if a treatment name was alphabetically before "control"
2) update: output the data from jagify() to include the greenhouse, site, and treatment random effects indices. This will allow combining posterior chains more easily when multiple sites and treatments are involved. "jagged.data.Rdata" is a list of dataframes (length = number of traits) that replaces the previous "madata.Rdata". The first 6 columns of jagged.data is equivalent to madata and are the direct inputs to the meta-analysis model. The remaining columns serve to link observations to their BETYdb-assigned greenhouse, site, and treatment ids as well as control designations and random effects indices. 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/db/R/rename_jags_columns.R
```diff
@@ -18,18 +18,21 @@
 rename_jags_columns <- function(data) {
   
   # Change variable names and calculate obs.prec within data frame
+  # Swap column names
+  colnames(data)[colnames(data) %in% c("greenhouse", "ghs")] <- c("ghs", "greenhouse")
+  colnames(data)[colnames(data) %in% c("site_id", "site")] <- c("site", "site_id")
```
infotroph, line 23: Can you explain a bit why this is needed? On a quick read it looks as if these columns get swapped upstream in `jagify` and then swapped back here, but I'm not sure what happens to them in between.

infotroph (changes requested):

base/db/R/rename_jags_columns.R
jessicaguo: Apologies for the confusion. Basically, I am trying to retain both the original name of the identifying column and the new name (the id given by BETYdb and the index assigned in the meta-analysis). They get swapped back because of the column names required downstream by pecan.ma().

infotroph: That makes sense -- I didn't think to check downstream. 

It might be useful to add a comment here to that effect for future maintainers, but I'll approve this now so that my change request isn't blocking.

infotroph (approved):

dlebauer (approved):

istfer:
Note that documentation still mentions `madata.Rdata`, e.g. documentation/tutorials/02_Demo_Uncertainty_Analysis/Demo02.Rmd:49, book_source/03_topical_pages/92_workflow_modules.Rmd:27

base/utils/tests/testthat/test.utils.R
```diff
@@ -70,7 +70,9 @@ test_that("summarize.result works appropriately", {
                            n = 1,
                            mean = sqrt(1:10),
                            stat = NA,
-                           statname = NA
+                           statname = NA,
+                           name = NA,
```
robkooper, line 74: make sure to update the documentation for these 2 variables.

base/utils/R/utils.R
```diff
@@ -209,7 +209,7 @@ summarize.result <- function(result) {
     dplyr::filter(n == 1) %>%
     dplyr::group_by(citation_id, site_id, trt_id,
                     control, greenhouse, date, time,
-                    cultivar_id, specie_id) %>%
+                    cultivar_id, specie_id, name, treatment_id) %>%
```
robkooper, line 212: I think this is where it complains. @infotroph how should this be fixed?

dlebauer: no idea but perhaps it is related to this https://www.r-bloggers.com/2019/08/no-visible-binding-for-global-variable/ and also covered here: https://community.rstudio.com/t/how-to-solve-no-visible-binding-for-global-variable-note/28887

image

which implies that it could be fixed by changing this to 

```r
dplyr::group_by(.data$citation_id, .data$site_id, .data$trt_id,
                    .data$control, .data$greenhouse, .data$date, .data$time,	.data$control, .data$greenhouse, .data$date, .data$time,
                    .data$cultivar_id, .data$specie_id) %>%	                    cultivar_id, specie_id, name, treatment_id)
```

but its not clear why only the two new variables are called out in the NOTE. Or if we have to start prepending variables everywhere with `.data$` 

and FWIW this is a note - note even a warning

jessicaguo: Well I just corrected them to include .data$, but it has still failed that check.

infotroph: @dlebauer beat me to my answer: `.data` is indeed what you need here, and the reason it's needed is that `R CMD check`'s static scope analysis doesn't know about nonstandard evaluation, so it can't tell that these variables will be provided by the dataframe.

As for why the NOTE only calls out the two newly added ones: `R CMD check` actually does complain about all of them, but they get special exceptions because they already existed before I wrote the CI script. For gory details, see section 8.4.3, "Continuous Integration", in the [developer guide](https://pecanproject.github.io/pecan-documentation/develop/developer-testing.html). 

Now that you've fixed this NOTE, the special exceptions are no longer needed and you can remove them by deleting all the lines starting `summarize.result: no visible binding for global variable` in `base/utils/test/Rcheck_reference.log`

base/utils/R/utils.R
```diff
@@ -207,9 +207,9 @@ zero.bounded.density <- function(x, bw = "SJ", n = 1001) {
 summarize.result <- function(result) {
```
infotroph, line 207:
```suggestion
##' @importFrom rlang .data
summarize.result <- function(result) {
```

infotroph: I guess you'll need to add rlang to the Imports section of DESCRIPTION too. I'd do it here, but I don't think GitHub's suggest changes feature works on files not already changed in the PR.

infotroph (approved):

base/utils/R/utils.R
dlebauer: @infotroph I think you can suggest changes on the source branch, e.g. here: https://github.com/jessicaguo/pecan/edit/track-RE/base/utils/DESCRIPTION

I can do it but if this is the right fix, is this something that we will have to do throughout a lot of packages with similar dplyr syntax (it wasn't clear from the links I found)?

infotroph: > you could suggest changes on the source branch

Cool! Thank you.

> is this something that we will have to do throughout

Eventually, yes. It's definitely annoying to write, but IMO not any more so than the various other workarounds (e.g. `globalVariables`), and is the approach recommended by the Tidyverse developers. 

When should "eventually" be? It's an easy but tedious fix for a problem that's mostly just annoying (it could be masking bugs, but in practice we'd probably notice them some other way). So I'd say "as time allows, probably one package at a time whenever reminded by a PR like this one." By my count we're currently ignoring 571 `no visible binding for global variable` NOTEs across all of PEcAn, and my wild guess is that half to two-thirds of these are Tidyverse calls.

PR #2761, opened 2021-01-28
Patch for get.trait.data.pft.R

jessicaguo:
Added an if/else to the existing_membership check, which pulls in either species.csv or cultivars.csv for comparison. Crucially, the column classes differ between the two types. 

## Description
'species.csv' has 4 columns, the first of which has to be numeric. 'cultivars.csv' has 6 columns, the first 2 of which have to be numeric. 

## Motivation and Context
When the meta-analysis update is set to FALSE in the pecan.xml file, pfts with cultivars can now have meta-analysis results ported over from a previous run, reducing the need to rerun the meta-analysis each time. 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible
## Types of changes

## Checklist:


base/db/R/get.trait.data.pft.R
```diff
@@ -136,15 +136,26 @@ get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon, trait.names,
...
-          existing_membership <- utils::read.csv(
-            need_paths[["pft_membership"]],
-            # Columns are: id, genus, species, scientificname
-            # Need this so NA values are formatted consistently
-            colClasses = c("double", "character", "character", "character"),
-            stringsAsFactors = FALSE,
-            na.strings = c("", "NA")
-          )
-          diff_membership <- symmetric_setdiff(
+          if (pfttype == "plant") {
```
mdietze, line 139: There seems to be a lot more duplicate code here than necessary. Could we just have the `if` be used to define the colClasses variable, and then use that variable in the `existing_membership` load?

mdietze (changes requested):

base/db/R/get.trait.data.pft.R
jessicaguo: Okay, fixed to be much more compact. Thanks!

mdietze (approved):

PR #2764, opened 2021-02-02
db pool was throwing errors

robkooper:
the pool created errors, will investigate later, this should fix the issue.

I think the pool does not play nice with some of the queries we have and dplyr queries we have. Right now this seems to make it work correctly.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


robkooper:
did not create changelog entry since API was introduced in this release.

ashiklom:
Can you provide more context on the errors? The reason I created the pool was because without it, the API would start to choke if you submitted even a relatively small number of queries at a time.

mdietze (approved):

ashiklom:
We should definitely revisit this. I introduced the connection pool because the API DB connection would choke under even relatively light strain. Otherwise, I'd never encountered any errors with submitting jobs through the API with the DB pool enabled. I wish we could have seen more on the error before merging this.

robkooper:
I agree, I like the idea of the pool. Problem was that when I was testing the code it would fail to get the basic /api/info endpoint.

PR #2765, opened 2021-02-02
run push only on master repo

robkooper:
This should make it so the depend image is build and pushed to docker hub.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


mdietze (approved):

.github/workflows/depends.yml
```diff
@@ -48,7 +49,6 @@ jobs:
 
       # this will publish to the actor (person) github packages
       - name: Publish to GitHub
-        if: github.event_name == 'push'
         uses: elgohr/Publish-Docker-Github-Action@2.22
         env:
           R_VERSION: ${{ matrix.R  }}
```
ashiklom, line 54: Do we want images for both R 3.6 and R 4.0? My inclination would be to pick one R version (preferably, the most recent one that works, so 4.0) and stick with it for our Docker containers. I wonder if this is where our errors are coming from.

ashiklom: Or, at the very least, if we _are_ building images for 2 different R versions (which I think we're doing now based on my understanding of how the build matrix works), we should at least tag the images with the R version. And, for backwards compatibility, have the `develop` point to one of them.

robkooper: I have version that is supported `  SUPPORTED: 4.0.2` and only for those that we/I call supported it is pushed as latest/develop. The other should be pushed as pecan/depends:4.0.2 and pecan/depends:3.6

ashiklom (approved):

PR #2766, opened 2021-02-04
Using ed2_pft_number tag in ED2 E files postprocessing

KristinaRiemer:
## Description
In the post processing of ED2 h5 results files, in order to get and put the E files into the netcdf format, there is some matching of ED2 PFT numbers. Currently it is assumed that the Pecan PFT is in the [pftmapping file](https://github.com/PecanProject/pecan/blob/eb55663d4362562e60ed42bbae93969e1ad4ee04/models/ed/data/pftmapping.csv), but ED2 values can also be specified in the pecan.xml `ed2_pft_number` tag. 

I tried to fix this as best as possible. It works for my situation, but I only have one PFT in my runs right now and I'm using the `ed2_pft_number` for it. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:


models/ed/R/model2netcdf.ED2.R
```diff
@@ -937,7 +937,9 @@ read_E_files <- function(yr, yfiles, efiles, outdir, start_date, end_date, pft_n
   
   npft <- length(pft_names)
   data(pftmapping, package = "PEcAn.ED2")
-  pfts <- sapply(pft_names, function(x) pftmapping$ED[pftmapping$PEcAn == x]) 
+  pfts <- sapply(pft_names, function(x) ifelse(x %in% settings$pfts$pft$name, 
+                                               as.numeric(settings$pfts$pft$ed2_pft_number[settings$pfts$pft$name == x]),
+                                               pftmapping$ED[pftmapping$PEcAn == x]))
```
ashiklom, line 942: This has two issues:

- If the PFT `name` is provided but the `ed2_pft_number` is not, it will return `NULL` instead of grabbing the number from the `pftmapping`
- Because of the deep hierarchy of the PEcAn XML PFT section (roughly, something like `settings <- list(pfts = list(pft = list(name = "tree"), pft = list(name = "shrub")))`, it won't work for multiple PFTs.

I might suggest rewriting this block to something like this:

```r
  pfts <- numeric(npft)
  names(pfts) <- pft_names

  # Extract the PFT names and numbers for all PFTs.
  xml_pft_names <- lapply(settings$pfts, "[[", "name")
  for (pft in pft_names) {
    which_pft <- which(xml_pft_names == pft)
    xml_pft <- settings$pfts[[which_pft]]
    # If the PFT number is provided in the XML, use that.
    if ("ed2_pft_number" %in% names(xml_pft)) {
      pft_number <- as.numeric(xml_pft$ed2_pft_number)
      if (!is.finite(pft_number)) {
        PEcAn.logger::logger.severe(
          "ED2 PFT number present but not parseable as number. Value was `",
          xml_pft$ed2_pft_number "`".
        )
      }
    } else {
      # Otherwise, fall back on the PFT number in the `pftmapping`
      pft_number <- pftmapping$ED[pftmapping$PEcAn == x]
    }
    pfts[pft] <- pft_number
  }
```

ashiklom (changes requested):

models/ed/R/model2netcdf.ED2.R
ashiklom: Also, a third issue (caught by our `check` GH Action!): `settings` is not currently passed to this function. A hacky workaround would be to add it as an argument with a default value of `NULL` and modify the calling code to also pass the `settings` argument.

KristinaRiemer:
Hi @ashiklom. I incorporated your suggested changes, and they seemed to work, until I added in the `settings` argument. Now the `pft_names` arg isn't being read into `read_E_files`, so I think I must have done something wrong here?

ashiklom:
Without more information/context on how you're calling `read_E_files` and more information on the error, not sure I can help debug this. 

My best guess from your description is that, by inserting `settings` as the first argument in these functions (usually not a great idea for back compatibility -- better to add it as the last argument before the `...`, and give it a default value of `NULL`), you've shifted all the other arguments back by one and, in a later call, have passed the wrong number of arguments (since, unless specified as keyword arguments -- `myfun(key1 = value1, key2 = value2, <etc...>)` -- R will assign functions in order).

models/ed/R/model2netcdf.ED2.R
dlebauer: instead of sending in the entire settings, it seems that it would be cleaner (and inline w/ prev. discussions) to only send the element of the settings file that is needed. e.g. in this case you could replace two arguments to the function (the pft name and number) with a list object that is either (`pfts = list(pft = list( name = 'SetariaWT', ed2_pft_number = "1"))` or `pfts = settings$pfts`).

Similarly, the model2netcdf.ED function you could have arguments models and pfts, e.g. `, models = settings$models, pfts = settings$pfts`). Not necessary for this PR, but a way to reduce the requirement to have a full settings object (or a list named settings when you just need one or two components) for a particular function - easier testing, easier to understand, and a step closer to package independence.

dlebauer: @ashiklom could you please re-review this PR - does it address your requested changes?

ashiklom (approved):
I only had time to do a quick skim of the code, but assuming you've tested it and it works, I think it's fine to merge.

KristinaRiemer:
All of the GitHub actions are passing now. I had to do `make document` within the repo to fix the out of date file problems.

dlebauer:
Have to update the branch and rerun tests...

mdietze:
Just had a chance to look at the PR and want to raise the issue @ashiklom raised a while back about whether you actually need to add the `settings` object as an argument in these functions. Looking through the added code it seems to just be referencing `settings$pfts`, but it also looks like `pfts` is also being passed in (but not always used). It would be my preference to just pass in `pfts` and not both `settings` and `pfts`.

KristinaRiemer:
The `settings` argument is optional as was suggested by @ashiklom in [this comment](https://pecanproject.slack.com/archives/C8RD9BRLK/p1623245972019800?thread_ts=1623108106.012300&cid=C8RD9BRLK), I'm not sure anyone suggested removing that arg entirely?

PR #2768, opened 2021-02-09
WIP Fix Docker reproducibility

ashiklom:
**NOTE: Hold off on pulling this until I've removed the WIP marker. I want to see if GH Actions pass under this configuration.**

- Do not set `repos = ...` in install.packages calls. We should not be overwriting the user- or system-specific setting in `.Rprofile`. In particular, this can produce insidious bugs when interacting with Docker containers that are configured to use specific CRAN snapshots. We should defer to the Docker R image configuration to handle this, unless strictly necessary.
- Undo manual installing rlang and testthat in GH actions . If we defer to the correct Docker images with the correct fixed RStudio package manager snapshots, this should not be necessary

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,9 +105,6 @@ jobs:
     - name: install new dependencies
       run: Rscript scripts/generate_dependencies.R && Rscript docker/depends/pecan.depends.R
 
-    - name: Manually update rlang and testthat (HACK)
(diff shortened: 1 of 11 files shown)


ashiklom:
OK, so that didn't work, but I think it's because the underlying `pecan/depends` image is still using the more recent version of `devtools` because my change to the Dockerfile hasn't propagated to it yet. I'll have to think of a way to test this.

ashiklom:
Closing this for now and migrating conversation to #2769 .

Issue #2769, opened 2021-02-09
Improve Docker reproducibility


ashiklom:
For one, I think we should implement the changes in the closed PR https://github.com/PecanProject/pecan/pull/2768. But that requires rebuilding the `depends` image and some additional testing that I don't have time to do right now.

Basically, I think we have three options that take full advantage of the current Docker reproducibility mechanisms:

1. Stick with R 4.0.2 and accept that our packages will be a few months out of date.
2. Bump to R 4.0.3 and accept using the latest CRAN versions of all packages (until R is bumped to 4.0.4, when this image will be frozen).
3. Instead of starting from the `rocker/tidyverse` image, build R in our `pecan/depends` image from scratch (based on the [existing template](https://github.com/rocker-org/rocker-versioned2/blob/master/dockerfiles/Dockerfile_r-ver_4.0.3)), and manually specify (by modifying the CRAN URL) exactly which R package snapshots we want to work with.

We also have a few hacky options (like manually updating specific packages to specific versions and/or adding a bit of text editing code into our Dockerfile to update the image's `repos` definition in `.Rprofile`). But, my personal favorite is (3) above: Doing it is much easier than it sounds (the parent Dockerfile is pretty small, I think because the parent Ubuntu 20 image has all the dependencies already), and gives us maximum control over package versions. The ideal system would be to build 2-3 version of the `depends` image based on different snapshots ("stable", "next", "latest"?), test against all of them, and periodically bump what is considered "stable" and "next" whenever we're comfortable.

robkooper:
I would like to use the fixed versions of R, like 4.0.2 with a fixed cran repo, that way we have a consistent build.

We can even tell people to use when working with R on their local machine.

robkooper:
4.0.2 => ` ENV CRAN=https://packagemanager.rstudio.com/all/__linux__/focal/344`
4.0.3 => `ENV CRAN=https://packagemanager.rstudio.com/all/__linux__/focal/latest`

Should we build 4.0.3 and see if it is ready for the future?

ashiklom:
+1 to building against 4.0.2 and 4.0.3 -- that's a great addition.

> 4.0.2 => ENV CRAN=https://packagemanager.rstudio.com/all/__linux__/focal/344
> 4.0.3 => ENV CRAN=https://packagemanager.rstudio.com/all/__linux__/focal/latest

[...]

PR #2770, opened 2021-02-09
Use package manager

robkooper:
Do not set repos = ... in install.packages calls. We should not be overwriting the user- or system-specific setting in .Rprofile. In particular, this can produce insidious bugs when interacting with Docker containers that are configured to use specific CRAN snapshots. We should defer to the Docker R image configuration to handle this, unless strictly necessary.
Undo manual installing rlang and testthat in GH actions . If we defer to the correct Docker images with the correct fixed RStudio package manager snapshots, this should not be necessary

This requires a modified version of pecan/depends:4.0.2 (which i pushed) to pass the build.

This also removes all travis pieces.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (approved):
This looks good to me but would love for @ashiklom or @infotroph to comment

robkooper:
This should address #2769

.travis.yml
```diff
@@ -1,88 +0,0 @@
-language: r
```
infotroph, line 1: All this code will be lost in time like tears in rain...

...and good riddance ;)

book_source/02_demos_tutorials_workflows/05_developer_workflows/04-testing.Rmd
```diff
@@ -124,49 +124,26 @@ The `batch_run.R` script can take the following command-line arguments:
...
-* Runs package unit tests (the same ones you run locally with `make test` or `devtools::test(pkgname)`).
-    - As discussed in [Unit testing](#developer-testing-unit), these tests should run quickly and test individual components in relative isolation.
-    - Any test that calls the `skip_on_ci` function will be skipped. This is useful for tests that need to run for a very long time (e.g. large data product downloads) or require resources that aren't available on Travis (e.g. specific models), but be sure to run these tests locally before pushing your code!
-* Runs R package checks (the same ones you run locally with `make check` or `devtools::check(pkgname)`), skipping tests and documentation rebuild because we just did those in the previous steps.
-    - Any ERROR in the check output will stop the build immediately.
-    - If there are no ERRORs, any WARNINGs or NOTEs are compared against a stored historic check result in `<package>/tests/Rcheck_reference.log`. If the package has no stored reference result, all WARNINGs and NOTEs are considered newly added and reported as build failures.
```
infotroph, line 147: I believe this section is the only written documentation of how the cached check comparisons work. Should it be retained (possibly in another location)?

infotroph (changes requested):

book_source/02_demos_tutorials_workflows/05_developer_workflows/04-testing.Rmd
robkooper: use to big of an axe.

robkooper:
4.0.3 uses latest so it will no have a consistent build, i feel that it is ok if that breaks and note make it a required pass.

book_source/02_demos_tutorials_workflows/05_developer_workflows/04-testing.Rmd
robkooper: added text back

infotroph (approved):

ashiklom (approved):

apps/api/Dockerfile
```diff
@@ -18,9 +18,9 @@ EXPOSE 8000
 RUN   apt-get update \
   &&  apt-get install libsodium-dev -y \
   &&  rm -rf /var/lib/apt/lists/* \
-  &&  Rscript -e "devtools::install_version('promises', '1.1.0', repos = 'http://cran.rstudio.com')" \
-  &&  Rscript -e "devtools::install_version('webutils', '1.1', repos = 'http://cran.rstudio.com')" \
-  &&  Rscript -e "install.packages('pool', repos = 'http://cran.rstudio.com')" \
+  &&  Rscript -e "devtools::install_version('promises', '1.1.0')" \
+  &&  Rscript -e "devtools::install_version('webutils', '1.1')" \
```
ashiklom, line 22:
```suggestion
  &&  Rscript -e "devtools::install_version('promises', '1.1.0')" \
  &&  Rscript -e "devtools::install_version('webutils', '1.1')" \
```

apps/api/Dockerfile
```diff
@@ -18,9 +18,9 @@ EXPOSE 8000
...
+  &&  Rscript -e "install.packages('pool')" \
   &&  Rscript -e "devtools::install_github('rstudio/swagger')" \
   &&  Rscript -e "devtools::install_github('rstudio/plumber')"
```
ashiklom, line 25: Ditto here.

```suggestion
  &&  Rscript -e "remotes::install_github('rstudio/swagger')" \
  &&  Rscript -e "remotes::install_github('rstudio/plumber')"
```

docker/depends/Dockerfile
```diff
@@ -39,8 +39,8 @@ RUN apt-get update \
 # INSTALL DEPENDENCIES
 # ----------------------------------------------------------------------
 COPY pecan.depends.R /
-RUN  Rscript -e "install.packages(c('devtools'), repos = 'http://cran.rstudio.com')" \
-  && Rscript -e "devtools::install_version('roxygen2', '7.0.2', repos = 'http://cran.rstudio.com')" \
+RUN  Rscript -e "install.packages(c('devtools'))" \
+  && Rscript -e "devtools::install_version('roxygen2', '7.0.2')" \
```
ashiklom, line 43: Ditto here:

```suggestion
RUN  Rscript -e "install.packages(c('remotes'))" \
  && Rscript -e "remotes::install_version('roxygen2', '7.0.2')" \
```

ashiklom:
Looks good, thanks! Not gonna hold up the PR for this, but for future reference, I'd probably recommend switching to `remotes::install_*` (all my suggested edits) -- that's basically (exactly?) what `devtools::install_*` calls under the hood. But, the benefit of `remotes` is that it doesn't pull in (i.e., tries to update) all the other `devtools` packages, which, AFAICT, was the cause of all these issues to begin with (we were installing a more recent version of `devtools` but with an older version of `testthat` -- that would have been avoided if we'd just used `remotes` instead, which has no connection to `testthat`).

robkooper:
Merging this, will create a new PR to do a the conversion from devtools::install to remotes::install

Issue #2758, opened 2020-12-17
Fix tidyverse "no visible binding for global variable" notes


dlebauer:
Adapted from discussion with @infotroph in https://github.com/PecanProject/pecan/pull/2756

- an easy but tedious and inelegant fix for a problem that's mostly just annoying
- We are currently ignoring 571 `no visible binding for global variable` NOTEs across all of PEcAn, and my wild guess is that half to two-thirds of these are Tidyverse calls.

**Shortcu:t** For anyone who runs into this error and doesn't want to implement this fix: just add the appropriate note causing the build to fail to the package's `<pkg>/test/Rcheck_reference.log`, like this: https://github.com/PecanProject/pecan/pull/2756/commits/4a303b7fe3b32914c1f329be5efffc5f55030339

**If you are in a good place to tackle this:** we support you, get comfy! Bonus points if you fix all of the issues in the package you are working on. A **special prize** if you fix all (currently 563) of these Notes.

TODO:

* [ ] find list of violations in `<pkg>/test/Rcheck_reference.log` files
  * ` find . -name Rcheck_reference.log | xargs grep "no visible binding for global variable"` 
  *  to count occurrences add ` | wc -l` to the end of the line above
* [ ] prepend variables with `.data$` as done here: https://github.com/PecanProject/pecan/pull/2756/files#diff-296112e719b8455a8be63e9cac941111d4ae1700081aa05b3a9adfa8d5b9738aR210
* [ ] add rlang to Imports for relevant package to description
* [ ] add `##' @importFrom rlang .data` to `<pkg>R/PEcAn.<pkgname>-package.R`. See [PEcAn.DB](https://github.com/PecanProject/pecan/blob/develop/base/db/R/PEcAn.DB-package.R#L8) for an example.
  * can also be done per function, but not sure when this would be preferred
* [ ] delete (relevant) lines starting `<function name>: no visible binding for global variable <variable>` in `<pkg>/test/Rcheck_reference.log`
* [ ] update documentation

Alternative solutions:
- change build checks to allow these notes
- or add notes as they appear to `<pkg>/test/Rcheck_reference.log` so that build ignores it
- wait until a more elegant solution evolves

See also 
- https://community.rstudio.com/t/how-to-solve-no-visible-binding-for-global-variable-note/28887
- https://www.r-bloggers.com/2019/08/no-visible-binding-for-global-variable/
- https://www.google.com/search?q=no+visible+binding+for+global+variable
[...]

PR #2773, opened 2021-02-18
Updated base/utils #2758

moki1202:
fixed tidyverse notes for the remaining variables in base/utils
convert.input :'id'
summarize.result: 'n' , 'statname'


base/utils/R/utils.R
```diff
@@ -207,21 +207,21 @@ zero.bounded.density <- function(x, bw = "SJ", n = 1001) {
...
 summarize.result <- function(result) {
   ans1 <- result %>%
-    dplyr::filter(n == 1) %>%
+    dplyr::filter(.data$n == 1) %>%
     dplyr::group_by(.data$citation_id, .data$site_id, .data$trt_id,
                     .data$control, .data$greenhouse, .data$date, .data$time,
                     .data$cultivar_id, .data$specie_id, .data$name, .data$treatment_id) %>%
     dplyr::summarize( # stat must be computed first, before n and mean
-      statname = dplyr::if_else(length(n) == 1, "none", "SE"),
+      .data$statname = dplyr::if_else(length(n) == 1, "none", "SE"),
```
infotroph, line 215:
```suggestion
      statname = dplyr::if_else(length(.data$n) == 1, "none", "SE"),
```

the `.data` verb isn't needed (nor allowed) on the left side of an assignment. If this is a surprise, recall that the reason for using `.data` is to avoid confusion about which scope a name is coming from. But here there's no ambiguity -- `foo = bar` always creates a new `foo` in the current scope whether or not one exists elsewhere.

infotroph (changes requested):

base/utils/R/utils.R
moki1202: @infotroph thank you for pointing this out , is there any other fixes to be done here?

base/utils/R/utils.R
```diff
@@ -207,21 +207,21 @@ zero.bounded.density <- function(x, bw = "SJ", n = 1001) {
...
+    dplyr::filter(.data$n == 1) %>%
     dplyr::group_by(.data$citation_id, .data$site_id, .data$trt_id,
                     .data$control, .data$greenhouse, .data$date, .data$time,
                     .data$cultivar_id, .data$specie_id, .data$name, .data$treatment_id) %>%
     dplyr::summarize( # stat must be computed first, before n and mean
-      statname = dplyr::if_else(length(n) == 1, "none", "SE"),
+    statname = dplyr::if_else(length(.data$n) == 1, "none", "SE"),
       stat = stats::sd(mean) / sqrt(length(n)),
-      n = length(n),
+      .data$n = length(n),
```
ashiklom, line 217:
```suggestion
      n = length(.data$n),
```

base/utils/R/utils.R
```diff
@@ -207,21 +207,21 @@ zero.bounded.density <- function(x, bw = "SJ", n = 1001) {
...
       mean = mean(mean)
     ) %>%
     dplyr::ungroup()
   ans2 <- result %>%
-    dplyr::filter(n != 1) %>%
+    dplyr::filter(.data$n != 1) %>%
     # ANS: Silence factor to character conversion warning
-    dplyr::mutate(statname = as.character(statname))
+    dplyr::mutate(.data$statname = as.character(.data$statname))
```
ashiklom, line 224:
```suggestion
    dplyr::mutate(statname = as.character(.data$statname))
```

ashiklom (changes requested):
A few other places where @infotroph 's comment about `.data` on the left side of expressions applies.

moki1202:
@ashiklom 
error 1
what about the 'n' inside sqrt() function . it will need a .data pronoun as well right?

ashiklom:
@moki1202 Yes, good find. Hard to say without more context in the file, but I suspect that line should have both `.data$mean` and `.data$sd`.

moki1202:
@ashiklom why didn't these variables ( 'mean' and 'sd' in this case ) throw a R CMD check error? because I didn't see them in the Rcheck_reference.log file.

ashiklom:
> why didn't these variables ( 'mean' and 'sd' in this case ) throw a R CMD check error? because I didn't see them in the Rcheck_reference.log file.

Good question! I'm pretty sure it's because `mean` and `sd` are pre-defined functions in base R, so the error that `No object "mean" / "sd" exists in the global environment` wouldn't actually apply! To `R CMD check`, it looks like you are referring to R's default `mean` and `sd` functions, which is a perfectly valid thing to do (even though that's not actually what the code is doing).

moki1202:
@ashiklom so if I use  '.data$' on every variable used inside "dplyr" functions, it wouldn't harm the overall result?

ashiklom:
> so if I use '.data$' on every variable used inside "dplyr" functions, it wouldn't harm the overall result?

That's correct _unless the target variable really is something outside of the data_. For example, consider the following two cases:

```r
my_constant <- 8.3145
my_data %>%
  mutate(new_variable = .data$old_variable * my_constant)

my_data %>%
  filter(.data$some_column < my_constant)
```

In both cases, `my_constant` really is an object outside of the data frame that I want to use inside the expression. 

Fortunately, `R CMD check` will (correctly) mark both of these cases as valid. That's why it's generally a good idea to limit this to the `R CMD check` warnings. The `mean` and `sd` cases above were places I happened to recognize as needing this too.

base/utils/R/utils.R
```diff
@@ -27,7 +27,7 @@
 ##' @return ncvar based on MstMIP definition
 ##' @author Rob Kooper
 mstmipvar <- function(name, lat = NA, lon = NA, time = NA, nsoil = NA, silent = FALSE) {
-  nc_var <- PEcAn.utils::mstmip_vars[PEcAn.utils::mstmip_vars$Variable.Name == name, ]
+nc_var <- PEcAn.utils::mstmip_vars[PEcAn.utils::mstmip_vars$Variable.Name == name, ]
```
infotroph, line 30:
```suggestion
  nc_var <- PEcAn.utils::mstmip_vars[PEcAn.utils::mstmip_vars$Variable.Name == name, ]
```

(To keep the code indented inside the function definition... and assuming my eyes aren't lying to me about the current indent!)

base/utils/R/utils.R
```diff
@@ -207,21 +207,21 @@ zero.bounded.density <- function(x, bw = "SJ", n = 1001) {
...
+    dplyr::filter(.data$n == 1) %>%
     dplyr::group_by(.data$citation_id, .data$site_id, .data$trt_id,
                     .data$control, .data$greenhouse, .data$date, .data$time,
                     .data$cultivar_id, .data$specie_id, .data$name, .data$treatment_id) %>%
     dplyr::summarize( # stat must be computed first, before n and mean
-      statname = dplyr::if_else(length(n) == 1, "none", "SE"),
-      stat = stats::sd(mean) / sqrt(length(n)),
-      n = length(n),
+      statname = dplyr::if_else(length(.data$n) == 1, "none", "SE"),
+      stat = stats::.data$sd(.data$mean) / sqrt(length(.data$n)),
```
infotroph, line 216:
```suggestion
      stat = stats::sd(.data$mean) / sqrt(length(.data$n)),
```

We actually do want the `stats::sd` function here rather than a variable in the data

infotroph (changes requested):
This is looking very good! One indentation fix and one place we want a  function (`stats::sd`) rather than `..data$sd`. 

Once these changes are made, would you be willing to also delete the warnings that have been fixed from `base/utils/test/Rcheck_reference.log`? That will make sure the build breaks if anyone adds the same warnings back in a future commit.

base/utils/R/utils.R
infotroph: @ashiklom beat me to it. Looks like you resolved these below.

base/utils/R/utils.R
moki1202: @infotroph updated both these files.

base/utils/R/utils.R
```diff
@@ -27,7 +27,7 @@
 ##' @return ncvar based on MstMIP definition
 ##' @author Rob Kooper
 mstmipvar <- function(name, lat = NA, lon = NA, time = NA, nsoil = NA, silent = FALSE) {
-  nc_var <- PEcAn.utils::mstmip_vars[PEcAn.utils::mstmip_vars$Variable.Name == name, ]
+ nc_var <- PEcAn.utils::mstmip_vars[PEcAn.utils::mstmip_vars$Variable.Name == name, ]
```
infotroph, line 30:
```suggestion
  nc_var <- PEcAn.utils::mstmip_vars[PEcAn.utils::mstmip_vars$Variable.Name == name, ]
```

base/utils/tests/Rcheck_reference.log
```diff
@@ -48,7 +48,6 @@ Non-standard file/directory found at top level:
 * checking foreign function calls ... OK
 * checking R code for possible problems ... NOTE
 convert.input: no visible binding for global variable ‘settings’
-convert.input: no visible binding for global variable ‘id’
 convert.input : log_format_df: no visible binding for global variable
```
infotroph, line 51:
```suggestion
convert.input: no visible binding for global variable ‘id’
convert.input : log_format_df: no visible binding for global variable
```

infotroph, line 51:
```suggestion
convert.input: no visible binding for global variable ‘id’
convert.input : log_format_df: no visible binding for global variable
```

infotroph (approved):
I took the liberty of committing my last three suggestions right away, so I could see whether they were enough to make the CI build pass:
* Spacing fix: this is the same change you already committed, no idea why it wasn't showing up 🤷‍♂️ 
* Re-added warning to Rcheck_reference.log about `id` in `convert.inputs`, since it's not fixed yet. Cleaning up that file can be a PR of its own!
* removed stray text at end of file

The remaining build error on 4.0.2 looks like a transient package manager issue -- the checks for the utils package all pass before it reaches the error -- so I think this PR is good to merge as it is.

moki1202:
@infotroph thankyou. I will look into the convert.inputs file once again just to verify that I didn't leave a variable unassigned. Also if I wanted to apply for GSoC this summer, what other tools would you recommend me to learn prior to applying?

infotroph:
@moki1202 That depends what kind of project you're interested in working on! 

We've listed some of the major prerequisites for each of the projects in our [list of GSoC project ideas](https://pecanproject.github.io/gsoc_ideas.html], but in broad strokes we're always looking for people who have skill/interest in one or more of:

* Programming and package development in R (R, devtools, renv, ...)
* Unix/sysadmin tools (bash, make, HPC, ...)
* Data provenance, storage, transformations, and databases (postgres, netcdf, ...)
* Web/API design and programming (HTML, Flask, REST, small amounts JS, Shiny, ...)
* Containerization and orchestration (Docker, kubernetes, Singularity, ...)
* Visualizing data (ggplot, Shiny, HTML, ...)
* Unit and integration testing (testthat, GitHub Actions, ...)
* And of course the whole PEcAn project exists to facilitate ecosystem modeling, so it helps (but is definitely not necessary!) to have some familiarity with the models we support and what they do.

This list is an "any of these areas" not an "all of these" -- *nobody* on the team is an expert in all of them -- and if you bring strong skills from any one group we can help you learn what you need from the others. If this skills list looks like you but you don't see anything that excites you in the project ideas, join the [PEcAn Slack community](https://pecanproject.slack.com/join/shared_invite/enQtMzkyODUyMjQyNTgzLWEzOTM1ZjhmYWUxNzYwYzkxMWVlODAyZWQwYjliYzA0MDA0MjE4YmMyOTFhMjYyMjYzN2FjODE4N2Y4YWFhZmQ) and ask around -- we can probably point you toward outstanding issues that could be developed into your own project proposal.

moki1202:
@ashiklom @infotroph if there's still some errors left  please update me , ill try to fix them ASAP.

ashiklom (approved):

moki1202:
@ashiklom @infotroph what about this issue? What's causing the 1 failed test?

infotroph:
@moki1202 It looks like the test failure here is intermittent trouble with the Rstudio package manager, and not related to your changes. I think this is ready to merge as it is.

PR #2775, opened 2021-02-19
updated modules/data.atmosphere

moki1202:
fixed tidyverse notes for the remaining variables.


modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -73,10 +73,10 @@ download.NARR_site <- function(outfolder,
 
   narr_proc <- result_full %>%
     dplyr::mutate(
-      data_nc = purrr::map2(data, file, prepare_narr_year, lat = lat, lon = lon)
+      data_nc = purrr::map2(.data$data, file, prepare_narr_year, lat = lat, lon = lon)
```
infotroph, line 76:
```suggestion
      data_nc = purrr::map2(.data$data, .data$file, prepare_narr_year, lat = lat, lon = lon)
```

infotroph: Hmm. @ashiklom is `narr_proc` needed at all? I don't see it used anywhere between here and the end of the function.

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -225,7 +225,7 @@ get_NARR_thredds <- function(start_date, end_date, lat.in, lon.in,
     PEcAn.logger::logger.info("Downloading in parallel")
     flx_df$flx <- TRUE
     sfc_df$flx <- FALSE
-    get_dfs <- dplyr::bind_rows(flx_df, sfc_df)
+    get_dfs <- dplyr::bind_rows(.data$flx_df, sfc_df)
```
infotroph, line 228:
```suggestion
    get_dfs <- dplyr::bind_rows(flx_df, sfc_df)
```

Needed because (1) we want the `flx_df` defined on line 207, and (2) the arguments to `bind_rows` need to be whole dataframes rather than single columns.

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -296,19 +296,19 @@ post_process <- function(dat) {
   dat %>%
     tidyr::unnest(data) %>%
     dplyr::ungroup() %>%
-    dplyr::mutate(datetime = startdate + lubridate::dhours(dhours)) %>%
-    dplyr::select(-startdate, -dhours) %>%
+    dplyr::mutate(datetime = .data$startdate + lubridate::.data$dhours(.data$dhours)) %>%
```
infotroph, line 299:
```suggestion
    dplyr::mutate(datetime = .data$startdate + lubridate::dhours(.data$dhours)) %>%
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
   )
   tibble::tibble(date = dates) %>%
     dplyr::mutate(
-      year = lubridate::year(date),
-      month = lubridate::month(date),
-      daygroup = daygroup(date, flx)
+      year = lubridate::.data$year(date),
+     month = lubridate::.data$month(date),
+      daygroup = daygroup(date, .data$flx)
```
infotroph, line 327:
```suggestion
      year = lubridate::.year(.data$date),
      month = lubridate::month(.data$date),
      daygroup = daygroup(.data$date, .data$flx)
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
...
     ) %>%
-    dplyr::group_by(year, month, daygroup) %>%
+    dplyr::group_by(.data$year, .data$month, daygroup) %>%
```
infotroph, line 329:
```suggestion
    dplyr::group_by(.data$year, .data$month, .data$daygroup) %>%
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
...
     dplyr::summarize(
       startdate = min(date),
```
infotroph, line 331:
```suggestion
      startdate = min(.data$date),
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
...
         "%s/%d/NARR%s_%d%02d_%s.tar",
         base_url,
-        unique(year),
+        unique(.data$year),
         tag,
-        unique(year),
-        unique(month),
+        unique(.data$year),
+        unique(.data$month),
         unique(daygroup)
```
infotroph, line 339:
```suggestion
        unique(.data$daygroup)
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
...
       )
     ) %>%
     dplyr::ungroup() %>%
-    dplyr::select(startdate, url)
+    dplyr::select(.data$startdate, url)
```
infotroph, line 343:
```suggestion
    dplyr::select(.data$startdate, .data$url)
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -382,12 +382,12 @@ get_narr_url <- function(url, xy, flx, pb = NULL) {
   if (dhours[1] == 3) dhours <- dhours - 3
   narr_vars <- if (flx) narr_flx_vars else narr_sfc_vars
   result <- purrr::pmap(
-    narr_vars %>% dplyr::select(variable = NARR_name, unit = units),
+    narr_vars %>% dplyr::select(variable = .data$NARR_name, unit = units),
```
infotroph, line 385:
```suggestion
    narr_vars %>% dplyr::select(variable = .data$NARR_name, unit = .data$units),
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -382,12 +382,12 @@ get_narr_url <- function(url, xy, flx, pb = NULL) {
...
     read_narr_var,
     nc = nc, xy = xy, flx = flx, pb = pb
   )
   names(result) <- narr_vars$CF_name
-  dplyr::bind_cols(dhours = dhours, result)
+  dplyr::bind_cols(dhours = .data$dhours, result)
```
infotroph, line 390:
```suggestion
  dplyr::bind_cols(dhours = dhours, result)
```

modules/data.atmosphere/R/downscaling_helper_functions.R
```diff
@@ -78,59 +78,59 @@ downscale_ShortWave_to_hrly <- function(df,lat, lon, hr = 1){
       data.hrly$group_6hr[i] <- group
     }
   }
-  
+
   ShortWave.ds <- data.hrly %>%
-    dplyr::mutate(hour = lubridate::hour(.data$time)) %>%
-    dplyr::mutate(doy = lubridate::yday(.data$time) + hour/(24/hr))%>%
-    dplyr::mutate(rpot = downscale_solar_geom(doy, as.vector(lon), as.vector(lat))) %>% # hourly sw flux calculated using solar geometry
+    dplyr::mutate(hour = lubridate::.data$hour(.data$time)) %>%
```
infotroph, line 83:
```suggestion
    dplyr::mutate(hour = lubridate::hour(.data$time)) %>%
```

infotroph (changes requested):
Most the suggested changes here are places I saw names in the data that reuse names defined elsewhere, and are therefore not showing up in the check warnings -- adding `.data` to these shouldn't change the code behavior, but disambiguating them should make it a lot easier to read the code!

Since most of these changes are additional edits rather than corrections, tagging @ashiklom to review them as the person most likely to know if I mangled a global that should stay global.

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -258,7 +258,7 @@ get_NARR_thredds <- function(start_date, end_date, lat.in, lon.in,
           url,
           robustly(get_narr_url, n = 20, timeout = 1),
           xy = xy,
-          flx = TRUE,
+          .data$flx = TRUE,
```
infotroph, line 261:
```suggestion
          flx = TRUE,
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -269,7 +269,7 @@ get_NARR_thredds <- function(start_date, end_date, lat.in, lon.in,
           url,
           robustly(get_narr_url, n = 20, timeout = 1),
           xy = xy,
-          flx = FALSE,
+          .data$flx = FALSE,
```
infotroph, line 272:
```suggestion
          flx = FALSE,
```

moki1202: @infotroph if we have a constant value at the right hand side of the argument (flx = constant) will flx here need ' .data' pronoun.

infotroph: @moki1202 I don't believe `.data` is ever allowed on the left no matter what's on the right side. 

But the key tricky bit here is that `flx = FALSE` gets parsed as an argument to `map` which is then passed on as an argument to `get_narr_url`,  not as an argument to `mutate`. The nesting is kind of confusing, but the overall effect is approximately equivalent to this loop (I'll ignore `robustly`, which is a wrapper that repeatedly calls `get_narr_url` until it succeeds):

```
xy <- "a lat/lon value defined outside the dataframe" 

sfc_data_raw <- sfc_df
sfc_data_raw$data <- NULL
for (i in seq_along(sfc_df$url)) {
  sfc_data_raw$data[[i]] <- get_narr_url(sfc_df$url[[i]], xy = xy, flx = FALSE)
}
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
   )
   tibble::tibble(date = dates) %>%
     dplyr::mutate(
-      year = lubridate::year(date),
-      month = lubridate::month(date),
-      daygroup = daygroup(date, flx)
+      year = lubridate::year(.data$date),
+     month = lubridate::month(.data$date),
```
infotroph, line 326:
```suggestion
      month = lubridate::month(.data$date),
```

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -86,8 +86,8 @@ download.NARR_site <- function(outfolder,
 #' @param file Full path to target file
 #' @param lat_nc `ncdim` object for latitude
 #' @param lon_nc `ncdim` object for longitude
-#' @param verbose 
-#' @return List of NetCDF variables in data. Creates NetCDF file containing 
+#' @param verbose
```
infotroph, line 89:
```suggestion
#' @param verbose logical: ask`ncdf4` functions to be very chatty while they work?
```

Documenting arguments is always good, but the reason we need this now is kind of 🙄: With the trailing space included, Roxygen was previousy complaining about "argument items with no description" and we ignored it in Rcheck_reference.log. Without the trailing space, Roxygen complains instead about "undocumented arguments" and causes a CI failure because the ignore-preexisting-errors script only knows how to skip exact test matches.

After fixing this we should delete the "with no description" from Rcheck_reference.log (currently lines 420-421).

modules/data.atmosphere/R/check_met_input.R
```diff
@@ -94,7 +94,7 @@ check_met_input_file <- function(metfile,
     test_raw = purrr::map(nc_vars, check_unit, nc = nc, variable_table = variable_table),
     test_passed = !purrr::map_lgl(test_raw, inherits, "try-error"),
     test_error_message = purrr::map_chr(test_raw, purrr::possibly(as.character, NA_character_))
```
infotroph, line 96:
```suggestion
    test_passed = !purrr::map_lgl(.data$test_raw, inherits, "try-error"),
    test_error_message = purrr::map_chr(.data$test_raw, purrr::possibly(as.character, NA_character_))
```

TIL you can reuse previously defined columns inside a tibble definition! I had assumed this kind of transformation would need to be written `df <- tibble::tibble(col_a = ...) %>% mutate(col_b = f(col_a, ...),...)`

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -296,19 +296,19 @@ post_process <- function(dat) {
   dat %>%
     tidyr::unnest(data) %>%
```
infotroph, line 297:
```suggestion
    tidyr::unnest(.data$data) %>%
```

infotroph (changes requested):
A few more reported by the CI checks.

modules/data.atmosphere/R/check_met_input.R
ashiklom: Yup! IIRC, this is one of the key differences between `tibble::tibble` and `data.frame` -- it will _not_ work in the latter.

ashiklom (approved):
I think this looks good (after @infotroph 's suggestions)!

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -1,4 +1,4 @@
-#' Download NARR time series for a single site
+=#' Download NARR time series for a single site
```
infotroph, line 1:
```suggestion
#' Download NARR time series for a single site
```

moki1202:
@ashiklom @infotroph can you guys suggest to me the next issue I should start working on?

moki1202:
@infotroph is there any other changes required here?

infotroph:
@moki1202 Sorry for the delay here. The test failure in https://github.com/PecanProject/pecan/runs/2056814058#step:9:1353 ("column `flx` not found in `.data`") probably means I was wrong about at least one of the `.data` additions I suggested, but at a glance it sure looks like flx ought to be a column. 

The next step is probably to run the tests locally and step through to verify what columns are actually present in the data at that step.

moki1202:
@infotroph gave it a thorough check in the file download.NARR_site.R and didn't find any line where ' fix ' variable was not properly called using the ' .data$ ' pronoun....although, I did find out that I actually missed deleting the respective line regarding global variable error for ' flx ' variable in the Rcheck_reference.R . could this be a problem?

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,25 +322,25 @@ generate_narr_url <- function(dates, flx) {
   )
   tibble::tibble(date = dates) %>%
     dplyr::mutate(
-      year = lubridate::year(date),
-      month = lubridate::month(date),
-      daygroup = daygroup(date, flx)
+      year = lubridate::year(.data$date),
+      month = lubridate::month(.data$date),
+      daygroup = daygroup(.data$date, .data$flx)
```
ashiklom, line 327: `flx` here is an argument to the parent function, not a variable in the data frame. I missed this on my first review.

```suggestion
      daygroup = daygroup(.data$date, flx)
```

moki1202: @ashiklom fixed this.

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -322,29 +322,29 @@ generate_narr_url <- function(dates, flx) {
...
       )
     ) %>%
     dplyr::ungroup() %>%
-    dplyr::select(startdate, url)
+    dplyr::select(.data$startdate, .data$url)
 }
 
 #' Assign daygroup tag for a given date
-daygroup <- function(date, flx) {
+daygroup <- function(date, .data$flx) {
```
infotroph, line 347:
```suggestion
daygroup <- function(date, flx) {
```

infotroph (approved):

modules/data.atmosphere/R/download.NARR_site.R
```diff
@@ -236,8 +236,8 @@ get_NARR_thredds <- function(start_date, end_date, lat.in, lon.in,
       ),
         robustly(get_narr_url)(url, xy = xy, flx = flx)
```
infotroph, line 237: This line now fails with a global variable warning -- It's the same problem as in dplyr functions (non-standard evaluation leading to the variable name being defined somewhere R's scope analysis doesn't know about), but we can't fix it by changing to `.data$flx` because foreach doesn't know how to use the rlang `.data` adverb. The two easiest options I see:

a. Re-add the line `get_NARR_thredds: no visible binding for global variable 'flx'` to `Rcheck_reference.log` and therefore keep ignoring the message
b. Add `flx <- NULL` between current lines 230 and 231, so that R can see it's defined but it doesn't interfere with the definitions we actually use (which are provided by `foreach`)

I'd slightly favor the NULL approach, but I'll defer to @ashiklom's opinion here if he has one.

moki1202: @infotroph just out of curiosity , how did you find the specific spot i.e. between line 230 and 231 for ' flx <- NULL ' to be added?

infotroph: @moki1202 I looked for the closest place that immediately before the problematic code but wasn't in the middle of another function or flow control operation, and since lines 232-238 were all part of the same variable assignment I decided to put it on the line before that. 

The other option would be to put it at the very top of the file, with a comment explaining what it does. That would make sense if we had a file with a lot of `foreach` calls in it that used the same variables over and over, but for the case where there's only one line complaining I think it makes sense to keep the fix as close to the issue as possible.

moki1202: @infotroph thank you for the clarification.

modules/data.atmosphere/tests/Rcheck_reference.log
```diff
@@ -448,17 +416,14 @@ Argument items with no description in Rd object 'met.process.stage':
...
-
 Argument items with no description in Rd object 'split_wind':
   ‘start_date’ ‘end_date’
 
 * checking for unstated dependencies in examples ... OK
 * checking contents of ‘data’ directory ... OK
 * checking data for non-ASCII characters ... OK
 * checking data for ASCII and uncompressed saves ... WARNING
-  
+
```
infotroph, line 426: Apparently the check script doesn't know how to handle whitespace changes, so this deletion was making the "significantly better compression..." warning reappear.

```suggestion
  
```

infotroph (approved):

moki1202:
@infotroph @ashiklom does this PR need any further work?

infotroph:
@moki1202 Nope, we're just waiting for the build to turn green. The current failures are "unable to verify current time", which is apparently caused by R itself trying to consult a flaky third-party API. So *definitely* not your problem!

PR #2777, opened 2021-02-22
WIP HACK: Fix GH actions coda and mvtnorm install

ashiklom:
Current builds are failing the `check` CI test for R4.0.2 because of the following:

```
  ── Newly failing
  
  ✖ checking whether package ‘PEcAn.allometry’ can be installed ... WARNING
  
  R check of modules/allometry reports the following new problems. Please fix these and resubmit:
  checking whether package ‘PEcAn.allometry’ can be installed ... WARNING
  Found the following significant warnings:
    Warning: package ‘coda’ was built under R version 4.0.3
    Warning: package ‘mvtnorm’ was built under R version 4.0.3
  See ‘/tmp/RtmpYbnRSL/PEcAn.allometry.Rcheck/00install.out’ for details.
  make: *** [Makefile:139: .check/modules/allometry] Error 1
```

As best I can tell, this is a bug in the R Package Manager, not something broken on our end. This PR implements a hack where we reinstall `coda` and `mvtnorm` from source to ensure R version consistency. Not great, but hopefully will fix the issue for now.

PR #2778, opened 2021-02-22
HACK: Install some packages from source by hand

ashiklom:
They are not cached properly by the RStudio Package Manager. For whatever reason, the default versions of these packages are compiled under R v4.0.3, even for R version 4.0.2. So we have to re-compile them from source by hand. This is an imperfect fix, but should (temporarily) stop our 4.0.2 builds from breaking.

The alternative is #2777 , which implements this only via GH Actions.

Another alternative is to advance to 4.0.3 -- now that R 4.0.4 is out, we can treat that as the latest and use 4.0.3 as the default for CI (though we can still check builds against 4.0.2 if we want).

NOTE: This will _not_ resolve current build errors until it is merged (because the base Docker image won't be pushed yet). However, once it is merged, subsequent builds should start succeeding again.


mdietze (approved):

ashiklom:
Superseded by #2779.

PR #2779, opened 2021-02-22
Bump default R version to 4.0.3

ashiklom:
Our 4.0.3 builds and checks are already passing, and now that R 4.0.4 is released, 4.0.3 packages are frozen to Feb 17, 2021. Meanwhile, our 4.0.2 builds are failing for no fault of our own (looks like an error with the R Package Manager).

NOTE: I think this will fail the 4.0.4 checks (not required) until it's merged because we haven't build the containers yet. I already changed the branch protection rules to use the 4.0.3 builds and checks instead of 4.0.2.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest  
 
     container: 
-      image: pecan/depends:R4.0.2
+      image: pecan/depends:R4.0.3
 
     steps:
     # checkout source code
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,8 +30,8 @@ jobs:
       fail-fast: false
       matrix:
         R:
-          - "4.0.2"
           - "4.0.3"
+          - "4.0.4"
 
     env:
       NCPUS: 2
@@ -76,8 +76,8 @@ jobs:
       fail-fast: false
       matrix:
         R:
-          - "4.0.2"
           - "4.0.3"
+          - "4.0.4"
 
     services:
       postgres:
@@ -129,8 +129,8 @@ jobs:
       fail-fast: false
       matrix:
         R:
-          - "4.0.2"
           - "4.0.3"
+          - "4.0.4"
 
     env:
       NCPUS: 2
@@ -175,8 +175,8 @@ jobs:
       fail-fast: false
       matrix:
         R:
-          - "4.0.2"
           - "4.0.3"
+          - "4.0.4"
 
     services:
       postgres:
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -12,7 +12,7 @@ on:
 
 env:
   # official supported version of R
-  SUPPORTED: 4.0.2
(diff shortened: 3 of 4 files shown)


mdietze (approved):

robkooper (approved):
LGTM, depends will be build immediatly when this is merged, and we need to update other PR to have them use the new 4.0.3 builds.

Should update the CHANGELOG to reflect that we now use 4.0.3 in our images.

infotroph:
This whole episode makes me worried about version stability -- I thought avoiding these issues was the point of building on Rocker versioned images! 

But I don't know how to fix this without making it unduly hard to update when we really do need the latest version of a package

ashiklom:
> This whole episode makes me worried about version stability -- I thought avoiding these issues was the point of building on Rocker versioned images!

@infotroph I think we're getting there. Now that R 4.0.4 is out, the packages associated with 4.0.3 [should now be frozen](https://github.com/rocker-org/rocker-versioned2/blob/master/dockerfiles/Dockerfile_r-ver_4.0.3#L13). My best guess is that the isolated R 4.0.2 errors were some kind of weird, 🤞 one-time 🤞 hiccup in the RStudio Package Manager that won't happen again.

I still think it's worth considering my suggestion in #2769 to build our own R images tied to a more specific CRAN snapshot. But I recognize that introduces its own burden, and I think R patches come around frequently enough (several times a year) that we'll be OK sticking to them.

The other alternative to all of this is to stick to combine an R-latest image with `renv` or similar -- that would give us absolute control over the versions of all packages.

infotroph:
> Now that R 4.0.4 is out, the packages associated with 4.0.3 should now be frozen.

@ashiklom That's actually why I'm worried -- I thought 4.0.2 should have been frozen when 4.0.3 came out, so seeing hiccups there around the time of the 4.0.4 release makes me think I'm either misunderstanding the system (Very possible! I haven't been following it that closely) or that 'frozen' might not be as correlated with stability as we might hope.

 If the PEcAn-on-CRAN project happens, that transition might be a good moment to rethink how we handle dependencies and whether we want to handle them package-by-package or once for all of PEcAn. 

@robkooper am I correct that depends:4.0.4 won't exist until after this PR is merged? If so, I'm OK to merge this.

ashiklom:
> That's actually why I'm worried -- I thought 4.0.2 should have been frozen when 4.0.3 came out, so seeing hiccups there around the time of the 4.0.4 release makes me think I'm either misunderstanding the system (Very possible! I haven't been following it that closely) or that 'frozen' might not be as correlated with stability as we might hope.

I think you're right that it definitely _should_ have worked. I played around with it (see issue below) and it definitely is something on the side of either the Rocker images or the RStudio Package Manager.

I opened an issue here: https://github.com/rocker-org/rocker-versioned2/issues/127

infotroph:
@ashiklom Does this mean further modifications are needed to this PR, or are we waiting for upstream changes in RSPM?

ashiklom:
No, it's good to go! I'll merge now, so we can have the images built in time to merge other PRs next week.

PR #2781, opened 2021-02-28
[build] Detect changes at any depth in package (fixes #2776)

### LINKED ISSUE #2776: Make does not pick up changes in tertiary subdirectories
### Bug Description

Make notices changes (and therefore marks the package as needing to be rebuilt) in any files that live in top-level package directories (`R/`, `man/`, `inst/`, etc) but not to files nested more deeply than that (e.g. `tests/testthat/`, `inst/examples/`).

All  deeply-nested files *are* installed correctly once the package is marked as needing rebuild by some other method, so a workaround is to update the timestamp of another file (e.g. `touch models/mymodel/DESCRIPTION`) before running `make`.

### Steps to reproduce

(This might be system-dependant! The following is true for me in bash 3.2.57 and zsh 5.3, both on OS X 10.13.6.)

Example reported by @bcow: Make notices updates to files that live directly in `inst/`, but not to files in a subdirectory of it.

```
$> touch models/jules/inst/template.job # noticed
$> make install
./scripts/time.sh "depends  models/jules" Rscript -e "options(Ncpus = 1)" -e "deps <- if (grepl('(base/utils|modules/benchmark)', ' models/jules')) { c('Depends', 'Imports', 'LinkingTo') } else { TRUE }" -e "devtools::install_deps('models/jules', dependencies = deps, upgrade=FALSE)"
=========== START depends models/jules ===========
[...snip...]
=========== END   install models/jules ===========
echo `date` > .install/models/jules
$> touch models/jules/inst/template_nml_4.2/ancillaries.nml # not noticed
$> make install
make: Nothing to be done for `install'.
$>
```

### More context

The root of the problem is that our make rules rely on shell globs (`**`, `*`) to build the prerequisite list:

```
.install/%: $$(wildcard %/**/*) $$(wildcard %/*) .doc/% | $$(@D)
```

At least on my and @bcow's systems, `**` does not recurse into subsubdirectories. So when `%` is set to (e.g.) `models/foo/`, `$$(wildcard %/**/*) ` expands to "every file in a directory that's an immediate child of `models/foo/`," `$$(wildcard %/*)` expands to "Every file in `models/foo/`," and nothing expands to "every file that's in a directory that's a grandchild or deeper of `models/foo/`." 

It's possible that a [newer version of bash](https://stackoverflow.com/a/28199633) would behave as intended without any change.

###  Proposed fix

Quick and dirty: Extend the existing glob matches to three levels by adding `$$(wildcard %/**/**/*)` to each rule, and hope nobody ever wants Make to notice changes in `models/foo/inst/examples/bar/baz/2012/fluffy/`

More robust: implement a fully recursive wildcard match in Make. Here's one I've used in a different project:

```
recursive_wildcard = $(foreach d, $(wildcard $1*), \
	$(call recursive_wildcard, $d/, $2)$(filter $(subst *, %, $2), $d))
```

Sample usage from that project: `SOURCES := $(call recursive_wildcard, $(SOURCE_DIR), *.cpp)`

It's *possible* we could use this as a drop-in replacement for the shell globs, perhaps something like:

```
.install/%: $$(call recursive_wildcard, %, *) .doc/% | $$(@D)
````

But I haven't tested this. Two key points to check when trying it:
 
* Whether custom function calls work like that inside a secondary expansion
* Whether it works with a match-anything pattern (`*`)-- I've only ever used it with the pattern `*.cpp`.

infotroph:
## Description

Implemented an arbitrary-depth recursive listing macro rather than rely on system-dependent shell glob behavior.

Also: Went back to using `devtools::test` now that `stop_on_error` is supported again.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


robkooper (approved):

ashiklom (approved):
Impressive Makefile wizardry!

PR #2786, opened 2021-03-14
change calls from RCurl to curl package

moki1202:


moki1202:
@infotroph tried my best to update convert.input file.although I am not at all sure that form_file would be an ideal function to replace POSTFORM function.

base/utils/R/convert.input.R
```diff
@@ -611,8 +611,8 @@ convert.input <-
     curloptions <- c(curloptions, followlocation = TRUE)
     
     # check if we can do conversion
-    out.html <- RCurl::getURL(paste0("http://dap-dev.ncsa.illinois.edu:8184/inputs/",
-                              browndog$inputtype), .opts = curloptions)
+    out.html <- curl::curl_download(paste0("http://dap-dev.ncsa.illinois.edu:8184/inputs/",
+                              browndog$inputtype), curloptions)
```
infotroph, line 615: I have not tested this yet, but it looks correct 👍

base/utils/R/convert.input.R
```diff
@@ -636,7 +636,7 @@ convert.input <-
     }
     
     # post zipped file to Brown Dog
-    html <- RCurl::postForm(url, fileData = RCurl::fileUpload(zipfile), .opts = curloptions)
+    html <- curl::form_file(url, type = "___")
```
infotroph, line 639: It looks like the recommended pattern for form posting is to build up the multipart request inside the handle, then pass the whole handle as a fetch. See [this example](https://cran.r-project.org/web/packages/curl/vignettes/intro.html#posting_forms) in the curl vignette.

infotroph (changes requested):
I've only looked at `convert_input.R` so far, but posting this to save my comments and will come back to it as soon as I have time. Thanks for your patience.

base/utils/R/convert.input.R
moki1202: @infotroph thanks for the review ,I'll start working on the other files .

base/utils/R/download.url.R
```diff
@@ -23,16 +23,16 @@
 download.url <- function(url, file, timeout = 600, .opts = list(), retry404 = TRUE) {
   dir.create(basename(file), recursive = TRUE)
   count <- 0
-  while (!RCurl::url.exists(url, .opts = .opts) && count < timeout) {
+  while (!curl::curl_fetch_memory(url, handle = curl::new_handle(nobody = 1)) && count < timeout) {
```
infotroph, line 26: I see two subtleties here:
* When the url doesn't exist, `curl_fetch memory` will still return a value that evaluates to TRUE, so we need to explicitly check the status code (e.g. `(curl_fetch_memory(...)$status_code %/% 200) == 1`)
* More trickily, when the requested hostname can't be resolved `curl_fetch_memory` will throw an error whereas `RCurl::url.exists` just returns FALSE. I'm not sure whether we need to handle this case or not but will ask around.

base/utils/R/download.url.R
```diff
@@ -23,16 +23,16 @@
...
   }
   if (count >= timeout) {
     return(NA)
   }
-  f <- RCurl::CFILE(file, mode = "wb")
-  RCurl::curlPerform(url = url, writedata = f@ref, .opts = .opts)
-  RCurl::close(f)
+  f <- open(file, mode = "wb")
+  curl::curl_download(url, file, handle = curl::new_handle(.list = .opts))
+  close(f)
```
infotroph, line 35:
```suggestion
  curl::curl_download(url, file, handle = curl::new_handle(.list = .opts))
```

`curl_download` handles the opening and closing for us

modules/data.atmosphere/R/met.process.R
```diff
@@ -531,7 +531,7 @@ browndog.met <- function(browndog, source, site, start_date, end_date, model, di
   userpass <- paste(browndog$username, browndog$password, sep = ":")
   curloptions <- list(userpwd = userpass, httpauth = 1L, followlocation = TRUE)
   result <- RCurl::postForm(paste0(browndog$url, formatname, "/"),
```
infotroph, line 533: I'm guessing this is intentionally incomplete for now and will use the same pattern you develop in `convert.input`, but here's a placeholder to remind us to finish it :)

infotroph:
Input requested: This PR will alter `PEcAn.utils::download.url` by replacing the call to `RCurl::url.exists` with a header request performed via `curl::curl_fetch_memory`, but these behave differently in cases where the host isn't resolvable: `url.exists` returns FALSE, curl throws an error.  

paging @kooper as architect and @serbinsh as user with an ongoing troublesome retry-the-URL use case: Does `download.url` need to handle temporary DNS failures, or is it strictly intended to wait for an otherwise-accessible host who is taking a while to prepare a particular file?

base/utils/R/download.url.R
moki1202: @infotroph should we be using httr:HEAD then?

infotroph: Looks like httr::HEAD has the same behavior both with respect to status codes and with respect to erroring on unresolvable hostnames, so we don't need to switch because of those. However HEAD definitely shows the intention much clearly than the `nobody = 1` incantation does, so it'd be worth using at least in packages that already import httr.

moki1202:
`out<-RCurl::curlPerform(url = "https://SDMDataAccess.nrcs.usda.gov/Tabular/SDMTabularService.asmx",
                          httpheader = headerFields,  postfields = body,
                          writefunction = reader$update`
@infotroph cant really figure out the transformation for this part. can you once explain these arguments (httpheader, writefunction, postfields)?

moki1202:
@infotroph I was wondering if we could talk a bit about the project you're mentoring in detail.

mdietze:
Similar to comments on other PRs, was curious what's left to do on this one @moki1202 @infotroph

infotroph:
@mdietze @moki1202 

1. I tested this locally and discovered I gave some bad advice in the comments above, so I'm fixing in a local branch and will push ASAP.

2. Branch `moki1202/patch-8` has been deleted and getting it back will be a hassle, so I'll open a replacement PR and close this one once it's open.

moki1202:
> @mdietze @moki1202
> 
> 1. I tested this locally and discovered I gave some bad advice in the comments above, so I'm fixing in a local branch and will push ASAP.
> 2. Branch `moki1202/patch-8` has been deleted and getting it back will be a hassle, so I'll open a replacement PR and close this one once it's open.

@infotroph Apologies from my side. Don't know why I deleted this branch. Might have been early days. Let me know how I can help to complete this.

infotroph:
Continuing work on this in PR #2944; closing this one

infotroph (commented):
Apparently LAST YEAR  I left a draft review unfinished, including some comments that would have been useful in later conversation. Posting now so they're visible to everyone else, with my apologies.

base/utils/R/convert.input.R
```diff
@@ -636,7 +636,11 @@ convert.input <-
     }
     
     # post zipped file to Brown Dog
-    html <- RCurl::postForm(url, fileData = RCurl::fileUpload(zipfile), .opts = curloptions)
+    h <- new_handle()
+    handle_setform(h,
+    fileData = curl::form_file(curl::curl_upload(zipfile, url))
```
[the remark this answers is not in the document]
infotroph: (I was wrong about this... see #2944)

PR #2794, opened 2021-03-28
add "/document all_packages" command

infotroph:
## Description

Extends the `/document` PR command by adding an optional `all_packages` flag that causes Roxygen to be invoked on *all* PEcAn packages rather than only the ones touched by the current PR.

In most cases this would be silly because it takes longer to achieve the same effect as a plain `document`; it was written for a single use case: To allow bumping the Roxygen version purely in CI.

To use:

* Submit a PR that changes the pinned Roxygen version in `docker/depends/Dockerfile` (Today from 7.0.2 -> 7.1.1)
* Add a commit that temporarily forces the Makefile to unconditionally reinstall Roxygen, by deleting `if(!requireNamespace('roxygen2', quietly = TRUE))` from the `install/roxygen2` rule. This step is needed because the depends image will still have the previously-pinned Roxygen until after the change is merged.
* comment `/document all_packages` on the PR
* Review the resulting changes -- hopefully most of these will just be updates to the RoxygenNote in every DESCRIPTION.
* Merge
* Immediately submit a new PR that reverts the Makefile change

Is this silly? Possibly. But it means we can bump the Roxygen version without anyone on the team needing to match their build environment exactly to what's used during CI.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/styler-actions.yml
+++ b/.github/workflows/styler-actions.yml
@@ -53,12 +53,20 @@ jobs:
         run: Rscript scripts/generate_dependencies.R
       - id: file_changes
         uses: trilom/file-changes-action@v1.2.3
+      - id: doc_all
+        name: check whether documenting all packages or only changed ones
+        if: startsWith(github.event.comment.body, '/document all_packages')
+        shell: bash
+        run: |
+          DESCS=$(find * -name DESCRIPTION | tr '\n' ' ')
+          echo "::set-output name=descriptions::${DESCS}"
       - name : make
         shell: bash
         env:
           FILES: ${{ join(fromJSON(steps.file_changes.outputs.files_modified), ' ') }}
+          EXTRA_FILES: ${{ steps.doc_all.outputs.descriptions }}
         run: |
-          echo ${FILES} \
+          echo ${FILES} ${EXTRA_FILES} \
           | tr ' ' '\n' \
           | grep -e base -e models -e modules \
           | cut -d / -f 1-2 \


robkooper:
This code here will create the PR https://github.com/clowder-framework/extractors-core/blob/master/.github/workflows/pyclowder.yml#L51

infotroph:
@robkooper I now remember why I dropped the new-PR idea before: It seemed confusing when working with cross-fork PRs.

The current workflow: 
1. From any branch in any fork, user opens a PR  that will merge into `PecanProject:develop`. I'll refer to the source branch below as `username:branchname` noting that `username` might happen to be `PecanProject` but usually isn't.
2. Review happens at PecanProject/pecan/pull/[number]
3. If any styling or documentation is needed, anyone can trigger it by invoking the appropriate slash command.
4. Once invoked, the `commands` GitHub action runs and pushes all changes back into the PR branch.
5. As soon as the action exits, changes are visible in the PR and available for everyone to review and to merge the whole PR if it's ready.


If we switch the `commands` action to create new PRs, I think the workflow will need to become:

1. Same as above
2. Same as above
3. Same as above
3. When triggered, the `commands` Github Action runs, makes the changes requested, pushes them to a new branch in the PecanProject repo named something like `GHA-update-[number]`, and opens a PR that will merge `PecanProject:GHA-update-[number]` into `username:branchname`.
4. Here's where the workflows diverge: No changes will be visible in PecanProject/pecan/pull/[number] until the user merges username/pecan/pull/[number2], so at this point the Actions bot should probably post in the review comments for PecanProject/pecan/pull/[number] saying something like 'I've proposed some documentation updates! `@username`, Please go to `https://github.com/username/pecan/pull/[number2]` to review and merge them."
5. User reviews and merges the bot PR into their own branch. PRs into user forks that update PRs into upstream seem to be a frequent source of confusion (including my own!), so we and/or the bot will likely need to coach people through this. I believe that since the PR happens in `username:pecan` rather than `PecanProject:pecan`, the user must do this themselves and we or the bot can't merge it for them. Correct?
6. If the user becomes unresponsive at this stage with the bot PR still open, I *think* the only way to resolve it is to close their original PR and create a new one in a fork that we control and then to pull `PecanProject:GHA-update-[number]` into that, or else to merge the original PR without the bot changes and then apply the bot changes as a separate commit.
7. After the bot PR is merged, and only then, the changes are visible in PecanProject/pecan/pull/[number] and available for review / CI / merging by people other than the originating user, as in step 5 of the push-to-existing-PR workflow.
8. If other branches have updated in the meantime (e.g. someone clicked "update branch" in PecanProject/pecan/pull/[number]) before username/pecan/pull/[number2] was merged), the order of updates needed might be confusing. I haven't thought this step through fully.

I'm hoping I've overcomplicated this and that some of the steps above won't happen or will be invisible once implemented properly, but it still seems like a lot of friction and a lot of steps that need input from the PR owner rather than from anyone who's free to do them. Thoughts?

robkooper:
So the other option maybe is to add a new command, called /build that will trigger a rebuild? That way we can do /document to make sure the documentation is correct, once that worked, we can do a /build to rebuild the system.

I started to look at this (and some other options) #2796

mdietze (approved):

infotroph:
I like @robkooper suggestion for a /build command — matking this as draft until #2796 is complete in case that changes how this should work

robkooper:
Ok, as it turns out, the testing of the code depends on the workflows in the develop branch. so to be able to test the `/build` command it is required to be in the develop branch (which it now is).

You can force a rebuild using /build, so you can do /document in comment1, check to see the comment worked, then do a /build to rebuild it.

infotroph:
On further consideration I'm now leaning toward thinking this is a bad idea. Bumping Roxygen versions in a local (possibly Docker) environment is not that hard to do.

Anyone object if I close this?

dlebauer:
RE https://github.com/PecanProject/pecan/pull/2794#issuecomment-898722740

I won't object, but will put forth for consideration that when I say something like 'this not that hard to do', 'should be trivial' or similar, I am _usually_ wrong.

With respect to

> ... this is a bad idea.

why is it a bad idea? 
- Is it redundant with the new /build from #2796 or
- Is there a downside beyond adding the 9 additional lines of code in this PR?
- other?

when you say 

>  Bumping Roxygen versions in a local (possibly Docker) environment is not that hard to do.

- `/document all_packages` is "not that hard to do"
- is the alternative: either [`make document`](https://pecanproject.github.io/pecan-documentation/develop/updatebety.html#pecan-make) ... (as discussed a few weeks ago [on Slack](https://pecanproject.slack.com/archives/C01KZDVDXSQ/p1627573420000100?thread_ts=1627357006.024900&cid=C01KZDVDXSQ)
   1. install docker (or R + dependencies + make ...)
   2. docker pull or compose or build or whatever see 3
   3. remember / find the correct docker command(s) 
   4. run the command 
   5. git add, commit, push, create PR
   6. auto-generated files in PR review ...


image

infotroph:
@dlebauer These are good things to consider! The point that "...isn't that hard..." is usually wrong is especially well-taken.

Basically I think it's a bad idea because it's adding arcane layers *that will only be used for this one task*, whereas bumping Roxygen from a local environment uses the same arcane layers we have to interact with for other PEcAn development.

Reviewing the workflow I proposed above, the "temporary change, merge, revert" part in particular felt like it was adding rather than removing complications. I think I may now see a way to simplify that (by moving the installation of Roxygen out of the Dockerfile and into pecan.depends.R), but the fix may add complications of its own. I'll try it and report back.

PR #2797, opened 2021-05-18
Edited the dbfile.input.check function if/else input statement

helge22a:
Bug Fix

## Description

The dbfile.input.check function was not able to find existing met data for NOAA sites. Traced the issue to the if/else statement that creates the input object, when you include the enddate argument the input object returns wit 0 observations. So I substituted in a new if/else statement that does not use the enddate argument to create the input object.
## Motivation and Context

With this fix I can now successfully run NOAA and NEON sites through the workflow for the EFI Challenge.
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


mdietze (approved):

infotroph (approved):
...Well, I think the style update made reviewing harder rather than easier, but it should pay off in the next round of editing. Apparently this file really needed a style cleanup! 

Thank you for the patch. The change looks good to me.

PR #2810, opened 2021-06-24
heteroskedastic laplace likelihood fix

istfer:
## Description

Bug fix in calculation. 

## Motivation and Context

Looks like I was undoing the inflation the old way and the calculation did not properly give the log-likelihood of h. laplace overall. 

I think we will change this logic some more (instead of inflating variance we will weigh the likelihood) but at least this is a bug that needs to be fixed. I thought I would make a PR already in case others are using these functions.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.batch/R/pda.define.llik.R
+++ b/modules/assim.batch/R/pda.define.llik.R
@@ -98,8 +98,8 @@ pda.calc.error <-function(settings, con, model_out, run.id, inputs, bias.terms){
         # weigh down log-likelihood calculation with neff
         # if we had one beta value (no heteroscadasticity), we could've multiply n_eff*beta
         # now need to multiply every term with n_eff/n 
-        SS_p <- - (inputs[[k]]$n_eff/inputs[[k]]$n) * log(beta_p) - resid[[1]][pos]/beta_p
-        SS_n <- - (inputs[[k]]$n_eff/inputs[[k]]$n) * log(beta_n) - resid[[1]][!pos]/beta_n
+        SS_p <- - log(2*beta_p) - resid[[1]][pos]/beta_p
+        SS_n <- - log(2*beta_n) - resid[[1]][!pos]/beta_n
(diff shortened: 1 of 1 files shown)


mdietze:
If you are removing the n_eff here, where is it being accounted for?

Also, why do you propose to shift from inflating the variance based on autocorrelation (as computationally-efficient approximation to avoid a full joint covariance matrix) to a "weight", which can be incredibly ad hoc and distort the probabilistic meaning of the Likelihood?

istfer:
> If you are removing the n_eff here, where is it being accounted for?

n_eff is accounted for, upstream in the function while calculating betas, this line was basically undoing it

> Also, why do you propose to shift from inflating the variance based on autocorrelation (as computationally-efficient approximation to avoid a full joint covariance matrix) to a "weight", which can be incredibly ad hoc and distort the probabilistic meaning of the Likelihood?

I thought this is what we discussed earlier in the context of HPDA, although, maybe it was about the posterior predictive intervals, as under the inflated data model there was a huge underdispersion issue. But this PR does not make any changes regarding that.

mdietze (approved):

modules/assim.batch/R/pda.define.llik.R
```diff
@@ -98,8 +98,8 @@ pda.calc.error <-function(settings, con, model_out, run.id, inputs, bias.terms){
         # weigh down log-likelihood calculation with neff
         # if we had one beta value (no heteroscadasticity), we could've multiply n_eff*beta
         # now need to multiply every term with n_eff/n
```
infotroph, line 100: @istfer Are these comment lines still in sync with the code after this change?

istfer: oops, actually no L90-91 is enough and accurate, these could be deleted, I can make a quick PR for that

PR #2815, opened 2021-06-30
Fix remaining check warnings from PEcAn.logger (fixes #2814)

### LINKED ISSUE #2814: make /base/logger (PEcAn.logger ) package CRAN ready
The primary aim of this issue is to make our package `PEcAn.logger`, CRAN ready. To achieve this, we would be following strict [_guidelines and policies_ ](https://cran.r-project.org/web/packages/policies.html)set by the CRAN team. 

Following are the necessary checkboxes that need to be ticked before we think about releasing the package-

- [x] R CMD check should not return any warnings, notes and errors. One exception: Since the package has never been on CRAN before, "checking CRAN incoming feasibility" will always return a NOTE stating the maintainer name and the package license.
- [x] An additional [NEWS.md](https://r-pkgs.org/release.html?q=NEWS#news) file should be added to each package. It should list all the API changes in each release.
- [x]  It should meet every requirement mentioned by CRAN in _[this document](https://cran.r-project.org/web/packages/policies.html)_
- [x] Any package about to be released to CRAN must not depend on any other NON-CRAN package!
- [x] Check that the package maintainer is correctly assigned and their email is up to date.
- [ ] Review descriptive text in DESCRIPTION and make sure it clearly describes the purpose of this package (not just PEcAn as a whole).
- [x] Delete tests/Rcheck_reference.log
- [ ] Bump version and date!
- [ ] no Roxygen example section for un-exported functions.
- [ ] exported functions should have return value documentation.

moki1202:
this pull request is directed towards getting the `logger` package ready to be released to CRAN! I will be constantly updating this pull request as we move forward until the task is finished!


moki1202 (commented):
fixes NOTE - `* checking DESCRIPTION meta-information ... NOTE
Authors@R field gives no person with name and roles.
Authors@R field gives no person with maintainer role, valid email
address and non-empty name.
Author field differs from that derived from Authors@R
  Author:    ‘Rob Kooper, Alexey Shiklomanov’
  Authors@R: ‘’

Maintainer field differs from that derived from Authors@R
  Maintainer: ‘Alexey Shiklomanov <ashiklom@bu.edu>’
  Authors@R:  ‘’
`

base/logger/DESCRIPTION
```diff
@@ -2,11 +2,10 @@ Package: PEcAn.logger
 Title: Logger functions for PEcAn
 Version: 1.7.1
 Date: 2019-09-05
-Authors@R: c(person("Rob","Kooper"),
-             person("Alexey", "Shiklomanov"))
-Author: Rob Kooper, Alexey Shiklomanov
-Maintainer: Alexey Shiklomanov <ashiklom@bu.edu>
+Authors@R: c(person("Rob", "Kooper", role = c("aut", "cre"), email = "kooper@illinois.edu"),
+             person("Alexey", "Shiklomanov", role = c("aut"), email = "ashiklom@bu.edu"))
```
infotroph, line 6: @moki1202 please add yourself here too!

base/logger/DESCRIPTION
```diff
@@ -2,11 +2,10 @@ Package: PEcAn.logger
...
 Description: Special logger functions for tracking execution status and the environment.
+imports: utils
```
infotroph, line 8:
```suggestion
Imports: utils
```

base/logger/R/logifnot.R
```diff
@@ -1,14 +1,15 @@
...
-#' Conditions can be vectorized, or can return non-logical values.The 
-#' underlying function automatically applies `isTRUE(all(.))` to the 
+#' Conditions can be vectorized, or can return non-logical values.The
+#' underlying function automatically applies `isTRUE(all(.))` to the
 #' conditions.
 #'
 #' @param msg Logger message to write, as a single character string.
-#' @param ... Conditions to evaluate 
+#' @param ... Conditions to evaluate
+#' @param ... other arguments passed on to severeifnot
```
infotroph, line 12: This line shouldn't be needed and I wasn't thinking clearly when I suggested it! 🤦 
 I think the actual change needed is to not have Roxygen document `check_conditions` at all -- see my other suggestions below.
 
```suggestion
```

base/logger/R/logifnot.R
```diff
@@ -30,6 +31,7 @@ severeifnot <- function(msg, ...) {
 }
 
 #' @rdname severeifnot
+#’ @param ... other arguments passed on to errorifnot
```
infotroph, line 34:
```suggestion
```

base/logger/R/logifnot.R
```diff
@@ -41,6 +43,7 @@ errorifnot <- function(msg, ...) {
 }
 
 #' @rdname severeifnot
+#’ @param ... other arguments passed on to warnifnot
```
infotroph, line 46:
```suggestion
```

base/logger/R/logifnot.R
```diff
@@ -52,6 +55,7 @@ warnifnot <- function(msg, ...) {
 }
 
 #' @rdname severeifnot
+#’ @param ... other arguments passed on to infoifnot
```
infotroph, line 58:
```suggestion
```

base/logger/R/logifnot.R
```diff
@@ -63,6 +67,7 @@ infoifnot <- function(msg, ...) {
 }
 
 #' @rdname severeifnot
+#’ @param ... other arguments passed on to debugifnot
```
infotroph, line 70:
```suggestion
```

base/logger/R/logifnot.R
```diff
@@ -74,6 +79,7 @@ debugifnot <- function(msg, ...) {
 }
 
 #' Check a list of conditions
+#’ @param ... other arguments passed on to check_conditions
```
infotroph, line 82: Here's the much simpler fix for the issue I was trying to fix with all the extra `@param ...` lines: This and `is_definitely_true` below are both very simple helper functions that aren't exported and aren't used anywhere outside this file, so they don't need Roxygen documentation at all! Let's just convert these from Roxygen comments (`#'`) to plain old R comments (`#`).

After making this change, you'll need to re-run Roxygen so make it delete the Rd files for these -- easiest will be to use `/document` and let the GitHub robot do it for you.

```suggestion
# Check a list of conditions
```

infotroph (changes requested):

base/logger/R/logifnot.R
moki1202: @infotroph understood, but what if these functions **were** used by any outside functions. What changes could fix this warning?

infotroph: Then we'd either need to add parameter descriptions here, or tell Roxygen to use the descriptions from somewhere else using one of [`@inheritParams`](https://roxygen2.r-lib.org/articles/rd.html#inheriting-documentation-from-other-topics), [`@describeIn`](https://roxygen2.r-lib.org/articles/rd.html#describein), or [`@rdname`](https://roxygen2.r-lib.org/articles/rd.html#rdname).

moki1202:
@infotroph same goes for the similar warning in logger.getLevelNumber function. it's also not used anywhere outside logger.R.

infotroph:
> @infotroph same goes for the similar warning in loggergetLevelNumber function? it's also not used anywhere outside logger.R.

Yep, that works! In all three of these cases it also would have worked to leave them as Roxygen comments but add a `@noRd` tag; which approach is "better" is personal preference and probably depends whether your text editor displays Roxygen comments more or less nicely than plain comments ;)

base/logger/R/logger.R
```diff
@@ -140,11 +140,12 @@ logger.severe <- function(msg, ..., wrap = TRUE) {
 ##' }
 logger.message <- function(level, msg, ..., wrap = TRUE) {
   if (logger.getLevelNumber(level) >= .utils.logger$level) {
-    dump.frames(dumpto = "dump.log")
-    calls <- names(dump.log)
+    call_dump <- NULL # to avoid "no visible binding" note from R check
+    utils::dump.frames(dumpto = "call_dump")
+    calls <- names(call_dump)
```
infotroph, line 145: This doesn't work and I'd have known that if I'd run the package tests before submitting like I was supposed to! 🙈 

(apparently `dump.frames` binds both a local AND a global variable, so creating a local one produces weird masking effects! So since we only care here about the names and not the frame contents, instead I'm generating the names in the same way `dump.frames` does internally)

@robkooper please check this -- do you see any problems that I and testthat didn't think of?

base/logger/NEWS.md
```diff
@@ -0,0 +1,7 @@
+# PEcAn.logger 1.7.1.9000
+
```
infotroph, line 2: (Recording the change in behavior produced by my `dump.frames`->`sys.calls` change below -- only accept this if accepting the other change too!)

```suggestion
## Fixed

* Logger calls no longer create a stray `dump.log` object in the global environment
```

base/logger/R/logger.R
moki1202: @infotroph what's the sys.calls( ) here ?

infotroph: @moki1202 It's a base R call that tells you the whole stack of functions currently being executed, i.e. it's how you find out whether you're calling `foo()` from inside `bar(baz(quux(foo())))` or just `quux(foo()))`. You can see `help("sys.calls")` for more details.

infotroph (approved):

infotroph (changes requested):

infotroph (approved):

infotroph:
This PR is ready to merge. HUGE thanks to @moki1202 for giving us our first PEcAn package that checks fully clean: 0 ERRORs, 0 WARNINGs, 0 NOTEs*, and all historically-ignored messages in `Rcheck_reference.log` resolved and deleted. More like this, please!

*Technically when you check with `--as-cran` there's still one NOTE saying who the maintainer is and what license the package has, but this is normal and can only be fixed by... submitting it to CRAN, which is the imminent next step for this package!

PR #2817, opened 2021-07-01
Handle exact.dates = FALSE in dbfile.input.check

juliusvira:
## Description
I added an else block under for the if (exact.dates). I kept the end_date logic added in 692b65aa588d2352bb3dfb2f88a235e32295bc1e but brought back the missing code from the earlier version d93f1b395279119cd0849c0bc34c121f7bf2edda.

## Motivation and Context

When trying to get CF met data processed I found the workflow to crash in dbfile.input.check, because no query is run whenever the exact.dates argument is false. It seems that the code to handle this case had gone missing in a recent commit. Assuming that this was accidental, I restored the missing code from an earlier version. 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/base/db/R/dbfiles.R
+++ b/base/db/R/dbfiles.R
@@ -285,6 +285,15 @@ dbfile.input.check <- function(siteid, startdate = NULL, enddate = NULL, mimetyp
         con = con
       )
     }
+  } else { # not exact dates
+    inputs <- db.query(
+        query = paste0(
+          "SELECT * FROM inputs WHERE site_id=", siteid,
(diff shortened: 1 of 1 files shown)


istfer (approved):
@helge22a please chime in if we are missing something but in `if(exact.dates){}else{}` the `else` case seems to be deleted by mistake

mdietze (approved):
Apologies for not catching that earlier when reviewing @helge22a's that introduced this change PR

PR #2821, opened 2021-07-06
Fix remaining check warnings from PEcAn.photosynthesis (fixes #2819)

### LINKED ISSUE #2819: make /modules/photosynthesis (PEcAn.photosynthesis ) package CRAN ready
The primary aim of this issue is to make our package `PEcAn.photosynthesis`, CRAN ready. To achieve this, we would be following strict [guidelines and policies](https://cran.r-project.org/web/packages/policies.html) set by the CRAN team.

Following are the necessary checkboxes that need to be ticked before we think about releasing the package-

- [ ]  R CMD check should not return any warnings, notes and errors. One exception: Since the package has never been on CRAN before, "checking CRAN incoming feasibility" will always return a NOTE stating the maintainer name and the package license.

- [ ] An additional [NEWS.md](https://r-pkgs.org/release.html?q=NEWS#news) file should be added to each package. it should list all the API changes in each release.
- [ ] It should meet every requirement mentioned by CRAN in [this document](https://cran.r-project.org/web/packages/policies.html).
- [ ] Any package about to be released to CRAN must not depend on any other NON-CRAN package!
- [ ] Check that package maintainer is correctly assigned and their email is up to date.
- [ ] Review descriptive text in DESCRIPTION and make sure it clearly describes the purpose of this package (not just PEcAn as a whole).
- [ ] Delete tests/Rcheck_reference.log
- [ ] Bump version and date!

moki1202:
this pull request is directed towards getting the `photosynthesis` package ready to be released to CRAN! I will be constantly updating this pull request as we move forward until the task is finished!


modules/photosynthesis/R/plots.R
```diff
@@ -1,9 +1,8 @@
-##' @title shaded confidence interval
-##' @name ciEnvelope
-##' @author Mike Dietze
-##' @export
+## @title shaded confidence interval
+## @name ciEnvelope
+## @author Mike Dietze
```
infotroph, line 3: For cases like this where there's already a reasonably complete Roxygen block, I suggest keeping them as Roxygen comments and replacing `@export` with `@noRd` instead of converting the whole block to plain comments.

This is a reversal from what I suggested in PEcAn.logger, but here's my reasoning: It's a smaller and easier change, and it makes it explicit that the function is intentionally not exported or documented, whereas if converting to a plain comment someone might later see `# @param foo` and think "Oh, obviously that was supposed to be a Roxygen comment, let me fix that..."
```suggestion
##' @title shaded confidence interval
##' @name ciEnvelope
##' @author Mike Dietze
##' @noRd
```

moki1202: would '/document' still create a .Rd file for ciEnvelope?

infotroph: Not with the `@noRd` tag present. If we were to remove the `@export` without adding `@noRd`, it would still create an Rd file -- which is sometimes desirable for cases where you want to document functions while keeping them internal, though not what we want in this case.

modules/photosynthesis/R/plots.R
```diff
@@ -1,21 +1,26 @@
...
-  return(polygon(cbind(c(x, rev(x), x[1]), c(ylo, rev(yhi), ylo[1])), 
+  return(graphics::polygon(cbind(c(x, rev(x), x[1]), c(ylo, rev(yhi), ylo[1])), 
                  col = col, border = NA, ...))
 } # ciEnvelope
 
 ##' @name plot_photo
 ##' @title plot_photo
 ##' @author Mike Dietze
+##' @param data input data
+##' @param out directory where output will be stored
```
infotroph, line 14: The parameter is misleadingly named, so I didn't catch this before: In this function `out` is the model output to be fitted, not an output directory.
```suggestion
##' @param out fitted model output from \code{\link{fitA}}
```

modules/photosynthesis/DESCRIPTION
```diff
@@ -18,7 +20,14 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
...
-    coda (>= 0.18)
+    coda (>= 0.18),
+    utils,
+    graphics,
+    stats
+Suggests:
+    knitr,
+    rmarkdown,
+    markdown
 SystemRequirements: JAGS2.2.0
```
infotroph, line 31: @mdietze Should the JAGS requirement be updated to `>=`?  I'm currently testing with JAGS 4.3.0 and it all seems  to be working as advertised.

mdietze: Yeah, that's fine

infotroph, line 31:
```suggestion
SystemRequirements: JAGS (>=2.2.0)
```

infotroph:
Currently failing build because the vignette is too large. This will need to be fixed before it can go to CRAN, but we may want to consider merging this and tackling that separately.

infotroph:
Per @mdietze in Slack discussion about reducing vignette size: "I think you could safely drop the gelman.plots from the vignette"

infotroph:
@Aariq At the moment compilation time is a secondary problem and the main issue, which precompilation won't fix, is that the finished vignette PDF is by itself larger than CRAN's 5 MB limit for a whole package! To resolve this we need to do some combination of (a) reducing the number of images and/or (b) making each image use fewer bytes.

Aariq:
Are we sure this is the images and not this: https://github.com/PecanProject/pecan/pull/2836#issuecomment-1182523692 ?

Aariq:
Looks like the branch this PR was coming from doesn't exist anymore.  Should this PR be closed?

infotroph:
This PR has a lot of good changes in it, so I'd like to keep it open until someone (me) has time to push a replacement and work out the minimal changes needed to merge.

Aariq:
That sounds good. I was just trying to take a look at it and couldn't figure out how to with my usual avoid-commandline-git workflow which said the branch didn't exist anymore.

moki1202:
Please accept my apologies here! I don't know what I was thinking when I deleted my pecan fork 😞. @infotroph please let me know how I can recover the work done in this branch.

infotroph:
I've opened #3015 to be the mergeable continuation of this work and am closing this PR, but please refer back to the discussion here when planning any further work in #3015.

infotroph:
> am closing this PR
...well, apparently not 🤥. But NOW I'm closing it!

infotroph:
> am closing this PR

...well, apparently not 🤥. But NOW I'm closing it!

PR #2822, opened 2021-07-07
Pass port explicitly to open_database()

istfer:
## Description

When a different port is used for bety the web UI could not connect to DB, passing the port explicitly solves it I think.

## Motivation and Context

On our server we use a different port for Bety and I can't use web UI because pages don't respond after open_database() fails. This change seems to fix it for me.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible
## Types of changes

## Checklist:

### DIFF
--- a/web/common.php
+++ b/web/common.php
@@ -76,6 +76,7 @@ function toXML($string) {
 # ----------------------------------------------------------------------
 function open_database() {
   global $db_bety_hostname;
+  global $db_bety_port;
   global $db_bety_username;
   global $db_bety_password;
   global $db_bety_database;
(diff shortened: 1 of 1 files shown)


istfer:
With this change I was able to move to the second page but also maybe @robkooper can help me with another (related?) web UI issue. Now on `02-modelsite` page, I cannot select the `Host` on UI, it's grayed out, unresponsive and shows "Any Host" only.  `Models` the same, "All models"

Why could it be disabled? `02-modelsite.php` should be able to set `$hostname`, and on the DB there are models associated with the server
Screenshot from 2021-07-07 13-20-31

mdietze:
On the second error, @istfer could you share the $hostlist section of your config.php, either here or on slack? Most likely you just have an error in that list.

mdietze:
On the first part, with the port, this PR needs an update to the config.example.php so that folks know to set that variable. Ideally, it would also be backwards compatible (set the port to a default value if the global value doesn't exist or is null)

istfer:
Thanks @mdietze 

for $hostlist: I didn't modify anything for hostlist in my config.php as `$fqdn` is [there](https://github.com/PecanProject/pecan/blob/develop/web/config.example.php#L73). Sorry, it's been a while setting these up but I don't recall modifying hostlist for localhost, is this in the documentation somewhere?

for port variable in config.example.php: I think there already is [a line for setting up the port](https://github.com/PecanProject/pecan/blob/develop/web/config.example.php#L6) on `config.example.php`. Did you mean something else?

mdietze (approved):
Got it. Didn't realize the port variable was already being used elsewhere.

I checked my own config.php and it was the same as yours for localhost. I know I've debugged issues like that in the past, but don't recall what the solution was.

PR #2825, opened 2021-07-19
Ic func

helge22a:
## Description

Specific changes to match_pft function to include the model object so that pfts can be matched with the appropriate model in BETY, and cohort2pool now has the dbh_name object which is intended to make it easier to identify the column in the dataset that has the DBH measurements (even if that column name does not include "DBH"). Also created a load_veg_manual script that allows a user to go through the process of creating IC data in pecan standards without having it registered in the BETY database. 

THIS IS A DRAFT
## Motivation and Context

Trying to add NEON IC data for 5 NEON sites in order to set up an iterative forecast. 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


models/ed/R/veg2model.ED2.R
```diff
@@ -1,10 +1,18 @@
 #' Writes ED specific IC files
-#'
+#' @title veg2model.ED2
+#' @name veg2model.ED2
```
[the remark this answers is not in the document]
infotroph: But while I'm here, *thank you* for all your work adding missing `@param` entries in this and previous PRs. It has cleaned up a *lot* of package check warnings and made the documentation much easier to understand. I'll take an occasional extra title tag over missing parameter descriptions any day!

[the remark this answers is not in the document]
helge22a: Thanks for the helpful feedback on the use of @title and @name. I am still learning how this all fits together. In my limited experience I have found that having the @name is helpful for NAMESPACE calls for the function. When I was going through ic_process line-by-line I was running into some errors where R would tell me it could not find function even with a namespace call. I found that when this would happen it was usually because the function did not have the @name param or because the @name param did not match the function name.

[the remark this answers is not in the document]
infotroph: > I was running into some errors where R would tell me it could not find function even with a namespace call

High-level point first: If the missing function is present in your package's NAMESPACE file you can be confident that Roxygen has done its job correctly and the issue lies elsewhere. In particular, my first stop for debugging function-not-found issues is to check the namespace *in the installed copy of the package* using `ls(getNamespace("pkgname_here"))`, and if the function is missing there I reinstall the package and try again. Even after a decade of building R packages, the overwhelming majority of my "why can't R find the function I know I added" problems turn out to be "I did not in fact reinstall the package after adding the function" problems.

Lower-level, I don't see how `@name` would affect function availability, because Roxygen only uses it change the name given in the documentation pages and not the name it exports in the namespace. Consider the following code file:

```
#' Rd name implicitly taken from function name
#' @export
foo1 <- function(){}

#' Rd name given explicitly but matches function name
#' @name foo2
#' @export
foo2 <- function(){}

#' Rd name does not match function name
#' @name bar
#' @export
foo3 <- function(){}
```

Once processed with Roxygen, this generates files documentation files `man/foo1.Rd`, `man/foo2.Rd`, and `man/bar.Rd` (which last will contain `\name{bar}` but also `\alias{foo3}` so that people typing `?foo3` can find the right help page), but the NAMESPACE file will be:

```
export(foo1)
export(foo2)
export(foo3)
```

Note that there's no `export(bar)` and that `pkgname::bar()` will still return an error.

mdietze:
@helge22a per the Slack discussion with @infotroph, you need to put back the `Roxygen: list(markdown = TRUE)` line into DESCRIPTION files where they got deleted. He thinks this is the source of your character escaping issue

modules/data.land/R/put_veg_module.R
```diff
@@ -1,4 +1,6 @@
-##' Match species to PFTs + veg2model module
+##' @description Match species to PFTs + veg2model module
+##' @name put_veg_module
+##' @title put_veg_module
```
infotroph, line 3:
```suggestion
##' Match species to PFTs + veg2model module
```

scripts/EFI_icprocess.R
```diff
@@ -0,0 +1,6 @@
+
+# EFI_icprocess -----------------------------------------------------------
+
+#required inputs
+
+
```
infotroph, line 6: This file does nothing -- does it need to be added now or should it wait until it contains code?

infotroph (changes requested):

infotroph:
And one more question that there's no where for me to put it in an inline comment: Was the rename from `scripts/dataprep_10_sites.csv` → to `scripts/EFScripts` really the rename you intended? Seems odd to switch from `.csv` to no file extension.

infotroph:
@helge22a I left a bunch of suggestions as inline comments. If you've used GitHub's suggestions before you know the drill; if not you can read about them [here](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) -- the short version is that suggestions are nice because they let you accept or reject the proposed changes right from the browser window instead of needing to go pull up the file I was commenting on in your local repo, retype my changes, commit, push, etc etc.

scripts/EFI_icprocess.R
helge22a: Yes that is a mistake I thought I removed that file. Should definitely wait to add it until it contains code.

infotroph (approved):
I see a few more suggestions still open, but most of them are spelling nits that don't need to hold up the PR. I do predict the checks will fail until you apply the `\dontrun` change or one like it, though.

Note that I only reviewed for build issues and changes introduced by merge conflicts, so I'm counting on another review from someone who understands the IC workflow better since I haven't evaluated the correctness of the code that's, y'know, the important part  of the PR :)

modules/data.land/R/extract_NEON_veg.R
```diff
@@ -0,0 +1,56 @@
...
+#' 
+#' @importFrom rlang .data
+#' 
+#' @examples start_date = as.Date("2020-01-01") 
+#' end_date = as.Date("2021-09-01")
+
+extract_NEON_veg <- function(lon, lat, start_date, end_date, ...){
+
+#Set store_dir to point to dietzelab/neon_store
+store_dir = "/projectnb/dietzelab/neon_data"
```
mdietze, line 24: This was fine for development, but you can't hardcode a directory within a package that could be used by anyone on any computer.

modules/data.land/R/extract_NEON_veg.R
```diff
@@ -0,0 +1,56 @@
...
+#' @export
+#' 
+#' @importFrom rlang .data
+#' 
+#' @examples start_date = as.Date("2020-01-01") 
+#' end_date = as.Date("2021-09-01")
+
+extract_NEON_veg <- function(lon, lat, start_date, end_date, ...){
+
+#Set store_dir to point to dietzelab/neon_store
```
mdietze, line 23: code within a function should be indented

modules/data.land/R/extract_veg.R
```diff
@@ -1,27 +1,53 @@
...
+##' @param format_name DEFAULT=NULL
+##' @param machine_host passed from ic_process
+##' @param dbparms taken from settings object, passed from ic_process
+##' @param outfolder passed from ic_process, location where to store files
+##' @param overwrite DEFAULT = FALSE
+##' @param ... Additional parameters
+##'
+##' @return results object to be passed back to get.veg.module
 ##' @export
 ##' @author Istem Fer
```
mdietze, line 19: general comment: @helge22a you should add yourself to the author list of functions you created or made major modifications to

modules/data.land/R/extract_veg.R
```diff
@@ -1,27 +1,53 @@
...
 extract_veg <- function(new_site, start_date, end_date, 
                      source, gridres, format_name = NULL, 
                      machine_host, dbparms, outfolder, overwrite = FALSE, ...){
-  
-  
+  #code taken from https://stackoverflow.com/questions/14183766/match-fun-provide-error-with-functions-defined-inside-functions
+   fget <- function(name, env = parent.frame()) {
```
mdietze, line 24: could you add a second comment explaining what this function does, not just where you got it?

modules/data.land/R/extract_veg.R
```diff
@@ -1,27 +1,53 @@
...
  # Extract veg info
   
  fcnx <- paste0("extract_", source) # e.g. extract_FIA
- 
+ #Need a better way to check if the function exists
  if (!exists(fcnx)) {
    PEcAn.logger::logger.severe(paste(fcnx, "does not exist."))
  }else{
-   fcn <- match.fun(fcnx)
+   fcn <- fget(fcnx) #Error cannot find the function
```
mdietze, line 43: Don't understand the comment here

modules/data.land/R/extract_veg.R
```diff
@@ -30,13 +56,19 @@ extract_veg <- function(new_site, start_date, end_date,
...
+      code_col    <- "species_USDA_symbol"
+      format_name <- "usda"
+      obs[obs$species_USDA_symbol != "2PLANT", ] #removes the rows with 2PLANT, this is a NEON specific code that means they could not identify the species 
+   }
  }
 
  
  # match code to species ID
- spp.info <- match_species_id(input_codes = obs[[code_col]], format_name = format_name)
+ spp.info <- PEcAn.data.land::match_species_id(input_codes = obs[[code_col]], format_name = format_name)
```
mdietze, line 71: you shouldn't need to reference functions names by namespace within the package itself.

modules/data.land/R/extract_veg.R
```diff
@@ -48,7 +80,7 @@ extract_veg <- function(new_site, start_date, end_date,
  # Write vegettion data as rds, return results to convert.input
  
  # need check for overwrite
- sppfilename <- write_veg(outfolder, start_date, end_date, veg_info = veg_info, source)
+ sppfilename <- PEcAn.data.land::write_veg(outfolder, start_date, veg_info = veg_info, source)
```
mdietze, line 83: ditto

modules/data.land/R/ic_process.R
```diff
@@ -92,22 +97,26 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
     settings$run$inputs[['poolinitcond']]$path <- newfile
     
     return(settings)
+  }else if (input$source == "NEON_veg"){
+    #For debugging purposes I am hard coding in the start and end dates, will revisit and adjust once extract_NEON_veg is working within ic_process
+    start_date = as.Date("2020-01-01")
+    end_date = as.Date("2021-09-01")
```
mdietze, line 103: Since you are no longer debugging, you need to fix this

modules/data.land/R/write_ic.R
```diff
@@ -16,41 +32,51 @@ write_ic <- function(in.path, in.name, start_date, end_date,
...
-  
-  # NOTE : match_pft may return NAs for unmatched dead trees
-  pft.info <- PEcAn.data.land::match_pft(obs$bety_species_id, pfts)
-  
-  ### merge with other stuff
+
+  # # NOTE : match_pft may return NAs for unmatched dead trees
+  pft.info <- PEcAn.data.land::match_pft(bety_species_id = obs$bety_species_id, pfts = pfts, model = model, con = NULL)
+
+  # ### merge with other stuff
```
mdietze, line 39: not sure why you modified the comments to be # ###

modules/data.land/R/write_ic.R
```diff
@@ -16,41 +32,51 @@ write_ic <- function(in.path, in.name, start_date, end_date,
...
   }
-  
-  out <- fcn(outfolder, veg_info, start_date, new_site, source)
-  
+# Cohort2Pool -------------------------------------------------------------
+  # read in registration xml for pool specific information
+  register.xml <- system.file(paste0("register.", model$type, ".xml"), package = paste0("PEcAn.", model$type))
+  register     <- PEcAn.data.atmosphere::read.register(register.xml, con = NULL)
+  #check if register,model.xml includes "POOL"
+  if (register$poolinitcond == "POOL") {
```
mdietze, line 60: see previous comment about register file tag name suggestion (initcond)

modules/data.land/R/write_ic.R
```diff
@@ -16,41 +32,51 @@ write_ic <- function(in.path, in.name, start_date, end_date,
...
   }else{
     fcn <- match.fun(fcnx)
   }
-  
-  out <- fcn(outfolder, veg_info, start_date, new_site, source)
-  
+# Cohort2Pool -------------------------------------------------------------
+  # read in registration xml for pool specific information
+  register.xml <- system.file(paste0("register.", model$type, ".xml"), package = paste0("PEcAn.", model$type))
+  register     <- PEcAn.data.atmosphere::read.register(register.xml, con = NULL)
```
mdietze, line 58: Given that IC is a new addition the register, I'd recommend adding some checks that the register file exists, and also add some checks (and a default) to see if IC tag exists. That way this code will run (though probably with a warning) for existing models (otherwise you need to go and update the registry for every existing model, which honestly you should probably do anyways)

mdietze (changes requested):

base/db/R/dbfiles.R
```diff
@@ -347,7 +347,7 @@ dbfile.input.check <- function(siteid, startdate = NULL, enddate = NULL, mimetyp
       print(inputs)
       dbfile <- dbfile.check(
         type = "Input",
-        container.id = inputs$id,
+        container.id = inputs$parent_id,
```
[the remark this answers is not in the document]
helge22a: So I made this change to address a bug I was having with the parent and db file ids passed to dbfile.input.check, since there are two ids sometimes the function would try to run the query with the dbfile id instead of the parent id and this would trigger an error. I think I also added a fix on the ic_process side of things once I realized hat was happening, so I am not sure if this is necessary anymore.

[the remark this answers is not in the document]
mdietze: I do remember an issue where it was using the dbfile id when it was supposed to be using the input id (or vice versa) but that should be a separate issue from whether you use the input id (pointer to the current input) or the input parent_id (pointer to the input the current input was derived from)

mdietze:
build is stuck. Closing PR and reopening

PR #2826, opened 2021-07-22
Last few tweaks for the logger package before it moves to CRAN

moki1202:


mdietze (approved):

base/logger/DESCRIPTION
```diff
@@ -2,14 +2,18 @@ Package: PEcAn.logger
...
              person("Shashank", "Singh", role = c("aut"),
                 email = "shashanksingh819@gmail.com"),
-             person("Chris", "Black", role = c("ctb")))
-Description: Special logger functions for tracking execution status and the environment.
+             person("Chris", "Black", role = c("ctb")),
+             person("University of Illinois, NCSA", role = c("cph")))
+Description: This package adds convenience functions for logging outputs. 
+    This is loosely based on the log4j package. This will enable the user to 
+    set what level messages are printed, and if these messages are written to a file, 
+    console or both.
```
infotroph, line 16:
```suggestion
Description: Convenience functions for logging outputs from 'PEcAn',
   the Predictive Ecosystem Analyzer (LeBauer et al. 2017;
   doi:10.1890/12-0137.1). Enables the user to set what level of messages are
   printed, as well as whether these messages are written to the console, 
   a file, or both. It also allows control over whether severe errors should
   stop execution of the PEcAn workflow; this allows strictness when deugging
   and lenience when running large batches of simulations that should not be
   terminated by errors in individual models. PEcAn.logger is loosely based on
   the 'log4j' package.
```

infotroph (changes requested):

infotroph:
@moki1202 I believe the last step needed here is to bump the version (to 1.8.0, in both DESCRIPTION and NEWS.md) and date. Then we can build, final-test, and submit!

base/logger/DESCRIPTION
```diff
@@ -1,15 +1,24 @@
 Package: PEcAn.logger
 Title: Logger Functions for PEcAn
-Version: 1.7.1
-Date: 2019-09-05
-Authors@R: c(person("Rob", "Kooper", role = c("aut", "cre"), 
+Version: 1.8.0
+Date: 2019-27-07
```
infotroph, line 4:
```suggestion
Date: 2021-07-27
```

infotroph (approved):

robkooper (approved):

robkooper:
@moki1202 if you feel it is ready, mark it ready for review so it can be merged. And we can push it to cran.

moki1202:
I believe we are ready to role!!

PR #2833, opened 2021-08-04
Soil moisture

CFranc22:
## Description

This is a function that pulls VSWC values from the EFI server.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

## Checklist:

### DIFF
--- a/modules/data.land/inst/Pulling_VSWC.R
+++ b/modules/data.land/inst/Pulling_VSWC.R
@@ -1,0 +1,60 @@
+# Script pulling soil moisture data from target files to create a netCDF file at the end
+## vswc = mass_fraction_of_unfrozen_water_in_soil_moisture
+
+# loading necessary packages
+library(readr)
+library(dplyr)
+library(ncdf4)
+
+# set function inputs
+homedir <- "/projectnb/dietzelab/cfranci1"
+site_ID <- "BART"
+
+call_vswc <- function(homedir, site_ID) {
+
+# reading in terrestrial targets data at daily level
+download.file("https://data.ecoforecast.org/targets/terrestrial/terrestrial_daily-targets.csv.gz",
+              "terrestrial_daily-targets.csv.gz")
+
+terr_daily <- read_csv(file.path(homedir, "terrestrial_daily-targets.csv.gz"),
+  col_types = cols(
+  time = col_datetime(format = ""),
+  siteID = col_character(),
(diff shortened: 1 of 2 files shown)


CFranc22:
I'm Christina, a student from the Harvard Forest REU program. I worked with Mike Dietze to incorporate site-specific soil moisture data into the PEcAn workflow.

mdietze (changes requested):
A good start, but a few things need fixing:
* the same script seems to have been committed twice, once to the R folder and once to the scripts folder (inst). The code seems to mostly be an R function, so I think we R folder version is the one we want to retain
* Given that, the R file needs to JUST contain the function, not the homedir and site_ID that come before that. But these things might be useful for an example.
* PEcAn convention is to call the directory you're writing files to `outdir` not `homedir`
* functions need ROxygen documentation -- @helge22a can show you how RStudio ca give you a template for that
* PEcAn also asks that functions be called by namespace rather than using `library` to load whole packages. So, for example, you'd drop `library(readr)` and instead replace `read_csv` with `readr::read_csv`
* I don't know if it's coming in a separate PR from @helge22a, but there should be parallel changes to `write.configs.SIPNET` where these files are read in and used
* The latter is a high-enough level new feature that the overall PEcAn documentation should be updated to indicate that this new feature exists

helge22a:
@mdietze I will make the changes you've requested today. The additions to the write.configs.SIPNET are apart of the larger pull request I am doing that will include all of the IC functions.

helge22a:
@mdietze How should I go about making changes to the overall documentation?

mdietze:
Documentation is in the `book_source` folder and its file organization mirrors the order of the compiled documentation

modules/data.land/inst/Pulling_VSWC.R
```diff
@@ -0,0 +1,60 @@
+# Script pulling soil moisture data from target files to create a netCDF file at the end
+## vswc = mass_fraction_of_unfrozen_water_in_soil_moisture
+
+# loading necessary packages
+library(readr)
+library(dplyr)
+library(ncdf4)
+
+# set function inputs
+homedir <- "/projectnb/dietzelab/cfranci1"
```
serbinsh, line 10: But we will want to pull these hard-coded paths right?  so they work on any machine

PR #2836, opened 2021-08-09
Fix remaining check warnings and undeclared dependencies from PEcAn.allometry

moki1202:
Fix remaining check warnings and undeclared dependencies from PEcAn.allometry


modules/allometry/inst/AllomDriver.R
```diff
@@ -13,8 +13,6 @@ if(!is.null(settings$Rlib)){ .libPaths(settings$Rlib)}
 #library(PECAn)
 
 ## libraries & database connection
-library(mvtnorm)
-library(MCMCpack)
```
infotroph, line 17: Let's leave these in place -- library calls OK inside `inst/` scripts

moki1202:
@infotroph are there any other changes required here?

modules/allometry/DESCRIPTION
```diff
@@ -1,29 +1,35 @@
 Package: PEcAn.allometry
 Type: Package
-Title: PEcAn allometry functions
+Title: PEcAn Allometry Functions
 Version: 1.7.2
 Date: 2021-10-04
-Authors@R: c(person("Mike", "Dietze", role = c("aut", "cre"),
-                     email = "dietze@bu.edu"),
+Authors@R: c(person("Mike","Dietze", role = c("aut","cre"), email = "dietze@bu.edu"),
```
infotroph, line 6:
```suggestion
Authors@R: c(person("Mike","Dietze", role = c("aut","cre"),
                                       email = "dietze@bu.edu"),
```

modules/allometry/DESCRIPTION
```diff
@@ -1,29 +1,36 @@
 Package: PEcAn.allometry
 Type: Package
-Title: PEcAn allometry functions
+Title: PEcAn Allometry Functions
 Version: 1.7.2
 Date: 2021-10-04
-Authors@R: c(person("Mike", "Dietze", role = c("aut", "cre"),
+Authors@R: c(person("Mike","Dietze", role = c("aut","cre"),
```
infotroph, line 6:
```suggestion
Authors@R: c(person("Mike", "Dietze", role = c("aut", "cre"),
```

mdietze:
@infotroph @moki1202 this PR looks like it's in pretty good shape. I'm happy to pull it in as soon as the GH Action tests pass. Are there other things y'all wanted to integrate into this PR before it is merged? Currently failing on pecan.depends.R

infotroph:
@mdietze @moki1202 I agree this looks good -- if there was anything else that needed to happen, I don't remember it now!

moki1202:
@mdietze @infotroph Let me just check if I can remember if this required some changes or not.

moki1202:
@infotroph Since I can't access the VM, could you please run the `R CMD` checks on allometry and let me know the check results? Right now I see a lot of `no visible binding errors` in the file, so just wanted to check.

infotroph (approved):
I'm not sure why CI isn't running, but I'm approving now in case that unsticks something

modules/allometry/DESCRIPTION
```diff
@@ -1,29 +1,38 @@
 Package: PEcAn.allometry
 Type: Package
-Title: PEcAn allometry functions
+Title: PEcAn Allometry Functions
 Version: 1.7.2
 Date: 2021-10-04
 Authors@R: c(person("Mike", "Dietze", role = c("aut", "cre"),
                      email = "dietze@bu.edu"),
+             person("Shashank", "Singh", role = c("ctb"),
```
infotroph, line 8:
```suggestion
             person("Shashank", "Singh", role = c("ctb")),
```

mdietze:
Please take a look at the failures in both test and check, which bring up two separate issues. On the file-size one, I wonder if you could force the docs to build using lower-res images, or even just force the graphics to be raster not vector. If not, I think I had earlier suggestions on slack for what plots could be cut or made optional (so that one could run them manually but they wouldn't knit into the vignette)

Aariq:
> Please take a look at the failures in both test and check, which bring up two separate issues. On the file-size one, I wonder if you could force the docs to build using lower-res images, or even just force the graphics to be raster not vector. If not, I think I had earlier suggestions on slack for what plots could be cut or made optional (so that one could run them manually but they wouldn't knit into the vignette)

It's not the images, it's the hundreds of warnings like this:
```
## Warning in u2 + 0.5 * crossprod(Y - X %*% beta): Recycling array of length 1 in vector-array arithmetic is deprecated.
##   Use c() or as.vector() instead.
```

If I turn off all the output except the plots the rendered vignette is 867.4kb

modules/allometry/R/allom.BayesFit.R
```diff
@@ -265,9 +263,9 @@ allom.BayesFit <- function(allom, nrep = 10000, form = "power", dmin = 0.1, dmax
         u2 <- u2 + 0.5 * crossprod(Y - X %*% beta)
```
mdietze, line 263:
```suggestion
        u2 <- u2 + as.numeric(0.5 * crossprod(Y - X %*% beta))
```

mdietze: @Aariq I think this will fix the warning you identified. Is this the only one or were there other similar warnings?

infotroph:
Thanks @Aariq for the diagnostics and @mdietze for the fix! Checks are now passing and this is ready to merge.

PR #2837, opened 2021-08-10
cleanup/fixes of API

robkooper:
- use NA instad of NULL in functions, openapi does not know what to do with NULL.
- don't pass in global_db_pool into function, this does not exist in openapi
- fix plumber calls, remove deprecated calls

@moki1202 this will fix problem (hopefully) with api endpoints.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible
- [ ] 
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -14,6 +14,8 @@ on:
 jobs:
   bookdown:
     runs-on: ubuntu-latest  
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
     container: 
       image: pecan/depends:R4.0.3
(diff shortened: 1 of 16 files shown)


mdietze (approved):

robkooper:
@mdietze @infotroph added a bonus fix for failing builds. Looks like the build in github_pat is no longer working. It will now use a trick I saw in devtools, where it sets [GITHUB_PAT from GITHUB_TOKEN](https://github.com/r-lib/devtools/blob/master/.github/workflows/pkgdown.yaml#L14).

infotroph (approved):

PR #2842, opened 2021-08-29
[GH Actions] remove special installation for processx

infotroph:
No longer needed because 3.5.2 is available from CRAN now.


infotroph:
...Well, OK, it's on CRAN but not yet in the Depends docker image, presumably because that uses date-locked snapshots from RSPM. 

@robkooper is there a way to prompt the image to update or should we hold this PR until the new processx shows up automatically (presumably when we switch to testing with a newer R release)?

robkooper:
we use the date limited to make sure we are building with a consistent version of packages, and not what CRAN thinks is the right set of packages.  So yes, I think we can remove it when we bump versions. Should we build using 4.0 and 4.1 instead?

infotroph:
@robkooper Yeah, that makes sense to me. How much trouble would it be to also test against R-devel? In an ideal world I'd like to add that, but I realize that's kinda the opposite of consistent versions and I don't have a sense how much trouble it adds to the Docker stack.

robkooper:
as long as we don't require that to pass, I think it is ok.

infotroph:
ping @robkooper -- I think this should be a quick review

robkooper (approved):

PR #2845, opened 2021-09-01
skip dependency check when on CI

infotroph:
Not needed because when on CI all known dependencies are listed in `docker/depends/pecan.depends.R` and we install them in bulk before invoking make

Question for @robkooper:  On checking, it looks like install-dependencies steps of the `test`, `check`, and `sipnet` jobs is 

```
Rscript scripts/generate_dependencies.R && Rscript docker/depends/pecan.depends.R
``` 

But in the `build` job it's just

```
Rscript scripts/generate_dependencies.R
```

Should that change, or does build not need depends?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,9 @@ $(ALL_PKGS_I) $(ALL_PKGS_C) $(ALL_PKGS_T) $(ALL_PKGS_D): | .install/devtools .in
 
 .SECONDEXPANSION:
 .doc/%: $$(call files_in_dir, %) | $$(@D)
+ifeq ($(CI),) # skipped on CI because we start the run by bulk-installing all deps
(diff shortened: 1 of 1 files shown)


dlebauer (approved):

robkooper:
so the only tricky part of this is a chicken and egg problem.

`Rscript scripts/generate_dependencies.R` would make sure that all dependencies listed in DESCRIPTION are in `pecan.depends.R` and then then install them. This would make sure that if you added a new depedency in the DESCRIPTION it would get installed and work during the compile/test phases

infotroph:
@robkooper I think what you're describing is already done in all the other steps (test, check, sipnet), just not in the build step. But addionally, it looks as if `devtools::install` may now install missing deps by  default (at least ones from known repos), so maybe that'll handle most cases by itself.

PR #2849, opened 2021-09-13
Very simple test cases for AllomAve

infotroph:
## Description

Some fairly token test cases for `PEcAn.allometry`: Verifies that raw output from `AllomAve` gets written to the expected working directory, and that `AllomUnitCoef` behaves sensibly when given either known or unknown units.

## Motivation and Context

Same motivation here as for #2848, so if you read that already you can skip to the code now.

"Once again, Chris, why are you spending your time writing tests for the simplest function in the package?" 

Well, because my real motivation is to prevent a future build problem: 

* The allometry and assim.sequential packages currently contain no unit tests.
* This is fine in our current CI builds because they currently use devtools 2.3.2, in which `devtools::test` simply reports that no tests were found and moves on to the next package.
* But devtools >= 2.4 stops with an error when no tests are found, so the CI builds will start failing for these packages the next time we bump the R version we test against.
* Therefore I'm setting out to make sure every package contains *some* tests, but I'm gonna start with the functions that are easiest to test 😉  I highly encourage follow-up PRs from package owners that focus on the functions actually central to each package.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


modules/allometry/tests/testthat/test_coefs.R
```diff
@@ -0,0 +1,22 @@
...
+	# unknown value of tp -> ignored
+	expect_equal(
+		AllomUnitCoef(x = "cm", tp = "invalid"),
+		1)
+
+	# length(tp) must equal length(x)
+	expect_error(
+		AllomUnitCoef(x = c("kg", "cm"), tp = "crc"),
+		"missing value")
+})
```
infotroph, line 22:
```suggestion
})

```

modules/allometry/tests/testthat/test_AllomAve.R
```diff
@@ -0,0 +1,36 @@
...
+	pfts <- list(FAGR = data.frame(spcd = 531, acronym = "FAGR"))
+	allom_stats <- AllomAve(
+		pfts,
+		components = 18,
+		ngibbs = 5,
+		nchain = 2)
+
+	expect_true(file.exists(file.path(outdir, "Allom.FAGR.18.Rdata")))
+	expect_true(file.exists(file.path(outdir, "Allom.FAGR.18.MCMC.pdf")))
+
```
infotroph, line 35:
```suggestion
```

mdietze (approved):

PR #2851, opened 2021-09-14
Fix remaining check warnings and undeclared dependencies from PEcAn.remote

moki1202:


moki1202:
I'll keep updating this PR with time.

infotroph:
@DongchenZ @araiho Can I get please your SDA-expert eyes on commit https://github.com/PecanProject/pecan/pull/2851/commits/ae3d578221d5c0bcc54984132f193fa240505363 in particular? It moves packages `plyr` and `reshape2` from Imports to Suggests of PEcAn.assim.sequential, to fix a check-time message about importing excessive numbers of packages. The reason it needed to happen in this PR is that we moved `start.model.runs` from PEcAn.remotes to PEcAn.workflow, therefore causing PEcAn.assim.sequential to need to import workflow, therefore triggering the check message by pushing the number of imported packages over 20.

I realize some of these changes are in places that are likely going to be refactored imminently in ongoing work on the SDA functions; if those refactorings remove the need to use plyr and/or reshape at all, then all these runtime checks can be happily deleted!

For reference/context on Imports vs. Suggests and why the runtime checks for availability are needed, see the [PEcAn book](https://pecanproject.github.io/pecan-documentation/develop/declaring-dependencies-depends-suggests-imports.html), especially the paragraph starting "Since the point of Suggests is that they are allowed to be missing..."

infotroph:
Clarifying my comment above: @araiho and @DongchenZ, you're very welcome to review the whole PR if you want to! But the changes to assim.sequential in commits https://github.com/PecanProject/pecan/pull/2851/commits/792378986713a9f6a50432dfb5f00a6006fb82df, https://github.com/PecanProject/pecan/pull/2851/commits/5a887341d9c3ef2ff6987efbca1242ad14eedc86, and https://github.com/PecanProject/pecan/pull/2851/commits/ae3d578221d5c0bcc54984132f193fa240505363 are the ones I want to be sure an SDA expert has seen and OK'd.

infotroph (approved):

PR #2855, opened 2021-09-18
QAQC: Convert tests for new.taylor to compare against stored snapshot

infotroph:
The existing test was writing an `Rplots.pdf` into the test directory but not performing any checks that it looked right. This version uses ~testthat's `expect_snapshot_file` function~ <ins>the `vdiffr` package</ins> to compare what's plotted against a stored version.

Note that the stored version is the plot output I got today -- it looked sensible to me, but whether it's "correct" could be debated.

And while I was in the package and had a test that would tell me if I broke it, I also refactored `new.taylor()` to avoid some check warnings 😄 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible
## Types of changes

## Checklist:


dlebauer:
I don't see where `expect_snapshot_file` is used, but the docs say this function is deprecated and should be replaced with `compare` https://testthat.r-lib.org/reference/expect_snapshot_file.html

`vidffr::expect_dopelganger` is also marked deprecated, but with no alternative suggested https://vdiffr.r-lib.org/reference/expect_doppelganger.html

infotroph:
@dlebauer Good catch -- I replaced `expect_snapshot_file` with `vdiffr::expect_doppelganger` (to keep snapshots made on OS X from failing on Linux and vise-versa) and neglected to update the PR description.

Re deprecation: My understanding is that both functions are still supported and recommended, but they each have one *argument* that's marked deprecated. That display confused me at first too.

Still to do before merging: This test seems to need the newest version of `vdiffr` to keep it from deleting the stored snapshots when tests fail (!) and the Docker images currently install an old version from Rstudio Package Manager. If you like the overall approach, I can probably work around this by installing `vdiffr` from github.

mdietze (approved):

PR #2883, opened 2021-11-03
update the CO2 data LPJ-GUESS uses

istfer:
## Description

A small PR adding new years to the CO2 file LPJ-GUESS uses. The new values come from [NOAA GML dataset](https://gml.noaa.gov/webdata/ccgg/trends/co2/co2_annmean_mlo.txt).

## Motivation and Context

This data load is probably not the best solution, but it's the existing solution and people would probably like to run simulations past 2012 now.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


models/lpjguess/R/write.config.LPJGUESS.R
```diff
@@ -96,16 +96,16 @@ write.config.LPJGUESS <- function(defaults, trait.values, settings, run.id) {
 #' @author Istem Fer
 write.insfile.LPJGUESS <- function(settings, trait.values, rundir, outdir, run.id) {
   
-  guessins  <- readLines(con = system.file("template.ins", package = "PEcAn.LPJGUESS"), n = -1)
-  paramsins <- readLines(con = system.file("pecan.ins", package = "PEcAn.LPJGUESS"), n = -1)
+  guessins  <- base::readLines(con = system.file("template.ins", package = "PEcAn.LPJGUESS"), n = -1)
+  paramsins <- base::readLines(con = system.file("pecan.ins", package = "PEcAn.LPJGUESS"), n = -1)
```
infotroph, line 100: Seems odd to need this -- it's usually safe / recommended to use base functions without namespace qualifiers. Is there a competing non-base version of readLines that this is avoiding?

istfer: at some point I thought we decided to use namespace even for the base functions, but looks like it was a misunderstanding on my end. Will just revert these

infotroph (approved):

PR #2886, opened 2021-11-03
Humidity issue, ERA5 met.process and Prep_OBS_SDA function

DongchenZ:
1. In the met2cf.ERA5.R file, I changed the way it calculates relative humidity such that it will give 0-1 value instead of percentage (the percentage value will give incorrect specific humidity value)

2. In the met2model.SIPNET.R file, I used the air pressure data in converting specific humidity to relative humidity such that the relative humidity will not exceed 1. (which gives a reasonable value for VPD and VPD soil)

3. I also updated several fixings for met.process function especially for ERA5 data.

4. I uploaded a function called Prep_OBS_SDA.R which serves for downloading and extracting AGB and LAI data and converting them into SDA preferred format.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -45,6 +45,7 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
   } else {
     jobsh <- readLines(con = system.file("template.job", package = "PEcAn.SIPNET"), n = -1)
   }
+  jobsh[1] <- paste(jobsh[1],"-l")
```
infotroph, line 48: Am I right that this makes the first line of the generated job.sh be `#!/bin/bash -l` (where `-l` indicates "login mode" and causes bash to read user profile files on startup) ? 

If so, I'm weakly inclined against this: The standard advice for shell scripting is to keep the script self-contained and to avoid chances for the user's personal configuration to affect script behavior. If you need to provide specific settings to be able to use sipnet on your machine, I think it's probably better to pass them in specifically -- perhaps in settings$model$prerun?

This is a tentative opinion, though -- I'm open to counterarguments.

DongchenZ: yes! should I just grab info from the setting object and create a if statement like that:
if(is geo){
jobsh[1] <- paste(jobsh[1],"-l")
}
such that, it won't break anything if it's not on geo.

DongchenZ: I haven't figured out how to pass this info to the pre-model run section.

infotroph: What specifically needs to be passed in?

mdietze: @DongchenZ what specifically needs to be set from your profile for the script to run? Is there a reason that can't be passed in using the `@HOST_SETUP@` tag in the template job https://github.com/PecanProject/pecan/blob/develop/models/sipnet/inst/template.job, which maps onto `settings$model$prerun`

modules/data.atmosphere/R/ERA5_met_process.R
```diff
@@ -0,0 +1,188 @@
...
+  }
+  
+  #loop over each site
+  #Record all IDs for inputs and dbfiles
+  if(Write){
+    Input_IDs <- list()
+  }
+  
+  #record paths of extracted .clim files if not writing into bety
+  if(!Write){
```
mdietze, line 75: should be an `else` not a second `if`

modules/data.atmosphere/R/ERA5_met_process.R
```diff
@@ -0,0 +1,188 @@
...
+    }else{
+      print(paste0("The output files for site ",as.character(site_info$site_id[i])," already exists jump to the next site"))
+      next
+    }
+    
+    #Write into inputs table for each site
+    if(Write){
+      #insert into inputs table
+      cmd <- paste0(
+        "INSERT INTO inputs ",
```
mdietze, line 96: Do the Input Insert AFTER the met has been downloaded, not before, in case the download fails

modules/data.atmosphere/R/ERA5_met_process.R
```diff
@@ -0,0 +1,188 @@
...
+    }
+    for (j in 1:length(ens_nc)) {
+      nc_path <- ens_nc[j]
+      
+      #find a proper in prefix for each ensemble member
+      ens_num <- strsplit(basename(nc_path),"_")[[1]][3]
+      in_prefix <- paste0("ERA5.", ens_num)
+      
+      #preparing for the met2model.SIPNET function
+      PEcAn.SIPNET::met2model.SIPNET(in.path = nc_path,
```
mdietze, line 132: Why is this function hard coded to only generate ERA5 met for just one model?

modules/data.atmosphere/R/ERA5_met_process.R
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }
+    
+    #record paths of clim files if not Write into Bety
+    if(!Write){
+      Clim_paths[i] <- list(in.path=list.files(path=site_outFolder, pattern = '*.clim', full.names = T))
+    }
+
+    #record input ID and dbfile IDs and extracted paths of clim files for each site
+    if(Write){
```
mdietze, line 155: This if(Write) and the previous if(!Write) should be joined by an `else`

modules/data.atmosphere/R/met.process.R
```diff
@@ -221,7 +221,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
       con,
       hostname = PEcAn.remote::fqdn(),
       exact.dates = TRUE,
-      pattern = met,
+#      pattern = met,
```
mdietze, line 224: not clear why this was commented out

modules/data.atmosphere/R/met.process.R
```diff
@@ -244,7 +244,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
           end_date <= end_date,
           format_id == formatid
         ) %>%
-        filter(grepl(met, name)) %>%
+        filter(grepl(met, "name")) %>%
```
mdietze, line 247: Could you explain this change too? Are we really looking for the literal string "name" in the met, not the contents of the variable `name`?

mdietze (changes requested):

docker.sh
```diff
@@ -158,7 +158,7 @@ for x in base web docs; do
 done
 
 # all files in subfolder
-for x in models executor data thredds monitor rstudio-nginx check; do
+for x in models executor data thredds monitor rstudio-nginx; do
```
mdietze, line 161: @robkooper there's a bunch of changes to Docker in this PR. Could you check to make sure these should be here and are not just the result of some sort of merge issue?

mdietze (approved):

PR #2889, opened 2021-11-09
Roxygen update step 1: Install on Depends image

infotroph:
## Description
Bumps Docker from Roxygen 7.0.2 to 7.1.2 (current release).  As soon as this is merged I will submit part 2, which updates all packages to match.

**This will break all CI jobs!** Sorry for the inconvenience; they should start working again as soon as part 2 is merged.
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/docker/depends/Dockerfile
+++ b/docker/depends/Dockerfile
@@ -40,7 +40,7 @@ RUN apt-get update \
 # ----------------------------------------------------------------------
 COPY pecan.depends.R /
 RUN  Rscript -e "install.packages(c('devtools'))" \
-  && Rscript -e "devtools::install_version('roxygen2', '7.0.2')" \
+  && Rscript -e "devtools::install_version('roxygen2', '7.1.2')" \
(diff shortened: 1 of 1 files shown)


docker/depends/Dockerfile
```diff
@@ -40,7 +40,7 @@ RUN apt-get update \
 # ----------------------------------------------------------------------
 COPY pecan.depends.R /
 RUN  Rscript -e "install.packages(c('devtools'))" \
-  && Rscript -e "devtools::install_version('roxygen2', '7.0.2')" \
```
robkooper, line 43: I think we need to update https://github.com/PecanProject/pecan/blob/develop/Makefile#L81 that is the check to make sure the right roxygen is installed.

robkooper (changes requested):

docker/depends/Dockerfile
infotroph: That's updated as part of #2890 -- if we change it here then the checks will fail

robkooper (approved):

PR #2892, opened 2021-11-12
for CI=true

robkooper:
Set CI=TRUE when building in docker to prevent running of credits for github.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of .zenodo.json
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2893, opened 2021-11-12
Less eager dependency check

infotroph:
## Description

Replaces the Makefile check for package dependencies, previously provided by a call to `devtools::install_deps` that always checked for updates, with a new check that checks locally first and exits if all dependencies are already satisfied, _even if newer versions of them are available_. This is a change in behavior, but I argue it's what we want: If your installed version of (say) `stringi` still works, running `make document` shouldn't waste your time looking for a new version of it.

Note that if any dependencies *are* missing for a given package, we still resolve them by calling `install_deps` with `upgrade = FALSE`, which will still _send queries_ to learn about updates to all dependencies (not just the ones that are missing) but will only  _install_ the missing ones. 

## Motivation and Context

The primary motivation is that during CI builds, running `install_deps` on every package was generating enough GitHub API queries to routinely use up GitHub's rate limit of 60 unauthenticated requests per hour. Providing a personal access token helped, but was a hassle to pass in to all the places that run Make. 

This PR takes the approach that it's fine if _a few_ packages find missing dependencies and install them during the build -- this allows easier testing of PRs that add new dependencies which aren't yet installed on the Docker images -- but that we needn't spam CRAN and GitHub repeatedly checking for new versions of packages that are already installed and working.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,11 +62,15 @@ drop_parents = $(filter-out $(patsubst %/,%,$(dir $1)), $1)
 # Generates a list of regular files at any depth inside its argument
 files_in_dir = $(call drop_parents, $(call recurse_dir, $1))
 
-# HACK: assigning to `deps` is an ugly workaround for circular dependencies in utils pkg.
+# HACK: NA vs TRUE switch on dependencies argument is an ugly workaround for
+# circular dependencies in utils pkg.
 # When these are fixed, can go back to simple `dependencies = TRUE`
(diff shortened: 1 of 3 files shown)


infotroph:
@robkooper If you like this approach, note that it makes #2892 not needed, because setting `CI` will no longer change Make's behavior.

robkooper (approved):

PR #2895, opened 2021-11-15
Report PEcAn package versions

infotroph:
## Description
Reports released and currently installed versions of all PEcAn packages, along with datasets for historical release versions and a gratuitously overengineered script for updating the datasets at release time.

Critique vigorously, please, and send me back to the drawing board if it's needed. I went ahead will a full version to show the concept, but I think there's room for debate about basic design decisions here: 
* Does this interface make sense?
* Is it storing the right information?
* Will it continue to scale acceptably as we add new releases and packages?
* Is PEcAn.all the right package for it to live in? Adding it there breaks the current conceptual explanation that "`PEcAn.all` is like `tidyverse`: It doesn't contain any code and doesn't export anything, it only makes sure the other packages are available." 
    * [Edit 2021-11-29: I had been under the impression `tidyverse` contained no code of its own, but apparently it too now exports a couple of version-reporter functions. So maybe the analogy holds up OK!]
* How should this be plumbed into the release process so that it stays up to date without undue manual work?

## Motivation and Context
With the move toward making PEcAn packages available on CRAN, it is no longer practical to assume that all package versions get incremented every PEcAn release, so we need a way to match a release of PEcAn to the versions of the packages it contains.

Note also that once this function is available, I think a natural next step will be to adopt the `x.y.z.9000` version convention for the development versions of all packages. That will help ensure you can always tell whether you're running e.g. the actually-released PEcAn.logger 1.8.0, or a development version of it that will eventually become 1.8.1.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [x] My name is in the list of .zenodo.json


base/all/data-raw/record_package_versions_for_release.R
```diff
@@ -0,0 +1,241 @@
...
+# START WITH A CLEAN GIT TREE. This script checks out tags to read their
+# versions. It complains before clobbering anything and restores state
+# after run, but no warranties etc
+#
+# No "save" => just write to console
+# ./base/all/data-raw/record_package_versions_for_release.R v1.2.3
+#
+# With "save" => add results to package data files
+# base/all/data/pecan_version_history.rda & base/all/data/pecan_releases.rda
+# ./base/all/data-raw/record_package_versions_for_release.R v4.5.6 v4.5.7 save
```
infotroph, line 15: When I wrote this I was picturing that we could use this to add new versions at release time, but that implies creating the tags before committing the updated data files! Probably need an option to report the versions at HEAD and write them to file with a specified tag name.

infotroph: Probably better yet:
* Change this script to report only date rather than datetime in `pecan_releases` table
* rerun and save result
* leave this script as a historical, overly-baroque artifact
* write a new script that only
    - takes the desired tag name, version, and release date as arguments
    - writes the tagname, version, and release date as a new row in `pecan_versions`
    - calls the guts of `scan_at_tag` without performing any Git operations (i.e. scans at HEAD)
    - writes the result to a new column of `pecan_version_history`

base/all/data-raw/record_versions.R
```diff
@@ -0,0 +1,124 @@
...
+    strict = FALSE
+  )
+}
+
+######## End function defs
+
+
+args <- commandArgs(trailingOnly = TRUE)
+
+if (length(args) < 3) {
```
infotroph, line 60:
```suggestion
if (length(args) < 3 || length(args) > 4) {
```

robkooper:
@infotroph why not save as csv file for the previous version. That way it is easy to look at this with other tools as well?

infotroph:
@robkooper I tried that at first and would be happy to switch back if others like that better, but for this application I liked rda for three-ish reasons:
* Biggest reason: Saving as rda keeps all the version numbers properly wrapped in class `package_version`, making it easy to do version comparisons without needing to remember to re-convert every time.
* Also a factor: R's lazy-loading rules for CSVs are frankly archaic: It needs all "CSV"s to be semicolon-delimited and always reads character data as factors (Or more precisely "files ending ‘.csv’ or ‘.CSV’ are read using `read.table(..., header = TRUE, sep = ";", as.is=FALSE)`", making it even more of a hassle to convert the package versions back. It would be possible to store both a normal comma-separated CSV and an R script that reads it back to a properly-formatted dataframe at package load time, but that felt to me like extra hassle to recreate an object that was easy to retain in Rda format.
* Minor consideration: The rda file is smaller than even a compressed CSV would be.

mdietze (approved):

PR #2904, opened 2022-01-18
Fixes no visible binding for global variable for PEcAn.benchmark package | Fixes #2758

### LINKED ISSUE #2758: Fix tidyverse "no visible binding for global variable" notes

Sarthakaga15:
@infotroph @ashiklom @moki1202 I tried fixing a few no visible binding errors; I would love your opinion on proceeding further. Please let me know if my changes are wrong so I can work on fixing it.


ashiklom (approved):
This looks great to me!

Minor request: Could you also remove the corresponding lines from the [Rcheck_reference.log](https://github.com/PecanProject/pecan/blob/develop/modules/benchmark/tests/Rcheck_reference.log#L68)? That help us catch this problem if it resurfaces in the future.

Thanks for your help!

Sarthakaga15:
@ashiklom Apologies . I forgot to convert this into the draft. I will be updating this PR further, fixing some more global binding errors ;)

infotroph:
@Sarthakaga15 To make the CI checks pass you'll eventually also need to make sure that `rlang::.data` is being imported into all the packages you use it in, via `@importFrom rlang .data`. Feel free to ping me if you need more details.

modules/benchmark/R/align_data_to_data_pft.R
```diff
@@ -50,6 +50,7 @@
 #'  custom_table = table)
 #' }
 #' @export
+#' @importFrom rlang .data
```
infotroph, line 53: I don't see anywhere in this function that uses .data, so unless I missed one I don't think you need this `importFrom`

infotroph (approved):

modules/benchmark/DESCRIPTION
```diff
@@ -35,6 +35,8 @@ Imports:
     SimilarityMeasures,
```
infotroph, line 35:
```suggestion
    rlang,
    SimilarityMeasures,
```

PR #2908, opened 2022-02-11
rough code for conducting flux validation with FLUXNET sites

serbinsh:
Making some interim code available to build validation scripts. hindsight, this was not the correct package to dump this code


infotroph:
@serbinsh How important is it that the images be added along with with the code? If my rough calculation is right (63 PNGs, roughly half weighing ~500 KB and half ~800 KB), merging this will add about 40 MB that need to be carried around forever in the Git history even if we delete the files later.

serbinsh:
@infotroph i removed most examples and just left a few.  Hopefully, all of this will get removed after we create the proper functions for conducting the model-data validation/comparisons!

@mdietze note I moved this from data.remote to assim.sequential

infotroph:
Whoever merges this (likely @mdietze or @robkooper): Please squash-merge to avoid keeping the removed sample files.

mdietze (approved):

mdietze:
@serbinsh could you check the CI error this PR is getting? seems related to file name lengths?

serbinsh:
@mdietze given this is just some temporary info that I expect we will remove later, do we need to worry about the filename lengths?

PR #2934, opened 2022-06-03
path issue fixes of Met files

DongchenZ:
## Description

path issue fixes of Met files
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -200,8 +200,8 @@ sda.enkf.multisite <- function(settings,
   ### Splitting/Cutting the mets to the start and the end  of SDA       ###
   ###-------------------------------------------------------------------###---- 
   #create a folder to store extracted met files
-  if(!file.exists(paste0(settings$outdir, "/Extracted_met"))){
-    dir.create(paste0(settings$outdir, "/Extracted_met"))
+  if(!file.exists(paste0(settings$outdir, "/Extracted_met/"))){
+    dir.create(paste0(settings$outdir, "/Extracted_met/"))
```
infotroph, line 204: This will work. You can also consider using `file.path` instead of `paste0` so that it will handle the path separators for you.

infotroph (approved):

modules/data.atmosphere/R/ERA5_met_process.R
```diff
@@ -66,7 +66,11 @@ ERA5_met_process <- function(settings, in.path, out.path, Write=FALSE){
   
   #setting up met2model function depending on model name from settings
   met2model_method <- paste0("PEcAn.", settings$model$type, "::met2model.", settings$model$type)
-
+  package_name <- strsplit(met2model_method, "::")[[1]][1]
+  function_name <- strsplit(met2model_method, "::")[[1]][2]
+  #load library otherwise the eval(call(package::function)) will not work
+  library(package_name, character.only = TRUE)
+
```
infotroph, line 73: This is the wrong approach for several reasons (`library()` calls should never occur in package code, and this particular library call will make PEcAn.atmosphere depend on every model package), but the good news is the right approach is much simpler than this, though admittedly strange-looking: construct the function lookup as an explicit call to `::`, then use the resulting function directly:

```
met2model_method <- do.call(
  "::", 
  list(paste0("PEcAn.", settings$model$type), paste0("met2model.", settings$model$type)))
```

Breaking this down, the two key pieces to know are that `::` can accept literal character strings (e.g. `"stats"::"var"(1:3)` => 1) but does not evaluate its arguments (that is, it won't expand variables: `x <- "stats"; x::"var"` => "there is no package called ‘x’"), and that `do.call` does evaluate its arguments by default and is often useful for exactly that.

Then instead of either `call(met2model_method, ...)` or `eval(cal(met2model_method, ...))` lines 115-119 below become simply `met2model_method(...)`.

infotroph (changes requested):

modules/data.atmosphere/R/ERA5_met_process.R
DongchenZ: got it! Thank you so much! I already fixed this!

mdietze (approved):

infotroph (approved):

Issue #2950, opened 2022-07-05
read.settings doesn't error when inputfile doesn't exist


Aariq:
### Bug Description

Looking at the documentation, this seems like maybe it was intended as a "feature", but it seems likely to cause headaches for some.  If I have a `pecan.xml` in my root directory, I can read it in with: `read.settings()`, `read.settings("pecan.xml")`, or `read.settings("blahblahblah.xml")`.  It seems to me like that last one should error!  If you had two settings files, say, `pecan.xml` and `test_sp.xml`, and you ran `read.settings("testsp.xml")` your typo would read in `pecan.xml` without an error alerting you that `testsp.xml` doesn't exist.

PR #2954, opened 2022-07-07
Remove Settings as Global Variable

Its-Maniaco:
This pull tries to fix the issue of 'Settings' being used as Global variable. This is done by creating a local variable 'Settings' and passing that as function argument as required (here in convert.input.R).


Aariq:
See https://github.com/PecanProject/pecan/issues/2950. I think it would be better practice to have a `file` argument that could have a default of `"pecan.xml"`, but it should error if the user tries to read in a .xml file that doesn't exist.

base/utils/R/convert.input.R
```diff
@@ -64,6 +64,7 @@
 ##' @export
 ##' @author Betsy Cowdery, Michael Dietze, Ankur Desai, Tony Gardella, Luke Dramko
 
+settings <- read.settings("pecan.xml")
```
Aariq, line 67:
```suggestion
settings <- read.settings(file = "pecan.xml")
```

Aariq:
I think this is the line that needs changing: 
https://github.com/PecanProject/pecan/blob/cd3e4c2dc1739b4165574dd1200f73886d80af9a/base/utils/R/convert.input.R#L100

`exists()` looks for an R object in the global environment.  Looks like it could probably be replaced with  `Rbinary <- settings$host$Rbinary` if `settings` is an argument to the function.

base/utils/R/convert.input.R
```diff
@@ -88,6 +89,7 @@ convert.input <-
            ensemble = FALSE,
            ensemble_name = NULL,
            dbparms=NULL,
+           settings,
```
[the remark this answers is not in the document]
mdietze: Also, if you look inside convert.input the only use of `settings is to get the Rbinary. 
https://github.com/PecanProject/pecan/blob/f7ee293df1d2ebd4cc7df9c019d1d43bf77a64ef/base/utils/R/convert.input.R#L100

Seems like a much better solution would be to make Rbinary an optional argument to convert.input, with the default value of "R". If it's optional you could concievably add it here at the end of the optional list.  Also, in either case, don't forget that if you're adding an arg to a function you need to add it in every place the function is used, not just where it's defined.

mdietze (changes requested):

base/utils/R/convert.input.R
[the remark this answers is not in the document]
Aariq: Good call.  The benefit of giving it an argument with a default is that the change is unlikely to break anything!

Its-Maniaco:
@Aariq @mdietze Based on your suggestions, this is what I came up with.

base/utils/R/convert.input.R
```diff
@@ -99,7 +98,7 @@ convert.input <-
                      mimetype, site.id, start_date, end_date))
   
   # TODO see issue #18
-  Rbinary <- ifelse(!exists("settings") || is.null(settings$host$Rbinary),"R",settings$host$Rbinary)
+  if(!is.null(settings$host$Rbinary) Rbinary <- settings$host$Rbinary)
```
infotroph, line 101: This line can be deleted entirely -- now that `Rbinary` is an argument to the function, it's the user's responsibility to set it, and this code can simply use the value they provide (or the default of "R" if they don't specify anything); if the user wants to set it from `settings$host$Rbinary` then they should do that in their `convert.input` call.

mdietze: I 100% agree with @infotroph. Unless this line is dropped, there's still an invisible dependency on a `settings` as a global variable.

That said, it is important to check the places that convert.input is called and to add settings$host$Rbinary as an argument iif it exists in the settings object. Note that it might not exist, so don't just pass settings$host$Rbinary by default as that will result in a whole lot Rbinary = NULL errors.

Its-Maniaco: Bleeding heart10052_square-01

modules/data.atmosphere/R/download.raw.met.module.R
```diff
@@ -91,7 +91,8 @@
       method = input_met$method,
       pattern = met,
       dbparms=dbparms,
-      ensemble = ensemble
+      ensemble = ensemble,
+      Rbinary = "R"
```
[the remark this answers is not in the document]
moki1202: also does the /document command no longer work?

modules/data.atmosphere/R/extract.nc.module.R
```diff
@@ -38,7 +38,9 @@ if (exists(paste0("extract.nc.", met))) fcn <- paste0("extract.nc.", met)
                             newsite = new.site$id, 
                             overwrite = overwrite,
                             exact.dates = FALSE, 
-                            ensemble = register$ensemble %>% as.numeric())
+                            ensemble = register$ensemble %>% as.numeric(),
+                            Rbinary = "R"
+                            )
```
infotroph, line 43:
```suggestion
                            Rbinary = "R")
```

modules/data.atmosphere/R/met2cf.module.R
```diff
@@ -35,7 +35,9 @@
                             write = TRUE, 
                             format.vars = format.vars, 
                             overwrite = overwrite,
-                            exact.dates = FALSE)
+                            exact.dates = FALSE,
+                            Rbinary = "R"
+                            )
```
infotroph, line 40:
```suggestion
                            Rbinary = "R")
```

modules/data.atmosphere/R/met2cf.module.R
```diff
@@ -49,7 +51,9 @@
                            start_date = start_date, end_date =  end_date, 
                            pkg = pkg, fcn = fcn, con = con, host = host, browndog = NULL,
                            write = TRUE, 
-                           overwrite = overwrite, exact.dates = FALSE)
+                           overwrite = overwrite, exact.dates = FALSE,
+                           Rbinary = "R"
+                           )
```
infotroph, line 56:
```suggestion
                           Rbinary = "R")
```

modules/data.atmosphere/R/met2cf.module.R
```diff
@@ -77,7 +81,9 @@
                           write = TRUE, 
                           lat = lat, lon = lon, 
                           format.vars = format.vars, 
-                          overwrite = overwrite, exact.dates = FALSE)
+                          overwrite = overwrite, exact.dates = FALSE,
+                          Rbinary = "R"
+                          )
```
infotroph, line 86:
```suggestion
                          Rbinary = "R")
```

modules/data.atmosphere/R/met2model.module.R
```diff
@@ -64,7 +64,9 @@
                               forecast = forecast,
                               ensemble = !is.null(register$ensemble) && as.logical(register$ensemble),
                               ensemble_name = ensemble_name,
-                              dbfile.id=ready.id$dbfile.id)
+                              dbfile.id=ready.id$dbfile.id,
+                              Rbinary = "R"
+                              )
```
infotroph, line 69:
```suggestion
                              Rbinary = "R")
```

modules/data.atmosphere/R/metgapfill.module.R
```diff
@@ -37,7 +37,9 @@
                             forecast = forecast,
                             pattern = met,
                             ensemble = !is.null(register$ensemble) && as.logical(register$ensemble),
-                            ensemble_name = ensemble_name)
+                            ensemble_name = ensemble_name,
+                            Rbinary = "R"
+                            )
```
infotroph, line 42:
```suggestion
                            Rbinary = "R")
```

modules/data.land/R/ens.veg.module.R
```diff
@@ -55,7 +55,9 @@ ens_veg_module <- function(getveg.id, dbparms,
                                          in.name = spp.file$file_name,
                                          n.ensemble = n.ensemble,
                                          machine_host = machine_host,
-                                         source = input_veg$source)
+                                         source = input_veg$source,
+                                         Rbinary = "R"
+                                         )
```
infotroph, line 60:
```suggestion
                                         Rbinary = "R")
```

modules/data.land/R/get_veg_module.R
```diff
@@ -53,7 +53,9 @@ get_veg_module <- function(input_veg,
                                 new_site = new_site,
                                 gridres = input_veg$gridres, dbparms = dbparms,
                                 machine_host = machine_host, input_veg = input,
-                                source = input_veg$source)
+                                source = input_veg$source,
+                                Rbinary = "R"
+                                )
```
infotroph, line 58:
```suggestion
                                Rbinary = "R")
```

modules/data.land/R/get_veg_module.R
```diff
@@ -92,7 +94,9 @@ get_veg_module <- function(input_veg,
                                ##   <age>70</age>
                                ##  </metadata>
                                ##
-                               icmeta = input_veg$metadata)
+                               icmeta = input_veg$metadata,
+                               Rbinary = "R"
+                               )
```
infotroph, line 99:
```suggestion
                               Rbinary = "R")
```

modules/data.land/R/put_veg_module.R
```diff
@@ -72,7 +72,9 @@ put_veg_module <- function(getveg.id, dbparms,
                                               pfts = pfts,
                                               source = input_veg$source,
                                               n.ensemble = n.ensemble,
-                                              host.inputargs = host.inputargs)
+                                              host.inputargs = host.inputargs,
+                                              Rbinary = "R"
+                                              )
```
infotroph, line 77:
```suggestion
                                              Rbinary = "R")
```

infotroph (changes requested):
This is looking very good and I think it's veeery nearly ready to merge. I've left some suggested changes with minor code style fixes; the easiest way to get them integrated will be to click "add to batch" for each one and then commit the whole batch as one change.

infotroph:
@its-Maniaco Please also add a Roxygen documentation string for the new `Rbinary` parameter by adding the followin between current lines 57 and 58 of `base/utils/R/convert.input.R`:

```
##' @param Rbinary Command (including path, if needed) to call the R executable on `host`

**Comment by Its-Maniaco** [comment]
cafe_hush_by_zen_master_ddfgn5v

**Review by infotroph — APPROVED** [review]

File: modules/data.atmosphere/R/download.raw.met.module.R (hunk 4 above)
Comment (mdietze) [inline modules/data.atmosphere/R/download.raw.met.module.R:95]: So it looks like the general pattern you've followed is that anywhere `convert.inputs` is called you've hardcoded in `Rbinary = "R"`. The result of this is that if someone passes in a non-standard Rbinary path in via the `settings` object that this path will never get used, which is going to break real-world workflows even if it doesn't break the Github Actions.

Comment (infotroph) [reply modules/data.atmosphere/R/download.raw.met.module.R:95]: It seems the existing `host` argument is likely to be passed the entire `settings$host` section, and therefore if Rbinary is set in `settings` I bet it will be accessible inside `convert.inputs` as host$Rbinary`. Is that a reliable enough assumption that we can eliminate the new `Rbinary` argument and read it from `host`?

Comment (mdietze) [reply modules/data.atmosphere/R/download.raw.met.module.R:95]: Doh, I didn't realize that `host` was an argument! If that's the case, then yes, we can drop Rbinary as an argument and just replace the original Rbinary line of code
```
Rbinary <- ifelse(!exists("settings") || is.null(settings$host$Rbinary),"R",settings$host$Rbinary)
```
with something like
```
Rbinary <- ifelse(is.null(host$Rbinary),"R",host$Rbinary)
```

Comment (infotroph) [reply modules/data.atmosphere/R/download.raw.met.module.R:95]: @Its-Maniaco I think this is the right way to go -- can you undo adding the `Rbinary` argument and implement @mdietze's proposal instead? 
Sorry for the design change! The fact that none of us remembers what all the arguments to `convert.input` do seems like good evidence it has too many parameters :wink:

File: base/utils/R/convert.input.R [hunk 15]
```diff
@@ -55,6 +55,7 @@
 ##' @param forecast Logical: Is the data product a forecast?
 ##' @param ensemble An integer representing the number of ensembles, or FALSE if it data product is not an ensemble.
 ##' @param ensemble_name If convert.input is being called iteratively for each ensemble, ensemble_name contains the identifying name/number for that ensemble.
+##' @param Rbinary command (including path, if needed) to call the R executable on `host`
```
infotroph, line 58: If removing `Rbinary`, delete this line and edit line 46 above (gitHub won't let me comment on it directly because it's not yet changed in this PR) to 
```
##'   Currently only \code{host$name} and \code{host$Rbinary} are used by \code{convert.input},
##'    but whole list is passed to other functions
```

infotroph:
@Its-Maniaco I took the liberty of working up a patch that implements the "read from host list" approach we recommended in the comments above. If you review and merge https://github.com/Its-Maniaco/pecan/pull/1 (note that it's a pull request in *your fork* of PEcAn, not in the PecanProject upstream that I'm typing this comment inside), my changes should immediately show up here... and if I did it right, this here PR should switch to showing only two files changed instead of 12.

infotroph (approved):
Thank you! I think this is now ready to go. Thanks for sticking with this, @Its-Maniaco.

@mdietze, does this revision fix your change requests?

mdietze (approved):

PR #2956, opened 2022-07-12
Fix some warnings and notes from PEcAn.MA package.

nanu1605:


modules/meta.analysis/R/single.MA.R
```diff
@@ -13,8 +13,8 @@
 ##' single.MA. This will allow power analysis  to run repeated MA outside of the
 ##' full loop over traits and PFTs.
 ##' @title Single MA
-##' @param data
 ##' @param j.chains number of chains in meta-analysis
+##' @param data data frame generated by jagify function with indexed values for greenhouse, treatment, and site (ghs, trt, site) as well as Y, SE, and n for each observation or summary statistic
```
infotroph, line 17: Please wrap lines similar to my suggestion in `meta.analysis.summary` above

modules/meta.analysis/DESCRIPTION
```diff
@@ -25,19 +25,19 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
     streamline the interaction between data and models, and to improve the
     efficacy of scientific investigation. The PEcAn.MA package contains
     the functions used in the Bayesian meta-analysis of trait data.
-Depends:
+Imports:
+    coda (>= 0.18),
     XML,
+    R2WinBUGS,
```
infotroph, line 31: We don't actually use any functionality from the R2WinBUGS package in PEcAn.MA, we only hyperlink to it in the documentation as an acknowledgement of design inspiration. 

You're totally correct that the R checks demand that we declare dependencies on all the packages whose documentation we link to, but in this case I think the correct fix is to avoid linking to it rather than to declare the dependency -- PEcAn users don't need to have R2WinBUGS installed to understand that we were inspired by it. 

Can you please remove this line and then rewrite the docs for `write.ma.model` to cite R2WinBUGS in plain text instead of with a link? Like so:

```
- ##'  Writes a meta-analysis model based on available data and prior specification. Adapted from \code{\link[R2WinBUGS]{write.model}}
+ ##'  Writes a meta-analysis model based on available data and prior specification. Adapted from \code{R2WinBUGS::write.model}
```

modules/meta.analysis/R/meta.analysis.summary.R
```diff
@@ -36,7 +36,7 @@ pecan.ma.summary <- function(mcmc.object, pft, outdir, threshold = 1.2, gg = FAL
   for (trait in names(mcmc.object)) {
     
     if (gg) {
-      gg <- require(ggmcmc)
+      gg <- requireNamespace(ggmcmc)
```
Aariq, line 39:
```suggestion
      gg <- requireNamespace("ggmcmc")
```
Also, this sort of thing is usually included in the part of the function that needs the package (i.e. line 42).  Typically something like:

```r
if(gg) {
    if(!requireNamespace("ggmcmc")) {
      stop("`gg=TRUE` requires the 'ggmcmc' package. Please install 'ggmcmc' or run with `gg=FALSE`")
    } else {
      <code that should run when gg = TRUE>
    }
  }
```

infotroph (commented):
This is good work. I left a few suggestions for edits -- some of them you can apply directly with the "commit suggestion" button (if you do this, make sure to pull the result back from GitHub to your local branch before making other edits), others will need more manual changes from you. 

I'll be offline again for the rest of this week and should be back to more regular Internet access starting on the 16th.

modules/meta.analysis/R/meta.analysis.summary.R
infotroph: Good catch, @Aariq! Looks like we were reviewing at the same time -- I'm marking yours resolved since I incorporated it into one of my suggestions below.

modules/meta.analysis/R/meta.analysis.summary.R
```diff
@@ -27,7 +27,7 @@
 ##' }
 ##' @author David LeBauer, Shawn Serbin
 pecan.ma.summary <- function(mcmc.object, pft, outdir, threshold = 1.2, gg = FALSE) {
-  
+  ggmcmc <- is.mcmc.list <- theme_set <- theme_bw <- ggs <- pdf<- box <- dev.off <- NULL
```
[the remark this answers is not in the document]
nanu1605: @Aariq I was fixing warnings regarding undefined functions. But can you help me out in fixing undefined variables too?

modules/meta.analysis/R/approx.posterior.R
```diff
@@ -111,15 +111,15 @@ approx.posterior <- function(trait.mcmc, priors, trait.data = NULL, outdir = NUL
...
+      posteriors[trait, "paramb"] <- stats::sd(dat)
       if (do.plot) {
         .dens_plot(posteriors, priors, ptrait, dat, trait, trait.data)
       }
     }
   }  ## end trait loop
   
   if (do.plot) {
-    dev.off()
+    imguR::dev.off()
```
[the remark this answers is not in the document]
mdietze: Pretty sure `dev.off` is in `grDevices`

modules/meta.analysis/DESCRIPTION
```diff
@@ -34,7 +34,8 @@ Imports:
     PEcAn.logger,
     MASS,
     PEcAn.settings,
-    rjags
+    rjags,
+    ggplot2
```
Aariq, line 38: If `ggplot2` is only used for optional arguments (that `gg=TRUE` argument) then it should probably be in `Suggests` rather than `Imports`.

modules/meta.analysis/R/meta.analysis.summary.R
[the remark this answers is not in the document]
Aariq: can you point me to an example?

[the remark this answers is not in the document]
Aariq: I don't know how to do this in base R, sorry. The help file for `transform()` explicitly says it shouldn't be used for programming, so probably the solution is to totally re-factor this section, unfortunately.  Here's a *hint* at how you might do it with `dplyr` and avoid the undefined global variable with the `.data` pronoun:

```
r <-
  r %>% 
  dplyr::mutate(site_id = as.integer(factor(.data$site_id, unique(.data$site_id))))
```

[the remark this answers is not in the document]
Aariq: See also: https://dplyr.tidyverse.org/articles/programming.html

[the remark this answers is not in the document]
Aariq: Opened an issue for this: https://github.com/PecanProject/pecan/issues/2961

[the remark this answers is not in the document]
nanu1605: Sure @Aariq, I will search for something to replace `transform()` and will share it with you on slack.

nanu1605:
Hi @infotroph @Aariq @mdietze, Should I manually update the `docker/depends/pecan.depends.R` to clear the PR checks?

mdietze:
@nanu1605 the script itself includes the answer to your question
```
# autogenerated do not edit
# use scripts/generate_dependencies.R
```

Makefile.depends
```diff
@@ -1,43 +1,55 @@
...
+$(call depends,models/lpjguess): | .install/base/utils .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/remote
+$(call depends,models/maat): | .install/modules/data.atmosphere .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/remote .install/base/settings .install/base/utils
+$(call depends,models/maespa): | .install/modules/data.atmosphere .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/remote .install/base/utils
+$(call depends,models/preles): | .install/base/utils .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/modules/data.atmosphere .install/base/utils
+$(call depends,models/sibcasa): | .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/utils
+$(call depends,models/sipnet): | .install/modules/data.atmosphere .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/remote .install/base/utils
+$(call depends,models/stics): | .install/base/settings .install/base/PEcAn.DB.Rcheck/PEcAn.DB .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/utils .install/base/remote
+$(call depends,models/template): | .install/base/PEcAn.DB.Rcheck/PEcAn.DB .install/base/PEcAn.logger.Rcheck/PEcAn.logger .install/base/utils
```
infotroph, line 55: All these added lines that contain `PEcAn.*.Rcheck` are in error. This is arguably a bug in the `generate_dependencies.R` script (It looks for everything that looks like an R package anywhere in the PEcAn directory, even inside Rcheck folders), but the easiest fix will be to remove all the `*.Rcheck` directories from your PEcAn folder and run `generate_dependencies.R` again.

modules/meta.analysis/R/meta.analysis.summary.R
```diff
@@ -36,13 +35,18 @@ pecan.ma.summary <- function(mcmc.object, pft, outdir, threshold = 1.2, gg = FAL
   for (trait in names(mcmc.object)) {
     
     if (gg) {
-      gg <- require(ggmcmc)
+      if (!requireNamespace(ggmcmc, quietly = TRUE)) {
```
infotroph, line 38:
```suggestion
      if (!requireNamespace("ggmcmc", quietly = TRUE)) {
```

Makefile.depends
nanu1605: Thanks @infotroph, I will do the changes ASAP

modules/meta.analysis/DESCRIPTION
```diff
@@ -25,21 +25,25 @@ Description: The Predictive Ecosystem Carbon Analyzer (PEcAn) is a scientific
...
+    PEcAn.DB,
     PEcAn.logger,
+    MASS,
     PEcAn.settings,
     rjags
 Suggests:
+    ggmcmc,
+    rmarkdown,
+    knitr,
+    ggplot2,
```
infotroph, line 42: Purely aesthetic: Keeping dependency lists in alphabetical order helps when humans are reading them (the R checks don't care about the order).

```suggestion
    ggmcmc,
    ggplot2,
    knitr,
    rmarkdown,
```

infotroph (approved):

infotroph:
All of these commits are also included in #2965, so  we should merge that and close this PR, but I'm leaving it open (marked as draft to prevent accidental merging) until then for easy reference to the review discussions.

infotroph:
These changes have been merged as part of #2965

PR #2963, opened 2022-07-16
Documentation of Some functions and dataset in PEcAn.MA package.

nanu1605:


infotroph:
The changes look good, but since all of these commits are also included in #2965 the easiest resolution is to merge that and close this.

PR #2965, opened 2022-07-17
fixed some global variables in PEcAn.MA package

nanu1605:


modules/meta.analysis/R/jagify.R
```diff
@@ -34,15 +34,15 @@ jagify <- function(result, use_ghs = TRUE) {
...
-                        citation_id = citation_id,
-                        ghs = greenhouse,
-                        site = site_id,
-                        trt_name = name), 
+                        r$citation_id <- r$citation_id,
+                        ghs = r$greenhouse,
+                        site = r$site_id,
+                        trt_name = r$name), 
               select = c("stat", "n", "site_id", "trt_id", "mean", "citation_id", "greenhouse", 
                          "ghs", "treatment_id", "site", "trt_name")) # original versions of greenhouse, treatment_id, site_id, and name
```
Aariq, line 47: I don't think this is valid R code.  You need to move the assignments to `r` outside of `subset()` and remove the `transform()` function.

infotroph:
It looks like this branch contains a lot of changes (everything up to commit 8ad821e9c67a877c9e3675849fef998379a0a756) that are already present in #2956 -- this will cause issues when trying to merge them all into develop. Can you please cherry-pick the last three commits (95880de, 49905f5, 57c8786) into a new branch that starts from the current tip of the `develop` branch, open a new PR from that branch, and close this one? 

I'm sorry that this will be some hassle, but it's better to resolve it now than to wait -- Git branch divergences only get more painful the longer they continue.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -7,7 +7,7 @@
 # http://opensource.ncsa.illinois.edu/license.html
 #-------------------------------------------------------------------------------
 
-run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.2, dbfiles, dbcon, use_ghs = TRUE) {
+run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.2, dbfiles, dbcon, use_ghs = TRUE, update) {
```
mdietze, line 10: would be good to have a default on `update`

mdietze (changes requested):

modules/meta.analysis/R/run.meta.analysis.R
nanu1605: Hi @mdietze, I think the default value of `update` should be `FALSE`, So whenever `TRUE` is passed as an argument, the `run.meta.analysis` will run. Am I right?

modules/meta.analysis/R/jagify.R
nanu1605: What I think is to restore this file for now and should create a sepreate PR for the removal of the `transform()` function.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -190,7 +190,7 @@ run.meta.analysis <- function(pfts, iterations, random = TRUE, threshold = 1.2,
   dbcon <- PEcAn.DB::db.open(database)
   on.exit(PEcAn.DB::db.close(dbcon), add = TRUE)
 
-  result <- lapply(pfts, run.meta.analysis.pft, iterations = iterations, random = random, 
+  result <- lapply(pfts, run.meta.analysis.pft(update = settings$meta.analysis$update), iterations = iterations, random = random,
```
mdietze, line 193: This isn't how lapply works. You need to pass update as one of the arguments to lapply, which then passes them to the function for each pft, same as is being done for the other arguments here.

mdietze (approved):

modules/meta.analysis/R/jagify.R
```diff
@@ -39,7 +39,7 @@ jagify <- function(result, use_ghs = TRUE) {
                         site_id = as.integer(factor(site_id, unique(site_id))), 
                         greenhouse = as.integer(factor(greenhouse, unique(greenhouse))),
                         mean = mean, 
-                        citation_id = citation_id,
+                        citation_id = =citation_id,
```
infotroph, line 42:
```suggestion
                        citation_id = citation_id,
```

modules/meta.analysis/R/single.MA.R
```diff
@@ -13,8 +13,10 @@
 ##' single.MA. This will allow power analysis  to run repeated MA outside of the
 ##' full loop over traits and PFTs.
 ##' @title Single MA
-##' @param data
 ##' @param j.chains number of chains in meta-analysis
+##' @param data data frame generated by jagify function
```
dlebauer, line 17: should this line come before j.chains so that it is in the same order as the function arguments?

dlebauer (changes requested):
looks great! 
- most important thing would be to add a comment about how and why new.env is being used 
- consider not exporting the p.point.in.prior function

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -43,10 +43,12 @@ run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.
   PEcAn.logger::logger.info("-------------------------------------------------------------------")
   
   ## Load trait data for PFT
-  load(file.path(pft$outdir, "trait.data.Rdata"))
-  load(file.path(pft$outdir, "prior.distns.Rdata"))
+  trait_env <- new.env()
```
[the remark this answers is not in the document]
infotroph: To load the file contents into their own object rather than the function's environment; this is (1) the recommended practice to avoid overwriting any existing objects with the same name as the loaded object(s), and (2) avoids "no visible binding for global variable" warnings from R CMD check.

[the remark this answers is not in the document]
infotroph: @dlebauer I now see you were asking for an explanatory comment in the code, not just in this thread. In the spirit of "make the code so clear it doesn't need comments", would better variable names have helped you follow what was happening here?

modules/meta.analysis/R/single.MA.R
nanu1605: Yes, I will make the changes

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -235,15 +242,12 @@ runModule.run.meta.analysis <- function(settings) {
 ##'
 ##' used to compare data to prior, meta analysis posterior to prior
 ##' @title find quantile of point within prior distribution
-##' @param point 
+##' @param point quantile of given prior to return
 ##' @param prior list of distn, parama, paramb
 ##' @return result of p<distn>(point, parama, paramb)
 ##' @export p.point.in.prior
```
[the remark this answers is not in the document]
nanu1605: @dlebauer Should I delete the whole `p.point.in.prior`? As it is now not used anywhere.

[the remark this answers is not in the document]
nanu1605: @dlebauer?

modules/meta.analysis/R/run.meta.analysis.R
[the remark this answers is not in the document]
infotroph: For cases where other file formats would actually be preferable, I support this change. But I see `load`'s "assign in the parent env by default" behavior as a minor historical annoyance rather than a reason to avoid the function entirely: .Rdata is a pretty good file format that's well suited for many applications, `load` is the obvious way to read it, and at this point even if we changed the format for new files we'd probably want to keep supporting .Rdata indefinitely so that old meta-analysis results can be read and updated.

modules/meta.analysis/R/run.meta.analysis.R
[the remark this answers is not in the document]
nanu1605: Sure I will do the changes and commit ASAP.

dlebauer (approved):

PR #2971, opened 2022-07-20
Fixed some global variables in PEcAn.DB package

nanu1605:


nanu1605:
@mdietze Are these changes ok?

base/db/R/query.prior.R
```diff
@@ -31,7 +32,7 @@
 ##'   con <- db.open(...)
 ##'   query.priors("ebifarm.pavi", c("SLA", "Vcmax", "leaf_width"), con = con)
 ##' }
-query.priors <- function(pft, trstr = NULL, con = NULL, ...){
+query.priors <- function(pft, trstr = NULL, con = NULL, ..., update){
```
robkooper, line 35: put update before the `...` that should be the final argument (list of leftover parameters)

robkooper (changes requested):

base/db/R/query.prior.R
```diff
@@ -50,7 +51,7 @@ query.priors <- function(pft, trstr = NULL, con = NULL, ...){
           "`settings` object not found. Unable to connect to database."
         )
       }
-      params <- settings[[c("database", "bety")]]
+      params <- update[[c("database", "bety")]]
```
infotroph, line 54: This won't work because `update` will just be a logical scalar, not the entire settings object. I think the change to make here is to have `query.prior` give up if no params are passed in `...`. Specifically, lines 45-54 can be deleted and  replaced with

```
PEcAn.logger::logger.severe(
  "No connection (`con`) specified and no connection parameters given in `...`.",
  "Unable to connect to database.")
```

infotroph (changes requested):

base/db/R/query.prior.R
moki1202: @infotroph let's say we need to make this work. can simply assigning settings to our update variable in line 53 solve this?

infotroph: @moki1202 Yes, it would be possible to change the design to pass the whole settings object rather than subsets of it, but I think in this case, since it's only needed as the last of *three* possible ways of creating the connection, it's cleaner to remove it and let the function fail if `con` isn't provided.

base/db/R/get.trait.data.R
```diff
@@ -56,7 +56,7 @@ get.trait.data <- function(pfts, modeltype, dbfiles, database, forceupdate,
     # `query_priors`, but haven't done so yet because that requires
     # prepared statements and therefore requires the Postgres driver. 
     all_priors_list <- lapply(format(pft_ids, scientific = FALSE), query.priors,
-                              con = dbcon, trstr = trait.names)
+                              con = dbcon, trstr = trait.names, update = settings$db$update)
```
infotroph, line 59: I'm not seeing where `settings` is defined inside `get.trait.data`, so I'm surprised this doesn't still throw a global variable warning. Am I missing a definition somewhere above?

base/db/R/query.prior.R
```diff
@@ -31,7 +32,7 @@
 ##'   con <- db.open(...)
 ##'   query.priors("ebifarm.pavi", c("SLA", "Vcmax", "leaf_width"), con = con)
 ##' }
-query.priors <- function(pft, trstr = NULL, con = NULL, ...){
+query.priors <- function(pft, trstr = NULL, con = NULL, update, ...){
```
infotroph, line 35:
```suggestion
query.priors <- function(pft, trstr = NULL, con = NULL, update = FALSE, ...){
```

base/db/R/query.prior.R
```diff
@@ -22,6 +22,7 @@
 ##' @param con Database connection object.
 ##' @param ... Optional arguments for connecting to database (e.g.
 ##'   password, user name, database).
+##' @param update
```
infotroph, line 25: The current CI failure is from Roxygen complaining that this parameter description is empty

base/db/R/get.trait.data.R
nanu1605: Hi @infotroph, I added the `update` variable in the `query.priors` function, and here is the call to that function that's why I passed `update` as an argument. I have done some changes now and removed `settings` from `get.trait.data`. Please review it once

base/db/R/get.trait.data.pft.R
```diff
@@ -17,11 +17,12 @@
 ##' @param dbcon database connection
 ##' @param forceupdate set this to true to force an update, auto will check to see if an update is needed.
 ##' @param trait.names list of trait names to retrieve
+##' @param update logical: Rerun the db if result files already exist?
 ##' @return updated pft with posteriorid
 ##' @author David LeBauer, Shawn Serbin, Rob Kooper
 ##' @export
 get.trait.data.pft <- function(pft, modeltype, dbfiles, dbcon, trait.names,
-                               forceupdate = FALSE) {
+                               forceupdate = FALSE, update = FALSE) {
```
infotroph, line 25: What's the difference between `forceupdate` and `update` here? Should they be combined into one argument?

nanu1605: I think both of them are different.

infotroph (changes requested):

base/db/R/query.prior.R
```diff
@@ -31,7 +32,7 @@
 ##'   con <- db.open(...)
 ##'   query.priors("ebifarm.pavi", c("SLA", "Vcmax", "leaf_width"), con = con)
 ##' }
-query.priors <- function(pft, trstr = NULL, con = NULL, ...){
+query.priors <- function(pft, trstr = NULL, con = NULL, update = FALSE, ...){
```
mdietze, line 35: Why is `update` being added to query.priors? I don't see anywhere this variable is used within the function, nor does the concept of `update` make sense in a function that only does a SELECT but no INSERT

base/db/R/get.trait.data.pft.R
mdietze: @nanu1605 I can't find anywhere in this function where `update` is used other than where you added it to `query.priors`, which as noted there seems to be an unnecessary addition. I think you can remove the `update` argument and usage in query.priors

base/db/R/get.trait.data.R
```diff
@@ -27,11 +27,12 @@
 ##' @param trait.names Character vector of trait names to search. If
 ##'   `NULL` (default), use all traits that have a prior for at least
 ##'   one of the `pfts`.
+##' @param update logical: Rerun the db if result files already exist?
 ##' @return list of PFTs with update posteriorids
 ##' @author David LeBauer, Shawn Serbin, Alexey Shiklomanov
 ##' @export
 get.trait.data <- function(pfts, modeltype, dbfiles, database, forceupdate,
-                           trait.names = NULL) {
+                           trait.names = NULL, update = FALSE) {
```
mdietze, line 35: Could you remove the addition of `update` as an argument to this function? As far as I can tell the only place it is used is where you added it to `query.priors` and `get.trait.data.pft`, and in both cases this appears to be a mistake.

mdietze (changes requested):

nanu1605:
Sorry @mdietze, I used `update` to remove settings as a global variable but later @infotroph suggested to removed the lines containing the `settings` variable and I forgot to remove `update` from the files. I am updating them now

infotroph (approved):

mdietze (approved):

base/db/R/query.prior.R
dlebauer: resolved by removing update as an argument in https://github.com/PecanProject/pecan/pull/2971/files/a2dc5df89b728d0b37a170d4434e445a9040b597#r928076659

dlebauer:
ready to be merged after tests pass; I dismissed @robkooper's requested changes which were no longer applicable after deciding not to pass the 'update' argument to the function.

PR #2645, opened 2020-06-30
GEE - SMAP script

ayushprd:
## Description

This PR adds the `gee2pecan_smap()` function which can be used for getting soil moisture data from Earth Engine.

## Motivation and Context

To be used in the remote data module.

## Installation instructions
As the module is not yet integrated with the PEcAn workflow, these steps will have to be performed manually.
This script requires Python3 and the python package manager pip

1. Sign up for the Google Earth Engine https://earthengine.google.com/new_signup/

2. Install the earth engine python api.
 `pip install earthengine-api` on Linux use `pip3` instead of `pip`
 Detailed steps including instructions for the alternative conda version are available [here](https://developers.google.com/earth-engine/python_installl)

3. Install the dependencies: download this [requirements.txt](https://raw.githubusercontent.com/ayushprd/PEcAn-GEE/master/requirements.txt) file to your working directory and run
 `pip install -r requirements.txt`

3. Once you have the earth engine account, run the following inside a python shell, (one time step)
 `import ee`
`ee.Authenticate()` [this will open up a browser and ask you to sign in with the google account registered for ee]

## Usage instructions
This module requires the input ROI (point and polygon type) in a geojson file format. The existing test file in `data.remote` cannot be used as SMAP data is not available for that region. This [test file](https://raw.githubusercontent.com/ayushprd/PEcAn-GEE/master/test2.geojson) can be downloaded for testing. I used [this](https://geojson.io/) for creating the geojson file. 
Please ensure that you have the name of the site/ROI inside the "properties" tag, this info is required for saving the file.

Example run:
At the location where `gee2pecan_smap.py` is present, open a python shell
```
from gee2pecan_smap import gee2pecan_smap

gee2pecan_smap("./test2.geojson", "./out/", "2018-01-01", "2018-12-31", "ssm")
```
The output netCDF file should be saved at the specified path. More info about the function arguments can be found out by `help(gee2pecan_smap)`

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
[...]

PR #2972, opened 2022-07-20
SMAP Download and Analysis

jbowers20:
## Description

This PR introduces three scripts and changes one pre-existing script.
Introduced scripts:
1. `SMAPvsModel_Comparison.R` 
2. `download_SMAP.R`
3. `download_SMAP_from_gee.R`  -- near copy of `download_SMAP_gee2pecan.R` script with slight modifications where necessary

Modified scripts:
1. `gee2pecan_smap.py`

## Motivation and Context

To be used in the remote data modules as a way to download and analyze SMAP data in comparison to soil moisture model output data.

## Usage instructions
`SMAPvsModel_Comparison` acts as the primary functional script of the download and analysis process.  From this script, you can define your input values.  Input values include:
1. Data range to be analyzed as (String)
2. File location for ensemble netCDF files 
3. File location containing XML file of all 39 CONUS NEON sites and their information 
4. File location to place output analytical data
5. File location for output of NEON site geoJson files
6. File location for output of SMAP netCDF files

Prior to integration into PEcAn workflow, some additional steps will need to be taken to initialize the gee2pecan_smap script.  See [this](https://github.com/PecanProject/pecan/pull/2645) PR for more details on doing so.  To initialize GEE on your personal system via the terminal, this command can be called after loading a python module: `earthengine authenticate --auth_mode=notebook`.

`download_SMAP.R`, `download_SMAP_from_gee`, and `gee2pecan_smap.py` are all being used as auxiliary scripts for the primary functional script and will need little to no modification prior to use.

Additionally, an XML file called `conus_sites.xml` containing all CONUS NEON sites in the desired format has been included as an additional document in this pull request.  

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
+##'@name SMAPvsModel_Comparison
+##'@description: This script collects ensemble model data from a specified file location, 
+##'analyzes that data to produce two exploratory graphs, and finally compares the
+##'model data to actual soil moisture data collected from NASA's SMAP satellite to
+##'produce an additional 5 graphs and a set of comparative statistics including: pbias, 
+##'RMSE, CRPS, Correlation, and r-squared values.
+##'
+##'This script is broken into 4 functions: extractSoilMoist, extractEnsambles,
```
mdietze, line 8: typo in Ensembles

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+library(verification)
+
+
+#################### INPUTS #################### 
+## Set Date Range ##
+start <- '2016-01-02'
+end <- '2016-07-16'
+
+## Directory of Ensembles Files ##
+ensamble_dir <- '/projectnb2/dietzelab/dongchen/All_NEON_SDA/NEON42/SDA_testrun/out/'
```
mdietze, line 68: fix spelling of ensemble

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+    } else {
+      soilMoist <- append(soilMoist, ncvar_get(nc,'SoilMoist'))
+      time <- append(time, PEcAn.utils::cf2datetime(nc[["dim"]][["time"]][["vals"]],nc$dim$time$units))
+    }
+  }
+  return(list(time, soilMoist))
+}
+
+
+#' @title extractEnsamlbes
```
mdietze, line 124: new, different misspelling of ensembles 😉

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+#' @return A list containing soil moisture and time values
+#'
+#' @examples 
+#' extractSoilMoist(betySiteId = '1000004945', yearStart = 2016, yearEnd = 2017, ENS = '25')
+#' 
+#' This extracts soil moisture values and time values from the 25th ensamble member netCDF file 
+#' in a specified file directory for the Harvard Forest ('1000004945') site during the years 
+#' 2016 and 2017.
+#' 
+extractSoilMoist <- function(betySiteId, yearStart, yearEnd, ENS){
```
mdietze, line 107: There's an argument to be made that a lot of these functions should be moved to the R directory so that they are more generally useable. If so, an important question is whether a function like this, which operates on model output not data, would be better placed in a different module

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+#' during the specified date range
+#' 
+#'
+#' @examples 
+#' extractEnsambles(betySiteId = '1000004945', start = as.Date('2016-01-02'), end = as.Date('2016-07-01'))
+#' 
+#' Extracts all 25 ensample members (default) from the Harvard Forest (1000004945) site between the
+#' '2016-01-02' - '2016-07-01' date range.
+#' 
+extractEnsambles <- function(betySiteId, ENSnum=25){
```
mdietze, line 141: would suggest a rename to make it clear this function just grabs soil moisture, not other variables. That said, it wouldn't take much work in either function to make the variable to be extracted an argument rather than something that's hardcoded

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+#' @param compare (boolean) to turn on/off comparison to actual smap data
+#'
+#' @return returns a list consisting of all of the analysis plots and statistics
+#'
+#' @examples 
+#' var <- runAnalysis('1000004945', '2016-01-02', '2016-07-16', TRUE)
+#' This runs an analysis on the Harvard Forest (1000004945) site with results between
+#' the date range of '2016-01-02' and '2016-07-16'
+#' 
+runAnalysis <- function(betySiteId, compare = TRUE){
```
mdietze, line 178: would benefit from a more specific name (especially as Analysis means something very specific in PEcAn's SDA module) e.g. runSMAPvalidation

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+    geom_line(data = mean_per_ENS, aes(x = time, y = mean, color = 'Mean')) +
+    geom_line(data = quantiles_lower, aes(x = time, y = q_lower)) +
+    geom_line(data = quantiles_upper, aes(x = time, y = q_upper)) + 
+    xlab('Date of Prediction') + ylab('Surface Soil Moisture (ssm)') +
+    labs(colour = 'Legend')
+  print(p2)
+  
+  if (compare) {
+    ## Analyses Requiring Actual SMAP Data ##
+    smap_data <- all_smap_data[[betySiteId]]
```
mdietze, line 225: all_smap_data appears to be a global variable. better to pass as a function argument, which could replace `compare` (i.e. make the default argument NULL and if it's not null then it's the data set you should compare to)

modules/data.remote/inst/SMAPvsModel_Comparison.R
```diff
@@ -0,0 +1,353 @@
...
+    runAnalysis(site, compare)
+    dev.off()
+  }
+}
+
+# Running All Sites #
+all_sites <- PEcAn.settings::read.settings(input_dir)
+sites_vector <- c()
+for (index in 1:length(all_sites)) {
+ sites_vector <- append(sites_vector, all_sites[[index]]$run$site$id)
```
mdietze, line 349: could replace this for loop with a sapply

modules/data.remote/inst/download_SMAP_from_gee.R
```diff
@@ -0,0 +1,110 @@
+##'@name download_SMAP_from_gee
```
mdietze, line 1: If this function is virtually identical to `download_SMAP_gee2pecan.R` then why do we have both (which will be really confusing to users)? If the old one doesn't work, wouldn't it make more sense to fix it than to create a duplicate? Also, by creating a duplicate I can't see the diff of what's old vs what you changed. Finally, is there any reason this should be in `inst` instead of `R` if its now ready to use?

mdietze: And you should add yourself to the author list

mdietze (changes requested):
At a high-level, the PR also needs updates to the PEcAn documentation and change log

modules/data.remote/inst/SMAPvsModel_Comparison.R
jbowers20: New module called netCDFvarExtraction created with the two functions extractPredictionVar (previously extractEnsemble) and extractVariableENS (previously extractSoilMoist).  extractPredictionVar and extractVariableENS have a new arg called predictVar that allows the user to enter the variable to extract (ex. "SoilMoist" for soil moisture values).

modules/data.remote/inst/SMAPvsModel_Comparison.R
jbowers20: Renamed to extractVariableENS.

modules/data.remote/inst/SMAPvsModel_Comparison.R
jbowers20: Renamed to runSMAPvalidation.

modules/data.remote/inst/download_SMAP_from_gee.R
jbowers20: `download_SMAP_from_gee` has been re-integrated into `download_SMAP_gee2pecan.R`.  I also added my name to the author list.

modules/data.remote/inst/SMAPvsModel_Comparison.R
mdietze: @jbowers20 while you pulled a few functions out of your main validation script, you kept them in the /inst directory so they're still not an exported part of the package.

mdietze (approved):

PR #2973, opened 2022-07-20
Fixed remaining Notes in PEcAn.MA

nanu1605:


modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -58,7 +58,7 @@ run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.
   dir.create(pathname, showWarnings = FALSE, recursive = TRUE)
   
   ## Convert data to format expected by pecan.ma
-  jagged.data <- lapply(trait_env$trait.data, PEcAn.MA::jagify, use_ghs = use_ghs)
+  jagged.data <- lapply(trait_env$trait.data, jagify, use_ghs = use_ghs)
```
dlebauer, line 61: I think `PEcAn.MA:::jagify` (3x :) will call non exported functions

modules/meta.analysis/R/jagify.R
```diff
@@ -33,16 +33,16 @@ jagify <- function(result, use_ghs = TRUE) {
...
-                        stat = as.numeric(stat), 
-                        n = as.numeric(n), 
-                        site_id = as.integer(factor(site_id, unique(site_id))), 
-                        greenhouse = as.integer(factor(greenhouse, unique(greenhouse))),
-                        mean = mean, 
-                        citation_id = citation_id,
-                        ghs = greenhouse,
-                        site = site_id,
-                        trt_name = name), 
+  r <- subset(dplyr::mutate(r,
```
dlebauer, line 36: If I recall, using 'transmute' instead of 'mutate' will only return the modified columns, making the following 'select' unnecessary.

dlebauer: Also, it is not clear - does 'subset' do anything here?

dlebauer (changes requested):
Can you please update the CHANGELOG.md with this and other PRs?

modules/meta.analysis/R/run.meta.analysis.R
nanu1605: Yes, but the R CMD check is throwing a NOTE when I use `PEcAn:::jagify` in place of `jagify`
`There are ::: calls to the package's namespace in its code. A package
  almost never needs to use ::: for its own objects:
  ‘jagify’`

moki1202: @nanu1605 we use `package::function` expression when the function here is exported, likewise we should ideally use `package:::function` expression when the function here is an internal function (not exported). Since in this case, you've unexported`jagify`, its a good call to use the latter. Let us know if the suggested changes throw a `NOTE`.

infotroph: I'm going to disagree here: Since `jagify` is defined in the same package that calls it, we don't need any namespace qualifier at all -- whether or not it's exported! Unless any tests complain, I think the code you have here is correct.

modules/meta.analysis/R/jagify.R
```diff
@@ -33,16 +33,16 @@ jagify <- function(result, use_ghs = TRUE) {
...
+                        n = as.numeric(.data$n), 
+                        site_id = as.integer(factor(.data$site_id, unique(.data$site_id))), 
+                        greenhouse = as.integer(factor(.data$greenhouse, unique(.data$greenhouse))),
+                        mean = .data$mean,
+                        citation_id = .data$citation_id,
+                        ghs = .data$greenhouse,
+                        site = .data$site_id,
+                        trt_name = .data$name), 
               select = c("stat", "n", "site_id", "trt_id", "mean", "citation_id", "greenhouse", 
                          "ghs", "treatment_id", "site", "trt_name")) # original versions of greenhouse, treatment_id, site_id, and name
```
infotroph, line 47:
```suggestion
  r$stat <- as.numeric(r$stat)
  r$n <- as.numeric(r$n)
  r$site_id <- as.integer(factor(r$site_id, unique(r$site_id)))
  r$greenhouse <- as.integer(factor(r$greenhouse, unique(r$greenhouse)))
  r$ghs <- r$greenhouse
  r$site <- r$site_id
  r$trt_name <- r$name
  
  r <- r[, c("stat", "n", "site_id", "trt_id", "mean", "citation_id", "greenhouse", 
             "ghs", "treatment_id", "site", "trt_name")]
```

infotroph: I believe if you do it this way there's no need to declare the `dplyr` import above in DESCRIPTION

infotroph (changes requested):

modules/meta.analysis/R/run.meta.analysis.R
moki1202: I must have overlooked here. I thought` jagify` was throwing a note. if there's no note then I agree with @infotroph too. No need to add namespace for daughter functions. Which makes me wonder what's the exact use of `:::` expression (Since all the non-exported functions are supposedly package's own functions.

nanu1605: The tests are passing but CI checks is throwing a NOTE

moki1202: @nanu1605 as suggested by @infotroph remove the namespace and then try again?

nanu1605: I did the suggested changes but I don't know why the CI checks are failing. Can you help me out here @infotroph @moki1202?

infotroph: Bah, the checks are failing on a place where `jagify` gets [called from inside the `PEcAn.priors` package](https://github.com/PecanProject/pecan/blob/develop/modules/priors/R/plots.R#L149). 

@dlebauer Do you see an easy way to remove the `jagify` usage in priors? If not I'm inclined to go back to exporting it and say we're stuck with it being part of the public interface.

nanu1605: @infotroph @moki1202 @dlebauer should I go back and export `jagify` ?

infotroph: @nanu1605 Yes, let's re-export `jagify`... Not exporting functions that don't need to be exported is good, but backwards compatibility, including supporting functions that were previously exported and are now depended on my other code, is arguably even more important.

mdietze:
@nanu1605 this pr is passing checks. If you address the changes requested by @dlebauer and @infotroph i'd be happy to pull it in

nanu1605:
Hi @mdietze, I have done the changes requested by @infotroph and @dlebauer but let them review it once.

mdietze:
@nanu1605 I don't see @dlebauer's requested CHANGELOG updates

nanu1605:
@mdietze I was thinking of updating the `CHANGELOG` and `Rcheck_reference.log` file of multiple packages in a separate PR once PR #2989 is merged.

infotroph:
@nanu1605 let's update the CHANGELOG as part of this PR and have Rcheck_reference be its own PR. Reasoning: It's much easier to write down what changed when you're looking at the changeset from the PR than later, whereas the Rcheck_reference updates are essentially cleanup operations.

mdietze:
@infotroph @dlebauer have the changes you both requested been addressed?

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ We are slowly change the license from NCSA opensource to BSD-3 to help with publ
 - rstudio was not working behind traefik.
 - plots now work in docker containers
 - when specifying diferent rstudio user, dev setup would mount pecan folder in wrong path.
+- transform() function from jagify.R in meta-analysis is removed (#2973).
+- fixed vignettes warning from PEcAn.DB package (#2966).
+- All the warnings from PEcAn.MA package is fixed except maintainers warning(#2965, #2973).
+- A global variable "settings" is fixed from PEcAn.DB package (#2971).
```
infotroph, line 43: The wording you have here does a good job of capturing all the changes but doesn't give users a clear idea what they need to do in response to the changes. This can be condensed more to focus on the high-level result:

```suggestion
- Code cleanup in PEcAn.MA to protect against global namespace pollution (#2965, #2973; @nanu1605) 
- Fixed vignettes and cleaned up internal warnings in PEcAn.DB (#2966, #2971; @nanu1605).
- A global variable "settings" is fixed from PEcAn.DB package (#2971).
```

infotroph (changes requested):

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ We are slowly change the license from NCSA opensource to BSD-3 to help with publ
 - rstudio was not working behind traefik.
 - plots now work in docker containers
 - when specifying diferent rstudio user, dev setup would mount pecan folder in wrong path.
+- Code cleanup in PEcAn.MA to protect against global namespace pollution (#2965, #2973; @nanu1605) 
+- Fixed vignettes and cleaned up internal warnings in PEcAn.DB (#2966, #2971; @nanu1605).
+- A global variable "settings" is fixed from PEcAn.DB package (#2971).
```
infotroph, line 42:
```suggestion
```

infotroph (approved):

dlebauer (approved):

PR #2982, opened 2022-07-24
Fixed Global function and variables from PEcAn.emulator

nanu1605:


modules/emulator/R/GaussProcess.R
```diff
@@ -240,12 +240,12 @@ GaussProcess <- function(x, y, isotropic = TRUE, nugget = TRUE, method = "bayes"
         }
       }
       psigibbs[g, ] <- psi
-      psijump <- update(psijump, psigibbs)
+      psijump <- stats::update(psijump, psigibbs)
```
mdietze, line 243: This is actually a call to the `update` function thats within this package, not the one in stats

modules/emulator/R/GaussProcess.R
```diff
@@ -256,17 +256,17 @@ GaussProcess <- function(x, y, isotropic = TRUE, nugget = TRUE, method = "bayes"
...
     if (is.numeric(anum) && 
         is.finite(anum) && 
-        exp(anum - aden) > runif(1)) {
+        exp(anum - aden) > stats::runif(1)) {
       tauw <- taustar
       S <- Sstar
     }
     tauwgibbs[g, ] <- tauw
-    tauwjump <- update(tauwjump, tauwgibbs)
+    tauwjump <- stats::update(tauwjump, tauwgibbs)
```
mdietze, line 264: This is actually a call to the `update` function thats within this package, not the one in stats

modules/emulator/R/gp_mle.R
```diff
@@ -39,7 +39,7 @@ gp_mle <- function(theta, d, nugget, myY, maxval = Inf) {
   
   S <- calcSpatialCov(d, phi, tauw)
   
-  val <- try(-sum(dmvnorm(myY, rep(mu, n), S + diag(tauv, n), log = TRUE)))
+  val <- try(-sum(emdbook::dmvnorm(myY, rep(mu, n), S + diag(tauv, n), log = TRUE)))
```
mdietze, line 42: This function comes from `mvtnorm`

modules/emulator/R/gp_mle.R
```diff
@@ -84,7 +84,7 @@ gp_mle2 <- function(theta, d, nugget, myY, maxval = Inf) {
   
   S <- calcSpatialCov(d, phi, tauw)
   
-  val <- try(-sum(dmvnorm(myY, rep(0, n), S + diag(tauv, n), log = TRUE)))
+  val <- try(-sum(emdbook::dmvnorm(myY, rep(0, n), S + diag(tauv, n), log = TRUE)))
```
mdietze, line 87: same

modules/emulator/R/predict.GP.R
```diff
@@ -142,12 +142,12 @@ predict.GP <- function(gp, xpred, cI = NULL, pI = NULL, splinefcns = NULL) {
     }
     
     if (nugget) {
-      Wprime    <- rmvnorm(1, S12 %*% S22inv %*% (W[i, ]), Sbar)
+      Wprime    <- SimDesign::rmvnorm(1, S12 %*% S22inv %*% (W[i, ]), Sbar)
```
mdietze, line 145: from mvtnorm

modules/emulator/R/predict.GP.R
```diff
@@ -142,12 +142,12 @@ predict.GP <- function(gp, xpred, cI = NULL, pI = NULL, splinefcns = NULL) {
...
       cInt[j, ] <- mu[i] + Wprime + y.trend
-      pInt[j, ] <- rnorm(npred * dim, cInt[j, ], sqrt(tauv1))
+      pInt[j, ] <- stats::rnorm(npred * dim, cInt[j, ], sqrt(tauv1))
     } else {
       cInt[j, ] <- mu[i] + S12 %*% S22inv %*% (y - mu[i]) + y.trend
-      mypred <- try(rmvnorm(1, cInt[j, ], Sbar), silent = TRUE)  ##wrap to prevent eigen failure
+      mypred <- try(SimDesign::rmvnorm(1, cInt[j, ], Sbar), silent = TRUE)  ##wrap to prevent eigen failure
```
mdietze, line 150: from mvtnorm

mdietze (changes requested):

modules/emulator/R/GaussProcess.R
nanu1605: Hi @mdietze, I can't find the `update` function in this package.

mdietze: https://github.com/PecanProject/pecan/blob/develop/modules/emulator/R/update.jump.R

nanu1605: So this update is call to the `update.jump` function?

mdietze: well, it's either to update.jump or update.mvjump depending on whether psijump is a member of the jump class or mvjump class. This is a deliberate use of R class polymorphism https://github.com/PecanProject/pecan/blob/a2c15b9dc3748649da2c39a7e3ae22e0acee46d4/modules/emulator/R/GaussProcess.R#L139

nanu1605: So all the three (line no: 218,249,264) `update` are not from `stats`

mdietze: correct

nanu1605:
@mdietze In case when no arguments passed for eg in line 105 of `modules/emulator/R/predict.GP.R` should we use `      utils::setTxtProgressBar()` directly? as we can not determine the min and max value

mdietze:
@nanu1605  1) if we switch to a new function we need to switch to that new functions args/usage, we can't just keep using it like the old function, nor can we use the function for updating the progress bar without defining it 2) in this case we most definitely do know the min, max, and state -- if you look on line 156 you see that the counter goes up to max = length(samp) and the counter variable is `i` instead of  `g`

modules/emulator/R/GaussProcess.R
```diff
@@ -156,7 +156,7 @@ GaussProcess <- function(x, y, isotropic = TRUE, nugget = TRUE, method = "bayes"
   haveTime <- require("time")
```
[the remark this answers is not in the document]
nanu1605: @infotroph will line 156 -160 will be replaced with `prevTime <- utils::txtProgressBar()` directly?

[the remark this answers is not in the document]
infotroph: @nanu1605 Yes, that assignment will replace all of lines 156-160. I'd name the variable `progress_bar` instead of `prevTime`, though.

nanu1605:
@infotroph I think changing `pda.calc.llik.par` to `PEcAn.assim.batch::pda.calc.llik.par` is creating a circular dependency between `emulator` and `assim.batch`. Should we move the function to some other package?

mdietze:
Poking around a bit I think the best option to remove the circular dependency would be to move the functions in minimize.GP.R that use pda.calc.llik* into assim.batch, but I haven't fully checked yet whether that would just create a different circular dependency in the other direction if those functions are used elsewhere in emulator.

nanu1605:
@mdietze I think we can move the whole file (minimize.GP.R) to assim.batch as all the functions in the minimize.GP.R are only used in assim.batch.

modules/emulator/R/predict.GP.R
```diff
@@ -100,11 +100,7 @@ predict.GP <- function(gp, xpred, cI = NULL, pI = NULL, splinefcns = NULL) {
   nsamp <- length(samp)
   # cInt <- pInt <- matrix(NA,nsamp,npred*dim)
   cInt <- pInt <- matrix(NA, nsamp, npred)
-  haveTime     <- require(time)
-  prevTime     <- NULL
-  if (haveTime) {
-    prevTime <- progressBar()
-  }
+  progress_bar <- utils::txtProgressBar()
```
infotroph, line 103:
```suggestion
  progress_bar <- utils::txtProgressBar(min = 0, max = length(samp), style = 3)
```

modules/emulator/R/predict.GP.R
```diff
@@ -134,35 +130,34 @@ predict.GP <- function(gp, xpred, cI = NULL, pI = NULL, splinefcns = NULL) {
...
       if (is.numeric(mypred)) {
         pInt[j, ] <- mypred
       }
     }
-    if (haveTime) {
-      prevTime <- progressBar(i/length(samp), prevTime)
-    }
+    pb <- utils::txtProgressBar(min = 0, max = length(samp), style = 3)
+    progress_bar <- utils::setTxtProgressBar(pb , i)
   }
```
infotroph, line 154:
```suggestion
    utils::setTxtProgressBar(progress_bar , i)
  }
  close(progress_bar)
```

infotroph (changes requested):

infotroph:
@nanu1605 I resolved the conflicts from merging #3027 and added a few more cleanup items to get checks passing -- can you please review my commits from today? 

I also left some suggestions above on the `txtProgressBar` calls -- I think they should be initialized once before the loop and then set inside it, not reinitialized.

nanu1605:
Sure @infotroph, I will do it ASAP

infotroph (approved):

infotroph:
@mdietze I think we've addressed your change requests and this is ready for re-review

mdietze (approved):

PR #2994, opened 2022-07-30
Change argument `env.samples` in `get.emsemble.samples()` to optional

ecamo19:
Changes argument `env.samples` in the `get.ensembles.samples()` function from required to optional

env.samples is a required argument in the get.ensemble.samples function. However not all the time these env samples are available. Thus, is necessary to change this argument to optional in case a user wants to get the ensemble samples without taking into account the env.samples.

## Description

The argument changes from required to optional, if the argument env.samples is NOT null the env.samples object will be appended to the pft.samples object.

## Motivation and Context

With the change proposed it will be possible to get ensemble samples without having the env.samples object  

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/uncertainty/R/ensemble.R
```diff
@@ -78,28 +78,32 @@ read.ensemble.output <- function(ensemble.size, pecandir, outdir, start.year, en
...
+                ans <- PEcAn.utils::get.sa.sample.list(pft.samples, env.samples, 0.5)
+                } else {
+                    pft.samples[[length(pft.samples) + 1]] <- env.samples
+                    names(pft.samples)[length(pft.samples)] <- "env"
+                    pft2col <- NULL
+                    for (i in seq_along(pft.samples)) {
+                        pft2col <- c(pft2col, rep(i, length(pft.samples[[i]])))
+                    }
+                }
+        }
```
Aariq, line 105: It's not clear to me what happens if `env.samples` is `NULL`.

modules/uncertainty/R/ensemble.R
```diff
@@ -78,28 +78,32 @@ read.ensemble.output <- function(ensemble.size, pecandir, outdir, start.year, en
 ##' @return matrix of (quasi-)random samples from trait distributions
 ##' @export
 ##' @author David LeBauer, Istem Fer
-get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples, 
-                                 method = "uniform", param.names = NULL, ...) {
+get.ensemble.samples <- function(ensemble.size, pft.samples,method = "uniform",
```
Aariq, line 81: Hey, while you're fixing things, could you change this so the possible options show up in the autocomplete better?
```suggestion
get.ensemble.samples <- function(ensemble.size, pft.samples,method = c("uniform", "halton", "sobol", "torus", "lhc"),
```
Then to the beginining of the function body add `method <- match.arg(method)`

Aariq (commented):
I'm not super familiar with this function but I thought I'd take a shot giving some feedback.  My only other suggestion would be to document the `env.samples` param a bit better.  Since you seem to know what it is, it'd be really helpful to add that info to the documentation.

Aariq:
Was your issue fixed by https://github.com/PecanProject/pecan/pull/3005 ?  If so, does this PR still need review?

ecamo19:
Issue was fixed  by [#3005](https://github.com/PecanProject/pecan/pull/3005 ) thus the PR proposed is no longer necessary

Aariq:
I converted this to a draft.  I may pick this up and implement some of my suggestions later.

PR #2999, opened 2022-08-02
remove leftover link to udunits2 package

infotroph:
R checks for utils now complain if run on a system that doesn't have `udunits2` installed. Fixed by rewording to point to new wrapper instead.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3001, opened 2022-08-02
removed udunits2 usages in `shiny/` and `scripts/`

nanu1605:
removed udunits2 from `shiny` and `scripts` folder


shiny/workflowPlot/server_files/model_data_plots_server.R
```diff
@@ -43,10 +43,10 @@ observeEvent(input$var_name_modeldata, {
 
 # Check that new units are parsible and can be used for conversion
 observeEvent(input$units_modeldata,{
-  if(udunits2::ud.is.parseable(input$units_modeldata)){
+  if(units::install_unit(input$units_modeldata)){
```
infotroph, line 46: As we discussed in PEcAn.DB, install_units isn't the right approach here: `ud.is.parseable` checks if the udunits library knows how to handle the unit, `install_unit` declares the unit as valid even if the udunits library has never seen it before. 

What we did in PEcAn.DB was to wrap `as_unit()` in a tryCatch, and that's probably the right approach here too. A shorter but probably much harder to understand hack would be `if( PEcAn.utils::ud_convert(1, input$units_modeldata, input$units_modeldata)) { #test if unit parseable by trying to convert to itself`

shiny/workflowPlot/server_files/model_plots_server.R
```diff
@@ -34,11 +34,11 @@ observeEvent(input$var_name_model, {
 
 # Check that new units are parsible and can be used for conversion
 observeEvent(input$units_model,{
-  parseable <- ifelse(udunits2::ud.is.parseable(input$units_model), "can", "cannot")
-  if(udunits2::ud.is.parseable(input$units_model)){
+  parseable <- ifelse(units::install_unit(input$units_model), "can", "cannot")
+  if(units::install_unit(input$units_model)){
```
infotroph, line 38: same comment as above

nanu1605:
@infotroph I think when PR #2999 will be merged the CI checks will pass

infotroph (changes requested):

shiny/workflowPlot/server_files/model_data_plots_server.R
moki1202: latter is a very cool approach indeed! @nanu1605 please go ahead and try this. If you're confused, ping me or Chris.

moki1202: I really like @infotroph 's latter suggestion, it's a very cool approach but now that you've written a replacement function for `ud.is.parseable`, go ahead and replace these calls with the `unit_is_parseable` function once you get the green light from other reviewers.

infotroph: After thinking more, I think my latter suggestion is a bad one -- it's clever but too easy to misunderstand. I can easily see a future reader (including myself in a year!) saying "The input and output units are the same, so this line does nothing! I can delete it!", so defining a new function with a clear name seems preferable.

shiny/workflowPlot/server_files/model_plots_server.R
```diff
@@ -34,11 +34,11 @@ observeEvent(input$var_name_model, {
 
 # Check that new units are parsible and can be used for conversion
 observeEvent(input$units_model,{
-  parseable <- ifelse(udunits2::ud.is.parseable(input$units_model), "can", "cannot")
-  if(udunits2::ud.is.parseable(input$units_model)){
+  parseable <- ifelse(PEcAn.utils::unit_is_parseable(input$units_model), "can", "cannot")
```
infotroph, line 37: This line seems unused, so I think we can delete it:
```suggestion
```

infotroph (changes requested):

shiny/workflowPlot/server_files/model_plots_server.R
infotroph: (I went ahead and committed this so that the final round of checks could kick off)

infotroph (approved):

PR #3005, opened 2022-08-04
Remove orphaned and troublesome query and logger message from ensemble.R

dlebauer:
Removing code implemented in PR https://github.com/PecanProject/pecan/pull/2126. The queried the DB and used the information to return some information. 

It was implemented as a quick workaround for a specific project to link pfts with specific sites. But at this point it has no effect other than to send a message via logger, or create an error if nothing is returned (perhaps because R has changed how it handles missing data). In [todays meeting](https://docs.google.com/document/d/1DUQyKEwGYTumr6ttFJA8vcwZXqgUTTxn4rjeVdomQtU/edit#bookmark=id.ucevo0tnkm6b) we decided to remove the code and punt the implementation of pft-site relationships to the next time it is needed


robkooper (approved):

robkooper:
checked more and the variables used in the code removed is not used anywhere else in the code.

PR #3006, opened 2022-08-04
Move ED2 internal data to unexported sysdata.rda

Aariq:
## Description
This moves history.csv and soil.csv files to `models/ed/data-raw/` which is in .Rbuildignore and instead includes them in sysdata.rda which is unexported by default.  These data files are necessary for functionality of `write.configs.ed.R` but do not need to be exported.

## Motivation and Context
Solves failing checks in #2985 

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


Aariq:
Failing test is undocumented dataset pftmapping.  This is documented in PR https://github.com/PecanProject/pecan/pull/2985.  Together, these two PRs should fix all (new) warnings for PEcAn.ED2

dlebauer:
Thanks for proposing this; I'm not familiar with the convention of using a data-raw folder; is this preferred over inst/extdata?

Aariq:
The approach I took was the one suggested in the R Packages book (https://r-pkgs.org/data.html).  They mention inst/extdata in the context of vignettes in that it's useful if you want to show examples of reading in data.  I can certainly try it though and see if `devtools::check()` complains.

Aariq:
> That also gives us extra flexibility on the format of the file; i.e., we can do a more standard CSV format, or other formats parsed in specific ways as appropriate.

I don't think there is any limitation on file format with the proposed solution (or any solution for this really).  A .rda file can hold any number of any kind of R objects and they can be read in and parsed in different ways in sysdata.R so this reading in and parsing doesn't have to happen inside of functions.

dlebauer (approved):
@Aariq I share @ashiklom 's concern that the sysdata.R needs to be rerun when the files are updated. However, these inputs seem like they won't change except when adding a new version of ED2.

And it does appear to follow recommendations / best practices in the textbook.

The other recommended option for internal data is to use dput and paste it in the code. That option is appealing in its simplicity but it also has downsides of not being easily readable.

Aariq:
Ok, so I think what I'll try to do then is merge this PR into #3006 since they have complimentary changes that, together, should result in passing checks.

PR #3009, opened 2022-08-06
Moved convert.input.R from utils to DB

nanu1605:
To move `convert.input` from `utils` to `DB`


infotroph:
Copying the checklist we discussed in Slack here for easy reference:

- [ ] check whether convert.input uses any unexported functions from PEcAn.utils. If so, see where else they're used and decide whether to move them to data.atm, copy them, or export them from utils
- [ ] As one commit, move the convert.input function definition and man page from PEcAn.utils to PEcAn.DB.
- [ ] As a separate commit, update the convert.input code to remove PEcAn.DB:: in front of calls to DB functions and to add PEcAn.utils:: in front of calls to utils. It's important to do it in a separate commit because otherwise Git will just show one giant blob deleted from utils and another added to DB, with no easy way to find the differences between them. Doing it in two steps makes it pretty easy to verify that the move part didn't change anything inside the blob
- [ ] Review the function documentation and make any edits needed — I doubt it will, but do check
- [ ] Implement any needed deprecation stub in PEcAn.utils. The usual best practice is to deprecate functions slowly over several releases to give users time to update, but the circular dependency demands quicker action. Decision needed: After this change, should a call to PEcAn.utils::convert.input() fail with the function not found or should it do nothing but throw a message to use the version in PEcAn.DB?
- [ ] Update the changelog and both packages’ NEWS files
- [ ] Update all calls to convert.inputs elsewhere in PEcAn

infotroph:
This looks great! Go head and tackle the last three points in my checklist above as part of this PR and we should be able to merge it all at once.

nanu1605:
@infotroph To change the `Rcheck_reference.log` PR #3000 is already opened. Should I merge it in this PR and close that one?

infotroph:
@nan1605 No, just update the files that this PR touches, probably by manual editing rather than a full new check run (i.e. only change the lines that need to change because of *this* PR). We'll see which PR gets merged first and resolve any conflicts after that in the remaining one.

base/db/tests/Rcheck_reference.log
```diff
@@ -85,15 +84,13 @@ Namespaces in Imports field not imported from:
 * checking replacement functions ... OK
 * checking foreign function calls ... OK
 * checking R code for possible problems ... NOTE
-db.exists: no visible binding for '<<-' assignment to ‘user.permission’
-db.exists: no visible binding for global variable ‘user.permission’
-load_data_single_run: no visible binding for global variable ‘var_name’
-load_data_single_run: no visible binding for global variable ‘vals’
+convert.input: no visible binding for global variable ‘settings’
+convert.input : log_format_df: no visible binding for global variable
+  ‘.’
```
infotroph, line 89: Let's fix these before merging this rather than write them into the log.

CHANGELOG.md
```diff
@@ -44,6 +44,11 @@ We are slowly change the license from NCSA opensource to BSD-3 to help with publ
   Note that both `units` and `udunits2` interface with the same underlying
   compiled code, so the `udunits2` *system library* is still required.
   (#2989; @nanu1605)
+- Added a new function `unit_is_parseable` in PEcAn.utils to replace `udunits2::ud.is.parseable`.
+  (#3002; @nanu1605)
```
infotroph, line 48: This should be in the "Added" section

CHANGELOG.md
```diff
@@ -44,6 +44,11 @@ We are slowly change the license from NCSA opensource to BSD-3 to help with publ
...
+- To remove the cyclic dependency between PEcAn.DB and PEcAn.utils `convert.input` function is moved
+  from PEcAn.utils to PEcAn.DB.
+  (#3009; @nanu1605)
```
infotroph, line 51: This should be in the "changed" section

infotroph: Also I recommend switching the order of the clauses to put the most important information first: "`convert.input is moved ... this was needed to resolve a cyclic dependency between ..."

base/db/tests/Rcheck_reference.log
nanu1605: @infotroph I have removed `settings` variable from `convert.input` but I am not sure how to remove `.` variable. Can you help me out here

nanu1605: @infotroph I have removed `.` and `settings` variable from `convert.input`. I am not sure if it's the correct method to do it.

base/db/R/convert.input.R
```diff
@@ -737,7 +737,7 @@ convert.input <-
     log_format_df = function(df){
       df  %>%
         format() %>%
-        rbind(colnames(.), .) %>%
+        rbind(colnames(.data$.), .data$.) %>%
```
Aariq, line 740: I think the `.` can be replaced with just `.data`.  Are there no tests to run this on to check? 

``` r
df %>% format() %>% rbind(colnames(.data), .data) %>%...
```

nanu1605: All the tests are passing before changing to `df %>% format() %>% rbind(colnames(.data), .data) %>%...`

Aariq: Whelp, that's not great because pretty sure what you had there was not valid R code.  Not your fault that there aren't tests for that function though.

You could also re-write this to not use the pipe.  Remember that `a %>% f(.)` is the same as `f(a)`

base/db/R/convert.input.R
```diff
@@ -736,7 +737,7 @@ convert.input <-
     log_format_df = function(df){
       df  %>%
         format() %>%
-        rbind(colnames(.), .) %>%
+        df %>% format() %>% rbind(colnames(.data), .data) %>%...
```
Aariq, line 740: Pretty sure this is not valid R code to pipe something into `...`.  Even if this line is valid, then the next line will error because nothing is getting piped into `purrr::reduce()`.

nanu1605: `rbind(colnames(df %>% format()), df %>% format())` @Aariq will this code work,

nanu1605: `rbind(colnames(df %>% format()), df %>% format())` @Aariq will this code work? I tried not to use pipe

Aariq: yeah, I think that works.  You could also do 
```suggestion
        rbind(colnames(format(df)), format(df))
```
I'm still a little confused about why this function exists.  Why would you want the column names to print twice?

Aariq:
``` r
df <- data.frame(x = 1:3, y = c("a", "b", "c"))

#why would you want this?
rbind(colnames(format(df)), format(df))
#>    x y
#> 1  x y
#> 11 1 a
#> 2  2 b
#> 3  3 c

#this does the same thing, but also WHY?
rbind(colnames(df), df)
#>   x y
#> 1 x y
#> 2 1 a
#> 3 2 b
#> 4 3 c
```

<sup>Created on 2022-08-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Aariq: Feels like someone tried to re-invent the wheel here.  `print(df, row.names = FALSE)` does something similar. 🤷‍♂️

nanu1605: @Aariq should I replace `df %>% format() %>% rbind(colnames(.data), .data) %>%...`  with `rbind(colnames(df), df) %>%` as it does the same thing? or should commit the suggestion?

Aariq: I don't think it does anything if you don't supply it with any arguments other than maybe strip attributes, but I guess I'd just leave `format()` in since there aren't any tests for this function.

base/db/R/convert.input.R
```diff
@@ -736,7 +737,7 @@ convert.input <-
     log_format_df = function(df){
       df  %>%
         format() %>%
-        rbind(colnames(.), .) %>%
+        rbind(colnames(format(df)), format(df))
         purrr::reduce( paste, sep=" ") %>%
         paste(collapse="\n")
     }
```
Aariq, line 743: If you do this, you need to get rid of `df %>%` as well.  Please test this function on any dataframe before asking for review.

infotroph (changes requested):

base/utils/NAMESPACE
```diff
@@ -7,7 +7,6 @@ export(cf2datetime)
 export(cf2doy)
 export(clear.scratch)
 export(convert.expr)
-export(convert.input)
```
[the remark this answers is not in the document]
nanu1605: It's a great idea. We change the name of `convert.input` to `convert_input` that eliminate the package ambiguity and the stub function will tell the user that the `convert. input` function is available in PEcAn.DB with a new name `convert_input`

moki1202:
@nanu1605 I think we'll need to add a line in both NEWS.md and changelog.md file stating that we've made a change in the function's name and why.

infotroph:
Replaced by #3026

Issue #2955, opened 2022-07-07
Update code to reflect new docker endpoint


Aariq:
Recently @robkooper changed the way the docker install works so that pecan is reached from http://pecan.localhost/ instead of http://localhost:8000/.  I've made the changes in the documentation in #2935, but there are still instances of "localhost:8000" in tests, shell scripts, and other misc. places that probably need to be updated.

Aariq:
I think this is where the change was made: https://github.com/PecanProject/pecan/pull/2936

Issue #2959, opened 2022-07-13
Better default for `host$qsub.jobid`


Aariq:
Currently the default value for `qsub.jobid` is `'Your job ([0-9]+) .*'`, but this is not very generic.  A pattern that matches any group of consecutive numbers, possibly with non-numbers before and/or after would be a better default.

``` r
qsub.jobid <- "\\D*([0-9]+)\\D*"

out_example <- c("job:12345", "submitted 12345", "12345 submitted", "job 12345 submitted")

sub(pattern = qsub.jobid, replacement = "\\1", x = out_example)
#> [1] "12345" "12345" "12345" "12345"
```

<sup>Created on 2022-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Aariq:
Relevant place in code:

https://github.com/PecanProject/pecan/blob/cd3e4c2dc1739b4165574dd1200f73886d80af9a/base/settings/R/check.all.settings.R#L401

Aariq:
somewhat related to #2958

PR #2935, opened 2022-06-07
Edits to documentation in book_source

Aariq:
## Description

As I go through the PEcAn bookdown site, I've been making edits where I find typos, missing links, or outdated instructions.  It's likely to be a lot of small commits as I go through the book, so let me know if you'd like me to "squash" them and I can figure out how to do that.

## Motivation and Context
Updating documentation.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [x] I have updated the CHANGELOG.md. 
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes. (not applicable)
- [x] All new and existing tests passed. (some tests failing, but unrelated to these changes)

PR #3010, opened 2022-08-07
fixes issues #2955 and #2959

hangyakuzero:
## Description

Recently @robkooper changed the way the docker install works so that pecan is reached from http://pecan.localhost/ instead of http://localhost:8000/. I've made the changes in the documentation in https://github.com/PecanProject/pecan/pull/2935, but there are still instances of "localhost:8000" in tests, shell scripts,etc. I have made the changes accordingly.
Also I have changed the regex pattern as required.

this fixes issue #2955 and  #2959
## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/settings/R/check.all.settings.R
```diff
@@ -398,7 +398,7 @@ check.settings <- function(settings, force = FALSE) {
         "qsub not specified using default value :", settings$host$qsub)
     }
     if (is.null(settings$host$qsub.jobid)) {
-      settings$host$qsub.jobid <- "Your job ([0-9]+) .*"
+      settings$host$qsub.jobid <- "Your job ^[A-Za-z0-9_-]*$"
```
Aariq, line 401: Not sure this will catch many Job IDs.  For example, if the HPC message for a submitted SLURM script is "Job 190231 submitted" or "Submitted as job 19203" this regex wouldn't work I think.  The idea in #2959 was to make the default regex able to catch any consecutive group of numbers anywhere in whatever the submission message is.

hangyakuzero: oh sorry I forgot about adding blank space in the regex, going to fix it now

Aariq: So it's not the blank space, it's that my HPC doesn't print "Your job", so this regex doesn't work. 

Try creating a few test submission messages like "Your job 812973 has been submitted", "Job 109238 submitted", "Submitted 1282", etc. and then using `sub()` to figure out what regex pattern returns only the numeric part.

I was pretty sure `"\\D*([0-9]+)\\D*"` would work, but when I tried it in an actual pecan.xml file it didn't.  I'm not sure why.  My HPC prints "Submitted batch job 4094596" upon SLURM script submission.  The regex that works for me is `".* ([0-9]+)"`

robkooper: you can change this by setting it in your pecan.xml:
```
<host>
  ...
  <qsub.jobid>.* ([0-9]+).*</qsub.jobid>
</host>
```
This is a very generic one, and might end up eating to much, or not mach. You can check it at: https://regex101.com/

base/settings/R/check.all.settings.R
```diff
@@ -398,7 +398,7 @@ check.settings <- function(settings, force = FALSE) {
         "qsub not specified using default value :", settings$host$qsub)
     }
     if (is.null(settings$host$qsub.jobid)) {
-      settings$host$qsub.jobid <- "Your job ([0-9]+) .*"
+      settings$host$qsub.jobid <- "Your job ^[A-Za-z0-9_ -]*$"
```
Aariq, line 401:
```suggestion
      settings$host$qsub.jobid <- "[^0-9]*([0-9]+)[^0-9]*"
```

Aariq: I'm going to go ahead and commit this suggestion since there hasn't been any activity on this PR for a while

infotroph:
This still won't catch jobids without a literal space before them (e.g. my old frenemy "ondemand52.aci.ics.psu.edu.568420642 submitted"), but my sense is that in practice every default involving qsub should be thought of less as a value that will work for most people and more as an example of how to specify your own value when the default inevitably fails on your system. Therefore I support merging this as it is.

infotroph (approved):

PR #3013, opened 2022-08-10
added API key authentication for REST API's

koolgax99:
## Description

This feature will introduce the concept of API keys in the PEcAn REST API's, an API key shall be passed by the user in the header X-API-Key

The API would be more secure now and in the future we can use this API key to track the requests from the user.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/R/auth.R
+++ b/apps/api/R/auth.R
@@ -41,6 +41,22 @@ validate_crypt_pass <- function(username, crypt_pass) {
   return(NA)
 }
 
+#* Check if the API key exists
+#* @param api_key API Key
+#* @return user details if apikey exists, else Invalid API Key
+#* @author Nihar Sanda
+validate_api_key <- function(api_key) {
+  res <- tbl(global_db_pool, "users") %>%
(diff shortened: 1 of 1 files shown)


apps/api/R/auth.R
```diff
@@ -41,6 +41,22 @@ validate_crypt_pass <- function(username, crypt_pass) {
...
+  res <- tbl(global_db_pool, "users") %>%
+    filter(apikey == api_key) %>%
+    collect()
+  
+  if (nrow(res) == 1) {
+    return(res)
+  }
+  
+  return(NA) 
+}
```
Aariq, line 58: Not sure how you're planning on having the users store the API key, but often packages have users put it in a .Renviron file.  Then, the `validate_api_key` can look for the key there by default.  An example: https://github.com/ropensci/webchem/blob/18e551609c8f1aa371f83aad1b281540e9df1066/R/chemspider.R#L22

mdietze (approved):

Issue #2987, opened 2022-07-25
Don't use `testthat` inside of functions


Aariq:
### Bug Description

The `testthat` package is made for unit testing, not use inside of functions.  `testhat` functions appear in at least a couple of places:
- `check_ed_metheader()`
- `check_ed_metfile()`

These should be replaced with calls to `stopifnot()` or something from the `assert` or `assertthat` packages

Aariq:
Other places:
- `PEcAn.atmosphere::check_met_input_file()`
- `PEcAn.atmosphere::check_unit()`

Aariq:
Partially fixed by #3016

Sarthakaga15:
Hi @Aariq , can I please work on this issue?

moki1202:
@Sarthakaga15 there's one more instance that you can work on, in `./models/ed/R/write_ed_veg.R`. if you need any further help on this, let me know.

PR #3016, opened 2022-08-22
Replace `testthat` funs used in package code

Aariq:
Adresses https://github.com/PecanProject/pecan/issues/2987
`testthat` is meant for testing, and not for using inside of functions.  This replaces `testthat` functions with `stopifnot()` or `if(cond) { stop("errormessage')}` where appropriate across all packages.


Aariq:
`PEcAn.data.atmosphere::check_met_input_file()` makes extensive use of `testthat` and will take some time to fix.  However, it's not currently used by anything.  Maybe this function doesn't need to get fixed in this PR.

dlebauer (approved):

PR #3017, opened 2022-08-25
moved minimize.GP.R from emulator to assim.batch

nanu1605:
To remove the circular dependency between `emulator` and `assim.batch`


moki1202:
@infotroph @mdietze Would it be a better idea here to leave an instance of `minimize.gp.R` in emulator mentioning it as "`deprecated`"?

nanu1605:
Referenced in #3027

Issue #2967, opened 2022-07-18
`run.write.configs()` opens too many database connections


Aariq:
https://github.com/PecanProject/pecan/blob/ae89856f24a62cd83a06fa0c59e7f902df4f63c5/base/workflow/R/run.write.configs.R#L37

Because this line is inside of a for-loop, I think it's opening (or trying to open) a number of database connections equal to the number of PFTs specified.  The `on.exit()` line also gets called once per PFT and prints a warning that the connection may have already been closed.

Aariq:
With 5 PFTs specified I get 4 warnings:

```
2022-07-18 22:22:38 WARN   [PEcAn.DB::db.close] : 
   Connection might have been closed already. 
2022-07-18 22:22:38 WARN   [PEcAn.DB::db.close] : 
   Connection might have been closed already. 
2022-07-18 22:22:38 WARN   [PEcAn.DB::db.close] : 
   Connection might have been closed already. 
2022-07-18 22:22:38 WARN   [PEcAn.DB::db.close] : 
   Connection might have been closed already. 
   ```

PR #3018, opened 2022-08-25
only open one connection to BETY instead of one per PFT (#2967)

Aariq:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


dlebauer (approved):

Aariq:
Looks like at least in the case of ED2, `write.ensemble.configs()`, which is called by `run.write.configs()` *also* opens a connection to BETY.  Probably only the outer connection should do this and pass the connection to the inner function.  Maybe a separate PR though?

PR #3021, opened 2022-08-26
Change `model2netcdf.ED2()` to include more variables from monthly -E- files and group by PFT

### LINKED ISSUE #2997: `model2netcdf.ED2()` output isn't grouped by PFT
### Bug Description

The .nc files resulting from conversion from ED2 output have variables that could be split up by PFT but aren't.  For example NPP in the .nc file only has the dimensions lat, lon, and time but it is possible to split up by PFT in the .h5 files produced by ED2.  The only variables that have PFT as a dimension are DBH, rate of change of DBH and plant density.

### To Reproduce

``` r
devtools::load_all("models/ed")
outdir <- tempfile()
unzip("models/ed/tests/testthat/data/ed2_run_output.zip", exdir = outdir)
file.copy("models/ed/tests/testthat/data/pecan_checked.xml", file.path(outdir, "pecan_checked.xml"))

settings <- PEcAn.settings::read.settings(file.path(outdir, "pecan_checked.xml"))
settings$outdir <- outdir

model2netcdf.ED2(settings = settings)

library(ncdf4)
x <- nc_open(list.files(outdir, pattern = "*.nc$", full.names = TRUE))

x
```

## Expected behavior

It would be helpful for at least the option to split variables up by PFT.  One way to do this to ensure backwards compatibility would be to add an argument `by_pft = FALSE` that changes nothing if kept default, but adds PFT as a dimension wherever possible when `TRUE`

Aariq:
Mainly tackling https://github.com/PecanProject/pecan/issues/2997.  Likely fixing other issues and changing some behaviors along the way.

## Description
- renamed and deprecated `begins`, `ends`, and `out` args in `put_E_values()` and `put_T_values()`
- Read in additional ED2 output variables in `read_E_files()`
- Behavior of `read_E_files()` changed somewhat significantly:  previously, variables were **weighted** by plant density in each cohort.  This weighting has been removed.  Instead, any variables in per-plant units are now **converted** to per area units using plant density. 
- add unit conversions to PEcAn standard
- fix bug/incorrect documentation of `out` argument in `put_E_values()`
- Revert changes to the `pft` argument for `model2netcdf.ED2()` and the `read_*_files()` functions to be a named numeric vector instead of the settings$pfts list.  If `settings` is supplied, new helper function `extract_pfts(settings)` will return a named vector.
- Document functions
- changed message about failed runs from "info" to "warning" (indicates problems since ensemble runs may not be failing at random)
- changed `put_E_values()` and `put_T_values()` to unexported.
- Added tests

## Motivation and Context

Only a small number of variables were being read from monthly E files and put into the .nc files output by `model2netcdf.ED()` and they were not being converted to PEcAn standard.  This PR adds more variables to that list and hopefully establishes some generalizable patterns for adding more variables in the future.

Closes #2997 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


Aariq:
Just for the record, I think it's pretty hard to work on this group of functions because `read_*_files()` does some conversions and `put_*_values()` does additional conversions and adds the units.  I think it would be a lot simpler and easier to work on if the "read" and "put" steps were combined, but that would be a much bigger re-write than I'm already doing.

Aariq:
Ok, all tests are passing.  I think the only thing left is to name the pft-level variables pulled from the monthly E files something standard.

models/ed/R/model2netcdf.ED2.R
```diff
@@ -166,8 +167,10 @@ model2netcdf.ED2 <- function(outdir, sitelat, sitelon, start_date,
     PEcAn.logger::logger.info("*** Writing netCDF file ***")
 
     out <- unlist(out_list, recursive = FALSE)
+    #create nc file with slots for all variables
     nc <- ncdf4::nc_create(file.path(outdir, paste(y, "nc", sep = ".")),
                            nc_var)
+    on.exit(ncdf4::nc_close(nc))
```
mdietze, line 173: Not sure about this design change over manually closing the files. Doesn't this delay the closing of the file until the exit of the function. Since this is in a loop, seems like we're opening connections to a lot of files at once, which seems inefficient and could lead to a "too many connections" error for a long timeseries. Also, since we're reusing `nc` in the loop, it seems like we're going to loose the connection to all but the last of the files opened.

models/ed/R/model2netcdf.ED2.R
```diff
@@ -1365,24 +1499,44 @@ read_S_files <- function(sfile, outdir, pfts, pecan_names = NULL, settings = NUL
...
+#' 
+#' A helper function to extract a named vector of pft numbers from
+#' `settings$pfts`.  Will use pft numbers in `settings` if they exist, otherwise
+#' it'll match using the `pftmapping` dataset
+#'
+#' @param pfts settings$pfts
+#'
+#' @return named numeric vector
+#'
+extract_pfts <- function(pfts) {
```
mdietze, line 1513: Not a required change, but this file has gotten long and unwieldy -- might be good to move some of these helper functions to either files of their own or a general file for model2netcdf.ED2 helpers.

models/ed/R/model2netcdf.ED2.R
Aariq: Thanks.  I think keeping these `on.exit()` lines will ensure the file gets closed if there is an error or interruption and they can co-exist with the manual closing of files within this loop.

Aariq: Nevermind, just tested that and it doesn't work like I expected.  I'll just revert this change for now then.

mdietze (approved):

Issue #3019, opened 2022-08-26
Some ensemble `run` directories don't get copied over to HPC


Aariq:
### Bug Description

When running an ED2 model, `start_model_runs()` sometimes fails to copy some ensembles in the run directory to a remote host (HPC). One possible reason for this this is that `rsync` is currently being run inside of a for-loop and maybe there are some limits to how many connections to the server are open or how often connections can be made.  It'll be more efficient to just `rsync` all the ensemble files over at once outside of a for-loop anyways, even if it doesn't fix this bug.

It's either happening here: 
https://github.com/PecanProject/pecan/blob/f5194f84b26a85bbfcb279cc4665d7560205efc7/base/workflow/R/start_model_runs.R#L97-L102

Or maybe here (can't remember)
https://github.com/PecanProject/pecan/blob/f5194f84b26a85bbfcb279cc4665d7560205efc7/base/workflow/R/start_model_runs.R#L132-L141

### To Reproduce

difficult to reproduce, sorry.

## Expected behavior

All files for ensemble runs should be copied over and if they can't be, there should be an informative warning or error.

PR #3025, opened 2022-08-31
Copy all run and out dirs at once, not in for-loop

Aariq:
Addresses #3019 with changes to `start_model_runs()`

## Description
- Moves creation of directories on the remote host outside of for loop to do it all at once
- Moves calls to `remote.copy.to` outside of for loop to copy over all /out and /run directories at once.
- Moves call to `remote.copy.from` outside of for loop to copy all outputs back to local at once.
- Adds notes for future improvments
- Add informative errors to `remote.copy.to()` and `remote.copy.from()`

There aren't any tests for `start_model_runs()` and I'm not sure how to write them, but I did install this PR and test it with my own setup.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


dlebauer (approved):

base/workflow/R/start_model_runs.R
```diff
@@ -271,17 +279,10 @@ start_model_runs <- function(settings, write = TRUE, stop.on.error = TRUE) {
...
-            dst = settings$modeloutdir)
-        }
-        
+      
         # TODO check output log
         if (is_rabbitmq) {
+          #TODO: unclear to me if this path is local or remote.  If local, then needs to be moved outside of for loop after files are copied back to local.
           data <- readLines(file.path(jobids[run], "rabbitmq.out"))
           if (data[-1] == "ERROR") {
             msg <- paste("Run", run, "has an ERROR executing")
```
Aariq, line 288: @robkooper Do you have the ability to check if this still works?  I don't have the docker setup with rabbitmq on my machine and I can't tell if I've broken this part by pulling lines 274-282 outside of this for-loop.

base/workflow/R/start_model_runs.R
```diff
@@ -271,17 +279,10 @@ start_model_runs <- function(settings, write = TRUE, stop.on.error = TRUE) {
       }
       
       if (job_finished) {
-
```
robkooper, line 274: this loop is needed since the file rabbitmq.out is generated on the remote side. see https://github.com/PecanProject/pecan/blob/master/docker/executor/executor.py#L88

One option is to only copy that one file (small) and at the end copy all files.

robkooper (changes requested):

base/workflow/R/start_model_runs.R
Aariq: What if we did this with `remote.execute.cmd()` instead?  I'm not actually sure which would be best for performance, but I suspect that would be more light weight than copying the log file.

base/workflow/R/start_model_runs.R
Aariq: What if we did this with `remote.execute.cmd()` instead? I'm not actually sure which would be best for performance, but I suspect that would be more light-weight than copying the log file.

The thing I'm confused about is if the path to `rabbitmq.out` is really `"/<jobid>/rabbitmq.out"`.  Shouldn't it be in `settings$modeloutdir` somewhere?

Aariq: If you could check and tell me where rabbitmq.out ends up relative to `settings$outdir` or `settings$modeloutdir` that would be super helpful.

base/workflow/R/start_model_runs.R
Aariq: I'm not sure this is correct.  See line 265/273 above:
```
if (is_rabbitmq) {
        job_finished <- file.exists(file.path(jobids[run], "rabbitmq.out"))
```
This was always happening before the files get copied back over.  It's the same path referenced in L285/286, so it must be a file that exists before `remote.copy.from()` runs, right?

Aariq:
Not sure if the `--exclude` flag is working as expected for just copying over log files

Aariq:
Ok, seems to be working consistently now.  Running two remote commands in a row seemed to be causing problems, but a 1 second pause seems to fix it. 🤷‍♂️

Aariq:
Now that I've been using this for a while I've noticed that it _still_ fails to copy over all the `run/` files sometimes despite `rsync` reporting "success".  This definitely works better than it did previously, but I can also continue to try to figure this out.

Aariq:
Ok, the remaining bug has nothing to do with PEcAn or with R for that matter.  I've been able to reproduce it with just `rsync`.  It also seems to be specific to our server as I can't reproduce it from my laptop.  Going to open this back up for review.  One possible safeguard is to wrap the `remote.copy.to` commands in `PEcAn.utils::retry.func()` but I'll wait to get feedback before doing this.

base/workflow/R/start_model_runs.R
```diff
@@ -195,14 +209,17 @@ start_model_runs <- function(settings, write = TRUE, stop.on.error = TRUE) {
...
-            host = settings$host,
-            src = file.path(settings$rundir, format(run, scientific = FALSE)),
-            dst = settings$host$rundir,
-            delete = TRUE)
+          PEcAn.utils::retry.func(
+            PEcAn.remote::remote.copy.to(
+              host = settings$host,
+              src = file.path(settings$rundir, format(run, scientific = FALSE)),
+              dst = settings$host$rundir,
+              delete = TRUE),
```
Aariq, line 220: Does this need to be `delete == TRUE`

mdietze:
@robkooper could you look at the changes you requested so we can figure out whether this is ready to pull?

base/remote/R/remote.copy.from.R
```diff
@@ -45,5 +45,37 @@ remote.copy.from <- function(host, src, dst, options = NULL, delete = FALSE, std
     }
   }
   PEcAn.logger::logger.debug("rsync", shQuote(args))
-  system2("rsync", shQuote(args), stdout = TRUE, stderr = as.logical(stderr))
+  out <-
+    system2("rsync", args, stdout = "", stderr = as.logical(stderr))
```
infotroph, line 49: is removing the `shQuote` around `args` intentional?

base/remote/R/remote.copy.from.R
```diff
@@ -45,5 +45,37 @@ remote.copy.from <- function(host, src, dst, options = NULL, delete = FALSE, std
...
+  
+  # Informative errors from rsync man page
```
infotroph, line 51: I appreciate the desire for clarity, but I'd rather send people to the rsync docs for these error messages. We use whatever rsync version is installed on the host system, so hard-coding values here feels like an invitation for them to get out of sync. 

I assume rsync doesn't often change the meaning of exit codes between versions, but if they add a new code that's not on this list then this switch will return an empty string instead of the numeric value.

base/workflow/R/start_model_runs.R
infotroph: No, `=` is correct -- we're passing a named argument to `remote.copy.to`, not performing a logical comparison.

base/remote/R/remote.copy.from.R
Aariq: Yes, it was intentional, but now that you bring it up I'm not sure it's correct.  `system2()` always `shQuote`s `command`, but I guess it doesn't do that for `args`.  I'll add it back and make sure everything still works.

robkooper (approved):
tested with rabbitmq. please address issues pointed out by @infotroph

base/remote/R/remote.copy.from.R
Aariq: Yeah, this makes sense.  I'm going to keep errors (anything other than 0) as `logger.severe()` though.

infotroph (approved):

PR #3026, opened 2022-08-31
Shifted convert.input.R from utils to DB

nanu1605:
To move `convert.input` from `utils` to `DB`

This replaces #3009, which accumulated enough conflicts that it was easier to redo the changes in a new branch.


nanu1605:
Copying the checklist we discussed in Slack here for easy reference:

- [x] check whether convert.input uses any unexported functions from PEcAn.utils. If so, see where else they're used and decide whether to move them to data.atm, copy them, or export them from utils
- [x] As one commit, move the convert.input function definition and man page from PEcAn.utils to PEcAn.DB.
- [x] As a separate commit, update the convert.input code to remove PEcAn.DB:: in front of calls to DB functions and to add PEcAn.utils:: in front of calls to utils. It's important to do it in a separate commit because otherwise Git will just show one giant blob deleted from utils and another added to DB, with no easy way to find the differences between them. Doing it in two steps makes it pretty easy to verify that the move part didn't change anything inside the blob
- [x] Review the function documentation and make any edits needed — I doubt it will, but do check
- [ ] Implement any needed deprecation stub in PEcAn.utils. The usual best practice is to deprecate functions slowly over several releases to give users time to update, but the circular dependency demands quicker action. Decision needed: After this change, should a call to PEcAn.utils::convert.input() fail with the function not found or should it do nothing but throw a message to use the version in PEcAn.DB?
- [x] Update the changelog and both packages’ NEWS files
- [x] Update all calls to convert.inputs elsewhere in PEcAn
- [x] Update Rcheck_reference.log, if needed, in all the packages touched by this PR

base/db/R/convert_input.R
```diff
@@ -0,0 +1,891 @@
+##' Convert between formats, reusing existing files where possible
```
mdietze, line 1: this looks like an existing file that should have been moved via `git mv` so as to retain version history

mdietze (changes requested):

base/db/R/convert_input.R
nanu1605: @mdietze when I try to do it using `git mv` it says `fatal: not under version control, source=base/utils/R/convert_input.R, destination=base/db/R/convert_input.R`

moki1202: @nanu1605 make sure you use the right case in the paths.

infotroph: @mdietze Rebased to address this -- long story short it was less about `mv` vs add/`rm` and more that the old filename was retained with new content. Placing the defunct stub in a separate file allowed Git to detect the rename correctly.

mdietze (approved):

infotroph (approved):

PR #3033, opened 2022-09-13
set defaults for new ED2IN vars

Aariq:
## Description

Sets some arbitrary defaults for new variables in ED2IN file. These are required for the ED2IN file to be parsed, but I don't think they actually are used.  The ED2IN template says "These variables are used to define the additional soil properties needed for SOIL_HYDRO_SCHEME=2." and in this ED2IN file I've set `SOIL_HYDRO_SCHEME=0`.

Defaults came from googling the description of the variable with the word "typical" and "soil". 

Also fixes a mistake in #3035 that added new variables not in the ED2IN template *after* the `$END` tag.

## Motivation and Context

Required variables for development version of ED2 to run.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

Extremely unlikely to break anything, and I can't test if it works until the PR gets merged and the docker container gets built by GitHub actions.

## Checklist:


Aariq:
Installed on Welsh server and tested with development version of ED2.  Both supplying the new template and supplying new tags with pecan.xml work.

dlebauer (approved):

models/ed/inst/ED2IN.r2.2.0.github
```diff
@@ -599,11 +599,13 @@ $ED_NL
...
+   ! SOIL_HYDRO_SCHEME=0                                                                   !
    !---------------------------------------------------------------------------------------!
-   NL%SLSOC   = myslsoc
-   NL%SLPH    = myslph
-   NL%SLCEC   = myslcec
-   NL%SLDBD   = mysldbd
+   NL%SLSOC   = 0.06
+   NL%SLPH    = 6
+   NL%SLCEC   = 0.5
+   NL%SLDBD   = 1330
```
mdietze, line 608: Not sure I understand why you switch out variables for hard-coded constants. Doesn't this eliminate the ability for PEcAn to set these variables?

mdietze (changes requested):

models/ed/inst/ED2IN.r2.2.0.github
Aariq: These are new variables that didn't exist in previous versions of ED2IN.  I assumed I would need knowledge of soil chemistry to connect these up to PEcAn in some way, but if that's not the case I can take a look at how other ED2IN variables are done and take a crack at this. Also, I'm pretty sure they aren't used unless another hard-coded variable is changed.

mdietze: So if the `mysl*` variables don't exist anywhere else in the code I'm OK with this PR, though I do think the soil.nc files do have most of these variables, if one wanted to set up the ability to specify these via code.

mdietze (approved):

models/ed/inst/ED2IN.r2.2.0.github
Aariq: I'll open an issue as a reminder to add this in the future

PR #3040, opened 2022-09-15
Add script that extracts Soilgrids SOC based on location of sites

Qianyuxuan:
## Description

Add a script that extracts total soil organic carbon and its standard deviation based on user-defined site location from SoilGrids250m version 2.0: https://soilgrids.org
## Motivation and Context

To provide SOC observations for SDA

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
+###Extraction of total soil organic carbon based on user-defined site location from SoilGrids250m version 2.0 : https://soilgrids.org
+###Input parameter: user-defined site longitude/latitude 
+###Extract variable ocd (organic carbon density)
+###Example:ocs<-ocs_extract(c(-77.904, -119.1944, -121.557), c(40.6658, 37.0678, 44.4524))
+
+rm(list = ls())
+library(terra)
+library(reshape2)
+library(dplyr)
```
serbinsh, line 9: For the final version of the function we put into data.land, we will need to populate the header with documentation.  you can look at other pecan functions to see how to do this.

You will also need to remove the calls to loading these libraries and make sure the dependent functions are part of the DESCRIPTION and NAMESPACE

mdietze: Clarification: the NAMESPACE file should not be edited manually. After you document your functions using ROxygen syntax you should run `make document` and commit the updated version of the NAMESPACE and any Rd documentation files that are generated.

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
...
+  ocs_sum <- mean_site[,1]*(5-0)*0.01+mean_site[,2]*(15-5)*0.01+mean_site[,3]*(30-15)*0.01+mean_site[,4]*(60-30)*0.01+mean_site[,5]*(100-60)*0.01+mean_site[,6]*(200-100)*0.01 
+  #calculate standard deviation of ocs as the square root of sum of variance of layer-specific ocs, the unit of ocs is kg/m2, based on Eq. (8) in paper https://www.sciencedirect.com/science/article/pii/S2215016122000462, except the correlation term due to the lack of information 
+  ocs_std <- sqrt((std_site[,1]*(5-0)*0.01)^2+(std_site[,2]*(15-5)*0.01)^2+(std_site[,3]*(30-15)*0.01)^2+(std_site[,4]*(60-30)*0.01)^2+(std_site[,5]*(100-60)*0.01)^2+(std_site[,6]*(200-100)*0.01)^2)
+  
+  if (!is.null(outdir)) {
+    PEcAn.logger::logger.info(paste0("Storing results in: ",file.path(outdir,"soilgrids_soc_data.RData")))
+    if (! file.exists(outdir)) dir.create(outdir,recursive=TRUE)
+    soilgrids_soc_data <- list("Total_soc" = ocs_sum, "Sdev_soc" = ocs_std)
+    save(soilgrids_soc_data, file = file.path(outdir,"soilgrids_soc_data.RData"))
```
mdietze, line 144: Given the size and simplicity of this object, why not just save it as a (more portable) CSV?

mdietze (changes requested):

modules/data.land/R/soilgrids_soc_extraction.R
serbinsh: To keep things moving I am going to post up here now a quick edit with docs etc to show you what we mean. We will need to still update the code to accept a dataframe of siteID, site name, lat, long, for the function. if we dont use BETY we can just show how site ID can be generic or NA perhaps but if we are using BETY sites that should be included as well as the site name

modules/data.land/R/soilgrids_soc_extraction.R
serbinsh: OK, thanks fine to do.

Again it needs to be: Site_ID, Site_Name, Latitude, Longitude, Total_OCS SD_OCS 

Or something like that for both the memory output and the CSV.  Its better if the result is a cleaner format that we can then transform into what we need

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
...
+    ocd_Q0.95_real <- ocd_Q0.95_map[, 2] / 10
+    
+    ocdquant[dep, ] <-c(ocd_mean_real,ocd_Q0.05_real,ocd_Q0.50_real,ocd_Q0.95_real)
+    siteid [dep, ] <- rep(1:length(lon_input), 4)
+
+    ### Display progress to console
+    if (verbose) {
+      setTxtProgressBar(pb, j)
+      j <- j+1
+      flush.console()}
```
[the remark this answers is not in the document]
serbinsh: I added the namespace

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
...
+    close(pb)
+  }
+  
+  # parse extracted data and prepare for output
+  ocd_fit <- melt(ocdquant, id.vars = c("Mean"))
+  id_fit <- melt(siteid, id.vars = c("Mean"))
+  colnames(ocd_fit) <- c("Depth", "Quantile", "Value")
+  ocd_fit$Variable <- rep("ocd", length(nrow(ocd_fit)))
+  ocd_fit$siteid <- id_fit$value
+  dat <- split(ocd_fit, list(ocd_fit$siteid, ocd_fit$Depth))
```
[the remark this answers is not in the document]
serbinsh: I added the namespace

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
...
+  mean_site <- matrix(mean, length(lon_input), 6)
+  rownames(mean_site) <- paste0("Site_", 1:length(lon_input))
+  colnames(mean_site) <- c("0-5cm",
+                           "5-15cm",
+                           "15-30cm",
+                           "30-60cm",
+                           "60-100cm",
+                           "100-200cm")
+  std_site <- matrix(std, length(lon_input), 6)
+  rownames(std_site) <- paste0("Site_", 1:length(lon_input))
```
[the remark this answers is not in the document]
serbinsh: Agree

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
...
+###Input parameter: user-defined site longitude/latitude 
+###Extract variable ocd (organic carbon density)
+###Example:ocs<-ocs_extract(c(-77.904, -119.1944, -121.557), c(40.6658, 37.0678, 44.4524))
+
+rm(list = ls())
+library(terra)
+library(reshape2)
+library(dplyr)
+
+ocs_extract <- function (lon_input, lat_input, outdir=NULL,verbose=TRUE) {
```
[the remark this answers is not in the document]
serbinsh: Thats a good suggestion.  @Qianyuxuan if you change make sure you also edit the documentation at the top of the function and then remember to do ```make document``` after to update the docs

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+##' 
+##' @return a dataframe containing the total soil carbon values  
+##' and the corresponding standard deviation values (uncertainties) for each location 
+##' Output column names are c("Site_ID","Site_Name","Latitude","Longitude",
+##' "Total_soilC","Std_soilC")
+##' 
+##' @export
+##' @author Qianyu Li, Shawn P. Serbin
+##' 
+soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE, ...) {
```
mdietze, line 59: ditto to my comment on your other PR. what is the `...` and why isn't it documented

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+##' site_info <- qry_results.2
+##' verbose <- TRUE
+##' system.time(result_soc <- soilgrids_soilC_extract(site_info=site_info, verbose=verbose))
+##' result_soc
+##' 
+##' }
+##' @importFrom reshape2 melt
+##' @importFrom terra vect project rast extract
+##' @importFrom stats qgamma optim
+##' @importFrom utils flush.console setTxtProgressBar txtProgressBar
```
mdietze, line 49: This is OK to do, but in general the PEcAn norm is to use the namespaces in the code itself rather than using @importFrom, which we generally restrict for instances where namespaces don't make sense (e.g. extended tidyverse pipes)

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,149 @@
...
+  # reproject locations to soilgrids projection
+  #Soilgrids data is using Homolosine projection https://www.isric.org/explore/soilgrids/faq-soilgrids 
+  p <- terra::vect(lonlat, crs = "+proj=longlat +datum=WGS84") # Users need to provide lon/lat
+  newcrs <- "+proj=igh +datum=WGS84 +no_defs +towgs84=0,0,0" 
+  p_reproj <- terra::project(p, newcrs) # Transform the point vector to data with Homolosine projection
+  
+  # setup progress bar
+  if (verbose) {
+    j <- 1
+    pb <- txtProgressBar(min = 0, max = length(depths), char="*", width=70, style = 3)
```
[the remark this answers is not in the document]
serbinsh: I thought I already fixed this?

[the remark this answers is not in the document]
serbinsh: Oh I did - this is outdated

modules/data.land/R/soilgrids_soc_extraction.R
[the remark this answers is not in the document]
serbinsh: Actually it looks like we also need to do

```utils::flush.console()```

File: modules/data.land/R/soilgrids_soc_extraction.R (hunk 6 above)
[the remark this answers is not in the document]
Comment (serbinsh) [reply modules/data.land/R/soilgrids_soc_extraction.R]: @mdietze are you ok with the new function name

"soilgrids_soilC_extract"?

File: modules/data.land/NAMESPACE [hunk 10]
```diff
@@ -49,12 +49,23 @@ export(soil.units)
...
 importFrom(rlang,.data)
+importFrom(stats,optim)
+importFrom(stats,qgamma)
+importFrom(terra,extract)
+importFrom(terra,project)
+importFrom(terra,rast)
+importFrom(terra,vect)
+importFrom(utils,flush.console)
+importFrom(utils,setTxtProgressBar)
+importFrom(utils,txtProgressBar)
```
infotroph, line 71: PEcAn house style is to call all of these directly from their namespaces, e.g. `stats::optim(...)` rather than using `importFrom`. The exceptions are cases where the `::` notation doesn't work at all, e.g. `%>%` or `.data`.

modules/data.land/R/soilgrids_soc_extraction.R
infotroph: I see that Mike beat me to my comment above, but I'll add that the norm is strong enough I'd like these importFrom's removed and the namespaces added to each call before merging.

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+##' Output column names are c("Site_ID","Site_Name","Latitude","Longitude",
+##' "Total_soilC","Std_soilC")
+##' 
+##' @export
+##' @author Qianyu Li, Shawn P. Serbin
+##' 
+soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE, ...) {
+
+    if (is.null(site_info)) {
+    stop("Please provide a BETY DB site list containing at least the site id and PostGIS geometry\
```
infotroph, line 62: I recommend making this a PEcAn.logger::logger.error() call

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+  colnames(siteid) <- c(rep("Mean", length(internal_site_info$site_info.lon)),
+                        rep("0.05", length(internal_site_info$site_info.lon)),
+                        rep("0.5", length(internal_site_info$site_info.lon)),
+                        rep("0.95", length(internal_site_info$site_info.lon)))
+  if (verbose) {
+    close(pb)
+  }
+  
+  # parse extracted data and prepare for output
+  ocd_fit <- reshape2::melt(ocdquant, id.vars = c("Mean"))
```
infotroph, line 136: Low priority: How difficult would it be to rewrite these manipulations to use something like `dplyr::pivot_longer` instead of adding reshape2 to the dependency list? data.land has a beastly number of dependencies already, so when there's a choice I recommend preferring solutions that use existing imports rather than add new ones.

modules/data.land/R/soilgrids_soc_extraction.R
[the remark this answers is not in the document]
Qianyuxuan: The namespace was added.

Qianyuxuan:
Thanks Chris! Great suggestion! I did test and found out "dplyr::pivot_longer" works well to replace "reshape2::melt". I will update my PR soon.

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -110,38 +104,31 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
...
-  colnames(ocd_fit) <- c("Depth", "Quantile", "Value")
-  ocd_fit$Variable <- rep("ocd", length(nrow(ocd_fit)))
-  ocd_fit$siteid <- id_fit$value
-  f1<-factor(ocd_fit$siteid,levels=unique(ocd_fit$siteid))
-  f2<-factor(ocd_fit$Depth,levels=unique(ocd_fit$Depth))
+ # parse extracted data and prepare for output
+  quantile_name <-c(paste("Mean_",site_info$id,sep=""),paste("0.05_",site_info$id,sep=""),paste("0.5_",site_info$id,sep=""),paste("0.95_",site_info$id,sep=""))
+  colnames(ocdquant) <- quantile_name
+  ocdquant_dep <- cbind(ocdquant,depths)
+  ocdquant_long <- tidyr::pivot_longer(as.data.frame(ocdquant_dep),cols=all_of(quantile_name))
```
mdietze, line 124:
```suggestion
  ocdquant_long <- tidyr::pivot_longer(as.data.frame(ocdquant_dep),cols=tidyr::all_of(quantile_name))
```

serbinsh:
Looks like reshape2 is indeed no longer a depends

```
[sserbin@modex data.land]$ grep -ri "reshape" .
./inst/FIA_allometry/README.Rmd:# Note: the "reshape" package is currently required, although this may
./inst/FIA_allometry/README.Rmd:library(reshape)
./inst/FIA_allometry/allometry.r:# Note: the "reshape" package is currently required, although this may
./inst/FIA_allometry/allometry.r:  #install.packages("reshape")
./inst/FIA_allometry/allometry.r:library(reshape)
./inst/FIA_allometry/biomass.r:# Note: the "reshape" package is currently required, although this may
./inst/FIA_allometry/biomass.r:library(reshape)
./DESCRIPTION:    reshape2,
```

but we still have some reshape requirements that could be updated at some point

serbinsh:
> @Qianyuxuan here are the most recent build errors.
> 
> ```
>  checking package dependencies ... NOTE: Imports includes 33 non-default packages.
>   checking R code for possible problems ... NOTE: soilgrids_soilC_extract: no visible binding for global variable 'name'
>   checking R code for possible problems ... NOTE: soilgrids_soilC_extract : cgamma: no visible global function definition for 'qgamma'
>   checking R code for possible problems ... NOTE: soilgrids_soilC_extract : fitQ: no visible global function definition for 'optim'
> ```



> @Qianyuxuan here are the most recent build errors.
> 
> ```
>  checking package dependencies ... NOTE: Imports includes 33 non-default packages.
>   checking R code for possible problems ... NOTE: soilgrids_soilC_extract: no visible binding for global variable 'name'
>   checking R code for possible problems ... NOTE: soilgrids_soilC_extract : cgamma: no visible global function definition for 'qgamma'
>   checking R code for possible problems ... NOTE: soilgrids_soilC_extract : fitQ: no visible global function definition for 'optim'
> ```

So "qgamma" is an R base function yes, same with optim yes (well its stats::optim).  Not sure what "name" is referring too

serbinsh:
@mdietze in the current errors, I dont understand what check is referring to with 

```
NOTE: soilgrids_soilC_extract: no visible binding for global variable 'name'
```

All I could find was in the docs

##' @name soilgrids_soilC_extract

Is that what its referring too?

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,199 @@
...
+  if (verbose) {
+    close(pb)
+  }
+  
+ # parse extracted data and prepare for output
+  quantile_name <-c(paste("Mean_",site_info$id,sep=""),paste("0.05_",site_info$id,sep=""),paste("0.5_",site_info$id,sep=""),paste("0.95_",site_info$id,sep=""))
+  colnames(ocdquant) <- quantile_name
+  ocdquant_dep <- cbind(ocdquant,depths)
+  ocdquant_long <- tidyr::pivot_longer(as.data.frame(ocdquant_dep),cols=tidyselect::all_of(quantile_name))
+  ocd_df <- tidyr::separate(ocdquant_long,name, c('Quantile', 'siteid'),"_")
```
mdietze, line 125: @Qianyuxuan @serbinsh this is where the undefined `name` is coming from.

Qianyuxuan: "name" is a default column name after using "pivot_longer". I guess it should be fixed with an assigned name. I will try it.

mdietze: @Qianyuxuan the issue isn't that the name is default, it's that because it's a column in the dataframe it's not defined outside of the dataframe. The solution is to add
```
##' @importFrom rlang .data
```
to the Roxygen for this function and then, in that line, to refer to the variable as `.data$name` instead of just `name`. Here .data is a reference to the dataframe being passed through the pipe. Normally this is implicit, but R's code checking forces us to make this explicit. This is one of rare cases where it actually does make more sense to do the importFrom at the function level, rather than using namespaces.

Qianyuxuan: @mdietze Just saw your comments. But I used another way to avoid the issue. You can test it to see if it is fixed.

infotroph:
Thanks! When I made the suggestion I may possibly have been forgetting that `pivot_longer` lived in tidyr instead of dplyr and therefore still meant adding a new dependency, but I like this change anyway :)

mdietze (approved):

PR #3049, opened 2022-09-27
Update of Soilgrids extraction codes with new documents and return format

Qianyuxuan:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+##' site_info <- qry_results.2
+##' verbose <- TRUE
+##' system.time(result_soc <- soilgrids_soilC_extract(site_info=site_info, verbose=verbose))
+##' result_soc
+##' 
+##' }
+##' @importFrom reshape2 melt
+##' @importFrom terra vect project rast extract
+##' @importFrom stats qgamma optim
+##' @importFrom utils flush.console setTxtProgressBar txtProgressBar
```
mdietze, line 49: This is OK to do, but in general the PEcAn norm is to use the namespaces in the code itself rather than using `@importFrom`, which we generally restrict for instances where namespaces don't make sense (e.g. extended tidyverse pipes)

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+##' 
+##' @return a dataframe containing the total soil carbon values  
+##' and the corresponding standard deviation values (uncertainties) for each location 
+##' Output column names are c("Site_ID","Site_Name","Latitude","Longitude",
+##' "Total_soilC","Std_soilC")
+##' 
+##' @export
+##' @author Qianyu Li, Shawn P. Serbin
+##' 
+soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE, ...) {
```
mdietze, line 59: what is the `...` here? It doesn't seem to be documented in the `@param` section, which will probably cause an error

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -0,0 +1,209 @@
...
+    
+    ocdquant[dep, ] <-c(ocd_mean_real,ocd_Q0.05_real,ocd_Q0.50_real,ocd_Q0.95_real)
+    ### Display progress to console
+    if (verbose) {
+      utils::setTxtProgressBar(pb, j)
+      j <- j+1
+      flush.console()}
+  
+  # cleanup interim results
+    rm(ocd_mean.url, ocd_Q0.05.url, ocd_Q0.50.url, ocd_Q0.95.url,
```
mdietze, line 116: This deallocation of memory, and reallocation at the start of the loop, probably slows down your code

modules/data.land/R/soilgrids_soc_extraction.R
Qianyuxuan: Okay! They occurred after the "make document" command. How can I get rid if them?

modules/data.land/R/soilgrids_soc_extraction.R
Qianyuxuan: Will delete it.

modules/data.land/R/soilgrids_soc_extraction.R
infotroph: These lines we're commenting on (in the Roxygen comment block) are the source of the ones that appeared after running `make`. If you delete these and then rerun `make document`, the corresponding `importFrom` lines in NAMESPACE should disappear.

Issue #2969, opened 2022-07-19
Allow for comments in pecan.xml by stripping them in `read.settings()`


Aariq:
Currently comments in `pecan.xml` (see example below) don't get stripped and can cause problems. In this example, the comment in the `<ed2in_tags>` ends up as a length 0 element in the settings list named `comment` which causes `write.settings.ED2` to error with `Error in vapply(custom_tags, function(x) grepl(numvec_rxp, x), logical(1)) : 
  values must be length 1,`
  
``` xml
<ed2in_tags>
      <PFT_1ST_CHECK>0</PFT_1ST_CHECK>
      <IED_INIT_MODE>0</IED_INIT_MODE>
      <!-- turn off cohort and patch fusion. Not necessary for short runs -->
      <MAXCOHORT>0</MAXCOHORT>
      <MAXPATCH>0</MAXPATCH>
    </ed2in_tags>
```

Here's a possible pattern to strip comments using the `xml2` package:

``` r
tt <-
  '<x>
     <a>text</a>
     <b foo="1"/>
     <!-- a comment -->
     <c bar="me">
        <d>a phrase</d>
        <!-- deeper comment -->
     </c>
  </x>'
xml <- xml2::read_xml(tt)
xml
#> {xml_document}
#> <x>
#> [1] <a>text</a>
#> [2] <b foo="1"/>
#> [3] <c bar="me">\n  <d>a phrase</d>\n  <!-- deeper comment -->\n</c>
# https://stackoverflow.com/questions/784745/accessing-comments-in-xml-using-xpath
comments <- xml2::xml_find_all(xml, "//comment()")
xml2::xml_remove(comments)
xml
#> {xml_document}
#> <x>
#> [1] <a>text</a>
#> [2] <b foo="1"/>
#> [3] <c bar="me">\n  <d>a phrase</d>\n</c>
```

I know pecan currently uses `XML`, but it might be a good opportunity to start the switch to `xml2` since `XML` is [no longer maintained](https://devguide.ropensci.org/building.html#recommended-scaffolding).

<details>
  <summary>full pecan.xml</summary>

``` xml

<?xml version="1.0"?>
<pecan>
  <info>
    <!--Edit this-->
    <notes>site name</notes>
  </info>
  <!-- Edit the outdir name to replace SN with the two-letter abreviation for this site -->  
  <outdir>/data/tests/ed2_transect_SN</outdir>
  <database>
    <bety>
      <driver>PostgreSQL</driver>
[...]

PR #3051, opened 2022-10-04
Allow HTML comments in pecan.xml and warn when named .xml doesn't exist

### LINKED ISSUE #2950: read.settings doesn't error when inputfile doesn't exist

Aariq:
Closes #2950 and #2969.

## Description

Uses a recursive function to strip any list elements named "comment" from a parsed pecan.xml.  This allows the use of HTML style comments (`<!-- a comment -->`) in the pecan.xml so annotated version can be used without breaking runs.  It also adds a warning to `read.settings()` when defaulting to "pecan.xml" when a named inputfile is provided and not found.

Another minor addition is that `logger.setLevel()` now returns (invisibly) the previous level.  This behavior is similar to how `par()` works and is useful for temporarily changing the logger level and then changing it back.

## Motivation and Context

I think adding comments in a pecan.xml file is a great way to give myself reminders about what needs to be changed and what needs to stay the same when preparing a series of runs based on a template.  This way, I can leave those comments in for others to see and don't have to remember to delete them all.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/settings/R/read.settings.R
```diff
@@ -97,3 +100,39 @@ read.settings <- function(inputfile = "pecan.xml") {
...
+      return(max(unlist(
+        lapply(this, depth, thisdepth = thisdepth + 1)
+      )))
+    }
+  }
+  
+  thisdepth <- depth(x)
+  if (thisdepth == 0) {
+    return(x)
+  } else if (length(nameIndex <- which(names(x) == "comment"))) {
```
infotroph, line 133: Matter of style, but I’d hoist the assignment to nameIndex up to between current  lines 130 and 131

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ see if you need to change any of these:
 - Added a new function `unit_is_parseable` in PEcAn.utils to replace `udunits2::ud.is.parseable`.
   (#3002; @nanu1605)
 - Initial LDNDC model coupling
+- `PEcAn.settings::read.settings()` now strips comments so HTML style comments (e.g. `<!-- a comment -->`) are now allowed in pecan.xml files
+- `PEcAn.logger::setLevel()` now invisibly returns the previously set logger level
```
infotroph, line 34: Can you add this to base/logger/NEWS.md as well? Sorry for the multiple scattered changelogs...

Aariq: Should I bump the version too?

infotroph: Yes, thanks!

base/logger/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# PEcAn.logger (development version)
```
infotroph, line 1:
```suggestion
# PEcAn.logger 1.8.1.9000
```

Push back if you like yours better, but I've been doing other NEWS files this way -- I have a vague memory there was some package-build-time magic that required the top heading to to match the package version. Don't remember what the magic was or what would break if the version was wrong, though...

PR #3052, opened 2022-10-06
Sda dongchen

DongchenZ:
Adding template job files specifically for SCC server in BU, and solved several issues so we can not export full year nc files on geo.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


models/template/inst/template_geo.job
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash -l
+
+# redirect output
+exec 3>&1
+exec &> "@OUTDIR@/logfile.txt"
+
+# host specific setup
+@HOST_SETUP@
+@CDO_SETUP@
```
serbinsh, line 9: Can you provide an example of what @CDO_SETUP@ should look like? is that the module command? the path?  both?

serbinsh: I also dont see a space in Create_Multi_settings.R where you put in the CDO options so that they get written to the XML. Does that info not go into the XML file?

DongchenZ: yeah, it should look like "module load cdo/2.0.6", it's SCC specific setup btw.

DongchenZ: good point, I am adding it right now.

DongchenZ: @serbinsh already updated the create_multi_settings script and it's now having the cdosetup in the host setction.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -55,6 +55,14 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     hostsetup <- paste(hostsetup, sep = "\n", paste(settings$host$prerun, collapse = "\n"))
   }
   
+  # create cdo specific settings
+  cdosetup <- ""
+  if (!is.null(settings$host$cdosetup)) {
+    cdosetup <- paste(cdosetup, sep = "\n", paste(settings$host$cdosetup, collapse = "\n"))
+    jobsh <- gsub("@CDO_SETUP@", cdosetup, jobsh)
```
mdietze, line 62: This statement needs to be outside of the if, otherwise this code will leave the `@CDOSETUP@` tag in the job.sh, which will cause the script to fail

models/template/inst/template_geo.job
mdietze: Dongchen, I see two things that need to happen before I can approve this PR:
1) The new tags you've added need to be added to the overall PEcAn documentation, including examples like the one in the thread above so that folks know how to use these new tag
2) These tags need to be added to the SIPNET job template, not the generic model template, for them to work. Furthermore, they need to be there in the default job template, not a geo specific one, otherwise your code will fail. Indeed, the problem you've identified and fixed is in no way specific to BU's geo cluster, only your example module load calls are.

mdietze (changes requested):

models/sipnet/R/write.configs.SIPNET.R
DongchenZ: solved

models/template/inst/template_geo.job
DongchenZ: solved the second point, but I am still working on how to add details into PEcAn documentation.

mdietze (approved):

PR #3058, opened 2022-10-20
find fqdn across platforms

### LINKED ISSUE #3044: Failing windows build of `PEcAn.ED2`: bug in `fqdn()`?
### Bug Description

The windows build of `PEcAn.ED2` on r-universe is [currently failing](https://github.com/r-universe/pecanproject/actions/runs/3094086051/jobs/5007429303) because of a failing test in test.met2model.R.  The only thing that makes sense to me that would cause this is if `fqdn()` returns NULL or a length 0 vector on Windows.  I don't have a Windows machine to do local testing on.  Tests for `fqdn()` would help debug this.

dlebauer:
## Description

Add `as.character(Sys.info()["nodename"])`  if no env variable FQDN and `hostname -f`  return empty strings.

## Motivation and Context

hostname doesn't work on Windows. Fixes #3044

**Perhaps a simpler option** 

replace function with `fqdn <- function() as.character(Sys.info()["nodename"])`

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/remote/R/fqdn.R
+++ b/base/remote/R/fqdn.R
@@ -19,5 +19,10 @@
 ##' @examples
 ##' fqdn()
 fqdn <- function() {
-  as.character(Sys.info()["nodename"])
+  if (Sys.getenv("FQDN") == "") {
+    fqdn <- as.character(Sys.info()["nodename"])
+  } else {
+    fqdn <- Sys.getenv("FQDN")
+  }
+  return(fqdn)
 } # fqdn
--- a/base/remote/R/fqdn.R
+++ b/base/remote/R/fqdn.R
@@ -19,12 +19,5 @@
 ##' @examples
 ##' fqdn()
 fqdn <- function() {
-  fqdn <- Sys.getenv("FQDN")
-  if (fqdn == "") {
-    fqdn <- system2("hostname", "-f", stdout = TRUE)
-  } 
-  if (fqdn == "") {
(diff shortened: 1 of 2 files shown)


base/remote/R/fqdn.R
```diff
@@ -19,9 +19,5 @@
 ##' @examples
 ##' fqdn()
 fqdn <- function() {
-  if (Sys.getenv("FQDN") != "") {
-    Sys.getenv("FQDN")
-  } else {
-    system2("hostname", "-f", stdout = TRUE)
-  }
+  as.character(Sys.info()["nodename"])
```
dlebauer, line 22: If there is a case when the environment variable FQDN should be checked first we could replace this with

```r
  fqdn <- Sys.getenv("FQDN")
  if (fqdn == "") {
    fqdn <- system2("hostname", "-f", stdout = TRUE)
  } 
  if (fqdn == "") {
    fqdn <- as.character(Sys.info()["nodename"])
  }
  return(fqdn)

robkooper: main reason for the FQDN is for an easy way to override the hostname, for example in the case of docker and kubernetes.

robkooper (changes requested):

base/remote/R/fqdn.R
dlebauer: OK now FQDN will override

PR #3071, opened 2022-11-21
Change from g to kg for SIPNET IC.process

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.land/R/cohort2pool.R
+++ b/modules/data.land/R/cohort2pool.R
@@ -70,7 +70,7 @@ cohort2pool <- function(dat, allom_param = NULL, dbh_name="DBH") {
       return(AGB)
     }
   }
-  #calculate total herbaceous biomass, already in gC
+  #calculate total herbaceous biomass, already in kgC
   tot_herb <- sum(dat[[1]][,"dryMass"])/(herb_plot*herb_num)
   
   #Calculate AGB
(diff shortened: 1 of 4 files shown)


modules/data.land/R/cohort2pool.R
```diff
@@ -83,12 +83,12 @@ cohort2pool <- function(dat, allom_param = NULL, dbh_name="DBH") {
   tot_leaf <- sum(leaf,na.rm = TRUE)
   
   #Divide by plot area, divide by 2 to convert from kg to kgC
-  leaf_biomass = ((tot_leaf/(total_area))/2 + tot_herb/1000)*1000#convert from kg to g
+  leaf_biomass = ((tot_leaf/(total_area))/2 + tot_herb)#in kg
```
mdietze, line 86: I think this fix (and the one on AGB) was done incorrectly. Specifically, I think tot_herb must have been in g. Line 73 changes the comment (already in gC) but doesn't change the units, so either the units need to be changed to kgC on line 74 or 86 and 91 need to be divided by 1000.

Finally, a reminder that explictly using PEcAn's unit conversion functions, rather than multiplying/dividing by hard-coded constants, remains best practices for the PEcAn code since it makes it much easier to see and understand unit conversions. It also reduces errors and inconsistances (e.g. different people using different numbers of significant digits during conversions)

modules/data.land/R/extract_NEON_veg.R
```diff
@@ -163,7 +164,7 @@ extract_NEON_veg <- function(lon, lat, start_date, end_date, store_dir, neonsite
         joined.soil$bulkDensity <- bulkDensity
         
         #convert from g/cm2 to g/m2, note that we have to divide by 100 because of percentage
-        joined.soil$SoilCarbon <- (joined.soil$organicCPercent * joined.soil$bulkDensity)*30*100 
+        joined.soil$SoilCarbon <- PEcAn.utils::ud_convert((joined.soil$organicCPercent * joined.soil$bulkDensity)*30*100, 'g m-2', 'kg m-2')#convert from g to kg.
```
mdietze, line 167: Comment says we have to divide by 100 but then the code multiplies by 100, which is confusing. Also unclear what the 30 is here.

modules/data.land/R/cohort2pool.R
DongchenZ: The tot_herb is in kg because I already converted it in the extract_NEON_veg function.

istfer:
Thanks @DongchenZ for looking into it so quickly. There is still `litterInit` left in write.configs.SIPNET in `g m-2` I believe, otherwise looks good to me

mdietze (approved):

PR #3075, opened 2022-12-12
Fix to job.sh model2netcdf line

Aariq:
## Description

With large numbers of PFTs the `model2netcdf.ED2()` line in job.sh would get cut off.  This is because `deparse()` has a default line width cutoff and would then split the output into multiple lines.  I've replaced it with `deparse1()` which always produces a character vector of length 1.

## Motivation and Context

The model2netcdf.ed2() step was failing for runs with more than a few PFTs 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/ed/inst/template.job
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -l
 
+# Generated by PEcAn.ED2::write.config.jobsh.ED2()
```
dlebauer, line 3:
```suggestion
# Parsed by PEcAn.ED2::write.config.jobsh.ED2() to generate job.sh files
```

models/ed/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.ED2
 Type: Package
 Title: PEcAn Package for Integration of ED2 Model
-Version: 1.7.2
+Version: 1.7.2.9000
```
dlebauer, line 4: minor detail, but why `*.9000` instead of `*.1` 

or `1.7.3`, which I think is appropriate way to increment a bug fix following https://semver.org/?

models/ed/tests/testthat/test.write.configs.ed.R
```diff
@@ -21,6 +21,7 @@ testdir <- tempfile()
 dir.create(testdir)
 withr::defer(unlink(testdir, recursive = TRUE))
 unzip("data/outdir.zip", exdir = testdir)
+# unzip("models/ed/tests/testthat/data/outdir.zip", exdir = testdir)
```
dlebauer, line 24: when is this used? for interactive debugging?

dlebauer (changes requested):
looks good. just few minor questions about bumping version and comments

models/ed/tests/testthat/test.write.configs.ed.R
Aariq: Exactly.  Tests are run with `testthat/` as the working directory rather than the project root.

models/ed/DESCRIPTION
Aariq: I'm used to bumping the version in pull requests with the development version (`usethis::use_version("dev")`) and only incrementing the patch or minor version for releases.  I think I noticed there are a couple other PEcAn packages with a version like 1.7.2.900x, so I figured maybe that was the way to go?

Aariq: From @infotroph on Slack:
>I like the tidyverse approach where releases are x.y.z, the dev version is always x.y.z.9000 with no increments between releases, and the next bump happens at release

infotroph (approved):

models/ed/NEWS.md
```diff
@@ -1,5 +1,6 @@
 # PEcAn.ED2 (development version)
 
+* Fixed a bug effecting the generation of job.sh for runs with many PFTs (#3075)
```
dlebauer, line 3:
```suggestion
* Fixed a bug affecting the generation of job.sh for runs with many PFTs (#3075)
```

dlebauer (approved):

PR #3076, opened 2022-12-14
Allow `model2netcdf.ED2()` to process partial output from failed runs

Aariq:
## Description

Adds an argument `process_partial` (default FALSE) to `model2netcdf.ED2()` that allows it to work even on partial runs that have been interrupted and not produced as many .h5 files as expected based on the dates in settings.

## Motivation and Context

Although ensembles of different lengths might screw up downstream analysis, sometimes you want to extract data from failed runs to inspect it.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/ed/DESCRIPTION
+++ b/models/ed/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: PEcAn.ED2
 Type: Package
 Title: PEcAn Package for Integration of ED2 Model
-Version: 1.7.2
+Version: 1.7.2.9000
 Date: 2021-10-04
 Authors@R: c(person("Mike", "Dietze", role = c("aut", "cre"),
                      email = "dietze@bu.edu"),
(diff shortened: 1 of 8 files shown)


models/ed/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.ED2
 Type: Package
 Title: PEcAn Package for Integration of ED2 Model
-Version: 1.7.2
+Version: 1.7.2.9000
```
dlebauer, line 4: same ? about versioning as #3075

dlebauer (approved):

models/ed/DESCRIPTION
Aariq: From @infotroph on slack:
>I like the tidyverse approach where releases are x.y.z, the dev version is always x.y.z.9000 with no increments between releases, and the next bump happens at release

Issue #3080, opened 2022-12-22
`write.config.xml.ED2()` doesn't use correct ED2 default dataset.


Aariq:
### Bug Description

Because the datasets `history`, `history.r46`, etc. are non-exported, `data()` doesn't work on them.  Current code to find the correct dataset based on the `<revision>` tag is therefore broken

https://github.com/PecanProject/pecan/blob/ab55731c300e7eb0d97b04fe30b97cb405dccba5/models/ed/R/write.configs.ed.R#L451

## Proposed solution

Internal data should be referenced with `packagename:::dataset`, so I'll replace the non-functioning code with a `switch()` statement like:

```
edhistory <- switch(revision, "85" = PEcAn.ED2:::history.r85, ...)
```

PR #3081, opened 2022-12-22
Bug fix for #3080 (write.config.xml.ED2() not using <revision> tag correctly)

Aariq:
## Description

`data()` can't be used for internal datasets.  The `history.csv` files are now stored as internal data and need to be accessed with `history`, `history.r81`, etc.  This PR fixes some code in `write.config.xml.ED2()` that wasn't working.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/ed/R/write.configs.ed.R
```diff
@@ -443,18 +443,22 @@ remove.config.ED2 <- function(main.outdir = settings$outdir, settings) {
...
-  histfile <- paste0("history.r", settings$model$revision) # set history file name to look for in ed2_package_data
-  if (histfile %in% ed2_package_data$results[, "Item"]) {
+  # TODO this should come from the database
+  
+  if(settings$model$revision %in% c("46", "81", "82", "85", "git")) {
+    histfile <- paste0("history.r", settings$model$revision)
     PEcAn.logger::logger.debug(paste0("--- Using ED2 History File: ", histfile))
-    data(list=histfile, package = 'PEcAn.ED2', envir = environment())
-    edhistory <- get(histfile)
+    edhistory <-
```
mdietze, line 451: This solution doesn't seem scalable -- if anyone adds a new ED version to the database they need to know to make manual modifications to this function.

Aariq: Yeah, that makes sense.  I'll see if I can make this more "programatic".  For what its worth, there aren't history files for every version of ED2 that there's an ED2IN file for, so it's already not scaling.  It's not clear to me how to update the history file for a new ED2 version.

Aariq: Ok, tried a different strategy and added some comments for why this code is there.

mdietze (approved):

PR #3083, opened 2023-01-05
[CI] Fix Docker action rate limits, hopefully

infotroph:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -45,8 +45,9 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
         with:
-          install: true
+          driver: docker
           buildkitd-flags: --debug
+          install: true
 
       # calculate some variables that are used later
       - name: get version tag
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -46,6 +46,7 @@ jobs:
         uses: docker/setup-buildx-action@v2
         with:
           install: true
+          buildkitd-flags: --debug
 
       # calculate some variables that are used later
       - name: get version tag
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -42,6 +42,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           set-safe-directory: false
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          install: true
 
       # calculate some variables that are used later
       - name: get version tag
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -77,7 +77,7 @@ jobs:
         env:
           PECAN_GIT_CHECKSUM: ${{ github.sha }}
           PECAN_GIT_BRANCH: ${GITHUB_BRANCH}
-          VERSION: ${PECAN_VERSION}
+          VERSION: ${{ env.PECAN_VERSION }}
 
       # push all images to github
       - name: Publish to GitHub
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,6 +29,7 @@ on:
 env:
   MASTER_REPO: PecanProject/pecan
   DOCKERHUB_ORG: pecan
+  GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   docker:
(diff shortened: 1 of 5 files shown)


mdietze (approved):

infotroph:
Questions for @robkooper:
* This change probably requires everyone to use BuildKit/buildx, including for local image builds. Is it safe to assume everyone building locally will have BuildKit available? Should that be documented somewhere?
* I find the `mount=type=secret,... export...cat... && ...` nonsense in the Dockerfiles to be very hard to read, and I'm worried it will make some users give up on understanding the build steps. Do you have ideas for how to make it more approachable? (Comments? Some kind of indirection to reduce the verbosity?) 
* Since the syntax is so verbose, I only passed the token to the RUN steps that actually need to use it to install R packages from GitHub. Would it be more or less confusing to use it everywhere? Do you have other thoughts on how to achieve this securely without making the build steps completely incomprehensible?

.github/workflows/docker.yml
```diff
@@ -41,6 +42,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           set-safe-directory: false
+      - name: Set up Docker Buildx
```
infotroph, line 45: Needed because the `--secret` syntax is only supported by BuildKit/buildx

docker.sh
```diff
@@ -121,6 +121,7 @@ echo "# ----------------------------------------------------------------------"
 if [ "${DEPEND}" == "build" ]; then
     ${DEBUG} docker build \
         --pull \
+        --secret id=github_token,env=GITHUB_PAT \
```
infotroph, line 124: I'd never have guessed what this does without reading the docs: "Take the env var `GITHUB_PAT` from the environment where the build is running, and make its contents available to the build, but not to the built image, as a file whose path will default to `/run/secrets/github_token`"

apps/api/Dockerfile
```diff
@@ -15,7 +15,9 @@ EXPOSE 8000
 # --------------------------------------------------------------------------
 
 # COMMAND TO RUN
-RUN   apt-get update \
+RUN  --mount=type=secret,id=github_token \
```
infotroph, line 18: "For the duration of this run step only, mount the secret named `github_token` (which we defined in the arguments to the `docker build` command) as a file whose path will default to `/run/secrets/github_token`"

apps/api/Dockerfile
```diff
@@ -15,7 +15,9 @@ EXPOSE 8000
...
+  export GITHUB_PAT=`cat /run/secrets/github_token` \
```
infotroph, line 19: Read the secret file back into an env var.
* Yes, this means we went from an env var in the environment surrounding the build to a file and back to an env var inside the build. If I'm reading the docs right, this is the only recommended way of passing build-time secrets around.
* I was expecting to be able to write `GITHUB_PAT=$(cat /run/secrets/github_token) apt-get update...`, but couldn't get it to work in local tests without the `export`. If y'all reviewers see a way to simplify this, please shout.

.github/workflows/docker.yml
```diff
@@ -72,7 +79,7 @@ jobs:
         env:
           PECAN_GIT_CHECKSUM: ${{ github.sha }}
           PECAN_GIT_BRANCH: ${GITHUB_BRANCH}
-          VERSION: ${PECAN_VERSION}
+          VERSION: ${{ env.PECAN_VERSION }}
```
robkooper, line 82: one option is for us take the docker.sh file and make multiple jobs (each bullet can be a matrix)
- depends
- base web docs
- models executor data thredds monitor rstudio-nginx
- dbsync
- all models
-

infotroph: @robkooper Can you say more about why? I'm guessing it's either:
a) If the build run as separate jobs, do they each get their own rate limit and let us avoid bothering to pass the PAT around?

robkooper: I'm partlly hoping that we might get around the dependencies issue, also it should speed up the build process since we can run things in parallel.

The depends image can be build only if the Dockerfile or pecan.depends.R file changes. I hope that we can also avoid running updates to packages, since that should have been captured in the pecan.depends.R file.

infotroph: @robkooper Here's a VERY off-the-cuff first stab at a matrix implementation, branched off this PR for experimenting: https://github.com/PecanProject/pecan/pull/3085

Will definitely fail at first because it doesn't handle order dependencies yet -- feel free to push your improvements straight to that branch.

infotroph: ...Actually I think #3805 is backwards: You said "each bullet can be a matrix", I made each bullet into an entry in one single matrix, so it dies on order dependencies.

PR #3096, opened 2023-01-25
moved robustly func to utils

### LINKED ISSUE #3073: Move `robustly()` to PEcAn.utils? Redundant of `retry.func()`?
`PEcAn.data.atmosphere::robustly()` and `PEcAn.utils::retry.func()` do almost the same thing.

`robustly()` is an adverb that returns a function (similar to `purrr::safely()`), while `retry.func()` returns the output of an expression (more like `try()`)

`robustly()` seems like it would be widely useful and should go in `PEcAn.utils` along side or replacing `retry.func()`

```
#' @param .f Function to call.
#' @param n Number of attempts to try
#' @param timeout Timeout between attempts, in seconds
#' @param silent Silence error messages?
#' @return Modified version of input function
robustly(.f, n = 10, timeout = 0.2, silent = TRUE)
```

```
#' @param expr The function to try running
#' @param maxErrors The number of times to retry the function
#' @param sleep How long to wait before retrying the function call
#' @param isError function to use for checking whether to try again.
#'   Must take one argument that contains the result of evaluating `expr`
#'   and return TRUE if another retry is needed 
#' @return retval returns the results of the function call
retry.func(
  expr,
  isError = function(x) inherits(x, "try-error"),
  maxErrors = 5,
  sleep = 0
)
```

meetagrawal09:
## Description
The PR moves `robustly` function to `utils` from `PEcAn.data.atmosphere`
This PR fixes #3073

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


meetagrawal09:
@mdietze can you please help me out with the failing tests and a review over the changes.

moki1202:
@meetagrawal09 Just want to check whether you manually moved the file or used `git mv`. I would highly recommend using the latter. Reviewers please correct me if I'm wrong here.

meetagrawal09:
> @meetagrawal09 Just want to check whether you manually moved the file or used `git mv`. I would highly recommend using the latter. Reviewers please correct me if I'm wrong here.

@moki1202 sir which file are we talking about here ? Is it the `robustly.Rd` file you are talking about ?

moki1202:
I'm talking about the robustly.R file that has been moved from `PEcAn.atmosphere` to `PEcAn.utils`.

moki1202:
> > @meetagrawal09 Just want to check whether you manually moved the file or used `git mv`. I would highly recommend using the latter. Reviewers please correct me if I'm wrong here.
> 
> @moki1202 sir which file are we talking about here ? Is it the `robustly.Rd` file you are talking about ? ( genetated again after moving the function using `roxygen2` ) And if you meant the `robustly.R` file then I added that function to `utils.R` ( thus no movement of file really was needed )

I 

> > @meetagrawal09 Just want to check whether you manually moved the file or used `git mv`. I would highly recommend using the latter. Reviewers please correct me if I'm wrong here.
> 
> @moki1202 sir which file are we talking about here ? Is it the `robustly.Rd` file you are talking about ? ( genetated again after moving the function using `roxygen2` ) And if you meant the `robustly.R` file then I added that function to `utils.R` ( thus no movement of file really was needed )

Removing the function entirely from the previous package would mean a little trouble for PEcAn users. A good practice here is to leave the deprecated version of the function at its original place and update the changelog file to let the users know so they don't have any troubles finding the function. Again I'm not a 100% sure about this so I'll leave it to @Aariq and @infotroph

moki1202:
Removing the function entirely from the previous package would mean a little trouble for `PEcAn` users. A good practice here is to leave the `deprecated version` of the function at its original place and update the `changelog` file to let the users know so they don't have any troubles finding the function. Again I'm not a 100% sure about this so I'll leave it to @Aariq and @infotroph

Aariq:
It sounds like it's not a good idea to move `robustly` to utils.  See: https://github.com/PecanProject/pecan/issues/3073#issuecomment-1405446745

infotroph:
Thanks for taking initiative on this, @meetagrawal09! Let's finish the discussion in #3073 before working more on this, but please do keep the PR open until there's a formal decision -- my comment there is only one opinion, and others may have counterpoints. I've been loudly wrong many times before on this project :)

meetagrawal09:
Sure, will keep the PR open until any formal decision.

base/utils/DESCRIPTION
```diff
@@ -57,5 +58,5 @@ License: BSD_3_clause + file LICENSE
 Copyright: Authors
 LazyData: true
 Encoding: UTF-8
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.3
```
mdietze, line 61: Roxygen version needs to be bumped down to 7.1.2

modules/data.atmosphere/DESCRIPTION
```diff
@@ -76,4 +76,4 @@ Copyright: Authors
 LazyLoad: yes
 LazyData: FALSE
 Encoding: UTF-8
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.3
```
mdietze, line 79: version also needs to be fixed here

mdietze (changes requested):

PR #3097, opened 2023-01-27
Replaced `transform` function

### LINKED ISSUE #2961: Replace `transform()` with something programming-friendly
>Actually, my advice would be to save this for a separate PR.  I did a quick search and there are **many** instances of `transform()` in PEcAn which will all likely result in a similar problem of undefined global variables.  It would probably easier to change them all at once when you've got the fix pattern figured out.

_Originally posted by @Aariq in https://github.com/PecanProject/pecan/pull/2956#discussion_r921198756_

`transform()` is similar to `dplyr::mutate()` but doesn't have a way of avoiding the 'undefined global variables' CRAN check error.  The help file for `transform()` explicitly says to not use it in programming.

For example, `transform(iris, Sepal.Length = Sepal.Length * 2)` will result in an undefined global variable `Sepal.Length`.  With `dplyr`'s `.data` pronoun this can be avoided: `dplyr::mutate(iris, Sepal.Length = .data$Sepal.Length * 2)`

meetagrawal09:
## Description

The aim of this PR is to replace the use of `transform()`
This PR fixes #2961

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/maat/R/write.config.MAAT.R
+++ b/models/maat/R/write.config.MAAT.R
@@ -59,19 +59,19 @@ convert.samples.MAAT <- function(trait.samples, runid) {
   }
   if ("Ha.vcmax" %in% names(trait.samples)) {
     ## Convert from kJ mol-1 to J mol-1
-    trait.samples <- transform(trait.samples, Ha.vcmax = PEcAn.utils::ud_convert(Ha.vcmax, "kJ", "J"))
+    trait.samples$Ha.vcmax <- PEcAn.utils::ud_convert(trait.samples$Ha.vcmax, "kJ", "J")
   }
   if ("Hd.vcmax" %in% names(trait.samples)) {
     ## Convert from kJ mol-1 to J mol-1
-    trait.samples <- transform(trait.samples, Hd.vcmax = PEcAn.utils::ud_convert(Hd.vcmax, "kJ", "J"))
+    trait.samples$Hd.vcmax <- PEcAn.utils::ud_convert(trait.samples$Hd.vcmax, "kJ", "J")
   }
   if ("Ha.jmax" %in% names(trait.samples)) {
     ## Convert from kJ mol-1 to J mol-1
-    trait.samples <- transform(trait.samples, Ha.jmax = PEcAn.utils::ud_convert(Ha.jmax, "kJ", "J"))
+    trait.samples$Ha.jmax <- PEcAn.utils::ud_convert(trait.samples$Ha.jmax, "kJ", "J")
   }
   if ("Hd.jmax" %in% names(trait.samples)) {
     ## Convert from kJ mol-1 to J mol-1
-    trait.samples <- transform(trait.samples, Hd.jmax = PEcAn.utils::ud_convert(Hd.jmax, "kJ", "J"))
+    trait.samples$Hd.jmax <- PEcAn.utils::ud_convert(trait.samples$Hd.jmax, "kJ", "J")
   }
   if ("leaf_reflect_vis" %in% names(trait.samples) & "leaf_trans_vis" %in% names(trait.samples) ){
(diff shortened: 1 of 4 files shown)


meetagrawal09:
Further occurrences of the `transform` function appear in `pecan.models.maat` and `pecan.modules.priors`
In these can we consider importing the `dplyr` package and using the `mutate` function to solve the issue ?

@Aariq any specific reason for doing the below for packages not importing `dplyr` ( anything affecting efficiency ? ): 
```
new_df <- old_df
new_df$new_col <- new_df$old_col * 2
```

Aariq:
> @Aariq any specific reason for doing the below for packages not importing `dplyr` ( anything affecting efficiency ? ): Reference [#2961comment](https://github.com/PecanProject/pecan/issues/2961#issuecomment-1188381697)

Just that it's sometimes a good idea to minimize the number of dependencies a package has.

meetagrawal09:
All instances of `transform()` have been replaced

models/maat/R/write.config.MAAT.R
```diff
@@ -59,19 +59,19 @@ convert.samples.MAAT <- function(trait.samples, runid) {
   }
   if ("Ha.vcmax" %in% names(trait.samples)) {
     ## Convert from kJ mol-1 to J mol-1
-    trait.samples <- transform(trait.samples, Ha.vcmax = PEcAn.utils::ud_convert(Ha.vcmax, "kJ", "J"))
+    trait.samples$Ha.vcmax <- PEcAn.utils::ud_convert(trait.samples$Ha.vcmax, "kJ", "J")
```
infotroph, line 62: I'm amused that the `transform`'s in this file were obviously intended to "save typing" but most of these lines are shorter without it!

modules/priors/vignettes/priors_demo.Rmd
```diff
@@ -294,7 +294,10 @@ vcmax.density <- with(vcmax.dist, pr.dens(distribution, a, b), xlim = c(0,50))
 
 ######### Vcmax Prior Plot
 vcmax.c4 <- all.vcmax.data$ft == 2
-vcmax.data <- transform(all.vcmax.data, mean = Y, se =  1/(sqrt(n) * obs.prec))[,c('ft', 'mean', 'se')]
+vcmax.data <- all.vcmax.data[,c('ft', 'Y')]
+vcmax.data$mean <- vcmax.data$Y
+vcmax.data$se <- 1/(sqrt(n) * obs.prec)
```
infotroph, line 299: This looks right if `obs.prec` is a variable in the local environment and not a column in `all.vcmax.data` -- if it's a column then this will ignore it where transform would use it and mask any `obs.prec` from the environment. 

(Apologies if the answer is obvious-- I'm reading this on my phone and didn't pull up the entire file)

infotroph (commented):
This is looking very good! Is it ready for full review or are you still adding changes?

infotroph (approved):

modules/priors/vignettes/priors_demo.Rmd
infotroph: I checked on a bigger screen and I think `obs.prec` *is* a column in the data frame, so I think it needs to be added to line 297 and then this line needs to become
```suggestion
vcmax.data$se <- 1/(sqrt(vcmax.data$n) * vcmax.data$obs.prec)
```

infotroph (changes requested):

modules/priors/vignettes/priors_demo.Rmd
meetagrawal09: Thanks for pointing that out : ) 
Committed the change.

infotroph (approved):

PR #3103, opened 2023-02-02
Adding SIPNET all states (snow, soil moisture, litter, wood, etc) to params$restart

helge22a:
## Description

Track water states in params$restart so that new model runs do not use default param values.
## Motivation and Context

Using default param values for water states adds cyclic behavior in the model output that is not present in the data. This should help work being done by @mdietze @serbinsh @DongchenZ @Qianyuxuan 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/db/R/dbfiles.R
```diff
@@ -350,9 +350,10 @@ dbfile.input.check <- function(siteid, startdate = NULL, enddate = NULL, mimetyp
         container.id = inputs$id,
         con = con,
         hostname = hostname,
-        machine.check = TRUE,
+        machine.check = FALSE, #issues with hostid not matching and resulting in error message dispite files existing on host machine
         return.all = return.all
       )
+      PEcAn.logger::logger.info("Set machine.check = FALSE, when TRUE results in error message even when files are on host machine")
```
Aariq, line 356: Should this just be a comment?  Doesn't seem like something the user has control over.  Either that or add it as an argument to the function and give it a default value?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -159,9 +159,14 @@ sda.enkf.multisite <- function(settings,
...
 
   #--get model specific functions
-  my.write_restart <- paste0("PEcAn.", model, "::write_restart.", model)
-  my.read_restart <- paste0("PEcAn.", model, "::read_restart.", model)
-  my.split_inputs  <- paste0("PEcAn.", model, "::split_inputs.", model)
+  #switch to depreciated version for debugging purposes
+  do.call("library", list(paste0("PEcAn.", model)))
+  my.write_restart <- paste0("write_restart.", model)
+  my.read_restart <- paste0("read_restart.", model)
+  my.split_inputs  <- paste0("split_inputs.", model)
```
Aariq, line 166: This looks kind of like you're trying to re-invent S3 dispatch.  Might be worth thinking about just implementing these functions as generic functions.  See http://adv-r.had.co.nz/S3.html

modules/assim.sequential/inst/Site_XMLS/testingMulti_HF.xml
```diff
@@ -115,7 +115,13 @@
...
   </run>
   <host>
-    <name>localhost</name>
+    <name>geo.bu.edu</name>
+    <user>ahelgeso</user>
+    <prerun>module load R/4.1.2</prerun>
+    <qsub>qsub -l h_rt=24:00:00 -q &apos;geo*&apos; -N SIPNETrun -S /bin/bash -o /projectnb/dietzelab/ahelgeso/Jobs/qsub -e /projectnb/dietzelab/ahelgeso/Jobs/qsub</qsub>
+    <qsub.jobid>Your job ([0-9]+) .*</qsub.jobid>
+    <qstat>'qstat -j @JOBID@ &amp;> /dev/null || echo DONE'</qstat>
+    <tunnel>/projectnb/dietzelab/ahelgeso/tunnel</tunnel>
```
Aariq, line 124: This doesn't seem like something that should be part of the R package

base/db/R/dbfiles.R
helge22a: It is already a function argument and the default value is TRUE. Recently, I've had issues with the machine check failing even though the files exist on the host machine (which is the point of the check). I added the message to aid in debugging and so that it would clear to others using the function, that for the time being I am changing the default value.

modules/assim.sequential/R/sda.enkf_MultiSite.R
helge22a: Good catch this is depreciated version of the code that I use when debugging, I will update to current version.

modules/assim.sequential/inst/Site_XMLS/testingMulti_HF.xml
helge22a: In order to execute model runs via a tunnel you need to make sure your host settings in the xml are configured properly. Including this as an example is helpful to other users, since the documentation can be a bit confusing.

base/db/R/dbfiles.R
Aariq: I mean that it doesn't look like it's an argument that's being passed from `dbfile.input.check()`.  So the logger message will display when`dbfile.input.check()` is run under certain conditions, but there's no way to change `machine.check` in `dbfile.input.check()`.

helge22a: I switched back to the default values for the PR and am keeping my changes locally.

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -22,9 +22,14 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   prior.sla <- params[[which(!names(params) %in% c("soil", "soil_SDA", "restart"))[1]]]$SLA
   
   forecast <- list()
-  # additional varnames, because we need these deterministic relationships
-  var.names <- c(var.names, "fine_root_carbon_content", "coarse_root_carbon_content")
-  
+  params$restart <-c() #state.vars not in var.names will be added here
+  #SIPNET inital states refer to models/sipnet/inst/template.param
+  state.vars <- c("SWE", "SoilMoistFrac", "AbvGrndWood", "TotSoilCarb", "LAI",
```
mdietze, line 27: litter moisture seems to still be missing

mdietze (changes requested):
Looks very close. In addition to adding litter moisture, make sure to update the change log and the documentation. Also, before we merge this in we really need confirmation that its been tested for a run with >1 site and that it is correctly writing the old restart states into the new files (i.e. that both the state and site matching is correct)

models/sipnet/R/read_restart.SIPNET.R
helge22a: While litter moisture is written to the sipnet.out files, this variable appears to be absent from the .nc files. I've review model2netcdf.SIPNET and litter water is not included in the variables transfered from sipnet.out to .nc. Since the read.restart.SIPNET function reads in the .nc files, I did not include litter water. If litter water should be included then this will require an update to model2netcdf as well as read.restart.SIPNET.

helge22a: Was this a previous oversight? Should I update model2netcdf to include litter water?

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -38,24 +43,18 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
...
   last <- which(as.Date(real_time)==as.Date(stop.time))[length(which(as.Date(real_time)==as.Date(stop.time)))]
   
-  params$restart <-c() ## This will be filled with some restart coefficient if above ground wood is in the state variables.
-  
   #### PEcAn Standard Outputs
   if ("AbvGrndWood" %in% var.names) {
     forecast[[length(forecast) + 1]] <- PEcAn.utils::ud_convert(ens$AbvGrndWood[last],  "kg/m^2", "Mg/ha")
     names(forecast[[length(forecast)]]) <- c("AbvGrndWood")
-    
+  }else{
```
DongchenZ, line 50: If AGB is in the var.names, then the root carbon will not be updated into params object.

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -111,6 +117,8 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   if ("TotSoilCarb" %in% var.names) {
     forecast[[length(forecast) + 1]] <- ens$TotSoilCarb[last]  ## kgC/m2
     names(forecast[[length(forecast)]]) <- c("TotSoilCarb")
+  }else{
+    params$restart["TotSoilCarb"] <- ens$TotSoilCarb[last]
   }
   
   print(runid)
```
DongchenZ, line 124: I think in addition to the read_restart function, you also need to make changes to the write_restart function, in the `write_restart_sipnet` function, the `params$restart` becomes `IC_extra` object, and you will need to write several `else` control just like what you did in the read_restart function to make sure that you pass the restart values from `IC_extra` to `analysis.save` object and name them accordingly.

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -22,9 +22,14 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
...
   
   forecast <- list()
-  # additional varnames, because we need these deterministic relationships
-  var.names <- c(var.names, "fine_root_carbon_content", "coarse_root_carbon_content")
-  
+  params$restart <-c() #state.vars not in var.names will be added here
+  #SIPNET inital states refer to models/sipnet/inst/template.param
+  state.vars <- c("SWE", "SoilMoistFrac", "AbvGrndWood", "TotSoilCarb", "LAI", 
+                  "litter_carbon_content", "fine_root_carbon_content", "coarse_root_carbon_content")
+  #pre-populate parsm$restart with NAs so state names can be added
```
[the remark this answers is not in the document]
helge22a: I will double check the spelling where I've added new calls for params$restart

Aariq:
This is probably a separate PR, but if for some reason you feel like doing some refactoring, a more robust alternative to the way model2netcdf works might be to use named lists.  Right now it relies a lot on indexing by position which is easy to screw up by accident when adding or removing variables.  

For example, instead of this ...
```r
out <- list()
out[[1]] <- sipnet.output$npp * conversion_factor # NPP
out[[2]] <- sipnet.output$snow * 10 # SWE

...

nc_var[[1]] <- PEcAn.utils::to_ncvar("NPP", dims)
nc_var[[2]] <- PEcAn.utils::to_ncvar("SWE", dims)

...
```

... do this ...

``` r
out <- list(
  "NPP" = sipnet.output$npp * conversion_factor,
  "SWE" = sipnet.output$snow * 10,
...
)

nc_var <- list(
  "SWE" = PEcAn.utils::to_ncvar("SWE", dims),
  "NPP" = PEcAn.utils::to_ncvar("NPP", dims),
...
)
#order doesn't matter anymore, because you can sort them by name if you need the two lists to be in the same order
nc_var[names(out)]

```

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -111,8 +125,13 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   if ("TotSoilCarb" %in% var.names) {
     forecast[[length(forecast) + 1]] <- ens$TotSoilCarb[last]  ## kgC/m2
     names(forecast[[length(forecast)]]) <- c("TotSoilCarb")
+  }else{
+    params$restart["TotSoilCarb"] <- ens$TotSoilCarb[last]
   }
   
+  #remove any remaining NAs from params$restart
+  na.omit(params$restart)
```
DongchenZ, line 133: please add global package name on the "na.omit" function. e.g. `stats::na.omit`

mdietze (commented):
still looking for parallel changes to write_restart

base/utils/data/standard_vars.csv
```diff
@@ -1,116 +1,117 @@
...
-Dens,NA,1 ha-1,Stem Density,Diversity,real,lon,lat,time,pft,Stem Density by PFT,
-DBH,NA,cm,Diameter at Breast Height,Diversity,real,lon,lat,time,pft,DBH by PFT,
-Fcomp,NA,kgC kgC-1,Aboveground Biomass Fractional Composition,Diversity,real,lon,lat,time,pft,Aboveground biomass Fractional composition of each PFT within each grid cell,
-Estab,NA,1 ha-1,New Individuals,Diversity,real,lon,lat,time,pft,New Individuals,
-Mort,NA,1 ha-1,Mortality,Diversity,real,lon,lat,time,pft,Individuals lost through death,
-SoilDepth,NA,m,Soil Depth Layer,Deprecated,real,depth,NA,NA,NA,Depth to the bottom of each model-defined soil layer,
-assimilation_rate,NA,kg C m-2 s-1,Leaf assimilation rate,Carbon Fluxes,real,lon,lat,time,NA,Rate of leaf photosynthesis / carbon assimilation,
-date_of_budburst,NA,day of year,Date of Budburst,Phenology,real,lon,lat,time,NA,Date of Budburst,
-date_of_senescence,NA,day of year,Date of Senescence,Phenology,real,lon,lat,time,NA,Date of Senescence,
+"","X.1","Variable.Name","standard_name","Units","Long.name","Category","var_type","dim1","dim2","dim3","dim4","Description","X"
```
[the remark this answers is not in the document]
helge22a: I loaded in the existing dataframe and added a new row. Not sure why it is showing up like this on github. I made no changes to the formatting.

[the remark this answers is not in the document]
mdietze: While I agree that you did not _intentionally_ make any changes to the formatting, this diff demonstrates that in practice you reformatted the whole file. Please revert and then in a text editor just add one new line to the bottom of the file.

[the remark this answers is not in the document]
helge22a: I fixed the formatting.

base/utils/data/standard_vars.csv
```diff
@@ -1,116 +1,117 @@
...
+"108","Fcomp",NA,"kgC kgC-1","Aboveground Biomass Fractional Composition","Diversity","real","lon","lat","time","pft","Aboveground biomass Fractional composition of each PFT within each grid cell"
+"109","Estab",NA,"1 ha-1","New Individuals","Diversity","real","lon","lat","time","pft","New Individuals"
+"110","Mort",NA,"1 ha-1","Mortality","Diversity","real","lon","lat","time","pft","Individuals lost through death"
+"111","SoilDepth",NA,"m","Soil Depth Layer","Deprecated","real","depth",NA,NA,NA,"Depth to the bottom of each model-defined soil layer"
+"112","assimilation_rate",NA,"kg C m-2 s-1","Leaf assimilation rate","Carbon Fluxes","real","lon","lat","time",NA,"Rate of leaf photosynthesis / carbon assimilation"
+"113","date_of_budburst",NA,"day of year","Date of Budburst","Phenology","real","lon","lat","time",NA,"Date of Budburst"
+"114","date_of_senescence",NA,"day of year","Date of Senescence","Phenology","real","lon","lat","time",NA,"Date of Senescence"
+"115","litter_mass_content_of_water",NA,"kg m-2","Average Layer Litter Moisture","Physical Variables","real","lon","lat","depth","time","Litter water content in litter layer, including liquid, vapor and ice"
```
mdietze, line 117: file still showing up as a major formatting change. Still needs to be reverted

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -2,7 +2,7 @@
 ##' 
 ##' @section Information on Units:
 ##' Information on NOAA weather units can be found below. Note that the temperature is measured in degrees C, 
-##' but is converted at the station and downloaded in Kelvin.
+##' but is converted at the station and downlaoded in Kelvin.
```
mdietze, line 5: You seem to have added a new spelling error

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -19,7 +19,7 @@
 ##' 
 ##' @section Data Save Format:
 ##' Data is saved in the netcdf format to the specified directory.  File names reflect the precision of the data to the given range of days.
-##' NOAA.GEFS.willow creek.3.2018-06-08T06:00.2018-06-24T06:00.nc specifies the forecast, using ensemble number 3 at willow creek on
+##' NOAA.GEFS.willow creek.3.2018-06-08T06:00.2018-06-24T06:00.nc specifies the forecast, using ensemble nubmer 3 at willow creek on
```
mdietze, line 22: same here

mdietze (changes requested):

modules/data.atmosphere/R/download.CRUNCEP_Global.R
```diff
@@ -9,7 +9,7 @@
 ##' @param overwrite logical. Download a fresh version even if a local file with the same name already exists?
 ##' @param verbose logical. Passed on to \code{\link[ncdf4]{ncvar_def}} and \code{\link[ncdf4]{nc_create}}
 ##'   to control printing of debug info
-##' @param maxErrors Maximum times to re-try following an error accessing netCDF data through THREDDS
+##' @param maxErrors Maximum times to re-try folloing an error accessing netCDF data through THREDDS
```
mdietze, line 12: another typo

Issue #2685, opened 2020-08-20
Roxygen2 v7.1.0 bump


istfer:
We are but it may be time to change that. If someone wants to take it on before I can, the process is basically:

* pick a date for the update and warn everyone
* ideally close out as many open PRs as possible before then, to minimize merge conflicts
* install the newest Roxygen on your dev machine
* git pull, make sure your tree is clean, then `make clean && make document`
* commit all the changed Rd and DESCRIPTION files
* `grep -R 7.0.2 .` -> review, update appropriately, commit. I think at least three scripts `scripts/travis/install.sh`, `.github/workflows/styler-actions.yml`, and `docker/depends/Dockerfile` will need updating, but I may be forgetting others.
* push, review, merge
* remind everyone to update Roxygen on their own machine
* monitor PRs for merge conflicts/accidental reversions (if it edits 300 files, it's probably an accidental reversion)

_Originally posted by @infotroph in https://github.com/PecanProject/pecan/pull/2672_

istfer:
Just created this issue so that @infotroph's helpful comment doesn't get buried. 

It would be nice if we can move from 7.0.2 to the most up-to-date version of Roxygen similar to what has been done [before](https://github.com/PecanProject/pecan/pull/2524)

Aariq:
Currently using v7.1.2 of roxygen2 so closing this

PR #3104, opened 2023-02-03
Updated base/qaqc

meetagrawal09:
## Description

The PR is aimed at fixing the "no visible binding for global variable" notes for `base/qaqc`

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


meetagrawal09:
@mdietze, @infotroph, @Aariq if you can please help me out with the failing `check 4.0`

Aariq:
Not sure.  I tried adding `rlang` to Imports and it didn't help.  But I did notice that the counterpart to this function, `find_inputs_without_formats()` probably doesn't work, and now I'm questioning whether the function you're editing even works.  Might want to write some tests before you mess around with them.

meetagrawal09:
Yes, will wait for some further clarification from you on this.
The only place where I see these functions being called is [Link](https://github.com/PecanProject/pecan/blob/353b4dd2eb632f5509a4dae56b6ca388c41873d8/base/qaqc/vignettes/Pre-release-database-cleanup.Rmd#L41)

Aariq:
> Yes, will wait for some further clarification from you on this. The only place where I see these functions being called is [Link](https://github.com/PecanProject/pecan/blob/353b4dd2eb632f5509a4dae56b6ca388c41873d8/base/qaqc/vignettes/Pre-release-database-cleanup.Rmd#L41)

You can see here that `user_id_code` isn't an argument for the function, so `!!user_id_code` will cause an error because it doesn't exist.

infotroph:
@meetagrawal09 Try rerunning Roxygen — looks like the new importFrom isn’t yet showing up in NAMESPACE

meetagrawal09:
> https://github.com/PecanProject/pecan/blob/353b4dd2eb632f5509a4dae56b6ca388c41873d8/base/qaqc/R/find_inputs_without_formats.R#L23
> 
> You can see here that `user_id_code` isn't an argument for the function, so `!!user_id_code` will cause an error because it doesn't exist.

Yes, I did see that issue. I believe it's a typo and should be `user_id_code` because the context and description for this variable in both the files `find_formats_without_inputs` and `find_inputs_without_formats` seems to be same.

meetagrawal09:
> @meetagrawal09 Try rerunning Roxygen — looks like the new importFrom isn’t yet showing up in NAMESPACE

Yes that did work. If you can give some suggestion about the `user_id` variable being changed to `user_id_code`  in file [pecan/base/qaqc/R/find_inputs_without_formats.R](https://github.com/PecanProject/pecan/blob/353b4dd2eb632f5509a4dae56b6ca388c41873d8/base/qaqc/R/find_inputs_without_formats.R#L23)

base/qaqc/DESCRIPTION
```diff
@@ -43,4 +44,4 @@ LazyData: FALSE
 Encoding: UTF-8
 VignetteBuilder: knitr
 Config/testthat/edition: 3
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.3
```
infotroph, line 47: Sorry about this undocumented requirement, but: Can you please rebuild this using Roxygen 7.1.3? You can install it by running `remotes::install_version("roxygen2", "7.1.3")`, then manually edit this line back to "7.1.3" (Roxygen doesn't run if it thinks a newer version touched the package last) and run Roxygen again.

Why are we picky about making everyone use an outdated Roxygen version? That's a fair question. Our reasoning for making everyone use the same version is that If different contributors are using different versions of Roxygen, then redocumenting after one change can wind up introducing spurious changes in another place you didn't touch, and then another contributor with a different version of Roxygen tries to redocument their work and winds up reverting your change, and so on.

We are due for a bump in our version requirement so that we insist everyone use Roxygen 7.2.3 instead of Roxygen 7.1.3, but we haven't gotten around to doing it yet. If anyone reading this wants to take on this version bump, see #2685 for an outline of the approach.

meetagrawal09: Thanks for the very clear explanation. It totally makes sense. 
Well actually there is no `version 7.1.3` for `roxygen2`. Checked releases and also running the suggested command gives 

> version '7.1.3' is invalid for package 'roxygen2'

[roxygen2-releases](https://github.com/r-lib/roxygen2/releases)

infotroph: 🤦 I meant 7.1.2, like in the old version of this line, but apparently I carried the typo through all four times I wrote it out... maaaaaybe I need to start wearing glasses at the computer.

infotroph:
> How can I update the [Rcheck_reference.log](https://github.com/PecanProject/pecan/blob/develop/base/qaqc/tests/Rcheck_reference.log)

The general pattern for editing `Rcheck_reference.log` is to delete the line that complains, and if that removes the last line of complains from a given check, edit the line for that check to end in `OK` instead of `NOTE` or `WARNING`. So if you fixed all the "no visible binding" errors, then you'll edit line 47 and delete lines 48-71:

```{diff}
- * checking R code for possible problems ... NOTE
- cull_database_entries: no visible global function definition for
-   ‘read.table’
- cull_database_entries: no visible global function definition for
-   ‘write.table’
- find_formats_without_inputs: no visible binding for global variable
-   ‘user_id’
- find_formats_without_inputs: no visible binding for global variable
-   ‘created_at’
- find_formats_without_inputs: no visible binding for global variable
-   ‘updated_at’
- find_inputs_without_formats: no visible binding for global variable
-   ‘user_id_code’
- find_inputs_without_formats: no visible binding for global variable
-   ‘created_at’
- find_inputs_without_formats: no visible binding for global variable
-   ‘updated_at’
- write_out_table: no visible global function definition for
-   ‘write.table’
- Undefined global functions or variables:
-   created_at read.table updated_at
-   user_id user_id_code write.table
- Consider adding
-   importFrom("utils", "read.table", "write.table")
- to your NAMESPACE file.
+ * checking R code for possible problems ... OK
* checking Rd files ... OK
```

After editing, confirm that the edit worked by running `./scripts/check_with_errors.R base/qaqc` in your terminal -- if it complains, then one of the deleted lines isn't actually fixed yet.

> Is there a need to add an explicit import for these ?

Add them to the `Imports:` section of `DESCRIPTION`. No need for any Roxygen `@importFrom` in this case.

base/qaqc/DESCRIPTION
```diff
@@ -12,10 +12,13 @@ Author: David LeBauer, Tess McCabe
 Maintainer: David LeBauer <dlebauer@illinois.edu>
 Description: PEcAn integration and model skill testing
 Imports:
+    dplyr,
     graphics,
+    rlang,
```
infotroph, line 17: Nit: we prefer to list imports in alphabetical order (R doesn't care, this just makes it easier to skim when asking "Let's see, does this package depend on ____?")

```suggestion
```

base/qaqc/DESCRIPTION
```diff
@@ -12,10 +12,13 @@ Author: David LeBauer, Tess McCabe
...
+    PEcAn.DB,
     PEcAn.logger,
     plotrix,
-    stats
+    stats
```
infotroph, line 21:
```suggestion
    rlang,
    stats
```

base/qaqc/R/cull_database_entries.R
```diff
@@ -17,63 +17,50 @@
...
 ##'
-cull_database_entries<-function(table = NULL, outdir, file_name = NULL, con, machine_id = NULL, table_name = NULL){
-  
-  
-  if(is.null(table)){
-    
-    if(is.null(file_name)){
-      
+cull_database_entries <- function(table = NULL, outdir, file_name = NULL, con, machine_id = NULL, table_name = NULL) {
+  if (is.null(table)) {
```
infotroph, line 21: Thank you for including the stylr run, and especially for doing it as its own commit at the end of the PR! (That makes it easy for us to review the substantive changes separately from any widespread-but-minor formatting fixes)👍

infotroph (approved):
One more tiny suggestion to make the DESCRIPTION file easier to read, otherwise this looks great. Thank you!

PR #3108, opened 2023-02-09
Check for `host` being `NULL`

### LINKED ISSUE #3107: PEcAn.remote::remote.copy.to and from don't check that `host` argument isn't NULL
### Bug Description

Made a typo and did `remote.copy.to(host = settings$hots, ...)` and rsync failed with uninformative error.  It ended up running this:

``` bash
rsync '-a' '-q' '/path/to/src' ' :/path/to/dst' 
```

instead of

``` bash
rsync '-a' '-q' '/path/to/src' ' username@host:/path/to/dst' 
```

## Expected behavior

Should fail early if `host` is `NULL` with informative message.

meetagrawal09:
## Description

PEcAn functions will fail early if `host` is `NULL` with informative message when making a call to `remote.copy.to()` 
This PR fixes #3107 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/remote/R/remote.copy.to.R
+++ b/base/remote/R/remote.copy.to.R
@@ -25,18 +25,12 @@ remote.copy.to <- function(host, src, dst, options = NULL, delete = FALSE, stder
   if (is.null(host)) {
     PEcAn.logger::logger.severe("`host` object passed to the function is NULL : Try passing a valid host object")
   }
-  else {
-    if (is.null(host$name)) {
-      PEcAn.logger::logger.severe("`name` parameter in the `host` object is NULL : Try passing a valid host object")
-    } else if (is.null(host$user)) {
-      PEcAn.logger::logger.severe("`user` parameter in the `host` object is NULL : Try passing a valid host object")
-    } else if (is.null(host$tunnel)) {
-      PEcAn.logger::logger.severe("`tunnel` parameter in the `host` object is NULL : Try passing a valid host object")
-    }
-  }
   if (is.localhost(host)) {
     args <- c(args, src, dst)
   } else {
+    if (is.null(host$name)) {
+      PEcAn.logger::logger.severe("`name` parameter in the `host` object is NULL : Try passing a valid host object")
+    }
     tunnel <- host$tunnel
     if (!is.null(host$data_tunnel)) {
       tunnel <- host$data_tunnel


base/remote/R/remote.copy.to.R
```diff
@@ -22,6 +22,9 @@ remote.copy.to <- function(host, src, dst, options = NULL, delete = FALSE, stder
   if (as.logical(delete)) {
     args <- c(args, "--delete")
   }
+  if (is.null(host)) {
+    PEcAn.logger::logger.severe("Host param provided is NULL : Possibly a typo in the host name variable or a missing host in the arguments passed")
+  }
```
Aariq, line 27: Another approach that might catch more problems is to assert that `host` is a character value of length 1
```suggestion
 if(!is.character(host) | length(host) > 1)
    PEcAn.logger::logger.severe("`host` should be a length 1 character vector, not ", class(host), " length ", length(host))
  }
```

base/remote/R/remote.copy.to.R
```diff
@@ -22,6 +22,9 @@ remote.copy.to <- function(host, src, dst, options = NULL, delete = FALSE, stder
   if (as.logical(delete)) {
     args <- c(args, "--delete")
   }
+  if (!is.character(host) | length(host) > 1) {
+    PEcAn.logger::logger.severe("`host` should be a length 1 character vector, not ", class(host), " length ", length(host))
```
infotroph, line 26: The rest of the function is expecting `host` to have at least 3 entries -- see the @examples block above.

Aariq: ah right, I forgot that `settings$host` has a bunch of stuff in it.

infotroph: I think I'd advise testing that each of the needed items _inside_ `host` are non-NULL and have the needed format... then you'll correctly fail on cases like `host = list(naem = ...)` as well as on `host = NULL`.

base/remote/R/remote.copy.to.R
```diff
@@ -22,6 +22,18 @@ remote.copy.to <- function(host, src, dst, options = NULL, delete = FALSE, stder
   if (as.logical(delete)) {
     args <- c(args, "--delete")
   }
+  if (is.null(host)) {
+    PEcAn.logger::logger.severe("`host` object passed to the function is NULL : Try passing a valid host object")
+  }
+  else {
+    if (is.null(host$name)) {
+      PEcAn.logger::logger.severe("`name` parameter in the `host` object is NULL : Try passing a valid host object")
+    } else if (is.null(host$user)) {
```
mdietze, line 31: Are name and user both required if tunnel exists? It's been a while since I've used this, but my recollection was that if you had a valid tunnel you didn't need the user?

mdietze (approved):

PR #3109, opened 2023-02-09
HF landscape

mdietze:
Code to doing a landscape-scale SDA, using Harvard Forest as the test case


modules/assim.sequential/R/sda.enkf_MultiSite.R
helge22a, line 1: Are you sure this is the most recent version of this function? The check for observations and for the analysis to skip if there are no observations was fixed in one of the last PR but it looks like this is the old version?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -125,9 +125,9 @@ sda.enkf.multisite <- function(settings,
   #Here I'm trying to make a temp config list name and put it into map to iterate
   if(multi.site.flag){
     conf.settings<-settings
-    site.ids <- conf.settings %>% purrr::map(~.x[['run']] ) %>% purrr::map('site') %>% purrr::map('id') %>% unlist() %>% as.character()
+    site.ids <- conf.settings$run %>% map('site') %>% map('id') %>% unlist() %>% as.character()
```
mdietze, line 128: note to self: somehow namespaces got lost in a merge

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -344,14 +345,16 @@ sda.enkf.multisite <- function(settings,
       new.params <- sda_matchparam(settings, ensemble.samples, site.ids, nens)
     }
       #sample met ensemble members
-      inputs <- conf.settings %>% purrr::map(function(setting) {
+      #TODO: incorporate Phyllis's restart work
+      #      sample all inputs specified in the settings$ensemble not just met
+      inputs <- papply(conf.settings,function(setting) {
```
mdietze, line 350: add namespace to papply

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -365,9 +368,9 @@ sda.enkf.multisite <- function(settings,
       if (t>1){
         #for next time step split the met if model requires
         #-Splitting the input for the models that they don't care about the start and end time of simulations and they run as long as their met file.
-        inputs.split <- PEcAnAssimSequential::metSplit(conf.settings, inputs, settings, model, no_split = FALSE, obs.times, t, nens, restart_flag = FALSE, my.split_inputs)
+        inputs.split <- metSplit(conf.settings, inputs, settings, model, no_split = FALSE, obs.times, t, nens, restart_flag = FALSE, my.split_inputs)
```
mdietze, line 371: namespace

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -500,18 +511,10 @@ sda.enkf.multisite <- function(settings,
...
-          recompileTobit = FALSE
-        }
-        
-        if(!exists('Cmcmc')) {
-          recompileGEF = TRUE
-        }else{
-          recompileGEF = FALSE
-        }
+
+        recompileTobit = !exists('Cmcmc_tobit2space')
```
DongchenZ, line 515: Great one!

mdietze:
@DongchenZ I think this code is ready to go. Could you give a quick look and then "approve" if you think its good.

@helge22a as far as I can tell my code doesn't actively revert any of the changes you mentioned, as much as they seem to have gotten lost along the way independent of this PR. Would be good to track those down and get them applied, but I don't think there's anything I can do in this PR to fix that.

PR #3114, opened 2023-02-12
deleted 'isFALSE'

### LINKED ISSUE #3102: isFALSE exists in base, don't need to have our own version
https://github.com/PecanProject/pecan/blob/353b4dd2eb632f5509a4dae56b6ca388c41873d8/base/utils/R/utils.R#L417

This line should be removed as base already has an `isFALSE()` function.

Its-Maniaco:
Fixes #3102


mdietze:
Equivalent change was already pulled in via PR #3115. Close #3102

PR #3118, opened 2023-02-16
Refactor `model2netcdf.SIPNET` variable lists

### LINKED ISSUE #3117: Refactor model2netcdf.SIPNET variable lists
@Aariq suggested the following in the discussion for PR #3103. I agree that its a good idea and also that it should be its own issue.
  
>This is probably a separate PR, but if for some reason you feel like doing some refactoring, a more robust alternative to the way model2netcdf works might be to use named lists.  Right now it relies a lot on indexing by position which is easy to screw up by accident when adding or removing variables.  

>For example, instead of this ...
```r
out <- list()
out[[1]] <- sipnet.output$npp * conversion_factor # NPP
out[[2]] <- sipnet.output$snow * 10 # SWE

...

nc_var[[1]] <- PEcAn.utils::to_ncvar("NPP", dims)
nc_var[[2]] <- PEcAn.utils::to_ncvar("SWE", dims)

...
```

>... do this ...

``` r
out <- list(
  "NPP" = sipnet.output$npp * conversion_factor,
  "SWE" = sipnet.output$snow * 10,
...
)

nc_var <- list(
  "SWE" = PEcAn.utils::to_ncvar("SWE", dims),
  "NPP" = PEcAn.utils::to_ncvar("NPP", dims),
...
)
#order doesn't matter anymore, because you can sort them by name if you need the two lists to be in the same order
nc_var[names(out)]

```

_Originally posted by @Aariq in https://github.com/PecanProject/pecan/issues/3103#issuecomment-1428597383_

meetagrawal09:
## Description
The PR fixes #3117 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/R/model2netcdf.SIPNET.R
+++ b/models/sipnet/R/model2netcdf.SIPNET.R
@@ -178,24 +178,22 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date, 
     bounds <- round(bounds,4) 
     
     ## Setup outputs for netCDF file in appropriate units
-    output       <- list()
-    output[[1]]  <- (sub.sipnet.output$gpp * 0.001) / timestep.s  # GPP in kgC/m2/s
-    ## output[[2]] <- (sub.sipnet.output$npp*0.001) / timestep.s # NPP in kgC/m2/s. Internal SIPNET
-    ## calculation
-    output[[2]]  <- (sub.sipnet.output$gpp * 0.001) / timestep.s - ((sub.sipnet.output$rAboveground *
-                                                                       0.001) / timestep.s + (sub.sipnet.output$rRoot * 0.001) / timestep.s)  # NPP in kgC/m2/s. Post SIPNET calculation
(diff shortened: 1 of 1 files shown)


mdietze (changes requested):
This is looking good, but I think there's a key idea in @Aariq's suggestion that was a bit subtle but I think still needs to be implemented. In the following loop:
```
      for (i in seq_along(nc_var)) {
        ncdf4::ncvar_put(nc, nc_var[[i]], output[[i]])
        cat(paste(nc_var[[i]]$name, nc_var[[i]]$longname), file = varfile, sep = "\n")
      }
```
We're still using the nc_var and output lists by number instead of by name. This means that the "protection" we're gaining by naming the variables in the list isn't actually being used in practice. As such, I think the new version actually makes the code _more_ susceptible to error since it's harder to verify that the two lists are aligned. The better solution would be to loop over the names in one of the lists and then reference both lists by those names. If that was done then they either list could be in any order and they'd still be paired correctly when building the netcdf file.

mdietze (approved):

Issue #3084, opened 2023-01-09
Broken link and missing documentation for `cos_solar_zenith_angle()`


Aariq:
### Bug Description

The documentation for `cos_solar_zenith_angle()`, an exported function in `PEcAn.data.atmosphere` has only a 1 sentence description that points the user to what is now a spam gambling website. The arguments are not well explained and there is no example. For example, I don't understand what the difference between `dt` (Timestep) and `hr` (Hours timestep) is.

## Expected behavior

Use the `@references` roxygen2 tag to add a reference
Use `@examples` to add examples (required by CRAN I think)

OR make this a non-exported, internal function.

deep-poharkar:
can you assign me this?

Sweetdevil144:
I see the issue is still open and still requires attention. i wanted to proceed with this and fix it and would propose the following changes for now : 

```R
#' Cosine of Solar Zenith Angle
#'
#' Calculates the cosine of the solar zenith angle based on the given parameters. 
#' This angle is crucial in determining the amount of solar radiation reaching a point on Earth.
#'
#' @param doy Day of year. Integer representing the day of the year (1-365).
#' @param lat Latitude in degrees. Positive for the Northern Hemisphere and negative for the Southern Hemisphere.
#' @param lon Longitude in degrees. Positive for East and negative for West.
#' @param dt Time interval in seconds. Represents the duration over which the measurement is averaged or integrated.
#' @param hr Hour of the day (0-23). Specifies the specific hour for which the calculation is made.
#'
#' @return Numeric value representing the cosine of the solar zenith angle.
#'
#' @references 
#' "Understanding Solar Position and Solar Radiation" - RAMMB: [Link](https://rammb.cira.colostate.edu/wmovl/vrl/tutorials/euromet/courses/english/nwp/n5720/n5720005.htm)
#'
#' @examples
#' cos_solar_zenith_angle(doy = 150, lat = 45, lon = -93, dt = 3600, hr = 12)
#'
#' @export
cos_solar_zenith_angle <- function(doy, lat, lon, dt, hr) {
  # Function implementation (remains the same)
}
```

Any recommendation @Aariq ?
We can also proceed with adding documentations for other similar functions too which have low descriptions.

Aariq:
I'm not an expert at this kind of thing, but the documentation is clear to me

dlebauer:
@Sweetdevil144 I agree, that looks clear. Feel free to submit a PR

PR #3120, opened 2023-02-17
fixed the missing documentation.

deep-poharkar:
## Description

I tried to fix (Broken link and missing documentation for cos_solar_zenith_angle() #3084) by providing description, references and example. The given link in description was working but I can update that if needed. I want you to review and tell me what exactly needs to be changed. I am not sure about what should I write in @example

## Motivation and Context

This will fix the #3084 which will improve the documentation.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


mdietze (changes requested):
Doesn't really address the core problems raised in https://github.com/PecanProject/pecan/issues/3084 (e.g. dt vs hr, what are the units on these things? which things are scalars vs vectors?) but instead just reiterates information that's already present in a slightly different (but not necessarily better) format.

FWIW, I don't know if someone already fixed the link issue mentioned in #3084, but that doesn't seem to be a problem for me.

deep-poharkar:
@mdietze can you review the improved one.

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,27 @@
...
 #'
+#' @description
+#' `cos_solar_zenith_angle` returns the cosine of solar zenith angle.
+#' The cosine of the solar zenith angle (cossza) is the angle between the sun’s rays and the vertical
+#' This solar zenith angle is normally used in combination with the solar azimuth angle to determine the position of the Sun as observed from a given location on the surface of the Earth.
+#' The cosine of the solar zenith angle (cossza) is also a key component in Mean Radiant Temperature(MRT).
+#' It is required for photolysis.
+#'
+#' doy Day of year : It is the day when you are calculating the angle.
+#' lat Latitude : Latitude measures the distance north or south of the equator. The primary unit of latitude is degrees (°).
```
mdietze, line 13: Ok, but does the function want decimal degrees, or degrees, minutes, seconds?

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,27 @@
...
+#' lon Longitude : Longitude measures distance east or west of the prime meridian. The primary unit of longitude is degrees (°).
+#' dt Timestep : Timestep is a scalar quantity.
+#' hr Hours timestep : It is a timestep in hours format.
```
mdietze, line 16: Still doesn't really explain what dt and hr actually are and what acceptable values are. Looking at the code, hr isn't a "timestep" it's the time of day that you're trying to predict zenith angle for (If I'd written this function I'd have put it next to doy to make this connection more obvious). Looking at the code I'd guess this should be a number between 0-24 and that decimal values are acceptable (i.e. one should use fractional hours instead of minutes). I agree dt is probably a scalar, but you likewise still haven't said what it actually is or what acceptable values are. My best guess it that the author of the function anticipated that many people would be passing in vectors of days and hours, and thus if you have a whole vector of times you might actually want to do the calculation for the midpoint between two timesteps. dt thus tells the function how far apart values in hr are. Seeing that this is calculated by dividing dt by 86400 and then multiplying by 24, it looks like dt needs to be in SECONDS. In my mind dt should probably be given a default of 0.

modules/data.atmosphere/R/solar_angle.R
deep-poharkar: decimal degrees.

modules/data.atmosphere/R/solar_angle.R
deep-poharkar: sure, I have made the changes in the code and made it more in detail. I have one doubt should I write an example, if so then it'll take too many lines.

deep-poharkar:
@mdietze what do you think about the updated information.

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#' When the solar zenith angle is small (i.e. the sun is close to directly overhead), the intensity of solar radiation is high.
+#' When the solar zenith angle is larger (i.e. the sun is closer to the horizon), the intensity of solar radiation is lower, resulting in cooler temperatures.
+#' It is required for photolysis.
+#'
+#' It depends upon the following -
+#'
+#' doy Day of year : It is the day when you are calculating the angle.
+#' January 1st being day 1 and December 31st being day 365 (or 366 in leap years).
+#' It is used to determine the position of the sun in the sky throughout the year.
+#' Unit - Dimensionless (a simple count of the day within a year).
```
Aariq, line 18: Use roxygen2 tags for this
```suggestion
#'
#' @param doy Day of year : It is the day when you are calculating the angle.
#' January 1st being day 1 and December 31st being day 365 (or 366 in leap years).
#' It is used to determine the position of the sun in the sky throughout the year.
#' Unit - Dimensionless (a simple count of the day within a year).
```

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
 #'
 #' For explanations of formulae, see https://web.archive.org/web/20180307133425/http://www.itacanet.org/the-sun-as-a-source-of-energy/part-3-calculating-solar-angles/
```
Aariq, line 3: CRAN is picky about URLs—can you run `urlchecker::url_check("modules/data.atmosphere")` to check that this is OK?

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#' With a positive sign for locations east of the prime meridian and a negative sign for locations west of the prime meridian.
+#' (e.g., -122.4194° for San Francisco)
+#'
+#' @references
+#' https://www.ecmwf.int/sites/default/files/elibrary/2022/20336-calculating-cosine-solar-zenith-angle-thermal-comfort-indices.pdf
+#' https://centaur.reading.ac.uk/104581/
+#' https://wiki.seas.harvard.edu/geos-chem/index.php/Centralized_chemistry_time_step
+
+#' @examplesIf interactive()
+#' browseURL("https://www.ecmwf.int/sites/default/files/elibrary/2022/20336-calculating-cosine-solar-zenith-angle-thermal-comfort-indices.pdf")
```
mdietze, line 47: This seems like it's a reference, not an example, and it duplicates an existing reference so I don't think it is needed

modules/data.atmosphere/R/solar_angle.R
mdietze: Further clarification: you already have a `@param` tag for every variable, and you don't want to create duplicates. Instead, you want to add the basic variable-by-variable details to the existing variable descriptions (which are too short). Beyond that, if you need to provide a more detailed explanation of what the function is doing, it should go in a `@details` section, not in the `@description`, which should be no more than a few sentences.

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#'
+#' lon Longitude : Longitude measures distance east or west of the prime meridian.
+#' Unit - Degrees (°) in decimal format.
+#' With a positive sign for locations east of the prime meridian and a negative sign for locations west of the prime meridian.
+#' (e.g., -122.4194° for San Francisco)
+#'
+#' @references
+#' https://www.ecmwf.int/sites/default/files/elibrary/2022/20336-calculating-cosine-solar-zenith-angle-thermal-comfort-indices.pdf
+#' https://centaur.reading.ac.uk/104581/
+#' https://wiki.seas.harvard.edu/geos-chem/index.php/Centralized_chemistry_time_step
```
mdietze, line 44: This reference seems tangential. I'd drop

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#'
+#' It depends upon the following -
+#'
+#' doy Day of year : It is the day when you are calculating the angle.
+#' January 1st being day 1 and December 31st being day 365 (or 366 in leap years).
+#' It is used to determine the position of the sun in the sky throughout the year.
+#' Unit - Dimensionless (a simple count of the day within a year).
+#'
+#' hr Hours timestep : It is the time of day that you're trying to predict zenith angle for.
+#' It's an integer value ranging from 0 to 24.
```
mdietze, line 21: Why integer? Code looks like it would be happy to work with fractional hours

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#' hour time step might be used to represent the change in solar position and angle between consecutive hours within a day.
+#' Unit - Hours (h)
+#'
+#' dt Timestep : It represents the time interval between two values.
+#' Therefore it is an integer value ranging from 0 to 60.
+#' It should have a default value of 0.
+#' It is designed to calculate the midpoint between two time steps if value is passed in vectors of days and hours.
+#' dt tells the function how far apart values in hours are.
+#' a day time step might be used to represent the change in solar position and angle between consecutive days.
+#' Unit - Minutes (min).
```
mdietze, line 31: pretty sure it's actually seconds

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#' Unit - Dimensionless (a simple count of the day within a year).
+#'
+#' hr Hours timestep : It is the time of day that you're trying to predict zenith angle for.
+#' It's an integer value ranging from 0 to 24.
+#' hour time step might be used to represent the change in solar position and angle between consecutive hours within a day.
+#' Unit - Hours (h)
+#'
+#' dt Timestep : It represents the time interval between two values.
+#' Therefore it is an integer value ranging from 0 to 60.
+#' It should have a default value of 0.
```
mdietze, line 27: Ok, but you didn't give the function a default

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#' It is used to determine the position of the sun in the sky throughout the year.
+#' Unit - Dimensionless (a simple count of the day within a year).
+#'
+#' hr Hours timestep : It is the time of day that you're trying to predict zenith angle for.
+#' It's an integer value ranging from 0 to 24.
+#' hour time step might be used to represent the change in solar position and angle between consecutive hours within a day.
+#' Unit - Hours (h)
+#'
+#' dt Timestep : It represents the time interval between two values.
+#' Therefore it is an integer value ranging from 0 to 60.
```
mdietze, line 26: 1. doesn't have to be an integer
2. I don't think you need to specify the range
3. the "between two values" comment doesn't make sense if you are feeding the functions scalar values, which is what a lot of people would assume you do. I think it would be better to add something like "Used if you are passing the function a vector times and want to calculate radiation at the midpoint between timesteps"

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -2,6 +2,50 @@
...
+#' It should have a default value of 0.
+#' It is designed to calculate the midpoint between two time steps if value is passed in vectors of days and hours.
+#' dt tells the function how far apart values in hours are.
+#' a day time step might be used to represent the change in solar position and angle between consecutive days.
+#' Unit - Minutes (min).
+#'
+#' lat Latitude : Latitude measures the distance north or south of the equator.
+#' Unit - Degrees (°) in decimal format (e.g., 37.7749° for San Francisco).
```
mdietze, line 34: could be shortened to "decimal degrees"

modules/data.atmosphere/R/solar_angle.R
```diff
@@ -11,18 +55,18 @@
 #' @return `numeric(1)` of cosine of solar zenith angle
 #' @export
```
Aariq, line 56: *usually* `@examples` goes at the end of the roxygen2 comments block.  There's no hard rule about that, but since examples appear last in rendered documentation, it makes sense to put them last here as well.

mdietze:
@deep-poharkar pinging you to see if you could finish up this PR

mdietze:
@deep-poharkar pinging you about finishing up this PR

infotroph:
Looks like #3084 has now been fixed by #3276. I'm sorry we weren't able to use this!

PR #3121, opened 2023-02-17
Updated `PEcAn.utils::download.file` to `PEcAn.utils::download_file`

### LINKED ISSUE #2957: `PEcAn.utils::download.file` masks `utils::download.file`.  Change name?
https://github.com/PecanProject/pecan/blob/cd3e4c2dc1739b4165574dd1200f73886d80af9a/base/utils/R/utils.R#L647

This might be annoying to users who find base R's `download.file()` unexpectedly working differently and with different arguments.  Consider renaming to `download_file()`?

meetagrawal09:
## Description
This PR fixes #2957

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ convert data for a single PFT fixed (#1329, #2974, #2981)
   This was needed to resolve a cyclic dependency between PEcAn.DB and PEcAn.utils.
   (#3026; @nanu1605)
 - Internal changes to keep up to date with tidyselect v1.2.0
+- The `PEcAn.utils::download.file()` function has now been renamed to `PEcAn.utils::download_file()`
 
 ### Removed
 
@@ -434,8 +435,7 @@ This is a major change:
 - Five functions from PEcAn.utils functions have been moved to other packages. The versions in PEcAn.utils are deprecated, will not be updated with any new features, and will be removed in a future release.
   - run.write.configs and runModule.run.write.configs have been moved to PEcAn.workflow
   - read.ensemble.output, get.ensemble.samples and write.ensemble.configs have been moved to PEcAn.uncertainty
-- Change the way packages are checked for and called in SHINY apps. DESCRIPTION files in SHINY apps are not the place to declare pacakge dpendencies.    
-- The `PEcAn.utils::download.file()` function has now been renamed to `PEcAn.utils::download_file()`
+- Change the way packages are checked for and called in SHINY apps. DESCRIPTION files in SHINY apps are not the place to declare pacakge dpendencies.
 
 ## [1.5.3] - 2018-05-15
 


mdietze:
@meetagrawal09 I think it's safe to replace all instances of download.file with download_file except the ones inside download_file itself. Looks like 99% of the time download_file just calls download.file except in the case where the file is on an ftp server and download.ftp.method is set to a specific value (ncftpget) that allows @serbinsh to get around a firewall issue.

mdietze:
Yes, this change should go in the change log

mdietze (approved):

modules/data.remote/R/NLCD.R
```diff
@@ -38,7 +38,7 @@ download.NLCD <- function(outdir, year = 2011, con = NULL) {
     ## Download the data
     dir.create(outdir, showWarnings = FALSE)
     destfile <- file.path(outdir, paste0("nlcd", year, ".zip"))
-    download.file(url, destfile = destfile)
+    PEcAn.utils::download_file(url, destfile = destfile)
```
infotroph, line 41: Please convert this to `utils::download.file` -- this was (and should remain) a call to the base R `download.file` and the updated call will fail because it doesn't know `destfile` is the same argument it calls `filename`.

modules/data.remote/inst/scripts/Download_OSU_AGB_data_v5.R
```diff
@@ -67,13 +67,13 @@ doParallel::registerDoParallel(cl)
 # get median data
 med_urls <- paste(URL,dataset_version,target_dataset,"median",med_files,sep="/")
 PEcAn.logger::logger.info("*** Downloading AGB median data")
-foreach::foreach(i=1:length(med_urls)) %dopar% try(download.file(med_urls[i], file.path(target_directory,med_files[i])))
+foreach::foreach(i=1:length(med_urls)) %dopar% try(PEcAn.utils::download_file(med_urls[i], file.path(target_directory,med_files[i])))
```
infotroph, line 70: This call won't fail since the argument is unnamed, but please keep using the base version (i.e. `utils::download.file`) so that this PR isn't changing which function existing calls use.

modules/data.remote/inst/scripts/Download_OSU_AGB_data_v5.R
```diff
@@ -67,13 +67,13 @@ doParallel::registerDoParallel(cl)
...
 rm(i)
 
 # get stdv data
 stdv_urls <- paste(URL,dataset_version,target_dataset,"stdv",stdv_files,sep="/")
 PEcAn.logger::logger.info("*** Downloading AGB stdv data")
-foreach::foreach(j=1:length(stdv_urls)) %dopar% try(download.file(stdv_urls[j], file.path(target_directory,stdv_files[j])))
+foreach::foreach(j=1:length(stdv_urls)) %dopar% try(PEcAn.utils::download_file(stdv_urls[j], file.path(target_directory,stdv_files[j])))
```
infotroph, line 76: Same here

modules/data.remote/tests/Rcheck_reference.log
```diff
@@ -98,8 +98,6 @@ call_MODIS: no visible global function definition for ‘write.csv’
 download.LandTrendr.AGB: no visible binding for global variable ‘k’
 download.NLCD: no visible global function definition for ‘dbfile.check’
 download.NLCD: no visible global function definition for ‘db.query’
-download.NLCD: no visible global function definition for
```
infotroph, line 101: 👍

infotroph:
@meetagrawal09 Thanks for this change! I see this has been merged already, but it needs a followup to correct a call that will fail (yes, I wish our test coverage was good enough to have caught this 😞 )

PR #3123, opened 2023-02-25
Adding tests to `PEcAn.remote`

meetagrawal09:
## Description
The PR aims to add unit tests for functions in `PEcAn.remote`.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


base/remote/tests/testthat/test-fqdn.R
```diff
@@ -1,3 +1,14 @@
...
 })
+
+test_that("`fqdn()` returns expected `FQDN` value", {
+  
+  #set FQDN
+  Sys.setenv(FQDN = "pecan_host")
+  expect_equal(fqdn(), "pecan_host")
+  
+  #unset FQDN
+  Sys.unsetenv("FQDN")
```
infotroph, line 12: Note that this can have persistent side effects after the test is finished (in this case: If another test downstream is expecting to find `$FQDN` set, this will affect it). Wrapping these tests in [withr::with_envvar()](https://withr.r-lib.org/reference/with_envvar.html) would be a clean alternative.

base/remote/tests/testthat/test.remote.execute.cmd.R
```diff
@@ -0,0 +1,16 @@
+context("Testing remote.execute.cmd() for different possible inputs")
```
infotroph, line 1: The testthat team [no longer recommends using `context`](https://testthat.r-lib.org/reference/context.html) -- they recommend (and I agree) having one context per file and picking filenames that convey it well.

base/remote/tests/testthat/test.remote.execute.cmd.R
```diff
@@ -0,0 +1,16 @@
...
+library(PEcAn.remote)
+library(testthat)
```
infotroph, line 3: The usual pattern is to load these packages in `tests/testthat.R`, and not in the individual test files.

mdietze (approved):

Issue #3124, opened 2023-03-01
ED2 defaults for allometry overwrite ED2 calculations?


Aariq:
### Bug Description

It appears that `PEcAn.ED2` sets the ED2 flag `NL%IALLOM = 3` in ED2IN by default.  It *also* writes `b1Ht` and `b2Ht` to config.xml.  **BUT** ED2 seems to want to calculate `b1Ht` and `b2Ht` when IALLOM = 3.  I believe that config.xml gets the final say though and overwrites the calculations ED2 is doing.  This could be a problem because `b1Ht` and `b2Ht` mean different things depending on the value for `NL%IALLOM`.  The PEcAn.ED2 defaults appear to be the ones for `NL%IALLOM = 1`.

https://github.com/EDmodel/ED2/blob/b9c93f806d6df7a38db306d944a0695d8b40b39a/ED/src/init/ed_params.f90#L3660-L3664

## Expected behavior

I'd expect PEcAn.ED2 to just not include `b1Ht` and `b2Ht` in config.xml.  Anything not in config.xml will just use the ED2 defaults store in ED.  That would eliminate the need to keep an updated copy of those defaults in PEcAn.ED2.  Either that, or use the default IALLOM mode (2).

## Additional context

Getting divide by zero errors from ED2 and trying to eliminate potential sources.  Allometry seems like a prime suspect.

Aariq:
Update.  Running ED2 without PEcAn.ED2 setting ED2's defaults (i.e. letting ED2 use it's own internal defaults) has fixed all the errors I was getting in ED2 runs.  I suspect most of the errors I have been getting have been a result of PEcAn.ED2 breaking allometry equations by overwriting coefficients with inappropriate values.

Aariq:
I can envision at least 3 possible solutions roughly in order of least to most control from PEcAn (and consequently, least to most difficulty in getting it right)

1) Don't write any paramaters to config.xml that don't have priors. (i.e. what I did in #3125)
2) Go through ED2 code and figure out which parameters mean different things or have different defaults depending on settings in ED2IN and don't write those to config.xml
3) Write different defaults depending on the settings in ED2IN

Aariq:
Another possibility is to make this work with the current machinery by:

1. re-naming ED2IN.r2.2.0.github to ED2IN.rgit
2. updating `history.rgit` so that the defaults there match the defaults that would be used by allometry equations in ED2.

[...]

PR #3125, opened 2023-03-01
Only write params with priors and/or data to config.xml

Aariq:
## Description

Testing out what happens when PEcAn.ED2 doesn't write default constants to config.xml.  An extreme fix for #3124 

## Motivation and Context

Would simplify PEcAn.ED2 code a lot and be less likely to accidentally overwrite constants in ways that break ED2

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


Aariq:
If this is a desirable PR, then I also would need to:
- [ ] edit tests
- [ ] remove history.* datasets
- [ ] update code comments
- [ ] update documentation
- [ ]

mdietze:
While I understand the disadvantages of the existing approach, and agree that things could be done to more easily keep the defaults more up-to-date, the decision to overwrite PFTs entirely was definitely done intentionally. If this approach is implemented I would strongly prefer that it be done as an optional flag rather than as the default behavior.

Aariq:
I understand the motive behind overwriting all the parameters, but there's at least one case where this is having unintended consequences.  For example *at least* `b1Ht`, `b2Ht`, and `hgt_ref` have different defaults (and meanings) depending on the choice of allometry equations set in ED2IN and in some cases these parameters don't have defaults, but are calculated from other parameters (`rho`) in ED2. There's more detail in #3124 as well as some ideas for alternative approaches to this PR.  I'll keep this open as a draft for the time being if that's alright, as I'm using this version of PEcAn.ED2 to wrap up a project, at least until there is a better alternative.

mdietze:
Very few ED2 parameters are actually *calculated* based on other parameters. Most "calculations" are actually empirical correlations among traits that are hard coded into ED2 and in general those equations are old and ignore both uncertainties and correlations with other variables. The decision to avoid ED2's "calculations", and to handle these correlations within PEcAn, was very deliberate.

I understand that some parameters only make sense in combination with specific flags. If individuals updated the default flags without updating parameters that's a genuine error. If one wants to use different flags from the default they should probably set that as a different model version and adjust the parameters & priors accordingly.

Aariq:
If I _wanted_ to use a different value for, say, IALLOM, how would I know which parameters in config.xml I would also have to change for it to work correctly? Would I have to look into the ED2 source code? Could I remove them from config.xml with PEcAn in some way, or would I have to write a script to edit all the config.xml files that PEcAn generates?  I'm just trying to figure out how this is supposed to work.

mdietze:
If you want to PEcAn to use a specific non-default value for an ED2 parameter, rather than sample it from a distribution, then you put them in the `<constants>` section of the pft config https://pecanproject.github.io/pecan-documentation/develop/xml-core-config.html#xml-pft

As for how you know what different modes do, in theory that should be in the ED2 wiki, but I don't think people have been keeping it updated as they've made changes. So yeah, you mostly end up looking at the code.

Aariq:
My work with `PEcAn.ED2` is done, so you can close this if you'd like, but to be clear `PEcAn.ED2` is currently broken without the changes in this PR (#3124), unless some other recent change has addressed this problem.

PR #3127, opened 2023-03-03
Update get.analysis.filenames.r

mdietze:
Discovered today that in some contexts this will throw an error because if outdir is NULL then is.na returns numeric(0) not FALSE.


infotroph (commented):
Good catch!

This reminds me that in the next release of R, `if` will start erroring on conditions with length > 1, meaning anywhere else we see `|` instead of `||` is a place we should probe now for bugs.

infotroph (approved):
Meant to approve before, sorry

PR #3132, opened 2023-03-11
Fix indexing issue when there is only one observation with one state variable for the SDA workflow.

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.sequential/R/sda.enkf_MultiSite.R
+++ b/modules/assim.sequential/R/sda.enkf_MultiSite.R
@@ -149,14 +149,7 @@ sda.enkf.multisite <- function(settings,
   obs.mean <- obs.mean[sapply(lubridate::year(names(obs.mean)), function(obs.year) obs.year %in% (assim.sda))] #checks obs.mean dates against assimyear dates
   obs.cov <- obs.cov[sapply(lubridate::year(names(obs.cov)), function(obs.year) obs.year %in% (assim.sda))] #checks obs.cov dates against assimyear dates
   #checking that there are dates in obs.mean and adding midnight as the time
-  if(settings$state.data.assimilation$forecast.time.step == "year"){
-    obs.times <- PEcAnAssimSequential::obs_timestep2timepoint(settings$state.data.assimilation$Obs_Prep$start.date,
-                                                              settings$state.data.assimilation$Obs_Prep$end.date,
-                                                              list(unit="year", num=1))
-  }else{
-    obs.times <- names(obs.mean)
-  }
-  
+  obs.times <- names(obs.mean)
   obs.times.POSIX <- lubridate::ymd_hms(obs.times)
   for (i in seq_along(obs.times)) {
     if (is.na(obs.times.POSIX[i])) {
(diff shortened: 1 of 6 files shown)


modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -149,7 +149,14 @@ sda.enkf.multisite <- function(settings,
   obs.mean <- obs.mean[sapply(lubridate::year(names(obs.mean)), function(obs.year) obs.year %in% (assim.sda))] #checks obs.mean dates against assimyear dates
   obs.cov <- obs.cov[sapply(lubridate::year(names(obs.cov)), function(obs.year) obs.year %in% (assim.sda))] #checks obs.cov dates against assimyear dates
   #checking that there are dates in obs.mean and adding midnight as the time
-  obs.times <- names(obs.mean)
+  if(settings$state.data.assimilation$forecast.time.step == "year"){
+    obs.times <- PEcAnAssimSequential::obs_timestep2timepoint(settings$state.data.assimilation$Obs_Prep$start.date,
```
mdietze, line 153: Conceptually, I wonder if it would make more sense to do this fix in the assembler (i.e. add more "empty" dates to obs.mean)? To me, it seems cleaner, more flexible, and more clear to users to have the SDA itself always run exactly what days its told to, rather than having to document the behaviour that the SDA function adds in dates magically. The automatic adding in of dates could potentially mess up someone who is trying to run "non-standard" assimilation experiments.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -359,13 +366,18 @@ sda.enkf.multisite <- function(settings,
         restart.list <-
           furrr::future_pmap(list(out.configs, conf.settings %>% `class<-`(c("list")), params.list, inputs.split),
                              function(configs, settings, new.params, inputs) {
+                               #if the new state for each site only has one row/col.
+                               #then we need to convert it to matrix to solve the indexing issue.
+                               new_state <- new.state[, which(attr(X, "Site") %in% settings$run$site$id)]
```
mdietze, line 371: This might work fine because of different scopes, but to me it feels a bit dangerous and simpler to just use a different variable name in the assignment.

modules/assim.sequential/R/sda.enkf_MultiSite.R
DongchenZ: Fixed.

modules/assim.sequential/R/sda.enkf_MultiSite.R
DongchenZ: I don't understand what you mean here.

mdietze: new_state isn't an argument to the function, so you're accessing a global variable, which you're then modifying (subsetting) within the function. In such situations it can be ambiguous whether those changes are local or global, and whether you WANT the changes to be local or global. I think you want the changes to be only local. doing something like `new_state_site = new_site[, which(attr(X, "Site") %in% settings$run$site$id)]` would be more clear and less likely to cause errors related to scope.

DongchenZ: Fixed.

mdietze (approved):

PR #3136, opened 2023-03-17
possibly last fixes for #18

### LINKED ISSUE #18: Remove settings as a global variable

infotroph:
## Description

After this patch, `grep --include="Rcheck_reference.log" -R 'global variable.*settings'` returns no results, while I take to mean this can close #18 before it reaches its decade mark.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.sequential/R/build_X.R
+++ b/modules/assim.sequential/R/build_X.R
@@ -19,7 +19,6 @@
 #' @return X ready to be passed to SDA Analysis code
 #' @export
 #'
-#' @examples
 build_X <- function(out.configs, settings, new.params, nens, read_restart_times, outdir, t = 1, var.names, my.read_restart, restart_flag = FALSE){
   if(t == 1 & restart_flag){
     reads <-
--- a/modules/assim.sequential/tests/Rcheck_reference.log
+++ b/modules/assim.sequential/tests/Rcheck_reference.log
@@ -94,7 +94,6 @@ Contruct.Pf: no visible global function definition for ‘filter’
 Contruct.Pf: no visible binding for global variable ‘Var1’
 Contruct.Pf: no visible binding for global variable ‘Var2’
 EnKF: no visible global function definition for ‘cov’
-EnKF.MultiSite: no visible binding for global variable ‘settings’
 EnKF.MultiSite: no visible binding for global variable ‘site.ids’
 EnKF.MultiSite: no visible binding for global variable ‘blocked.dis’
 GEF: no visible global function definition for ‘logger.warn’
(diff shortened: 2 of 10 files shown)


modules/assim.sequential/R/build_X.R
```diff
@@ -19,7 +19,6 @@
 #' @return X ready to be passed to SDA Analysis code
 #' @export
 #'
-#' @examples
```
Aariq, line 22: Is removing @examples necessary to get the documentation to build or something?  CRAN will require examples I believe.  It might be better to put _something_ here wrapped in `\dontrun{}` rather than remove the roxygen tag entirely?

Aariq: This could probably even be something like :

``` r
#' @examples
#' #TODO: add examples!
```

dlebauer: Is this an internal function (?) if so could remove `@export` and skip examples?

infotroph: I deleted it because an empty `@examples` triggers complaints from Roxygen, but this was an add-on that could be handled in a different PR. Reverting.

infotroph: @helge22a Do you use `build_X` or `sda_matchparam` anywhere outside `sda.enkf.multisite`? If not I'll make it internal. (I'll handle all the code changes, just need confirmation I'm not breaking your workflows when I do it.)

mdietze: @infotroph these can be internal functions

modules/assim.sequential/tests/Rcheck_reference.log
```diff
@@ -729,18 +726,8 @@ File ‘PEcAnAssimSequential/R/Analysis_sda.R’:
...
-  ‘settings’
-
-Undocumented arguments in documentation object 'EnKF'
-  ‘setting’
-Documented arguments not in \usage in documentation object 'EnKF':
-  ‘settings’
-
 Undocumented arguments in documentation object 'GEF'
-  ‘H’ ‘setting’
+  ‘H’
```
mdietze, line 730: If anyone wants to get rid of this error, H is the observation operator matrix

modules/assim.sequential/R/build_X.R
infotroph: Fixed in 33bfb007b

PR #3144, opened 2023-03-21
Added partial namespaces for the SDA workflow; Fixed bugs of Obs_prep function and read && write restart within the SDA workflow.

DongchenZ:
1. Added partial namespaces within the SDA functions and updated the log file accordingly;
2. Fixed bugs where the arguments are not passing correctly during the extract_AGB function calls within the AGB_prep function;
3. Fixed bugs where the AGB is not updating due to empty wood/roots fractions from the read restart function;
4. Fixed bugs where the soil moisture is not updating correctly under the `percentage` unit due to incorrect if control.
5. Fixed bugs so that the `litterInit`, `snowInit`, and `litterWFracInit` are actually written into the param file.
6. Fixed bugs where `na.omit` is not working for the reading restart function.
7. Fixed bugs where the unit conversion is not employed properly when variable not in X.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


models/sipnet/R/write_restart.SIPNET.R
```diff
@@ -104,14 +104,14 @@ write_restart.SIPNET <- function(outdir, runid, start.time, stop.time, settings,
   }
   
   if ("SoilMoistFrac" %in% variables) {
-    analysis.save[[length(analysis.save) + 1]] <- new.state$SoilMoistFrac  ## unitless
-    if (new.state$SoilMoistFrac < 0 || new.state$SoilMoistFrac > 1) analysis.save[[length(analysis.save)]] <- 0.5
+    analysis.save[[length(analysis.save) + 1]] <- new.state$SoilMoistFrac/100  ## unitless
+    if (analysis.save[[length(analysis.save)]] < 0 || analysis.save[[length(analysis.save)]] > 1) analysis.save[[length(analysis.save)]] <- 0.5
```
mdietze, line 108: not following the change in logic here

DongchenZ: So if we use the percentage of Soil moisture profile (0-100) and compare the percentage value against the 0 - 1 range, the soilMoistFrac will not be updated correctly and will be fixed at 0.5 in most cases.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -707,3 +703,12 @@ sda.enkf.multisite <- function(settings,
...
 } # sda.enkf
+
+#' @title loadRData
+#' @param fileName path to the Rdata.
+#' Enable the feature of passing loaded object into new variable.
+loadRData <- function(fileName){
+  #loads an RData file, and returns it
+  load(fileName)
+  mget(ls()[ls() != "fileName"])
+}
```
infotroph, line 714: This looks like a reimplementation of `PEcAn.utils::load_local` -- I recommend using that instead.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -50,7 +50,11 @@ sda.enkf.multisite <- function(settings,
   }else{
     restart_flag = FALSE
   }
-  future::plan(multiprocess)
+  if (future::supportsMulticore()) {
+    future::plan(future::multicore)
+  } else {
+    future::plan(future::multisession)
+  }
```
infotroph, line 57: Apologies if this is documented somewhere already: Does this function provide a way for the user to control how many processes it runs in parallel? The ability to use cores when they're available is great, but having something use lots of cores when you didn't expect it is a lot less great, and CRAN is firm that package code should only run multicore when the user specifically tells it to.

DongchenZ: Fixed. I added a new argument called `run_parallel` in the SDA function.

modules/assim.sequential/R/sda.enkf_MultiSite.R
DongchenZ: I reverted back because this is much more complicated than I thought.

modules/assim.sequential/R/Multi_Site_Constructors.R
```diff
@@ -24,20 +24,20 @@ Contruct.Pf <- function(site.ids, var.names, X, localization.FUN=NULL, t=1, bloc
...
-    pf.matrix [pos.in.matrix, pos.in.matrix] <- cov( X [, pos.in.matrix] ,use="complete.obs")
+    pf.matrix [pos.in.matrix, pos.in.matrix] <- stats::cov( X [, pos.in.matrix] ,use="complete.obs")
   }
   
   # This is where we estimate the cov between state variables of different sites
   #I put this into a sperate loop so we can have more control over it
-  site.cov.orders <- expand.grid(site.ids,site.ids) %>%
-                        filter( Var1 != Var2)
+  site.combs <- expand.grid(site.ids,site.ids)
+  site.cov.orders <- site.combs[which(site.combs$Var1 != site.combs$Var2),]
```
mdietze, line 33: Not sure why you got rid of the `filter` here. I'd recommend putting it back because the original code was cleaner and easier to understand

modules/assim.sequential/tests/Rcheck_reference.log
```diff
@@ -602,22 +588,10 @@ sda.enkf: no visible global function definition for
...
-  for ‘input.ens.gen’
-sda.enkf.multisite : <anonymous>: no visible global function definition
-  for ‘write.ensemble.configs’
-sda.enkf.multisite : <anonymous>: no visible binding for global
-  variable ‘ensemble.samples’
-sda.enkf.multisite: no visible global function definition for
-  ‘setNames’
-sda.enkf.multisite: no visible global function definition for ‘map_dfc’
+sda.enkf.multisite: no visible global function definition for ‘write.ensemble.configs’
+sda.enkf.multisite: no visible global function definition for ‘input.ens.gen’
```
mdietze, line 594: do these two really need to be added? Above it looked like you provided the namespace for both

DongchenZ: Those two should be deleted.

mdietze (approved):

PR #3156, opened 2023-03-31
Updated `write.configs.ed` to throw an error if `revision` not specified

### LINKED ISSUE #3142: write.configs.ed should error if revision not specified
> I understand the logic of this bit, and I know you didn't add the line @Aariq, but I wonder if it would be better for the code to throw an error if the revision isn't specified. In general we're trying to reduce/isolate the dependency of pecan modules and functions on the database -- by fixing this one error-checking line we're making users of the ED2 module install the database.

_Originally posted by @mdietze in https://github.com/PecanProject/pecan/pull/3135#discussion_r1142050704_
            
☝️ In reference to the following line of code in models/ed/write_configs.ed.R: 
```
 model <- PEcAn.DB::db.query(paste("SELECT * FROM models WHERE id =", settings$model$id), params = settings$database$bety)
```

meetagrawal09:
## Description
This PR fixes #3142 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/Makefile.depends
+++ b/Makefile.depends
@@ -28,7 +28,7 @@ $(call depends,models/cable): | .install/base/logger .install/base/utils
 $(call depends,models/clm45): | .install/base/logger .install/base/utils
 $(call depends,models/dalec): | .install/base/logger .install/base/remote .install/base/utils
 $(call depends,models/dvmdostem): | .install/base/logger .install/base/utils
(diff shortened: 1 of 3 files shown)


meetagrawal09:
@mdietze, @Aariq should I turn to using `logger.error` in-place of `logger.severe` for the fix ?

mdietze (approved):

PR #3158, opened 2023-04-06
Refinements and bug fixes for the current SDA workflow.

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.sequential/R/Multi_Site_Constructors.R
+++ b/modules/assim.sequential/R/Multi_Site_Constructors.R
@@ -172,12 +172,12 @@ block_matrix <- function (x = NULL, b = NULL, byrow = FALSE, dimnames = NULL) {
 ##' @return Returns a matrix with block sizes determined by the b argument. Each block is filled with the same value taken from x.
 ##' @export
 Construct.H.multisite <- function(site.ids, var.names, obs.t.mean){
-  
-  site.ids.matrix <- sort(rep(site.ids, length(var.names)))
+  #we first create a matrix containing site.ids, var.names, observations, and the index of observations across obs.mean.
+  site.ids.matrix <- rep(site.ids, each = length(var.names))#this is 
   var.names.matrix <- rep(var.names, length(site.ids))
   H.pre.matrix <- data.frame(site.ids.matrix, var.names.matrix, NA, NA) %>% `colnames<-` (c("site.id", "var.name", "obs", "obs.ind"))
   obs.ind <- 1
-  for (i in 1:dim(H.pre.matrix)[1]) {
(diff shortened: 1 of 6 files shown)


modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -419,8 +418,8 @@ sda.enkf.multisite <- function(settings,
         
         #I'm rewrting the runs because when I use the parallel appraoch for wrting configs the run.txt will get messed up; because multiple cores want to write on it at the same time.
         runs.tmp <- list.dirs(rundir, full.names = F)
+        runs.tmp <- runs.tmp[grepl("ENS-*", runs.tmp)]
```
mdietze, line 421: Not sure about this change. Depending on how the pecan.xml and bety are set-up you I'm not sure you can depend on this naming of ensemble members. Ultimately, a parrallel-safe approach to writing the runs list would be a better solution (e.g. data.frame::fwrite with append=TRUE looks promising)

modules/assim.sequential/R/Multi_Site_Constructors.R
```diff
@@ -172,45 +172,28 @@ block_matrix <- function (x = NULL, b = NULL, byrow = FALSE, dimnames = NULL) {
...
-    for (n in seq_along(choose))
-    {
-      H.this.site[n, choose[n]] <- 1
-      #sapply(choose[n], agrep, x = seq_along(var.names), max = 1, USE.NAMES = FALSE) %>% unlist
-      #H.this.site[n, choose.col] = 1
+  site.ids.matrix <- sort(rep(site.ids, length(var.names)))
+  var.names.matrix <- rep(var.names, length(site.ids))
+  H.pre.matrix <- data.frame(site.ids.matrix, var.names.matrix, NA, NA) %>% `colnames<-` (c("site.id", "var.name", "obs", "obs.ind"))
+  obs.ind <- 1
+  for (i in 1:dim(H.pre.matrix)[1]) {
```
mdietze, line 180: This is probably fine for now, but (1) doing this with a for loop will be slow when the size of obs.mean gets big and (2) coming down the pipeline pretty soon is the case where H will need to include fractions (e.g. multiple model patches located within a single MODIS pixel) so we're going to have to generalize the building of H significantly.

mdietze, line 180: Also, seq_len(nrow(H.pre.matrix)) is simpler and safer than 1:dim(H.pre.matrix)[1], which will fail in a nonintuitive way if nrow = 0

mdietze (approved):

PR #3162, opened 2023-04-26
Added a new SDA plotting function.

DongchenZ:
## Description

The new function surpassed the previous version in the following ways:
1. It creates figures based on the alphabetic order of forecast, analysis, and observations.
2. It provides user-defined plotting order by each site, each state variable, or each PFT if the pft.csv path is provided.
3. It provides a user-defined plotting style (so far only colors) for creating figures.
4. The function also creates a database of forecasts, analyses, and observations in a more scalable and generalizable way for future ggplot purposes.
5. For the site option, each page of the PDF file will contain all state variables for each site.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/assim.sequential/R/sda_plotting.R
```diff
@@ -845,3 +845,143 @@ post.analysis.multisite.ggplot <- function(settings, t, obs.times, obs.mean, obs
...
+      return(vector)
+    }
+  }
+  #read pft.csv file for the option by == pft.
+  if(!is.null(pft.path)){
+    pft <- read.csv(pft.path)
+  }
+  #create database
+  DB <- data.frame()
+  for (id in sort(unique(site_ids))) {
```
mdietze, line 875: not essential, but I suspect a `lapply` or `map` approach would be a lot quicker than this level of nested looping. I also bet that within a timepoint and site you could handle the variable reorganization with pivot_wider and/or pivot_longer

mdietze (approved):

base/workflow/NAMESPACE
```diff
@@ -2,11 +2,14 @@
 
 export(create_execute_test_xml)
 export(do_conversions)
+export(merge_job_files)
 export(model_specific_tags)
+export(qsub_parallel)
```
mdietze, line 7: wasn't there a suggestion to move this to `remote`?

PR #3166, opened 2023-05-01
use pecan/models to build

robkooper:
Use the pecan/models image to build. This will make sure we use the same version of the OS to build the models as the one that will run the models.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/Dockerfile
+++ b/models/sipnet/Dockerfile
@@ -4,18 +4,10 @@ ARG IMAGE_VERSION="latest"
 # ----------------------------------------------------------------------
 # BUILD SIPNET BINARY
 # ----------------------------------------------------------------------
-FROM debian:stretch as sipnet-binary
+FROM pecan/models:${IMAGE_VERSION} as model-binary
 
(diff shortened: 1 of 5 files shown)


models/sipnet/Dockerfile
```diff
@@ -4,19 +4,11 @@ ARG IMAGE_VERSION="latest"
...
 ARG MODEL_VERSION=git
 
-# install dependencies
-RUN apt-get update \
-    && apt-get install -y \
-       gcc \
-       git \
-       make \
-    && rm -rf /var/lib/apt/lists/*
-
```
infotroph, line 19: Am I right these are deleted because they're all already installed on the pecan/models image?

infotroph (approved):

models/sipnet/Dockerfile
robkooper: correct

mdietze (approved):

PR #3168, opened 2023-05-16
Allow `ud_convert()` to work with class "difftime" (fix for #3012)

### LINKED ISSUE #3012: units::set_units cannot work with class "difftime"
### Bug Description
Recent `udunits2` to `units` change introduced a bug to one of our common met process pipeline. [This met2CF.csv line ](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/met2CF.csv.R#L246)is now throwing an error:
```
Error in UseMethod("set_units") : 
  no applicable method for 'set_units' applied to an object of class "difftime"
```

### To Reproduce
Try the demo run or see:

```
days_since_1700 <- as.POSIXct(c("2003-01-01 00:00:00 UTC", "2003-01-01 00:30:00 UTC")) - lubridate::ymd_hm("1700-01-01 00:00")
as.numeric(mean(PEcAn.utils::ud_convert(diff(days_since_1700), "d", "s")))
```

### Proposed solution

We can change class from `difftime` to `numeric` before conversion:
```
mean(PEcAn.utils::ud_convert(as.numeric(diff(days_since_1700)), "d", "s"))
```
or any other solutions are welcome.

Aariq:
## Description

Fixes #3012 .  Checks if argument `x` is difftime, and if it is converts it to a `units` object with `as_units()`.  Units are inferred from the class, so argument `u1` isn't actually used in the conversion in this case.  UI is a little weird here, but seems like this is mostly an internal function so probably better than adding argument defaults and allowing `u1` to be `NULL` or something like that.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/utils/R/ud_convert.R
+++ b/base/utils/R/ud_convert.R
@@ -15,6 +15,9 @@ ud_convert <- function(x, u1, u2) {
   stopifnot(units::ud_are_convertible(u1, u2))
   if(inherits(x, "difftime")) {
     x1 <- units::as_units(x)
+    if(units::units(x1) != units::units(units::as_units(u1))) {
+      warning("Units of `x` don't match `u1`, using '", units::deparse_unit(x1), "' instead")
+    }
   } else {
     x1 <- units::set_units(x, value = u1, mode = "standard")
   }
--- a/base/utils/R/ud_convert.R
+++ b/base/utils/R/ud_convert.R
@@ -15,7 +15,7 @@ ud_convert <- function(x, u1, u2) {
   stopifnot(units::ud_are_convertible(u1, u2))
   if(inherits(x, "difftime")) {
     x1 <- units::as_units(x)
-    if(units::units(x1) != units::units(units::as_units(u1))) {
+    if(units(x1) != units(units::as_units(u1))) {
       warning("Units of `x` don't match `u1`, using '", units::deparse_unit(x1), "' instead")
     }
   } else {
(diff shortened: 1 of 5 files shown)


dlebauer (changes requested):
Looks great- just make sure to update the change log!

base/utils/R/ud_convert.R
```diff
@@ -3,15 +3,21 @@
...
+##'   to be supplied and needs to be convertible to `u2` for consistency.
 ##' @param u2 string parseable as the units to convert to
 ##' 
 ##' @return numeric vector with values converted to units in `u2`
 ##' @export
 ud_convert <- function(x, u1, u2) {
   stopifnot(units::ud_are_convertible(u1, u2))
-  x1 <- units::set_units(x, value = u1, mode = "standard")
+  if(inherits(x, "difftime")) {
+    x1 <- units::as_units(x)
```
infotroph, line 17: I'm nervous about ignoring `u1` here -- I don't know how often units(x) and u1 will mismatch in practice, but when they do it means at a minimum that the user's expectations are wrong and the earlier we alert them the better.

Maybe something like...?
```suggestion
    stopifnot(units::as_units(units(x)) == units::as_units(u1))
    x1 <- units::as_units(x)
```

Aariq:
> Looks great- just make sure to update the change log!

I can do this, but I'll just point out that the changelog already includes a bullet about this function being re-factored for this version.  This bug was an unforseen side-effect of that refactoring

infotroph (approved):

dlebauer (approved):

PR #3176, opened 2023-06-02
[feat]: Reduce Workflow Time for CI GitHub Action

RohanSasne:
## Description

 
In this PR, I have created 3 Makefiles for base, module and model, also I have updated the CI.yml such that these checks will run in parallel, there by reducing the time from nearly [31 minutes](https://github.com/PecanProject/pecan/actions/runs/5132766441) to roughly around [17 to 20 minutes](https://github.com/RohanSasne/pecan/actions/runs/5152546372), the main checks for base,module and model would take around 17 - 18 minutes and the extra 1 - 2 minutes would be required to download the utilities (Which totally depends on the environment). 

## Motivation and Context

Currently GitHub Actions will check to see if there are newer versions of the packages installed.
Biggest problem was that right now we run out of minutes and compiles take a lot longer than needed.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [x ] I have added tests to cover my changes.
- [ x] All new and existing tests passed.

 
cc: @robkooper

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+      - Reduce-Workflow-Time
 
     tags:
       - '*'
@@ -82,7 +83,7 @@ jobs:
   # ----------------------------------------------------------------------
   # R CHECK
   # ----------------------------------------------------------------------
-  check:
+  check_base:
     if: github.event_name != 'issue_comment' || startsWith(github.event.comment.body, '/build')
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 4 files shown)


mdietze (commented):
One thing I'm worried about with this PR is that it appears to creates two versions of the same thing, the original makefile and then this new set, that will need to maintained in parallel. Is there a way to have the original make leverage these three new files?

infotroph:
If I'm understanding the approach right, think you could achieve the same effect by defining new phony targets in the existing Makefile, one for each operation on a subset of packages: 

```
--- a/Makefile
+++ b/Makefile
@@ -96,10 +96,20 @@ depends = .doc/$(1) .install/$(1) .check/$(1) .test/$(1)
 
 ### Rules
 
-.PHONY: all install check test document shiny
+.PHONY: all install check test document shiny \
+               install_base install_models install_modules \
[snip snip snip]
+               doc_base doc_models doc_modules
 
 all: install document
 
+install_base: $(BASE_I)
+install_models: $(MODELS_I)
[snip snip snip]
+doc_models: $(MODELS_D)
+doc_modules: $(MODULES_D)
+
 document: $(ALL_PKGS_D) .doc/base/all
 install: $(ALL_PKGS_I) .install/base/all
 check: $(ALL_PKGS_C) .check/base/all
```

Then each operation can call the command it needs, e.g. `make install_base` or `make doc_modules`.

infotroph:
Followup thought: there's a good chance we don't need all 12 of those like my previous comment implies. Maybe for CI purposes it's sufficient to define `check_*` and leave install/test/doc as single commands that work on all packages.

RohanSasne:
thank you @mdietze , @infotroph  for reviewing this PR, using phony targets is surely a workaround, I have made the necessary changes and also removed the individual makefiles. Do check the latest changes once :)

Makefile
```diff
@@ -96,10 +96,16 @@ depends = .doc/$(1) .install/$(1) .check/$(1) .test/$(1)
 
 ### Rules
 
-.PHONY: all install check test document shiny
+.PHONY: all install check test document shiny \
++               check_base check_models check_modules
```
mdietze, line 100: Is the + here needed?

Makefile
```diff
@@ -96,10 +96,16 @@ depends = .doc/$(1) .install/$(1) .check/$(1) .test/$(1)
...
 
 all: install document
 
+
+check_base: $(BASE_C) 
+check_models: $(MODELS_C) 
+check_modules: $(BASE_C) $(MODULES_C)
```
mdietze, line 107: This seems to imply that modules depends on base, but models doesn't. Is this true? Doesn't make know the dependencies among packages? (e.g., if models already know which subset of base it needs to build, doesn't modules also). And these seem to be checks, not just builds -- running a check on all the base packages before running the modules checks seems redundant with the check_base

RohanSasne: hey @mdietze, I had a discussion with @robkooper on google meet few days back and The problem with modules is that, it has a circular dependency with base, which requires the base check to run first, also we have kept check_base and check_modules separately as we further plan to trigger the checks only when changes are made to those specific directories. So it would be wise for now to keep check_base and check_modules separate for this PR.

mdietze: Yes but doesn't the circular dependency have to do with running the *install* for base, not on running the *checks*. Just installing base would be quicker than also running the checks on all the base packages.

RohanSasne: Oh, yes you are right sir, I tried running the workflow with [installing the base packages and then running the check on modules](https://github.com/RohanSasne/pecan/actions/runs/5163977571) and it is surely quicker than before, I have made the necessary changes, please have a look. :)

Makefile
RohanSasne: typo from my end, it's resolved now

mdietze (approved):

.github/workflows/ci.yml
```diff
@@ -131,7 +131,121 @@ jobs:
 
     # run PEcAn checks
     - name: check
-      run: make -j1 check
+      run: make -j1 check_base
```
infotroph, line 134: I bet we could further reduce code duplication here by setting the check command as a matrix variable, but that can be a separate PR -- what you have here is already a big improvement!

infotroph (approved):
Thank you! A couple more minor suggestions, but I think this is basically ready to go.

Makefile
```diff
@@ -180,4 +186,4 @@ $(ALL_PKGS_I) $(ALL_PKGS_C) $(ALL_PKGS_T) $(ALL_PKGS_D): | .install/devtools .in
 # Install dependencies declared by Shiny apps
 .shiny_depends/%: $$(call files_in_dir, %) | $$(@D)
 	Rscript scripts/install_shiny_deps.R $(subst .shiny_depends/,shiny/,$@)
-	echo `date` > $@
+	echo `date` > $@
```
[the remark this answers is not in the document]
RohanSasne: Oh didn't know about this, I will keep this in mind moving forward :sweat_smile:

Makefile
RohanSasne: Cool, I have added comments to address why we install BASE first and then run check on modules.

.github/workflows/ci.yml
RohanSasne: Thank you :smile:

.github/workflows/ci.yml
```diff
@@ -201,4 +315,4 @@ jobs:
 
     # run SIPNET test
     - name: integration test
-      run: ./tests/integration.sh ghaction
+      run: ./tests/integration.sh ghaction
```
infotroph, line 318:
```suggestion
      run: ./tests/integration.sh ghaction
```

PR #3183, opened 2023-06-11
build: Test Models Against the Full Docker Stack

### LINKED ISSUE #3184: Should the new SIPNET test run on every PR or should it be scheduled on weekly basis
## Description

**Is your feature request related to a problem? Please describe.**

We are creating a YAML file workflow which will test the current SIPNET model against the full docker stack brought up by docker compose.
It is resource intensive as we pull all the images in the docker compose and build them.

The testing workflow can be found in the draft PR: #3183 

## Proposed Solution

So should we schedule it for weekly basis or let it run on every PR ?

I was of the opinion to run it weekly, but open to the suggestions from the community.

cc. @robkooper @mdietze @infotroph

allgandalf:
## Description

In this PR, we have used the full docker stack to test SIPNET by using Docker Compose, this test will run weekly on Mondays. 

## Motivation and Context

From Discussions in #3184, It would be wise to have a script file and have our cURL POST requests for different models in it as the Matrix option would have to bring up the whole docker stack for each job of the Matrix. 

In this PR, we have added the SIPNET model test, which outputs the sipnet.out **_(model output_**) file, [Test results can be found in this workflow](https://github.com/PecanProject/pecan/actions/runs/5272813345/jobs/9535516799)

In the script file, template has been provided to correctly input data into the cURL request so that in future, People can add more models in the script file against which tests can be performed.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

This PR closes #3184


.github/workflows/docker-stack-test.yml
```diff
@@ -0,0 +1,150 @@
...
+  CI: true
+
+jobs:
+ 
+
+  # ----------------------------------------------------------------------
+  # DOCKER STACK TESTS
+  # ----------------------------------------------------------------------
+  Stack-Test:
+    if: github.event_name != 'issue_comment' || startsWith(github.event.comment.body, '/build')
```
infotroph, line 36: Does the `/build` comment command need to run these tests? If not, these `if`s could likely be dropped.

RohanSasne: Updated the file accordingly :)

docker-compose.yml
```diff
@@ -300,6 +300,7 @@ services:
   api:
     image: pecan/api:${PECAN_VERSION:-latest}
     user: "${UID:-1001}:${GID:-1001}"
+    restart: unless-stopped
```
RohanSasne, line 303: It can be useful to automatically restart the api container if it fails when we bring up the whole docker stack

tests/Docker-Stack-Test.sh
```diff
@@ -0,0 +1,47 @@
...
+#     -F 'email=' \
+#     -F 'notes=' \
+# 'http://172.17.0.1/pecan/04-runpecan.php'
+
+# Adjust the sleep time according to the model you are running
+
+# Append the workflowid and ID after `out/` by 1 for every consecutive run
+
+# curl -v -L -H "Host: pecan.localhost" \
+# 'http://172.17.0.1/pecan/dataset.php?workflowid=(ID)&type=file&name=out/(ID)/sipnet.out'
```
infotroph, line 47: These lines seem redundant with the ones above
```suggestion
```

tests/Docker-Stack-Test.sh
```diff
@@ -0,0 +1,47 @@
...
+    -F 'input_met=CRUNCEP' \
+    -F 'email=' \
+    -F 'notes=' \
+'http://172.17.0.1/pecan/04-runpecan.php'
+
+sleep 200
+
+curl -v -L -H "Host: pecan.localhost" \
+'http://172.17.0.1/pecan/dataset.php?workflowid=99000000001&type=file&name=out/99000000001/sipnet.out'
+
```
infotroph, line 19: Add a step here verifying that sipnet.out is free from errors / contains the expected output values?

infotroph (changes requested):

infotroph:
Left some comments above focused on technical implementation, but from a broader perspective it's worth making sure we're aligned that "Whole-docker-stack SIPNET test via GitHub Actions every Monday" is a good approach. 

Pinging @meetagrawal09 for thoughts on how this aligns/interacts with your planned CI development this summer

meetagrawal09:
@infotroph Some points that come to my mind when I check this implementation are : 

1. Should we be more selective as to which containers we are running as part of this full docker stack rather than running all of them at once.

2. We are restricting ourselves with the features provided by the PEcAn API by using curl queries to trigger model runs. A more extensible approach seems to be by using a `settings` file and triggering workflows by running the `workflow.R` script.

3. We can run multiple models here as part of the script in the same job but I prefer having separate models to be running as separate jobs. E.g. Here I can run `SIPNET` and `ED2` as part of the shell script but they both are part of the same job. In case `SIPNET` fails to run correctly this entire job fails without even trying if `ED2` was able to run or not. Rather if I configure them as 2 separate jobs we can analyse issues with each model separately. 
By creating a new job for each model run we are separating out concerns with their failures and also reducing time on each run. It also gives us the liberty to choose which model we want to trigger a run for as part of the CI.

allgandalf:
Closing this PR in favour of https://github.com/PecanProject/pecan/pull/3345, please reopen if you disagree

PR #3193, opened 2023-07-10
Hf landscape updates [WIP]

mdietze:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


computate (approved):

base/workflow/R/run.write.configs.R
```diff
@@ -34,68 +35,82 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
...
+      ## Sample parameters
+      PEcAn.uncertainty::get.parameter.samples(settings, posterior.files, ens.sample.method)
+      load(file.path(settings$outdir, "samples.Rdata")) ## loads ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples
+      
+      ## end use.existing.samples == FALSE
+    } ## is.logical(use.existing.samples)
   } else {
-    PEcAn.logger::logger.error(samples.file, "not found, this file is required by the run.write.configs function")
+    ### use.existing.samples must be a filename
+    load(use.existing.samples)
```
infotroph, line 111: I did a doubletake here ("wait, why is it calling load on a bool?") even after reading at the top of the function that `use.existing.samples` could be a filepath.

Possible alternate design: Call this param `existing.sample.file`, have the argument always be a path, and make the default value be NULL (behaving like FALSE does now). Then passing `existing.sample.file = "samples.Rdata"` would behave like TRUE does now with no need for a special case.

base/workflow/R/run.write.configs.R
```diff
@@ -34,68 +35,82 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
...
+#  PEcAn.uncertainty::get.parameter.samples(settings, posterior.files, ens.sample.method)
+#  samples.file <- file.path(settings$outdir, "samples.Rdata")
+#  if (file.exists(samples.file)) {
+#    samples <- new.env()
+#    load(samples.file, envir = samples) ## loads ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples
+#    trait.samples <- samples$trait.samples
+#    ensemble.samples <- samples$ensemble.samples
+#    sa.samples <- samples$sa.samples
+#    runs.samples <- samples$runs.samples
+#    ## env.samples <- samples$env.samples
```
infotroph, line 51: I strongly recommend keeping this pattern of loading all Rdata files into their own environment -- not only does it keep the package checks from complaining about variables with no visible binding, it means there's no ambiguity about what names were loaded from which file.

The hack of reading the file into an environment and then immediately assigning back out to separate variables was just to minimize lines changed while quieting the check note.  I'd support a further refactor to use all of these directly from thee loading environment instead (e.g. write all references to `sa.samples` as `samples$sa.samples`)

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,16 +2,17 @@
...
-runModule.run.write.configs <- function(settings, overwrite = TRUE) {
+runModule.run.write.configs <- function(settings, overwrite = TRUE, use.existing.samples = TRUE) {
 
   if (PEcAn.settings::is.MultiSettings(settings)) {
     if (overwrite && file.exists(file.path(settings$rundir, "runs.txt"))) {
       PEcAn.logger::logger.warn("Existing runs.txt file will be removed.")
       unlink(file.path(settings$rundir, "runs.txt"))
     }
-    return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = FALSE))
+    return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = overwrite,use.existing.samples=use.existing.samples))
```
infotroph, line 15:
```suggestion
    return(PEcAn.settings::papply(
      settings,
      runModule.run.write.configs,
      overwrite = overwrite,
      use.existing.samples = use.existing.samples))
```

modules/assim.sequential/R/Analysis_sda_multiSite.R
```diff
@@ -359,8 +360,14 @@ GEF.MultiSite <- function(settings, Forecast, Observed, H, extraArg,...){
...
                nthin = extraArg$nthin,
-               monitors = c("Xall", "qq"))
-  outputs <- furrr::future_map(lapply(rep("data", as.numeric(settings$state.data.assimilation$chains)), get), MCMC_function)
+               monitors = c("Xall", "qq"),
+               t=t)
+  tmp = list()
+  for(i in seq_len(as.numeric(settings$state.data.assimilation$chains))){
+    tmp[[i]] = MCMC_data
+  }
+  MCMC_data = tmp
```
infotroph, line 369: Is this equivalent to `MCMC_data = rep(list(MCMC_data), settings$state.data.assimilation$chains)`? That form would make it clearer to me that this is a single list of identical copies of the same list.
I haven't tested, but `rep` is probably more memory efficient too, because it can be allocated once and doesn't need to grow

base/workflow/R/run.write.configs.R
mdietze: to be honest, this was me trying to resolve a merge conflict and not having time to go back and check the code in more detail (it's been months since I worked on this PR). My plan had been to take a closer look at this function and once I had pulled in the latest code

infotroph:
Status: @mdietze is working with Yaang to replace this PR with a new approach. Leaving open until then.

PR #3200, opened 2023-07-22
fix check warnings/notes in assim.batch package

moki1202:


mdietze (approved):

mdietze (approved):

mdietze (changes requested):
Rcheck_reference.log

infotroph:
@moki1202 Looks like this is just waiting for an updated Rcheck_reference.log

moki1202:
> @moki1202 Looks like this is just waiting for an updated Rcheck_reference.log

@infotroph @mdietze done!

modules/assim.batch/tests/Rcheck_reference.log
```diff
@@ -140,9 +71,32 @@ File ‘PEcAn.assim.batch/R/pda.get.model.output.R’:
  OK
 * checking for unstated dependencies in vignettes ... OK
 * checking package vignettes in ‘inst/doc’ ... OK
-* checking re-building of vignette outputs ... OK
+* checking re-building of vignette outputs ... WARNING
```
mdietze, line 74: I understand that this gets us around the isFalse issue, but I don't feel great about switching a reference log from an OK to a WARNING (the idea of the log was to only move in the direction towards fewer warnings)

modules/assim.batch/tests/Rcheck_reference.log
```diff
@@ -87,37 +38,17 @@ use conditionally.
...
+* checking dependencies in R code ... OK
 * checking S3 generic/method consistency ... OK
 * checking replacement functions ... OK
 * checking foreign function calls ... OK
 * checking R code for possible problems ... NOTE
-get.da.data: no visible binding for global variable ‘ensemble.samples’
-get.da.data : <anonymous>: no visible binding for global variable
-  ‘sa.samples’
-get.da.data: no visible binding for global variable ‘sa.samples’
 get.da.data : <anonymous>: no visible global function definition for
```
mdietze, line 46: Not required for the PR, but I'm curious what's up with `read.output.type` and why this one undefined variable was left?

moki1202: @mdietze I think I remember this. I was unable to find which package does read.output.type function belong to.

mdietze: Looking at those functions, I suspect `get.da.data.growth.R` and `get.da.data.R` can both be moved from the `/R` folder to the `/inst` folder. I don't have time to go through all the code in detail, but I think most of what's here is redundant with newer code by @istfer and others. This code doesn't appear to be operationally linked to other bits of the PDA and the value of retaining this code is primarily because it provides a record of what Carl Davidson did in his (still unpublished) Master thesis.

modules/assim.batch/tests/Rcheck_reference.log
```diff
@@ -140,9 +71,32 @@ File ‘PEcAn.assim.batch/R/pda.get.model.output.R’:
...
+Error: processing vignette 'MultiSitePDAVignette.Rmd' failed with diagnostics:
+The function xfun::isFALSE() will be deprecated in the future. Please consider using base::isFALSE(x) or identical(x, FALSE) instead.
+--- failed re-building ‘MultiSitePDAVignette.Rmd’
+
+SUMMARY: processing the following files failed:
+  ‘AssimBatchVignette.Rmd’ ‘MultiSitePDAVignette.Rmd’
+
+Error: Vignette re-building failed.
+Execution halted
+
```
infotroph, line 92: @moki1202 I think we can manually re-delete this warning -- I'm still not fully sure why xfun::isFALSE thing keeps coming back, but we can deal with it separately without having it in Rcheck_reference.log.
```suggestion
* checking re-building of vignette outputs ... OK

```

mdietze (approved):

PR #3203, opened 2023-07-27
Added Integration Tests for `download.CRUNCEP`

meetagrawal09:
## Description
This PR aims to add integration tests for the data item : `CRUNCEP`

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


mdietze (commented):
Seems like a good start, but an integration test checking that we can download and correctly process the CRUNCEP met data is much more important than a unit test checking if we download the land/water mask, which is a really small part of the workflow and which doesn't need to be updated (i.e. we could download it once and then use that one version forever after).

PR #3208, opened 2023-08-06
Added Integration Tests for `download.AmerifluxLBL`

meetagrawal09:
## Description
This PR aims to add integration tests for the data item : `AmerifluxLBL`

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
+++ b/modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
@@ -1,0 +1,165 @@
+library(testthat)
+
+tmpdir <- tempfile(pattern = "amerifluxData")
+dir.create(tmpdir)
+on.exit(teardown(unlink(tmpdir, recursive = TRUE)))
+
+sitename <- "US-Akn"
+# outfolder <- "./amerifluxData"
+outfolder <- tmpdir
+start_date <- "2011-01-01"
+end_date <- "2011-12-31"
+username <- "pecan"
+useremail <- "@"
+data_product <- "BASE-BADM"
+data_policy <- "CCBY4.0"
+overwrite <- FALSE
+verbose <- FALSE
+
+
+start_date <- as.POSIXlt(start_date, tz = "UTC")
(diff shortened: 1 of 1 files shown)


mdietze (changes requested):
While this PR and the other met download PRs do represent useful tests of whether the underlying APIs we use are responding, they aren't actually integration tests. I'm not even sure they count as unit tests since they're not actually testing the download functions that we're currently using. (aside: I wonder if it'd be possible, once files are downloaded in the current piecemeal fashion, to use them in mocking the responses to the overall function unit tests). More importantly, they current unit tests are definitely not integration testing the larger met.process workflow. As such, I worry that we may be missing opportunities to identify the larger issues in that workflow and are investing time in fixing things that may not be the parts that are actually broken.

modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
```diff
@@ -0,0 +1,165 @@
+library(testthat)
+
+tmpdir <- tempfile(pattern = "amerifluxData")
+dir.create(tmpdir)
+on.exit(teardown(unlink(tmpdir, recursive = TRUE)))
```
Aariq, line 5: FYI `teardown()` is superseded.  It is now recommended to use the `withr` package for this kind of thing.  This is still valid code though and will continue to work.

modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
```diff
@@ -0,0 +1,31 @@
+# putting logger to debug mode
+PEcAn.logger::logger.setUseConsole(TRUE, FALSE)
+on.exit(PEcAn.logger::logger.setUseConsole(TRUE, TRUE), add = TRUE)
```
Aariq, line 3: Not sure that `on.exit()` works as expected outside of a function.  It will trigger only when an enclosing function environment is exited I think so this may not happen until after all tests are run if run with `test()`

mdietze (approved):

PR #3213, opened 2023-09-06
Adding Workflow to run Integration Tests

meetagrawal09:
## Description
This PR aims at adding a new workflow to run integration tests for `data.atmosphere` download functions. These tests are to be scheduled weekly.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,21 @@
+name : Integration Tests
+on : 
+  schedule:
+  - cron: '30 4 * * 1'
```
mdietze, line 4: Right now this is set to run at the exact same time as the other weekly integration tests, which is going to max-out our time and download quotas. I'd recommend changing to a different day of the week. I'd also recommend modifying to include "workflow_dispatch", which should allow us to manually trigger the tests (which, if nothing else, is really helpful for debugging github actions. i.e. the top of this file should look like 
```
on:
  workflow_dispatch:
  schedule:
```

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,21 @@
...
+  - cron: '30 4 * * 1'    
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+    
+    container: 
+      image: pecan/depends:R4.1
```
mdietze, line 13: You need to either change this to an image that contains the current full pecan install or you have to install the latest version of pecan. The pecan/depends container just contains all the libraries and R packages that pecan needs to build, not pecan itself. Check with @robkooper but I think you want to use pecan/base or pecan/executor (Rob, note that none of these tests need the models themselves, just the pecan libraries)

mdietze (changes requested):

.github/workflows/integration-test.yml
meetagrawal09: Changed the day to Thursday.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+    - name: Checkout source code
+      uses: actions/checkout@v3
+    - name: Run tests
+      run: |
+        for FILE in modules/data.atmosphere/inst/integrationTests/*; do
+          Rscript $FILE;
+        done
```
RohanSasne, line 30:
```suggestion
    - name: Install Packages
      run: make -j1 install
      
    - name: Run tests
      run: |
        for FILE in modules/data.atmosphere/inst/integrationTests/*; do
          Rscript $FILE;
        done
```

Let us try with install only the libraries

RohanSasne:
The updated workflow installs all the required libraries, now can you figure out why we get the follow [error](https://github.com/meetagrawal09/pecan/actions/runs/6135784582/job/16649887912#step:5:1)? :

```bash
Error: '"stub"' is not an exported object from 'namespace:mockery'
Execution halted
Error: Process completed with exit code 1.
````

File: .github/workflows/integration-test.yml (hunk 2 above)
Comment (RohanSasne) [reply .github/workflows/integration-test.yml:21]: Do you think that, we should create a phony object in the makefile to only install base and modules and skip the models ? The current make command will install all the 3 during the workflow.

**Comment by meetagrawal09** [comment]
`mockery::stub` is something that a lot of our unit tests depend upon and all of those seem to work fine on CI.

File: .github/workflows/integration-test.yml [hunk 4]
```diff
@@ -0,0 +1,35 @@
...
+    container: 
+      image: pecan/base:develop
+
+    steps:
+    - name: Checkout source code
+      uses: actions/checkout@v3
+
+    - name: Install mockery
+      run: |
+        Rscript -e "install.packages('mockery', repos='https://ftp.belnet.be/mirror/CRAN')"
```
infotroph, line 29: I don't /think/ this is what's causing the CI failure for this PR, but is this mirror known to be reliable and up-to-date? In general I recommend leaving `repos` unset wherever unless there's a specific reason to override it, so that all installations use the mirror(s) configured by the base image.

meetagrawal09: Actually without this as well the CI was throwing the same error, thus on a trial basis I added this to test if this might solve the issue. Thus will be removing this out anyway.

mdietze (approved):

meetagrawal09:
@mdietze, @infotroph I am getting the following error when trying to have python setup in Github Actions. `ERA5` needs the `cdsapi` python package and I am using `pip` to install it. Can you please help me in fixing this one.


```
python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by python)
python: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
```

meetagrawal09:
> @mdietze, @infotroph I am getting the following error when trying to have python setup in Github Actions. `ERA5` needs the `cdsapi` python package and I am using `pip` to install it. Can you please help me in fixing this one.
> 
> ```
> python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by python)
> python: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
> python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
> python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
> python: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /__t/Python/3.9.18/x64/lib/libpython3.9.so.1.0)
> ```

Resolved it by changing Ubuntu version to `20.04`

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: |
+        cat .cdsapirc
+     
+    - name: Set up Python
+      uses: actions/setup-python@v2
+
+    - name: Install dependencies
+      run: |
+        sudo -H apt install python3-pip
+        sudo -H pip install cdsapi
```
RohanSasne, line 41:
```suggestion
        sudo apt-get update
        sudo apt-get install python3-pip -y
        sudo -H python3 -m pip install --upgrade pip
        sudo -H pip install cdsapi
```

This should do the work @meetagrawal09

RohanSasne (changes requested):

.github/workflows/integration-test.yml
RohanSasne: This change will make sure that the job to install cdsapi will be completed successfully :)

meetagrawal09:
@mdietze I have updated the PR to run integration tests for all other data items except `ERA5` since that requires additional setup to access the CDS API. You can proceed with merging this one so that tests for other data items can run. I will raise another PR for changes needed by `ERA5`. 

Successful test run : [Job Link](https://github.com/PecanProject/pecan/actions/runs/7328822487/job/19957623264?pr=3213)

PR #3218, opened 2023-09-15
notes/warnings fixes in data.atmosphere package

moki1202:


modules/data.remote/tests/Rcheck_reference.log
```diff
@@ -80,49 +33,24 @@ Author field differs from that derived from Authors@R
...
-download.thredds.AGB: no visible global function definition for
-  ‘stopCluster’
-extract_NLCD: no visible global function definition for ‘dbfile.check’
-extract_NLCD: no visible global function definition for ‘db.query’
-extract_NLCD: no visible global function definition for ‘raster’
-extract_NLCD: no visible global function definition for ‘SpatialPoints’
-extract_NLCD: no visible global function definition for ‘CRS’
-extract_NLCD: no visible global function definition for ‘spTransform’
-extract_NLCD: no visible global function definition for ‘crs’
-extract_NLCD: no visible global function definition for ‘extract’
```
infotroph, line 116: Look at all those lines removed! 🥇

moki1202: @infotroph I think I messed up a little. data.remote log file wasn't supposed to be here. I might have messed up while rebasing.

modules/data.atmosphere/R/met_functions.R
```diff
@@ -24,6 +24,10 @@ get.latlonbox <- function(lati, loni, Lat = Lat, Lon = Lon) {
 
 get.cruncep <- function(lat, lon, start.date, end.date) {
   result <- load.cfmet(lat, lon)
+  Lat <- ncdf4::ncvar_get(nc = met.nc, varid = "lat")
+  Lon <- ncdf4::ncvar_get(nc = met.nc, varid = "lon")
+  lati <- which.min(abs(Lat - lat))
```
infotroph, line 29: I don't think think is right -- note that met.nc isn't an argument to this function, so I don't think it's available to read from here. 

But I also think something was broken here already -- `load.cfmet` has five mandatory arguments but is being passed only two of them on line 26.

infotroph: `get.weather`, `is.land`, `get.latlonbox`, and `get.cruncep` all are not exported and the latter three aren't called anywhere, while `get.weather` is called only in the very out-of-date `vignettes/compare_narr_cruncep_met.R`. I vote we delete all four of them.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -330,7 +330,7 @@ met_temporal_downscale.Gaussian_ensemble <- function(in.path, in.prefix, outfold
         if (precip[i] == 0) {
           p <- 2
         }
-        hdry[i] <- A * p * (1 - exp(-1 * bmlist[i] * ((temp_max[i] - temp_min[i])^C)))
+        hdry[i] <- A * p * (1 - exp(-1 * bmlist[i] * ((df$temp_max[i] - df$temp_min[i])^C)))
```
infotroph, line 333: How sure are you that `temp_max` and `temp_min` come from `df`? Since `df` starts out empty on line 112  I think we'd be able to see where they were defined if they're there... but this code is definitely confusing and I don't see where else they would have come from

modules/data.atmosphere/R/temporal.downscaling.R
```diff
@@ -60,7 +60,7 @@ cfmet.downscale.subdaily <- function(subdailymet, output.dt = 1) {
   new.date <- subdailymet[,list(hour = 0:(23 / output.dt) / output.dt),
                     by = c("year", "month", "day", "doy")]
 
-  new.date$date <- new.date[,list(date = lubridate::ymd_h(paste(year, month, day, hour)))]
+  new.date$date <- new.date[,list(date = lubridate::ymd_h(paste(new.date$year, new.date$month, new.date$day, new.date$hour)))]
```
infotroph, line 63: I think this line does some data.frame magic that may or may not work with $. Leaving this note as a reminder for me to investigate more.

modules/data.atmosphere/R/tdm_lm_ensemble_sims.R
```diff
@@ -227,8 +227,13 @@ lm_ensemble_sims <- function(dat.mod, n.ens, path.model, direction.filter, lags.
       }
 
       # Load the saved model
-      load(file.path(path.model, v, paste0("model_", v, "_", day.now,
-                                           ".Rdata")))
+      model.file <- file.path(path.model, v, paste0("model_", v, "_", day.now,
+                                                    ".Rdata"))
+      if(file.exists(model.file)) {
+        env = new.env()
+        load(model.file, env = env)
```
infotroph, line 234:
```suggestion
        load(model.file, envir = env)
```

modules/data.atmosphere/R/met_functions.R
infotroph: Deleted in 0b0d2434e

modules/data.atmosphere/R/met.process.R
```diff
@@ -227,24 +227,24 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
       
-      machine.id <- get.id(table = "machines", "hostname", PEcAn.remote::fqdn(), con)
+      machine.id <- PEcAn.DB::get.id(table = "machines", "hostname", PEcAn.remote::fqdn(), con)
       # Fidning the tiles.
       raw.tiles <- tbl(con, "inputs") %>%
         filter(
-          site_id == register$ParentSite,
+          .data$site_id == register$ParentSite,
           start_date >= start_date,
           end_date <= end_date,
```
infotroph, line 242: These lines are a bug! Both sides of the comparison are evaluated in the data mask, so they always compare equal.
```suggestion
          .data$start_date <= .env$start_date,
          .data$end_date >= .env$end_date,
```
Note the sign flip -- I think this step is looking for available inputs that cover the requested range, so I assume the original author was thinking of it as `.env$start_date >= .data$start_date`, so I reversed it to keep .data consistently on the left.

modules/data.atmosphere/R/temporal.downscaling.R
```diff
@@ -100,85 +104,116 @@ cfmet.downscale.subdaily <- function(subdailymet, output.dt = 1) {
...
+        each = n)
+      * light$resC2,
+    # Note: When dt >= 6, all times get the same T prediction
+    Temp = df$tmin +
+      (sin(2 * pi * (tseq - 10) / n) + 1) /
+      2 * (df$tmax - df$tmin),
+    relative_humidity = df$rhmin + rhscale * (df$rhmax - df$rhmin),
+    # TODO: Why do we divide by n?
+    #   isn't precipitation_flux already an intensity?
+    precipitation_flux = rep(df$precipitation_flux / n, each = n),
```
infotroph, line 150: The CF units on `precipitation_flux` are documented as kg m^-2 s^-1, so I would have expected it to be an average rate across the day and therefore to not need dividing by `n`. Is the division incorrect or am I missing something else? Is there a chance this input quantity was supposed to be named `precipitation_amount` (with units kg m^-2 = mm)?

modules/data.atmosphere/R/temporal.downscaling.R
```diff
@@ -100,85 +104,116 @@ cfmet.downscale.subdaily <- function(subdailymet, output.dt = 1) {
...
+    wind = rep(df$wind_speed, each = n)) %>%
+  dplyr::mutate(
+    # TODO: Computation of solarR above already multiplies by resC2.
+    #   Is multiplying it again here really correct?
+    #   That's how the old data.table version did it
+    #   (once when computing `solarR` and again when computing `SolarR`),
+    #   so keeping it until proven wrong.
+    downwelling_photosynthetic_photon_flux = .data$solarR * light$resC2)
```
infotroph, line 158: I don't know radiation calculations well enough to know if this is correct or not, just seemed odd to me especially since they're done in two separate steps. Can a met expert weigh in here, please?

infotroph: (see line 133 for the first multiplication by `resC2`)

modules/data.atmosphere/R/temporal.downscaling.R
infotroph: resolved in [32b7afb](https://github.com/PecanProject/pecan/pull/3218/commits/32b7afbb942812a3327acc7658d1089ec1ad56d6)

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
infotroph: I'm going to revert this change and go back to ignoring the NOTE about `temp_max` and `temp_min`.

I went back through the history of this file and as far as I can tell `temp_max` and `temp_min` have *never* had any visible binding. Whether the code is working as intended or not, it's been this way for a long time and can wait for someone to investigate in more detail.

infotroph: Done in 691aa6b4f

modules/data.atmosphere/R/temporal.downscaling.R
infotroph: (And see deleted lines 136 and 138 of the for the same calculations as implemented before this PR)

modules/data.remote/tests/Rcheck_reference.log
```diff
@@ -33,22 +80,47 @@
...
+download.thredds.AGB: no visible global function definition for
+  ‘stopCluster’
+extract_NLCD: no visible global function definition for ‘dbfile.check’
+extract_NLCD: no visible global function definition for ‘db.query’
+extract_NLCD: no visible global function definition for ‘raster’
+extract_NLCD: no visible global function definition for ‘SpatialPoints’
+extract_NLCD: no visible global function definition for ‘CRS’
+extract_NLCD: no visible global function definition for ‘spTransform’
+extract_NLCD: no visible global function definition for ‘crs’
+extract_NLCD: no visible global function definition for ‘extract’
```
mdietze, line 114: Why are a bunch of new warnings being added to the log? I thought the ideas was that we would only ever reduce them.

infotroph: These were accidentally added way up at the top of the PR -- notice this is the log for data.remote in a data.atmosphere PR. Sure thought we'd removed them before, but will do it now.

infotroph: Fixed in [f6b0857](https://github.com/PecanProject/pecan/pull/3218/commits/f6b08570fe7654f0e6c2c61ed09c8ae5620352cb)

infotroph (approved):

PR #3219, opened 2023-09-15
fix notes/warnings in data.land package

moki1202:


mdietze (changes requested):
Rcheck_reference.log

modules/data.land/R/plot2AGB.R
```diff
@@ -84,34 +84,34 @@ plot2AGB <- function(combined, out, outfolder, allom.stats, unit.conv = 0.02) {
...
-    mbiomass_thoc2[i, ] <- apply(biomass_thoc2[i, , ], 2, mean, na.rm = TRUE)
-    sbiomass_thoc2[i, ] <- apply(biomass_thoc2[i, , ], 2, sd, na.rm = TRUE)
+    mbiomass_tsca[i, ]  <- apply(biomass_tsca[i, , ], 2, stats::mean, na.rm = TRUE)
+    sbiomass_tsca[i, ]  <- apply(biomass_tsca[i, , ], 2, stats::sd, na.rm = TRUE)
+    mbiomass_acsa3[i, ] <- apply(biomass_acsa3[i, , ], 2, stats::mean, na.rm = TRUE)
+    sbiomass_acsa3[i, ] <- apply(biomass_acsa3[i, , ], 2, stats::sd, na.rm = TRUE)
+    mbiomass_beal2[i, ] <- apply(biomass_beal2[i, , ], 2, stats::mean, na.rm = TRUE)
+    sbiomass_beal2[i, ] <- apply(biomass_beal2[i, , ], 2, stats::sd, na.rm = TRUE)
+    mbiomass_thoc2[i, ] <- apply(biomass_thoc2[i, , ], 2, stats::mean, na.rm = TRUE)
+    sbiomass_thoc2[i, ] <- apply(biomass_thoc2[i, , ], 2, stats::sd, na.rm = TRUE)
```
Aariq, line 99: `mean()` isn't in `stats`.  These changes should be reverted

mdietze: Good catch @Aariq 

@moki1202 rather than reverting these should be updated to where the functions are, which is `base`

Aariq: You can do that, but I think specifying `base::` isn't usually required by `check()`?

moki1202: Resolved.

mdietze:
Rcheck_reference.log?

infotroph (changes requested):
This looks very good! I left a couple of optional formatting suggestions to help tame a few of the very long lines. 

The one remaining  change I see needed before merging: Please make sure to update Rcheck_reference.log (`RESAVE_CHECKS=true Rscript scripts/check_with_errors.R modules/data.land`). We should see a LOT of messages disappear from it!

modules/data.land/R/soil_utils.R
```diff
@@ -20,16 +20,38 @@
 #' @return list of soil hydraulic and thermal parameters
 #' @export
 #' @importFrom rlang %||%
-    # null-filling operator: A %||% B == "B if A is null, A in all other cases"
+# null-filling operator: A %||% B == "B if A is null, A in all other cases"
```
infotroph, line 23: Nit: I preferred the space here, to visually distinguish this comment from the Roxygen block around it.
Though I do admit that interleaving `#'` and `#` like this is always ripe for confusion and maybe is a sign this needs a bigger rewrite :)
```suggestion
    # null-filling operator: A %||% B == "B if A is null, A in all other cases"
```

modules/data.land/tests/testthat/test-match_species_id.R
```diff
@@ -0,0 +1,41 @@
+test_that("Species matching works", {
```
infotroph, line 1: Not actually a new file, just renamed from `test.match_species_id.R`. Apparently the rearragements I did in other commits before the move are enough to make gitHub treat this and a delete-and-add rather than a rename.

modules/data.land/R/soil_utils.R
Aariq: The `%||%` operator is in base r-devel and will be in the next big release of R! (still good to import from rlang to not require a min R version though)

infotroph: @Aariq This is good news! Would you agree that means it's well-known enough we don't need to explain the import and could delete this comment entirely?

Aariq: Haha, no, I think it's actually really confusing and not obvious what it does at all 😅.  I find that comment very helpful! However, it might make more sense to move it to wherever in the function(s) `%||%` is first used rather than mixing roxygen comments (which _do_ something) and comments about those comments.

modules/data.land/R/soil_utils.R
```diff
@@ -44,27 +66,27 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
     mysoil$soil_n <- which(toupper(soil.name) == toupper(soil_type))
     sand <- xsand.def[soil_type]
     clay <- xclay.def[soil_type]
-
+
```
infotroph, line 69:
```suggestion

```

modules/data.land/R/soil_utils.R
```diff
@@ -44,27 +66,27 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
     mysoil$soil_n <- which(toupper(soil.name) == toupper(soil_type))
     sand <- xsand.def[soil_type]
     clay <- xclay.def[soil_type]
-
+    
   } else {
-
+
```
infotroph, line 71:
```suggestion

```

modules/data.land/R/soil_utils.R
```diff
@@ -44,27 +66,27 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
...
-
+    
     if (any(c(sand, silt, clay) > 2)) {
       # assume values reported in % not proportion
       sand <- if (is.null(sand)) { NULL } else { sand / 100 }
       silt <- if (is.null(silt)) { NULL } else { silt / 100 }
       clay <- if (is.null(clay)) { NULL } else { clay / 100 }
     }
-
+
```
infotroph, line 78:
```suggestion

```

modules/data.land/R/soil_utils.R
```diff
@@ -44,27 +66,27 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
...
       clay <- if (is.null(clay)) { NULL } else { clay / 100 }
     }
-
+    
     # compute up to one missing value (>1 missing was handled above)
     sand <- sand %||% (1-silt-clay)
     silt <- silt %||% (1-sand-clay)
     clay <- clay %||% (1-sand-silt)
-
+
```
infotroph, line 83:
```suggestion

```

modules/data.land/R/soil_utils.R
```diff
@@ -44,27 +66,27 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
...
     clay <- clay %||% (1-sand-silt)
-
+    
     #normalize
     stot <- sand + silt + clay
     sand <- sand / stot
     silt <- silt / stot
     clay <- clay / stot
-
+
```
infotroph, line 89:
```suggestion

```

modules/data.land/R/soil_utils.R
```diff
@@ -44,27 +66,27 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
...
+    
     if (any(c(sand, silt, clay) > 2)) {
       # assume values reported in % not proportion
       sand <- if (is.null(sand)) { NULL } else { sand / 100 }
       silt <- if (is.null(silt)) { NULL } else { silt / 100 }
       clay <- if (is.null(clay)) { NULL } else { clay / 100 }
     }
-
+    
     # compute up to one missing value (>1 missing was handled above)
```
infotroph, line 79:
```suggestion
    # compute up to one missing value (>1 missing was handled above)
    # %||% is a null-filling operator: A %||% B == "B if A is null, A in all other cases"
```

modules/data.land/R/soil_utils.R
infotroph: Good call -- will do that below.

modules/data.land/R/soil_utils.R
```diff
@@ -118,15 +140,19 @@ soil_params <- function(soil_type=NULL, sand=NULL, silt=NULL, clay=NULL, bulk=NU
...
+      "inch/hour",
+      "meters/second") 
     
     # Soil moisture at saturation [ m^3/m^3 ]
     mysoil$volume_fraction_of_water_in_soil_at_saturation[z]  <- (50.5 - 14.2*mysoil$fraction_of_sand_in_soil[z] - 3.7*mysoil$fraction_of_clay_in_soil[z]) / 100.
     
     # Soil field capacity[ m^3/m^3 ]
     mysoil$volume_fraction_of_water_in_soil_at_field_capacity[z] <- mysoil$volume_fraction_of_water_in_soil_at_saturation[z] * ( fieldcp.K/mysoil$soil_hydraulic_conductivity_at_saturation[z])^ (1. / (2.*mysoil$soil_hydraulic_b[z]+3.))
-  
+
```
infotroph, line 155:
```suggestion

```

modules/data.land/R/soil_utils.R
```diff
@@ -20,16 +20,38 @@
 #' @return list of soil hydraulic and thermal parameters
 #' @export
 #' @importFrom rlang %||%
-    # null-filling operator: A %||% B == "B if A is null, A in all other cases"
+    # `A %||% B` is a null-filling operator: "B if A is null, A in all other cases"
```
infotroph, line 23:
```suggestion
```

infotroph (approved):

infotroph:
@mdietze Have all your requested changes been addressed?

mdietze (approved):

PR #3221, opened 2023-09-18
Description title for data.mining package switched to upper case

moki1202:


mdietze (approved):

modules/data.mining/DESCRIPTION
```diff
@@ -21,4 +21,4 @@ LazyLoad: yes
 LazyData: FALSE
 Collate:
 Encoding: UTF-8
-RoxygenNote: 6.1.1
+RoxygenNote: 7.1.2
```
infotroph, line 24: Might as well use the same Roxygen version as the rest of the packages... Since there's no code to document, this is the only change created by the version bump.

```suggestion
RoxygenNote: 7.2.3
```

moki1202: Will change this in all the PRs.

PR #3222, opened 2023-09-18
fix warnings/notes in data.remote package

moki1202:


modules/data.remote/R/NLCD.R
```diff
@@ -104,14 +104,14 @@ extract_NLCD <- function(buffer, coords, data_dir = NULL, con = NULL, year = 201
         print(paste("File not found:", filename))
         return(NULL)
     }
-    nlcd <- raster(filename)
+    nlcd <- raster::raster(filename)
```
mdietze, line 107: Not part of this PR, but we should open a separate issue about switching all of our `raster` calls to `terra` since `raster` is deprecated

mdietze (changes requested):
Rcheck_reference.log

modules/data.remote/R/NLCD.R
moki1202: I had the exact suspicion! raster is being used at a lot of places, so I'll leave it for now.

moki1202:
> Rcheck_reference.log

@mdietze Can I run R CMD check without --as-cran tag and paste the result in the log file? or should I be doing it manually?

mdietze:
> > Rcheck_reference.log
> 
> @mdietze Can I run R CMD check without --as-cran tag and paste the result in the log file? or should I be doing it manually?

How about you try running R CMD check for one of the packages and then do a quick diff between what you get and what's there now in the log file. If the changes are just the lines you expect then the cut-and-paste approach should work. If the diff is huge then there's something else different and pasting the result into the log will just cause all subsequent builds to fail.

modules/data.remote/R/NLCD.R
Aariq: Not to be pedantic, but `raster` has been *superseded*, not deprecated.  The difference is that `raster` isn't going away anytime soon. Still probably a good idea to do a spatial package makeover at some point!

Aariq:
More info on failing check: https://yihui.org/en/2023/02/xfun-isfalse/

moki1202:
> More info on failing check: https://yihui.org/en/2023/02/xfun-isfalse/

@Aariq I tried looking for a general import or a library declaration of xfun in pecan but couldn't find one, so assuming the isFALSE functions we use are from base, why do we get this error?

Aariq:
The error is because the GitHub actions are somehow using old versions of R packages.  Someone just needs to update them.  The vignettes build without error locally.  It *could* be something to do with the DESCRIPTION asking for a fixed version of rmarkdown from GitHub. 
https://github.com/PecanProject/pecan/blob/b3b3191b6d171cf141dac893ed1f591b7ff6c0f8/base/db/DESCRIPTION#L84-L86
I'm guessing someone set these for development reasons and forgot to remove them.  They'll have to be removed before going to CRAN anyways, so worth a try getting rid of them?

Aariq:
I copied just base/db to its own repo and set up standard checks with `usethis::use_github_action()` and looks like the vignettes are building fine. https://github.com/Aariq/PEcAn.DB/actions/runs/6641355603

These standard checks only fail because of an invalid email address which apparently isn't getting caught here.

infotroph:
> The error is because the GitHub actions are somehow using old versions of R packages. Someone just needs to update them. The vignettes build without error locally. It _could_ be something to do with the DESCRIPTION asking for a fixed version of rmarkdown from GitHub.
> 
> https://github.com/PecanProject/pecan/blob/b3b3191b6d171cf141dac893ed1f591b7ff6c0f8/base/db/DESCRIPTION#L84-L86
> 
> 
> I'm guessing someone set these for development reasons and forgot to remove them. They'll have to be removed before going to CRAN anyways, so worth a try getting rid of them?

@Aariq Yes, as is detailed in the comment directly above that.

https://github.com/PecanProject/pecan/blob/b3b3191b6d171cf141dac893ed1f591b7ff6c0f8/base/db/DESCRIPTION#L78-L83

infotroph (approved):

mdietze (approved):

PR #3225, opened 2023-09-26
Updated Integration Tests

meetagrawal09:
## Description

This PR tries to change the way we access functions from a package in integration tests.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
+++ b/modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
@@ -1,19 +1,23 @@
 library(testthat)
+library(mockery)
+library(PEcAn.DB)
+library(PEcAn.logger)
+library(withr)
 
 test_download_AmerifluxLBL <- function(start_date, end_date, sitename, lat.in, lon.in) {
   # putting logger to debug mode
-  PEcAn.logger::logger.setUseConsole(TRUE, FALSE)
-  on.exit(PEcAn.logger::logger.setUseConsole(TRUE, TRUE), add = TRUE)
-  PEcAn.logger::logger.setLevel("DEBUG")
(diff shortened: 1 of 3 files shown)


modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
```diff
@@ -1,19 +1,23 @@
 library(testthat)
+library(mockery)
+library(PEcAn.DB)
+library(PEcAn.logger)
+library(withr)
 
 test_download_AmerifluxLBL <- function(start_date, end_date, sitename, lat.in, lon.in) {
   # putting logger to debug mode
-  PEcAn.logger::logger.setUseConsole(TRUE, FALSE)
```
mdietze, line 5: In general I'm surprised by the need to add the library calls. I'm even more surprised that you need to actively remove the namespace info. Did you try a version that added the library calls without deleting the namespace?

meetagrawal09: Yes it does seem kind of strange. But I did try to add library calls and keep the namespaces but that too resulted in a failure while running the tests.

meetagrawal09:
These tests work with all versions of `mockery` and can now be run using the command: 
```
 Rscript modules/data.atmosphere/inst/integrationTests/test.download.ERA5.R
 ```

modules/data.atmosphere/inst/integrationTests/test.download.AmerifluxLBL.R
meetagrawal09: I tried out removing just the `PEcAn.DB` namespace and putting that to a library call. This works well.

mdietze (approved):

PR #3227, opened 2023-10-06
Add resave mode to check_with_errors.R

infotroph:
## Description
Adds a new usage variant: To update the cached check results in `[package_dir]/tests/Rcheck_reference.log`, run the check script with env var `RESAVE_CHECKS` set, e.g.:

```
RESAVE_CHECKS=true Rscript scripts/check_with_errors.R base/db
```

After running, review the changes to `Rcheck_reference.log` and commit them if they look good.

## Motivation and Context

Thanks to the whole team in general and @moki1202 in particular, lots of our cached check results are stale in the sense that they contain messages about errors/warnings we have since fixed. We've mostly been updating these by manually editing the saved files, but this is tedious and error-prone. This patch provides a one-shot method to ease the process.

As always, my hope is that we'll use this to remove fixed messages (so that we can eventually remove all `Rcheck_reference.log` files entirely) rather than to ignore newly added messages, but it will do either one as needed.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/scripts/check_with_errors.R
+++ b/scripts/check_with_errors.R
@@ -7,6 +7,8 @@ log_level <- Sys.getenv("LOGLEVEL", unset = NA)
 die_level <- Sys.getenv("DIELEVEL", unset = NA)
 redocument <- as.logical(Sys.getenv("REBUILD_DOCS", unset = NA))
 runtests <- as.logical(Sys.getenv("RUN_TESTS", unset = TRUE))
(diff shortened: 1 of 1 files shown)


mdietze (approved):

moki1202:
@infotroph This is awesome, thankyou! I'll push the changes for my PRs.

Issue #2914, opened 2022-02-25
Migrate geospatial tools from the rgdal ecosystem (rgdal, rgeos, maptools) depends to sf & terra


serbinsh:
## Description

Apparently, the R spatial world is shifting away from rgdal, rgeos, and maptools toward sf and terra.  rgdal, rgeos, and maptools all scheduled for retirement at the end of 2023.  Thanks for the heads up @infotroph 

https://mobile.twitter.com/RogerBivand/status/1407705212538822656
https://cran.r-project.org/web/packages/rgdal/index.html

raster doesnt yet have a retirement date, but its suggested to move toward terra

## Proposed Solution

So to me this means right now is a great time to overhaul our spatial dependencies.  First start with simpler functions toward sf/terra (e.g. in data.remote) but we need to develop a list of all functions that use these tools and then schedule the order for updating.

## Alternatives Considered

none

## Additional Context 
```
[](https://twitter.com/@RogerBivand)[Roger Bivand](https://twitter.com/@RogerBivand)Roger Bivand [@RogerBivand](https://twitter.com/@RogerBivand)
[@edzerpebesma](https://twitter.com/edzerpebesma) I'm not sure that taking over maintenance is a sensible use of effort (add maptools to that list). The main condition for this will be transition from raster to terra, as raster uses rgdal heavily. If sf/stars and terra become the main workflow bases, things get simplified.
TwitterTwitter | [Jun 23rd, 2021](https://mobile.twitter.com/RogerBivand/status/1407705212538822656)
```

Some other issues - if we need to buffer points to average pixels when extracting does terra support that like raster?

Aariq:
related: https://github.com/PecanProject/pecan/issues/2913

Aariq:
more details: https://github.com/r-spatial/evolution

Aariq:
Appearances of `rgdal`:

https://github.com/PecanProject/pecan/blob/1043bfd57becbb79885e9455439308c68b9816ca/modules/data.land/R/gis.functions.R#L71-L72

https://github.com/PecanProject/pecan/blob/1043bfd57becbb79885e9455439308c68b9816ca/modules/data.remote/R/NLCD.R#L72

https://github.com/PecanProject/pecan/blob/1043bfd57becbb79885e9455439308c68b9816ca/modules/data.atmosphere/R/GEFS_helper_functions.R#L248

Plus a LOT of scripts and vignettes

Aariq:
[...]

PR #3211, opened 2023-08-10
Add the down_scale function to pecan.

JoshuaPloshay:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3229, opened 2023-11-01
removed rgdal from data.land and data.remote

istfer:
## Description

Per issue https://github.com/PecanProject/pecan/issues/2914 made changes to data.land. It's a very small change, rgdal functions were being used for printing info only.

I just removed rgdal from data.remote `DESCRIPTION` as well, although there was an `require(rgdal)` in the `extract_NLCD` function, no rgdal package function was being used within this function, also seems to be the case in #3222.

## Motivation and Context

See https://github.com/PecanProject/pecan/issues/2914.

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.remote/R/NLCD.R
+++ b/modules/data.remote/R/NLCD.R
@@ -69,8 +69,7 @@ download.NLCD <- function(outdir, year = 2011, con = NULL) {
 ##' @description Based on codes from Christy Rollinson and from Max Joseph (http://mbjoseph.github.io/2014/11/08/nlcd.html)
 extract_NLCD <- function(buffer, coords, data_dir = NULL, con = NULL, year = 2011) {
     library(raster)
-    require(rgdal)
-    
+
     if (!is.null(con)) {
         library(PEcAn.DB)
         if (year == 2001) {
--- a/modules/data.remote/R/NLCD.R
+++ b/modules/data.remote/R/NLCD.R
@@ -103,21 +103,21 @@ extract_NLCD <- function(buffer, coords, data_dir = NULL, con = NULL, year = 201
         print(paste("File not found:", filename))
         return(NULL)
     }
-    nlcd <- raster(filename)
+    nlcd <- terra::rast(filename)
     
     # transform points
-    sites <- SpatialPoints(coords = coords, proj4string = CRS("+proj=longlat +datum=WGS84"))
-    sites <- spTransform(sites, crs(nlcd))
+    sites <- terra::vect(coords, geom=c("long", "lat"), crs="+proj=longlat +datum=WGS84")
+    sites <- terra::buffer(x, width=buffer)
     
     # extract
-    sum.raw <- table(extract(nlcd, sites, buffer = buffer))
+    sum.raw <- table(terra::extract(nlcd, sites))
     summ <- prop.table(sum.raw)
-    mydf <- data.frame(cover = names(summ), percent = as.vector(summ), count = as.vector(sum.raw))
+    mydf <- data.frame(cover.name = colnames(summ), percent = as.vector(summ), count = as.vector(sum.raw))
+    mydf <- mydf[mydf$count!=0,]
     
-    # land cover number to name conversions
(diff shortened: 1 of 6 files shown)


modules/data.remote/R/NLCD.R
```diff
@@ -69,8 +69,7 @@ download.NLCD <- function(outdir, year = 2011, con = NULL) {
 ##' @description Based on codes from Christy Rollinson and from Max Joseph (http://mbjoseph.github.io/2014/11/08/nlcd.html)
 extract_NLCD <- function(buffer, coords, data_dir = NULL, con = NULL, year = 2011) {
     library(raster)
-    require(rgdal)
```
mdietze, line 72: Do we know if raster's dependencies on rgdal have been fixed? If not, this fix could be pretty short lived. I know in other cases we're trying to migrate from raster to terra, but that's a harder fix than the current PR

mdietze (approved):

modules/data.remote/R/NLCD.R
istfer: Ah right, I don't know much about `raster`, but checking `terra` some lines can be changed, e.g.:

```
nlcd <- raster::raster(filename)
```
to
```
nlcd <- terra::rast(filename)
```
and
```
sites <- sp::spTransform(sites, raster::crs(nlcd))
```
to
```
sites <- sp::spTransform(sites, terra::crs(nlcd, proj=TRUE))
```
then probably instead of `raster::extract`, `terra::extract` could be used, but I'm not familiar with this function. Is there someone who can quickly check if such changes would work as intended?

mdietze: @istfer Given that this function isn't actively being used in any automated workflow at the moment (i.e. there's no one else who has things set up for a quick and easy test), I'd be OK with you making the obvious changes and leaving a note or warning that they haven't been rigorously tested.  

FYI the closest analog is @JoshuaPloshay work on PR https://github.com/PecanProject/pecan/pull/3211 which uses terra to extract a number of other data products.

istfer: got it, I made an attempt to replace with terra functions (I found an nlcd image to test it with) and put a warning

istfer:
I'm guessing CI's API rate limit errors are not something I can fix on my end?

mdietze:
Nope, you just have to wait for the API to refresh and then try running the ones that failed again

mdietze:
@istfer the build hit the following errors in check_modules
```
check modules/data.remote
  2 warnings found in modules/data.remote.
  1 notes found in modules/data.remote.
  R check of modules/data.remote returned new problems:
  Execution halted
  checking dependencies in R code ... WARNING: '::' or ':::' import not declared from: 'terra'
  checking dependencies in R code ... WARNING: 'library' or 'require' calls not declared from: 'doParallel' 'PEcAn.DB'
  checking dependencies in R code ... WARNING: 'library' or 'require' calls in package code: 'doParallel' 'PEcAn.DB' Please use :: or requireNamespace() instead. See section 'Suggested packages' in the 'Writing R Extensions' manual.
  ```
  
  At a minimum, I think you need to remove `raster` from the DESCRIPTION and add `terra`. Looks like there are additional issues with 'doParallel' and 'PEcAn.DB' that aren't your fault but may need to be resolved too. For example there's a `require(doParallel)` in `download.thredds.R` that we can probably get rid of since the functions appear to be using the namespace. Same with the `library(PEcAn.DB)` in NLCD.R

modules/data.remote/R/NLCD.R
Aariq: `raster` no longer imports `rgdal`.  `raster` has replaced `rgdal` imports with `terra` so you do not need to change this (but probably good to do it eventually).  If a packages is on CRAN still, it no longer uses the retired r-spatial packages. You can also  check if you've gotten rid of all the `rgdal` dependencies with `pak::local_deps_explain("rgdal", "modules/data.land")`

PR #3232, opened 2023-11-03
Update write.configs.ed.R

mdietze:
third straw on isFalse

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/ed/R/write.configs.ed.R
+++ b/models/ed/R/write.configs.ed.R
@@ -393,7 +393,7 @@ write.config.ED2 <- function(trait.values, settings, run.id, defaults = settings
   }
 
   barebones <- settings$model$barebones_ed2in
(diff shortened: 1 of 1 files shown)


Aariq:
I don't think these errors with `xfun::isFALSE()` are anything in PEcAn—the error is more likely coming from an old version of `rmarkdown` or `knittr` or something (not exactly sure, because Yihui doesn't really say where exactly the error comes from in his [blog post](https://yihui.org/en/2023/02/xfun-isfalse/)). When I've seen this error, I've been able to fix it just by using recent versions of packages. For example, the vignettes for PEcAn.DB are built without error here: https://github.com/Aariq/PEcAn.DB/actions/runs/6641355603/job/18043682007#step:6:42

infotroph:
Closing since dropping support for R 4.0 seems to have ~fixed this.

PR #3234, opened 2023-11-21
Update Actions to R4.2

mdietze:
Swaps the outdated R4.0 build to R4.2 to see if that version builds successfully -- was overdue and might get the GH Actions working again.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3235, opened 2023-11-22
Build more depends versions

infotroph:
## Description
* Adds builds of the depends image with R 4.2 and 4.3
* Adds the `patch` utility, which is needed to compile hdf5r on images (e.g. Rdevel) whose CRAN doesn't provide a precompiled binary version.

## Motivation and Context

Building the depends image for new versions of R is a necessary upstream step before using them in CI as proposed in #3234, and will facilitate local testing as well. Once this PR is merged and the `depends` workflow has run, the R 4.2 and 4.3 images will be available for other PRs to use.

I suspect we don't want to keep building 4 versions forever, but we can wait until the new versions are working before turning off any old ones.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (approved):

mdietze:
@infotroph despite pulling this in, there's still not R4.2 and R4.3 versions of pecan/depends at https://hub.docker.com/r/pecan/depends/tags

infotroph:
@mdietze The build job that pushes to Docker Hub runs at midnight UTC daily -- looks like [yesterday's run](https://github.com/PecanProject/pecan/actions/runs/6964229201) was just before this merged, but hopefully today's run should pick it up.

PR #3237, opened 2023-11-27
Misc fixes found when running package checks on newer releases of R

infotroph:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/utils/R/mcmc.list2init.R
```diff
@@ -30,7 +30,7 @@ mcmc.list2init <- function(dat) {
   
   ## define variables
   ic <- list()
-  n <- nrow(dat[[1]])
+  nr <- nrow(dat[[1]])
```
robkooper, line 33: surprised this was not found earlier.

infotroph:
Probably need to merge #3244 first to resolve current CI failures

mdietze:
@infotroph there are merge conflicts that need to be resolved in order to pull this one in

infotroph:
Local debugging of the current check failures on 4.1 shows they're from dplyr refusing to load because the installed version of tidyselect was too old.

This was caused by the `remotes::install_version(pkg, dependencies = TRUE, upgrade = FALSE)` added to pecan.depends.R in #3244: Apparently "dependencies = TRUE" only installs dependencies that are _not already installed_, and does not upgrade ones with installed but unsuitable versions.

Options I see: switch to `upgrade = TRUE` and accept that will pull in large numbers of updates to old systems, or parse dependencies ourselves inside `ensure_version` to upgrade only as far as the package requires.

mdietze (approved):

modules/rtm/R/gpm.R
```diff
@@ -12,6 +12,6 @@ generalized_plate_model <- function(k, refractive, N) {
     length(N) == 1
   )
   RT <- matrix(0, 2101, 2)
-  outlist <- .Fortran("gpm", k, refractive, N, RT)
+  outlist <- .Fortran("gpm", k, refractive, N, RT, PACKAGE="PEcAnRTM")
```
infotroph, line 15: @ashiklom are you still lurking in the PEcAn world? Would it be possible to get your eyes on this commit?
TL;DR: Does this patch look OK to you?

This is a proposed fix for intermittent test failures from RTM that have me puzzled for three reasons:
1. Tests are only failing on R 4.2, not on older or newer R versions.
2. Failure is traceable to .Fortran calls to `foursail` sometimes, but not always, returning NaNs. Repeated calls with the same input fail maybe 2/3 of the time but succeed often enough to make me nervous.
3. Adding the PACKAGE argument to each .Fortran call seems to fix it, which would seem to imply some kind of namespace collision... but I don't know what other package would contain a `foursail` symbol or why symbol lookup would behave nondeterministically from one invocation to another.

infotroph:
All checks are passing, so I'll merge this today unless anyone speaks up before then.

Substantive changes since the last review:
* Reorganized, resaved, and documented datasets in data.land, and made package data be lazy-loaded.
* CI checks now ignore timestamps printed by newer versions of R CMD check
* include PACKAGE argument in all .Fortran calls from RTM, fixing a test failure that only happens on R 4.2

PR #3252, opened 2024-02-07
remove V8 installation script

infotroph:
cc @DongchenZ since I believe you're the main user of this script -- any reason this can't live alongside other installation tools in `scripts/`?


DongchenZ:
Sorry I have no idea about this.

mdietze:
@infotroph since this script isn't used by anyone, or called by anything else, I think we can just delete it rather than move it

mdietze (approved):

PR #3253, opened 2024-02-09
Require newer knitr and rmarkdown everywhere

infotroph:
Hopefully for real this time: Fixes `xfun::isFALSE` deprecation warnings by updating knitr and rmarkdown to versions that use base::isFALSE internally. 

Note that as wirtten, the check and renderbook jobs will fail one more time on this PR:

* The current build of the depends image is in an inconsistent state where {knitr} is still old but {xfun} has already been updated to a newer version that complains when you use an old {knitr}.
* Therefor the check job is going to fail because `pecan.depends.R` tries to install BioCro (and fails because the old knitr can't compile its vignettes) before it gets to the line that will update knitr. 😞 
* If the Docker CI job passes, it should be OK to merge this with the other checks failing. That will bring the depends image back into a version-consistent state.
* Once that deployment finishes and the depends image rebuilds overnight, the next round of package checks should be able to compile BioCro (and all of PEcAn) without error again.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -25,10 +25,12 @@ jobs:
     steps:
     # checkout source code
     - uses: actions/checkout@v3
-    # install rmarkdown
-    - name: Install rmarkdown
+    # install bookdown
+    - name: Install bookdown
       run: |
-        Rscript -e 'install.packages(c("rmarkdown","bookdown"))'
+        Rscript \
+          -e 'repos <- c(getOption("repos"), sub(r"(\d{4}-\d{2}-\d{2})", "latest", getOption("repos"))' \
+          -e 'remotes::install_version("bookdown", ">= 0.31", dependencies = TRUE, upgrade = FALSE, repos = repos)'
     # copy files
     - name: copy extfiles
       run: |
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install bookdown
       run: |
         Rscript \
-          -e 'repos <- c(getOption("repos"), sub(r"(\d{4}-\d{2}-\d{2})", "latest", getOption("repos"))' \
+          -e 'repos <- c(getOption("repos"), sub(r"(\d{4}-\d{2}-\d{2})", "latest", getOption("repos")))' \
           -e 'remotes::install_version("bookdown", ">= 0.31", dependencies = TRUE, upgrade = FALSE, repos = repos)'
     # copy files
     - name: copy extfiles
--- a/docker/docs/Dockerfile
+++ b/docker/docs/Dockerfile
@@ -9,7 +9,12 @@ MAINTAINER Rob Kooper <kooper@illinois.edu>
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends pandoc \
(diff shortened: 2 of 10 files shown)


docker/docs/Dockerfile
```diff
@@ -9,7 +9,7 @@ MAINTAINER Rob Kooper <kooper@illinois.edu>
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends pandoc \
-    && install2.r -e -s -n -1 bookdown \
+    && install2.r -e -s -r cloud.r-project.org -n -1 bookdown \
```
infotroph, line 12: I'm not entirely happy with this hack -- It feels a little like a step backwards from always using the repositories set by Rocker. But since this image just serves the compiled docs without needing to interact with the rest of the system, it feels simplest to jump straight to the latest CRAN version of bookdown here.

.github/workflows/book.yml
```diff
@@ -25,10 +25,12 @@ jobs:
...
     # checkout source code
     - uses: actions/checkout@v3
-    # install rmarkdown
-    - name: Install rmarkdown
+    # install bookdown
+    - name: Install bookdown
       run: |
-        Rscript -e 'install.packages(c("rmarkdown","bookdown"))'
+        Rscript \
+          -e 'repos <- c(getOption("repos"), sub(r"(\d{4}-\d{2}-\d{2})", "latest", getOption("repos"))' \
```
infotroph, line 32: Borrowed from pecan.depends.R -- this makes install_version use the bookdown from the existing rocker-specified CRAN mirror if it's suitable, or the latest version if the Rocker-specified mirror is too old.

docker/docs/Dockerfile
```diff
@@ -9,7 +9,12 @@ MAINTAINER Rob Kooper <kooper@illinois.edu>
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends pandoc \
-    && install2.r -e -s -n -1 bookdown \
+    && Rscript \
+        -e 'repos <- c(getOption("repos"),' \
+        -e '           sub(r"(\d{4}-\d{2}-\d{2})", "latest", getOption("repos")))' \
+        -e 'remotes::install_version("rmarkdown", ">= 2.19", dependencies = TRUE, upgrade = FALSE, repos = repos)' \
+        -e 'remotes::install_version("knitr", ">= 1.42", dependencies = TRUE, upgrade = FALSE, repos = repos)' \
+        -e 'remotes::install_version("bookdown", ">= 0.31", dependencies = TRUE, upgrade = FALSE, repos = repos)' \
```
infotroph, line 17: Would it make more sense just to unconditionally install the latest CRAN versions of all three of these?

.github/workflows/book.yml
```diff
@@ -25,10 +25,12 @@ jobs:
     steps:
     # checkout source code
     - uses: actions/checkout@v3
-    # install rmarkdown
-    - name: Install rmarkdown
+    # install bookdown
+    - name: Install bookdown
```
infotroph, line 29:
```suggestion
    # install rmarkdown
    - name: Install rmarkdown
```

mdietze (approved):

PR #3260, opened 2024-02-15
[Docker, CI] Allow building pecan/depends from pecan/depends

infotroph:
## Description

For discussion -- definitely needs input from @robkooper before merging!

Adds a third option for the dependency image when building Docker images: In addition to "use the existing image" and "build from scratch", adds "try to update an existing image".

Also edits the Docker CI job to check for changed dependencies when building Docker images in a PR branch, and if detected to update the depends image before building other images on top of it.

Additionally and unrelatedly, adds `workflow_dispatch` hooks  to several more workflows for easier debugging.

## Motivation and Context

#3237 is currently failing the Docker CI check because it introduces new version requirements on `dplyr` and `dbplyr` that will be satisfied at Depends build time but that Make doesn't know how to resolve. Trying to fix this in the Makefile would add a lot of only-rarely-used complexity, and pre-emptively rebuilding the depends image from scratch on every CI run would be slow and wasteful. 

Running the depends build only when it will actually change anything (i.e. when the PR touches `docker/depends/*`) is a first step, but it still feels wasteful to wait for all hojillion deps to reinstall when we really only added one or two. Then I realized that the two time-consuming steps in the depends build (apt updates and R package installation) will both skip packages that are already present on the disk -- so if we start from an existing pecan/depends image, adding a layer with the changed dependencies should be fast.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


docker/depends/Dockerfile
```diff
@@ -1,20 +1,12 @@
...
 
-# ----------------------------------------------------------------------
-# UPDATE GIT
-# This is needed for stretch and github actions
-# ----------------------------------------------------------------------
-RUN if [ "$(lsb_release -s -c)" = "stretch" ]; then \
-      echo 'deb http://deb.debian.org/debian stretch-backports main' >> /etc/apt/sources.list \
-      && apt-get update \
-      && apt-get -t stretch-backports upgrade -y git \
-    ; fi
```
infotroph, line 17: This deletion is just cleanup -- none of our images have used Stretch since before R 4.0, so this command never runs anymore.

infotroph:
@robkooper Discussion questions: 
* Do you agree we _don't_ want to push images built with this approach back to Docker hub? The point here is to unwedge CI statuses so that we can merge with confidence; better for the pushed images to spend the time for a full clean build.
* If yes, do we need any explicit safeguards to prevent such pushing? I think that as currently implemented they won't be pushed because the update-build happens only on pull requests (technically only when github.base_ref is set, but it's documented to only be set for pull requests), but maybe that's too indirect to rely on as a safeguard.

.github/workflows/docker.yml
```diff
@@ -74,14 +72,26 @@ jobs:
...
             echo "PECAN_TAGS=develop" >> $GITHUB_ENV
           fi
-
+      # If any dependencies changed in this PR, try to update depends image
+      # Only runs on pull_request -- others it's not clear what base to compare to
+      - name: check for changed dependencies
+        if: github.base_ref != ''
+        run: |
+          dep_chgs=$(
+            git diff --name-status ${GITHUB_BASE_REF}...${GITHUB_HEAD_REF} -- docker/depends)
```
infotroph, line 81: Currently getting "fatal: bad revision develop...allow-depends-update" in the GitHub run but not locally... investigating.

docker/depends/Dockerfile
infotroph: Done, thanks!

docker/depends/Dockerfile
```diff
@@ -1,20 +1,12 @@
-ARG R_VERSION="4.0.2"
+ARG R_VERSION="4.1"
+ARG FROM_IMAGE="rocker/tidyverse"
```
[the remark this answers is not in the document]
infotroph: Being able to change the source image is the point of this PR. Whether that’s a good idea is a key question, but I don’t think it can be removed without changing approaches — this is what allows the same script to either build the image fresh from the upstream Rocker image or be updated from an existing pecan/develop.

My hope is that in the cases where we want to use this, the gain in build speed will offset the cost of the (thin) extra layers, but please do burst my bubble if that won’t be true.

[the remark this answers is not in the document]
robkooper: ARG is not written as a layer in the docker image so we don't have to worry about. This use of the ARGS allows us to customize the depends image to point to any version any base image. These are then used in all subsequent builds. If you do a `docker history pecan/depends:develop` you will not see any of the ARGS.

Removing the stretch section below saves 4KB and one layer.

robkooper (approved):

PR #3196, opened 2023-07-11
fix-notes/warnings/errors in base packages

moki1202:

PR #3261, opened 2024-02-17
Fixing Notes in `PEcAn.workflow`

meetagrawal09:
## Description
This PR fixes notes and warning raised by `R CMD check` on `PEcAn.workflow`.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


base/workflow/R/run.write.configs.R
```diff
@@ -82,6 +82,11 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
   scipen <- getOption("scipen")
   options(scipen = 12)
 
+  trait.samples <- NULL
+  sa.samples <- NULL
+  ensemble.samples <- NULL
+  runs.samples <- NULL
```
infotroph, line 88: I think this was fixed in #3196 and we neglected to update the cached check outputs -- can you please try running checks _with_ the edit to Rcheck_Reference.log but _without_ this change and see if they pass?

base/workflow/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.workflow
 Type: Package
-Title: PEcAn functions used for ecological forecasts and
-    reanalysis
+Title: PEcAn Functions Used for Ecological Forecasts and
+    Reanalysis
```
infotroph, line 4: 👍

base/workflow/R/run.write.configs.R
meetagrawal09: @infotroph, I do not see the use of the variable `env.samples` in the entire function. Can you please explain the significance of having it ?

infotroph: It's not used at the moment, but we kept it here because it seemed weird to load all the other objects from the samples file and ignore that one. I'd be fine with deleting it as unused -- doing more with env samples is a wishlist item (as evidenced by the "### NEED TO IMPLEMENT: Load Environmental Priors and Posteriors" on line 129 below), but it'd be easy enough to re-add the env.samples line when that gets implemented.

meetagrawal09: We can just keep the declaration of `env.samples` as a comment (something like below). That might be a good option. What is your take on this ?

```
 if (file.exists(samples.file)) {
    samples <- new.env()
    load(samples.file, envir = samples) ## loads ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples
    trait.samples <- samples$trait.samples
    ensemble.samples <- samples$ensemble.samples
    sa.samples <- samples$sa.samples
    runs.samples <- samples$runs.samples
    # env.samples <- samples$env.samples
  } else {
    PEcAn.logger::logger.error(samples.file, "not found, this file is required by the run.write.configs function")
  }
```

meetagrawal09: This way we don't have it declared now and can tell any future implementer that the variable can be picked here.

infotroph: I like that! Thanks.

infotroph (approved):

PR #3263, opened 2024-02-21
Fix bug for building H when there is any site that has zero observation.

DongchenZ:
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -130,6 +130,10 @@ build.block.xy <- function(settings, block.list.all, X, obs.mean, obs.cov, t) {
   }
   
   #Handle observation
+  #observation number per site
+  obs_per_site <- obs.mean[[t]] %>% 
+    purrr::map(function(site.obs){length(site.obs)}) %>% 
+    unlist()
```
infotroph, line 136: This can be simplified to
```suggestion
  obs_per_site <- purrr::map_int(obs.mean[[t]], length)
```

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -167,16 +171,20 @@ build.block.xy <- function(settings, block.list.all, X, obs.mean, obs.cov, t) {
...
     }
     #create H
-    H <- construct_nimble_H(site.ids = site.ids,
-                            var.names = var.names,
-                            obs.t = obs.mean[[t]],
-                            pft.path = settings[[1]]$run$inputs$pft.site$path,
-                            by = "block_pft_var")
+    # if there is any site that has zero observation.
+    if (any(obs_per_site == 0)) {
+      f.2.y.ind <- which(grepl(unique(names(unlist(obs.mean[[t]] %>% purrr::set_names(NULL)))), var.names, fixed = T))
```
infotroph, line 176: This looks to me as if the unique(...) will always return NULL and make grepl throw an error. Am I missing something?

infotroph: Oh, maybe the `unlist()` has the effect of bringing names out from objects inside the list? If so I wonder if there's a clearer way to write that.

DongchenZ: So here we are actually matching the names of vector to the names of each state variable. The `set_names(NULL)` is trying to remove names of site ids, otherwise we will have something like `1000004945.LAI`, which will then fail the name matching.

modules/assim.sequential/R/Analysis_sda_block.R
DongchenZ: Fixed, thank you!

modules/assim.sequential/R/Analysis_sda_block.R

infotroph: Nice improvement! Thanks.

infotroph (approved):

PR #3267, opened 2024-03-01
Rebuild logic between partial missing observations and complete missing observation (free run case), and fixed some edge cases for the wishart Q.

DongchenZ:
Rebuild logic between partial missing observations and complete missing observation (free run case), and fixed some edge cases for the wishart Q.

## Description

The free run is different than a normal run with partial missing observations for specific sets of sites. For the free run, we will estimate all state variables based on previous states and the potential Q from outside. For the normal run with partial missing observations, we will estimate state variables that match up with maximum types of observations across all sites. For the Wishart Q, some edge cases such as single site with single observation will be treated as Normal-Gamma conjugacy instead of MVN-Wishart conjugacy.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.sequential/R/Analysis_sda_block.R
+++ b/modules/assim.sequential/R/Analysis_sda_block.R
@@ -102,7 +102,6 @@ build.block.xy <- function(settings, block.list.all, X, obs.mean, obs.cov, t) {
   } else if (settings$state.data.assimilation$q.type == "wishart") {
     q.type <- 4
   }
-  
   #grab basic arguments based on X.
   site.ids <- unique(attributes(X)$Site)
   var.names <- unique(attributes(X)$dimnames[[2]])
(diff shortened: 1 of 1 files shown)


modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -174,8 +174,8 @@ build.block.xy <- function(settings, block.list.all, X, obs.mean, obs.cov, t) {
       #name matching between observation names and state variable names.
       f.2.y.ind <- obs.mean[[t]] %>% 
         purrr::map(\(x)which(var.names %in% names(x))) %>% 
-        base::unlist %>% 
-        base::unique
+        unlist %>%
```
mdietze, line 177: did you have to remove the namespace here? There's a chance GH Actions is going to complain

infotroph: I haven't yet found a case where R complains about removing `base::` -- the builtins really are special that way

mdietze (approved):

PR #3272, opened 2024-03-07
Add downscale_function and covariates scripts

JoshuaPloshay:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/assim.sequential/R/downscale_function.R
+++ b/modules/assim.sequential/R/downscale_function.R
@@ -3,10 +3,10 @@
 ##' @author Joshua Ploshay
 ##'
 ##' @param data  In quotes, file path for .rds containing ensemble data.
-##' @param focus_year In quotes, if SDA site run, format is yyyy/mm/dd, if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'data'.
+##' @param coords In quotes, file path for .csv file containing the site coordinates, columns named "lon" and "lat".
+##' @param date In quotes, if SDA site run, format is yyyy/mm/dd, if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'data'.
 ##' @param C_pool In quotes, carbon pool of interest. Name must match carbon pool name found within file supplied to 'data'.
-##' @param covariates In quotes, file path of SpatRaster stack, used as predictors in randomForest. Layers within stack should be named.
-##' @param cords In quotes, file path for .csv file containing the site coordinates, columns named "lon" and "lat".
+##' @param covariates Should be loaded in using 'covariates' instructions in inst folder
 ##' @details This function will downscale forecast data to unmodeled locations using covariates and site locations
 ##'
 ##' @description This function uses the randomForest model.
@@ -14,25 +14,21 @@
 ##' @return It returns the `downscale_output` list containing lists for the training and testing data sets, models, and predicted maps for each ensemble member.
 
 
-NA_downscale <- function(data, cords, covariates, focus_year, C_pool){
-  
-  # Read in the covariates and set CRS to EPSG:4326
-  covariates <- terra::rast(covariates) # ADD package to every function
-  terra::crs(covariates) <- "EPSG:4326"
+NA_downscale <- function(data, coords, date, C_pool){
   
   # Read the input data and site coordinates
   input_data <- readRDS(data)
-  site_coordinates <- terra::vect(readr::read_csv(cords), geom=c("lon", "lat"), crs="EPSG:4326")
+  site_coordinates <- terra::vect(readr::read_csv(coords), geom=c("lon", "lat"), crs="EPSG:4326")
   
   # Extract the carbon data for the specified focus year
-  index <- which(names(input_data) == focus_year)
+  index <- which(names(input_data) == date)
   data <- input_data[[index]]
   carbon_data <- as.data.frame(t(data[which(names(data) == C_pool)]))
   names(carbon_data) <- paste0("ensemble",seq(1:ncol(carbon_data)))
   
   # Extract predictors from covariates raster using site coordinates
   predictors <- as.data.frame(terra::extract(covariates, site_coordinates)) 
-  predictors <- dplyr::select(predictors, -1)
+  predictors <- dplyr::select(predictors, -ID)
   
   # Combine each ensemble member with all predictors
   ensembles <- list()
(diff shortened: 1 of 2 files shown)


modules/assim.sequential/R/downscale_function.R
```diff
@@ -3,36 +3,32 @@
 ##' @author Joshua Ploshay
 ##'
 ##' @param data  In quotes, file path for .rds containing ensemble data.
-##' @param focus_year In quotes, if SDA site run, format is yyyy/mm/dd, if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'data'.
+##' @param coords In quotes, file path for .csv file containing the site coordinates, columns named "lon" and "lat".
+##' @param date In quotes, if SDA site run, format is yyyy/mm/dd, if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'data'.
 ##' @param C_pool In quotes, carbon pool of interest. Name must match carbon pool name found within file supplied to 'data'.
-##' @param covariates In quotes, file path of SpatRaster stack, used as predictors in randomForest. Layers within stack should be named.
-##' @param cords In quotes, file path for .csv file containing the site coordinates, columns named "lon" and "lat".
+##' @param covariates Should be loaded in using 'covariates' instructions in inst folder
```
mdietze, line 9: Previous comment was actually more informative. The existence of the covariates script seems like something that would go in the details

modules/assim.sequential/R/downscale_function.R
```diff
@@ -3,36 +3,32 @@
...
 ##'
 ##' @return It returns the `downscale_output` list containing lists for the training and testing data sets, models, and predicted maps for each ensemble member.
 
 
-NA_downscale <- function(data, cords, covariates, focus_year, C_pool){
-  
-  # Read in the covariates and set CRS to EPSG:4326
-  covariates <- terra::rast(covariates) # ADD package to every function
-  terra::crs(covariates) <- "EPSG:4326"
+NA_downscale <- function(data, coords, date, C_pool){
```
mdietze, line 17: why has `covariates` been removed as a function argument when you clearly need this information for the function to work?

modules/assim.sequential/R/downscale_function.R
```diff
@@ -3,36 +3,32 @@
...
-  index <- which(names(input_data) == focus_year)
+  index <- which(names(input_data) == date)
   data <- input_data[[index]]
   carbon_data <- as.data.frame(t(data[which(names(data) == C_pool)]))
   names(carbon_data) <- paste0("ensemble",seq(1:ncol(carbon_data)))
   
   # Extract predictors from covariates raster using site coordinates
   predictors <- as.data.frame(terra::extract(covariates, site_coordinates)) 
-  predictors <- dplyr::select(predictors, -1)
+  predictors <- dplyr::select(predictors, -ID)
```
mdietze, line 31: I don't see where `ID` is being defined/calculated in the code above.

modules/assim.sequential/R/downscale_function.R
```diff
@@ -61,25 +57,25 @@ NA_downscale <- function(data, cords, covariates, focus_year, C_pool){
...
-    output[[i]] <- randomForest::randomForest(ensembles[[i]][[1]][["carbon_data"]] ~ land_cover+tavg+prec+srad+vapr+nitrogen+phh2o+soc+sand,
-                                data = ensembles[[i]][[1]],
-                                ntree = 1000,
-                                na.action = stats::na.omit,
-                                keep.forest = T,
-                                importance = T)
+    rf_output[[i]] <- randomForest::randomForest(ensembles[[i]][[1]][["carbon_data"]] ~ land_cover+tavg+prec+srad+vapr+nitrogen+phh2o+soc+sand,
+                                                 data = ensembles[[i]][[1]],
+                                                 ntree = 1000,
+                                                 na.action = na.omit,
```
mdietze, line 65: why has `stats::na.omit` been reverted to `na.omit`, this will lead to a GH Actions error

modules/assim.sequential/inst/covariates.R
```diff
@@ -0,0 +1,109 @@
...
+
+#### WorldClim ####
+# WorldClim v2.1 (1970-2000) historical climate data
+# Data can be found at https://www.worldclim.org/data/worldclim21.html
+# Product is a zip file containing 12 .tif files, one for each month
+# Use code below to average the 12 files to obtain one map
+# 2023 REU project used 10 minute spatial resolution
+
+## Solar Radiation (kJ m-2 day-1)
+srad <- terra::rast(list.files(path = "/projectnb/dietzelab/jploshay/pecan_copy/jploshay/10m_srad",
```
mdietze, line 17: For anyone to be able to reuse this script they will have to redefine a LOT of local paths. Better would be to define ONE folder where all the data is located at the top of the script and then express all the other paths relative to that (e.g. using `file.path()`)

modules/assim.sequential/inst/covariates.R
```diff
@@ -0,0 +1,109 @@
...
+## Soil Organic Carbon (g kg-1)
+soc <- geodata::soil_world(var = "soc", depth = 5, stat = "mean", path = tempdir())
+
+## Soil Sand (%)
+sand <- geodata::soil_world(var = "sand", depth = 5, stat = "mean", path = tempdir())
+
+
+#### Land Cover ####
+GLanCE_extract <- function(pattern, path) {
+  files <- list.files(path = "/projectnb/dietzelab/dietze/glance2012/e4ftl01.cr.usgs.gov/MEASURES/GLanCE30.001/2012.07.01", #make this path default
```
mdietze, line 68: I'm pretty sure you don't want to hard code a file path within a function. Much better to make it an argument. indeed, this function already has a `path` argument that doesn't seem to be used anywhere.

modules/assim.sequential/inst/covariates.R
```diff
@@ -0,0 +1,109 @@
...
+nitrogen <- geodata::soil_world(var = "nitrogen", depth = 5, stat = "mean", path = tempdir())
+
+## Soil Organic Carbon (g kg-1)
+soc <- geodata::soil_world(var = "soc", depth = 5, stat = "mean", path = tempdir())
+
+## Soil Sand (%)
+sand <- geodata::soil_world(var = "sand", depth = 5, stat = "mean", path = tempdir())
+
+
+#### Land Cover ####
```
mdietze, line 66: For files that are manually downloaded, your earlier examples provided some indication of where the data is located and how to download it. Equivalent info is missing here.

modules/assim.sequential/inst/covariates.R
```diff
@@ -0,0 +1,109 @@
...
+  return(GLanCE_product)
+}
+
+# Integer identifier for class in the current year
+land_cover <- GLanCE_extract(pattern = "NA_LC.tif$")
+
+
+#### Resample and Stack Maps ####
+# Define the extent to crop the covariates to North America
+NA_extent <- terra::ext(-178.19453125, -10, 7.22006835937502, 83.5996093750001)
```
mdietze, line 85: This would be good to have at the top of the script in case a user wants/needs to set a different extent.

mdietze (changes requested):

modules/assim.sequential/R/downscale_function.R
mdietze, line 9:
```suggestion
##' @param covariates SpatRaster stack, used as predictors in randomForest. Layers within stack should be named. Recommended that this stack be generated using 'covariates' instructions in assim.sequential/inst folder
```

modules/assim.sequential/R/downscale_function.R
mdietze:
```suggestion
NA_downscale <- function(data, coords, date, C_pool, covariates){
```

modules/assim.sequential/R/downscale_function.R
mdietze:
```suggestion
                                                 na.action = stats::na.omit,
```

mdietze (changes requested):

mdietze:
@JoshuaPloshay this PR is failing because of a conflict between the updated Roxygen and the not-yet-updated man page. Could you pull the changes I've made into your local branch, then rerun make document, then commit the updated NA_downscale.Rd?

mdietze:
@JoshuaPloshay I believe you have code that loads up existing PEcAn output, calls your downscale function, and does some analyses and visualizations of the outputs (including quantile checks). It would be good to add such code to this PR (e.g. in the /inst folder) so that others don't need to recreate this to be able to pick up on where your analyses left off.

modules/assim.sequential/R/downscale_function.R
infotroph: Looks like it was added by `terra::extract` on line 34 above, and could be suppressed by by passing `ID = FALSE` there

mdietze (approved):

PR #3274, opened 2024-03-08
update to STICS wrappers to use stics R packages

istfer:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


base/db/R/dbfiles.R
```diff
@@ -43,10 +43,10 @@ dbfile.input.insert <- function(in.path, in.prefix, siteid, startdate, enddate,
                                 parentid = NA, con, hostname = PEcAn.remote::fqdn(), allow.conflicting.dates = FALSE, ens = FALSE) {
   name <- basename(in.path)
   hostname <- default_hostname(hostname)
-
+
```
infotroph, line 46: Looks like all the changes to this file are adding trailing whitespace -- can it be dropped from the PR?

models/stics/inst/prof.mod
```diff
@@ -0,0 +1,4 @@
+2
```
infotroph, line 1: I'm guessing this filename is set by Stics and not under our control, but if it can be changed: I'm used to assuming `*.mod` are compiled Fortran modules that should be deleted any time I want a clean compilation. Is there a risk of someone naively trying something like `find pecan/ -name '*.mod' -delete` and then wondering why Stics broke?

infotroph (approved):

models/stics/R/write.config.STICS.R
```diff
@@ -54,65 +159,1017 @@ write.config.STICS <- function(defaults, trait.values, settings, run.id) {
...
+    # skipping the other parameters related to this switch, they don't seem influential, at least on NPP and LAI
+    # potgermi: soil water potential under which seed imbibition is impeded
+    # nbjgerlim: maximum number of days after grain imbibition allowing full germination
+    # propjgermin: minimal proportion of the duration nbjgerlim when the temperature is higher than the temperature threshold Tdmax
+    
+    
+    # parameter of the curve of coleoptile elongation
+    if ("belong" %in% pft.names) { 
+      SticsRFiles::set_param_xml(plant_file, "belong", pft.traits[which(pft.names == "belong")], overwrite = TRUE)
+    }
```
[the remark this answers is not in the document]
istfer: the idea was to change these STICS specific names (e.g. `belong`) to more generic Bety names in the future, but it's a bit of a wishful thinking :) 

Also writing param names out explicitly helps keep track of the "overwritten" variables and gives a bit more specific control over what to pass via the priors dataframe

models/stics/inst/prof.mod
istfer: yes the filename is what STICS expects and I'm not sure if it can be changed. Stics will break for many other reasons before it comes to this 😄 but joking aside, Stics writes out error logs and I'm sure it will say that it cannot find `X.mod` file in such a case, which should help find the cause

PR #3275, opened 2024-03-10
[Bug fix] : Fixed `gsub` issue in `write.config.SIPNET`

meetagrawal09:
## Description
Fixed issues with local deployment for `SIPNET` model. 

The variables `cpruncmd` and `cpoutcmd` were added to `write.config.SIPNET` to extend it for cloud based PEcAn deployment. On a local deployment setup the workflow for SIPNET model was breaking with the below error : 

```
Error in gsub("@CPRUNCMD@", cpruncmd, jobsh) :
  invalid 'replacement' argument
Calls: <Anonymous> ... <Anonymous> -> do.call -> write.config.SIPNET -> gsub
> proc.time()
   user  system elapsed
  7.132   3.362   7.877
``` 

The error arises since local `pecan.xml` does not contain `settings$host$rabbitmq$cpfcmd` and `settings$host$rabbitmq$cpfcmd` which is used to set  `cpruncmd` and `cpoutcmd` in the `write.config.SIPNET` file.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -102,8 +102,12 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
   jobsh <- gsub("@BINARY@", settings$model$binary, jobsh)
   jobsh <- gsub("@REVISION@", settings$model$revision, jobsh)
   
-  jobsh <- gsub("@CPRUNCMD@", cpruncmd, jobsh)
-  jobsh <- gsub("@CPOUTCMD@", cpoutcmd, jobsh)
+  if(!is.identical(cpruncmd, character(0)) {
+    jobsh <- gsub("@CPRUNCMD@", cpruncmd, jobsh)
```
mdietze, line 106: Couple questions:
* are we sure that if these variables aren't defined in the pecan.xmxl that they'll be set to character(0)?
* If these variables don't exist, are we sure the job.sh will run correctly with the text of these tags left in the shell script? looking at the template job, it seems like they should be removed (i.e. gsubbed with an empty string)
https://github.com/PecanProject/pecan/blob/456433715edc46e9f06319cb66771f2f90a03697/models/template/inst/template_geo.job#L46

If the right thing to do is gsub with an empty string, and the is.identical is assuming the default is an empty string, this begs the question of why the default case is failing

infotroph: @meetagrawal09 I agree with your diagnosis, but I think the issue should be fixed on lines 76-77 rather than here. Maybe that could be something like `cpruncmd <- gsub("@OUTDIR@", settings$host$rundir, settings$host$rabbitmq$cpfcmd %||% "")`, then the replacements on lines 105-106 remain unconditional?

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -73,6 +73,7 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
   cpruncmd <- cpoutcmd <- rmoutdircmd <- rmrundircmd <- ""
   if (!is.null(settings$host$rabbitmq)) {
     #rsync cmd from remote to local host.
+    settings$host$rabbitmq$cpfcmd <- if(settings$host$rabbitmq$cpfcmd == NULL) "" else settings$host$rabbitmq$cpfcmd
```
mdietze, line 76: @meetagrawal09 this syntax doesn't work in R. First, `if(settings$host$rabbitmq$cpfcmd == NULL)` throws an "argument is of length zero" error if the variable is NULL and should be replaced with is.null(). Second, the `if ... else` syntax should be replaced with the `ifelse()` function.

mdietze (changes requested):

models/sipnet/R/write.configs.SIPNET.R
meetagrawal09: Don't know but this operator did not work for me. Shows me the below error. 
```
Error in settings$host$rabbitmq$cpfcmd %||% "" :  could not find function "%||%"
```

mdietze (approved):

models/sipnet/R/write.configs.SIPNET.R
infotroph: Oh, hm. 
(1) You'll need to import it[1] from rlang with `#' @importFrom rlang %||%` or just copy in the definition: `function (x, y) { if (is.null(x)) y else x}` ...
(2) ...which shows I was remembering wrong and `%||%` only takes scalar inputs, so my advice was bad to start with 🤦 

So `ifelse` it is!

[1]  (`%||%` has been added to base R in 4.4.0, so a few version bumps from now we'll be able to use it with no import)

infotroph (approved):

PR #3282, opened 2024-04-02
Create Landtrender.AGB.Model [Work In Progress]

tamigord:
Creation of Landtrender.AGB.Model - Help still needed with checking system for old csv files

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+        "agbd_pi_lower",
+        "agbd_pi_upper",
+        "agbd"
+      )
+    
+     gediL4_output <- l4_getmulti(l4,add_col = col, source= T)
+```
+```{r}
+# If using predownloaded dataset (saved as HDF5 file format) - (!) Could be automated to ask or check computer for file paths 
+    site_info = list.files(path =".", pattern = "h5", all.files = T, full.names = T)
```
mdietze:123]: 1. earlier you defined site_info as a function input argument, so you don't want to redefine it here
2. in all of the other data constraint functions, site_info provides the site ids, lat, lon, etc. Your function won't work with the larger PEcAn system if you're using the same variable to mean something very different here (vector of hdf5 files)
3. looking for files in the current directory also isn't a general solution. This seems like a folder that should be passed into the function as an argument (e.g. is this outdir? or a subdir off of outdir?)

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+    View(site_info)
+    if(length(site_info) != 0){
+       dataname <- l4_getmulti(site_info[[1]],just_colnames = T)
+      gediL4_path <- l4_getmulti(site_info, catch = T, merge=T)
+      ## Dataset with predownloaded data
+        gediL4_pre <- l4_getmulti(site_info,just_colnames = F, add_col = col, source=T, catch = T)
```
mdietze:129]: Unclear why the same function needs to be called 3 different times with different arguements. I don't know these functions well enough to know if this is right or wrong, but it definitely seems like a place where more documentation is needed to explain the intent of this block of code

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+        "agbd"
+      )
+    
+     gediL4_output <- l4_getmulti(l4,add_col = col, source= T)
+```
+```{r}
+# If using predownloaded dataset (saved as HDF5 file format) - (!) Could be automated to ask or check computer for file paths 
+    site_info = list.files(path =".", pattern = "h5", all.files = T, full.names = T)
+    View(site_info)
+    if(length(site_info) != 0){
```
mdietze:125]: If length is 0, will the later code still work? Or do you need an `else` that either defines what all the variables below should be or to end the function with a return?

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+       dataname <- l4_getmulti(site_info[[1]],just_colnames = T)
+      gediL4_path <- l4_getmulti(site_info, catch = T, merge=T)
+      ## Dataset with predownloaded data
+        gediL4_pre <- l4_getmulti(site_info,just_colnames = F, add_col = col, source=T, catch = T)
+        rbind(gediL4_output, add_col = gediL4_pre, use.names = FALSE, fill = T)
```
mdietze:130]: The output of this rbind isn't being assigned to anything

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+
+```
+
+```{r}
+ ## Find and download GEDI Data within your study area:
+    ## site_info (list as data.table) tells the model the data files' site_id and its site_info 
+      #list all dataset in h5 file
+ 
+  ## Clipping Dataset
+  l4_data <- l4_getmulti(l4)
```
mdietze:143]: Similar to earlier comment, this seems to be the 4th call to l4_getmulti without clear explanation of how they're different

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+  #clip using vector of coordinates
+  b_box <- c(-50,35,52,37)
```
mdietze:145]: This seems like something that's leftover from the tutorial you were building from, not something that should be part of the function.  More generally, does our function need to clip to a bounding box in order to extract data for a site, or do we end up doing that in another way? If the latter this can be skipped. If the former, then we probably need a loop where we clip each site +/- whatever we decide the search radius is for an individual site.

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+  clipped <- l4_clip(l4_data,clip=b_box)
+  
+  #using Shapefile to clip
+  bound <- system.file("extdata","bound4326.shp",package="GEDI4R")
+  #with  extension
+  clipped <- l4_clip(l4_data,clip=bound,usegeometry = F)
+  #with  polygon boundaries
+  clipped2 <- l4_clip(l4_data,clip=bound,usegeometry = T)
```
mdietze:153]: We definitely don't need this bit on using a shapefile to clip -- seems leftover from the tutorial

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+  
+  ## l4_convert, which can also reproject the data to a user-defined coordinate reference system
+  converted <- l4_convert(l4_data,epsg = 4326, filename=paste0(outdir,"/example.shp"),return_path = T)
+  example <- sf::read_sf(converted)
+  file.remove(list.files(outdir,pattern = "example",full.names = T))
```
mdietze:158]: Do we need to reproject the GEDI data? If so, we should be doing that before extracting site data to ensure things are aligned correctly.

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+    "agbd_pi_lower",
+    "agbd_pi_upper",
+    "agbd",
+    "date",
+    "source")
+gediL4 <- gediL4_output[,..var_needed]
+
+
+## Save file as csv 
+write.csv(gediL4, "gediL4_output.csv")
```
mdietze:174]: This data doesn't seem to be organized in the way that our downstream data assimilation code is looking for.

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+```
+
+```{r}
+## Plotting Data
+
+#footprints locations and AGBD distribution against elevation
+l4_plotagb(gedil4,type="distribution")
```
mdietze:181]: download function shouldn't be making plots

Creation of Landtrendr.AGB.Model [Work In Progress]
```diff
@@ -0,0 +1,203 @@
...
+
+```
+
+```{r}
+## Check for old CSV and create new CSV 
+outdir = "/Users/tamigordon/Desktop/rGEDI/data files"
```
mdietze:187]: outdir shouldn't be hard coded, it should be a function input

mdietze (commented):
Current code has a good way to go still. I'd recommend getting an example site_info from Dongchen (e.g. the 39 NEON CONUS sites) and work through the steps of grabbing data for different sites than the ones the tutorial you were using are built around. I think that will help move things towards the function we are looking for and help generalize the things that were specific to the tutorial.

infotroph:
@DongchenZ is incorporating / updating this code in his work on GEDI download for the data assimilation workflow, and will include it in a new PR soon. Closing this PR in the meantime to keep the queue manageable.

PR #3291, opened 2024-05-17
Add Script to Identify `Orphaned` Functions in Codebase

Sweetdevil144:
## Description
This script identifies orphaned functions within our system. It iterates through modules, listing functions and checking if they are utilized within their module or externally. Unutilized functions are categorized as purely orphaned, while those used externally are noted separately. The results are output to a CSV file for further analysis.

## Motivation and Context
The motivation behind this change is to clean up the codebase by identifying functions that are no longer in use. This can help in reducing the code size, improving maintainability, and potentially uncovering areas of the code that lack proper integration. It addresses the need for ongoing codebase optimization and cleanliness.

## Review Time Estimate

- [ ] Immediately
- [X] Within one week
- [ ] When possible

## Types of changes

## Checklist:


scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,249 @@
...
+            if (
+                root.__contains__(func_location_module)
+                or root.__contains__("/man")
+                or root.__contains__("/vignettes")
+            ):
+                # print(f"{color_codes[13]} Skipping the module directory {root}{ENDC}")
+                continue
+            else:
+                # print(f"{color_codes[14]} Searching for the function in the directory {root}{ENDC}")
+                for file in files:
```
mdietze, line 162: The code below seems redundant with your earlier search function. Why not just call the "search within module" function here?

mdietze (changes requested):

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,249 @@
...
+    "\033[96m",  # BRIGHT_CYAN-14
+    "\033[97m",  # BRIGHT_WHITE-15
+]
+
+# Reset
+ENDC = "\033[0m"
+
+# List of modules to check for orphaned functions
+# Modules Not included : data.hydrology, DART due to differing file structures. Script should be enhanced further to remove this limitation.
+module_list = [
```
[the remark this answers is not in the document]
Sweetdevil144: For this, we can extract all files from the `modules` directory **IGNORING** those present in `ignored_modules` list. Also, I added our `DART` and `data.hydrology` to `ignored_modules` due to Different Code (`F90` for `DART`) & lack of code (in `data.hydrology`).

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,243 @@
...
+    "\033[95m",  # BRIGHT_MAGENTA-13
+    "\033[96m",  # BRIGHT_CYAN-14
+    "\033[97m",  # BRIGHT_WHITE-15
+]
+
+# Reset
+ENDC = "\033[0m"
+
+# Modules to ignore while checking for orphaned functions
+ignored_modules = ["data.hydrology", "DART"]
```
infotroph, line 32: Thanks for this tip -- I hadn't looked closely at `data.hydrology` in a long time. Now that I have, I think we don't need it and have opened #3292 to delete it entirely.

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,243 @@
...
+    return [module for module in modules if module not in ignored_modules and not module.startswith('.')]
+
+# List of modules to check for orphaned functions
+module_list = get_module_list()
+
+# Initialize two lists to store the orphaned functions
+# List_1: Purely orphaned functions
+# List_2: Functions that are utilized externally
+list_1 = []
+list_2 = []
```
infotroph, line 47: I'd suggest picking more informative variable names here. If these are named something like `funs_unused` and `funs_used_externally`, all the code below will be easier to read / edit without needing to remember whether the unused ones are list 1 or 2, and therefore less susceptible to bugs too.

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # Construct the path to the R directory of the module
+        module_r_code_path = os.path.join(modules_dir, module_name, "R")
+
+        # Print the module name and path
+        # print(color_codes[1] + "Processing module:", module_name, "\n", ENDC)
+        # print(color_codes[6] + "Module path:", module_r_code_path, "\n", ENDC)
+
+        # List all functions in the module
+        functions = list_functions_in_module(module_r_code_path)
```
infotroph, line 215: For me this fails here on modules/data.mining, which has code in `inst` but no `R` directory.

infotroph: Easiest fix is probably to add data.mining to ignored_modules

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # Writing the header for list_1
+        writer.writerow(["Purely Orphaned functions", "module", "function_name"])
+
+        # Writing data from list_1
+        for module, function in list_1:
+            writer.writerow(["", module, function])
+
+        # Writing the header for list_2
+        writer.writerow(["Externally Utilized Orphaned Functions"])
```
infotroph, line 191: I'm puzzled by this classification -- if a function is used externally, in what sense is it orphaned?

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,243 @@
...
+    ]
+
+    # print(color_codes[2] + "Function name:",function_name," \t Function's Location modules :",func_location_module,ENDC,)
+
+    for directory in search_directories:
+        for root, dirs, files in os.walk(directory):
+            if (
+                root.__contains__(func_location_module)
+                or root.__contains__("/man")
+                or root.__contains__("/vignettes")
```
infotroph, line 149: I would count a usage in another package's vignette as an external usage.

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    List all functions defined in the R files of a module.
+
+    Args:
+    module_path (str): The path to the module's R directory.
+
+    Returns:
+    list: A list of function names defined in the module.
+    """
+    function_pattern = re.compile(r"\b([a-zA-Z_]\w*(?:\.\w+)*)\s*<-\s*function")
```
infotroph, line 61: This won't catch `funcname = function(...`, which is legal R and my quick grep says we've used 258 times in this repo.

scripts/get_orphaned_functions.py
Sweetdevil144: Thanks for the tip. Resolved now. Renamed them as below :
```python
purely_orphaned_functions = []
externally_utilized_functions = []
```

Would these be apt for current usage or sgould we change it to any other name?

scripts/get_orphaned_functions.py
Sweetdevil144: Thanks for the heads up. Removed `vignettes` from excluded directory

scripts/get_orphaned_functions.py
infotroph: Those are good. Thanks!

scripts/get_orphaned_functions.py
Sweetdevil144: The Function would be orphaned within it's `OWN` directory while at the same time it is being utilised externally. These are functions that are not used within their own module (hence considered `orphaned` in a local sense) but are called or used by external `packages` that depend on the prior module. These functions might not be directly referenced within the project's codebase but are still necessary for the functionality provided to external consumers. I've some of my own doubts regarding this. Many a times, we'll see such functions' utilization in external `packages` as a coupled dependency within two (or more) different packages. Shifting functions that are identified as "externally utilized orphaned functions" to external packages or creating a new directory to handle all coupled dependencies within different packages in the repository can be a good approach.

scripts/get_orphaned_functions.py
Sweetdevil144: Even better if we recursively recall the function to scan within our `inst` directory too, maybe by passing the directory to be scanned as a function requirement. Maybe we can term this as `scanning_directory` which may be `/R`, `/inst` or any other directory we want to include within it.
To remove redundancy for duplicate count, we can replace our `list` by a `set`.

scripts/get_orphaned_functions.py
infotroph: There's no requirement that a function be called within its own package to be considered useful, so I strongly suggest not referring to these as "orphaned".  I take your point that "defined only here, used only there" might be a candidate for relocation, but my hunch is these will be a small fraction of the functions identified by this criterion.

But! A function that's not called within its own package is very likely _untested_, so it may be worth identifying them on those grounds. But if we want to go into that level of detail it may be worth moving from regexes to a parsing approach that can give more information / allow more control over the exact types of call matched.

scripts/get_orphaned_functions.py
Sweetdevil144: > Note that this won't catch `funcname = function(...`

I guess this expression is only being utilized for our `lambda` functions. However, there still exists a possibility of these lambda's being exported to outer packages for utilization. But, as far as I seem to know, I haven't encountered any such lambda function being utilised externally (still does not rules out any such possibility).

Maybe changing our expression as such may help us in identification of such `lambda` functions: 
```r
function_pattern = re.compile(r"\b([a-zA-Z_]\w*(?:\.\w+)*)\s*(?:<-\s*function|\s*=\s*function)")
```

Do you have any other opinions?

scripts/get_orphaned_functions.py
Sweetdevil144: Oh. That just didn't clicked my mind. Refracted it now.

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,246 @@
...
+    # print(label)
+    color_index = 0  # Start with the first color
+    for i in range(0, len(list_to_print), 5):
+        batch = list_to_print[i : i + 5]
+        # Get the current color
+        color = color_codes[color_index % len(color_codes)]
+        # Print the batch in the current color
+        print(color + ", ".join(map(str, batch)) + ENDC)
+        # Move to the next color for the next batch
+        color_index += 1
```
Sweetdevil144, line 177: Maybe We can remove this Function.
I created this function for better printing in the Terminal so that readability issues are removed.
However, we may still find their usage if anyone tried debugging this code.

So, shall I remove this or Keep this for future use?

scripts/get_orphaned_functions.py
Sweetdevil144: For now, I've moved `data.mining` to `ignored_modules`

scripts/get_orphaned_functions.py
mdietze: Not sure what you mean by "this expression is only being utilized for our lambda functions". I agree with @infotroph that `funcname = function(` is valid and needs to be searched for

scripts/get_orphaned_functions.py
mdietze: I think it'd just warn against trying to over-engineer a script to try and solve multiple different use cases and problems, and instead just focus on the question being asked (are there functions that aren't getting used). Personally, I wouldn't rely on just this function alone to make decisions about whether to delete, deprecate, or move a function, but would instead follow up on proposed orphans (e.g. using grep) to see how and where it is being used.

Per @infotroph point about tests, that cuts both ways -- there could be functions that were designed with lots of tests, but are never used OUTSIDE of their build tests, that would also be considered orphans.

scripts/get_orphaned_functions.py
Sweetdevil144: My bad. I got confused. I've added searching functionality for functional declarations of type `funcname = function(` as well. Can you review it as well?

Sweetdevil144: Also, I've filtered out all such occurances of `error = function(e) ....`

scripts/get_orphaned_functions.py
Sweetdevil144: Refactored it as follow : 
```python
else:
     if is_function_utilized_within_module(function_name, MODULE_DIRECTORY_PATH):
        # print(f'{color_codes[12]} Function "{function_name}" is utilized EXTERNALLY in module: {MODULE_DIRECTORY_PATH}{ENDC}')
        return True
```

scripts/get_orphaned_functions.py
```diff
@@ -0,0 +1,249 @@
...
+
+    Returns:
+    bool: True if the function is utilized within the module, False otherwise.
+    """
+
+    # Regex pattern to detect usage of the function
+    function_pattern = re.compile(r"\b" + re.escape(function_name) + r"\b\s*\(")
+
+    # Traverse all files in the given module R directory
+    for r_file_name in os.listdir(module_r_path):
```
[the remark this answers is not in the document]
Sweetdevil144: Updated file as per requirements in a test branch [here](https://github.com/Sweetdevil144/pecan/blob/orphan-test-branch/scripts/get_orphaned_functions.py). Can you verify if this is correctly performed?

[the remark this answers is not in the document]
infotroph: @Sweetdevil144 Please go ahead and commit those changes to the existing PR instead of leaving them in another branch -- it takes a fair number of extra steps for us to go find the diffs between your branches, and once we've found it there's no easy UI for us to leave comments on the changes. If instead you commit it to the existing `develop-ophanedFunctions` branch that this PR is using, then they'll show up here and we can easily discuss them in context.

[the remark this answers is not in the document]
Sweetdevil144: @infotroph Pushed. Can you review the changes once?

mdietze (approved):

PR #2813, opened 2021-06-26
Adding ICOS download functionality

ayushprd:
## Description
This PR adds the functions for downloading and processing the ICOS ecosystem data products.
- [x] download function
- [x] met2cf conversion
- [x] registration file

## Motivation and Context
Closes #2594 Closes #2802 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3295, opened 2024-05-21
Add NEWS.md to data.atmosphere

infotroph:
Tried to summarize the changes I could see in Git history, but a lot of these are components I don't know well -- please correct my descriptions where I got it wrong, and add developments I've missed

Help wanted especially on the entries with question marks in them.


modules/data.atmosphere/NEWS.md
```diff
@@ -0,0 +1,36 @@
...
+* `download.MERRA()` now respects `overwrite = FALSE` [@meetagrawal09, #3169]
+* `download.MERRA()` can now find data from September of 2020 [@KristinaRiemer, #2888]
+* `download.Fluxnet2015()` now gives a more useful message when passed a malformed URL [@meetagrawal, #3130]
+
+## Added
+
+* Newly supported met data sources:
+	- ERA5 via `ERA5_met_process()` [@DongchenZ, #2886]
+	- NOAA GEFS via `download_NOAA_GEFS_EFI()` [@helge22a, #3174]
+		(replaces NOAA_GEFS_downscale?)
```
mdietze, line 17: No, doesn't replace. The EFI version can only grab files for the subset of sites on data.ecoforecast.org, which at the moment is just NEON. We still need to maintain the original GEFS download functions for all other sites

modules/data.atmosphere/NEWS.md
```diff
@@ -0,0 +1,36 @@
...
+	- ICOS?
+	- MERRA?
+* half-hourly downscaling via `downscale_*_to_half_hrly()`?
```
mdietze, line 20: I'm not sure about these 3. What PRs were they associated with?

infotroph: Hmm, looks like all of these were added prior to the 1.7.2 release:
* ICOS: https://github.com/PecanProject/pecan/pull/2813
* MERRA: https://github.com/PecanProject/pecan/pull/2725
* half-hourly: https://github.com/PecanProject/pecan/pull/2800

I'll remove and let people find them in the big changelog.

modules/data.atmosphere/NEWS.md
infotroph, line 17:
```suggestion
	- NOAA GEFS data hosted by the Ecological Forecasting Initiative, via `download_NOAA_GEFS_EFI()` [@helge22a, #3174] (For GEFS data directly from NOAA, continue to use `download.NOAA_GEFS`)
```

modules/data.atmosphere/NEWS.md
infotroph, line 20:
```suggestion
```

mdietze (approved):

PR #3298, opened 2024-05-27
Unit Testing functions in `meta-analysis` sub-workflow

meetagrawal09:
## Description

This PR aims at adding unit tests for all the pivotal function in the `meta-analysis` sub-workflow.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

- [] I have added tests to cover my changes.


modules/meta.analysis/tests/testthat/test.run.meta.analysis.R
```diff
@@ -1,42 +1,11 @@
-#-------------------------------------------------------------------------------
-# Copyright (c) 2012 University of Illinois, NCSA.
-# All rights reserved. This program and the accompanying materials
-# are made available under the terms of the 
-# University of Illinois/NCSA Open Source License
-# which accompanies this distribution, and is available at
-# http://opensource.ncsa.illinois.edu/license.html
-#-------------------------------------------------------------------------------
+test_that("`runModule.run.meta.analysis`", {
```
dlebauer, line 1: note that the argument to `test_that` should be a description of the tests within. The `context` function can name the function or set of functions being tested.

infotroph: The testthat docs now discourage use of `context` and recommend encoding that info in the filename

modules/meta.analysis/tests/testthat/test.jagify.R
```diff
@@ -0,0 +1,33 @@
...
+                           cultivar_id = 1,
+                           specie_id = 1,
+                           n = 2,
+                           mean = sqrt(1:10),
+                           stat = 1,
+                           statname = "SE",
+                           treatment_id = 1:10
+  )
+  i <- sapply(testresult, is.factor)
+  testresult[i] <- lapply(testresult[i], as.character)
```
infotroph, line 29: It's probably save to drop these lines:

1. You can get the same effect by passing `stringsAsFactors = FALSE` to the `data.frame` call
2. In R >= 4.0.0 (as all our supported versions are), stringsAsFactors is FALSE by default, so all of this is a no-op.

modules/meta.analysis/R/meta.analysis.R
```diff
@@ -9,7 +9,7 @@
```
meetagrawal09, line 9: @infotroph, I think we should remove these outdated license blocks before a release. Should I remove all of them for the `meta.analysis` package and make a commit or doing it as a separate PR would be better ?

infotroph: @meetagrawal09 It should be a separate PR that will merge into branch `release/v1.8.0` instead of into `develop`. We'll need to double-check that all code owners have signed off on the license change before we merge, but do go ahead and open the PR please.

modules/meta.analysis/tests/testthat/test.jagify.R
```diff
@@ -0,0 +1,31 @@
+#-------------------------------------------------------------------------------
+# Copyright (c) 2012 University of Illinois, NCSA.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the 
+# University of Illinois/NCSA Open Source License
+# which accompanies this distribution, and is available at
+# http://opensource.ncsa.illinois.edu/license.html
+#-------------------------------------------------------------------------------
```
infotroph:
```suggestion
```

infotroph (approved):

PR #3300, opened 2024-05-30
Remove dependency on `data.atmosphere` from `data.land`

Sweetdevil144:
Updated corresponding changes to  `CHANGELOG.md` as per suggestions

## Description

Replaced the `PEcAn.data.atmosphere::db.site.lat.lon` function with `PEcAn.DB::query.site` for retrieving site latitude and longitude in the `soil_process` function.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (changes requested):
Could you describe how these functions have been tested to ensure that they still work? It's been a while since I've used either `PEcAn.DB::query.site` or `PEcAn.data.atmosphere::db.site.lat.lon` to know how/if the data structures they return might differ and how that might impact downstream code.

infotroph (changes requested):

Sweetdevil144:
@infotroph was correct. Seems like I failed to `push` my changes after I only **committed** them. My fault :) .
More clarity into the above changes @mdietze . To check whether The above changes were valid or not, I created my custom `test case` as below, other than `make test` to be ran :

```r
test_that("query.site and db.site.lat.lon return the same results", {
  db_params <- PEcAn.DB::get_postgres_envvars(
    dbname = "bety",
    user = "bety",
    password = "bety",
    host = "localhost",
    driver = "Postgres"
  )
  con <- PEcAn.DB::db.open(db_params)
  site <- list(id = 1, sitename = "Aliartos")

  latlon1 <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")]
  latlon2 <- PEcAn.data.atmosphere::db.site.lat.lon(site$id, con = con)
  expect_equal(latlon1$lat, latlon2$lat)
  expect_equal(latlon1$lon, latlon2$lon)

  PEcAn.DB::db.close(con)
})
```

Results were as follow :

```vbnet
root@3c52500f363a:/pecan# Rscript -e "testthat::test_file('modules/data.atmosphere/tests/testthat/test-site-latlon.R')"

══ Testing test-site-latlon.R ════════════════════════════════════════════════════════════════
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 2 ] Done!
```

mdietze (approved):

Sweetdevil144:
_By @infotroph in Slack,_
>I disagree — seems safe to deleting it after replacing all the uses of `db.site.lat.lon` inside `met.process` with `query.site` same as you’re doing in data.land.

If we're doing this Fix, then shouldn't we add the above recommended changes too in this PR?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ see if you need to change any of these:
   to star PEcAn with https.
 - TRAEFIK_HTTPS_OPTIONS is no longer used, this is the default when you use https.
 - TRAEFIK_HTTP_REDIRECT is no longer used, this is the default when you use https.
+- Updated `soil_process` function in `soil_process.R` and `ic_process` to use `PEcAn.DB::query.site` instead of `PEcAn.data.atmosphere::db.site.lat.lon` for retrieving site latitude and longitude.
```
infotroph, line 23: This change won't be visible to end users, so it doesn't really need a line in the PEcAn changelog. 
Instead, please add an entry in the NEWS.md of data.land stating that  data.land no longer depends on data.atmosphere, including a reference to this PR number and a credit to yourself for the change.

infotroph:
@Sweetdevil144 Last step is to remove `PEcAn.data.atmosphere` from the DESCRIPTION file of `PEcAn.data.land`, then run `./scripts/generate_dependencies.R` to propagate that change through the dependency lists.

modules/data.land/R/write_ic.R
```diff
@@ -55,7 +55,8 @@ write_ic <- function(in.path, in.name, start_date, end_date,
   # read in registration xml for pool specific information
   register.xml <- system.file(paste0("register.", model$type, ".xml"), package = paste0("PEcAn.", model$type))
   if(file.exists(register.xml)){
-    register     <- PEcAn.data.atmosphere::read.register(register.xml, con = NULL)
+    register     <- XML::xmlToList(XML::xmlParse(register.xml))
+    PEcAn.logger::logger.debug(as.data.frame(register))
```
infotroph, line 59: I assume this is leftover from testing the change
```suggestion
```

infotroph (changes requested):

modules/data.land/R/write_ic.R
Sweetdevil144: My bad. I forgot to remove these.

Sweetdevil144:
Not sure why the tests fail?

infotroph:
@Sweetdevil144 The tests are failing because we use exact string matches to ignore an existing warning that "Imports includes 35 non-default packages.", and you just improved it to "Imports includes 34 non-default packages." 🤣 

This is one of the rare-ish cases where the right fix is to manually edit the saved check outputs. Will you please add a commit that changes `35` to `34` on line 16 of `modules/data.land/tests/Rcheck_reference.log` (and does nothing else)?

Sweetdevil144:
>  we use exact string matches to ignore an existing warning

That's my first time seeing this in PEcAn atmosphere 🥲.
I've been debugging what's wrong since morning. Thanks a lot for the aid !!

infotroph:
> I've been debugging what's wrong since morning

I'm sorry for the frustration! You're not alone: This cached-result-comparison step is way too complicated and causes a lot of grief for the whole team. Has it produced enough of an increase in code quality to be worth it? Weeeeell... we hope so but can't prove it.

infotroph (approved):
💯

PR #3311, opened 2024-06-24
release v1.8.0

robkooper:
This is the same as #3305

Things to do:
- [x] update version numbers
- [ ] test containers

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (approved):

dlebauer (commented):
**General comments**

I was able to successfully get through running the sipnet workflow. 

I've added a few suggestions inline - not sure if these should be required for v1.8.0 release, but at a high level I'd suggest

- changing all mentions of deprecated `docker-compose` to `docker compose`
- removing any documentation specific to Windows. Since docker requires WSL2, we can assume it is installed and explain that all subsequent instructions assume users are using WSL2 or Ubuntu unless otherwise indicated (e.g. for MacOS).

_if you agree with either or both of these, I am happy to implement_

**missing traefik**

At the 

```
docker-compose up -d
```

Step I got the error

```
docker-compose up -d
# lots of Container pecan-xxxx-1 Creating
Error response from daemon: No such image: traefik:v2.9 
```

Following guidance from @infotroph, this worked

```
docker pull traefik:v2.9
docker compose up -d
```

Did I miss a step? Is something missing in a docker-compose*yml file, or does this need to be added to DEV-INTRO.md? other?

**One other comment:**

these instructions appear in docker-compose.override.yml

```
# cp docker/web/config.docker.php web/config.php
```

should it, along with more specific instructions about what to change in this file, be in the DEV-INTRO.md?

infotroph (approved):

Issue #3307, opened 2024-06-19
Refactor `convert_input.R` to Optimise Workflow


Sweetdevil144:
## Tasks

- [ ] Combine multiple queries using `PEcAn.DB::db.query` to fetch required data at once and make further checks on that basis.
> *condense database lookups so that they happen all at once in one or a couple steps rather than spread out through every code path*

- [ ] Reuse, Reduce and Pass functional Arguments wherever they can be rather than looking up in `DB` every time we need the required information
> >How can we be so sure we have the info available offline to us

> I’m thinking less of maintaining the same information in another location and more of thinking about things 
> + that can be passed around as function arguments rather than looked up again
> + are currently looked up but never used 
> + that are currently used but we can find a way to do without them entirely etc

*--> Chris in Slack* [Link](https://pecanproject.slack.com/archives/C0256U1S48Y/p1718722630477789)

- [ ] Big picture goal is to make entire workflow less dependent on `PEcAn.DB`

### Low Priority Tasks

**These may be decided after discussion with `PEcAn` admins. This can be ignored for now, but it can also be tackled.**

- [ ] Improper variable declaration within restricted code blocks which may cause `variable out-of-scope` errors
- [ ] Small support as discussed with Chris: remove support of `db.site.lat.lon` function completly and replace it with `query.site(params)['lat','lon']`

### Suggestion

*Maybe we can try this for a better lead*

- [ ] Break the `convert_input` and `do_conversions` into several smaller functions. Once this is done, many other steps would become clear to us as for where to optimise the file.
- [ ] Eliminate Bad system config. For example in `convert_input` we have `} else if (ensemble > 1) {` where ensemble was previously declared to be of type `bool`.
- [ ] Eliminate other `WARN` types which shouldn't have been occuring?

## Context

The proposed changes are part of the **GSoC project** aimed at optimizing the `PEcAn` workflow. Currently, the workflow involves multiple database lookups and queries, which can lead to performance Issues.

## Possible Implementation

TBD *or* as mentioned in above task points.

Sweetdevil144:
Fixed by #3338

PR #3312, opened 2024-06-24
Combine multiple `PEcAn.db` calls in a single query

Sweetdevil144:
## Description

This is an example demonstration of how we may proceed with : Refactor `convert_input.R` to Optimise Workflow #3307

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/read.register.R
+++ b/modules/data.atmosphere/R/read.register.R
@@ -35,10 +35,14 @@ read.register <- function(register.xml, con) {
     } else if ((!is.null(register$format$id) & is.null(register$format$name)) 
                | 
                (!is.null(register$format$id) & is.null(register$format$mimetype))) {
-      register$format$name <- PEcAn.DB::db.query(
-        paste("SELECT name from formats where id = ", register$format$id), con)[[1]]
-      register$format$mimetype <- PEcAn.DB::db.query(
-        paste("SELECT mime_type from formats where id = ", register$format$id), con)[[1]]
+      # get name and mime_type from formats table in a singular dataframe
+      format.info <- PEcAn.DB::db.query(
+        paste("SELECT name, mime_type FROM formats WHERE id = ", register$format$id), con
+      )
+
+      # Assign the values from the data frame to the respective columns in register$format
+      register$format$name <- format.info$name
+      register$format$mimetype <- format.info$mime_type
     } else if (is.null(register$format$id) & !is.null(register$format$name) & !is.null(register$format$mimetype)) {
       register$format$id <- PEcAn.DB::db.query(
         paste0("SELECT id from formats where name = '", register$format$name,
--- a/modules/data.atmosphere/R/read.register.R
+++ b/modules/data.atmosphere/R/read.register.R
@@ -36,14 +36,17 @@ read.register <- function(register.xml, con) {
                | 
                (!is.null(register$format$id) & is.null(register$format$mimetype))) {
       # Retrieve format name and mimetype from the database
-      register$format <- PEcAn.DB::db.query(
-        paste("SELECT name, mime_type AS mimetype FROM formats WHERE id = ", register$format$id), con
+      query.format.info <- PEcAn.DB::db.query(
+        paste("SELECT name, mimetype_id AS mimetype FROM formats WHERE id = ", register$format$id), con
       )
+      
+      register$format$name <- query.format.info$name
+      register$format$mimetype <- query.format.info$mimetype
 
     } else if (is.null(register$format$id) & !is.null(register$format$name) & !is.null(register$format$mimetype)) {
       register$format$id <- PEcAn.DB::db.query(
         paste0("SELECT id from formats where name = '", register$format$name,
-               "' and mime_type = '", register$format$mimetype, "'"), con)[[1]]
+               "' and mimetype_id = '", register$format$mimetype, "'"), con)[[1]]
     }
   }
   return(invisible(register))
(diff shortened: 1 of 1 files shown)


modules/data.atmosphere/R/read.register.R
```diff
@@ -35,10 +35,14 @@ read.register <- function(register.xml, con) {
...
-      register$format$mimetype <- PEcAn.DB::db.query(
-        paste("SELECT mime_type from formats where id = ", register$format$id), con)[[1]]
+      # get name and mime_type from formats table in a singular dataframe
+      format.info <- PEcAn.DB::db.query(
+        paste("SELECT name, mime_type FROM formats WHERE id = ", register$format$id), con
+      )
+
+      # Assign the values from the data frame to the respective columns in register$format
+      register$format$name <- format.info$name
+      register$format$mimetype <- format.info$mime_type
```
mdietze, line 45: Looks like this could be done in one line, register$format = format.info, if you do the name change in the query, mime_type as mimetype.

mdietze (approved):

modules/data.atmosphere/R/read.register.R
```diff
@@ -35,10 +35,11 @@ read.register <- function(register.xml, con) {
...
                | 
                (!is.null(register$format$id) & is.null(register$format$mimetype))) {
-      register$format$name <- PEcAn.DB::db.query(
-        paste("SELECT name from formats where id = ", register$format$id), con)[[1]]
-      register$format$mimetype <- PEcAn.DB::db.query(
-        paste("SELECT mime_type from formats where id = ", register$format$id), con)[[1]]
+      # Retrieve format name and mimetype from the database
+      register$format <- PEcAn.DB::db.query(
+        paste("SELECT name, mime_type AS mimetype FROM formats WHERE id = ", register$format$id), con
+      )
```
infotroph, line 41: Two problems here: 
1. Assigning the result to register$format will overwrite the existing format object (which contains at least `id` and `inputtype` -- see lines 25 and 35) with one that contains only `name` and `mimetype`. Better to query once into a temporary result and then assign from there into `register$format$name` and `register$format$mimetype.
2. The existing line 41 is broken because the formats table doesn't have a `mime_type` column, it has a `mimetype_id` tht then needs to be looked up in the `mimetypes` table. I think you want `"SELECT name, type_string AS mimetype FROM formats JOIN mimetypes ON formats.mimetype_id = mimetypes.id  WHERE formats.id ="`

Sweetdevil144: 1. I seemed to miss it. Fixing it now.
2. Ahh, I recall us having a conversation about this earlier and that seems to be the case.

modules/data.atmosphere/R/read.register.R
```diff
@@ -35,14 +35,18 @@ read.register <- function(register.xml, con) {
...
+      
+      register$format$name <- query.format.info$name
+      register$format$mimetype <- query.format.info$mimetype
+
     } else if (is.null(register$format$id) & !is.null(register$format$name) & !is.null(register$format$mimetype)) {
       register$format$id <- PEcAn.DB::db.query(
         paste0("SELECT id from formats where name = '", register$format$name,
-               "' and mime_type = '", register$format$mimetype, "'"), con)[[1]]
+               "' and mimetype_id = '", register$format$mimetype, "'"), con)[[1]]
```
Sweetdevil144, line 49: I changed this below occurrence too to keep in check the Point 2 you mentioned.

infotroph: This won't work because mimetype_id is expecting a bigint (e.g. `1093`) while register$format$mimetype will be a string (e.g. "application/x-netcdf").

modules/data.atmosphere/R/read.register.R
```diff
@@ -35,14 +35,18 @@ read.register <- function(register.xml, con) {
     } else if ((!is.null(register$format$id) & is.null(register$format$name)) 
                | 
                (!is.null(register$format$id) & is.null(register$format$mimetype))) {
-      register$format$name <- PEcAn.DB::db.query(
-        paste("SELECT name from formats where id = ", register$format$id), con)[[1]]
-      register$format$mimetype <- PEcAn.DB::db.query(
-        paste("SELECT mime_type from formats where id = ", register$format$id), con)[[1]]
+      # Retrieve format name and mimetype from the database
+      query.format.info <- PEcAn.DB::db.query(
+        paste("SELECT name, mimetype_id AS mimetype FROM formats WHERE id = ", register$format$id), con
```
infotroph, line 40:
```suggestion
        paste(
          "SELECT name, type_string AS mimetype",
          "FROM formats JOIN mimetypes ON formats.mimetype_id = mimetypes.id",
          "WHERE formats.id = ", register$format$id),
        con
```

modules/data.atmosphere/R/read.register.R
```diff
@@ -35,14 +35,18 @@ read.register <- function(register.xml, con) {
...
+      )
+      
+      register$format$name <- query.format.info$name
+      register$format$mimetype <- query.format.info$mimetype
+
     } else if (is.null(register$format$id) & !is.null(register$format$name) & !is.null(register$format$mimetype)) {
       register$format$id <- PEcAn.DB::db.query(
         paste0("SELECT id from formats where name = '", register$format$name,
```
infotroph, line 48:
```suggestion
        paste0(
        "SELECT id FROM formats JOIN mimetypes ON formats.mimetype_id = mimetypes.id ",
        "WHERE name = '", register$format$name,
```

modules/data.atmosphere/R/read.register.R
infotroph:
```suggestion
               "' AND type_string = '", register$format$mimetype, "'"), con)[[1]]
```

PR #3319, opened 2024-06-27
Refactor `met.process` and `dbfiles`

Sweetdevil144:
## Description

+ Optimized the standardization process by combining the two loops and improving efficiency.
+ Added `return` statement where we intended to return but didn't implement

## Motivation and Context

Removed the redundant loop that iterated over `standardize_result` to update `ready.id`. Instead, `ready.id` is now directly assigned the values from the current iteration's `standardize_result` if `is_standardized` is TRUE every time we perform a standardization.

For now, I've commented out the previous changes and will remove them once the current changes are approved.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


base/db/R/dbfiles.R
```diff
@@ -251,13 +251,13 @@ dbfile.input.check <- function(siteid, startdate = NULL, enddate = NULL, mimetyp
   formatid <- get.id(table = "formats", colnames = c("mimetype_id", "name"), values = c(mimetypeid, formatname), con = con)
 
   if (is.null(formatid)) {
-    invisible(data.frame())
+    return(invisible(data.frame()))
```
infotroph, line 254: It's an R style convention to omit the `return` unless returning early.

base/db/R/dbfiles.R
```diff
@@ -251,13 +251,13 @@ dbfile.input.check <- function(siteid, startdate = NULL, enddate = NULL, mimetyp
   formatid <- get.id(table = "formats", colnames = c("mimetype_id", "name"), values = c(mimetypeid, formatname), con = con)
 
   if (is.null(formatid)) {
-    invisible(data.frame())
+    (invisible(data.frame()))
```
infotroph, line 254: Sorry, my previous comment was misreading the code, and I missed that you were making it return early on purpose. Yes, please _do_ add `return` here.

Sweetdevil144: Adding it now !! :)

Sweetdevil144: I think `invisible` prevents the automatic printing of data rather than stopping execution of the external called function all together.

Would it be better to feasible early `return` statements in more such instance wherever we will want to urgently return early and not continue?

infotroph: Yep, exactly right. Anywhere you see an `invisible(...)` that's not assigned to anything there's a great chance it was intended to stop execution and should be wrapped to `return(invisible(...))`. Always worth considering if an early return is still the right behavior, though.
(and at much lower priority, it's also worth considering whether the return should really be invisible or not, but in most cases the only difference that makes is in how verbose the log output gets)

base/db/R/dbfiles.R
```diff
@@ -251,13 +251,13 @@ dbfile.input.check <- function(siteid, startdate = NULL, enddate = NULL, mimetyp
...
 
   # setup parent part of query if specified
-  if (is.na(parentid)) {
-    parent <- ""
-  } else {
+  parent <- ""
+  
+  if (!is.na(parentid)) {
     parent <- paste0(" AND parent_id=", parentid)
   }
```
infotroph, line 262: 👍 This same pattern appears in dbfile.input.insert above  too -- can you modify it there (near line 60) as well?

modules/data.atmosphere/R/met.process.R
```diff
@@ -314,15 +317,17 @@ met.process <- function(site, input_met, start_date, end_date, model,
                                        host = host, 
                                        overwrite = overwrite$standardize,
                                        ensemble_name = i)
+        is_standardized <- TRUE
+      } else {
+        is_standardized <- FALSE
+      }
+
+      if(is_standardized) {
```
infotroph, line 325: I think this changes behavior: previously `ready.id` would have an entry for each id whether it was standardized or not; this only adds the ids that were standardized. Is that intentional?

(As currently implemented I think ids not standardized would probably have errored out earlier, but this step should not rely on that)

Sweetdevil144: I've implemented a `NA` appending too to make sure to append even when we have face a `!is_standardized` condition. The initial though process I had was whether it would be wise pass the `input.id`'s to `.met2model.module` and later on to `convert_input` if we have not standardised our results. Although it seems I failed and ignored the order in array at check which may change the results.

Sweetdevil144: @infotroph Can you take a look into this?

base/db/R/dbfiles.R
```diff
@@ -648,9 +647,9 @@ dbfile.file <- function(type, id, con, hostname = PEcAn.remote::fqdn()) {
 
   if (nrow(files) > 1) {
     PEcAn.logger::logger.warn("multiple files found for", id, "returned; using the first one found")
-    invisible(file.path(files[1, "file_path"], files[1, "file_name"]))
+    (invisible(file.path(files[1, "file_path"], files[1, "file_name"])))
```
infotroph:
```suggestion
    return(invisible(file.path(files[1, "file_path"], files[1, "file_name"])))
```

base/db/R/dbfiles.R
```diff
@@ -648,9 +647,9 @@ dbfile.file <- function(type, id, con, hostname = PEcAn.remote::fqdn()) {
...
   } else if (nrow(files) == 1) {
-    invisible(file.path(files[1, "file_path"], files[1, "file_name"]))
+    (invisible(file.path(files[1, "file_path"], files[1, "file_name"])))
```
infotroph:
```suggestion
    return(invisible(file.path(files[1, "file_path"], files[1, "file_name"])))
```

base/db/R/dbfiles.R
```diff
@@ -648,9 +647,9 @@ dbfile.file <- function(type, id, con, hostname = PEcAn.remote::fqdn()) {
...
   } else {
     PEcAn.logger::logger.warn("no files found for ", id, "in database")
     invisible(NA)
```
infotroph:
```suggestion
    return(invisible(NA))
```

infotroph: For the record I _think_ all three lines above do work properly (if confusingly) before this change:
* only one leg of the if is evaluated, 
* so whichever invisible(...) is selected becomes the last evaluated expression 
* whose result the function returns by default. 
 
But the fact I had to write all that out to convince myself is exactly why the explicit returns are a good addition ;)

modules/data.atmosphere/R/met.process.R
infotroph: @Sweetdevil144 I was wrong above about the existing behavior -- In the current code `standardize_result` has an entry for each id, but it is NULL for ids not standardized and therefore gets silently dropped when appending to `ready.id`. I don't know how advisable that is, but I bet downstream code relies on it.

I'll suggest changes here that eliminate the second for-loop but keep the original behavior -- please edit further if you see problems I missed.

modules/data.atmosphere/R/met.process.R
```diff
@@ -284,7 +284,9 @@ met.process <- function(site, input_met, start_date, end_date, model,
   # Change to Site Level - Standardized Met (i.e. ready for conversion to model specific format)
   if (stage$standardize) {
     standardize_result <- list()
-    
+    is_standardized <- FALSE
+    ready.id <- list(input.id = NULL, dbfile.id = NULL)
+
```
infotroph:
```suggestion
    ready.id <- list(input.id = NULL, dbfile.id = NULL)

```

modules/data.atmosphere/R/met.process.R
```diff
@@ -284,7 +284,9 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
     for (i in seq_along(cf.id[[1]])) {
 
       if (register$scale == "regional") {
```
infotroph, line 263: GitHub only allows suggestions within 3 lines of an existing change, so commenting here: On line 294 below, replace `standardize_result[[i]] <-` with `id_stdized <-`

modules/data.atmosphere/R/met.process.R
```diff
@@ -303,6 +305,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
                                        host = host, 
                                        overwrite = overwrite$standardize)
                                        # Expand to support ensemble names in the future
+        is_standardized <- TRUE
```
infotroph:
```suggestion
```

modules/data.atmosphere/R/met.process.R
```diff
@@ -303,6 +305,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
       } else if (register$scale == "site") {
         ##### Site Level Processing
         standardize_result[[i]] <- .metgapfill.module(cf.id = list(input.id = cf.id$input.id[i], dbfile.id = cf.id$dbfile.id[i]),
```
infotroph:
```suggestion
        id_stdized <- .metgapfill.module(cf.id = list(input.id = cf.id$input.id[i], dbfile.id = cf.id$dbfile.id[i]), 
```

modules/data.atmosphere/R/met.process.R
```diff
@@ -316,15 +319,20 @@ met.process <- function(site, input_met, start_date, end_date, model,
                                        host = host, 
                                        overwrite = overwrite$standardize,
                                        ensemble_name = i)
+        is_standardized <- TRUE
+      } else {
+        is_standardized <- FALSE
```
infotroph:
```suggestion
      } else {
        # No action taken. These ids will be dropped from ready.id
        id_stdized <- NULL
```

modules/data.atmosphere/R/met.process.R
```diff
@@ -316,15 +319,20 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
+      }
+
+      if(is_standardized) {
+        ready.id$input.id <- c(ready.id$input.id, standardize_result[[i]]$input.id)
+        ready.id$dbfile.id <- c(ready.id$dbfile.id, standardize_result[[i]]$dbfile.id)
+      } else {
+         ready.id$input.id <- c(ready.id$input.id, NA)
+          ready.id$dbfile.id <- c(ready.id$dbfile.id, NA)
       }
```
infotroph:
```suggestion
      ready.id$input.id <- c(ready.id$input.id, id_stdized$input.id)
      ready.id$dbfile.id <- c(ready.id$dbfile.id, id_stdized$dbfile.id)
```

infotroph (changes requested):

modules/data.atmosphere/R/met.process.R
infotroph: It's possible this should be an error instead, with a message along the lines of "Only scales 'site' and 'regional' are currently supported". Thoughts?

modules/data.atmosphere/R/met.process.R
Sweetdevil144: Applied the changes

Sweetdevil144 (commented):
I think this PR is ready for review. Mentioning @infotroph and @moki1202 for approval

infotroph (approved):

PR #3320, opened 2024-06-27
Update CONTRIBUTING.md

dlebauer:
added instructions to update CHANGELOG, NEWS, CITATION files

do we have a threshold for authorship? currently 'non-trivial change'


mdietze (approved):
Threshold is hard, as pull request vary widely in their size and complexity. It's easy to point to examples where a single file of ~50 lines might be over a month's worth of work, and other cases where some quick and easy clean up might touch dozens of files and hundreds of lines of code, but only take a day.

PR #3321, opened 2024-06-29
#GSOC PR : Add Preprocess Function for Data Cleaning and Validation

sambhavnoobcoder:
## Description:

This PR introduces a new preprocess function designed to streamline the data cleaning and validation process. This function reads input data and site coordinates, validates the presence of specified date and carbon pool, and ensures the consistency of data dimensions. It outputs a structured list containing the cleaned data, ready for further analysis. Below is an extended description of the new function and its components.

### Purpose:
The preprocess function is created to read and validate input data and site coordinates, ensuring that the data is correctly formatted and consistent for further processing. It handles potential inconsistencies in the data, providing informative messages and adjustments where necessary.

### Parameters:
data_path: Path to the RDS file containing the input data. coords_path: Path to the CSV file containing site coordinates. date: The specific date for which the carbon data is to be extracted. C_pool: The specific carbon pool within the input data to focus on. Process:

### Reading Data:

Reads the input data from the provided RDS file.
Reads the site coordinates from the provided CSV file. 

### Validation:

Checks if the specified date exists in the input data. If not, the function stops and returns an error message. Extracts the carbon data for the specified date and validates the existence of the specified carbon pool. If the carbon pool is not found, the function stops and returns an error message. 

### Data Transformation:

Transposes the extracted carbon data to a data frame format, ensuring each column represents an ensemble. Renames the columns to a consistent naming convention (e.g., "ensemble1", "ensemble2", etc.). Coordinate Validation:

Ensures that the site coordinates data contains 'lon' and 'lat' columns. If these columns are missing, the function stops and returns an error message. 

### Data Consistency:

Validates that the number of rows in the site coordinates matches the number of rows in the carbon data. If there is a mismatch in the number of rows, the function truncates either the site coordinates or the carbon data to match the row counts, ensuring consistency.

### Output:
The function returns a list containing:

input_data: The original input data read from the RDS file. site_coordinates: The validated and possibly truncated site coordinates. carbon_data: The validated and possibly truncated carbon data. Messages:
The function provides informative messages during the preprocessing steps, alerting the user to any adjustments made to the data to ensure consistency.

### Example Usage:

```preprocessed_data <- preprocess("path/to/input_data.rds", "path/to/site_coords.csv", "2022-01-01", "TotalCarbon")```

### Benefits:

Efficiency: Streamlines the data preparation process, reducing manual validation and transformation steps.
Error Handling: Provides clear error messages and handles common data issues, improving robustness.
[...]

PR #3322, opened 2024-07-02
Downscale Function created for for hourly data in nc

Snafkin547:
## Description
This new function, called 'NA_downscale_hrly', is extended from the original 'NA_downscale' function.
This newly created downscale function handles hourly data stored in nc file and generates outputs including the weights of each ensemble member for each observatory site. These weights are not currently used, however, they will be used effectively in the aggregation step in the future, once we come up with an idea of how to distribute to each coordinate.

## Motivation and Context
In the previously developed 'NA_downscale' function, monthly-yearly data was handled. This newly developed downscale function enables to deal with the hourly data. Due to the difference in data structures and absence of weights in previous data structure, we needed this change.

## Review Time Estimate
- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


Snafkin547:
Example usage
```
# Set working directory
my_wd <- "<Your working directory>"
setwd(my_wd)

# Load necessary library
library(terra)
library(readr)
library(sf)
library(ncdf4)

nc_file <- "/Users/hishii/Desktop/PEcAn/osfstorage-archive/NCDF/NEE.nc"
nc_data <- nc_open(nc_file)
coords <- paste0(my_wd, "data/SDA/sites.csv")
covariates <- rast(paste0(my_wd, "data/covariates.tif"))
date <- "1986-01-01 00:00:00 EST"
downscale_output <- NA_downscale_hrly(nc_data, coords, date, covariates)

```

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,96 @@
+#' @title North America Downscale Function
+#' @name NA_downscale_hrly
```
mdietze: Note that in PR #3321 we change the name of the original downscale function to SDA_downscale because there's nothing North America specific about this code.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,96 @@
...
+#'
+#' @description This function uses the randomForest model.
+#' @return It returns the `downscale_output` list containing lists for the training and testing data sets, models, and predicted maps for each ensemble member.
+#' @import ncdf4
+#' @export
+
+NA_downscale_hrly <- function(nc_data, coords, date, covariates){
+  
+  # Read the input data and site coordinates
+  input_data <- ncvar_get(nc_data, "NEE")
```
mdietze: Wanted to check on inputs. The function says it's a file name, but ncvar_get requires an opened connection to a nc file, not a file name. Is the code wrong or the function documentation? FWIW I suspect you'll want to pass in an `nc_open()`ed nc object, not a nc file name.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,96 @@
...
+  weights_rrel <- ncvar_get(nc_data, "weights_rrel")
+  
+  # Timereadable
+  time <- nc_data$dim$time$vals
+  time_units <- nc_data$dim$time$units
+  time_origin <- as.POSIXct(substr(time_units, 12, 31), format="%Y-%m-%dT%H:%M")
+  time_readable <- time_origin + time * 3600  # Convert hours to seconds
```
mdietze: Just want to check on the logic here. Two lines up extracts the time units, but then this line does a time unit conversion without first checking the units. More generally, I'd recommend using more formal datetime functions (e.g. using `lubridate`) over hard-coding conversions.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,96 @@
...
+  
+  # Extract predictors from covariates raster using site coordinates
+  site_coordinates <- terra::vect(readr::read_csv(coords), geom=c("lon", "lat"), crs="EPSG:4326")
+  index <- which(time_readable == date)
+  data <- input_data[index, , ]
```
mdietze: As written, this code seems to downscale a single timeslice, not a timeseries. Furthermore, it doesn't seem to be set up to efficiently allow one to apply the function over every timepoint in a timeseries, which would involve a lot of redundant opening of files, copying of data objects, etc.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,96 @@
...
+  }
+  
+  # Rename the carbon_data column for each ensemble member
+  for (i in 1:length(ensembles)) {
+    ensembles[[i]] <- dplyr::rename(ensembles[[i]], "carbon_data" = "carbon_data[[i]]")
+  }
+  
+  # Split the observations in each data frame into two data frames based on the proportion of 3/4
+  ensembles <- lapply(ensembles, function(df) {
+    sample <- sample(1:nrow(df), size = round(0.75*nrow(df)))
```
mdietze: Note: if you're downscaling a timeseries, to me it seems to make more sense to split the data once, not to generate different splits for every time point in the timeseries.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,96 @@
...
+  # Rename the training and testing data frames for each ensemble member
+  for (i in 1:length(ensembles)) {
+    # names(ensembles) <- paste0("ensemble",seq(1:length(ensembles)))
+    names(ensembles[[i]]) <- c("training", "testing")
+  }
+  
+  # Train a random forest model for each ensemble member using the training data
+  rf_output <- list()
+  for (i in 1:length(ensembles)) {
+    rf_output[[i]] <- randomForest::randomForest(ensembles[[i]][[1]][["carbon_data"]] ~ land_cover+tavg+prec+srad+vapr+nitrogen+phh2o+soc+sand,
```
mdietze: note improvements in PR #3321 that avoids hard-coding the covariate list
```
 formula = stats::as.formula(paste(ensemble_col, "~", paste(covariate_names, collapse = " + ")))
```

mdietze (changes requested):
Overall, function is a good first start but doesn't really address the need to downscale timeseries, but instead provides a way to downscale a single time point. I'm happy to pull in this PR as an initial milestone without that generalization once the other smaller points are addressed.

modules/assim.sequential/R/downscale_function_hrly.R
infotroph: Chiming in with a Roxygen style point: I recommend not including a `@name` line at all -- if you omit it, Roxygen will default to using the function name, which is usually what you want and when they differ it can be confusing. 

Relatedly I recommend leaving the `@title`, `@description`, and `@details` tags implicit: Write the introductory block with no tags, then Roxygen will [take these](https://roxygen2.r-lib.org/articles/rd.html#the-introduction) from the first, second, and third-and-later paragraphs respectively.  In most cases typing the tags out doesn't hurt anything, but can be confusing when they're specified in a different order. We have some legacy functions where I keep looking for the `@details` and being surprised to find them below the parameter descriptions.

None of this changes Mike's point about changing the name of the function itself, just saying it could be changed in one line of this file instead of two.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -9,6 +9,7 @@
 #' @param covariates SpatRaster stack, used as predictors in randomForest. Layers within stack should be named. Recommended that this stack be generated using 'covariates' instructions in assim.sequential/inst folder
 #' @return It returns the `downscale_output` list containing lists for the training and testing data sets, models, and predicted maps for each ensemble member.
 #' @import ncdf4
+#' @import lubridate
```
mdietze: Code below looks great. But note that our preference would be to use function namespaces (e.g. ncdf4::nc_open, lubridate::ymd_hm) over `@import`

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -20,8 +21,16 @@ SDA_downscale_hrly <- function(nc_file, coords, date, covariates){
   # Timereadable
   time <- nc_data$dim$time$vals
   time_units <- nc_data$dim$time$units
-  time_origin <- as.POSIXct(substr(time_units, 12, 31), format="%Y-%m-%dT%H:%M")
-  time_readable <- time_origin + time * 3600  # Convert hours to seconds
+  time_origin_str <- substr(time_units, 12, 31)
+  time_origin <- ymd_hm(time_origin_str, tz="EST")
```
mdietze: Check timezone. If not specified in a file, standard would be to use UTC/GMT not EST

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,108 @@
...
+  downscale_output<- list()
+  
+  # Train & Test split
+  sample <- sample(1:nrow(predictors), size = round(0.75*nrow(predictors)))
+  
+  # Predict for each time stamp of the year selected
+  time_indices <- which(year(time_readable) == yyyy)
+  for (index in time_indices) {
+    if(index == 37986){
+      break
```
mdietze: This feels like it's a bit of a hack. Could you explain why this bit is needed and where this specific index number comes from?

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,108 @@
...
+    full_data <- cbind(carbon_data, predictors)
+    train_data <- full_data[sample, ]
+    test_data <- full_data[-sample, ]
+    
+    # Combine each ensemble member with all predictors
+    models <- list()
+    maps <- list()
+    predictions <- list()
+    ensembles <- list()
+    for (i in seq_along(carbon_data)) {
```
mdietze, line 71: Doesn't need any change for this PR, but making a note for those doing future development that this would be a great place to either parallelize this work or to modify the ML model to take time into account explicitly.

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,108 @@
...
+    for (i in 1:length(curr_downscaled$data)) {
+      names(curr_downscaled$data[[i]]) <- paste0("ensemble", seq(1:ncol(carbon_data)))
+    }
+    names(curr_downscaled$models) <- paste0("ensemble", seq(1:ncol(carbon_data)))
+    names(curr_downscaled$maps) <- paste0("ensemble", seq(1:ncol(carbon_data)))
+    names(curr_downscaled$predictions) <- paste0("ensemble", seq(1:ncol(carbon_data)))
+    
+    downscale_output[[as.character(time_readable[index])]]<-curr_downscaled
+  }
+  nc_close(nc_data)
```
mdietze: One thing we find helpful is to embed the nc_close in an `on.exit()` and to put it right after the nc_open, which ensures the file is closed in the event the function exits prematurely.

modules/assim.sequential/R/downscale_function_hrly.R
Snafkin547: Sorry for my oversight. I used this control flow for debugging. Removed it in the latest commit

modules/assim.sequential/R/downscale_function_hrly.R
Snafkin547: Moved this after the nc_open: `on.exit(ncdf4::nc_close(nc_data))`

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,103 @@
...
+#' @param covariates SpatRaster stack, used as predictors in randomForest. Layers within stack should be named. Recommended that this stack be generated using 'covariates' instructions in assim.sequential/inst folder
+#' @return It returns the `downscale_output` list containing lists for the training and testing data sets, models, and predicted maps for each ensemble member.
+#' @export
+
+SDA_downscale_hrly <- function(nc_file, coords, yyyy, covariates){
+  
+  # Read the input data and site coordinates
+  nc_data <- ncdf4::nc_open(nc_file)
+  on.exit(ncdf4::nc_close(nc_data))
+  input_data <- ncvar_get(nc_data, "NEE")
```
mdietze: Needs namespace

modules/assim.sequential/R/downscale_function_hrly.R
```diff
@@ -0,0 +1,103 @@
...
+    time_origin <- lubridate::ymd_hm(time_origin_str, tz = "UTC")
+  } else if (grepl("EST", time_units)) {
+    time_origin <- lubridate::ymd_hm(time_origin_str, tz = "EST")
+  } else {
+    time_origin <- lubridate::ymd_hm(time_origin_str, tz = "UTC")  # Default to UTC if not specified
+  }
+  
+  # Timereadable
+  if (grepl("hours", time_units)) {
+    time_readable <- time_origin + dhours(time)
```
mdietze: dhours and dseconds need namespace (unless they are base functions)

mdietze (approved):

infotroph:
Hi, @Snafkin547! With this PR you became an author of the PEcAn.assim.sequential package, so we'd like to make sure we have your agreement to change the package license from NCSA to BSD-3. Commenting "yes" here is enough. More details are in #3364 if you want them.

We started the relicensing process (with notes in CHANGELOG.md) before you opened this PR, but at merge time we hadn't yet updated any license files in this package, so for full clarity we'd like to hear your explicit consent even if you had been assuming your contribution would be BSD-3 all along 😁

PR #3324, opened 2024-07-05
Add function to Optionally get `site.info` if not present

Sweetdevil144:
## Description

This PR aims to refactor site.id and improve str_ns. The end goal is to make the the System independent of DB. Currently, I'm refactoring to create a siteID if not present already. I'll add test cases to check this util function too.
Work still in pending

Some comments from @mdietze :
> 1. Here's my slightly different interpretation :
> 2. Require the user to input a dataframe with lat/lon and optionally siteID; if siteID is not provided, >construct a unique identifier from lat/lon
> 3. replace str_ns with provided unique identifier
> 4. Provide a helper function that takes in a BETY connection and site IDs and returns lat, lon, and str_ns

## Motivation and Context

May Fix a Subtask of #3307 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


Sweetdevil144:
Seems the tests fail as the package is not published with with my Updated function. Should I utilize `source` to Update Imports and later on turn it to `PEcAn.remote::` before it's published?

Sweetdevil144:
@meetagrawal09 can you cross check if corresponding changes in `test.met.process` are valid or not? :)

modules/data.atmosphere/tests/testthat/test.met.process.R
```diff
@@ -8,7 +8,7 @@ test_that("`met.process` able to call .download.raw.met.module based on met proc
   mockery::stub(met.process, 'PEcAn.DB::query.format.vars', list())
   mockery::stub(met.process, 'PEcAn.DB::dbfile.check', list(id = 1))
   mockery::stub(met.process, 'assign', 1)
-  mockery::stub(met.process, 'PEcAn.DB::query.site', list(lat = 0, lon = 0))
+  mockery::stub(met.process, 'get.new.site', list(new.site = data.frame(id = 1, lat = 0, lon = 0), str_ns = "0-0"))
```
meetagrawal09, line 11: Update this to `PEcAn.DB::get.new.site` rather than just `get.new.site`.

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,53 @@
+#-------------------------------------------------------------------------------
+# Copyright (c) 2012 University of Illinois, NCSA.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the
+# University of Illinois/NCSA Open Source License
+# which accompanies this distribution, and is available at
+# http://opensource.ncsa.illinois.edu/license.html
+#-------------------------------------------------------------------------------
+
+##' Get new site info using provided site information
```
infotroph, line 10: Can you explain why this function is needed? Using the database to generate new IDs isn't going to help with the goal of enabling runs of `do_conversions` when no database is present at all.

base/workflow/R/do_conversions.R
```diff
@@ -9,6 +9,10 @@
 ##' @author Ryan Kelly, Rob Kooper, Betsy Cowdery, Istem Fer
 
 do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALSE, overwrite.ic = FALSE) {
+
+  # extract site info from settings to be utilizes later in processing.
+  site.info <- settings$run$site
+
```
infotroph, line 15: Better to use it directly from inside the existing `settings` object

infotroph:
I think you're taking the wrong approach to task 2 here: Yes, "if siteID is not provided, construct a unique identifier from lat/lon", but it needs to be constructed without using the DB. As we discussed in Slack, this could be as simple as something like `id_str <- paste0(lat, "_", lon)`.

base/db/R/get.new.site.R
Sweetdevil144: This version of file handles only when con != NULL.
I've updated the code to handle db-less ID handling. Can you check it out once?

base/workflow/R/do_conversions.R
Sweetdevil144: I'd been experimenting to see if we have any instance which can be directly utilized with `site` rather than querying it. Although now I've shifted it to a new function, it's nice to use it from within. Thanks for suggestion

Sweetdevil144:
Current code is updated for db-less interaction. What more improvements can be made? Would creating a new site-id be correct for a pre existing `lat` and `lon` or should we consider fetching it from `db` if connection is present. Else-if `con=NULL`, we can generate a new siteID. We can create a `int+str` (For eg : `dcf544e6c` etc) styled site-id to distinguish this `str_ns` from those `site-id`'s when `con!=NULL`.

Sweetdevil144:
Should we add a unit test to check if `get.new.site()` works correctly?

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,153 @@
+#-------------------------------------------------------------------------------
+# Copyright (c) 2012 University of Illinois, NCSA.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the
+# University of Illinois/NCSA Open Source License
+# which accompanies this distribution, and is available at
+# http://opensource.ncsa.illinois.edu/license.html
+#-------------------------------------------------------------------------------
```
infotroph:
```suggestion
```

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,153 @@
...
+##' get.new.site(site=data.frame(id=1,lat=40.1,lon=-88.2,time_zone="UTC"),con=NULL,latlon=NULL)
+
+get.new.site <- function(site, con = NULL, latlon = NULL) {
+    if (is.null(con)) {
+        PEcAn.logger::logger.debug("DB connection is closed. Trying to generate a new site ID or use pre-existing one.")
+        # No DB connection present. Generate a new ID using one of below steps:
+
+        if (is.null(site$id) | is.na(site$id)) {
+            if ((!is.null(site$lat) && !is.null(site$lon)) |
+                (!is.na(site$lat) && !is.na(site$lon))
```
infotroph, line 21: Weird little gotcha here -- the !is.null will fire when lat and lon are NA, and the !is.na will return a missing value when lat and lon are null.

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,145 @@
...
+            ) {
+                site.id <- generate_siteID(site$lat, site$lon)
+                new.site <- data.frame(
+                    id = as.numeric(site.id),
+                    lat = site$lat,
+                    lon = site$lon
+                )
+                str_ns <- paste0(new.site$lat, "-", new.site$lon)
```
infotroph: Might want to use a string other than "-" to concatenate here, to avoid confusion between separator characters and negative lon values

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,145 @@
...
+                    id = as.numeric(site$id),
+                    lat = latlon$lat,
+                    lon = latlon$lon
+                )
+                str_ns <- paste0(new.site$lat, "-", new.site$lon)
+            }
+        }
+    }
+
+    site.info <- list(new.site = new.site, str_ns = str_ns)
```
infotroph, line 101: Q in live review: why this structure? A: to match what's expected at met.process line 165

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,145 @@
...
+    # Return the siteID
+    return(siteID)
+}
+
+# Function to generate a new siteID (db-less runs ONLY)
+generate_new_siteID <- function() {
+    # Generate a random number. Assuming higher order integers to increase randomness in IDs
+    random_id <- sample(10000:99999999, 1)
+    return(as.numeric(random_id))
+}
```
infotroph: We want IDs to be deterministic -- don't need this

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,145 @@
...
+    # Scale normalized values (Assuming 4 digits each for lat and lon)
+    scaled_lat <- as.integer((norm_lat / 180) * 9999) # Scaled latitude
+    scaled_lon <- as.integer((norm_lon / 360) * 9999) # Scaled longitude
+
+    # Create a unique ID by concatenating scaled values
+    siteID <- as.integer(paste0(scaled_lat, scaled_lon))
+
+    # Return the siteID
+    return(siteID)
+}
```
infotroph: replace with paste0(lat,lon) or slight variant -- it's OK (at least in the met processing workflow) for IDs to be strings that contain minus signs.

base/workflow/R/do_conversions.R
```diff
@@ -52,7 +53,7 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     # IC conversion : for now for ED only, hence the css/pss/site check
     # <useic>TRUE</useic>
     if (ic.flag) {
-      settings <- PEcAn.data.land::ic_process(settings, input, dir = dbfiles, overwrite  = overwrite.ic)
+      settings <- PEcAn.data.land::ic_process(settings, input, dir = dbfiles, overwrite  = overwrite.ic, site = settings$run$site)
```
infotroph: Since site is already in settings, no need to pass as a new argument

modules/data.atmosphere/R/met.process.R
```diff
@@ -161,11 +161,12 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
-  latlon <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")] 
-  new.site <- data.frame(id = as.numeric(site$id), 
-                         lat = latlon$lat, 
-                         lon = latlon$lon)
-  str_ns <- paste0(new.site$id %/% 1e+09, "-", new.site$id %% 1e+09)
+  latlon <- NULL
+  site.info <- PEcAn.DB::get.new.site(site, con=con, latlon = latlon)
+
+  # extract new.site and str_ns from site.info
+  new.site <- site.info$new.site
```
infotroph, line 148: Hypothesis to confirm before acting: No downstream object uses new.site$id, can remove it here

modules/data.land/R/ic_process.R
```diff
@@ -6,14 +6,14 @@
 ##' @param input Taken from settings$run$inputs. This should include id, path, and source
 ##' @param dir settings$database$dbfiles
 ##' @param overwrite Default = FALSE. whether to force ic_process to proceed
+##' @param site Current site information
 ##'
 ##' @author Istem Fer, Hamze Dokoohaki
-ic_process <- function(settings, input, dir, overwrite = FALSE){
+ic_process <- function(settings, input, dir, overwrite = FALSE, site = settings$run$site){
```
infotroph: don't need this new argument, can just keep line 16 below

modules/data.land/R/soil_process.R
```diff
@@ -28,13 +28,14 @@ soil_process <- function(settings, input, dbfiles, overwrite = FALSE,run.local=T
...
-
-  str_ns <- paste0(new.site$id %/% 1e+09, "-", new.site$id %% 1e+09)
+  
+  # setup site database number, lat, lon and name and copy for format.vars if new input
+  latlon <- NULL
+  site.info <- PEcAn.DB::get.new.site(site, con=con, latlon = latlon)
+
+  # extract new.site and str_ns from site.info
+  new.site <- site.info$new.site
+  str_ns <- site.info$str_ns
```
infotroph: I need to think more, but I think we could leave this as it is to be refactored separately -- soil_process already assumes DB is available for other reasons

infotroph (changes requested):
Posting my notes from our live review -- they're a bit cryptic, but hopefully enough to remind you of the key points from the conversation. 

Basically I think the entire get.new.site function is trying to handle too many cases, and instead when the db isn't available we should simply paste together lat and lon to use as a siteid.

base/db/R/get.new.site.R
infotroph: You probably want `&&` instead of `|`

modules/benchmark/DESCRIPTION
```diff
@@ -40,8 +40,7 @@ Imports:
     zoo 
 Suggests:
     PEcAn.data.land,
-    testthat (>= 2.0.0),
-    BrownDog
+    testthat (>= 2.0.0)
```
infotroph: This should be in #3348 instead of this branch

modules/data.atmosphere/R/met.process.R
Sweetdevil144: `new.site$id` is later on passed on to `download.raw.met.module` for further conversions in `convert_input` function. I guess for that instance we will have to keep it rather than duplicating the code.

modules/data.land/R/soil_process.R
Sweetdevil144: Refactoring the same!!

base/db/R/get.new.site.R
Sweetdevil144: Can you suggest a method to perform so? I am unable to determine one right now :) !!

base/db/R/get.new.site.R
Sweetdevil144: Detailed Explanation : Current structure is followed to match data flow of what information is later on utilised in `met.process` for other function calls

base/db/R/get.new.site.R
Sweetdevil144: Replaced current logic with a "_" sign

base/db/R/get.new.site.R
Sweetdevil144: One approach I suggest is as follow : Hashing to generate a unique id using lat and lon of the site. This would ensure greater precision upto 8 decimal points.

Sweetdevil144 (commented):
From previous comments on #3348, I couldn't resist removing unnecessary changes. I will post a discussion regarding this PR and gather insights from other 'members & contributors' too.

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,123 @@
+##' Get new site info using provided site information
+##'
+##' @title Get New Site Info
+##' @param site a dataframe with site information including id, lat, lon, and time_zone.
+##' @param con Database connection object.
+##' @param latlon Optional global latlon object which will store updated lat/lon.
```
mdietze: I'm not following why latlon exists as an option here, as information is never extracted from it in any of the cases below within this function, and none of the calls you implemented to this function pass in anything other than NULL.

modules/data.atmosphere/R/met.process.R
```diff
@@ -133,11 +133,12 @@ met.process <- function(site, input_met, start_date, end_date, model,
   
   
   # setup site database number, lat, lon and name and copy for format.vars if new input
-  latlon <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")] 
-  new.site <- data.frame(id = as.numeric(site$id), 
-                         lat = latlon$lat, 
-                         lon = latlon$lon)
-  str_ns <- paste0(new.site$id %/% 1e+09, "-", new.site$id %% 1e+09)
+  latlon <- NULL
```
mdietze: Why is latlon defined to be null here? Why not just delete this line and the `, latlon = latlon` in the function call as you already set the default for latlon to NULL.

modules/data.land/R/ic_process.R
```diff
@@ -49,15 +50,25 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
   
-  #grab site lat and lon info
-  latlon <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")] 
   # setup site database number, lat, lon and name and copy for format.vars if new input
-  new.site <- data.frame(id = as.numeric(site$id),
-                         lat = latlon$lat,
-                         lon = latlon$lon)
+  latlon <- NULL
+  if(is.null(site$lat) | is.null(site$lon)) {
+    site.info <- PEcAn.DB::get.new.site(site, con=con, latlon = latlon)
```
mdietze: Same point as earlier about latlon not needing to be defined here.

modules/data.land/R/ic_process.R
```diff
@@ -49,15 +50,25 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
+    # extract new.site and str_ns from site.info
+    new.site <- site.info$new.site
+    str_ns <- site.info$str_ns
+  } else {
+    latlon <- list(lon = site$lat, lon=site$lon)
+    new.site <- data.frame(
+      id = as.numeric(site$id),
+      lat = site$lat,
+      lon = site$lon
+    )
```
mdietze: I'm not following why this if/else exists. Doesn't get.new.site already support this `else` case? And if not, shouldn't it?

modules/data.land/R/soil_process.R
```diff
@@ -30,6 +30,7 @@ soil_process <- function(settings, input, dbfiles, overwrite = FALSE,run.local=T
   # set up bety connection
   con <- PEcAn.DB::db.open(dbparms$bety)
   on.exit(PEcAn.DB::db.close(con), add = TRUE)
+  
   # get site info
   latlon <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")]
```
mdietze: Not following why this function only has whitespace change as it seems like the code chunk here is doing the same thing as what you were trying to support with your new function, and also would benifit from having a database-independent option.

mdietze (changes requested):

base/db/R/get.new.site.R
dlebauer: I've implemented hasing in #3423 thus:

```r
mutate(id = digest::digest(geometry, algo = "xxhash64"))
```

where 'geometry' could be replaced by `paste0(lat, lon)` or, if we wanted to be consistent and convert it to an sf object (I'm not certain I see the point) but could do 

```r
data.frame(lat=100, lon = -90) |> 
  sf::st_as_sf(coords = c("lon", "lat"), crs = 4326) |> 
  digest::digest(algo="xxhash64")
```

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    site.info <- list(new.site = new.site, str_ns = str_ns)
+
+    return(site.info)
+}
+
+
+# Function to generate a new siteID using hashing (db-less runs ONLY)
+generate_new_siteID <- function(lat, lon) {
```
dlebauer: I think this is a useful function to have - and would like to use it. 

I've implemented something similar as:

```
mutate(id = digest::digest(geometry, algo = "xxhash64"))
```

where 'geometry' could be replaced by paste0(lat, lon) or, if we wanted to be consistent and convert it to an sf object (I'm not certain I see the point) but could do

```
data.frame(lat=lat, lon = lon) |> 
  sf::st_as_sf(coords = c("lon", "lat"), crs = 4326) |> 
  digest::digest(algo="xxhash64")
```

I think both approaches have the same goal - I'd go with whichever one has a lower probability of generating the same id for different locations (unless both are trivially small like <<10^-9). Also, I'd suggest the function be documented, exported, and named generate_site_id (following db field name convention)

base/db/R/get.new.site.R
Sweetdevil144: `latlon` may exist within `site` as `site$lat` and `site$lon`. In case user had previously extracted this data within `latlon`, then passing `latlon` would seem like an ideal approach. Else the functions extracts required site information from `site` info itself.

Should I delete the `latlon` variable and try to extract information from `site` info itself everytime?

base/db/R/get.new.site.R
Sweetdevil144: I've documented the function & exported as well. I'd like to go with below functional convention because this seems like the simplest thing to do for now :)
Although if you suggest I can change it to use the latter implementation.
```r
  latlon_str <- paste0(lat, lon)
  uid <- digest::digest(latlon_str, algo = "xxhash64")
```

base/db/R/get.new.site.R
dlebauer:
```suggestion
```

Sweetdevil144:
Fixing the CI checks and test errors

base/db/R/get.new.site.R
dlebauer: Removed `latlon` because it is never used in the function. It was created later whether or not it existed, but not returned. https://github.com/PecanProject/pecan/pull/3324/commits/8132ae69e4d40482c969e1e5f6e07ce9371ed13e

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,121 @@
+##' Get new site info using provided site information
+##'
+##' @title Get New Site Info
+##' @param site a dataframe with site information including id, lat, lon, and time_zone.
+##' @param con Database connection object.
+##' @return a dataframe with new site information on lat, lon and time_zone
+##' @export
+##' @author Abhinav Pandey
+##' @importFrom digest digest
```
mdietze, line 9: What does this do? First, as far as I can tell this function is never used here. Second, where it is used I don't see why it can't be called by namespace

base/db/NAMESPACE
```diff
@@ -63,6 +65,7 @@ export(try2sqlite)
 export(var_names_all)
 export(workflow)
 export(workflows)
+importFrom(digest,digest)
```
mdietze, line 68: Should not be needed

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,121 @@
...
+#' @param lon Longitude value
+#' @return A unique string ID generated by hashing the lat/lon values
+#' @details This function is intended for database-less runs only by generating
+#'          a unique site ID by hashing the rounded lat/lon coordinates.
+#'
+#' @author Abhinav Pandey
+#' @export
+generate_site_id <- function(lat, lon) {
+  latlon_str <- paste0(lat, lon)
+  uid <- digest::digest(latlon_str, algo = "xxhash64")
```
mdietze, line 119: Do we really need to add an entirely new package dependency to PEcAn just for this one call in one function? Is there an easier solution, or one that relies on existing packages?

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,121 @@
...
+        PEcAn.logger::logger.debug("DB connection is closed. Trying to generate a new site ID or use pre-existing one.")
+        # No DB connection present. Generate a new ID using one of below steps:
+
+        if (is.null(site$id) | is.na(site$id)) {
+            if ((!is.null(site$lat) && !is.null(site$lon)) &&
+                (!is.na(site$lat) && !is.na(site$lon))
+            ) {
+                site.id <- paste0(site$lat, "_", site$lon)
+                new.site <- data.frame(
+                    id = as.numeric(site.id),
```
mdietze, line 25: the as.numeric here seems like it's going to always return NA given how site.id is constructed

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,121 @@
...
+                    lat = site$lat,
+                    lon = site$lon
+                )
+                str_ns <- paste0(new.site$lat, "_", new.site$lon)
```
mdietze, line 29: this seems redundant with site.id

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,121 @@
...
+            } # lat/lon present but ID is absent
+            # Use Pre-existing normalised lat/lon
+            else {
+                PEcAn.logger::logger.warn("Site dataframe does not have an id column")
+                site.id <- generate_site_id()
```
mdietze, line 34: This function has two non-optional arguments that aren't being specified, so how is this code not throwing an error when you run it?

base/db/R/get.new.site.R
```diff
@@ -0,0 +1,121 @@
...
+        # Check if site dataframe has an id column
+        if (is.null(site$id) | is.na(site$id)) {
+            PEcAn.logger::logger.warn("Site dataframe does not have an id column. Generating a unique ID")
+            if ((!is.null(site$lat) && !is.null(site$lon)) &&
+                (!is.na(site$lat) && !is.na(site$lon))
+            ) {
+                PEcAn.logger::logger.info(paste0("Generated siteID using lat and lon:", site.id))
+                site.id <- generate_site_id(site$lat, site$lon)
+                new.site <- data.frame(
+                    id = as.numeric(site.id),
```
mdietze, line 78: does this id need to be numeric? If so, why? Is the algorithm generating it guaranteed to produce numbers?

modules/data.atmosphere/R/met.process.R
```diff
@@ -142,11 +142,11 @@ met.process <- function(site, input_met, start_date, end_date, model,
   
   
   # setup site database number, lat, lon and name and copy for format.vars if new input
-  latlon <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")] 
-  new.site <- data.frame(id = as.numeric(site$id), 
-                         lat = latlon$lat, 
-                         lon = latlon$lon)
-  str_ns <- paste0(new.site$id %/% 1e+09, "-", new.site$id %% 1e+09)
+  site.info <- PEcAn.DB::get.new.site(site, con=con, latlon = latlon)
```
mdietze, line 145: latlon undefined

modules/data.land/R/ic_process.R
```diff
@@ -49,21 +50,16 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
   
-  #grab site lat and lon info
-  latlon <- PEcAn.DB::query.site(site$id, con = con)[c("lat", "lon")] 
-  # setup site database number, lat, lon and name and copy for format.vars if new input
-  new.site <- data.frame(id = as.numeric(site$id),
-                         lat = latlon$lat,
-                         lon = latlon$lon)
+  # Setup site database number, lat, lon and name and copy for format.vars if new input
+  # Then extract new.site and str_ns from site.info 
+  site.info <- PEcAn.DB::get.new.site(site, con=con, latlon = latlon)
```
mdietze, line 55: latlon undefined

modules/data.land/R/ic_process.R
```diff
@@ -6,14 +6,15 @@
 ##' @param input Taken from settings$run$inputs. This should include id, path, and source
 ##' @param dir settings$database$dbfiles
 ##' @param overwrite Default = FALSE. whether to force ic_process to proceed
+##' @param site Current site information
```
mdietze, line 9: This line is defining a new argument to the function, but the argument has't been added to the function itself

base/db/DESCRIPTION
```diff
@@ -71,7 +72,8 @@ Suggests:
     rmarkdown (>= 2.19),
     testthat (>= 2.0.0),
     tidyverse,
-    withr
+    withr,
+    openssl
```
mdietze, line 76: Package added as a dependency, but I can't find any places in your PR where it's actually used. Can this be removed?

mdietze (changes requested):
What are the examples you are using to test your code as part of development? I ask because (1) there are no testthat tests defined for your new functions and (2) there's a good bit of code here that doesn't look like it would run at all. I want to make sure your not just writing code that "makes sense" and submitting it without verifying that the individual functions  work as expected under all the different cases and that the overall workflow still works

base/db/R/get.new.site.R
dlebauer: Doesn't necessarily warrant using it, but that was my suggestion since I was using it in ccmmf workflows, and it was already a dependency b/c of the api. https://github.com/PecanProject/pecan/blob/7faa280e4eaeecdd2370d3fd5abd14180cc1fafb/apps/api/R/auth.R#L14

base/db/R/get.new.site.R
dlebauer: It won't be numeric if using digest as is.

base/db/R/get.new.site.R
infotroph: There's also `rlang::hash`

infotroph:
@Sweetdevil144 Thank you for the extensive work you put into this PR. It has become quite large compared to the simple idea it started from, and at the same time we've been gaining more experience running PEcAn with no database and learning how much we can simplify things by making more inputs truly required rather than trying to fill them in for you. Specifically, it's now clear that we almost always want to force the user to provide a site id rather than try to create it for them, and that in the cases we create an ID on the fly we should generally be thinking of it as a site name (i.e. something that lets a human figure out which site's in this folder) rather than guarantee its permanence or uniqueness as a global identifier.

I've taken the liberty of cherry-picking commits from this PR into a much smaller set of changes in #3656, which will close this one. I think the replacement PR captures the core ideas of this one (avoiding DB access when we know location already, allowing met download from locations with no preassigned siteid) with a much smaller footprint and better alignment with our current run setup approaches.

Sweetdevil144:
Let's close this Pr once #3656 is merged

PR #3326, opened 2024-07-06
Dev intro tweaks

infotroph:
## Description

Updating some versions and clarifying some points that have confused me repeatedly. I believe this also addresses most of @dlebauer's comments in https://github.com/PecanProject/pecan/pull/3311#pullrequestreview-2156700604

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


DEV-INTRO.md
```diff
@@ -78,70 +72,57 @@ The steps in this section only need to be done the first time you start working
...
-echo "COMPOSE_PROJECT_NAME=pecan" >> .env
-echo "PECAN_VERSION=develop" >> .env
-echo "UID=$(id -u)" >> .env
-echo "GID=$(id -g)" >> .env
+COMPOSE_PROJECT_NAME=pecan
+PECAN_VERSION=develop
 ```
 
-For MacOSX
+If you have a Linux system you will need to set UID and GID (these are needed by rstudio when sharing files between host and container):
```
infotroph, line 96: Is this actually Linux-specific? I have UID and GID set on my Mac and have not been able to get it working without. 
Should we recommend this step for every OS?

DEV-INTRO.md
```diff
@@ -78,70 +72,57 @@ The steps in this section only need to be done the first time you start working
...
 
 ```sh
-echo "COMPOSE_PROJECT_NAME=pecan" >> .env
-echo "PECAN_VERSION=develop" >> .env
+echo "UID=$(id -u)" >> .env
+echo "GID=$(id -g)" >> .env
```
infotroph, line 100: Is this actually Linux-specific? I have UID and GID set on my Mac and have not been able to get it working without.
Should we recommend this step for every OS?

DEV-INTRO.md
```diff
@@ -78,70 +72,57 @@ The steps in this section only need to be done the first time you start working
...
 
 If you have uncommented the volumes in `docker-compose.override.yml` you will need to create the folders. Assuming you have not modified the values, you can do this with:
 
 ```sh
-mkdir -p $HOME/volumes/pecan/{lib,pecan,portainer,postgres,rabbitmq,traefik}
+mkdir -p $HOME/volumes/pecan/{R_library,pecan,portainer,postgres,rabbitmq,traefik}
```
infotroph, line 118: 1. Why this renaming, you ask? Because I kept forgetting what was stored in `pecan_lib` and eventually realized a more descriptive name could help.
2. Question: Given we now recommend letting Docker manage most of these, is this still the right list of dirs to create?

DEV-INTRO.md
Sweetdevil144: I don't think I required it while setting up PEcAn on my Mac. It works fine without these lines for me

DEV-INTRO.md
```diff
@@ -78,70 +72,57 @@ The steps in this section only need to be done the first time you start working
...
+The goal of the development is to share the development folder with your container, whilst minimizing the latency. What this will do is setup the folders to allow for your pecan folder to be shared, and keep the rest of the folders managed by docker. Some of this is based on Dave Scott's DockerCon 2020 presentation ["New Docker Desktop Filesharing Features"](https://www.youtube.com/watch?v=gyddZyc8r48). In this talk it is recommended to keep the database on the filesystem managed by docker, as well as any other folders that are not directly modified on the host system (not using the docker managed volumes could lead to a large speed loss when reading/writing to the disk). The `docker-compose.override.yml` can be modified to copy all the data to the local filesystem, by uncommenting the appropriate blocks. If you are sharing more than the pecan home directory you will need to make sure that these folder exist. As from the video, it is recommended to keep these folders outside of the actual pecan folder to allow for better caching capabilities of the docker system.
```
infotroph, line 113:
```suggestion
The goal of the development is to share the development folder with your container, whilst minimizing the latency. What this will do is setup the folders to allow for your pecan folder to be shared, and keep the rest of the folders managed by docker. Some of this is based on Dave Scott's DockerCon 2020 presentation ["New Docker Desktop Filesharing Features"](https://www.youtube.com/watch?v=gyddZyc8r48&t=512s). In this talk it is recommended to keep the database on the filesystem managed by docker, as well as any other folders that are not directly modified on the host system (not using the docker managed volumes could lead to a large speed loss when reading/writing to the disk). The `docker-compose.override.yml` can be modified to copy all the data to the local filesystem, by uncommenting the appropriate blocks. If you are sharing more than the pecan home directory you will need to make sure that these folder exist. As from the video, it is recommended to keep these folders outside of the actual pecan folder to allow for better caching capabilities of the docker system.
```

DEV-INTRO.md
robkooper: https://docs.google.com/document/d/1ipa3ROpHgZYLsTMBClgMMzAbVm_vI67iZsbTn29uDWg/edit#heading=h.157oh7kug6t6

robkooper: I think this might work for the mac but it could depend on what way the file system in shared into the container.

DEV-INTRO.md
robkooper: 1. sounds good.
2. if you want to develop, you really want to have the R_library shared, since this is the folder in the base folder and if you make changes in the `sipnet` model, it will not be seen the sipnet container, unless you recompile pecan there.

robkooper (approved):
LGTM

DEV-INTRO.md
infotroph: Let's leave it as it is for now and I'll explore more.

PR #3332, opened 2024-07-12
Update NASA DAAC download function to include all DAAC server.

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


modules/data.remote/R/NASA_DAAC_download.R
```diff
@@ -46,37 +48,38 @@ NASA_DAAC_download <- function(ul_lat,
...
+                               outdir = NULL,
                                doi,
                                netrc_file = NULL,
                                just_path = FALSE) {
   # if there is no credential file in the outdir.
   # we will create a new one.
   # this function is located within the GEDI_AGB_prep script.
-  if (length(list.files(outdir, pattern = "netrc")) == 0) {
-    netrc_file <- getnetrc(outdir)
+  if (is.null(outdir) & !just_path) {
```
mdietze, line 58: I suspect a better default for outdir is the current directory (e.g. getwd() or "."). It's unexpected to provide a default value and then have that default value throw an error, as that means the argument is required and shouldn't have a default.

mdietze (changes requested):

modules/data.remote/R/NASA_DAAC_download.R
DongchenZ: Fixed.

mdietze (approved):

PR #3337, opened 2024-07-15
relicense QAQC and PEcAn.STICS as BSD3

infotroph:
Updating the license file upon noting we have agreement from all contributors whose code is still present in that package.
* QAQC: @ashiklom , @infotroph, @DongchenZ, @mdietze, @robkooper, @allgandalf, @mccabete, @tonygardella, @istfer, @meetagrawal09     
* STICS: @infotroph, @DongchenZ, @robkooper, @nanu1605, @istfer, @qdbell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - The following components have changed their licensing. With approval of all their contributors, we now provide them under a BSD 3-clause license rather than the previously used NCSA Open Source license. As a reminder, we intend to relicense the entire system and this list will expand as we gather permission from the relevant copyright owners.
     * Shiny apps `dbsync`, `Data-Ingest`, and `Elicitation`
     * Base packages `PEcAn.all` and `PEcAn.QAQC`
-    * Model packages `PEcAn.LDNDC` and `PEcAn.SIBCASA`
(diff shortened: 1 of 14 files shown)


dlebauer:
I don't see the new licenses?

infotroph:
@dlebauer Good point. I've just added the BSD3 text to the LICENSE in the project root. 

We're following CRAN's (weird IMO, but consistently documented) rule that for BSD3 the LICENSE file is a template that should contain only the year organization, and copyright holder, and that the built R package should _not_ contain the entire text of the license. 

I think providing one copy for the whole repo is enough, but if folks feel strongly about keeping license text in the source of every individual package we can add it as a file named something other than `LICENSE` (maybe `LICENSE.txt`?) and then add that to .Rbuildignore.

robkooper:
Should this not have the License file anymore? https://github.com/PecanProject/pecan/blob/cc86b20d0575eb8e2acfc17bd7d73f66adf08ac3/models/stics/DESCRIPTION#L28

infotroph:
@robkooper DESCRIPTION should still say "+ file LICENSE", and the file should still be there but with changed contents -- See my answer to David above.

dlebauer:
I agree one license file for the whole repo is enough. Mostly I think it is helpful when the license is clear when people find the GitHub repo.

mdietze (approved):

PR #3338, opened 2024-07-18
Refactor `convert_input` to Perform tasks via helper function

Sweetdevil144:
## Description

Following changes were performed :
- [x] Shift functions to check for missing files
- [x] Return from convert_input via a helper function
- [x] Refactor extra variables in `run.meta.analysis`
- [x] Update corresponding test files and add tests to ensure do_conversions isn't affected by current applied changes

## Motivation and Context
The main motive for these changes is to simplify convert_input by trying to break some of its components and branch it to other functions.

This PR may fix a task within #3307 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/db/R/convert_input.R
```diff
@@ -734,143 +734,23 @@ convert_input <-
...
-  if (exists("existing.dbfile") && is.data.frame(existing.dbfile)) {
-    existing.dbfile <- list(existing.dbfile)
-  }
+  checked.missing.files <- check_missing_files(result, outname, existing.input, existing.dbfile)
+
+  # Unwrap parameters after performing checks for missing files
+  result_sizes <- checked.missing.files$result_sizes;
+  outlist <- checked.missing.files$outlist;
+  existing.input <- checked.missing.files$existing.input;
+  existing.dbfile <- checked.missing.files$existing.dbfile;
```
Sweetdevil144: Is their any other way to unwrap these variables in a single context line. For example, inheriting the above list and utilising the `list$var` syntax in-place. A lot of variable operations(assignment and reassignment) would consume extra memory in the heap. Should we also implement a Garbage Collection (`gc()`) or is it automatically applied in R ?

infotroph: Rather than try to unwrap them in one line I'd consider whether they need to be assigned separate names at all -- e.g. can all places below that refer to `result_sizes` refer to `checked.missing.files$result_sizes` instead? If that's what you're suggesting by "utilising the list$var syntax in-place", then yes, I like that approach.

That said these objects are all very small and the runtime/memory overhead from reassignment will be negligible and R's reference-counting garbage collection will take care of the cleanup for us. Happily though, the same optimization gives (what I think is) a substantial improvement in code clarity, so I endorse it on those grounds.

base/db/R/convert_input.R
```diff
@@ -734,143 +734,23 @@ convert_input <-
...
-      
-    } #End for loop
-    
-    successful <- TRUE
-    return(newinput)
-  } else {
-    PEcAn.logger::logger.warn("Input was not added to the database")
-    successful <- TRUE
-    return(NULL)
-  }
```
infotroph, line 773: I don't yet see where this chunk is moved to

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,49 @@
...
+  
+  # Wrap in a list for consistant processing later
+  if (exists("existing.input") && is.data.frame(existing.input)) {
+    existing.input <- list(existing.input)
+  }
+  
+  if (exists("existing.dbfile") && is.data.frame(existing.dbfile)) {
+    existing.dbfile <- list(existing.dbfile)
+  }
+  return(list(result_sizes, outlist, existing.input, existing.dbfile))
```
infotroph: Can you check how many of these values are actually used later? On a quick look I think `result_sizes` is only used on lines 22 and 31 above and thus doesn't need to be returned from this function, but please confirm that before removing.

base/db/R/convert_input.R
Sweetdevil144: This section was moved to `add.database.entries.R`

base/db/R/check.missing.files.R
Sweetdevil144: `result_sizes` can be safely removed. It is only utilised within `check.missing.files` to verify if _Requested Processing produced empty files or Nonexistent files_. Similarly, `outlist` is also not utilised after its declaration as follow :

```r
## base/db/R/check.missing.files.R

 # Insert into Database
  outlist <- unlist(strsplit(outname, "_"))
```

`outlist` too is not utilised anytime further. So, that too can be safely deleted since it wasn't used after initialisation.

Sweetdevil144:
Why do docker stacks fail with following msg?

```
Dockerfile:29
--------------------
  28 |     WORKDIR /src
  29 | >>> RUN git -c http.sslVerify=false clone https://github.com/EDmodel/ED2.git \
  30 | >>>     && cd ED2/ED/build \
  31 | >>>     && curl -o make/include.mk.VM http://isda.ncsa.illinois.edu/~kooper/EBI/include.mk.opt.`uname -s` \
  32 | >>>     && if [ "${MODEL_VERSION}" != "git" ]; then git checkout "v.${MODEL_VERSION}"; fi \
  33 | >>>     && ./install.sh -g -p VM \
  34 | >>>     && mv /src/ED2/ED/build/ed_${BINARY_VERSION}-opt /src/ED2/ED/build/ed 
  35 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c git -c http.sslVerify=false clone https://github.com/EDmodel/ED2.git     && cd ED2/ED/build     && curl -o make/include.mk.VM http://isda.ncsa.illinois.edu/~kooper/EBI/include.mk.opt.`uname -s`     && if [ \"${MODEL_VERSION}\" != \"git\" ]; then git checkout \"v.${MODEL_VERSION}\"; fi     && ./install.sh -g -p VM     && mv /src/ED2/ED/build/ed_${BINARY_VERSION}-opt /src/ED2/ED/build/ed" did not complete successfully: exit code: 1
```

base/db/R/get.machine.info.R
```diff
@@ -0,0 +1,83 @@
+#' Get machine information from db
+#' @param host host information
+#' @param input.args input args for existing records
+#' @param input.id input id for existing records
+#' @param con database connection
+#'
+#' @return list of machine, input, and dbfile records
+#' @author Betsy Cowdery, Michael Dietze, Ankur Desai, Tony Gardella, Luke Dramko
+
+get.machine.info <- function(host, input.args, input.id = NULL, con = NULL) {
```
infotroph: Might as well use underscore instead of dot in the name here to avoid possible confusion with S3 methods. These days I only use dot where it's needed for consistency with existing code, and the db package is already inconsistent about it 😉 
```suggestion
get_machine_info <- function(host, input.args, input.id = NULL, con = NULL) {
```

infotroph: Need to change the usages below to match, naturally

base/db/R/get.machine.info.R
```diff
@@ -0,0 +1,83 @@
...
+
+  return(list(machine = machine, input = input, dbfile = dbfile))
+}
+
+#' Helper Function to retrieve machine host and machine informations
+#' @param host host information
+#' @param con database connection
+#' @return list of machine host and machine information
+#' @author Abhinav Pandey
+get.machine.host <- function(host, con = NULL) {
```
infotroph: 1. rename as discussed for get.machine.info above
2. db.query will fail if con = NULL, might as well make it mandatory by providing  no default value
```suggestion
get_machine_host <- function(host, con) {
```

base/db/R/get.machine.info.R
```diff
@@ -0,0 +1,83 @@
...
+  machine.host.info <- get.machine.host(host, con = con)
+  machine.host <- machine.host.info$machine.host
+  machine <- machine.host.info$machine
+
+  if (nrow(machine) == 0) {
+    PEcAn.logger::logger.error("machine not found", host$name)
+    return(NULL)
+  }
+
+  if (missing(input.id) || is.na(input.id) || is.null(input.id)) {
```
infotroph: Don't need the `missing()` here because get.machine.info now defaults it to NULL
```suggestion
  if (is.na(input.id) || is.null(input.id)) {
```

base/db/R/get.machine.info.R
```diff
@@ -0,0 +1,83 @@
...
+
+get.machine.info <- function(host, input.args, input.id = NULL, con = NULL) {
+
+  machine.host.info <- get.machine.host(host, con = con)
+  machine.host <- machine.host.info$machine.host
+  machine <- machine.host.info$machine
+
+  if (nrow(machine) == 0) {
+    PEcAn.logger::logger.error("machine not found", host$name)
+    return(NULL)
```
infotroph, line 18: Note that returning from get.machine.info could have different behavior than returning from convert.input. I'll need to read to the end to know whether that difference matters, but mentioning it now before I forget

base/db/tests/testthat/test.check.missing.files.R
```diff
@@ -0,0 +1,19 @@
...
+
+    res <- check_missing_files(
+        result = list(data.frame(file = c("A", "B"))),
+        outname = "test",
+        existing.input = data.frame(),
+        existing.dbfile = data.frame()
+    )
+
+    # Print the structure of `res` for debugging
+    str(res)
```
infotroph:
```suggestion
```

base/db/R/convert_input.R
```diff
@@ -382,7 +380,7 @@ convert_input <-
     if (!is.null(ensemble) && ensemble) {
       return.all <-TRUE
       
-    }else{
+    } else{
```
infotroph:
```suggestion
    } else {
```

base/db/R/convert_input.R
```diff
@@ -174,17 +174,15 @@ convert_input <-
         # Date/time processing for existing input
         existing.input[[i]]$start_date <- lubridate::force_tz(lubridate::as_datetime(existing.input[[i]]$start_date), "UTC")
         existing.input[[i]]$end_date   <- lubridate::force_tz(lubridate::as_datetime(existing.input[[i]]$end_date), "UTC")
-        
+
         ## Obtain machine information
+        machine.host.info <- get.machine.host(host, con = con)
+        machine.host <- machine.host.info$machine.host
+        machine <- machine.host.info$machine
```
infotroph, line 181: This block appears a bunch of times in this file -- is it identical each time? Could it be pulled up to the top of the function rather than being rewritten in every else block?

base/db/R/convert_input.R
```diff
@@ -512,66 +510,17 @@ convert_input <-
...
-    if (nrow(dbfile) == 0) {
-      PEcAn.logger::logger.error("dbfile not found", input.id)
-      return(NULL)
-    }
-    if (nrow(dbfile) > 1) {
-      PEcAn.logger::logger.warn("multiple dbfile records, using last", dbfile)
-      dbfile <- dbfile[nrow(dbfile), ]
-    }
-  }
+  machine <- machine.info$machine
```
infotroph, line 496: Aha, here's where my comment above about return values comes back:
* in the existing code, if we don't find any results when looking up the machine, input, or dbfile then the whole `convert.input` function will return NULL before performing any conversions
* with this change, if we don't find any results then these `machine`, `input`, and `dbfile` variables will all be NULL, but execution of `convert.input` will continue into the next section if not stopped by a `logger.error()`.

Might be worth checking for this just in case:
```suggestion
  if (any(sapply(machine.info, is.null))) {
    PEcAn.logger::logger.error("failed lookup of inputs or dbfiles")
    return(NULL)  
  }
  machine <- machine.info$machine
```

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,47 @@
...
+      empty = file_size == 0
+    )
+  )
+
+  if (any(result_sizes$missing) || any(result_sizes$empty)) {
+    log_format_df <- function(df) {
+      formatted_df <- rbind(colnames(df), format(df))
+      formatted_text <- purrr::reduce(formatted_df, paste, sep = " ")
+      paste(formatted_text, collapse = "\n")
+    }
```
infotroph: Let's pull this function definition out to the end of the file -- I think the runtime overhead of nested function definitions is small, but I like to avoid them for clarity anyway.

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,47 @@
...
+    PEcAn.logger::logger.severe(
+      "Requested Processing produced empty files or Nonexistent files:\n",
+      log_format_df(result_sizes[, c(1, 8, 9, 10)]),
+      "\n Table of results printed above.",
+      wrap = FALSE
+    )
+  }
+
+
+  # Wrap in a list for consistant processing later
```
infotroph:
```suggestion
  # Wrap in a list for consistent processing later
```

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,47 @@
...
+  if (exists("existing.input") && is.data.frame(existing.input)) {
```
infotroph: Unless I'm forgetting something, `exists(x)` will always be TRUE inside a function that defines x as an argument, whether x or not x was specified at call time.
```suggestion
  if (is.data.frame(existing.input)) {
```

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,47 @@
...
+    existing.input <- list(existing.input)
+  }
+
+  if (exists("existing.dbfile") && is.data.frame(existing.dbfile)) {
```
infotroph:
```suggestion
  if (is.data.frame(existing.dbfile)) {
```

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,47 @@
+#' Function to check if result has empty or missing files
+#'
+#' @param result A list of dataframes with file paths
+#' @param outname Name of the output file
```
infotroph: Not used (right?)
```suggestion
```

base/db/R/check.missing.files.R
```diff
@@ -0,0 +1,47 @@
...
+#' @param existing.input Existing input records
+#' @param existing.dbfile Existing dbfile records
+#' @return A list of dataframes with file paths, a list of strings with the output file name, a list of existing input records, and a list of existing dbfile records
+#'
+#' @author Betsy Cowdery, Michael Dietze, Ankur Desai, Tony Gardella, Luke Dramko
+
+check_missing_files <- function(result, outname, existing.input = NULL, existing.dbfile = NULL) {
```
infotroph:
```suggestion
check_missing_files <- function(result, existing.input = NULL, existing.dbfile = NULL) {
```

infotroph: (need to update calls too)

base/db/R/add.database.entries.R
```diff
@@ -0,0 +1,121 @@
...
+
+add.database.entries <- function(
+    result, con, start_date,
+    end_date, write, overwrite,
+    insert.new.file, input.args,
+    machine, mimetype, formatname,
+    allow.conflicting.dates, ensemble,
+    ensemble_name, existing.input,
+    existing.dbfile, input) {
+    if (write) {
```
infotroph: I think it'd be clearer to leave this `if` outside of this function, so that convert.input contains something that amounts to `if (write)  { add.database.entries } else { logger.warn("Input was not added to the database")}`

base/db/R/add.database.entries.R
```diff
@@ -0,0 +1,121 @@
...
+        # This list will be returned.
+        newinput <- list(input.id = NULL, dbfile.id = NULL) # Blank vectors are null.
+
+        for (i in 1:length(result)) { # Master for loop
+            id_not_added <- TRUE
+
+            if (!is.null(existing.input) && nrow(existing.input[[i]]) > 0 &&
+                (existing.input[[i]]$start_date != start_date || existing.input[[i]]$end_date != end_date)) {
+                # Updating record with new dates
+                db.query(paste0("UPDATE inputs SET start_date='", start_date, "', end_date='", end_date, "' WHERE id=", existing.input[[i]]$id), con)
```
infotroph: just for readability
```suggestion
                db.query(
                  paste0(
                    "UPDATE inputs SET start_date='", start_date, "', end_date='", end_date,
                    "' WHERE id=", existing.input[[i]]$id),
                  con)
```

base/db/R/add.database.entries.R
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            if (overwrite) {
+                # A bit hacky, but need to make sure that all fields are updated to expected values (i.e., what they'd be if convert_input was creating a new record)
+                if (!is.null(existing.input) && nrow(existing.input[[i]]) > 0) {
+                    db.query(paste0("UPDATE inputs SET name='", basename(dirname(result[[i]]$file[1])), "' WHERE id=", existing.input[[i]]$id), con)
+                }
+
+                if (!is.null(existing.dbfile) && nrow(existing.dbfile[[i]]) > 0) {
+                    db.query(paste0("UPDATE dbfiles SET file_path='", dirname(result[[i]]$file[1]), "', file_name='", result[[i]]$dbfile.name[1], "' WHERE id=", existing.dbfile[[i]]$id), con)
```
infotroph:
```suggestion
                    db.query(
                      paste0(
                        "UPDATE dbfiles SET file_path='", dirname(result[[i]]$file[1]),
                          "', file_name='", result[[i]]$dbfile.name[1],
                        "' WHERE id=", existing.dbfile[[i]]$id),
                      con)
```

base/db/R/add.database.entries.R
```diff
@@ -0,0 +1,121 @@
...
+                    hostname = machine$hostname
+                )
+
+                newinput$input.id <- c(newinput$input.id, existing.input[[i]]$id)
+                newinput$dbfile.id <- c(newinput$dbfile.id, dbfile.id)
+            } else if (id_not_added) {
+                # This is to tell input.insert if we are writing ensembles
+                # Why does it need it? Because it checks for inputs with the same time period, site, and machine
+                # and if it returns something it does not insert anymore, but for ensembles, it needs to bypass this condition
+                ens.flag <- if (!is.null(ensemble) | is.null(ensemble_name)) TRUE else FALSE
```
infotroph:
```suggestion
                ens.flag <- (!is.null(ensemble) || is.null(ensemble_name))
```

base/db/R/add.database.entries.R
```diff
@@ -0,0 +1,121 @@
...
+                    allow.conflicting.dates = allow.conflicting.dates,
+                    ens = ens.flag
+                )
+
+                newinput$input.id <- c(newinput$input.id, new_entry$input.id)
+                newinput$dbfile.id <- c(newinput$dbfile.id, new_entry$dbfile.id)
+            }
+        } # End for loop
+
+        successful <- TRUE
```
infotroph: Looks like `successful` will need to be passed back to `convert.input` somehow -- it uses it to determine what files to clean up when it exits

infotroph: On a quick skim it looks like none of the cases that can set  `successful = FALSE` occur inside `add.database.entries`. Maybe a possible solution is to remove them here and have convert.input set `successful` after it calls `add.database.entries` and before returning?

base/db/tests/testthat/test.check.missing.files.R
```diff
@@ -0,0 +1,19 @@
+test_that("`check_missing_files()` able to return correct missing files", {
+    # Mock `purrr::map_dfr`
+    mocked_res <- mockery::mock(data.frame(file = c("A", "B"), file_size = c(100, 200), missing = c(FALSE, FALSE), empty = c(FALSE, FALSE)))
+    mockery::stub(check_missing_files, "purrr::map_dfr", mocked_res)
```
infotroph: I think what happens inside map_dfr is part of what we want to be testing here -- might be better to stub out `file.size` since that's the only place that would be looking at the filesystem.
```suggestion
    mocked_size <- mockery::mock(100,200)
    mockery::stub(check_missing_files, "file.size", mocked_res)
```

infotroph: It would also be good to include cases with missing and empty files and confirm that it produces the expected error

base/db/tests/testthat/test.check.missing.files.R
```diff
@@ -0,0 +1,19 @@
+test_that("`check_missing_files()` able to return correct missing files", {
```
infotroph, line 1: Filename should match function name (so use `_` not `.`)
Come to think of it, same for the files containing the function definitions

base/db/tests/testthat/test.convert_input.R
```diff
@@ -1,10 +1,26 @@
...
+      file = c("A", "B"),
+      file_size = c(100, 200),
+      missing = c(FALSE, FALSE),
+      empty = c(FALSE, FALSE)
+    ),
+    outlist = "test",
+    existing.input = list(data.frame(file = character(0))),
+    existing.dbfile = list(data.frame(file = character(0)))
+  ))
+  mockery::stub(convert_input, "add.database.entries", list(input.id = 1, dbfile.id = 1))
```
infotroph: This feels like a lot of mocks and stubs, but I don't see any that aren't needed. Hopefully as we break up the function further we'll be able to simplify these.

infotroph (changes requested):

base/db/R/add.database.entries.R
Sweetdevil144: Applied `successful <- TRUE` outside the function block in `convert_input`. Can you review it once?

base/db/R/check.missing.files.R
Sweetdevil144: Yes. Forgot to erase the var!!

Sweetdevil144 (commented):
> (need to update calls too) 

Updating calls as well as tests

Sweetdevil144 (commented):
Extra checks fail due to some reasons

Sweetdevil144 (commented):
`extras` check fail?

base/db/R/convert_input.R
Sweetdevil144: That depends on the host machine we are writing it to. If the host machine calls a seperate machine, then the info may change. Does any such occurance occurs in PEcAN's model runs?

base/db/tests/testthat/test.check.missing.files.R
Sweetdevil144: Fixed

base/db/tests/testthat/test.convert_input.R
Sweetdevil144: Another option would be to pass a custom dataset with known results through the test comparing it with corresponding output for input file. In simple terms, processing data similar to how we wrote test for `met2CF.csv.R`. Is it feasible? Or maybe I am going in the wrong direction?

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
+#' Return new arrangement of database while adding code to deal with ensembles
```
mdietze: This doesn't tell me anything about what this function does.

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+#' @param ensemble ensemble id
+#' @param ensemble_name ensemble name
+#' @param existing.input existing input records
+#' @param existing.dbfile existing dbfile records
+#' @param input input records
+#' @return list of input and dbfile ids
+#'
+#' @author Betsy Cowdery, Michael Dietze, Ankur Desai, Tony Gardella, Luke Dramko
+
+add_database_entries <- function(
```
mdietze: The name of this function is very generic. We have LOTS of functions in PEcAn.DB that add db entries. Could you update the name to clarify what type of entries this function is adding?

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
+#' Return new arrangement of database while adding code to deal with ensembles
+#'
+#' @param result list of results from the download function
```
mdietze: PEcAn has a ton of different download functions, so you can't just say "the". Could you clarify which function or set of functions and, more importantly, how this list needs to be structured

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+#' @param con database connection
+#' @param start_date start date of the data
+#' @param end_date end date of the data
```
mdietze: What formats are acceptable for start and end date? Character string? Date object? POSIX object?

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+#' @param write whether to write to the database
```
mdietze: Clarify that I think you mean this needs to be a boolean/logical variable. Also this variable is in the Roxygen here but not in the function definition below

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+#' @param overwrite Logical: If a file already exists, create a fresh copy?
+#' @param insert.new.file whether to insert a new file
```
mdietze: clarify that this should be boolean/logical

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+#' @param input.args input arguments obtained from the convert_input function
```
mdietze: Is input.args a documented output of the convert_input function or an undocumented internal variable? Just thinking you might want to provide a bit more info about what info this argument actually contains and how it's organized or even basic things like what variable type it is (I'm guessing list?)

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+#' @param machine machine information
+#' @param mimetype data product specific file format
+#' @param formatname format name of the data
+#' @param allow.conflicting.dates whether to allow conflicting dates
+#' @param ensemble ensemble id
+#' @param ensemble_name ensemble name
+#' @param existing.input existing input records
+#' @param existing.dbfile existing dbfile records
+#' @param input input records
```
mdietze: Rather than continuing to provide line-by-line comments, I'll just leave the general comment that there's not enough info about most variables. Would be good to know what things are list vs data frames vs vectors vs scalars. Which things are logical vs stings vs numeric. etc.

base/db/R/add_database_entries.R
```diff
@@ -0,0 +1,130 @@
...
+    insert.new.file, input.args,
+    machine, mimetype, formatname,
+    allow.conflicting.dates, ensemble,
+    ensemble_name, existing.input,
+    existing.dbfile, input) {
+    # Setup newinput. This list will contain two variables: a vector of input IDs and a vector of DB IDs for each entry in result.
+    # This list will be returned.
+    newinput <- list(input.id = NULL, dbfile.id = NULL) # Blank vectors are null.
+
+    for (i in 1:length(result)) { # Master for loop
```
mdietze: pretty sure seq_along would be better here

base/db/R/check_missing_files.R
```diff
@@ -0,0 +1,46 @@
+#' Function to check if result has empty or missing files
```
mdietze: result of what? Need more context

base/db/R/check_missing_files.R
```diff
@@ -0,0 +1,46 @@
...
+#'
+#' @param result A list of dataframes with file paths
+#' @param existing.input Existing input records
+#' @param existing.dbfile Existing dbfile records
```
mdietze: expand param definitions

base/db/R/convert_input.R
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
-      }
-      
-    } #End for loop
-    
-    successful <- TRUE
-    return(newinput)
-  } else {
-    PEcAn.logger::logger.warn("Input was not added to the database")
+    if(write) {
+      add_entries_result <- return (add_database_entries(result, con, start_date,
```
mdietze: odd formatting here

base/db/R/get_machine_info.R
```diff
@@ -0,0 +1,104 @@
+#' Get machine information from db
+#' @param host host information
+#' @param input.args input args for existing records
+#' @param input.id input id for existing records
+#' @param con database connection
```
mdietze, line 5: expand parameter info

base/db/R/get.machine.info.R
Sweetdevil144: I decided to instead terminate the current processing conversions rather than proceeding further when we don't get a machine info. Although, I don't think this piece of code is going to be invoked anytime soon until we introduce breaking changes

base/db/R/add_database_entries.R
Sweetdevil144: Does `update_ensemble_writes` or `update_ensemble_entries` sound good? Or would `update_input_and_dbfile_entries.R` be better?

base/db/R/convert_input.R
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
-      }
-      
-    } #End for loop
-    
-    successful <- TRUE
-    return(newinput)
-  } else {
-    PEcAn.logger::logger.warn("Input was not added to the database")
+    if(write) {
+      add_entries_result <- return (update_ensemble_writes(result, con, start_date,
```
harshagr70: add_entries_result <- update_ensemble_writes( result, con, start_date, end_date, overwrite, insert.new.file, input.args, machine, mimetype, formatname )
return(add_entries_result)


The return values assigned here to a variable likely is a bug , this patch might fix this up .

harshagr70: ctx <- list(
  con = con,
  start_date = start_date,
  end_date = end_date,
  overwrite = overwrite,
  insert_new_file = insert.new.file,
  input_args = input.args,
  machine = machine,
  mimetype = mimetype,
  formatname = formatname
)

update_ensemble_writes(result, ctx)

The long argument list can be concatenated into the ctx , this will help cater the too many arguments issue.

base/db/R/get_machine_info.R
```diff
@@ -0,0 +1,104 @@
...
+      PEcAn.logger::logger.error("dbfile not found", input.id)
+      return(NULL)
+    }
+    if (nrow(dbfile) > 1) {
+      PEcAn.logger::logger.warn("multiple dbfile records, using last", dbfile)
+      dbfile <- dbfile[nrow(dbfile), ]
+    }
+  }
+
+  return(list(machine = machine, input = input, dbfile = dbfile))
```
harshagr70, line 66: A better way would be to just return a named list here .
Downstream code in convert_input() currently accesses with position ([[1]], [[2]], etc.), which is fragile. Using $machine, $input, etc. would be cleaner and safer

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Change Log
+
```
```suggestion
```

base/db/R/get_machine_info.R
```diff
@@ -0,0 +1,104 @@
...
+#' @author Abhinav Pandey
+get_machine_host <- function(host, con) {
+  #Grab machine info of host machine
+  machine.host <- ifelse(host$name == "localhost", PEcAn.remote::fqdn(), host$name)
+  machine <- db.query(paste0(
+    "SELECT * from machines where hostname = '",
+    machine.host, "'"
+  ), con)
+
+  return(list(machine.host, machine))
```
infotroph:
```suggestion
  list(machine.host = machine.host, machine = machine)
```

base/db/R/convert_input.R
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
+                                                          end_date, overwrite,
+                                                          insert.new.file, input.args,
+                                                          machine, mimetype, formatname,
+                                                          allow.conflicting.dates, ensemble,
+                                                          ensemble_name, existing.input,
+                                                          existing.dbfile, input))
```
infotroph:
```suggestion
      return(
        update_ensemble_writes(result, con, start_date,
                               end_date, overwrite,
                               insert.new.file, input.args,
                               machine, mimetype, formatname,
                               allow.conflicting.dates, ensemble,
                               ensemble_name, existing.input,
                               existing.dbfile, input)
      )
```

base/db/R/convert_input.R
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
-        newinput$dbfile.id <- c(newinput$dbfile.id, new_entry$dbfile.id)
-      }
-      
-    } #End for loop
-    
-    successful <- TRUE
-    return(newinput)
-  } else {
-    PEcAn.logger::logger.warn("Input was not added to the database")
+    if(write) {
```
infotroph:
```suggestion
  if (write) {
```

base/db/R/convert_input.R
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
+                                                          existing.dbfile, input))
+    } else {
+      PEcAn.logger::logger.warn("Input was not added to the database")
+      successful <- TRUE
+      return(NULL)
+    }
     successful <- TRUE
-    return(NULL)
-  }
+    return (add_entries_result)
```
infotroph: unreachable
```suggestion
```

base/db/R/convert_input.R
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
+                                                          machine, mimetype, formatname,
+                                                          allow.conflicting.dates, ensemble,
+                                                          ensemble_name, existing.input,
+                                                          existing.dbfile, input))
+    } else {
+      PEcAn.logger::logger.warn("Input was not added to the database")
+      successful <- TRUE
+      return(NULL)
+    }
     successful <- TRUE
```
infotroph: To make it clearer this is the last possible exit point from the function
```suggestion
    }
    # if we got here, nothing left to do
    PEcAn.logger::logger.warn("Input was not added to the database")
    return(NULL)
    ```

**Review by infotroph — APPROVED** [review]

File: base/db/R/convert_input.R [hunk 47]
```diff
@@ -634,143 +558,29 @@ convert_input <-
...
+                               ensemble_name, existing.input,
+                               existing.dbfile, input)
+      )
+    }
+    # if we got here, nothing left to do
     PEcAn.logger::logger.warn("Input was not added to the database")
-    successful <- TRUE
     return(NULL)
-  }
+    ```
```
infotroph:
```suggestion
```

PR #3340, opened 2024-07-22
Improved the current SDA workflow to reach the North American runs with 6400 sites.

DongchenZ:
## Motivation and Context

1. The `foreach` package seems to be better compared to the `furrr` package concerning memory allocation. Thus, in this PR, I replaced every `furrr` with `foreach` during the general SDA workflow.
2. The computational power and memory are limited when executing certain SDA procedures locally (e.g., splitting meteorology files, writing configuration files, reading SDA outputs, running Bayesian MCMC analysis part, and removing files (e.g., removing NC files after the first SDA run)). Therefore, in the PR, I developed features of `qsub` submissions (specified by the `batch.settings` section in the XML file, which is also documented in the pecan book) during the SDA workflow.
3. To avoid the complex `if-else` usage within the current `sda.enkf_Multisite` function for the above `batch` job submissions, I developed a new `sda.enkf_NorthAmerica` function (although most of them are just copy-and-paste from the `sda.enkf_Multisite` function), which is cleaner and will only be used if the `batch.settings` is not empty.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/assim.sequential/NAMESPACE
```diff
@@ -61,6 +63,7 @@ import(furrr)
 import(lubridate)
 import(nimble)
 importFrom(dplyr,"%>%")
+importFrom(foreach,"%dopar%")
```
infotroph, line 67: Current CI failure translates to "You're not allowed to importFrom if the package is in Suggests rather than Imports"

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +639,99 @@ block.2.vector <- function (block.list, X, H) {
...
+    qsub <- gsub("@STDERR@", file.path(folder.path, "stderr.log"), qsub)
+    qsub <- strsplit(qsub, " (?=([^\"']*\"[^\"']*\")*[^\"']*$)", perl = TRUE)
+    cmd <- qsub[[1]]
+    out <- system2(cmd, file.path(folder.path, "job.sh"), stdout = TRUE, stderr = TRUE)
+  }
+  # checking results.
+  PEcAn.logger::logger.info("Checking results.")
+  while (length(list.files(outdir, pattern="results.Rdata", recursive = T)) < folder.num) {
+    Sys.sleep(60)
+  }
```
infotroph: Flagging that this has high potential to become an infinite loop if any job fails / is rejected by the queue / etc

infotroph: More generally, I'm cautious of submitting anything to qsub and waiting for it to complete before returning from the submitting function -- Part of the point of qsub is that it does the waiting for you.

modules/assim.sequential/NAMESPACE
DongchenZ: It doesn't allow me to include more than 20 dependencies.

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +639,99 @@ block.2.vector <- function (block.list, X, H) {
...
+  # assemble results.
+  PEcAn.logger::logger.info("Assembling results.")
+  analysis <- c()
+  for (path in seq_along(folder.paths)) {
+    load(file.path(path, "results.Rdata"))
+    analysis <- c(analysis, results)
```
infotroph:
```suggestion
    res_env <- new.env()
    load(file.path(path, "results.Rdata"), envir = res_env)
    analysis <- c(analysis, res_env$results)
```

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +639,99 @@ block.2.vector <- function (block.list, X, H) {
...
+##' @title qsub_analysis
+##' @param folder.path character: path where the `block.Rdata` file is stored.
+##' @importFrom foreach %dopar%
+##' @export
+qsub_analysis <- function(folder.path) {
+  # load file.
+  load(file.path(folder.path, "block.Rdata"))
+  # initialize parallel.
+  cores <- parallel::detectCores()
+  if (cores > 28) cores <- 28
```
infotroph: This looks like a BU-specific number and should be configurable

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +639,99 @@ block.2.vector <- function (block.list, X, H) {
...
+  cl <- parallel::makeCluster(cores)
+  doSNOW::registerDoSNOW(cl)
+  # progress bar
+  pb <- utils::txtProgressBar(min=1, max=length(blocks), style=3)
+  progress <- function(n) utils::setTxtProgressBar(pb, n)
+  opts <- list(progress=progress)
+  # parallel computation.
+  l <- NULL # fix GitHub check issue.
+  results <- NULL # fix GitHub check issue.
```
infotroph:
```suggestion
  l <- NULL # suppress R CMD check complaint (it doesn't know foreach assigns `l` below)
```

infotroph: Note that I'm fairly sure the `results <- NULL` is'nt needed at all -- I bet the "no binding" complaint was from line 706 above, and my suggestion there should hopefully fix it.

modules/assim.sequential/R/Analysis_sda_block.R

modules/assim.sequential/R/Analysis_sda_block.R

modules/assim.sequential/R/Analysis_sda_block.R
DongchenZ: I am not sure if I resolve this correctly. Here, instead, I detect the job completion by checking if the job still exists on the server. This will ensure a finite `while` loop even if, for some reason, some jobs never get finished and are killed by the server.

infotroph: > detect the job completion by checking if the job still exists on the server

👍 That seems like a solid improvement. -- still has the potential to be waiting a long time if the queue is stalled / moving slowly, but will only be waiting for outputs the server still plans to provide.

modules/assim.sequential/NAMESPACE
infotroph: Three mutually exclusive ideas:
* Find an existing package in Imports that's easier to move to Suggests to make room for `foreach`
* Add the "more than 20 imports" message to tests/Rcheck_reference.log so CI checks ignore it
* Use `%dopar%` directly from the namespace rather than importFrom, which is hideously ugly but appears to work in this test: 
```
x = 1:5
val <- NULL
foreach::foreach(val = x) |> foreach::`%dopar%`(log(val))
```

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,117 @@ block.2.vector <- function (block.list, X, H) {
...
+  doSNOW::registerDoSNOW(cl)
+  # progress bar
+  pb <- utils::txtProgressBar(min=1, max=length(blocks), style=3)
+  progress <- function(n) utils::setTxtProgressBar(pb, n)
+  opts <- list(progress=progress)
+  # parallel computation.
+  l <- NULL # fix GitHub check issue.
+  results <- foreach::foreach(l = blocks, .packages=c("Kendall", "purrr"), .options.snow=opts) %dopar% {
+    MCMC_block_function(l)
+  }
```
infotroph: Naive question: How closely equivalent is this to `parallel::parLapply(cl, blocks, MCMC_block_function)`?  Would it be worth considering that approach since it only uses existing dependencies?

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,117 @@ block.2.vector <- function (block.list, X, H) {
...
+      tail.num <- i*job.per.folder
+    }
+    # naming and creating folder.
+    folder.name <- paste0("From_", head.num, "_to_", tail.num)
+    folder.path <- file.path(outdir, "qsub_analysis", folder.name)
+    folder.paths <- c(folder.paths, folder.path)
+    dir.create(folder.path)
+    # save corresponding block list to the folder.
+    blocks <- block.list[head.num:tail.num]
+    save(blocks, file = file.path(folder.path, "block.Rdata"))
```
mdietze: I find it hard to tell what the structure of `blocks` actually is, but is this something that could be saved as a csv dataframe or something similar to avoid Rdata? At the minimum I'd recommend switching to RDS, which is more transparent about what objects are being reloaded.

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,117 @@ block.2.vector <- function (block.list, X, H) {
...
+    # create job file.
+    jobsh <- readLines(con = system.file("analysis_qsub.job", package = "PEcAn.ModelName"), n = -1, warn=FALSE)
```
mdietze: Why is the generic template model being hard-coded here? Do we really want to create a package dependency of the SDA module on the generic template model?

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,117 @@ block.2.vector <- function (block.list, X, H) {
...
               mu.a = mu.a,
               Pa = Pa))
+}
+
+##' This function provides means to split large SDA analysis (MCMC) runs into separate `qsub` jobs.
+##' Including job creation, submission, and assemble.
+##' @title qsub_analysis_submission
+##' @param block.list list: MCMC configuration lists for the block SDA analysis.
+##' @param outdir character: SDA output path.
+##' @param job.per.folder numeric: number of jobs per folder.
```
mdietze: This logic seems backwards to me. Generally with a HPC or cloud queueing system you want to specify the number of cores and/or nodes that you have, and then divide jobs across them.

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,117 @@ block.2.vector <- function (block.list, X, H) {
...
+  }
+  # create new folder.
+  dir.create(file.path(outdir, "qsub_analysis"))
+  # loop over sub-folders.
+  folder.paths <- job.ids <- c()
+  PEcAn.logger::logger.info(paste("Submitting", folder.num, "jobs."))
+  for (i in 1:folder.num) {
+    # create folder for each set of job runs.
+    # calculate start and end index for the current folder.
+    head.num <- (i-1)*job.per.folder + 1
```
mdietze: Are all jobs expected to take the same amount of time? If not (e.g. could one block have 1 site while another block has 1000 sites?), can we estimate which are expected to be longer or shorter so that we can load balance a bit more intelligently than doing so uniformly?

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,117 @@ block.2.vector <- function (block.list, X, H) {
...
+    dir.create(folder.path)
+    # save corresponding block list to the folder.
+    blocks <- block.list[head.num:tail.num]
+    save(blocks, file = file.path(folder.path, "block.Rdata"))
+    # create job file.
+    jobsh <- readLines(con = system.file("analysis_qsub.job", package = "PEcAn.ModelName"), n = -1, warn=FALSE)
+    jobsh <- gsub("@FOLDER_PATH@", folder.path, jobsh)
+    writeLines(jobsh, con = file.path(folder.path, "job.sh"))
+    # qsub command.
+    qsub <- "qsub -l h_rt=48:00:00 -l buyin -pe omp 28 -V -N @NAME@ -o @STDOUT@ -e @STDERR@ -S /bin/bash"
```
mdietze: qsub settings are system and user specific, and thus need to be read from the `settings` object, not hard-coded. Also, I'd STRONGLY recommend setting up a single array-style qsub over submitting multiple jobs in a loop.

modules/assim.sequential/R/Analysis_sda_block.R
DongchenZ: So, I haven't used this parallel function before. The reason that I used `foreach` is that it has great management of memory usage with large data inputs. But I can try this function out to see if it performs similarly to the `foreach` package.

modules/assim.sequential/R/Analysis_sda_block.R
DongchenZ: based on folder numbers.

modules/assim.sequential/R/Analysis_sda_block.R
DongchenZ: Should be balanced by number of sites within each block.

models/template/inst/analysis_qsub.job
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -l
+module load R/4.1.2
```
robkooper: this is very server specific, in the pecan.xml we have a section for qsub that allows you to specify the modules that need to be loaded. See https://pecanproject.github.io/pecan-documentation/master/xml-core-config.html#xml-host for an example.

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -632,4 +641,116 @@ block.2.vector <- function (block.list, X, H) {
...
+  # if we have previous outputs, remove them.
+  if (file.exists(file.path(outdir, "qsub_analysis"))) {
+    unlink(file.path(outdir, "qsub_analysis"), recursive = T)
+  }
+  # create new folder.
+  dir.create(file.path(outdir, "qsub_analysis"))
+  # loop over sub-folders.
+  folder.paths <- job.ids <- c()
+  PEcAn.logger::logger.info(paste("Submitting", folder.num, "jobs."))
+  for (i in 1:folder.num) {
```
robkooper: some systems will penalize you if you do many submissions in parallel, or only run two and then wait for those to be done, and more jobs you submit, the lower your priority in the queue. To overcome some of this I added the `modellauncher` this will create a text file with as first line the command to execute, and will have a list of folders it needs to run the command in.

modules/assim.sequential/R/Analysis_sda_block.R
robkooper: See [remote execution](https://pecanproject.github.io/pecan-documentation/master/xml-core-config.html#xml-host). Also a lot of this code seems to be in the PEcAn::remote package for example `start_qsub`

robkooper (changes requested):

base/remote/R/check_qsub_status.R
```diff
@@ -25,7 +26,9 @@ qsub_run_finished <- function(run, host, qstat) {
   }
 
   if (length(out) > 0 && substring(out, nchar(out) - 3) == "DONE") {
-    PEcAn.logger::logger.debug("Job", run, "for run", run_id_string, "finished")
+    if (verbose) {
+      PEcAn.logger::logger.debug("Job", run, "for run", run_id_string, "finished")
```
infotroph: PEcAn.logger already has built-in verbosity control via `logger.setLevel()` -- is a function-specific verbose flag needed here or is it enough for the user to set the logger level to something higher than debug so that this message isn't printed?

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -430,7 +439,7 @@ MCMC_block_function <- function(block) {
   conf$addSampler(target = samplerLists[[X.mod.ind]]$target, type = "ess",
                   control = list(propCov= block$data$pf, adaptScaleOnly = TRUE,
                                  latents = "X", pfOptimizeNparticles = TRUE))
-
+
```
infotroph:
```suggestion

```

modules/assim.sequential/DESCRIPTION
```diff
@@ -65,4 +67,4 @@ Suggests:
 License: BSD_3_clause + file LICENSE
 Copyright: Authors
 Encoding: UTF-8
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.2
```
infotroph:
```suggestion
RoxygenNote: 7.3.2

```

modules/assim.sequential/R/Multi_Site_Constructors.R
```diff
@@ -6,9 +6,6 @@
 ##' @param var.names vector names of state variable names.
 ##' @param X a matrix of state variables. In this matrix rows represent ensembles, while columns show the variables for different sites.
 ##' @param localization.FUN This is the function that performs the localization of the Pf matrix and it returns a localized matrix with the same dimensions.
-##' @param t not used
-##' @param blocked.dis passed to `localization.FUN`
-##' @param ... passed to `localization.FUN`
```
infotroph: This is undoing a fix I made in #3346 (and apparently forgot to update Rcheck_reference.log, sorry! That's why the checks didn't complain about this being undone.)

modules/assim.sequential/R/Multi_Site_Constructors.R
```diff
@@ -57,13 +54,13 @@ Contruct.Pf <- function(site.ids, var.names, X, localization.FUN=NULL, t=1, bloc
...
   labelss <- paste0(rep(var.names, length(site.ids)) %>% as.character(),"(",
-         rep(site.ids, each=length(var.names)),")") 
+                    rep(site.ids, each=length(var.names)),")") 
   
   colnames(pf.matrix.out ) <-labelss
   rownames(pf.matrix.out ) <-labelss
   
   return(pf.matrix.out)
-
+
```
infotroph:
```suggestion

```

modules/assim.sequential/R/Multi_Site_Constructors.R
```diff
@@ -27,15 +24,15 @@ Contruct.Pf <- function(site.ids, var.names, X, localization.FUN=NULL, t=1, bloc
...
+    #foreach site let's get the Xs
     pf.matrix [pos.in.matrix, pos.in.matrix] <- stats::cov( X [, pos.in.matrix] ,use="complete.obs")
   }
   
   # This is where we estimate the cov between state variables of different sites
   #I put this into a sperate loop so we can have more control over it
   site.cov.orders <- expand.grid(site.ids,site.ids) %>%
     dplyr::filter( .data$Var1 != .data$Var2)
-
+
```
infotroph:
```suggestion

```

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -187,14 +187,14 @@ GEF.MultiSite.Nimble <-  nimbleCode({
       # Sorting out qs
       q[1:YN, 1:YN] ~ dwish(R = aq[1:YN, 1:YN], df = bq) ## aq and bq are estimated over time
     }
-
+
```
infotroph:
```suggestion

```

modules/assim.sequential/R/Nimble_codes.R
```diff
@@ -187,14 +187,14 @@ GEF.MultiSite.Nimble <-  nimbleCode({
...
-
+    
     for (i in 1:nH) {
       tmpX[i]  <- X.mod[H[i]]
       Xs[i] <- tmpX[i]
     }
     ## add process error to x model but just for the state variables that we have data and H knows who
     X[1:YN]  ~ dmnorm(Xs[1:YN], prec = q[1:YN, 1:YN])
-
+
```
infotroph:
```suggestion

```

modules/assim.sequential/R/sda.batch.functions.R
```diff
@@ -0,0 +1,212 @@
...
+                                          "nimble",
+                                          "PEcAnAssimSequential"), 
+                              .options.snow=opts) %dopar% {
+                                MCMC_block_function(l)
+                              }
+  # wrap results.
+  parallel::stopCluster(cl)
+  writeLines("finished",con = file.path(folder.path, "result.txt"))
+  save(results, file = file.path(folder.path, "results.Rdata"))
+}
```
infotroph:
```suggestion
}

```

modules/assim.sequential/R/sda.enkf_NorthAmerica.R
```diff
@@ -0,0 +1,381 @@
...
+#' `forceRun` decide if we want to proceed the Bayesian MCMC sampling without observations;
+#' `MCMC.args` include lists for controling the MCMC sampling process (iteration, nchains, burnin, and nthin.).
+#'
+#’ @details
+#’ Restart mode:  Basic idea is that during a restart (primary case envisioned as an iterative forecast), a new workflow folder is created and the previous forecast for the start_time is copied over. During restart the initial run before the loop is skipped, with the info being populated from the previous run. The function then dives right into the first Analysis, then continues on like normal.
+#' 
+#' @description State Variable Data Assimilation: Ensemble Kalman Filter and Generalized ensemble filter. Check out SDA_control function for more details on the control arguments.
+#' 
+#' @return NONE
+#' @import nimble
```
infotroph: Can we importFrom just the functions we need instead of bringing all of nimble into the namespace?

modules/assim.sequential/R/Multi_Site_Constructors.R
infotroph: ...But if you have better descriptions for the parameters, please do improve my wording!

base/remote/R/check_qsub_status.R
DongchenZ: 1) I am unsure if I fully understand how `logger.setLevel()` works inside the `PEcAn.logger` package.
2) For a small amount of jobs, I think it will still be helpful to have job info printed instead of creating a progress bar.

modules/assim.sequential/R/Multi_Site_Constructors.R
DongchenZ: Fixed.

modules/assim.sequential/R/sda.enkf_NorthAmerica.R
DongchenZ: I removed it cause I found there is no place where `nimble` is applied.

mdietze:
@DongchenZ says that the changes here were superseded by the now committed PR #3544 and that this PR can be closed

PR #3345, opened 2024-07-27
[Integration Testing] PEcAn Docker Stack Run : Single Site, Single Model 

meetagrawal09:
## Description
PR aims at adding a workflow for running a complete PEcAn workflow using PEcAn's docker stack.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:


.github/workflows/model-run.yml
```diff
@@ -0,0 +1,59 @@
...
+      uses: actions/checkout@v4
+
+    - name: Set up Docker Compose
+      run: |
+          sudo apt-get update
+          sudo apt-get -y install docker-compose
+
+    - name: Build Containers
+      run: |
+        cp docker/env.example .env
```
infotroph: Since everything in env.example is commented out, we can skip the copy and create .env from scratch
```suggestion
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,59 @@
...
+  - cron: '30 4 * * 4'    
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+    
+    container: 
+      image: pecan/base:develop
```
infotroph: 👍 to hard-coding the tag for a first pass. Once it's working this way, we can consider taking it from the run context to allow e.g. testing against multiple R versions.

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Build Containers
+      run: |
+        cp docker/env.example .env
+        echo "COMPOSE_PROJECT_NAME=pecan" >> .env
+        echo "PECAN_VERSION=develop" >> .env
+        echo "UID=$(id -u)" >> .env
+        echo "GID=$(id -g)" >> .env
+        docker-compose up -d postgres
+        docker run --rm --network pecan_pecan pecan/db
```
infotroph: For the next person (me in a year 😄) who wonders "Wouldn't it make sense to write this as a separate step, like...?"
```
- name: db setup
   uses: docker://pecan/db:ci
```
...It might be worth leaving a comment explaining that no, when you run it in a separate step the `--network` option [isn't supported](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container#setting-container-resource-options)

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,59 @@
...
+        docker-compose run  bety user carya illinois "Carya Demo User" carya@example.com 1 1
+
+        docker run --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
+        docker run --rm --network pecan_pecan --volume pecan_pecan:/data pecan/data:develop chown -R "$(id -u).$(id -g)" /data
+
+        docker run --user="$(id -u)" --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
+    
+    - name: Run required containers
+      run: |
+        docker-compose up pecan bety sipnet rstudio -d
```
infotroph, line 50:
```suggestion
        docker-compose up -d pecan bety sipnet rstudio
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Check if containers are running # This will be updated soon to a loop
+      run: |
+        docker ps
+        sleep 50
```
infotroph: I see there [exists](https://docs.docker.com/reference/cli/docker/compose/up/#options) a `--wait` option to `docker compose up`, but haven't used it myself. Would that be able to handle this step for us?

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Run SIPNET
+      run: |
+        docker exec --workdir /pecan/tests pecan-rstudio-1 R CMD ../web/workflow.R --settings test.xml
```
infotroph, line 59: Should this be `docker compose exec ...`?

tests/test.xml
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0"?><pecan>
+  <outdir>/data/workflows/PEcAn_99000000021</outdir>
```
infotroph: Nit: I think it'd be clearer to name this in a way that's obviously for testing, rather than hardcode a workflow ID that won't have an equivalent in the database. `/tmp/docker_stack_tests/sipnet` or the like?

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,57 @@
...
+
+  # allow manual triggering
+  workflow_dispatch:
+
+  schedule:
+    # run Thursday 4:30 AM UTC
+  - cron: '30 4 * * 4'    
+jobs:
+  test:
+    runs-on: ubuntu-20.04
```
infotroph:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Build Containers
+      run: |
+        cp docker/env.example .env
+        echo "COMPOSE_PROJECT_NAME=pecan" >> .env
+        echo "PECAN_VERSION=develop" >> .env
+        echo "UID=$(id -u)" >> .env
+        echo "GID=$(id -g)" >> .env
+        
+        docker-compose up -d postgres
```
infotroph:
```suggestion
        docker compose up -d postgres
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,57 @@
...
+        docker run --rm --network pecan_pecan pecan/db
+        
+        docker-compose run bety user guestuser guestuser "Guest User" guestuser@example.com 4 4
+        docker-compose run bety user carya illinois "Carya Demo User" carya@example.com 1 1
```
infotroph: We can skip creating a guest user for headless runs
```suggestion
        docker compose run --rm bety user carya illinois "Carya Demo User" carya@example.com 1 1
```

.github/workflows/model-run.yml
infotroph: This reminds me that this line should include an explicit tag too -- the `db:ci` image is currently a little smaller than `db:latest`, but will soon be _much_ smaller (we're working on updating it to contain only the records we need for testing)
```suggestion
        docker run --rm --network pecan_pecan pecan/db:ci
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,57 @@
...
+
+        docker run --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
+        docker run --rm --network pecan_pecan --volume pecan_pecan:/data pecan/data:develop chown -R "$(id -u).$(id -g)" /data
+
+        docker run --user="$(id -u)" --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
+    
+    - name: Run required containers
+      run: |
+        docker-compose up pecan bety sipnet rstudio -d
```
infotroph: Assuming your goal is to start only the services we need:
```suggestion
        docker compose up -d sipnet executor
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    - name: Check if containers are running # This will be updated soon to a loop
+      run: |
+        docker ps
+        sleep 50
+
+    - name: Run SIPNET
+      run: |
+        docker exec --workdir /pecan/tests pecan-rstudio-1 R CMD ../web/workflow.R --settings test.xml
```
infotroph:
```suggestion
        docker compose exec --workdir /pecan/tests executor R CMD ../web/workflow.R --settings test.xml

```

.github/workflows/model-run.yml
infotroph: Also a note of caution for anyone testing this file: this step runs in the same Docker network as the desktop deployment, and WILL DELETE any existing pecan DB it finds there. On GitHub Actions you're starting from a clean VM, but if running locally with e.g. `act -W .github/workflows/model-run.yml -j test`, be prepared to lose your local DB contents.

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,57 @@
...
+        docker-compose up -d postgres
+        docker run --rm --network pecan_pecan pecan/db
+        
+        docker-compose run bety user guestuser guestuser "Guest User" guestuser@example.com 4 4
+        docker-compose run bety user carya illinois "Carya Demo User" carya@example.com 1 1
+
+        docker run --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
+        docker run --rm --network pecan_pecan --volume pecan_pecan:/data pecan/data:develop chown -R "$(id -u).$(id -g)" /data
+
+        docker run --user="$(id -u)" --rm --network pecan_pecan --volume pecan_pecan:/data --env FQDN=docker pecan/data:develop
```
infotroph: Do we need all three of these lines? Looks like the data wind up getting copied twice.

meetagrawal09: Good catch : ) Made the change.

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,56 @@
...
+  - cron: '30 4 * * 4'    
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+    - name: Checkout source code
```
robkooper: maybe add this:
```suggestion
    - name: Free Disk Space (Ubuntu)
      uses: jlumbroso/free-disk-space@main
      with:
        tool-cache: true
        android: true
        dotnet: true
        haskell: true
        large-packages: true
        docker-images: false
        swap-storage: true

      - name: Checkout source code
```

.github/workflows/model-run.yml
```diff
@@ -0,0 +1,72 @@
...
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
+      with:
+        # this might remove tools that are actually needed,
+        # if set to "true" but frees about 6 GB
+        tool-cache: false
```
robkooper: you can set this to true to remove 6 more GB

.github/workflows/docker-stack-sipnet.yml
```diff
@@ -0,0 +1,63 @@
+name : Docker Stack Test
```
infotroph:
```suggestion
name : Docker Stack Model Test
```

robkooper (approved):

.github/workflows/docker-stack-sipnet.yml
```diff
@@ -62,6 +62,10 @@ jobs:
         sleep 50
         docker ps
 
+    - name: Add models
+      run: |
+        docker compose exec --workdir /pecan/scripts executor sh add.models.sh
```
infotroph: Latest logs are showing that the executor image doesn't have psql installed.

There are probably better ways, but I think this should work:

```
docker compose cp scripts/add.util.sh postgres:/tmp/add.util.sh
docker compose cp scripts/add.models.sh postgres:/tmp/add.models.sh
docker compose exec postgres /tmp/add.models.sh
```

infotroph: Starting to suspect I'm sending you on a goose chase here. Hey @robkooper is this the right approach or should we instead be, like, starting the monitor service and waiting for it to register sipnet for us? If so what would that look like here?

allgandalf:
@meetagrawal09 what are we waiting for here? is this PR ready to get merged ?

meetagrawal09:
> @meetagrawal09 what are we waiting for here? is this PR ready to get merged ?

There is some issue while using rabbitmq. I see that a job is getting put in the queue but we are not able to poll the final status of this job. The Actions run just ends up in a timeout after waiting for completion for about ~6 hrs. 

[Example Run](https://github.com/PecanProject/pecan/actions/runs/13326409240/job/37220530132)

robkooper:
potential fix at: https://github.com/PecanProject/pecan/pull/3519

PR #3349, opened 2024-07-31
GEFS download fixes

infotroph:
## Description

Getting `download.NOAA_GEFS` working again, most notably by not scraping  two different webpages to extract dates that we can predict by knowing the forecasts are retained for four days. Best I can tell there have been some changes in the NOAA server configuration and the format of the grib files since this function last worked. 

See comments inline for my reasoning on each change, but I'd like a critical eye from someone more familiar with GEFS.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.atmosphere/NEWS.md
```diff
@@ -1,5 +1,12 @@
 # PEcAn.data.atmosphere 1.8.0.9000
 
+## Fixed
+* `download.NOAA_GEFS` is updated to work again with GEFS v12.3
+	(the current release as of this writing in July 2024).
```
infotroph:
```suggestion
* `download.NOAA_GEFS` is updated to work again with GEFS v12.3,
	the current release as of this writing in July 2024 (#3349).
```

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -17,21 +17,11 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
   
   
   download_grid <- function(ens_index, location, directory, hours_char, cycle, base_filename1, vars,working_directory){
-    #for(j in 1:31){
-    if(ens_index == 1){
-      base_filename2 <- paste0("gec00",".t",cycle,"z.pgrb2a.0p50.f")
-      curr_hours <- hours_char[hours <= 384]
```
infotroph, line 23: Most of my changes in this hunk are just to simplify the logic, but flagging that I can't figure out why this line was here at all: All ensemble members from a given forecast cycle have the same number of hours available, so my unconfident best guess is someone got confused between cycle ids and  ensemble ids. But cycle 00 is the one with _more_ hours, so this would have been wrong anyway.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -73,36 +63,11 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
   
   model_dir <- file.path(output_directory, model_name_raw)
   
+  #Availability: most recent 4 days
   curr_time <- lubridate::with_tz(Sys.time(), tzone = "UTC")
   curr_date <- lubridate::as_date(curr_time)
-  
-  noaa_page <- readLines('https://nomads.ncep.noaa.gov/pub/data/nccf/com/gens/prod/')
```
infotroph, line 79: These pages (here and on deleted line 91 below) load fine in a browser, but inside R and via curl they throw `Stream error in the HTTP/2 framing layer`. But all this scraping and extracting can be replaced by knowing that the forecasts are retained for four days -- am I missing cases where data posting isn't reliable enough to  assume "today and the three days before that"?

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -73,36 +63,11 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
-        potential_cycle <- c(potential_cycle, cycles)
-      }
-    }
-  }
-  
-  potential_dates <- lubridate::as_date(potential_dates)
-  
+  potential_dates <- curr_date - lubridate::days(3:0)
+
   potential_dates = potential_dates[which(potential_dates == forecast_date)]
```
infotroph: I think as written this always has length one and therefore the nested loops on lines 123 and 129 below aren't needed, but I elected not to mess with them in this edit.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -143,11 +111,11 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
         
         if(cycle == "00"){
-          hours <- c(seq(0, 240, 3),seq(246, 384, 6))
-          hours <- hours[hours<=end_hr]
+          hours <- c(seq(0, 240, 3),seq(246, 840, 6))
         }else{
-          hours <- c(seq(0, 240, 3),seq(246, min(end_hr, 840) , 6))
+          hours <- c(seq(0, 240, 3),seq(246, 384 , 6))
         }
+        hours <- hours[hours<=end_hr]
```
infotroph: The logic was backwards here -- cycle 00 is the longer one, not shorter.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -163,12 +131,12 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
-                           hours_char,
-                           cycle,
-                           base_filename1,
-                           vars,
+                           location = location,
+                           directory = directory,
+                           hours_char = hours_char,
+                           cycle = cycle,
+                           base_filename1 = base_filename1,
+                           vars = vars,
```
infotroph, line 84: Just for clarity

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -222,23 +193,13 @@ process_gridded_noaa_download <- function(lat_list,
...
-      if(ens_index-1 < 10){
-        ens_name <- paste0("0",ens_index-1)
-      }else{
-        ens_name <- as.character(ens_index-1)
-      }
-      base_filename2 <- paste0("gep",ens_name,".t",cycle,"z.pgrb2a.0p50.f")
-    }
+    ens_base <- if (ens_index == 1) { "gec" } else { "gep" }
+    ens_name <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
+    base_filename2 <- paste0(ens_base,ens_name,".t",cycle,"z.pgrb2a.0p50.f")
```
infotroph: Just for clarity -- this block was not producing errors

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -222,23 +193,13 @@ process_gridded_noaa_download <- function(lat_list,
...
     
     lats <- round(lat_list/.5)*.5
     lons <- round(lon_list/.5)*.5
     
-    if(lons < 0){
-      lons <- 360 + lons
-    }
```
infotroph, line 241: We're comparing these longitudes to the raw values in the grib file, which have negative values for west lon. I assume this must have changed at some point since these lines were written, but can't find any reference to it in the GEFS changelogs.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -263,8 +224,13 @@ process_gridded_noaa_download <- function(lat_list,
...
           if(curr_hours[hr] != "000"){
-            apcpsfc[s, hr]  <- grib_data_df$`SFC=Ground or water surface; 03 hr Total precipitation [kg/(m^2)]`[index]
-            tcdcclm[s, hr]  <- grib_data_df$`RESERVED(10) (Reserved); Total cloud cover [%]`[index]
+            # total precip alternates being named as 3 or 6 hr total
+            # TODO: not sure if the contents actually differ or if this is a labeling bug in the grib files
+            precip_hr <- if ((as.numeric(curr_hours[hr]) %% 2) == 1) { "03" } else { "06" }
+            precip_name <- paste("SFC=Ground or water surface;", precip_hr, "hr Total precipitation [kg/(m^2)]")
+            apcpsfc[s, hr]  <- grib_data_df[[precip_name]][index]
+
+            tcdcclm[s, hr]  <- grib_data_df$`EATM=Entire Atmosphere; Total cloud cover [%]`[index]
```
infotroph, line 235: This is weird and I don't understand why the grib files do it (or whether that means the interpretation of the precip totals should change from one time to the next), but I checked multiple sites and forecast days and the alternation is definitely there in the raw gribs.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -301,17 +267,15 @@ process_gridded_noaa_download <- function(lat_list,
   
   
   
-  cycle <-forecast_time
+  cycle <- forecast_time
```
infotroph, line 272: All changes from here to the bottom of this file are for clarity, not bugfixes.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -6,36 +6,43 @@
...
-##' This function downloads NOAA GEFS weather data.  GEFS is an ensemble of 21 different weather forecast models.  
-##' A 16 day forecast is avaliable every 6 hours.  Each forecast includes information on a total of 8 variables.  
-##' These are transformed from the NOAA standard to the internal PEcAn
-##' standard.
-##' 
-##' @section Data Avaliability:
-##' NOAA GEFS weather data is avaliable on a rolling 12 day basis; dates provided in "start_date" must be within this range. The end date can be any point after
-##' that, but if the end date is beyond 16 days, only 16 days worth of forecast are recorded.  Times are rounded down to the previous 6 hour forecast.  NOAA
-##' GEFS weather data isn't always posted immediately, and to compensate, this function adjusts requests made in the last two hours
-##' back two hours (approximately the amount of time it takes to post the data) to make sure the most current forecast is used.
```
infotroph, line 18: I do not think this two-hour adjustment was being done in the current code, and am sure it isn't being done after this PR. Deleted it from the docs.

modules/data.atmosphere/R/download.NOAA_GEFS.R
```diff
@@ -50,8 +57,6 @@
 ##' @author Quinn Thomas, modified by K Zarada 
 ##' 
 download.NOAA_GEFS <- function(site_id,
-                               sitename = NULL,
-                               username = 'pecan',
```
infotroph, line 54: These are not used by the function, so dropping

modules/data.atmosphere/R/half_hour_downscale.R
```diff
@@ -223,8 +223,16 @@ downscale_ShortWave_to_half_hrly <- function(df,lat, lon, hr = 0.5){
   
   for (k in 1:nrow(data.hrly)) {
     if(is.na(data.hrly$surface_downwelling_shortwave_flux_in_air[k])){
-      SWflux <- as.matrix(subset(df, .data$day == data.hrly$day[k] & .data$hour == data.hrly$hour[k], data.hrly$surface_downwelling_shortwave_flux_in_air[k]))
```
infotroph, line 226: I'm guessing a bit on what this was supposed to be, because it could not have worked as written -- not only is `.data` invalid inside base::subset (it's specific to tidy evaluation)  , but I'm not sure what subset would do with a bare number for its third argument. 

My replacement below seems reasonable if the goal of this step is "when SWflux is null in the hourly data, take it from the matching day and hour of the coarser-scale data", but please speak up if you see a better fix.

modules/data.atmosphere/R/half_hour_downscale.R
```diff
@@ -284,11 +292,9 @@ downscale_repeat_6hr_to_half_hrly <- function(df, varName, hr = 0.5){
     #previous 6hr period
     dplyr::mutate(lead_var = dplyr::lead(df[,varName], 1))
   #check for NA values and gapfill using closest timestep
-  for(k in 1:dim(df)[1]){
+  for(k in 2:dim(df)[1]){
```
infotroph:
```suggestion
  for(k in 1:dim(df)[1]){
```

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph:
```suggestion
  potential_dates <- potential_dates[which(potential_dates == forecast_date)]
```

modules/data.atmosphere/R/GEFS_helper_functions.R
mdietze: I actually think there are some ensemble members that are the full 35 days, and others that are shorter, even for cycle 00

modules/data.atmosphere/R/GEFS_helper_functions.R
mdietze: My experience is there can be delays in when files show up and occasionally gaps, which is why we didn't just assume forecasts are always there.

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph: Can you point to any examples of the shorter members, or documentation that mentions them? I've been poking around and haven't yet found any yet, but if this is an intermittent thing it'd be easy for me to have missed it.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -73,37 +63,12 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
-  
-  last_cycle_page <- readLines(paste0('https://nomads.ncep.noaa.gov/pub/data/nccf/com/gens/prod/gefs.', dplyr::last(potential_dates)))
-  
-  potential_cycle <- NULL
-  for(i in 1:length(last_cycle_page)){
-    if(stringr::str_detect(last_cycle_page[i], 'href=\"')){
-      end <- stringr::str_locate(last_cycle_page[i], 'href=\"')[2]
-      cycles <- stringr::str_sub(last_cycle_page[i], start = end+1, end = end+2)
-      if(cycles %in% c("00","06", "12", "18")){
-        potential_cycle <- c(potential_cycle, cycles)
```
infotroph, line 99: This variable was never used below, so this entire URL request and loop had no effect.

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph: Note also the change in cloud cover name from "RESERVED..." to "EATM..." -- I would have expected this to show up in a changelog somewhere as well, but haven't been able to find any references to it.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -73,37 +63,12 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
-      }
-    }
-  }
```
infotroph, line 102: `potential_cycle` was never used below, so this entire URL request and loop had no effect.

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -17,21 +17,11 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
-        ens_name <- as.character(ens_index -1)
-      }
-      base_filename2 <- paste0("gep",ens_name,".t",cycle,"z.pgrb2a.0p50.f")
-      curr_hours <- hours_char
-    }
-    
-    
+    ens_base <- if (ens_index == 1) { "gec" } else { "gep" }
+    ens_name <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
+    base_filename2 <- paste0(ens_base,ens_name,".t",cycle,"z.pgrb2a.0p50.f")
```
infotroph:
```suggestion
    ens_idx <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
    base_filename2 <- paste0(ens_base,ens_idx,".t",cycle,"z.pgrb2a.0p50.f")
```

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph: That's good to know, but note that the existing code only checked whether there exists a date folder that matches the requested forecast date, not whether all forecasts (or even all cycles) are present in it. How often do delays/gaps occur at the whole-day level?

Each individual file download (line 50 above) is wrapped in a tryCatch, so for both old and new code the likeliest outcome when files are missing will be a series of "skipping" messages (possibly then followed by a failure when downscaling can't bridge the time gap). Is that acceptable for this PR or do we need to dig further in on pre-download availability checks?

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -17,21 +17,11 @@ noaa_grid_download <- function(lat_list, lon_list, forecast_time, forecast_date,
...
-        ens_name <- as.character(ens_index -1)
-      }
-      base_filename2 <- paste0("gep",ens_name,".t",cycle,"z.pgrb2a.0p50.f")
-      curr_hours <- hours_char
-    }
-    
-    
+    ens_base <- if (ens_index == 1) { "gec" } else { "gep" }
+    ens_idx <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
+    base_filename2 <- paste0(ens_base,ens_idx,".t",cycle,"z.pgrb2a.0p50.f")
```
infotroph:
```suggestion
    member_type <- if (ens_index == 1) { "gec" } else { "gep" } # "_c_ontrol", "_p_erturbed"
    ens_idxname <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
    base_filename2 <- paste0(member_type,ens_idxname,".t",cycle,"z.pgrb2a.0p50.f")
```

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph:
```suggestion
    member_type <- if (ens_index == 1) { "gec" } else { "gep" } # "_c_ontrol", "_p_erturbed"
    ens_idxname <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
    base_filename2 <- paste0(member_type,ens_idxname,".t",cycle,"z.pgrb2a.0p50.f")
```

modules/data.atmosphere/R/GEFS_helper_functions.R
```diff
@@ -222,23 +193,13 @@ process_gridded_noaa_download <- function(lat_list,
...
-        ens_name <- as.character(ens_index-1)
-      }
-      base_filename2 <- paste0("gep",ens_name,".t",cycle,"z.pgrb2a.0p50.f")
-    }
+    member_type <- if (ens_index == 1) { "gec" } else { "gep" } # "_c_ontrol", "_p_erturbed"
+    ens_idxname <- stringr::str_pad(ens_index - 1, width = 2, pad = "0")
+    base_filename2 <- paste0(member_type,ens_idxname,".t",cycle,"z.pgrb2a.0p50.f")
     
     lats <- round(lat_list/.5)*.5
     lons <- round(lon_list/.5)*.5
```
infotroph, line 203: Not for this PR, but: GEFS is now also [available](https://www.nco.ncep.noaa.gov/pmb/products/gens/gec00.t00z.pgrb2s.0p25.f000.shtml) on a 0.25 degree grid. Would it be of interest to add support for that?

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph: Checking back on this discussion: @mdietze is it possible the time differences you're remembering were from an older release of GEFS? I'm still not seeing any in the current system.

Over the last ~week I've reviewed the list of available files in https://nomads.ncep.noaa.gov/pub/data/nccf/com/gens/prod/ daily at varying times, and run test downloads on at least three different days. In all cases I've seen that individual files are sporadically missing, but overall there are nearly always 181 files from each of 31 ensemble members in cycle 0 and 105 from each in cycles 6, 12, 18. So I'm confident there's no _systematic difference in what files exist on the server_  between control (id 1) and perturbed (id 2-31) ensemble members.

But! If instead the deleted code here was intended to _skip using_ a file that does exist, then my research won't have detected that.

modules/data.atmosphere/R/GEFS_helper_functions.R
infotroph: @mdietze Checking back in on my last point here -- is the existing tryCatch around download of individual grib files sufficient handling of delayed file availability? 

If I've read right that the code deleted here was only checking for the existence of a folder for a given date+cycle and not for full sets of files inside it, I don't think those checks were making much differences in the reliability of the process.

infotroph:
Changes since last review:

* Added unit tests, including both download from live server (skipped on CI) and interface verification with `download.file` mocked out so we don't thrash the NOAA server: bf096e2595eb1f1c584fd95d18cf28bf8ebdb057
* Further whitespace cleanup: 08928f664ea150cac04bf4d46193e04961bb4f99
* Simplified code without changing logic: ac30fddec038c98f0a35d51073b9a3223fc7fb8d, 65ea6457d70add60d2a69f3050ab8f300557bb01
* Moved `download_grid`, previously nested inside `noaa_grid_download`, out to a top-level but unexported function definition and added Roxygen doc for params: 756b95250c18da9b4c2df3942481923a3af26eae

mdietze (approved):

PR #3352, opened 2024-08-03
make a named volume for the Bety logs mountpoint

infotroph:
Best I can tell nothing gets written to this, but because the Bety image [specifies](https://github.com/PecanProject/bety/blob/develop/Dockerfile#L65) this path as a VOLUME, we were creating a new anonymous-but-permanent volume every time Compose spun up. Having one named one that stays empty seems a touch cleaner.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


robkooper (approved):
probably should add this to the bety repo as well, just to be consistent.

mdietze (approved):

infotroph:
> probably should add this to the bety repo as well, just to be consistent.

👍 Done in https://github.com/PecanProject/bety/pull/762

PR #3354, opened 2024-08-07
start new docker build

robkooper:
update build process
- fixes ed

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


.github/workflows/docker.yml
```diff
@@ -1,136 +1,332 @@
...
+          #push: true
+          platforms: "linux/amd64,linux/arm64"
+          # cache-from: type=registry,ref=classtranscribe/frontend-pipeline:buildcache
+          # cache-to: type=registry,ref=classtranscribe/frontend:buildcache,mode=max
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          outputs: type=image,dest=/tmp/depends.tar
+          build-args: |
+            VERSION=${{ steps.meta.outputs.version }}
+            R_VERSION=${{ github.event.inputs.r_version}}
```
infotroph:
```suggestion
            R_VERSION=${{ github.event.inputs.r_version }}
```

docker/base/Dockerfile
```diff
@@ -26,9 +25,7 @@ COPY models   /pecan/models/
 # install all PEcAn packages
 # `make clean` is to remove artifacts copied in from host system
 #   (e.g. basgra.so)
-RUN --mount=type=secret,id=github_token \
-    export GITHUB_PAT=`cat /run/secrets/github_token` \
```
infotroph, line 30: I'm glad to see this go, but I don't immediately see what replaces it -- are github tokens no longer needed inside this image, or is there now a cleaner way of passing them around?

infotroph: (Or am I asking too soon about a temporary change?)

robkooper: I have not needed the GITHUB_PAT, lets see what happens. I'll create an issue so we can keep track in case we need it.

docker/web/Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM php:8-apache
-MAINTAINER Rob Kooper <kooper@illinois.edu>
```
infotroph, line 2: I bet you considered this already, but to confirm: Is deleting better than replacing with `LABEL org.opencontainers.image.authors="kooper@illinois.edu"`?

.github/workflows/docker.yml
```diff
@@ -1,136 +1,615 @@
...
 on:
-  push:
-    branches:
-      - master
-      - develop
+  # push:
+  #   branches:
+  #     - main
+  #   tags:
+  #     - "v*.*.*"
```
infotroph: Remove or uncomment?

.github/workflows/docker.yml
```diff
@@ -1,136 +1,615 @@
...
+  # ----------------------------------------------------------------------
+  # base image has PEcAn compiled and installed, and depends on depends
+  # ----------------------------------------------------------------------
+  base:
+    runs-on: ubuntu-latest
+    needs: depends
+    permissions:
+      packages: write
+
+    steps:
```
infotroph, line 117: Each image's job has steps that are identical or close to it, correct? If so, I think we could likely save a fair amount of duplication by encapsulating these steps in a called workflow (see e.g. how check.yml/test.yml/sipnet.yml are used by ci.yml and ci-weekly.yml). 

(Probably just as easily done in a separate PR, so no need to hold this one up)

docker/base/Dockerfile
infotroph: And just to check: docker.depends/Dockerfile still contains the same `mount=type=secret` dance; is that intentional or should it go too?

models/ed/Dockerfile
```diff
@@ -10,27 +10,23 @@ FROM pecan/models:${IMAGE_VERSION} as model-binary
 ARG MODEL_VERSION="2.2.0"
 ARG BINARY_VERSION="2.2"
 
-# specify fortran compiler
```
infotroph, line 13: s/as/AS/ on line 7 above

docker-compose.yml
```diff
@@ -65,6 +70,11 @@ services:
       - pecan
     volumes:
       - postgres:/var/lib/postgresql/data
+    healthcheck:
+      test: pg_isready -U postgres
```
infotroph, line 74: 👍

.github/workflows/docker.yml
```diff
@@ -1,136 +1,615 @@
...
+        default: true
+      r_version:
+        description: 'R version to use'
+        required: true
+        type: choice
+        default: "4.1"
+        options:
+          - 4.1
+          - 4.2
+          - 4.3
```
infotroph, line 22: Should devel (currently built weekly) and 4.4 (currently built daily) be added here? 

Is this approach enough faster that we no longer need to limit some builds to weekly?

.github/workflows/docker.yml
```diff
@@ -1,136 +1,615 @@
...
+  R_VERSION: ${{ github.event.inputs.r_version || '4.1' }}
 
+# there are 3 jobs to build different images
 jobs:
-  docker:
-    if: github.event_name != 'issue_comment' || startsWith(github.event.comment.body, '/build')
+  # ----------------------------------------------------------------------
+  # depends image has all the dependencies installed
+  # ----------------------------------------------------------------------
+  depends:
```
infotroph, line 35: Am I right this step will completely replace depends.yml? How promptly? (I'm guessing we should delete it as soon as this is in production and confirmed to be behaving right?)

.github/workflows/docker.yml
```diff
@@ -1,136 +1,616 @@
...
+          - 4.2
+          - 4.3
 
-# Certain actions will only run when this is the master repo.
+# set up the environment, either input or default
 env:
-  MASTER_REPO: PecanProject/pecan
-  DOCKERHUB_ORG: pecan
-  GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+  R_VERSION: ${{ github.event.inputs.r_version || '4.1' }}
```
infotroph, line 28: Is the `|| '4.1'` needed here? I was assuming the `default: "4.1"` on line 21 would ensure a value, but haven't tested my assumption.

infotroph (approved):

docker/base/Dockerfile
robkooper: that should go as well 🤞

docker/web/Dockerfile
infotroph: Conclusion from further discussion in Slack: Yes, the existing MAINTAINER tags don't really line up well with who is actually maintaining each package, so better to delete them now and add more informative ones as time allows.

.github/workflows/docker.yml
infotroph: Saving this for later as #3374

.github/workflows/docker.yml
infotroph:
```suggestion
          - 4.3
          - 4.4
          - devel
```

mdietze: Is this extending the set of R versions for just the Docker build, or for all the GH Actions? The other actions are currently running 4.1 and 4.2 with the former being the required version. If this is an overall change, it's outside the scope of the PR. If this is a Docker specific change, then I don't think we should be pushing a bunch of untested versions.

.github/workflows/docker.yml
robkooper: correct, this will automatically build a new depends image for all builds. Once this is merged in, and we have a few builds going it would be good to see how well the caching works as well.

.github/workflows/docker.yml
robkooper: the 4.1 is needed because the inputs are ONLY set when using a dispatch. :(

.github/workflows/docker.yml
robkooper: this is just for the docker build. The nice thing is that the default is 4.1 but you can either use a curl or from a pull down test with others as well. Once we decide to change the default we can do that here as well, and when we want test with all other versions.

.github/workflows/docker.yml
robkooper: Also this will remove the need to do a /build you can now trigger it manually

PR #3357, opened 2024-08-13
#GSOC PR: Enhance CNN Model with L2 Regularization, K-Fold CV, and Ensemble Techniques

sambhavnoobcoder:
## Description

This PR enhances the CNN model in the SDA_downscale function with L2 Regularization, K-Fold Cross-Validation, and Ensemble techniques. Key improvements include:

- L2 Regularization added to dense layers to prevent overfitting.
- K-Fold Cross-Validation (k=5) implemented for robust model evaluation.
- Ensemble methods introduced with bagging (5 models per fold).
- CNN architecture refined with dual batch normalization and dropout layers.
- Enhanced performance evaluation with per-fold metrics and final test set evaluation.

These changes aim to improve model robustness, reduce overfitting, and increase prediction accuracy for environmental data downscaling.

## Motivation and Context

These enhancements address common challenges in machine learning, particularly for complex environmental data. They provide a more reliable framework for downscaling, improving generalization to unseen data and stability in training.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/assim.sequential/R/downscale_function.R
```diff
@@ -140,84 +140,141 @@ SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_typ
...
+        val_indices <- fold_indices[[fold]]
+        
+        x_train_fold <- x_train[train_indices, , drop = FALSE]
+        y_train_fold <- y_train[train_indices, i]
+        x_val_fold <- x_train[val_indices, , drop = FALSE]
+        y_val_fold <- y_train[val_indices, i]
+        
+        # Create bagged models for this fold
+        fold_models <- list()
+        for (bag in 1:num_bags) {
```
mdietze, line 204: Unless I'm misunderstanding how things are done below, this looks like a place where the folds could in theory be done in parallel, rather than sequentially. You might ping @DongchenZ as he'd been looking closely recently at the different parallelization libraries in R and their relative efficiency and memory overhead.

modules/assim.sequential/R/downscale_function.R
```diff
@@ -140,84 +140,141 @@ SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_typ
...
+      all_models <- list()
       
-      # Learning rate scheduler
-      lr_schedule <- keras3::learning_rate_schedule_exponential_decay(
-        initial_learning_rate = 0.001,
-        decay_steps = 1000,
-        decay_rate = 0.9
-      )
+      # Create k-fold indices
+      fold_indices <- caret::createFolds(y = 1:nrow(x_train), k = k_folds, list = TRUE, returnTrain = FALSE)
```
mdietze: This introduces a new dependency for what is actually appears to be a fairly straightforward task. Are there analogous functions you could use from packages that are already in the dependencies? Could you do this manually with base functions like `sample` and/or a bit of tidyverse?  If the answers to this are no, you need to add the new package to the dependencies and update the PR.

dlebauer: FWIW the 'tidyverse' equivalent would be `rsample::vfold_cv`, but still might not be worth adding this dependency since implementation "should be" 🤞🏻 simple, like:

```
# <incomplete function documentation>
# @param i vector of indices created by seq_along(x)
# @param k number of groups
# based on https://stackoverflow.com/a/57085061/199217
create_folds <- function(i, k) {
  indices <- sample(i)
  folds <- split(indices, cut(indices, breaks = k, labels = FALSE))
  return(folds)
}
```

Ok it wasn't as simple as I hoped, but that should work!

modules/assim.sequential/R/downscale_function.R
```diff
@@ -140,84 +140,141 @@ SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_typ
...
-        monitor = 'val_loss',
-        patience = 10,
-        restore_best_weights = TRUE
-      )
+      #initialise operations for each fold
+      for (fold in 1:k_folds) {
+        cat(sprintf("Processing ensemble %d, fold %d of %d\n", i, fold, k_folds))
+        
+        # Split data into training and validation sets for this fold
+        train_indices <- setdiff(1:nrow(x_train), fold_indices[[fold]])
```
dlebauer: to improve code robustness, replace uses of `1:length(x)` with `seq_along(x)`

modules/assim.sequential/R/downscale_function.R
```diff
@@ -140,84 +140,141 @@ SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_typ
...
-      
-      # Early stopping callback
-      early_stopping <- keras3::callback_early_stopping(
-        monitor = 'val_loss',
-        patience = 10,
-        restore_best_weights = TRUE
-      )
+      #initialise operations for each fold
+      for (fold in 1:k_folds) {
+        cat(sprintf("Processing ensemble %d, fold %d of %d\n", i, fold, k_folds))
```
dlebauer, line 191: I like the idea of providing updates on where we are. But do we want perhaps 100s of messages? This might be a good place to use the progress bar instead (see e.g. https://github.com/PecanProject/pecan/blob/325d7fc32fd38ebe53ce5ee81262dc307202daf5/modules/emulator/R/predict.GP.R#L103) 

Still it is a great idea to provide messages that contain useful information about the progress. If that is the case here, I think using the PEcAn logger functions would be preferred to `cat`, e.g.

```
logger.info(sprintf("Processing ensemble %d, fold %d of %d", i, fold, k_folds))
```

(though as mentioned above, this might generate too many messages to be useful, could just print once per fold? And if there is some useful information that could be provided when processing of each fold, perhaps add another logger.info when the fold is finished?

dlebauer (commented):
Left a few comments as an initial review, mostly on code style.

modules/assim.sequential/R/downscale_function.R
sambhavnoobcoder: Addressed in commit [bf02478](https://github.com/PecanProject/pecan/pull/3357/commits/bf024780296e8c05680ad19b9abaa3d61c0b4611)

modules/assim.sequential/R/downscale_function.R
sambhavnoobcoder: Addressed in the commits [4ae1974](https://github.com/PecanProject/pecan/pull/3357/commits/4ae1974d7aa78fa27e188d222e85ab0eba14b773) , [5d86312](https://github.com/PecanProject/pecan/pull/3357/commits/5d86312d92fe03adeb3a77665b96db4ec2dcda45) and [70bfdaf](https://github.com/PecanProject/pecan/pull/3357/commits/70bfdaf5ac40a69309d78b063c28d41576d5603d)

PR #3366, opened 2024-09-02
[build] avoid recursive installation of a zillion indirect Suggests

infotroph:
## Description
Avoids an undocumented behavior/possible bug in `remotes::install_deps`, which correctly handles `dependencies = TRUE` (by installing mandatory + optional dependencies of your package, but only mandatory dependencies of the dependencies), but the supposedly equivalent `dependencies = c("Depends", "Imports", "LinkingTo", "Suggests")` will recursively install Suggests of Suggests of Suggests.

Tested with `docker run --rm -it -v ${PWD}:/work -w /work rocker/tidyverse:4.4 ./scripts/confirm_deps.R models/stics TRUE`

<details><summary>Output before</summary>

```
Downloading GitHub repo SticsRPacks/SticsRFiles@HEAD
Installing 12 packages: lazyeval, hunspell, renv, markdown, checkmate, rex, spelling, learnr, formatR, covr, xslt, XML
Installing packages into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
Warning: dependencies ‘marray’, ‘affy’, ‘Biobase’, ‘limma’, ‘cmdstanr’, ‘glmmADMB’, ‘rnaturalearthhires’, ‘graph’, ‘ComplexHeatmap’, ‘starsdata’, ‘subselect’, ‘INLA’, ‘globaltest’, ‘sylly.de’, ‘sylly.es’, ‘M3C’, ‘Rgraphviz’, ‘gurobi’, ‘rrelaxiv’, ‘Biostrings’, ‘seqLogo’, ‘spDataLarge’, ‘biomaRt’, ‘KEGGgraph’, ‘Rcampdf’, ‘tm.lexicon.GeneralInquirer’, ‘ExactData’, ‘rtracklayer’, ‘GenomicRanges’, ‘IRanges’, ‘S4Vectors’, ‘RandomFields’, ‘bold’, ‘Icens’, ‘loe’, ‘cairoDevice’, ‘RGtk2’, ‘rhdf5’, ‘RDCOMClient’, ‘genefilter’, ‘sva’, ‘mixOmics’, ‘taxidata’, ‘aods3’, ‘RTCGA.rnaseq’, ‘snpStats’ are not available
[...]

PR #3367, opened 2024-09-02
WIP: use pak instead of confirm_deps for dependency installation

infotroph:
## Description

EXPERIMENTAL, NEEDS TESTING: Use pak::local_install_deps() instead of our confirm_deps script. 

Before merging, need to show that local_install_deps (1) behaves the same as confirm_deps with respect to installation behavior, and (2) does not query overeagerly for updates to dependencies. Recall that the original impetus for `confirm_deps` was `devtools::install_deps` making too many queries to remote repos that we'd already checked earlier in the process. My read of the documentation suggests we should _expect_ pak to be better about this, but I have not yet checked. Probably the easiest way to check would be by running the install process inside a container instrumented to monitor outgoing IP requests.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,8 +70,10 @@ files_in_dir = $(call drop_parents, $(call recurse_dir, $1))
 # HACK: NA vs TRUE switch on dependencies argument is an ugly workaround for
 # a circular dependency between benchmark and data.land.
 # When this is fixed, can go back to simple `dependencies = TRUE`
-depends_R_pkg = ./scripts/time.sh "depends ${1}" ./scripts/confirm_deps.R ${1} \
-	$(if $(findstring modules/benchmark,$(1)),NA,TRUE)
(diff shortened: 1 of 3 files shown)


infotroph:
When testing: Check whether pak retains the same behavior as `remotes::install_deps` with respect to choosing a CRAN repository, viz. 'try repositories in the order they are specified and stop when the version requirement is satisfied, even if a newer version is available from a repo later in the list'. We rely very heavily on this behavior to avoid version hell in the older rocker images.

infotroph:
Closing this until I have a chance to actually test it -- #3366 seems to have fixed the biggest current pain point in confirm_deps, so I'm no longer in as much of a hurry to replace it.

PR #3373, opened 2024-09-06
Move datapack from Depends to Suggests in data.land

dlebauer:
## Description

This reverts one of the changes from https://github.com/PecanProject/pecan/pull/2670 moving datapack from Imports to Suggests. As recommended by @infotroph in that PR:

> my advice these days is to just put things into Imports unless there's a specific reason to move them to Suggests.

A [search of the codebase](https://github.com/search?q=repo%3APecanProject%2Fpecan%20datapack&type=code
) shows the datapack package is only used once in the data.land package, and it is used directly following a call to the dataone package in the `dataone_download` function https://github.com/PecanProject/pecan/blob/1b181347a2b632f7a53bac4976777d1dea587483/modules/data.land/R/dataone_download.R#L39

Rationale for change:
- it is only used in one very specialized function 
- The dataone package is in Suggests so by symmetry it seems datapack should / could also be in Suggests 
 
## Motivation and Context

Proposed by BU Research Computing Services in Slack

Motivating discussion:

> Hello.  I am part of Research Computing Services at Boston University.  I am working with a researcher that is trying to do a native install of PEcAn on our compute cluster.  The instructions we are following are linked below:
https://pecanproject.github.io/pecan-documentation/v1.7.1/pecan-manual-setup.html#install-native
For step 3, when we run `make`, we encounter the error that the `datapack` R library failed to install because the redlands library is missing on our system.  I looked up the source code for redlands library and it seems it hasn't been updated since 2014.  Does PEcAn still use this library?
https://download.librdf.org/source/
A few months ago, I worked with another researcher, that was attempting to install the "data.land" module and at that point we also ran into the same issue.  While investigating that issue I noticed that the DESCRIPTION file (linked below) for "data.land" module had marked "redlands" R package as optional and "datapack" as required. Since "datapack" requires the R "redlands" package, there is an overriding dependency happening here.  So this suggests maybe redland is no longer required?
https://github.com/PecanProject/pecan/blob/develop/modules/data.land/DESCRIPTION
We did resolve the issue by making "datapack" as an optional install and it seemed that solution was satisfactory for the researcher.
So before I put in the effort to try and install redland on our cluster, I am wondering if the R package "datapack" is required for the installation of PEcAn? Or could it be marked as optional

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

_I will do this if implemented_


dlebauer:
- [x] Update News.md
- [x] Need to update pecan_package_dependencies.csv https://github.com/PecanProject/pecan/actions/runs/10742976918/job/29796748098?pr=3373#step:8:15 
   - implemented manually, should this be done w/ https://github.com/PecanProject/pecan/blob/develop/scripts/generate_dependencies.R?

infotroph:
> dataone package is in Suggests so by symmetry it seems datapack should / could also be in Suggests

I fully endorse this logic 👍

infotroph:
Current test failure is from brittle string-matching in our ignore-existing-check-messages system:

```
  R check of modules/data.land returned new problems:
  checking package dependencies ... NOTE: Imports includes 33 non-default packages.
```

To fix, edit s/34/33/ on line 16 of modules/data.land/tests/Rcheck_reference.log

infotroph (changes requested):
> implemented manually, should this be done w/ generate_dependencies.R?

Yes, it should. In this case your result is _probably_ identical, but pls use the script

> - [x] update Changelog

I think this is a modules/data.land/NEWS.md change but not a CHANGELOG.md change.

dlebauer:
NEWS.md is empty for v 1.8.0.9000, which is the version in DESCRIPTION. Should I add information there? 

At the repository CHANGELOG.md level, the section for adding changes is [Unreleased]. Is there an intentional discrepancy in these protocols? Or should I create a new section for Unreleased changes? Should this protocol be documented?

infotroph:
> NEWS.md is empty for v 1.8.0.9000, which is the version in DESCRIPTION. Should I add information there?

Yep, please

> At the repository CHANGELOG.md level, the section for adding changes is [Unreleased]. Is there an intentional discrepancy in these protocols? Or should I create a new section for Unreleased changes? Should this protocol be documented?

[unreleased] is what we've been using in CHANGELOG for as long as I've been on the project; x.y.z.9000 is the convention I see in the NEWS files of other R packages that use the .9000 = dev version convention.  

In either case it seems fairly transparent to me that in both files the basic  convention is "add new changes at the top, then at release time we label them with the correct version"

infotroph (changes requested):
One more piece I should have asked for to start with: Need to add an if to make the dataone_download function throw a useful error if datapack isn't installed. @dlebauer unless you object, I'll make that change and push it straight to this branch.

dlebauer:
NEWS.md updated

dlebauer:
> fairly transparent to me 

Between different filenames and conventions it may not be as clear to others - it's still not clear to me when we should update news vs changelog or both

dlebauer:
>  add an if to make the dataone_download function throw a useful error

Go for it, thanks!

infotroph (approved):

PR #3375, opened 2024-09-08
rename master to main/latest

robkooper:
This wll go through and setup the code to allow us to rename master branch to main. This is to match our branch names with many of the other github repositories.

github.io == latest
github.com == main

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


mdietze (approved):

robkooper:
once this is merged we can rename master to main

.github/workflows/depends.yml
```diff
@@ -1,102 +0,0 @@
...
-
-on:
-  push:
-    branches:
-      - develop
-      - master
-  workflow_dispatch:
-
-  # this runs on the develop branch
-  schedule:
```
infotroph, line 11: Confirming my understanding: Am I correct that the new Docker build only _routinely_ builds the depends image for R 4.1, builds other R versions only when manually triggered, and does not have an equivalent of these scheduled builds? 

If so: Note that once we delete this, the weekly checks (Rdevel etc) will be relying on images that are now only manually updated. I am OK with that as a temporary state -- seems wise get the new system settled in before adding  more versions to the scheduled steps.

infotroph (approved):
This looks great, thanks.

I do find that with three branches (develop, main, latest) I get confused about the relationships between them a lot more than when considering any two of those, and it's unintuitive that the same branch is `main` in some places but `latest` in others. I vaguely remember this being needed because of something to do with how github.io handles default branches, right? Would it be worth documenting the reasons behind that somewhere in our docs?

.github/workflows/depends.yml
robkooper: correct, goal is to have those build on a schedule like before.

robkooper: What I wanted to do was use "latest" for the documentation so that it will use that in https://github.com/PecanProject/pecan-documentation

I did notice that the documentation at https://pecanproject.github.io/documentation does not match the documentation in that repository.

dlebauer:
> once this is merged we can rename master to main

So should we rename now?

PR #3377, opened 2024-09-11
[docker] health check fixes

infotroph:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


robkooper (approved):
nice catch

dlebauer (approved):
are there any guidelines for when we should use the `--no-install-recommends` flag for `apt-get`? It seems to be used inconsistently (~50:50) in our Docker images.

robkooper:
I try to use it as much as possible. The recommended packages are often good for normal use, but we want to try and keep the image as small as possible, hence the `--no-recommends`

Issue #3392, opened 2024-10-21
VignetteBuilder: field of DESCRIPTION should declare all builder deps


infotroph:
From ["Writing R Extensions"](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-DESCRIPTION-file-1) (emphasis theirs):

> The ‘VignetteBuilder’ field names (in a comma-separated list) packages that provide an engine for building vignettes. These may include the current package, or ones listed in ‘Depends’, ‘Suggests’ or ‘Imports’. The utils package is always implicitly appended. See Section 1.4.2 [Non-Sweave vignettes], page 48, for details. Note that if, for example, a vignette has engine ‘knitr::rmarkdown’, then knitr (https://CRAN.R-project.org/package=knitr) provides the engine but both knitr and rmarkdown (https://CRAN.R-project.org/package=rmarkdown) are needed for using it, so _both_ these packages need to be in the ‘VignetteBuilder’ field and at least suggested (as rmarkdown is only suggested by knitr, and hence not available automatically along with it). Many packages using knitr (https://CRAN.R-project.org/package=knitr) also need the package formatR (https://CRAN.R-project.org/package=formatR) which it suggests and so the user package needs to do so too and include this in ‘VignetteBuilder’.

Since all our vignettes use `knitr::rmarkdown`, my takeaway is that we need to replace `VignetteBuilder: knitr` with `VignetteBuilder: knitr, rmarkdown` in all packages that have vignettes.

harshagr70:
Hi @infotroph ! I’d like to work on this issue if it’s still available. Could you please assign it to me? I’m ready to get started on a fix. Thanks!

harshagr70:
Hi @infotroph  I've opened a PR #3394 to address this issue. I believe it addresses the problem, but I'd welcome any feedback or suggestions for improvement. Thanks!

infotroph:
Fixed by #3394. Thank you, @harshagr70!

PR #3394, opened 2024-10-28
Updated VignetteBuilder field to include knitr and rmarkdown

harshagr70:
This PR updates the DESCRIPTION file to include both knitr and rmarkdown in the VignetteBuilder field. This change ensures that all necessary dependencies for building vignettes are explicitly declared.

Motivation and Context
This change addresses issue #3392 to prevent errors when building vignettes that depend on rmarkdown.


modules/data.atmosphere/DESCRIPTION
```diff
@@ -80,6 +80,6 @@ License: BSD_3_clause + file LICENSE
 Copyright: Authors
 LazyLoad: yes
 LazyData: FALSE
-VignetteBuilder: knitr
+VignetteBuilder: knitr, rmarkdown
```
infotroph, line 84: To fix the current  `check/modules` CI failure, add rmarkdown to the Suggests section, between current lines 71 and 72 above. (I thought all packages already had rmarkdown in Suggests, but apparently we missed this one).

harshagr70: Thanks for the feedback! I’ve added rmarkdown to the Suggests section as required .

infotroph (approved):
Thanks! This looks great.

I forgot to mention the last step, which was to run `./scripts/generate_dependencies.R` after editing the Suggests section.  I went ahead and did that and will merge as soon as CI finishes.

infotroph:
The remaining CI failures are an unrelated known issue (Docker deployment only works right from same-fork branches atm) and do not need to delay merge.

PR #3396, opened 2024-11-17
Remove obsolete qsub_build.sh script

### LINKED ISSUE #3133: remove obsolete scripts/qsub_build.sh
### Bug Description

qsub_build.sh uses install.dependencies.R which is no longer there, It should use install_pecan.sh instead as I thought

https://github.com/PecanProject/pecan/blob/develop/scripts/qsub_build.sh
### To Reproduce

Steps to reproduce the behaviour:
1. Go to ' https://github.com/PecanProject/pecan/blob/develop/scripts '
2. Click on 'qsub_build.sh'

harshagr70:
## Description
This PR removes the qsub_build.sh script from the repository as it has been determined to be obsolete and no longer in active use.

# fixes 
fixes #3133 

## Types of changes
I have removed the script - qsub_build.sh that was no longer in use. 

## Checklist:

1. limited activity in logs - the logs show only a few reference to scripts/qsuib_build.sh which suggests that its not frequently used and updates 
2. no direct reference - i found no direct reference for the script and also found no functions calling for the script , also checked for the workflow files but found no direct reference to the script .
3. also i have checked for the script as mentioned in the comment - referring to docker/depends/pecan.depends.R , this script caters the same functionality as of the qsuib_build.sh .
4. I found no reference or any detail about the script in the documentation . 
5. The last commits made to that script was about 9 years ago and since its not maintained and updated , the script currently do not caters the requirements of installing the dependencies .

Concluding all these points its evident that the script is no longer in use hence its completely safe to remove the script .


infotroph (approved):

infotroph (approved):
Thank you for doing the PR and especially for documenting the reasoning so thoroughly! I agree this can go and don't know of any other files that need to change because of it.

harshagr70:
thank you !! @infotroph for reviewing and merging the changes !

PR #3397, opened 2024-11-20
bump version add write

robkooper:
this should fix cache issues (expect many issues to be marked stale)

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -8,10 +8,15 @@ jobs:
   stale:
     if: github.repository == 'PecanProject/pecan'
 
+    permissions:
+      contents: write
+      issues: write
+      pull-requests: write
+
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -8,10 +8,15 @@ jobs:
   stale:
     if: github.repository == 'PecanProject/pecan'
 
+    permissions:
```
infotroph, line 11: For my reference: Is a permissions section needed in every workflow now?

infotroph (approved):

.github/workflows/stale.yml
robkooper: no, but it is needed here for teh cache, otherwise it keeps processing the same PR/Issues.

PR #3398, opened 2024-11-22
Draft: Workflow improvements to run independently of BETY

yinghaoSunn:
## Description
This draft Pull Request introduces changes to allow the LPJ-GUESS workflow to run independently of the BETY database. The main changes include:
- Updated database connection logic to handle missing BETY settings.
- Added fallback mechanisms for standalone workflow operation.

Feedback and further suggestions are welcome.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


base/workflow/R/runModule.run.write.configs.R
```diff
@@ -13,7 +13,6 @@ runModule.run.write.configs <- function(settings, overwrite = TRUE) {
     }
     return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = FALSE))
   } else if (PEcAn.settings::is.Settings(settings)) {
-    write <- settings$database$bety$write
```
infotroph: Can you explain the intended effect of deleting this line? My first guess is maybe the idea was for `settings$database$bety$write` to no longer get consulted at `run.write.configs` time and instead to have the ensemble and get.parameter.samples steps check it directly? But then I'm not sure why the changes in `run.write.configs` were needed, so I think I'm off-track already.

I ask because this deletion is what's causing the current CI failure -- it leaves `write` undefined in the call to run.write.configs on line 29 below, which then means the references to `write` inside run.write.configs get interpreted as the base function `write()` instead of as a logical value. I can think of several possible fixes ranging from re-adding this line to renaming the argument so it doesn't conflict with a base function name, but without knowing more about the intention I don't know which approach to recommend.

base/workflow/R/run.write.configs.R
```diff
@@ -25,14 +25,20 @@
...
-    con <- PEcAn.DB::db.open(settings$database$bety)
-    on.exit(PEcAn.DB::db.close(con), add = TRUE)
-  }, error = function(e) {
-    PEcAn.logger::logger.severe(
-      "Connection requested, but failed to open with the following error: ",
-      conditionMessage(e))
-  })
+  ## Skip database connection if settings$database is NULL
+  if (write == FALSE) {
+    PEcAn.logger::logger.info("Database connection skipped: No database settings provided.")
```
```suggestion
  if (!isTRUE(write)) {
    PEcAn.logger::logger.info("Not writing this run to database, so database connection skipped")
```

modules/uncertainty/R/ensemble.R
```diff
@@ -236,6 +237,7 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
       PEcAn.logger::logger.warn("We were not able to successfully establish a connection with Bety ")
```
dlebauer:
```suggestion
      PEcAn.logger::logger.warn("We were not able to successfully establish a connection with BETYdb ")
```

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -202,7 +206,7 @@ get.parameter.samples <- function(settings,
                                                env.samples, ens.sample.method, param.names)
     }
   }
-  
+  #ensemble.samples <- NULL
```
dlebauer: why add this commented out line?

base/workflow/R/runModule.run.write.configs.R
yinghaoSunn: Sorry for missing your comment earlier; I was busy with finals at the time. You’re absolutely right—I was trying to skip consulting `settings$database$bety$write`, but that shouldn’t have been done in `run.write.configs`. Your valuable comment has helped me since I was in the middle of troubleshooting the CI failure. Thanks!

modules/uncertainty/R/get.parameter.samples.R
yinghaoSunn: Sorry, it was part of a previous attempt. It should be deleted, but I forgot.

base/workflow/R/runModule.run.write.configs.R
infotroph: In that case re-adding this line, together with my suggested change in run.write.configs above that @dlebauer already applied for us, should be what you need to get CI passing 🤞

modules/uncertainty/R/ensemble.R
```diff
@@ -222,6 +222,7 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
   }
   
   # See if we need to write to DB
+  if (!is.null(settings$database) && !is.null(settings$database$bety) && !is.null(settings$database$bety$write)) {
```
infotroph: If a list entry exists we know all its parents exist too, so I think this can be safely simplified to just 
```suggestion
  if (!is.null(settings$database$bety$write)) {
```

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -202,7 +206,6 @@ get.parameter.samples <- function(settings,
                                                env.samples, ens.sample.method, param.names)
     }
   }
-
```
infotroph, line 205: nit: I suggest re-adding the empty line here -- It seems to be accepted R style  to have whitespace before the final (and thus implicitly returned) statement of a function.

infotroph (approved):
Thanks! this looks good and I'm happily building on it already for DB-free SIPNET runs.

modules/uncertainty/R/get.parameter.samples.R
yinghaoSunn: Thanks a lot！

base/workflow/R/run.write.configs.R
```diff
@@ -25,11 +25,14 @@
...
                               overwrite = TRUE) {
-  ## Skip database connection if settings$database is NULL
+  ## Skip database connection if settings$database is NULL or write is False
   if (!isTRUE(write)) {
     PEcAn.logger::logger.info("Not writing this run to database, so database connection skipped")
     con <- NULL  # Set con to NULL to avoid errors in subsequent code
-  } else {
+  } else if(is.null(settings$database)) {
+    PEcAn.logger::logger.info("There's no database connection, so database connection skipped")
+    con <- NULL  # Set con to NULL to avoid errors in subsequent code
```
infotroph: I'm only about 60% sure of this, but I'm inclined to think this should be an error -- we only hit this condition when `write = TRUE`, which seems like reasonably strong evidence the user really does want to use the database. Thoughts?

mdietze: Correct. I discussed this with @yinghaoSunn last week as the most likely source of his GH Actions failure and we discussed him changing the first `if` case to be `!isTRUE(write) && is.null(settings$database)` and then within the `else` to have the second `if` throwing an error (not a warning) with an informative error message if `is.null(settings$database)` because that means you want to write but dont have a db connection. The second else would then have the tryCatch to open the connection.  It's important that the very first `if` check both write and db being null because GH actions is failing in the case where write == FALSE but the db connection is still needed to READ from the db to get information that's not being provided in the pecan.xml (which presumably IS provided by the user when they don't provide settings$database)

yinghaoSunn: Thank you for pointing this out and for the detailed explanation! It was an oversight on my part to skip over this scenario without throwing an informative error. I’ve updated the logic to handle the condition where `write == TRUE` but `settings$database == NULL`.

modules/uncertainty/R/ensemble.R
```diff
@@ -222,6 +222,7 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
   }
   
   # See if we need to write to DB
+  if (!is.null(settings$database$bety$write)) {
   write.to.db <- as.logical(settings$database$bety$write)
```
infotroph: Should have the same effect but make it a bit clearer what's happening (I think the fact the existing function completely ignored the `write.to.db` in the function args was a bug).
Might also be worth updating the Roxygen entry on line 198 to something like `@param write.to.db logical: Record this run in BETY? Note that if settings$database$bety$write is set, it will override this`
```suggestion
  if (!is.null(settings$database$bety$write)) {
    # specifying `write` in settings overrides write.to.db in fn args
    write.to.db <- as.logical(settings$database$bety$write)
  }
```

modules/uncertainty/R/ensemble.R
```diff
@@ -233,9 +234,10 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
     # If we fail to connect to DB then we set to NULL
     if (inherits(con, "try-error"))  {
       con <- NULL
-      PEcAn.logger::logger.warn("We were not able to successfully establish a connection with Bety ")
+      PEcAn.logger::logger.warn("We were not able to successfully establish a connection with BETYdb ")
     }
   }
+  }
```
infotroph: Only accept this one if also accepting my change above ;)
```suggestion
```

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -202,7 +204,7 @@ get.parameter.samples <- function(settings,
                                                env.samples, ens.sample.method, param.names)
     }
   }
-  
+
```
infotroph:
```suggestion

```

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -202,7 +204,6 @@ get.parameter.samples <- function(settings,
                                                env.samples, ens.sample.method, param.names)
     }
   }
-  
   save(ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples,
```
infotroph, line 207:
```suggestion

  save(ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples, 
```

modules/uncertainty/R/ensemble.R
yinghaoSunn: I agree. It will help understand the parameter priority.

base/workflow/R/run.write.configs.R
```diff
@@ -25,14 +25,25 @@
...
-  })
+  ## Skip database connection if settings$database is NULL or write is False
+  if (!isTRUE(write) && is.null(settings$database)) {
+    PEcAn.logger::logger.info("Not writing this run to database, so database connection skipped")
+    con <- NULL  # Set con to NULL to avoid errors in subsequent code
+  } else if(is.null(settings$database)) {
+  PEcAn.logger::logger.error(
+    "Database is NULL but writing is enabled. Provide valid database settings in pecan.xml."
+  )
+  stop("Database connection required but settings$database is NULL.")
```
```suggestion
    PEcAn.logger::logger.error(
      "Database is NULL but writing is enabled. Provide valid database settings in pecan.xml."
    )
    stop("Database connection required but settings$database is NULL.")
```

base/workflow/R/start_model_runs.R
```diff
@@ -170,9 +172,10 @@ start_model_runs <- function(settings, write = TRUE, stop.on.error = TRUE) {
         )
       }
       
-      # write finished time to database
-      PEcAn.DB::stamp_finished(con = dbcon, run = run)
-      
+      if (write){
+        # write finished time to database
+        PEcAn.DB::stamp_finished(con = dbcon, run = run)
+        }
```
infotroph:
```suggestion
      }
```

infotroph:
@yinghaoSunn Thank you for the work on this! I would be ready to merge it now, but since it's still marked as a draft I wanted to check first if there are other changes you're planning to add.

yinghaoSunn:
> @yinghaoSunn Thank you for the work on this! I would be ready to merge it now, but since it's still marked as a draft I wanted to check first if there are other changes you're planning to add.

Thank you for checking! I don’t have anything else to add. I really appreciate the support and guidance from you, David, and Mike throughout this process. This is my first Pull Request, and I’m thrilled to contribute to the community.

PR #3406, opened 2025-01-02
Add codes to extract SoilGrids soil texture data and derive ensemble …

Qianyuxuan:
Add codes to extract SoilGrids soil texture data and derive ensemble soil parameter files

## Description
1. Add a script "soilgrids_texture_extraction.R" that can extract and save three types of soil texture data in parallel for a single or group of lat/long locations based on user-defined site location from SoilGrids250m data.
2. Add a script "soil_params_ensemble.R" that can estimate the soil parameters based on soil texture data and write the ensemble parameter paths into settings.
3. Modify "soil_utils.R" and "extract_soil_nc.R" to fix the bugs when generating soil parameter files.
4. Modify "write.config.SIPNET.R" to include codes that can read and write soil water holding capacity into parameter settings.
5. Fix the bug that the met ensemble members for t=1 and t>1 in SDA codes differ.

## Motivation and Context
Provide ensemble soil parameter files based on SoilGrids soil texture data for SDA.

## Review Time Estimate

- Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- My change requires a change to the documentation.
- My name is in the list of CITATION.cff
- I agree that PEcAn Project may distribute my contribution under any or all of
- I have updated the CHANGELOG.md.
- I have updated the documentation accordingly.
- I have read the **CONTRIBUTING** document.
- I have added tests to cover my changes.
- All new and existing tests passed.


models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,20 +462,23 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-      #LitterWHC
-      #param[which(param[, 1] == "litterWHC"), 2] <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[1]*100
-    }
-    if("soil_hydraulic_conductivity_at_saturation"%in%names(soil_IC_list$vals)){
-      #litwaterDrainrate
-      param[which(param[, 1] == "litWaterDrainRate"), 2] <- unlist(soil_IC_list$vals["soil_hydraulic_conductivity_at_saturation"])[1]*100/(3600*24)
+  #working on reading soil file (only working for soilWHC)
+  if (length(settings$run$inputs$soilinitcond$path) > 0) {
+    soil_ICs_num <- length(settings$run$inputs$soilinitcond$path)
+    soil_ICs_path <- settings$run$inputs$soilinitcond$path[[sample(1:soil_ICs_num, 1)]]
```
mdietze: I'm guessing the logic of this bit is that if an ensemble input is passed in then you choose one at random? Personally, I'd prefer to throw an error at this point, as by this point in the workflow ensemble inputs should have already been converted to specific inputs for each run. If they're still ensemble at this point it's an error and we should stop. Proceeding on with a random choice means (a) we don't catch an error and (b) we don't know/record which ensemble member mapped to which run.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,20 +462,23 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    if ("volume_fraction_of_water_in_soil_at_saturation" %in% names(soil_IC_list$vals)) {
+      # Calculate the soilWHC for each soil layer in cm
+      soilWHC_0.025 <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[which(soil_IC_list$dims$depth == 0.025)] * 5
+      soilWHC_0.1 <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[which(soil_IC_list$dims$depth == 0.1)] * 10
+      soilWHC_0.225 <-unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[which(soil_IC_list$dims$depth == 0.225)] * 15
+      soilWHC_0.45 <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[which(soil_IC_list$dims$depth == 0.45)] * 30
+      soilWHC_0.8 <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[which(soil_IC_list$dims$depth == 0.8)] * 40
+      soilWHC_1.5 <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[which(soil_IC_list$dims$depth == 1.5)] * 100
+      # Calculate the soilWHC for the whole soil profile in cm
+      soilWHC_total <- soilWHC_0.025 + soilWHC_0.1 + soilWHC_0.225 + soilWHC_0.45 + soilWHC_0.8 + soilWHC_1.5
```
mdietze: This bit is hard-coding the specific depths associated with a specific product and will fail if different depths are used (e.g. from a different product). The logic here should be easy to generalize to extracting the vector of soilWHC's and the vector of depths, then calculating the layer thicknesses, then performing a vector sum

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,20 +462,23 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-    soil_IC_list <- PEcAn.data.land::pool_ic_netcdf2list(settings$run$inputs$soilinitcond$path)
-    #SoilWHC and LitterWHC
-    if("volume_fraction_of_water_in_soil_at_saturation"%in%names(soil_IC_list$vals)){
-      #SoilWHC
-      param[which(param[, 1] == "soilWHC"), 2] <- mean(unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"]))*100
-      
-      #LitterWHC
-      #param[which(param[, 1] == "litterWHC"), 2] <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[1]*100
-    }
-    if("soil_hydraulic_conductivity_at_saturation"%in%names(soil_IC_list$vals)){
```
mdietze, line 504: Modified code seems to loose the litter drainage rate. Also not sure why the litterWHC was commented out, but doesn't seem like we want to loose that either.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -382,7 +386,12 @@ soil.units <- function(varname = NA){
                                       "soil_thermal_conductivity","W m-1 K-1", 
                                       "soil_thermal_conductivity_at_saturation","W m-1 K-1", 
                                       "soil_thermal_capacity","J kg-1 K-1",
-                                      "soil_albedo","1"
+                                      "soil_albedo","1",
+                                      "slpotwp","1", #intermediate variable, meaningless and unit "1" is used for convenience
+                                      "slpotcp","1", #intermediate variable, meaningless and unit "1" is used for convenience
```
mdietze: If these two are meaningless intermediate values, why are we adding them to the output?

mdietze: Same applies to slcpd below, which is used to calculate soil_thermal_capacity, and slden, which seems to be an unused duplicate of soil_bulk_density. Indeed, in my original code all 4 variables were dropped from the output, and then someone later commented this out. I suspect the commented out part needs to be restored:
https://github.com/PecanProject/pecan/blob/882766d7a010a091b8de30075951c57e11cb43e5/modules/data.land/R/soil_utils.R#L238

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+      alpha <- means * alpha0
+      q5 <- quantiles$q5
+      q50 <- quantiles$q50
+      q95 <- quantiles$q95
+      
+      # Compute differences with observed quantiles
+      samples <- MCMCpack::rdirichlet(10000, alpha) # Generate samples
+      sim_q5 <- apply(samples, 2, quantile, probs = 0.05, na.rm = TRUE)
+      sim_q50 <-apply(samples, 2, quantile, probs = 0.50, na.rm = TRUE)
+      sim_q95 <-apply(samples, 2, quantile, probs = 0.95, na.rm = TRUE)
```
mdietze: would be simpler to do these three in one line by specifying probs = c(0.05, 0.5, 0.95).

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+##' @param means A vector of means of sand, clay, and silt proportion for one soil layer at one site from SoilGrids data
+##' @param quantiles A list of 5th, 50th, and 95th percentiles for sand, clay, and silt for one soil layer at one site from SoilGrids data
+##'
+##' @example
+##' \dontrun{
+##' means <- c(0.383,0.213,0.403) # means of sand, clay, and silt at one site and one depth
+##' quantiles <-list(
+##' q5 = c(0.551,0.055,0.069), # 5th percentile for each category: sand, clay, and silt at one site and one depth
+##' q50 = c(0.691,0.124,0.156), # 50th percentile (median) for each category: sand, clay, and silt at one site and one depth
+##' q95 = c(0.786,0.258,0.297))  # 95th percentile for each category: sand, clay, and silt at one site and one depth
```
mdietze: In this example, the first and third means fall outside the interval estimates by a LOT. Suspect the test example needs fixing

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+      lower = 0.01  # alpha0 must be positive
+    )
+    return(result$par)
+  }
+ 
+  alpha0 <- estimate_alpha0(means, quantiles)
+  if (alpha0 <= 0) {
+    stop("Estimated alpha0 is non-positive, which is invalid.")
+  }
+  alphas <- means * alpha0
```
mdietze, line 53: Would be good to provide more comments in the code about the logic of what's being done here (both this specific line and the more general approach)

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+         means$fraction_of_clay_in_soil <- soil_rescaled$clay
+         means$fraction_of_silt_in_soil <- soil_rescaled$silt
+    
+         # Estimate Dirichlet parameters
+         alpha_est <- estimate_dirichlet_parameters(as.matrix(means), quantiles)
+
+         # Generate the ensemble soil texture data based on the ensemble size (ens_n) defined in the settings
+         samples <- MCMCpack::rdirichlet(10000, alpha_est)
+         colnames(samples) <-c("fraction_of_sand_in_soil","fraction_of_clay_in_soil","fraction_of_silt_in_soil")
+         samples_ens_new <-list(samples[sample(1:10000, ens_n), ]) %>% setNames(depths)
```
mdietze: I'm not following why you take more samples than you need, then subsample down to the number needed. Why not just ask rdirichlet to generate ens_n samples in the first place?

modules/data.land/R/soil_utils.R
```diff
@@ -264,7 +264,16 @@ sclass <- function(sandfrac,clayfrac){
...
-  silt <- 100. - sand - clay
+  # Prevent silt from being negative due to numerical precision issue
+  silt <- c()
+  for (i in seq_along(sand)) {
+    if (sand[i] + clay[i] >= 100.) {
+      silt[i] <- 0
+    } else {
+      silt[i] <- 100. - sand[i] - clay[i]
+    }
+  }
```
mdietze: simpler would have been `silt = pmax(100. - sand - clay, 0)`

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+##' soil_params_ensemble(settings,sand,clay,silt,outdir)
+##'  }
+##'  
+##'  @return Ensemble soil parameter files defined in outdir and file paths in xml file
+##'  @export 
+##'  @author Qianyu Li
+##'  @importFrom magrittr %>%
+##'  
+
+soil_params_ensemble <- function(settings,sand,clay,silt,outdir,write_into_settings=TRUE){
```
mdietze: Conceptually, it would be nice to have a more clean separation between things that are specific to one product vs general. Right now this function is specific to SoilGrids. I think you might want to either generalize the function a bit more (e.g. so that it would still work with a different soil map with different layer thicknesses and different data structures) or modify the function name to make it clear that it is only for this one product.

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+  texture_all <-merge(sand, clay, by=c("Depth", "Quantile", "Siteid"))  %>% merge(silt, by=c("Depth", "Quantile", "Siteid")) %>%
+  `colnames<-`(c(
+      "soil_depth", #"soil_depth" will be used in "soil2netcdf" function
+      "quantile",
+      "siteid",
+      "fraction_of_sand_in_soil",
+      "fraction_of_clay_in_soil",
+      "fraction_of_silt_in_soil"))
+
+   # Substitute the depth range with the middle depth value in meter
```
mdietze: Why middle? I'm pretty sure earlier soil code was working with the bottoms of each layer (with the assumption that the first layer's top is always at 0 by definition). Using the middles will generate ambiguity about layer thicknesses.

mdietze (changes requested):

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Yeah I followed the logic as choosing one from ensemble initial conditions here:https://github.com/PecanProject/pecan/blob/develop/models/sipnet/R/write.configs.SIPNET.R#L546. So you suggested we should just randomly choose and write one ensemble member in the script "soil_params_ensemble.R" and read that specific path in "write.configs.SIPNET.R"?

mdietze: So your `soil_params_ensemble` should generate a bunch of ensembles and load them *ALL* into the overall settings object. What I'm saying is that write.configs.SIPNET should only end up seeing ONE of those ensembles because `run.write.configs` should do the joint sampling of all the different types of ensembles (met, soils, IC, etc), record which ensemble member was assigned which inputs, and then ensure that the settings object reaching write.configs.SIPNET is only for a specific ensemble member. If you're finding that write.configs.SIPNET is being handed all of the soil ensemble  members then there's something wrong upstream in the pecan.xml (e.g. not including soils in the <ensemble> section) or in run.write.configs

modules/data.land/R/extract_soil_nc.R
Qianyuxuan: Yeah I agree we should restore the change. I added them to fix the bug due to the missing definition of these variables in "extract_soil_nc.R" but they are only meaningful internally.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+   } else {
+    future::plan(future::multisession,workers=3)
+  }
+  
+  sand_data_url <-
+    "/vsicurl?max_retry=30&retry_delay=60&list_dir=no&url=https://files.isric.org/soilgrids/latest/data/sand/sand_"
+  clay_data_url <-
+    "/vsicurl?max_retry=30&retry_delay=60&list_dir=no&url=https://files.isric.org/soilgrids/latest/data/clay/clay_"
+  silt_data_url <-
+    "/vsicurl?max_retry=30&retry_delay=60&list_dir=no&url=https://files.isric.org/soilgrids/latest/data/silt/silt_"
```
infotroph: Maybe a bigger change than you're interested in doing here, but have you looked into grabbing these from the soilgrids Web Coverage Service? I noticed for the SOC function that the vsicurl approach winds up pulling the entire ~80MB map for every layer.

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,224 @@
...
+    rescaled_clay <- clay / total
+    rescaled_silt <- silt / total
+    return(list(
+      sand = rescaled_sand,
+      clay = rescaled_clay,
+      silt = rescaled_silt))
+  }
+  
+  # A function to reformat the nested list as inputs to "soil2netcdf" function
+  reformat_soil_list <- function(samples_all_depth) {
```
infotroph: Nit: these helper functions are long enough it's hard to scan for where their definitions stop and the parent function resumes. I'd find it more readable if they were defined at the bottom of this file outside of soil_params_ensemble().

modules/data.land/R/soilgrids_texture_extract.R
mdietze: I think this is on purpose. I think at this point Cherry and Dongchen have found that grabbing the entire map is faster than trying to get an API to grab each point one-at-a-time. We're grabbing >6400 points for model execution and all of North America for spatial downscaling.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: @mdietze It seems that "run.write.configs" does not play a role in our current SDA workflow but I did find ways to do the sampling of soil ensemble members through the codes in "ensemble.R" by providing "soilinitcond" tag under "ensemble" in the XML file: https://github.com/PecanProject/pecan/blob/develop/modules/uncertainty/R/ensemble.R#L401-L406. However, it is only for a new fresh run and seems not applicable when it is a restart. The current restart components only include met inputs and parameters as https://github.com/PecanProject/pecan/blob/develop/modules/uncertainty/R/ensemble.R#L422C5-L424. What is your suggestion on that?

mdietze: 1. write.ensemble.configs is what's called by run.write.configs, so if that's what SDA is calling too we should be fine. But as noted earlier, it's important for the code to confirm that it's only receiving on ensemble member for each input, and to throw an error if it's not rather than resampling.
2. Restart is grabbing `inputs` and soils *should* be part of inputs (as should phenology, initial conditions, etc), not just met. You should verify this is working correctly, as it is is critical that ensemble sampling is preserved (not repeated) both iteratively within a SDA run and across SDA runs (e.g., when running a forecast today that starts from yesterday's forecast). Resampling things (params, inputs, IC, etc) that have already been samples is going to seriously mess up the covariance structures in our products

Qianyuxuan: @mdietze Based on the current argument passing of restart in "sda.enkf_MultiSite.R": https://github.com/PecanProject/pecan/blob/develop/modules/assim.sequential/R/sda.enkf_MultiSite.R#L358-L368, it seems we didn't grab soil, phenology and IC other than met now for inputs (note only "met" is specified in "input.ens.gen"). There is also a comment:  #TODO: incorporate Phyllis's restart work. #sample all inputs specified in the settings$ensemble not just met". So I guess it is what we should add here?

modules/data.land/R/soil_utils.R
```diff
@@ -264,7 +264,16 @@ sclass <- function(sandfrac,clayfrac){
   #----- Define the percentage of sand, clay, and silt. ----------------------------------#
   sand <- 100. * sandfrac
   clay <- 100. * clayfrac
-  silt <- 100. - sand - clay
+  # Prevent silt from being negative due to numerical precision issue
+  silt <- c()
+  for (i in seq_along(sand)) {
+    if (sand[i] + clay[i] >= 100.) {
```
dlebauer: if sand + clay > 100 shouldn't it be renormalized?

```
if sand + clay > 100{
  sand <- sand/(sand+clay) * 100
  clay <- clay/(sand+clay) * 100
}
```

and is this an error in SoilGrids that warrants contacting the creators?

dlebauer: Actually, now I see this normalizing is done in the rescale_sum_to_one function; would it make sense to add this silt correction that function?

Qianyuxuan: The normalizing in rescale_sum_to_one mainly solved the data issue (the sum of means doesn't equal to 1) of SoilGrids. But the codes in soil_utils.R are only for the R numeric rounding issue as the texture percentages used here are from Dirichlet distribution sampling and their sums strictly equal to one. So I guess non-normalization here seems fine.

> Actually, now I see this normalizing is done in the rescale_sum_to_one function; would it make sense to add this silt correction that function?

modules/data.land/R/soilgrids_texture_extract.R
dlebauer: It looks this function grabs the virtual map (VRT is essentially an index to soilgrids) and then uses that to query points. ~Has anyone checked if it would be faster to download the rasters and extract data from them locally?~

It looks like it is much (~100x) faster to download rasters locally and then query them. I was wondering and have started developing a function to do the downloading. I can take a whack at refactoring this code, including generalizing it to extract any of the soilgrids variables.

<details>

<summary>Benchmark code</summary>

```text
library(terra)
library(sf)
library(dplyr)
library(microbenchmark)

# approximately California extent
extent <- ext(-124, -120, 34, 40)  # Updated extent for California

# random points for testing
set.seed(42)
n_points <- 50  # Number of points for testing
site_info <- tibble(  # Using tibble for cleaner data manipulation
  site_id = paste0("Site", 1:n_points),
  lon = runif(n_points, min = -124, max = -120),
  lat = runif(n_points, min = 34, max = 40)
) |> 
  st_as_sf(coords = c("lon", "lat"), crs = 4326)  # Convert to sf object

soil_property <- "clay"
depth <- "0-5cm"
quantile <- "mean"

vrt_url <- sprintf(
  "/vsicurl/https://files.isric.org/soilgrids/latest/data/%s/%s_%s_%s.vrt",
  soil_property, soil_property, depth, quantile
)

test_raster <- file.path(tempdir(), "local_clay_0-5cm_CA.tif")


download_raster <- function() {
  rast(vrt_url) |>
    crop(extent) |> 
    writeRaster(test_raster, overwrite = TRUE)
}

# Method 1: query VRT
query_vrt <- function() {
  rast(vrt_url) |>  
    crop(extent) |> 
    extract(vect(site_info))  
}

# Method 2: query a local raster
query_local <- function() {
  rast(test_raster) |>  
    extract(vect(site_info))  
}

download_raster()

benchmark_results <- microbenchmark(
  VRT_Query = query_vrt(),
  Local_Query = query_local(),
  times = 3
)

print(benchmark_results)
```
</details>

```
Unit: milliseconds
        expr        min         lq       mean   median         uq        max neval cld
   VRT_Query 1763.45699 1766.54618 1786.76361 1769.635 1798.41693 1827.19849     3  a 
 Local_Query   13.66672   14.61436   16.07034   15.562   17.27215   18.98229     3   b
```

infotroph: Well sure, you're comparing a function that does all its downloading before you start the timer against once that does all its downloading after you start the timer. How long does download_raster() take?

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: I dropped litter drainage rate as I found the current little water is off in the master branch of SIPNET. I added it this time but noted that the soil hydraulic conductivity for the whole soil profile based on layered values differed for vertical flow and horizontal flow so I provided equations for both.

mdietze: Does having parameters in the sipnet.in file that correspond to options that are on/off for any particular sipnet build cause sipnet itself to crash (or parse the parameter file incorrectly)? If not, we might as well have PEcAn provide all the parameters it can cause otherwise we need some way of tracking what compilation flags are currently set in the SIPNET executable you're trying to call, and then adjusting the inputs accordingly. And even if we did that, it would be very easy to change the SIPNET flags and then run with PEcAn-generated inputs that are out-of-date.

modules/data.land/R/soilgrids_texture_extract.R
dlebauer: download_raster takes 1s for a 1x1 degree tile that is 126kb. Also, it would only be done once, in the background, then further testing and development would be 100x quicker each time. Downloading once also respects the data providers.

Here is a test on a 1x1 degree grid with 50 points sampled. By the time we sample 100k points across CA, the advantage would be even greater.

Also, a single map layer covering CA and much of NV is ~20MB

```
Unit: milliseconds
           expr    min     lq mean median   uq  max neval
 Local_Download 1007.0 1020.2 1104   1033 1152 1271     3
    Local_Query    8.5    8.8   32      9   44   79     3
      VRT_Query  306.2  311.9  320    318  327  337     3
```

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: I added soil parameter files to inputs for the restart list. Phenology parameters are sampled within parameter sampling and initial conditions don't apply for t>1 situation, so I guess the current solution seems fine.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Fixed.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Yeah I agree we should provide as many parameters as we can.

modules/data.land/R/soil_params_ensemble.R
Qianyuxuan: Fixed.

modules/data.land/R/soil_params_ensemble.R
Qianyuxuan: Fixed.

modules/data.land/R/soil_params_ensemble.R
Qianyuxuan: Fixed.

modules/data.land/R/soil_utils.R
Qianyuxuan: Fixed.

modules/data.land/R/soil_params_ensemble.R
Qianyuxuan: Fixed with change of function name.

modules/data.land/R/soil_params_ensemble.R
Qianyuxuan: Fixed with layer bottom value.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: I will incorporate the direct download function once it is done.

modules/data.land/R/soil_params_ensemble.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
infotroph: We should take this to another thread and get out of Cherry's way, but @dlebauer my point is that the benchmark you showed is not a meaningful comparison. Please do not rely on it for any decisions.

 If Mike, Cherry, and Dongchen have all considered this and found the vrt approach to be faster than individual queries for large runs, then I believe them and am happy to pay the overhead on small runs.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+  
+  
+  if (future::supportsMulticore()) {
+    future::supportsMulticore()
+    future::plan(future::multicore)
+   } else {
+    future::plan(future::multisession,workers=3)
+  }
+  
+  sand_data_url <-
```
DongchenZ: In a real situation where we might already have the `SoilGrids` data downloaded on the server, we might want to replace the `VRT` URL with the actual physical path to the data. To do this, we might need to set those VRT paths as the function arguments (e.g., `list(sand_data_url = "/vsicurl?max_retry=30&retry_delay=60&list_dir=no&url=https://files.isric.org/soilgrids/latest/data/sand/sand_")`). The downloaded file path, for example on the SCC, can be `/projectnb/dietzelab/dongchen/anchorSites/NA_runs/soil_nc/soilgrids_250m/sand/sand_0-5cm_mean/sand/sand_0-5cm_mean.tif`.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+    }
+    else {
+      PEcAn.logger::logger.error("No output directory found.")
+    }
+    # Return the results to the terminal as well
+    return(soilt_df)
+  }
+  
+  
+  if (future::supportsMulticore()) {
```
DongchenZ, line 114: It's not recommended to have two layers of `parallel` settings (e.g., the one you parallel over the site locations and this one parallel over the products). In practice, because of the limitation of `SoilGrids API`, the parallel over the site locations can be turned off.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+    # Reproject locations to soilgrids projection
+    # Soilgrids data is using Homolosine projection https://www.isric.org/explore/soilgrids/faq-soilgrids
+    p <-terra::vect(lonlat, crs = "+proj=longlat +datum=WGS84") # Users need to provide lon/lat
+    newcrs <- "+proj=igh +datum=WGS84 +no_defs +towgs84=0,0,0"
+    p_reproj <- terra::project(p, newcrs) # Transform the point vector to data with Homolosine projection
+    data_tag <- c("_mean.vrt", "_Q0.05.vrt", "_Q0.5.vrt", "_Q0.95.vrt")
+    name_tag <- expand.grid(depths, data_tag, stringsAsFactors = F) #find the combinations between data and depth tags.
+    L <- split(as.data.frame(name_tag), seq(nrow(as.data.frame(name_tag))))#convert tags into lists.
+    
+    if ("try-error" %in% class(try(soilt_real <- L %>% furrr::future_map(function(l){
```
DongchenZ: When the site number goes up, this parallel will be more likely to produce errors due to the limitation of the `SoilGrids` server. The extraction speed is similar with or without the parallel in my previous `6400 site runs`. Thus, it's recommended to just delete the parallel extraction part from the script.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+    
+    if (is.null(site_info)) {
+      PEcAn.logger::logger.error(
+        "No site information found. Please provide a BETY DB site list containing at least the site id and PostGIS geometry\
+    as lon and lat"
+      )
+    }
+    
+    # Prepare site info for extraction
+    internal_site_info <- data.frame(site_info$site_id, site_info$site_name, site_info$lat,site_info$lon)
```
DongchenZ: I think the `site_id` is enough to describe each site rather than combining `site_id` and `site_name`. Thus, you might want to just remove the `site_name` from the list to make things simpler.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+    #Create a variable to store mean and quantile of soil texture data for each soil depth
+    soiltquant <- matrix(NA, nrow = 6, ncol = length(internal_site_info$site_info.lon) * 4)
+    lonlat <-cbind(internal_site_info$site_info.lon, internal_site_info$site_info.lat)
+    depths <-c("0-5cm", "5-15cm", "15-30cm", "30-60cm", "60-100cm", "100-200cm")
+    
+    # Reproject locations to soilgrids projection
+    # Soilgrids data is using Homolosine projection https://www.isric.org/explore/soilgrids/faq-soilgrids
+    p <-terra::vect(lonlat, crs = "+proj=longlat +datum=WGS84") # Users need to provide lon/lat
+    newcrs <- "+proj=igh +datum=WGS84 +no_defs +towgs84=0,0,0"
```
DongchenZ: The projection to this specific `CRS` will be used if you only extract values through the `SoilGrids VRT URL`. If you are using local files, you should use the CRS from the target raster files (see my comments below).

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+##' @export
+##' @author Qianyu Li
+##' @importFrom magrittr %>%
+
+soilgrids_texture_extraction <- function(site_info, outdir=NULL, verbose=TRUE){
+  
+  # A function to extract and save one type of soil texture data 
+  download_and_extraction <- function(base_data_url, site_info, outdir) {
+    
+    if (future::supportsMulticore()) {
```
DongchenZ: Given the recommendations that you'll want to remove the parallel over the site locations, the parallel settings might also be removed from the script, if you don't mind.

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
+##' soilgrids_texture_extraction function
+##' A function to extract and save three types of soil texture data in parallel for a single or group of 
+##' lat/long locations based on user-defined site location from SoilGrids250m 
+##' version 2.0 : https://soilgrids.org
+##' @title soilgrids_texture_extraction
+##' @name soilgrids_texture_extraction
+##' 
+##' @param site_info A data frame of site info containing the BETYdb site ID, 
+##' site name, latitude, and longitude, e.g. 
+##' (site_id, site_name, lat, lon)
```
DongchenZ: The `site_name` is not necessary for setting up the site lists (see my below comments).

modules/data.land/R/soilgrids_texture_extract.R
```diff
@@ -0,0 +1,129 @@
...
+##' @param outdir  Provide the path to store the texture data file
+##' @param verbose Provide progress feedback to the terminal? TRUE/FALSE
+##' @return a data frame containing the soil texture data with columns "Depth", "Quantile", "Siteid", and "Value"
+##' 
+##' @export
+##' @author Qianyu Li
+##' @importFrom magrittr %>%
+
+soilgrids_texture_extraction <- function(site_info, outdir=NULL, verbose=TRUE){
```
DongchenZ: It's recommended to have extra arguments `data_paths` with the default `data_paths = list(sand_data_url = URL1, clay_data_url = URL2, silt_data_url = URL3)`, so the URLs provided here can be either the `SoilGrids` VRT URLs or physical paths to the local raster files (see my comments below for more details).

DongchenZ (changes requested):
So the general script is well established followed by a similar style as the previous SOC extraction function. The questions that I raised here are: 1) for the parallel design when over the site locations and data products, and 2) for the usage of downloaded local raster files (if they exist).

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,22 +462,40 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-    soil_IC_list <- PEcAn.data.land::pool_ic_netcdf2list(settings$run$inputs$soilinitcond$path)
-    #SoilWHC and LitterWHC
-    if("volume_fraction_of_water_in_soil_at_saturation"%in%names(soil_IC_list$vals)){
-      #SoilWHC
-      param[which(param[, 1] == "soilWHC"), 2] <- mean(unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"]))*100
-      
-      #LitterWHC
-      #param[which(param[, 1] == "litterWHC"), 2] <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[1]*100
+  #working on reading soil file
+  if (length(settings$run$inputs$soilinitcond$path) > 0) {
```
mdietze: Still no clear how you're handling the case if >1 file is passed in

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,22 +462,40 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+      ## override if specified in inputs
+      if ("soilinitcond" %in% names(inputs)) {
+        template.soilinit <- inputs$soilinit$path
+      }
     }
-    if("soil_hydraulic_conductivity_at_saturation"%in%names(soil_IC_list$vals)){
+    soil_IC_list <- PEcAn.data.land::pool_ic_netcdf2list(template.soilinit)
+    # Calculate the thickness of soil layers based on the assumption that the depth values are at bottoms and the first layer top is at 0
+    if ("depth" %in% names(soil_IC_list$dims)) {
+      thickness<-c(soil_IC_list$dims$depth[1],diff(soil_IC_list$dims$depth))
```
mdietze: Does this line still work if the product only has one layer?

mdietze: Also, should/could there be an "else" here? A bunch of code below continues to check for "depth" being defined, but doesn't ever use depth again --  instead you're just using this to ensure thickness is defined.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,22 +462,40 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    }
+    #SoilWHC
+    if ("depth" %in% names(soil_IC_list$dims) && "volume_fraction_of_water_in_soil_at_saturation" %in% names(soil_IC_list$vals)) {
+      # Calculate the soilWHC for the whole soil profile in cm
+      soilWHC_total <- sum(unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])*thickness)
+      param[which(param[, 1] == "soilWHC"), 2] <- soilWHC_total
+    #LitterWHC
+      param[which(param[, 1] == "litterWHC"), 2] <- soilWHC_total
```
mdietze: This is incorrect. If there's a litter layer it would have a much much smaller WHC than the entire soil pool.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,22 +462,40 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    if ("depth" %in% names(soil_IC_list$dims) && "soil_hydraulic_conductivity_at_saturation" %in% names(soil_IC_list$vals)) {
+      # Calculate the soil_hydraulic_conductivity for the whole soil profile (soilHC_total) if vertical flow dominates for
+      # unsaturated conditions or in deep, well-drained profiles. If horizontal flow dominates such as for 
+      # saturated conditions or where barriers to vertical flow exist, the overall conductivity is the 
+      # weighted arithmetic mean based on the proportion of the total thickness each layer represents as:
+      # soilHC_total <- sum(unlist(soil_IC_list$vals["soil_hydraulic_conductivity_at_saturation"])*thickness)/sum(thickness)
+      soilHC_total <- sum(thickness)/sum(thickness/(unlist(soil_IC_list$vals["soil_hydraulic_conductivity_at_saturation"])))
       #litwaterDrainrate
-      param[which(param[, 1] == "litWaterDrainRate"), 2] <- unlist(soil_IC_list$vals["soil_hydraulic_conductivity_at_saturation"])[1]*100/(3600*24)
+      param[which(param[, 1] == "litWaterDrainRate"), 2] <- soilHC_total
```
mdietze: First, I'm not sure why you're using the average hydraulic conductivity of the entire soil column to control the rate at which moisture leaves the litter and enters the first soil layer. Second, we should check that the way SIPNET uses the conductivity (as a proportional decay rate, see https://github.com/PecanProject/sipnet/blob/becec4d2d6d857fa25dc47f974c48621a0b9044b/sipnet.c#L1516) is compatible with the way the parameter is defined in the soils file (Darcy's law) or whether additional assumptions  or unit conversions are needed

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,22 +462,40 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    }
+    #SoilWHC
+    if ("depth" %in% names(soil_IC_list$dims) && "volume_fraction_of_water_in_soil_at_saturation" %in% names(soil_IC_list$vals)) {
+      # Calculate the soilWHC for the whole soil profile in cm
+      soilWHC_total <- sum(unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])*thickness)
+      param[which(param[, 1] == "soilWHC"), 2] <- soilWHC_total
+    #LitterWHC
+      param[which(param[, 1] == "litterWHC"), 2] <- soilWHC_total
+    }
+    if ("depth" %in% names(soil_IC_list$dims) && "soil_hydraulic_conductivity_at_saturation" %in% names(soil_IC_list$vals)) {
```
mdietze: per comment below about this being for surface litter, I'm not sure you need "depth" here

models/sipnet/R/write.configs.SIPNET.R
mdietze: > Phenology parameters are sampled within parameter sampling 

Not sure what you mean here, phenology files are separate from the parameter posterior files.

modules/assim.sequential/R/metSplit.R
```diff
@@ -32,17 +32,18 @@ metSplit <- function(conf.settings, inputs, settings, model, no_split = FALSE, o
       # Loading the model package - this is required bc of the furrr
       library(paste0("PEcAn.",model), character.only = TRUE)
       
-      inputs.split <- list()
+      # Initialize inputs.split with inputs to keep the sub-list of soil parameters "soilinitcond"
```
mdietze: I'm not following the logic here. The metSplit code should only ever be touching the met. The comment here about soil parameters is confusing, as metSplit shouldn't be touching ANY of the other inputs, not just soils. I'm having a hard time telling if the code below is incorrect, or if it's fixing a long standing bug, or if the larger issue is that metSplit shouldn't ever have been passed the entirety of inputs to begin with.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,28 @@ sda.enkf.multisite <- function(settings,
       #reformatting params
       new.params <- sda_matchparam(settings, ensemble.samples, site.ids, nens)
     }
-      #sample met ensemble members
+      #sample met and soil parameter ensemble members
```
mdietze: should be sampling ALL input ensemble members, not just met, not just (met + soils)

mdietze: >       #TODO: incorporate Phyllis's restart work
>      #      sample all inputs specified in the settings$ensemble not just met

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,28 @@ sda.enkf.multisite <- function(settings,
...
       #      sample all inputs specified in the settings$ensemble not just met
-      inputs <- PEcAn.settings::papply(conf.settings,function(setting) {
-        PEcAn.uncertainty::input.ens.gen(
-          settings = setting,
-          input = "met",
-          method = setting$ensemble$samplingspace$met$method,
-          parent_ids = NULL 
-        )
-       })
+      input_name <- c("met","soilinitcond")
```
mdietze: should not be hard coded. The set of inputs that are ensemble-based needs to be detected from the settings

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,28 @@ sda.enkf.multisite <- function(settings,
...
+      inputs <- list()
+      new_inputs <- list()
+      for (i_input in seq_along(input_name)){
+       inputs[[input_name[i_input]]]<-PEcAn.settings::papply(conf.settings,function(setting) {
```
mdietze: could you explain what the papply is doing here?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,28 @@ sda.enkf.multisite <- function(settings,
...
-          settings = setting,
-          input = "met",
-          method = setting$ensemble$samplingspace$met$method,
-          parent_ids = NULL 
-        )
-       })
+      input_name <- c("met","soilinitcond")
+      inputs <- list()
+      new_inputs <- list()
+      for (i_input in seq_along(input_name)){
```
mdietze: Inputs can't be sampled in an arbitrary order. See the logic in write.ensemble.configs. Indeed, if the SDA can't call write.ensemble.configs directly, then there's a lot of code there than needs to be refactored into a function that can be called by both forward and SDA code, since maintaining two separate versions of the sampling code is clearly resulting in code divergence.

modules/assim.sequential/inst/MultiSite-Exs/SDA/Create_Multi_settings.R
```diff
@@ -225,6 +225,7 @@ template <- PEcAn.settings::Settings(list(
   ensemble = structure(list(size = 25, variable = "NPP", 
                             samplingspace = structure(list(
                               parameters = structure(list(method = "lhc")),
+                              soilinitcond = structure(list(method = "sampling")),
```
mdietze: Check to make sure there aren't other inputs that are being left out

mdietze: Also, is the operational SDA using a LHC sampling of the posterior parameters? That's not random.

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: FYI Phyllis's work lives on the BU server at /projectnb/dietzelab/yixuanli and still needs going through to make sure (a) write.ensemble.configs is saving ALL the info it needs about ensembles, not just params and (b) that you can have the option to pass in a specific already-sampled ensemble of params, met, soil, IC, etc. without having the code resample it. This is important in the general case to allow for sensitivity analysis designs. It's also important for the SDA case as the sample from t==1 needs to be the same sample that's used for t>1.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: So the current design is when t=1, only one file path is sampled and passed into the setting based on codes in "write.ensemble.configs". When t>1, file paths of total ensemble size will be passed in but then "template.soilinit" will be assigned with the specific path in the inputs.

mdietze: @Qianyuxuan I was asking about n=1 versus n>1, not t=1 vs t>1, and I'm still confused -- if you're passed in a whole vector of paths instead of a single one, you should throw an error not continue on, but I'm not seeing that check anywhere.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -462,22 +462,40 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-      #LitterWHC
-      #param[which(param[, 1] == "litterWHC"), 2] <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[1]*100
+  #working on reading soil file
+  if (length(settings$run$inputs$soilinitcond$path) > 0) {
+    template.soilinit <- settings$run$inputs$soilinitcond$path  ## read from settings
+    
+    if (!is.null(inputs)) {
+      ## override if specified in inputs
+      if ("soilinitcond" %in% names(inputs)) {
+        template.soilinit <- inputs$soilinit$path
```
mdietze: I'm now realizing that I'm also confused by this bit. Why do settings$run$inputs$soilinitcond and settings$run$inputs$soilinit both exist?

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: I followed the steps for the pass of met files: https://github.com/PecanProject/pecan/blob/develop/models/sipnet/R/write.configs.SIPNET.R#L21-L29. I did find that a whole vector of met paths (10 ERA5 ensemble met paths defined in the settings) was passed in when t>1. But I don't think it is an error as the specific path will be overwritten by the one listed in the restart met input later.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: "soilinit" is only a typo and will be corrected.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Fixed.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Fixed.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Fixed.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Fixed. Only the conductivity for first soil layer is used for litWaterDrainRate and the equation complies to Darcy's law.

modules/assim.sequential/R/metSplit.R
Qianyuxuan: I didn't edit "metSplit" this time, instead I added similar modified codes in "sda.enkf_MultiSite.R". In my own opinion, the function "metSplit" lacks some flexibility as the formats "inputs = inputs$samples[[I]]", but we need to add another sublist such as "soilintcond" under inputs soil we need to specify "met".

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: I did find the previous design had two separate sampling in the SDA workflow, leading to met file for t==1 differs from that for t>1. I kept the sampling in SDA codes instead, saved the inputs as "samples", and passed it into "write.ensemble.configs" as an additional argument. If it is a fresh new run, those samples will be read and recorded directly. And those samples will be passed into the restart list when t>1 as well. Although it might not be the perfect solution, but I did test it and found it work well to make sure the ensemble member keeps the same through time.

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: Fixed.

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: In my new edits, the ids for samples that were generated in SDA codes and passed into the "write.ensemble.configs" will be recorded in "README.txt".

modules/assim.sequential/inst/MultiSite-Exs/SDA/Create_Multi_settings.R
Qianyuxuan: I think so. The current SDA workflow used "get.parameter.samples(settings, ens.sample.method = settings$ensemble$samplingspace$parameters$method)" to sample parameters and the method is "lhc".

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed. So now you can provide the local file folder to "local" in "data_paths". For each folder, there should be one type of texture data for different sites, layers, and statistics.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soilgrids_texture_extract.R
Qianyuxuan: Fixed.

modules/data.land/R/soil_params_ensemble.R
```diff
@@ -0,0 +1,239 @@
...
+##' quantiles <-list(
+##' q5 = c(0.127,0.034,0.052), # 5th percentile for each category: sand, clay, and silt at one site and one depth
+##' q50 = c(0.615,0.15,0.191), # 50th percentile (median) for each category: sand, clay, and silt at one site and one depth
+##' q95 = c(0.799,0.66,0.616))  # 95th percentile for each category: sand, clay, and silt at one site and one depth
+##' alpha_est <- estimate_dirichlet_parameters(means, quantiles)
+##' }
+##'  @return The individual alphas that work best to fit the observed quantiles
+##'  @export 
+##'  @author Qianyu Li
+estimate_dirichlet_parameters <- function(means, quantiles) {
```
dlebauer, line 21: I don't think it should be combined with the existing function `prior_fn` (that is similar in approach but for univariate distributions though it has its own limitations). But perhaps it should be included in the same file in the priors package?

https://github.com/PecanProject/pecan/blob/edc14b81a8da4fe71f93602137ddf0dd0308a4d1/modules/priors/R/priors.R#L90

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -490,22 +489,36 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+      if ("volume_fraction_of_water_in_soil_at_saturation" %in% names(soil_IC_list$vals)) {
+         # Calculate the soilWHC for the whole soil profile in cm
+         soilWHC_total <- sum(unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])*thickness)
+         param[which(param[, 1] == "soilWHC"), 2] <- soilWHC_total
+         #LitterWHC in cm (top soil layer)
+         param[which(param[, 1] == "litterWHC"), 2] <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[1]*thickness[1]
+       }
+      if ("soil_hydraulic_conductivity_at_saturation" %in% names(soil_IC_list$vals)) {
+         #litwaterDrainrate in cm/day
+         param[which(param[, 1] == "litWaterDrainRate"), 2] <- unlist(soil_IC_list$vals["soil_hydraulic_conductivity_at_saturation"])[1]*100/(3600*24)
```
DongchenZ:
```suggestion
         param[which(param[, 1] == "litWaterDrainRate"), 2] <- PEcAn.utils::ud_convert(unlist(soil_IC_list$vals["soil_hydraulic_conductivity_at_saturation"])[1], "m s-1", "cm day-1")
```

DongchenZ: Or you can check if the `ud_convert` function provides this conversion.

modules/uncertainty/R/ensemble.R
```diff
@@ -461,12 +463,12 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
                            settings = settings,
                            new.state = new.state[i, ], 
                            new.params = new.params[[i]], #new.params$`646`[[i]] for debugging
-                           inputs =list(met=list(path=inputs$samples[[i]])), 
+                           inputs =list(met=list(path=inputs$met$samples[[i]]),soilinitcond=list(path=inputs$soilinitcond$samples[[i]])),
```
DongchenZ: This line of code, which requires the `met` and `soilinitcond` paths inside the `restart` object, doesn't fit the creation of the `restart.list` object in the `sda.enkf_MultiSite` function (see L452).

Qianyuxuan: If I understood your question correctly, my latest edits seemed to have already fixed it. Firstly, "soilinitcond" was added in the XML file in "Create_Multi_settings.R". Then "inputs" were generated for all tags under the "ensemble" here: https://github.com/PecanProject/pecan/pull/3406/files#diff-8678ad23f7d92ce4d8c0e81c8a51298c2b7a1ee76759f9c848fd54492f533c9e. Finally, "inputs" was passed to "samples" when t=1 and restart=FALSE (https://github.com/PecanProject/pecan/blob/6cbf62b6cf78e129a938c7a9a0d3258f51d86748/modules/assim.sequential/R/sda.enkf_MultiSite.R#L478), and restart.list when t>1 here: https://github.com/PecanProject/pecan/blob/6cbf62b6cf78e129a938c7a9a0d3258f51d86748/modules/assim.sequential/R/sda.enkf_MultiSite.R#L452. Let me know if you still have concerns.

models/sipnet/R/write.configs.SIPNET.R
mdietze: per previous conversation, I recommend adding a check to make sure that `length(settings$run$inputs$soilinitcond$path) == 1` and throwing an error otherwise

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,33 @@ sda.enkf.multisite <- function(settings,
       #reformatting params
       new.params <- sda_matchparam(settings, ensemble.samples, site.ids, nens)
     }
-      #sample met ensemble members
-      #TODO: incorporate Phyllis's restart work
```
mdietze, line 359: please retain this comment, which will be helpful to the folks working on uncertainty partitioning.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,33 @@ sda.enkf.multisite <- function(settings,
...
+   #performing the sampling
+   inputs <- vector("list", length(conf.settings))
+   # For the tags specified in the xml I do the sampling
+   for (s in seq_along(conf.settings)){
+      if (is.null(inputs[[s]])) {
+        inputs[[s]] <- list() 
+      }
+     for (i in seq_along(samp.ordered)){
+         #call the function responsible for generating the ensemble
+        inputs[[s]][[names(samp.ordered)[i]]] <- input.ens.gen(settings=conf.settings[[s]],
```
mdietze: is this sampling occurring for each site independently or across all sites simultanously?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -379,7 +395,42 @@ sda.enkf.multisite <- function(settings,
       if (t>1){
         #for next time step split the met if model requires
         #-Splitting the input for the models that they don't care about the start and end time of simulations and they run as long as their met file.
-        inputs.split <- metSplit(conf.settings, inputs, settings, model, no_split = FALSE, obs.times, t, nens, restart_flag = FALSE, my.split_inputs)
+        #set start and end date for splitting met
```
mdietze: Conceptually, I'm not seeing why you're replacing a simple call to a function with a big chunk of code, which undo's the concept of modularization and refactoring. If there's something wrong within the function, I'd fix the function.

Also, conceptually note that MOST models don't need their met split, so make sure that any implementation of metSplit

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -379,7 +395,42 @@ sda.enkf.multisite <- function(settings,
...
+        }else{
+          stop.time = obs.times[t]
+        }
+        
+        
+        #-Splitting the input for the models that they don't care about the start and end time of simulations and they run as long as their met file.
+        inputs.split <- 
+          furrr::future_pmap(list(conf.settings %>% `class<-`(c("list")), inputs, model), function(settings, inputs, model) {
+            # Loading the model package - this is required bc of the furrr
+            library(paste0("PEcAn.",model), character.only = TRUE)
```
mdietze: We shouldn't have to load library's like this. But if you do, load it once, before the pmap, not repeatedly.

modules/assim.sequential/inst/MultiSite-Exs/SDA/Create_Multi_settings.R
mdietze: Code still retains lhc instead of sampling

modules/uncertainty/R/ensemble.R
```diff
@@ -393,6 +394,7 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
           "site        : ", settings$run$site$name, "\n",
           "site  id    : ", format(settings$run$site$id, scientific = FALSE), "\n",
           "met data    : ", samples$met$samples[[i]], "\n",
+          "soil parameter   : ", samples$soilinitcond$samples[[i]], "\n",
```
mdietze: This line will fail for any model that doesn't have soilinitcond. You should dynamically detect the set of sampled inputs present

modules/uncertainty/R/ensemble.R
mdietze: Here you want to pass in the sampled set of inputs, regardless of what variables are sampled, rather than hard coding a specific subset. Not every model even has the concept of soilinitcond. Plus, if you add new inputs you'd have to change these hard coded variables.

modules/uncertainty/R/ensemble.R
```diff
@@ -461,12 +463,12 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
                            new.state = new.state[i, ], 
                            new.params = new.params[[i]], #new.params$`646`[[i]] for debugging
-                           inputs =list(met=list(path=inputs$samples[[i]])), 
+                           inputs =list(met=list(path=inputs$met$samples[[i]]),soilinitcond=list(path=inputs$soilinitcond$samples[[i]])), 
                            RENAME = rename)#for restart from previous model runs, not sharing the same outdir
       )
     }
     params<-new.params
-    return(invisible(list(runs = data.frame(id=run.id), ensemble.id = ensemble.id, samples=list(met=inputs)
+    return(invisible(list(runs = data.frame(id=run.id), ensemble.id = ensemble.id, samples=list(met=inputs$met,soilinitcond=inputs$soilinitcond)
```
mdietze: same need to generalize to any set of inputs

mdietze (changes requested):

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: Previously, I directly modified "metSplit" but it did not seem to be suggested. The challenge is it that we only need to do the split function for "met" sublist but now it lacks the specification: https://github.com/PecanProject/pecan/blob/develop/modules/assim.sequential/R/metSplit.R#L39. Do you suggest we change it to:"           inputs.split$met$samples[i] <- do.call(
            my.split_inputs,
            args = list(
              settings = settings,
              start.time = (lubridate::ymd_hms(start.time, truncated = 3) + lubridate::second(lubridate::hms("00:00:01"))),
              stop.time =   lubridate::ymd_hms(stop.time, truncated = 3),
              inputs = inputs$met$samples[[i]])

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: For each site independently.

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: Has removed this block of code.

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: Fixed.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: Fixed.

modules/uncertainty/R/ensemble.R
Qianyuxuan: Fixed.

models/sipnet/R/write.configs.SIPNET.R
mdietze: @Qianyuxuan sorry for not clarifying, the request isn't to have the code fail if thickness is missing, but to to only have depth/thickness within the if, to take all the other variable out of the if/else, and to use a default or user-specified soil depth if one isn't present in the file.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -490,20 +489,41 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    # Calculate the thickness of soil layers based on the assumption that the depth values are at bottoms and the first layer top is at 0
+    if ("depth" %in% names(soil_IC_list$dims)) {
+      thickness<-c(soil_IC_list$dims$depth[1],diff(soil_IC_list$dims$depth))
+      #SoilWHC
+      if ("volume_fraction_of_water_in_soil_at_saturation" %in% names(soil_IC_list$vals)) {
+         # Calculate the soilWHC for the whole soil profile in cm
+         soilWHC_total <- sum(unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])*thickness)
+         param[which(param[, 1] == "soilWHC"), 2] <- soilWHC_total
+         #LitterWHC in cm (top soil layer)
+         param[which(param[, 1] == "litterWHC"), 2] <- unlist(soil_IC_list$vals["volume_fraction_of_water_in_soil_at_saturation"])[1]*thickness[1]
```
mdietze: We should be very careful here  around assumptions about the inputs -- depending on the soil product there's no guarantee that the first thickness will be the litter layer. Indeed, I could easily see cases where the first layer is 0-30 cm or where there's only 1 layer provided (e.g. first and only thickness is 0-100). Also, your current code seems to double count the first layer in both the soilWHC and litterWHC.

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: Is independent sampling (which is NOT what we want) something new here or are you just continuing the status quo. If it's the status quo I can see punting the fix to a new PR, but if you moved from joint to independent sampling you'll need to revert it back to joint.

modules/uncertainty/R/ensemble.R
```diff
@@ -381,6 +382,20 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
       }
       dir.create(file.path(settings$rundir, run.id), recursive = TRUE)
       dir.create(file.path(settings$modeloutdir, run.id), recursive = TRUE)
+
```
mdietze, line 385: nice generalization!

modules/uncertainty/R/ensemble.R
```diff
@@ -461,12 +474,12 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
                            settings = settings,
                            new.state = new.state[i, ], 
                            new.params = new.params[[i]], #new.params$`646`[[i]] for debugging
-                           inputs =list(met=list(path=inputs$samples[[i]])), 
+                           inputs = input_list,
```
mdietze, line 477: :+1:

modules/assim.sequential/R/sda.enkf_MultiSite.R
Qianyuxuan: I didn't do anything new here and just continued the status quo. The ensemble size for met is 10 at each site, does it mean we may need to sample from the joint distribution using bootstrapping?

mdietze: No, what I mean is that if model ensemble member 1 at site 1 samples met ensemble k (e.g. 6) then model ensemble member 1 needs to also use met ensemble member k at sites 2 through 6400. Even though the specific met data is different from site to site, the spatiotemporal values within a specific met ensemble member share a covariance structure.  Similarly if ensemble member 1 at site 1 samples a particular parameter vector with a specific row number, then all other sites need to use the same posterior row number (though row k for one PFT will obviously point to a different set of parameters than row k of another PFT). By having the same parameters, met, etc across sites we induce the correct spatiotemporal covariance structure across sites.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -358,30 +358,42 @@ sda.enkf.multisite <- function(settings,
...
+                                                        parent_ids=NULL)
+    #replicate the same ids for the remaining sites
+    for (s in seq_along(conf.settings)) {
+      if (s!= random_site) {
+        if (is.null(inputs[[s]])) {
+          inputs[[s]] <- list() 
+        }
+        input_path <- conf.settings[[s]]$run$inputs[[tolower(input_tag)]]$path
+        inputs[[s]][[input_tag]]$ids<-inputs[[random_site]][[input_tag]]$ids
+        inputs[[s]][[input_tag]]$samples<- input_path[inputs[[random_site]][[input_tag]]$ids]
```
mdietze, line 392: This is a good start. Tagging @blesson07asd so he's aware of these changes and can account for them when he further generalizes ensemble sampling. Eventually this will need to be expanded to account for parameter sampling too, but we can do that later.

mdietze (approved):
Thanks @Qianyuxuan for all your hard work on this PR!

DongchenZ (approved):
Great work @Qianyuxuan

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -355,17 +355,47 @@ sda.enkf.multisite <- function(settings,
...
+  inputs <- vector("list", length(conf.settings))
+  #for the tags specified in the xml, do the sampling for a random site and then replicate the same sample ids for the remaining sites for each ensemble member 
+  for (i in seq_along(samp.ordered)) {
+    random_site <- sample(1:length(conf.settings),1)
+    if (is.null(inputs[[random_site]])) {
+      inputs[[random_site]] <- list() 
+    }
+    input_tag<-names(samp.ordered)[i]
+    #call the function responsible for generating the ensemble for the random site
+    inputs[[random_site]][[input_tag]] <- input.ens.gen(settings=conf.settings[[random_site]],
```
mdietze:
```suggestion
    inputs[[random_site]][[input_tag]] <- PEcAn.uncertainty::input.ens.gen(settings=conf.settings[[random_site]],
```

mdietze:
@Qianyuxuan there's a small fix that needs to occur in data.land/DESCRIPTION to add the following new libraries to Suggests:   ‘doSNOW’ ‘foreach’ ‘MCMCpack’.  After that you'll probably need to run scripts/generate_dependencies.R and add any updated files to the PR (e.g., docker/depends/pecan_package_dependencies.csv)

PR #3409, opened 2025-01-07
SoilGrids: avoid zero-row result when no sites are NA (+ clean up df names)

infotroph:
## Description
One bugfix and two minor code-cleanup tweaks in `soilgrids_soilC_extract()`:

* When the results don't contain any sites that return NAs, the overall result was truncated to zero rows. This is because `which(vector_with_all_items_evaluating_to_FALSE)` returns  an empty integer vector, then `df[-integer()]` always returns zero rows (despite the minus sign it gets parsed as "select nothing from df", not as "remove nothing from df"). Fixed by passing the logical vector directly instead of using `which()`.
* Made column names in `internal_site_info` match those in `site_info` for symmetry/ease of typing and remembering.
* Wrapped the layer retrieval code in a little helper function so it's easier to see that the parallel and sequential approaches really do exactly the same thing, just with or without wrapping it in furrrr

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.land/R/soilgrids_soc_extraction.R
+++ b/modules/data.land/R/soilgrids_soc_extraction.R
@@ -65,10 +65,10 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
   }
   
   # prepare site info for extraction
-  internal_site_info <- data.frame(site_info$site_id, site_info$site_name, site_info$lat,site_info$lon)
+  internal_site_info <- site_info[, c("site_id", "site_name", "lat", "lon")]
   #create a variable to store mean and quantile of organic carbon density (ocd) for each soil depth
-  ocdquant <- matrix(NA, nrow = 6, ncol = length(internal_site_info$site_info.lon) * 4) #row represents soil depth, col represents mean, 5%, 50% and 95%-quantile of ocd for all sites 
-  lonlat <- cbind(internal_site_info$site_info.lon, internal_site_info$site_info.lat)
+  ocdquant <- matrix(NA, nrow = 6, ncol = length(internal_site_info$lon) * 4) #row represents soil depth, col represents mean, 5%, 50% and 95%-quantile of ocd for all sites
+  lonlat <- cbind(internal_site_info$lon, internal_site_info$lat)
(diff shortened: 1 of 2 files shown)


modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -65,10 +65,10 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
   }
   
   # prepare site info for extraction
-  internal_site_info <- data.frame(site_info$site_id, site_info$site_name, site_info$lat,site_info$lon)
+  internal_site_info <- site_info[, c("site_id", "site_name", "lat", "lon")]
```
infotroph, line 68: Mostly to simplify the column names when working with them below, but has a nice side benefit: If the user passes extra columns in `site_info` that this function doesn't need, the code will now ignore them rather than complain.

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -101,12 +102,9 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
...
     ocdquant[dep, ] <- ocd_real[dep.ind] %>% unlist
   }
-  na.ind <- which(is.na(ocdquant[1, 1:length(site_info$site_id)]))
-  internal_site_info <- data.frame(site_info$site_id[-na.ind], 
-                                   site_info$site_name[-na.ind], 
-                                   site_info$lat[-na.ind],
-                                   site_info$lon[-na.ind]) %>% `colnames<-`(names(site_info))
-  
+  na_sites <- is.na(ocdquant[1, seq_len(nrow(internal_site_info))])
+  internal_site_info <- internal_site_info[!na_sites, ]
```
infotroph, line 106: Here's the bugfix: `x[-which(all_FALSEs), ]` returns an empy df, but `x[!all_FALSEs, ]` returns x as desired.

modules/data.land/R/soilgrids_soc_extraction.R
infotroph: Edit for clarity: The complaint about extra columns didn't happen here, but in the `colnames<-` on now-removed line 108.

dlebauer (approved):

PR #3414, opened 2025-01-15
Add Doxygen to sipnet action

dlebauer:
Fixes build error https://github.com/PecanProject/pecan/actions/runs/12740766141/job/35506419353#step:11:43

By installing Doxygen. Similar to this commit to the SIPNET repository. https://github.com/PecanProject/sipnet/commit/f379cb9a95f5790a465aa15bac7d7ce26170c190

Perhaps a 'cleaner' option (so that PEcAn tests don't depend on SIPNET's build process would be for SIPNET's action to automatically compile and store a binary that can later be accessed by this GHA. But this is the easiest, quickest fix.

### DIFF
--- a/.github/workflows/sipnet.yml
+++ b/.github/workflows/sipnet.yml
@@ -38,7 +38,7 @@ jobs:
 
     # install additional tools needed
     - name: install utils
-      run: apt-get update && apt-get install -y doxygen postgresql-client qpdf
+      run: apt-get update && apt-get install -y postgresql-client qpdf
     - name: install new dependencies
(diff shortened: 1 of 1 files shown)


infotroph (changes requested):
I don't think installing Doxygen on a headless CI system where nobody will read any docs is the right approach here. I'm used to seeing Doxygen-using builds call it as a separate `make document` that *isn't* part of the default Make run, but if you're confident having Sipnet build docs by default is the right design then let's use `make sipnet` on GHA

dlebauer:
Make sipnet makes sense 

fixed 9d0212475532fede363f6cc67907bc16873ec61f

infotroph (approved):

PR #3419, opened 2025-01-23
take deciduous status from fracLeafFall instead of PFT name + handle missing database$bety$write

infotroph:
## Description

Found a few more places where it was ambiguous whether a NULL settings$database$bety$write should be treated as TRUE or FALSE, and set them to treat it as FALSE. For consistency I also changed the one place I saw that was treating missing as TRUE -- If I was following previous discussion in #3398, we've at least implicitly decided missing should mean do not write.

<del>Also: write.configs.SIPNET() now respects `run$site$site.pft` and uses it preferentially to `run$inputs$pft.site`. If anyone feels strongly I'd be open to switching the priority so pft.site wins if both are defined -- I picked this way because it felt like it was keeping effects best localized to the place they were set.</del>

Edit to the above: After iterating in review, write.config.SIPNET does "now respect site PFTs," but by a completely different method: Instead of consulting either `run$site$site.pft` or `run$inputs$pft.site`, it now extracts information about deciduous/evergreen status from the trait values that are passed to it. These were calculated from PFT identity in an upstream step.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/settings/R/check.all.settings.R
```diff
@@ -1024,8 +1024,9 @@ check.database.settings <- function(settings) {
       # should runs be written to database
       if (is.null(settings$database$bety$write)) {
         PEcAn.logger::logger.info(
-          "Writing all runs/configurations to database.")
-        settings$database$bety$write <- TRUE
+          "database$bety$write not set.",
+          "Will not write any runs/configurations to database.")
+        settings$database$bety$write <- FALSE
```
infotroph, line 1029: Here's the change in existing behavior: when `settings$database$bety$write` was undefined it used to default to TRUE, but defaulting to FALSE means DB-free runs can delete the entire settings$database section.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -560,12 +560,17 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-              param[which(param[, 1] == "laiInit"), 2] <- 0       
-          }
+      #Growing seasons are coarsely defined as months from May to September for non-conifers in the US
+      if (!(lubridate::month(settings$run$start.date) %in% seq(5, 9))) {
+        site.pft.name <- settings$run$site$site.pft
+        if (is.null(site.pft.name)) {
+          site_pft <- utils::read.csv(settings$run$inputs$pft.site$path)
+          site.pft.name <- site_pft$pft[site_pft$site == settings$run$site$id]
+        }
+        if (site.pft.name != "boreal.coniferous") {
```
mdietze: I know this is just a reorganization of existing code, but the hard-coded PFTs and Months here is dangerous because it represents hidden assumptions that are now getting applied at continental scales without the users being aware of these assumptions.

mdietze (approved):

mdietze:
I think I prefer the idea that if the pft.site file is defined it wins. Otherwise you could end up in a situation in multisite runs where some sites use pft.site and others use site$pft, which doesn't seem like the intended behavior and would make it hard to extract exactly what PFT was run at each site. In general, pft.site shouldn't be used for single-site runs, but is awfully handy for multisite runs.

base/settings/R/check.all.settings.R
infotroph: Turns out we have existing test coverage for this! I'll update the test with the new behavior.

models/sipnet/R/write.configs.SIPNET.R
infotroph: How much additional work would it be to remove these? Do existing PFT definitions contain information we could use to pick the right LAI here without the hard-coding?

infotroph:
@mdietze I've switched the logic to make pft.site win when both are defined. Thoughts on whether to merge now or tackle the hard-coded PFT/month logic?

models/sipnet/R/write.configs.SIPNET.R
mdietze: So we should be able to figure out the phenological type of a PFT based on it's parameters. I'm not entirely sure, off the top of my head, how the evergreen PFTs have been parameterized as evergreen, but for example I would expect `fracLeafFall` to be a number near 1 for deciduous species and a number <0.5 for evergreen (evergreens often do have a seasonal leaf phenology, but they might drop more like 25% of their LAI in the winter). Similarly, rather than hard coding dates we should look at the MODIS phenology file (if provided) or the leaf on/off parameters to know whether (for any DOY) we should initialize a specific site in a leaf on or off state

infotroph: You know, what's the chance we could push this logic all the way into a different workflow? If the fundamental question is "how many leaves should we assume we're starting from", why not answer it when creating the IC pool files rather than when reading them in to Sipnet?

mdietze: I get that argument, but I think there are a number of clear disadvantages. First, routine use of IC files in PEcAn runs is very new and I don't think we can assume it's universally adopted. Second, this would be a HUGE increase in the required temporal precision of initialization. Right now it's not uncommon to grab the "nearest available" IC data, which for vegetation and C pools can sometimes be up to a decade shifted in time. Pushing phenology into the IC means having to make IC files that are specific to an individual start date, and rebuilding the files if you change the start date at all. To me, adding in a simple check to see if, for the specified start date, whether the site should be in a leaf-on or leaf-off state isn't a huge burdern for write.configs that pays off with greater IC flexibility.

infotroph: @mdietze I took a stab at implementing this as discussed -- can you review, please?

Key points to note:
- As written, leaf on and leaf off days will come from template.param unless phenology is provided (or I guess unless passed as a trait or constant), so expect them to be DOY 144 and 285 by default. This will change behavior in simulations that start between May 1 and 24 (was leaf-on, now leaf-off) or Oct 1 and 13 (was leaf-off, now leaf-on).
- Both the previous hack and this update are only done when reading from `settings$run$inputs$poolinitcond` and not if `IC` is passed directly into write.config.SIPNET. Would it make sense to move it out of the conditional block so that initial LAI gets adjusted no matter how it was set?
- I don't fully understand how the loop over PFTs in lines 147-471 works. I don't _think_ that affects the needed solution for LAI, but please be alert for anywhere it could have made me misunderstand the data structures. Specifically:
     - I don't see any change in the assignment targets on successive loop iterations. Do the `param` values just get overwritten by each PFT and the last one wins?
     - In what cases will  write.configs.SIPNET be passed a `trait.values` that contains more than one pft in the first place? If that happens, do we need a way to handle configs for an evergreen PFT and a deciduous one in the same call to write.config.SIPNET?

infotroph:
[Hoisting this comment out of the previous review thread because GitHub seems to have made it hard to see now that the commented line has changed. Apologies if this is a duplicate notification.]

@mdietze I took a stab at implementing the fracLeafFall / leafOnDay lookup as discussed -- can you review, please?

Key points to note:
- As written, leaf on and leaf off days will come from template.param unless phenology is provided (or I guess unless passed as a trait or constant), so expect them to be DOY 144 and 285 by default. This will change behavior in simulations that start between May 1 and 24 (was leaf-on, now leaf-off) or Oct 1 and 13 (was leaf-off, now leaf-on).
- Both the previous hack and this update are only done when reading from `settings$run$inputs$poolinitcond` and not if `IC` is passed directly into write.config.SIPNET. Would it make sense to move it out of the conditional block so that initial LAI gets adjusted no matter how it was set?
- I don't fully understand how the loop over PFTs in lines 147-471 works. I don't _think_ that affects the needed solution for LAI, but please be alert for anywhere it could have made me misunderstand the data structures. Specifically:
     - I don't see any change in the assignment targets on successive loop iterations. Do the `param` values just get overwritten by each PFT and the last one wins?
     - In what cases will  write.configs.SIPNET be passed a `trait.values` that contains more than one pft in the first place? If that happens, do we need a way to handle configs for an evergreen PFT and a deciduous one in the same call to write.config.SIPNET?

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -559,14 +567,27 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+      # Sipnet always starts from initial LAI whether day 0 is in or out of the
+      # growing season -> set LAI=0 when a deciduous PFT starts with leaves off
+      #
+      # Note: At this writing in Jan 2025, leafOnDay and LeafOffDay are taken
+      # from the model defaults (template.param) unless:
+      # - settings$run$inputs$leaf_phenology is provided, or
+      # - the PFT sets leafOnDay/leafOffday as traits.
+      # So unless you set something different, it's probably using DOY 144/285
+      # ==> leaves are on from late May through mid-October.
+      is_deciduous_pft <- isTRUE(param[param[, 1] == "fracLeafFall", 2] < 0.5)
```
mdietze: I think the logic here is reverse. If a PFT is deciduous then leaf fall should be 1, while non-deciduous species should be near zero.

More complicated would be if someone tried to capture evergreen phenology with an intermediate value (e.g., 0.3 leaf loss would be reasonable for many confiers) you'd still need to invoke phenology but you wouldn't want to start from LAI = 0. LAI*fracLeafFall is an option, as noted in a comment below, but not only does that assume the LAI IC is near peak, but also that the LAI isn't already being nudged by the SDA (which, presumably, would see the seasonal cycling of LAI)

mdietze (commented):
Other than the one sign error, the phenology update looks good to me (definitely a lot better than what was already there!). Tagging @Qianyuxuan to make sure what is happening here isn't going to have unintended consequences for how the SDA works on seasonal timescales (e.g. monthly assimilation).

models/sipnet/R/write.configs.SIPNET.R
infotroph: Fixed in 8b6f55b

PR #3421, opened 2025-01-23
Fix "No Visible Binding for Global Variable" Warnings in lpjguess

harshagr70:
This PR focuses on addressing "no visible binding for global variable" warnings specifically within the lpjguess package of the Pecan codebase , these warnings were not specifically from the tidyverse calls . These issues were identified during rcmdcheck::rcmdcheck() runs, and while some changes have been implemented, further guidance is needed for unresolved errors. , fixes a part of the issue #2758 

Summary of Changes:

Scoped Variables:
Passed the ZZ variable as an argument in the relevant functions to resolve scoping issues flagged by R CMD check.

Namespace Updates:
Added required imports to the NAMESPACE file as suggested by rcmdcheck.

Log File Updates:
Updated the .log file to reflect the removal of errors for the issues that were fixed in this PR.

The following warnings remain unresolved and require input from maintainers:

write.insfile.LPJGUESS:
No visible binding for global variable: lpjguess_param_list.
No visible binding for global variable: co2.1850.2020.
Undefined global functions or variables:
co2.1850.2020 and lpjguess_param_list

These warnings may need clarification on:
Whether lpjguess_param_list and co2.1850.2020 should be explicitly declared as global variables or passed as arguments.


mdietze (approved):

models/lpjguess/NAMESPACE
```diff
@@ -6,6 +6,9 @@ export(pecan2lpjguess)
 export(readStateBinary)
 export(write.config.LPJGUESS)
 export(write.insfile.LPJGUESS)
+importFrom(utils,data)
+importFrom(utils,read.table)
+importFrom(utils,write.table)
```
infotroph: Thanks for working on this! Instead of an importFrom directive let's follow PEcAn house style by finding these calls where they appear in the code and adding the utils namespace qualifier, so that `write.table(df, filename,...)` becomes `utils::write.table(df, filename, ...)` and so on. (In concept we'd like to do the same for all instances of the `ncdf4::nc*` functions imported below, but that can be a separate project)

But FYI we do use importFrom for cases that don't work at all in `::` form (e.g. rlang's `.data` pronoun) or that do work but become a hassle (e.g. infix functions like `%>%`), so feel free to propose an importFrom where it seems like the right move. 

However in those cases be aware you'll need to use Roxygen directives (e.g. `#' @importFrom utils write.table` in the Roxygen comments for the function that uses write.table) rather than edit NAMESPACE directly -- as written Roxygen will overwrite this change the next time it runs.

models/lpjguess/R/readStateBinary.LPJGUESS.R
```diff
@@ -208,13 +208,13 @@ readStateBinary <- function(out.path, npft){
...
-    Patchpft_list[[pat]] <- getClass_Patchpft()
-    Vegetation_list[[pat]] <- getClass_Vegetation()
-    Individual_list[[pat]] <- getClass_Individual(Vegetation_list[[pat]]$indv)
-    Soil_list[[pat]] <- getClass_Soil()
-    Sompool_list[[pat]] <- getClass_Sompool()
-    SompoolCent_list[[pat]] <- getClass_SompoolCent() 
-    Fluxes_list[[pat]] <- getClass_Fluxes()
+    Patchpft_list[[pat]] <- getClass_Patchpft(zz,npft)
+    Vegetation_list[[pat]] <- getClass_Vegetation(zz)
+    Individual_list[[pat]] <- getClass_Individual(Vegetation_list[[pat]]$indv,zz)
```
infotroph: 👍 to this solution to the scoping problem. Nit: Spaces after commas, please.

```suggestion
    Patchpft_list[[pat]] <- getClass_Patchpft(zz, npft)
    Vegetation_list[[pat]] <- getClass_Vegetation(zz)
    Individual_list[[pat]] <- getClass_Individual(Vegetation_list[[pat]]$indv, zz)
```

models/lpjguess/R/readStateBinary.LPJGUESS.R
```diff
@@ -226,7 +226,7 @@ readStateBinary <- function(out.path, npft){
 
 
 ##################### Class : Patchpft #####################
-getClass_Patchpft <- function(){
+getClass_Patchpft <- function(zz,npft){
```
infotroph:
```suggestion
getClass_Patchpft <- function(zz, npft){
```

models/lpjguess/R/readStateBinary.LPJGUESS.R
```diff
@@ -380,7 +380,7 @@ getClass_Vegetation <- function(){
 
 
 ##################### Class : Individual #####################
-getClass_Individual <- function(nind){      #  nind <- Vegetation$indv 
+getClass_Individual <- function(nind,zz){      #  nind <- Vegetation$indv
```
infotroph:
```suggestion
getClass_Individual <- function(nind, zz){      #  nind <- Vegetation$indv 
```

infotroph (changes requested):

models/lpjguess/NAMESPACE
harshagr70: Thank you for the feedback! I’ll update the code to use utils::write.table and similar namespace qualifiers as per the PEcAn house style. I appreciate the clarification on using importFrom with Roxygen directives when needed—I'll keep this in mind for future contributions. I’ll make the changes and push the updates shortly. Thanks again !!

harshagr70:
@infotroph  i have added the changes as suggested !! and also checked for any errors by R CMD check . This fixes warnings  in the lpjguess package but there are two more warnings on which i need your guidance , 
Screenshot 2025-01-24 at 11 59 34 PM
Whether lpjguess_param_list and co2.1850.2020 should be explicitly declared as global variables or any other alternative fix for this !!

infotroph:
@harshagr70 both `lpjguess_param_list` and `co2.1850.2020` are created by `write.config.LPJGUESS()` loading files into the global environment, specifically the calls to `load(...)` on line 101 and `data(...)` on line 186. The fix is to load them into their own environments: 

```
co2_data <- new.env()
data(co2.1850.2020, package = "PEcAn.LPJGUESS", envir = co2_data)
co2.1850.2020 <- co2_data$co2.1850.2020
```

`load(...)` can use the same pattern to become `load(..., envir = some_env)`, and we have implemented the wrapper `PEcAn.utils::load_local` to help with this, but do note that it returns a list (because Rdata files are allowed to contain multiple objects), so you'll still need to decide whether it's cleaner to immediately extract the single object: 

```
lpjg_param_data <- PEcAn.utils::load_local(...)
lpjguess_param_list <- lpjg_param_data$lpjguess_param_list
[...]
for(t in seq_along(lpjguess_param_list)){
[...]
```

...or to pass the list around to the places that use the single object:
```
lpjg_param_data <- PEcAn.utils::load_local(...)

[...]
for(t in seq_along(lpjg_param_data$lpjguess_param_list)){
[...]
```
We have code that uses both patterns and I go back and forth on which one I prefer.

harshagr70:
Thank you @infotroph  for the detailed explanation and guidance. I will apply the suggested changes to load co2.1850.2020 and lpjguess_param_list into their respective environments, as outlined. I appreciate your assistance and will proceed with the implementation accordingly.

harshagr70:
@infotroph this package is now free of prompts of "no visible bindings for global variable " , made the changes as per your suggestions , Thankyou !!

models/lpjguess/R/write.config.LPJGUESS.R
```diff
@@ -183,7 +186,8 @@ write.insfile.LPJGUESS <- function(settings, trait.values, rundir, outdir, run.i
   if (end.year < 1850) {
     CO2 <- data.frame(start.year:end.year, rep(280, n.year))
   } else if (end.year < 2021) {
-    data(co2.1850.2020, package = "PEcAn.LPJGUESS")
+    utils::data(co2.1850.2020, package = "PEcAn.LPJGUESS", envir = co2_data)
```
infotroph, line 187: @harshagr70 unless I missed where it's defined above, I think you need to define `co2_data` as a new environment before assigning into it:
```suggestion
    co2_data <- new.env()
    utils::data(co2.1850.2020, package = "PEcAn.LPJGUESS", envir = co2_data)
```

models/lpjguess/tests/Rcheck_reference.log
```diff
@@ -46,29 +46,13 @@ Strong dependencies not in mainstream repositories:
...
 write.insfile.LPJGUESS: no visible binding for global variable
   ‘co2.1850.2020’
-write.insfile.LPJGUESS: no visible global function definition for
-  ‘write.table’
 Undefined global functions or variables:
-  co2.1850.2020 data lpjguess_param_list npft read.table write.table zz
-Consider adding
-  importFrom("utils", "data", "read.table", "write.table")
-to your NAMESPACE file.
+  co2.1850.2020 lpjguess_param_list
```
infotroph: You've fixed these now too! 🎉 
I think all of lines 49-60 can now be deleted and line 48 changed to `* checking R code for possible problems ... OK`

models/lpjguess/R/write.config.LPJGUESS.R
harshagr70: @infotroph i defined it on line 16 , but i guess it would be much better to define it just before we assign it

models/lpjguess/tests/Rcheck_reference.log
harshagr70: @infotroph I missed up on deleting those errors from the reference.log , i will make the relevant changes !! thankyou for the review  !!

models/lpjguess/R/write.config.LPJGUESS.R
infotroph: So you did! Yeah, I think since this is effectively a temporary variable that we write to once and immediately read back out to `co2.1850.2020`, defining it just before use is clearer.

harshagr70:
@infotroph added suggested changes, i guess its complete now !!

models/lpjguess/tests/Rcheck_reference.log
```diff
@@ -45,30 +45,7 @@ Strong dependencies not in mainstream repositories:
...
-write.insfile.LPJGUESS: no visible binding for global variable
-  ‘co2.1850.2020’
-write.insfile.LPJGUESS: no visible global function definition for
-  ‘write.table’
-Undefined global functions or variables:
-  co2.1850.2020 data lpjguess_param_list npft read.table write.table zz
-Consider adding
-  importFrom("utils", "data", "read.table", "write.table")
-to your NAMESPACE file.
+* checking R code for possible problems ... OK
```
infotroph: Sorry, I think I botched the line numbers 🤦 -- you can delete these next four lines as well (49-53; from this empty line down through "See section ..."), so that what's  left after deletion is 
```
* checking R code for possible problems ... OK
* checking Rd files ... OK

infotroph (approved):

models/lpjguess/tests/Rcheck_reference.log
infotroph: (This is me giving GitHub a grumpy look for making it so hard to comment on blocks of lines that aren't yet marked as changed)

infotroph:
Thank you for these fixes, @harshagr70!

harshagr70:
Welcome @infotroph , i will go ahead and fix the remaining packages as well , i will make sure to re-check the changes this time to save your time as well !! 😊

PR #3423, opened 2025-01-26
[WIP] Proposed vector data format and application to LandIQ data

dlebauer:
## Description

This PR introduces two new functions to the PEcAn.data.land package:

1. landiq2std: Processes LandIQ crop map Shapefiles into a standardized format with a GeoPackage + CSV, including mapping crops to PFTS.
2. shp2gpkg: Converts a Shapefile to a GeoPackage while ensuring spatial data integrity and optional geometry repairs. This is mostly a helper function

#### The goal here is to propose a new format for handling geospatial data with large tables.

This is motivated by the CCMMF project, and the use of the LandIQ crop datasets in particular, but should be generalizable to other workflows that use vector geospatial data.

It is also motivated by the desire to decouple workflows from BETYdb and its associated dependence on Posgres+PostGIS that has often been more of a barrier than originally envisioned.

#### Other options:

- store tables in a relational geospatial database like SQLite or duckdb. A geopackage is the OGC standard alternative to shapefiles and is built on SQLite, to this is a viable alternative. The disadvantages here include 
  - 1. could get large with lots of data (though that could be solved by using multiple files and 
  - 2. CSVs are easier to use

#### Linking CSVs and GPKG

Spatial joins can be slow and we don't want to store geometries in the CSV (they are large as text and that would be redundant).

As proposed the tables are linked by an id generated as a hash by the `digest::digest()` function based on the geometry. This adds a new dependency (though it is already used in the api). 

An alternative to using the hash as an id would be to would be to store lat+lon of the centroid in both the GPKG and all associated CSVs. Then joins could be on lat+lon. 

This would have the advantage of allowing some (many) uses of the CSVs independent of spatial files and libraries. 

There is a nonzero chance that distinct geometries could have the same centroid (eg if two cells from raster's with different resolution and perhaps other edge cases.

TODO
- [ ] rename geom --> geometry
- [ ] incorporate code used to produce 2018-2023 dataset (see /projectnb/dietzelab/malmborg/CARB/CARB_R/*R)

## Types of changes

## Checklist:

- [ ] Add digest to DESCRIPTION (or replace with lat+lon as unique id)
- [x] Split into separate files (?)
- [ ] Write Vignette
- [ ] Create a separate issue to replace hardcoded PFT mapping with a CSV file that can use defaults in the data.land package or be provided by the user.


modules/data.land/R/shp2gpkg.R
```diff
@@ -0,0 +1,79 @@
...
+#' # Prevent overwriting existing layers
+#' shp2gpkg("data/roads.shp", "output/roads.gpkg", overwrite = FALSE)
+#'
+#' @importFrom sf st_read st_write
+#' @export
+shp2gpkg <- function(input_shp, output_gpkg, layer_name = NULL, overwrite = TRUE) {
+  # Load the Shapefile
+  shapefile <- sf::st_read(input_shp, quiet = TRUE)
+  
+  # Check validity of geometries
```
dlebauer, line 31: @camalmborg this is the logic for reparing geometries. The general approach within pecan is to bring in external data and immediately converted to a common standard. shp2gpkg can be part of that workflow, as it is here.

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,107 @@
...
+#' # Run the conversion function
+#' landiq2std(input_shp, output_gpkg, output_csv)
+#'
+#' @importFrom sf st_read st_transform st_centroid st_coordinates st_write
+#' @importFrom dplyr mutate select rename filter case_when distinct
+#' @export
+landiq2std <- function(input_shp, output_gpkg, output_csv) {
+  
+  # Check input file format
+  if (!grepl(pattern = "\.shp$", input_shp)) {
```
dlebauer:
```suggestion
  if (!grepl(pattern = "\\.shp$", input_shp)) {
```

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
+#' Convert a LandIQ Shapefile into a Standardized Format consisting of
```
infotroph: 1. In CCMMF we decided early on to refer to this dataset as the "DWR crop maps" (after the CA department that leads development of the maps and is responsible for ongoing upkeep) not "LandIQ" (the company that was contracted to generate this version); I'd suggest following that convention here.
2. In what sense is the result standardized?

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+#' a GeoPackage file to store geospatial information and an associated
+#' CSV file with attributes
+#'
+#' This function reads a LandIQ crop map shapefile downloaded from
+#' https://www.landiq.com/land-use-mapping and processes the data into
```
infotroph: If using DWR nomenclature, canonical URL is probably https://data.cnra.ca.gov/dataset/statewide-crop-mapping

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+#' a standardized GeoPackage and CSV format.
+#'
+#' @param input_file Character. Path to the input Shapefile
+#' (GeoPackage created by shp2gpkg is also valid)
+#' @param output_gpkg Character. Path to the output GeoPackage
+#' @param output_csv Character. Path to the output CSV.
+#'
+#' @return Invisibly returns a list with paths to the output files.
```
infotroph: Will the output paths ever differ from the ones specified as input? Might be more useful to return a success/failure status instead of information the user already has.

infotroph: (I also think returning invisibly is more often confusing than useful, but side-effecting functions like this are a place it can work)

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+#' # Specify input and output file paths
+#' input_file <- "path/to/landiq.shp"
+#' output_gpkg <- "path/to/your_output.gpkg"
+#' output_csv <- "path/to/your_output.csv"
+#'
+#' # Run the conversion function
+#' landiq2std(input_file, output_gpkg, output_csv)
+#'
+#' @importFrom sf st_read st_transform st_centroid st_coordinates st_write
+#' @importFrom dplyr mutate select rename filter case_when distinct
```
infotroph: Please remove imports and use `sf::...` and `dplyr::...` instead

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+landiq2std <- function(input_file, output_gpkg, output_csv) {
+  # Check input file format
+  if (grepl(pattern = "\\.shp$", input_file)) {
+    PEcAn.logger::logger.info(
+      "Converting Shapefile:", basename(input_file),
+      "to GeoPackage:", basename(output_gpkg)
+    )
+    # read in, repair geometries, write out repaired geopackage
+    shp2gpkg(input_file, output_gpkg, overwrite = TRUE)
+    input_file <- output_gpkg # now gpkg is input file
```
infotroph: This seems potentially unwanted if something fails later in the landiq2std function and leaves the user wondering why the output file was updated even though the command stopped with an error message. Consider waiting until success?

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  # Read the Shapefile
+  landiq_polygons <- sf::st_read(input_file, quiet = TRUE)
+
+  # Determine crop year from column name
+  crop_col <- grep("Crop[0-9]{4}", colnames(landiq_polygons), value = TRUE)
+  year <- gsub("Crop", "", crop_col)
```
infotroph: I don't recognize these from the DWR column names. Does this function  actually operate on raw DWR shapefiles as documented, or is it built around Charlotte's multi-year file?

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+
+  # Check required columns
+  required_cols <- c("Acres", crop_col, "Source", "Comments", "County", "geom")
+  missing_cols <- setdiff(required_cols, colnames(landiq_polygons))
+  if (length(missing_cols) > 0) {
+    stop("Input file is missing the following columns: ", paste(missing_cols, collapse = ", "))
```
infotroph: use logger instead of stop

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  # possible spedup by pre-computing lat and lon
+  landiq_polygons <- landiq_polygons |>
+    sf::st_transform(4326) |>
+    dplyr::mutate(
+      year = year,
+      lon = sf::st_coordinates(sf::st_centroid(geom))[, "X"],
+      lat = sf::st_coordinates(sf::st_centroid(geom))[, "Y"],
```
infotroph: Does `st_coordinates` take enough time that it'd be worth breaking the pipeline to call it once and then extract X and Y from the result?

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+      area_ha = PEcAn.utils::ud_convert(Acres, "acre", "ha")
+    ) |>
+    dplyr::rowwise() |>
+    dplyr::mutate( # generate ids rowwise separately because
+      # rowwise geospatial operations are very slow
```
infotroph: I'm confused by this wording -- is rowwise the slow way or the fast one?

infotroph: looks like (at least on my machine, with this particular shapefile, right now with the moon in this phase, etc) rowwise is the slow way:

```
> system.time(dwr |> rowwise() |> mutate(id = digest::digest(geometry, algo = "xxhash64")))
   user  system elapsed 
  8.849   0.107   8.951 
> system.time(dwr |> mutate(id = digest::digest(geometry, algo = "xxhash64")))
   user  system elapsed 
  0.113   0.015   0.128 
```

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+    tidyr::as_tibble() |>
+    dplyr::mutate(
+      crop = .data[[crop_col]],
+      pft = case_when(
+        crop %in% c(
+          "Cherries", "Almonds", "Plums, Prunes and Apricots",
+          "Walnuts", "Citrus", "Miscellaneous Deciduous", "Pears", "Olives",
+          "Apples", "Pistachios", "Bush Berries", "Peaches/Nectarines",
+          "Miscellaneous Subtropical Fruits", "Pomegranates"
+        ) ~ "woody perennial crop",
```
infotroph: Another column I don't recognize -- The DWR shapefiles have alphanumeric codes (eg cherry = D3, almond = D12, etc), not cropnames. Am I missing where you merge in the lookup?

(If you have the codes available, I think the woody perennials will likely consist of all of C = "citrus and subtropical", D = "deciduous fruits and nuts", and V = "vineyards", plus T19 = "bush berries", and T28 = "blueberries")

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+        TRUE ~ paste0("no PFT for ", crop)
+      )
+    ) |>
+    dplyr::rename(
+      source = Source,
+      notes = Comments
+    ) |>
+    dplyr::select(id, lat, lon, year, crop, pft, source, notes)
```
infotroph: Nit: This is a place I might do the renaming inside the select like `...select(..., source = Source, notes = Comments)`, but this works

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,124 @@
...
+
+  # Warn about crops without a PFT
+  unassigned_pft <- csv_data |>
+    filter(grepl("no PFT for", pft)) |>
+    distinct(crop, pft)
+  if (nrow(unassigned_pft) > 0) {
+    PEcAn.logger::logger.warn( # or should this be error?
```
infotroph: Warn seems right, especially since it'll come out in the "No PFT for" strings too

modules/data.land/R/shp2gpkg.R
```diff
@@ -0,0 +1,84 @@
...
+#'
+#' @param input_shp Character. Path to the input Shapefile (e.g., `"data/myfile.shp"`).
+#' @param output_gpkg Character. Path to the output GeoPackage (e.g., `"output/myfile.gpkg"`).
+#' @param layer_name Character. Optional. Name of the layer in the GeoPackage. Defaults to the base name of the Shapefile.
+#' @param overwrite Logical. Whether to overwrite an existing layer in the GeoPackage. Defaults to `TRUE`.
+#'
+#' @return Invisibly returns the path to the GeoPackage (`output_gpkg`) upon successful conversion.
+#' @details
+#' This function reads a Shapefile, converts it to an `sf` object, and writes it to a GeoPackage.
+#' The `sf` package handles the conversion and ensures spatial data integrity.
```
infotroph: Nit: I recommend moving details above the parameter listing and removing the @details tag -- that's the layout Roxygen favors and it treats paragraph 1 as the title, paragraph 2 as the description, and paragraphs 3-and-onward as details. Yes, it differs from the layout of an Rd page, but I find it makes it easier to find the parameters when reading in source form.

modules/data.land/R/landiq2std.R
infotroph: Also I see your TODO above, but just to make sure: the mapping definitely needs to be made configurable before merging rather than left hardcoded into the function.

modules/data.land/R/landiq2std.R
dlebauer: ~I think I found the same result as you but borked the comment + implementation~

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,128 @@
...
```
dlebauer:
```suggestion
    PEcAn.logger::logger.warn( 
```

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,128 @@
...
+  landiq_polygons <- sf::st_read(input_file, quiet = FALSE)
+
+  # Determine crop year from column name
+  crop_col <- grep("Crop[0-9]{4}", colnames(landiq_polygons), value = TRUE)
+  year <- gsub("Crop", "", crop_col)
+  # Check required columns
+  required_cols <- c("Acres", crop_col, "Source", "Comments", "County", "geom")
+  missing_cols <- setdiff(required_cols, colnames(landiq_polygons))
+  if (length(missing_cols) > 0) {
+    stop("Input file is missing the following columns: ", paste(missing_cols, collapse = ", "))
```
dlebauer:
```suggestion
    PEcAn.logger::logger.error("Input file is missing the following columns: ", paste(missing_cols, collapse = ", "))
```

modules/data.land/R/landiq2std.R
dlebauer: 1. I'll change it here; can you please update this info in the [Data Needs spreadsheet](https://docs.google.com/spreadsheets/d/1lTLbAN2JhxfLtad9jUTJjFH7OhTLmp-GOVmh86SqkUM/edit?gid=0#gid=0)?
2. In the sense of, we need a standard to program to so this PR (see title) reflects a proposal.

@infotroph with respect to "In CCMMF we decided early on", please see [this Slack thread](https://pecanproject.slack.com/archives/C06M2FKU69W/p1738615806589429?thread_ts=1738612193.764499&cid=C06M2FKU69W).

modules/data.land/R/landiq2std.R
dlebauer: Thanks for pointing that out!
```suggestion
#' https://data.cnra.ca.gov/dataset/statewide-crop-mapping and processes the data into
```

modules/data.land/R/landiq2std.R
dlebauer: I built this around the 2016 file that Charlotte shared. All of these names are from the orignial DWR 2016 except `geom` (which is `geometry` in the source). I thought `geom` was the name required by OGC specification but now see that there is no restrictions on name, and I prefer the original geometry anyway.

It will still be a todo (thus WIP) to align this with her work.

modules/data.land/R/landiq2std.R
infotroph: Dude, I'm saying "let's be consistent here in the place we're currently discussing", not "I volunteer to fix inconsistencies everywhere they might exist". If you don't want to make the change here, say so and skip the whataboutism.

dlebauer: @infotroph, I agree that it makes sense to refer to the dataset as DWR, but I don’t recall a formal decision. I also couldn’t find it reflected in our documents, including the Modeling Framework we submitted to CARB.

If I missed a reference somewhere, let me know—I want to make sure we're aligned, which is why I used the Modeling Framework (including the snapshot of the Data Needs spreadsheet) as a reference.

More broadly how can we ensure we’re maintaining common points of reference so that we can avoid future misunderstandings? (perhaps we can discuss further in slack or our next meeting).

modules/data.land/R/landiq2std.R
dlebauer: good catch - fixed by replacing this overuse of the same file with a temp file that is removed on exit

modules/data.land/R/landiq2std.R
dlebauer: done w/ additional mutate step

modules/data.land/R/landiq2std.R
dlebauer: What I meant was that `digest::digest()` requires `rowwise()` so I did it in a separate mutate step. Fixed.

modules/data.land/data-raw/landiq_pft_map.R
```diff
@@ -0,0 +1,37 @@
+non_crop <- c(
```
dlebauer: replace these with a dataset (where) perhaps it should be user provided?

divine7022:
heads up -- yet to update `landiq2std()` and add tests

divine7022:
documenting to understand scope:

LandIQ shapefiles have two distinct formats:
- 2014: legacy format (`Crop2014`, `Acres`, `Source`, `Comments`)
- 2016: mix format (has both `Crop2016` AND `CLASS1`/`SUBCLASS1`)
- 2018+: new format (`CLASS1`/`SUBCLASS1`, `ACRES`, no `Source`/`Comments`)
currently i have implemented support for both formats. However, given that the harmonized `crops_all_years.csv` already covers 2016-2023, should `landiq2std()` focus only on the new format (2018+)? or do we want backward compatibility(support for pre-2018) with the legacy shapefiles for potential future use cases?

divine7022:
confirmed with @dlebauer -- 2014 is out of scope

mdietze:
I would prioritize the harmonized format (what you call "new"). I don't think there needs to be backwards compatibility

modules/data.land/R/landiq2std.R
```diff
@@ -0,0 +1,221 @@
...
+    dplyr::select(site_id, geom, lat, lon, area_ha, county) |>
+    sf::st_transform(crs_output)
+
+  # Create output CSV
+  csv_data <- crop_data |>
+    dplyr::select(
+      site_id, lat, lon, year, county,
+      class, subclass, crop, pft,
+      multiuse, source, notes
+    )
```
ashiklom, line 188: FYI: A GeoPackage is just a wrapper around an SQLite database. So it's valid (and, in fact, idiomatic, and has certain advantages) to store the `polygonID -> geometry` table and the long `polygonID -> attribute` table (no geoms) as two tables ("layers") in the same geopackage file. You can even configure SQLite to enforce the foreign key constraint to ensure the polygon IDs match across the two tables.

Separate CSV (or parquet, which I would recommend since the resulting table is fairly large) is conceptually simpler and creates a smaller file that we can pass around as needed...but then you have to make sure you keep both in sync.

Food for thought.

dlebauer: These are good points and we discussed this early on. Generally we will have lots of tables that reference these polygons, and it didn't make sense to stuff everything into a single file. 

It could make sense to keep all the landiq data together this way (I don't recall if there was an objection to that).

More broadly, most joins will be with other tables (eg environmental covariates, model inputs and outputs). We ended up keeping things separate for a few reasons, including: Many tables will be very large and independently used and consumed at different steps. Many steps need the metadata but not the polygons.

dlebauer:
closing as obsolete --> implemented in https://github.com/ccmmf/cadwr-landuse

PR #3425, opened 2025-01-27
Replace original testing server for `api testing`

Sweetdevil144:
## Description

Replace `pecan.localhost` with `localhost` as API testing on docker somehow references api through `localhost` when docker containers are ran.


robkooper (approved):
LGTM i think the url is normally if the full stack is deployed in docker, but this should work for running in the API server using the test script

PR #3430, opened 2025-02-04
Add make help

dlebauer:
The goal is to make it easier to use PEcAn's `make` functionality by adding documentation. 

This PR
- Adds help target to enable make help
- Documents all targets and added helpful info

TODO:
- [x] Update documentation
- [ ] Update changelog

## Types of changes

## Checklist:


mdietze (approved):

Makefile
```diff
@@ -99,30 +99,46 @@ doc_R_pkg = \
 
 depends = .doc/$(1) .install/$(1) .check/$(1) .test/$(1)
 
+### Help
+
+help: ## Show this help message
```
infotroph: Please move down to appear below the `all:` target, so that `all` keeps being the default target (by virtue of being the first one in the file) and is what's run when someone types `make` with no target specified.

(I have a vague notion there might also be a way to explicitly mark a target as the default rather than relying on Makefile order, but I don't remember what that way is or how widely supported it is.)

Makefile
```diff
@@ -99,30 +99,46 @@ doc_R_pkg = \
...
+	@echo "Usage: make [target]"
+	@echo ""
+	@echo "Examples:"
+	@echo "  make all     "
+	@echo "  make document"
+	@echo ""
+	@echo "Notes:"
+	@echo "  - Components not included: cable (models), data.mining and DART (modules)."
+	@echo "  - Standard workflow: install packages, run checks, test, and document before submitting a PR."
+	@echo "  - Before submitting a PR, please ensure that all tests pass, code is linted, and documentation is up-to-date."
```
infotroph: I think this is intended to be normative ("Your standard workflow when preparing to submit a PR should include all of installing packages, running checks, ..."), but it could also be read as a descriptive summary of what this Makefile's standard workflow does ("The effect of running the standard workflow is to install packages, run checks, ...") , raising the question of how one invokes said workflow or switches to a nonstandard one when needed.

Consider pointing to the PEcAn book for conventions on developer workflows and saying a bit more for folks in first-time installation mode who haven't yet connected that their setup process is "just" the install part of the standard dev workflow?

Makefile
```diff
@@ -99,30 +99,46 @@ doc_R_pkg = \
...
 
-document: $(ALL_PKGS_D) .doc/base/all
-install: $(ALL_PKGS_I) .install/base/all
-check: $(ALL_PKGS_C) .check/base/all
-test: $(ALL_PKGS_T) .test/base/all
-shiny: $(SHINY_I)
 
-# Render the PEcAn bookdown documentation
-book:
+book: ## Render the PEcAn bookdown documentation
```
infotroph: This is a good specific example of my comment above about end-of-line comments making the code harder to read: When I open a new-to-me Makefile, one of the first things I skim for is short-named targets without dependencies, because they're often phony targets where the interesting work gets done. This change would make this entry harder for me to see while skimming.

Makefile
```diff
@@ -99,30 +99,46 @@ doc_R_pkg = \
...
+	@echo "  make document"
+	@echo ""
+	@echo "Notes:"
+	@echo "  - Components not included: cable (models), data.mining and DART (modules)."
+	@echo "  - Standard workflow: install packages, run checks, test, and document before submitting a PR."
+	@echo "  - Before submitting a PR, please ensure that all tests pass, code is linted, and documentation is up-to-date."
+	@echo ""
+	@echo "Available targets:"
+	@grep -h -E '^[a-zA-Z0-9_-]+:.*?## ' $(MAKEFILE_LIST) | \
+		awk 'BEGIN {FS = ":.*?## "}; {printf "  %-20s %s\n", $$1, $$2}'
```
infotroph: This is clever but looks quite brittle -- it relies on a very specific comment format (one that--to me at least--makes it harder to read the code!) and won't pick up any newly-added targets that don't use it. Given the list of targets that need help is fairly short, would you consider either...? 

1. hard-coding the help text here and relying on authors* to update as they edit the file, or 
2. putting the help strings on their own lines above each target (a la https://stackoverflow.com/a/71734849) instead of trying to parse the target line itself? I'd be in favor of a format like
 ```
## @help foo: frabulate the crampons
foo:
    [...]
```

*Yeah, relying on authors is error-prone, but FWIW makefiles are one place I _do_ reliably need to read the whole file every time I edit, so I contend this is a context where the people doing the editing will have at least a fighting chance of remembering to update the help text.

**Review by infotroph — CHANGES_REQUESTED** [review]

File: Makefile [hunk 5]
```diff
@@ -99,30 +99,46 @@ doc_R_pkg = \
...
+help: ## Show this help message
+	@echo "Usage: make [target]"
+	@echo ""
+	@echo "Examples:"
+	@echo "  make all     "
+	@echo "  make document"
+	@echo ""
+	@echo "Notes:"
+	@echo "  - Components not included: cable (models), data.mining and DART (modules)."
+	@echo "  - Standard workflow: install packages, run checks, test, and document before submitting a PR."
```
dlebauer:
```suggestion
	@echo "  - A common workflow is to: install packages, run checks, test, and document before submitting a PR."
	@echo "    See the Developer Guide in PEcAn Documentation for additional details https://pecanproject.github.io/pecan-documentation/develop/updatebety.html#pecan-make"
```

Makefile
dlebauer: see above https://github.com/PecanProject/pecan/pull/3430/files#r1942059169 and feel free to propose a specific change

Makefile
dlebauer: I considered writing it all out in one block but this also seems to be a common approach (though regex varies). Regarding the critique that it "won't pick up any newly-added targets that don't use it", that is true for hard coded docs - it will only pick up what has been documented.

The link in 2 doesn't point to a particular answer, but it would require reworking the regex. 

Changes and improvements are welcome if you want to implement an alternative solution.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Added
 
+- Documentation of `make` options including addition of `make help`
```
dlebauer, line 12:
```suggestion
- Documentation of `make` options including addition of `make help` 
- Add make option to document a single package with `make documentation pathto/package` 
```

Makefile
```diff
@@ -10,10 +10,10 @@ MODULES := allometry assim.batch assim.sequential benchmark \
...
 				 photosynthesis priors rtm uncertainty
 
-# Components not currently included in the build
-# (Most need more development first)
-# 	models: cable
-#	modules: data.mining, DART
+## Components not currently included in the build
+## (Most need more development first)
+## 	models: cable
+##	modules: data.mining, DART
```
infotroph, line 16: Why the extra hashes here?

Makefile
```diff
@@ -99,36 +99,64 @@ doc_R_pkg = \
...
-check_models: $(MODELS_C) 
+check_base: $(BASE_C)     ## Check base packages  
+check_models: $(MODELS_C) ## Check model packages   
 
 # Install base first as Modules has a circular dependency on base,
 # and then run a check on modules
-check_modules: $(BASE_I) $(MODULES_C) 
+check_modules: $(BASE_I) $(MODULES_C) ## Install base packages, check module packages
+
+.PHONY: document
```
infotroph, line 116: I'd weakly prefer to keep the existing approach of declaring all phony targets on a single line, but if you feel strongly about having a .PHONY next to each phony target then please move all of them, not just some.

Makefile
```diff
@@ -99,36 +99,64 @@ doc_R_pkg = \
...
+document: ## Generate documentation for all packages or a single package: make document path/to-package
+	@if [ "$(filter-out $@,$(MAKECMDGOALS))" ]; then \
+		PKG="$(filter-out $@,$(MAKECMDGOALS))"; \
+		echo "Documenting package: $$PKG"; \
+		$(call doc_R_pkg, $$PKG); \
+	else \
+		echo "Documenting all packages"; \
+		$(MAKE) $(ALL_PKGS_D) .doc/base/all; \
+	fi
```
infotroph, line 125: I agree it would be really nice to have support for `make document path/to/pkg`! But I recommend deferring this to another PR -- what you have here needs more work because it:
* doesn't skip up-to-date packages (notice that `make document base logger && make document base/logger && make document base.logger` runs Roxygen 3 times)
* is noisy (especially the screenful of "...is up to date" when invoked with no arguments)
* should be applied to the other goals at the same time -- seems like it'd be pretty confusing to support `make document base/logger` but then still need to use `make .check/base/logger`.

Makefile
```diff
@@ -99,36 +99,64 @@ doc_R_pkg = \
...
+book: 
+	cd ./book_source && make build ## Render the PEcAn bookdown documentation
 
 # Make the timestamp directories if they don't exist yet
 .doc .install .check .test .shiny_depends $(call depends,base) $(call depends,models) $(call depends,modules):
 	mkdir -p $@
 
+### Help
+
+help: ## Show this help message
```
infotroph, line 142: I think this should be marked .PHONY too

Makefile
```diff
@@ -99,36 +99,64 @@ doc_R_pkg = \
...
-	cd ./book_source && make build
+
+book: 
+	cd ./book_source && make build ## Render the PEcAn bookdown documentation
 
 # Make the timestamp directories if they don't exist yet
 .doc .install .check .test .shiny_depends $(call depends,base) $(call depends,models) $(call depends,modules):
 	mkdir -p $@
 
+### Help
```
infotroph, line 140: Nit: move up two spots to land between `shiny` and `book`,  so that all the phony rules come before all the non-phony ones

Makefile
infotroph: > doesn't point to a particular answer

Huh, does for me, but the core idea I was trying to link to was just that parsing the target definition is a minefield and since any approach needs a human editor to type their doc string into a comment, it's easier to pick a custom comment syntax and use it on its own line above each target.

I volunteer to rewrite the regex and I propose to use Roxygen syntax:
```
#' all: Install all packages and generate documentation
all: install document
```

@dlebauer would you like me to push directly to this branch or open a PR into this one?

Makefile
infotroph: GitHub's making it hard for me to tell if you already applied the above or if it's a pending proposal, but yes it seems like an improvement.

infotroph (changes requested):

dlebauer:
@infotroph I welcome your suggestions implemented as a PR against this branch.

Makefile
```diff
@@ -99,36 +99,70 @@ doc_R_pkg = \
...
+	@echo "  make document"
+	@echo "  make document modules/assim.sequential  # Generate documentation for a specific package"
+	@echo ""
+	@echo "Notes:"
+	@echo "  - Components not included: cable (models), data.mining and DART (modules)."
+	@echo "  - Standard workflow: install packages, run checks, test, and document before submitting a PR."
+	@echo "  - Before submitting a PR, please ensure that all tests pass, code is linted, and documentation is up-to-date."
+	@echo ""
+	@echo "Available targets:"
+	@grep -h "^#' " $(MAKEFILE_LIST) | cut -d' ' -f2- | sort | column -t -s ':' | sed 's/^/  /'
```
infotroph: I got excited about how nice this alignment looked on Mac and then realized `column` isn't universally available; this fails inside our Docker images 😿. 
I don't think alignment is worth installing more tools for; maybe instead indent all the target lines but leave the `:` ragged?
```suggestion
	@grep -h "^#' " $(MAKEFILE_LIST) | sed "s/^#' /  /" | sort
```

PR #3431, opened 2025-02-04
Update SDA Downscaling to allow sf 

dlebauer:
## Description

* Refactored SDA functions to accept direct R objects (lists, data.frames, and sf objects) in addition to file paths.
* Added a helper function (.convert_coords_to_sf()) to standardize coordinate conversion.
* Updated man pages to reflect changes in input parameters.
* Implemented new tests to verify the updated behavior.

 

## Motivation and Context

Existing downscaling code required rasters. In the CCMMF project we are using vector data, so this PR generalizes the downscaling code to support vector file types as well.

It also allows ensemble data to be passed as either an RDS file or an R list of dataframes.

## Review Time Estimate

- [x] Without unnecessary delay

## Types of changes

## Checklist:


modules/assim.sequential/R/downscale_function.R
```diff
@@ -114,62 +153,70 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
+
+  # Convert site coordinates to an sf object using the helper function
+  site_coordinates_sf <- .convert_coords_to_sf(preprocessed$site_coordinates)
+
+  # Extract predictors from covariates <U+2013> support both SpatRaster and sf objects:
+  if (inherits(covariates, "SpatRaster")) {
+    # For raster objects, convert the sf coordinates to SpatVector before extraction
+    predictors <- as.data.frame(terra::extract(covariates, terra::vect(site_coordinates_sf), ID = FALSE))
+  } else if (inherits(covariates, "sf")) {
+    # For sf objects, perform a spatial join using st_intersects and then drop the geometry column
```
dlebauer:
```suggestion
    # For sf objects, spatial join
```

modules/assim.sequential/R/downscale_function.R
```diff
@@ -1,67 +1,104 @@
...
-##' @param data_path Character. File path for .rds containing ensemble data.
-##' @param coords_path Character. File path for .csv file containing the site coordinates, with columns named "lon" and "lat".
-##' @param date Date. If SDA site run, format is yyyy/mm/dd; if NEON, yyyy-mm-dd. Restricted to years within the file supplied to 'data_path'.
-##' @param carbon_pool Character. Carbon pool of interest. Name must match the carbon pool name found within the file supplied to 'data_path'.
+##' @param ensemble_data Character or list. Either a file path to an `.rds` file containing ensemble data, or an in-memory list where each element is a data.frame indexed by date.
+##' @param site_coordinates Character, data.frame, or sf object. Either:
+##'   - A file path to a `.csv` containing site coordinates with columns `"id"`, `"lat"`, and `"lon"`.
+##'   - A `data.frame` with the same structure.
+##'   - An `sf` object with point geometries.
+##' @param date Date. If SDA site run, format is yyyy/mm/dd; if NEON, yyyy-mm-dd. Restricted to years within the file or object supplied to 'ensemble_data'.
```
mdietze: When this function was originally written I asked that this be fixed to just take in an actual Date. While you're fixing things could you fix that too.

modules/assim.sequential/R/downscale_function.R
```diff
@@ -1,67 +1,104 @@
...
     stop(paste("Carbon pool", carbon_pool, "not found in the input data."))
   }
-  
+
   carbon_data <- as.data.frame(t(data[which(names(data) == carbon_pool)]))
   names(carbon_data) <- paste0("ensemble", seq(ncol(carbon_data)))
-  
+
+  # If sf object, convert to data.frame with 'lon' and 'lat'
+  if (inherits(site_coordinates, "sf")) {
```
mdietze: Curious why this is down here rather than up earlier where the other two site_coordinates cases were

modules/assim.sequential/R/downscale_function.R
```diff
@@ -82,28 +119,30 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
+
 ##' @title SDA Downscale Function
 ##' @name SDA_downscale
 ##' @author Joshua Ploshay, Sambhav Dixit
 ##'
 ##' @param preprocessed List. Preprocessed data returned as an output from the SDA_downscale_preprocess function.
-##' @param date Date. If SDA site run, format is yyyy/mm/dd; if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'preprocessed' from the 'data_path'.
-##' @param carbon_pool Character. Carbon pool of interest. Name must match carbon pool name found within file supplied to 'preprocessed' from the 'data_path'.
-##' @param covariates SpatRaster stack. Used as predictors in downscaling. Layers within stack should be named. Recommended that this stack be generated using 'covariates' instructions in assim.sequential/inst folder
+##' @param date Date. If SDA site run, format is yyyy/mm/dd; if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'preprocessed' from the 'ensemble_data'.
```
mdietze: same comment about fixing date

modules/assim.sequential/R/downscale_function.R
```diff
@@ -114,62 +153,70 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
       ensemble_col <- paste0("ensemble", i)
       formula <- stats::as.formula(paste(ensemble_col, "~", paste(covariate_names, collapse = " + ")))
       models[[i]] <- randomForest::randomForest(formula,
-                                                data = train_data,
-                                                ntree = 1000,
-                                                na.action = stats::na.omit,
-                                                keep.forest = TRUE,
-                                                importance = TRUE)
-      
+        data = train_data,
```
mdietze, line 152: revert indentation change

modules/assim.sequential/R/downscale_function.R
```diff
@@ -114,62 +153,70 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
+        ntree = 1000,
+        na.action = stats::na.omit,
+        keep.forest = TRUE,
+        importance = TRUE
+      )
+
```
mdietze: Another suggestion I've made in the past is that there may be logical/computational advantages to separating the function that builds the downscaling models and a separate function that makes the predictions to a map. Right now one can't just fit the model and evaluate the results without also running the (more computationally costly) mapping

modules/assim.sequential/R/downscale_function.R
```diff
@@ -302,18 +349,19 @@ SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_typ
       prediction_rast <- terra::rast(covariates)
 
       # Generate spatial predictions using the trained model
-      maps[[i]] <- terra::predict(prediction_rast, model = models[[i]],
-                                  fun = cnn_ensemble_predict,
-                                  scaling_params = scaling_params)
+      maps[[i]] <- terra::predict(prediction_rast,
+        model = models[[i]],
```
mdietze, line 293: fix indent

modules/assim.sequential/tests/testthat/test_downscaling.R
```diff
@@ -0,0 +1,79 @@
+library(testthat)
+library(sf)
+library(terra)
+library(randomForest)
+library(withr)
```
infotroph: Consider removing these -- testthat gets loaded by the package-level testthat.R, and since we want package functions to behave right when a dependency isn't attached, I think there's virtue in making ourselves run the tests without attaching things.

modules/assim.sequential/R/downscale_function.R
dlebauer: @mdietze I'm not sure if I fully understand, and how I can make sure not to brake existing code. 

What I have in mind is to 
- require the date parameter be a date object?
- if it is not a date object, try to convert from the two yyyymmdd options?

is that what you had in mind?

modules/assim.sequential/R/downscale_function.R
dlebauer: good point, no real reason

modules/assim.sequential/R/downscale_function.R
dlebauer: @mdietze I'm not sure if I fully understand, and how I can make sure not to brake existing code. 

What I have in mind is to 
- require the date parameter be a Date object and update the docs accordingly and
- if it is not a date object, try to convert from the two yyyymmdd options? (already happens)

is that what you had in mind?

modules/assim.sequential/R/downscale_function.R
dlebauer: In this case, it doesn't look like the function actually uses date, so I've removed it

modules/assim.sequential/R/downscale_function.R
```diff
@@ -112,64 +153,85 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
 ##' @return A list containing the training and testing data sets, models, predicted maps for each ensemble member, and predictions for testing data.
 
-SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_type = "rf", seed = NULL) {
+SDA_downscale <- function(preprocessed, date = NULL, carbon_pool, covariates, model_type = "rf", seed = NULL) {
+  if (!missing(date)) {
+    ## If you see this and it is after 2026-04-01, please remove
+    #    1. line starting with ##' @param date
+    #    2. date = NULL from function call
+    #    3. this conditional starting with if (!missing(date)) and ending after the warning below
+    PEcAn.logger::logger.warn("'date' argument is not used and will be removed on or after 2026-04-01. It is currently ignored.", call. = FALSE)
```
infotroph: logger.warn doesn't take `call.` and treats its contents as another message to print
```suggestion
    PEcAn.logger::logger.warn("'date' argument is not used and will be removed on or after 2026-04-01. It is currently ignored.")
```

(Also `call.` is high up my list of the base R arguments that look the most like typos, but that's not our problem to fix 😉 )

infotroph (commented):
A high-level question that you likely considered already: 

Accepting both files and in-memory data seems like it adds a lot of complexity to functions that were nice and clean before that, and I naively expect it would be both "easy" and desirable for error recovery to instead write data out to files before running this. What tradeoff factors make this the better approach?

mdietze:
Per @infotroph last comment, I'm inclined to push in the other direction. For SDA_downscale, I'm fairly confident that I asked the original authors of these functions to stop loading the data within the function, as you have greater flexibility if you load the constraint data first then call the function. This way you can easily control what covariate variables are part of any particular downscaling, and what domain you're doing the downscaling for, by changing what stack of variables you're passing in. In other words, for that function you could probably drop the option to pass in file paths.

infotroph:
@mdietze 's point makes sense and yes, load-first seems sensible. I was trying to question the need to support both rather than pushing for write-first specifically.

modules/assim.sequential/R/downscale_function.R
```diff
@@ -1,67 +1,106 @@
...
   if (nrow(site_coordinates) != nrow(carbon_data)) {
     message("Number of rows in site coordinates does not match the number of rows in carbon data.")
     if (nrow(site_coordinates) > nrow(carbon_data)) {
       message("Truncating site coordinates to match carbon data rows.")
-      site_coordinates <- site_coordinates[1:nrow(carbon_data), ]
+      site_coordinates <- site_coordinates[seq_len(nrow(carbon_data)), ]
     } else {
       message("Truncating carbon data to match site coordinates rows.")
-      carbon_data <- carbon_data[1:nrow(site_coordinates), ]
+      carbon_data <- carbon_data[seq_len(nrow(site_coordinates)), ]
```
dlebauer: @sambhavnoobcoder it looks like you implemented this, could you help me understand the logic? My suspicion is that if `nrow(site_coordinates) != nrow(carbon_data)`, something wrong with the inputs, and truncating one or the other could result in carbon_data being associated with the wrong sites.

sambhavnoobcoder: Hi @dlebauer ,my original implementation assumed that the site coordinates and carbon data were already aligned (i.e., that the order of rows in both datasets corresponds one‐to‐one). The idea was that if one of the inputs had extra rows, then simply truncating the longer dataset would let the function run without error. In that context, the truncation was a quick fix for handling extra or misordered rows.

That said, the concern is valid: if the rows aren’t truly aligned, then truncating without verifying a matching site identifier could indeed lead to carbon data being paired with the wrong sites. In other words, the current approach works correctly only if the inputs are pre‐validated or guaranteed to be in the correct order.

dlebauer:
> @mdietze 's point makes sense and yes, load-first seems sensible. I was trying to question the need to support both rather than pushing for write-first specifically.

I was maintaining the 'read from file' functionality to remain backwards-compatible. If these functions don't need to be backward compatible, they will be much easier to refactor.

mdietze:
Right now, this code has a small enough user community that backwards compatibility isn't an issue. That will change once we start publishing using this workflow and need outward-facing reproducibility

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 ### Fixed
 - updated github action to build docker images
 - PEcAn.SIPNET now accepts relative paths in its input XML (#3418). Previously all files referenced in the autogenerated `job.sh` needed to be specified as absolute paths.
+- Refactored `SDA_downscale` and `SDA_downscale_preprocess` to allow direct use of R objects in addition to file paths. Added tests for these functions.
```
dlebauer:
```suggestion
- Refactored and created new version of `SDA_downscale` named `ensemble_downscale`.
```

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 ### Fixed
 - updated github action to build docker images
 - PEcAn.SIPNET now accepts relative paths in its input XML (#3418). Previously all files referenced in the autogenerated `job.sh` needed to be specified as absolute paths.
+- Refactored and created new version of `SDA_downscale` named `ensemble_downscale`.
+- Added helper function `.convert_coords_to_sf()` for consistent conversion of data with lat lon data to `sf` pts.
```
infotroph: Nits: 
* refactors seem like belong in "changed" not "fixed"
* internal helper functions don't need to be mentioned in the changelog (and I'd argue it's confusing to do so, since the point is for users to not know or care about them)

modules/assim.sequential/R/Helper.functions.R
```diff
@@ -6,10 +6,10 @@
 #' @description This function performs a simple outlier replacement on all the columns of dataframes inside a list
 #' @return A list the same dimension as X, with each column of each dataframe
 #'   modified by replacing outlier points with the column median
-#' @export
+#' @export outlier.detector.boxplot
```
infotroph: I bet this is trying to prevent this name being treated as an S3 method, but I recommend renaming the function instead. Life's too short to fight with R about this.

modules/assim.sequential/R/downscale_function.R
```diff
@@ -1,67 +1,113 @@
...
   names(input_data) <- input_date_names
-  
-  # Convert the input date to a Date object
-  standard_date <- lubridate::ymd(date)
-  
+
+  # Ensure 'date' is a Date object, if not, convert
+  if (!inherits(date, "Date")) {
+    standard_date <- lubridate::ymd(date)
+  }
```
infotroph: `ymd` accepts Date objects and returns them unchanged -- can skip the if and the rename

modules/assim.sequential/R/downscale_function.R
```diff
@@ -1,67 +1,113 @@
...
+    )
+    PEcAn.logger::logger.error("I am not sure how to reconcile these differences.")
+    # see https://github.com/PecanProject/pecan/pull/3431/files#r1953601359
+  #    if (nrow(site_coordinates) > nrow(carbon_data)) {
+  #      PEcAn.logger::logger.info("Truncating site coordinates to match carbon data rows.")
+  #      site_coordinates <- site_coordinates[seq_len(nrow(carbon_data)), ]
+  #    } else {
+  #      PEcAn.logger::logger.info("Truncating carbon data to match site coordinates rows.")
+  #      carbon_data <- carbon_data[seq_len(nrow(site_coordinates)), ]
+  #    }
```
infotroph: The link is taking me to a whole-PR view so I can't tell what specific comment/commit it's referring to, but the commented lines seem safe to delete -- I agree that unequal row counts should make us stop with an error rather than truncate and continue.

modules/assim.sequential/R/downscale_function.R
```diff
@@ -82,28 +128,30 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
+
 ##' @title SDA Downscale Function
 ##' @name SDA_downscale
 ##' @author Joshua Ploshay, Sambhav Dixit
 ##'
 ##' @param preprocessed List. Preprocessed data returned as an output from the SDA_downscale_preprocess function.
-##' @param date Date. If SDA site run, format is yyyy/mm/dd; if NEON, yyyy-mm-dd. Restricted to years within file supplied to 'preprocessed' from the 'data_path'.
-##' @param carbon_pool Character. Carbon pool of interest. Name must match carbon pool name found within file supplied to 'preprocessed' from the 'data_path'.
-##' @param covariates SpatRaster stack. Used as predictors in downscaling. Layers within stack should be named. Recommended that this stack be generated using 'covariates' instructions in assim.sequential/inst folder
+##' @param date *Deprecated*. This argument has never been used and will be removed after 2026-04-01
```
infotroph: I vote drop it now, mention in NEWS, and let users figure it out

modules/assim.sequential/R/downscale_function.R
```diff
@@ -112,64 +160,82 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
-  site_coordinates <- terra::vect(preprocessed$site_coordinates, geom = c("lon", "lat"), crs = "EPSG:4326")
-  
-  # Extract predictors from covariates raster using site coordinates
-  predictors <- as.data.frame(terra::extract(covariates, site_coordinates, ID = FALSE))
-  
+
+  # Convert site coordinates to an sf object using the helper function
+  site_coordinates_sf <- .convert_coords_to_sf(preprocessed$site_coordinates)
+
+  # Extract predictors from covariates <U+2013> support both SpatRaster and sf objects:
```
infotroph: `<U+2013>`?

modules/assim.sequential/R/downscale_function.R
```diff
@@ -112,64 +160,82 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
+  if (inherits(covariates, "SpatRaster")) {
+    # For raster objects, convert the sf coordinates to SpatVector before extraction
+    predictors <- as.data.frame(terra::extract(covariates, terra::vect(site_coordinates_sf), ID = FALSE))
```
infotroph: Comment and code mismatch -- Doesn't the `as.data.frame` mean this converts to plain df rather than SpatVector?

modules/assim.sequential/R/downscale_function.R
```diff
@@ -302,18 +368,19 @@ SDA_downscale <- function(preprocessed, date, carbon_pool, covariates, model_typ
...
+        scaling_params = scaling_params
+      )
 
       # Make predictions on held-out test data
       predictions[[i]] <- cnn_ensemble_predict(models[[i]], x_data[-sample, ], scaling_params)
-      
     }
   } else {
-    stop("Invalid model_type. Please choose either 'rf' for Random Forest or 'cnn' for Convolutional Neural Network.")
+    PEcAn.logger::logger.error("Invalid model_type. Please choose either 'rf' for Random Forest or 'cnn' for Convolutional Neural Network.")
```
infotroph: I know you're just changing stop to logger.error here, but boy is this else a long way from the other cases. A nice refactor would be to do the argument validation with `match.arg(model_type)` way up at the top of the function rather than needing to come back to it at the very bottom.

infotroph: I might even go one step further and pull the if cases out to helper functions so that the entire control flow could be 
```
SDA_downscale <- function ([...], model_type = c("rf", "cnn"), [...]) {
  model_type <- match_arg(model_type)
  [...]
  if (model_type == "rf") {
    predictions <- downscale_rf([...])
  } else {
    predictions <- downscale_cnn([...])
  }
[...]
```

modules/assim.sequential/R/ensemble_downscale.R
```diff
@@ -0,0 +1,266 @@
...
+  if (nrow(ensemble_data) == 0) {
+    PEcAn.logger::logger.error("No carbon data found for the specified carbon pool.")
+  }
+
+  PEcAn.logger::logger.info("Ensemble data subset completed successfully.")
+  return(ensemble_data)
+}
+
+## Helper function to convert table with lat, lon into an sf object
+.convert_coords_to_sf <- function(coords) {
```
infotroph: Same function name used here and in downscale_function.R. Only whichever definition is parsed last at package byte-compilation time will be used (which I think is determined by collation order). Given Helper.functions.R already exists, I recommend moving all these dotted-name functions there.

modules/assim.sequential/R/ensemble_downscale.R
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  # Ensure the sites are in the ensemble data
+  if (!all(unique(site_coords$id) %in% unique(ensemble_data$site_id))) {
+    PEcAn.logger::logger.error("Some sites in site_coords are not present in the ensemble_data.")
+  }
+
+  # Filter the ensemble data to the specified date and carbon pool
+  ensemble_data <- ensemble_data |>
+    filter(
```
infotroph:
```suggestion
    dplyr::filter(
```

modules/assim.sequential/R/ensemble_downscale.R
```diff
@@ -0,0 +1,266 @@
...
+      lubridate::date(datetime) == lubridate::date(date),
+      site_id %in% unique(site_coords$id),
+      variable == carbon_pool
+    ) |>
+    select(site_id, ensemble, prediction)  # use site_id instead of site
```
infotroph:
```suggestion
    dplyr::select("site_id", "ensemble", "prediction")  # use site_id instead of site
```
not sure what the comment means -- is it needed?

modules/assim.sequential/R/ensemble_downscale.R
infotroph: you'll probably need `.data$` on the column names here too

infotroph: Same for other tidyverse fn calls below

modules/assim.sequential/R/ensemble_downscale.R
```diff
@@ -0,0 +1,266 @@
...
+  ## TODO: Use groups from the 01_cluster_and_select_design_points.R
+  sample <- sample(seq_len(nrow(design_pt_data)),
+    size = round(0.75 * nrow(design_pt_data))
+  )
+  train_data <- design_pt_data[sample, ]
+  test_data <- design_pt_data[-sample, ]
+
+  ensembles <- unique(ensemble_data$ensemble)
+
+  results <- furrr::future_map(seq_along(ensembles), function(i) {
```
infotroph: I strongly encourage pulling this anonymous function out to its own definition and giving it an informative name

infotroph: Also it appears `i` is only used to subset `ensemble` -- is there a reason not to pass each one directlly (i.e `map(ensembles, function(ens) ...`?

modules/assim.sequential/R/ensemble_downscale.R
```diff
@@ -0,0 +1,266 @@
...
+##'
+##' @description This function takes the output from the downscale function and computes various performance metrics for each ensemble. 
+##' It provides a way to evaluate the accuracy of the downscaling results without modifying the main downscaling function.
+##'
+##' @return A list of metrics for each ensemble, where each element contains MAE , MSE ,R_squared ,actual values from testing data and predicted values for the testing data
+##'
+##' @export
+downscale_metrics <- function(downscale_output) {
+  
+  test_data_list <- lapply(downscale_output$test_data, function(x) dplyr::pull(x, prediction))
```
infotroph: curious why the extra lambda -- does this differ from `lapply(downscale_output$test_data, dplyr::pull, prediction)`, or indeed from ```lapply(downscale_output$test_data, `[[`, "prediction")```?

modules/assim.sequential/tests/testthat/test_downscaling.R
```diff
@@ -0,0 +1,82 @@
+# Test setup: temporary datasets and files
+withr::with_tempdir({
+  temp_ensemble_data_rds <- "ensemble_data.rds"
+  temp_coords_csv <- "final_design_points.csv"
+  file.remove(temp_ensemble_data_rds, temp_coords_csv)
```
infotroph: Why do you need to start by removing these if the whole test happens in a fresh tempdir?

modules/assim.sequential/tests/testthat/test_downscaling.R
```diff
@@ -0,0 +1,82 @@
...
+
+  set.seed(123)
```
infotroph: Beware that the effects of this `set.seed` persist outside this test block! I think the easiest way to contain it would be to wrap all the lines that use random nunbers in one big `withr::with_seed(123, {...code that uses randomness here...})`.

(I think supposedly you could also replace this line with `withr::local_seed(123)` to have the seed reset at the end of the current execution context, but in my local tests that still affected code outside the function. Dunno if I'm misunderstanding "context" or I did something else wrong)

CHANGELOG.md
infotroph: Also: Need to move up to "unlreased" (this section is now for already-tagged 1.9.0)

modules/assim.sequential/R/downscale_function.R
dlebauer: good idea, though this will be nontrivial refactor; will add as TODO 

```suggestion
# TODO: Separate model fitting and held-out evaluation from spatial map prediction,
# so models can be evaluated without the cost of generating raster maps.
```

modules/assim.sequential/R/downscale_function.R
```diff
@@ -112,64 +160,82 @@ SDA_downscale_preprocess <- function(data_path, coords_path, date, carbon_pool)
...
   y_train <- y_data[sample, ]
   y_test <- y_data[-sample, ]
-  
+
   # Initialize lists for outputs
   models <- list()
   maps <- list()
   predictions <- list()
-  
+
```
dlebauer:
```suggestion
# TODO: Consider extracting the RF and CNN branches into dedicated helper functions.
```

modules/assim.sequential/R/downscale_function.R
dlebauer: Implemented match.arg() validation at top of SDA_downscale. Will leave extracting the RF and CNN branches into helpers for future refactor. 

Added TODO w/ suggestion about refactoring branches into helper functions

modules/assim.sequential/R/ensemble_downscale.R
dlebauer: removed here

modules/assim.sequential/R/ensemble_downscale.R
dlebauer: implemented [[ approach

modules/assim.sequential/R/Helper.functions.R
dlebauer: renamed replacing `.` w/ `_`

dlebauer (commented):
implemented all requested changes, and most suggestions, other than the nontrivial refactor required  to separate model fitting from spatial map prediction in https://github.com/PecanProject/pecan/pull/3431#discussion_r1940365525

mdietze (approved):

PR #3434, opened 2025-02-11
Stop making Docker build required until someone debugs it

mdietze:


robkooper:
hopefully docker builds are fixed, the requirements for the builds is listed in the settings tab of pecan project https://github.com/PecanProject/pecan/settings/branches

PR #3435, opened 2025-02-11
Pre-Built SIPNET Binary for GitHub Actions

### LINKED ISSUE #3427: Pre-Built SIPNET Binary for GitHub Actions
#### **Problem**

Currently, `sipnet.yml` in `pecanproject/pecan` checks out `pecanproject/sipnet`, builds SIPNET from source using `make`, and then runs PEcAn tests. This adds an unnecessary failure point and compute/time for PEcAn GHAS.

#### **Proposed Solution**

- Modify `ci.yml` in `pecanproject/sipnet` to build and upload the SIPNET binary as an artifact and update `sipnet.yml` in `pecanproject/pecan` to download and use the pre-built binary instead of compiling from source.

#### **Benefits**

- Eliminates failed runs due to sipnet build issues 
- Eliminates redundant SIPNET builds in PEcAn CI  
- Reduces build time and improves efficiency  

#### **Next Steps**
- [ ] Update `ci.yml` in `pecanproject/sipnet` to build and upload SIPNET binary.  
- [ ] Update `sipnet.yml` in `pecanproject/pecan` to download and use the pre-built SIPNET binary.  
- [ ] Test and verify the workflows.

AritraDey-Dev:
Fixes #3427

Add CI workflow to build and upload SIPNET binary as an artifact and update PEcAn workflow to use pre-built SIPNET binary.

* **ci.yml**
  - Add steps to build SIPNET binary.
  - Add steps to upload SIPNET binary as an artifact.

* **sipnet.yml**
  - Remove steps to check out and build SIPNET from source.
  - Add steps to download pre-built SIPNET binary.
  - Add steps to use pre-built SIPNET binary.

---


infotroph (changes requested):
Thanks for working on this! To meet the requirements of #3427 you'll need to make some changes:

* Building the binary should be done in the [sipnet repo](https://github.com/pecanproject/sipnet). Determining how to get the built binary from there to here is the crux of the task.
* In all repositories Github Actions workflows have to be in `.github/workflows`, not in the project root.
* The changes to PEcAn's CI should modify existing workflows rather than add new ones.

AritraDey-Dev:
> Thanks for working on this! To meet the requirements of #3427 you'll need to make some changes:
> 
> * Building the binary should be done in the [sipnet repo](https://github.com/pecanproject/sipnet). Determining how to get the built binary from there to here is the crux of the task.
> * In all repositories Github Actions workflows have to be in `.github/workflows`, not in the project root.
> * The changes to PEcAn's CI should modify existing workflows rather than add new ones.

Sure I will look into this changes.

AritraDey-Dev:
i have made the required changes.can you please look into those.

dlebauer:
@AritraDey-Dev Looks good to me! I am not sure if Lines 61-73 of ci.yml need to be there, now that the binary building will be done elsewhere. 

I put a note in https://github.com/PecanProject/sipnet/issues/27 and will update the curl statements once we have a release with binaries.

.github/workflows/ci.yml
```diff
@@ -58,7 +58,16 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Build SIPNET
+        run: make sipnet
+
+      - name: Upload SIPNET binary
+        uses: actions/upload-artifact@v2
+        with:
+          name: sipnet-binary
+          path: ./bin/sipnet
```
infotroph: No changes are needed in this file -- everything related to getting Sipnet installed is encapsulated inside sipnet.yml and nothing should change in how we call it.

AritraDey-Dev:
@infotroph i have Updated upload-artifact to v3. SIPNET installation remains encapsulated in sipnet.yml, so no further modifications were needed. Let me know if anything else is required!

.github/workflows/ci.yml
AritraDey-Dev: i have done the modifications.

.github/workflows/ci.yml
```diff
@@ -61,4 +61,4 @@ jobs:
     uses: ./.github/workflows/sipnet.yml
     with:
       R-version: ${{ matrix.R }}
-    secrets: inherit
+    secrets: inherit
```
allgandalf: I don't understand this code diff, what has changed here? did you add a extra space or something ? if so why?

.github/workflows/sipnet.yml
```diff
@@ -49,16 +49,15 @@ jobs:
...
+    - name: Download SIPNET for macOS
+      run: |
+        curl -L -o sipnet-macos "https://drive.google.com/uc?export=download&id=1psNvSw77VcD5QdbSWeGNVQsbCmJokG2W"
+        chmod +x sipnet-macos
+
+    - name: Download SIPNET for Linux
       run: |
-        cd ${GITHUB_WORKSPACE}/sipnet
-        make sipnet
+        curl -L -o sipnet-linux "https://drive.google.com/uc?export=download&id=1sRTFfs9Z9osmMrtLlwVTqxeUDxgVlrEv"
```
allgandalf: @dlebauer @infotroph are we sure that this drive link will always be live for our project ? joining mid way so do not have the best context

.github/workflows/sipnet.yml
```diff
@@ -49,16 +49,15 @@ jobs:
...
+        chmod +x sipnet-linux
```
allgandalf, line 55: Do we make use of `sipnet-linux` while building ?

allgandalf (commented):
few comments

.github/workflows/ci.yml
AritraDey-Dev: that is mistake from my side while changing the sipnet file and testing it.Thanks for pointing out this.

AritraDey-Dev: i am also not sure,may be while coding there is some unnecessary spaces that came.I will revert it.

.github/workflows/sipnet.yml
AritraDey-Dev: #3427 here the binary shared for linux i have updated that.

.github/workflows/sipnet.yml
dlebauer: That was just a placeholder until I got builds posted with different releases, the latest is 1.3.0 https://github.com/PecanProject/sipnet/releases/tag/v1.3.0

This will always resolve to the latest release https://github.com/PecanProject/sipnet/releases/latest but the version number in the binary name will change.

dlebauer: This should be the Linux build https://github.com/PecanProject/sipnet/releases/download/v1.3.0/sipnet-linux-v1.3.0

AritraDey-Dev:
@dlebauer @allgandalf  @infotroph I was thinking that how to get the binary file...I have solution please check how much this is possible.After building the binary file based on version or some other parameter,every time it will upload to some location(for now g-drive,or some other location) and then some kind of github Actions which will update the binary file in the sipnet.yml file.What do you think can something like this be done?

AritraDey-Dev:
@dlebauer @allgandalf Does this change work?

.github/workflows/sipnet.yml
```diff
@@ -49,16 +49,15 @@ jobs:
...
-      uses: actions/checkout@v4
-      with:
-        repository: PecanProject/sipnet
-        path: sipnet
-        set-safe-directory: false
-    - name: install sipnet
+    - name: Download SIPNET for macOS
+      run: |
+        curl -L -o sipnet-macos "https://drive.google.com/uc?export=download&id=1psNvSw77VcD5QdbSWeGNVQsbCmJokG2W"
+        chmod +x sipnet-macos
```
infotroph: Don't need to download Mac binary to the Linux check machine

.github/workflows/sipnet.yml
infotroph: You can see what binary name we use for testing by following the files referenced from line 69 below and finding that `tests/ghaction.sipnet_PostgreSQL.xml` expects to find it at`$GITHUB_WORKSPACE}/sipnet`, so you'll need to update either this line or that file to match each other

infotroph:
@AritraDey-Dev This PR is currently [failing CI checks](https://github.com/PecanProject/pecan/actions/runs/13576219822/job/3795294562) for the same reasons discussed in #3427: The downloaded binary needs to be at the path expected by the test files, and the test needs to be updated to handle retrieving its climate driver file (currently taken from the sipnet repo as `${GITHUB_WORKSPACE}/sipnet/Sites/Niwot/niwot.clim`, which won't work if we're not checking out the repo anymore).

AritraDey-Dev:
@infotroph i have a fix for this not sure it will work..will test it and let you know if it works.

AritraDey-Dev:
@infotroph i have made the changes in test.I have updated the path for the climate driver file in the test configuration to ensure it points to the correct location during the CI process.The CI checks for sipnet is working fine now.

AritraDey-Dev:
i have made the required changes.can you please look into those.

.github/workflows/sipnet.yml
```diff
@@ -49,16 +49,15 @@ jobs:
...
+    - name: Download SIPNET for Linux
+      run: |
+        curl -L -o sipnet-linux "https://github.com/PecanProject/sipnet/releases/download/v1.3.0/sipnet-linux-v1.3.0"
+        chmod +x sipnet-linux
+
+    - name: Download climate file from GitHub
       run: |
-        cd ${GITHUB_WORKSPACE}/sipnet
-        make sipnet
+        curl -L -o niwot.clim "https://raw.githubusercontent.com/PecanProject/sipnet/refs/heads/master/Sites/Niwot/niwot.clim"
```
infotroph: Apologies that we moved the ground under you here -- @Alomir and @dlebauer just did a big (and overdue, thank you!) reorganization of the Sipnet repository, and this file has moved. I think for this PR it's enough to change the path:

```suggestion
        curl -L -o niwot.clim "https://raw.githubusercontent.com/PecanProject/sipnet/refs/heads/master/tests/smoke/niwot.clim"
```


Long term there's no guarantee the Sipnet repo will keep providing this file, so we may want to consider providing it somewhere under PEcAn's control, but this works for now.

.github/workflows/sipnet.yml
infotroph: 👍 for now; once https://github.com/PecanProject/sipnet/pull/36 is merged we can come back to update the path to one that will stay up to date as new versions of Sipnet are released.

infotroph:
@AritraDey-Dev This PR is now very timely, because the same Sipnet reorganization I mentioned above has also broken our existing checks! If my suggestions above look right to you, let's merge this as our path forward.

AritraDey-Dev:
@infotroph Thanks for the review! Merging this now as the best path forward.Just to confirm, are there any other potential issues we should address before merging?

infotroph:
@AritraDey-Dev No other issues I know of! If we've done everything right, when the currently running CI checks finish we should see that the Sipnet job is passing, at which point I'll press merge (or if not passing, I'll have a look and comment on what's causing the new failure)

infotroph (approved):

AritraDey-Dev:
Sounds good! Let's wait for the CI checks to complete. Hopefully, everything passes like last time it did, but if not, i can debug as needed.

AritraDey-Dev:
@infotroph The Sipnet checks have passed, so I believe we can proceed with merging.

.github/workflows/sipnet.yml
```diff
@@ -49,16 +49,15 @@ jobs:
...
-    - name: Check out SIPNET
-      uses: actions/checkout@v4
-      with:
-        repository: PecanProject/sipnet
-        path: sipnet
-        set-safe-directory: false
-    - name: install sipnet
+    - name: Download SIPNET for Linux
+      run: |
+        curl -L -o sipnet-linux "https://github.com/PecanProject/sipnet/releases/download/v1.3.0/sipnet-linux-v1.3.0"
```
AritraDey-Dev, line 54: Hi @infotroph,
I noticed something important in this PR that I thought was worth mentioning.

Currently, the workflow uses a hardcoded SIPNET release version. While the latest release may remain stable for a while, it’s generally a good practice to fetch the release dynamically to ensure the workflow stays up to date automatically with any new versions.

This can be achieved by querying the GitHub API for the latest release tag:
```
     id: get_release
      run: |
        release_tag=$(curl -s https://api.github.com/repos/PecanProject/sipnet/releases/latest | jq -r .tag_name)
        echo "release_tag=$release_tag" >> $GITHUB_OUTPUT
```
Then, you can use the output from this step to construct the download URL dynamically, like so:

`https://github.com/PecanProject/sipnet/releases/download/${{ steps.get_release.outputs.release_tag }}/sipnet-linux-${{ steps.get_release.outputs.release_tag }}`
Let me know your thoughts on this approach and whether we should proceed with integrating it into the project.

PR #3437, opened 2025-02-12
fix:updated Rcheck_reference.log in settings package

harshagr70:
Removed warning for no visible bindings in package base/settings
reassured and rechecked by rcmdcheck if the warnings still persist , this package seems to have no warnings for visible bindings of global variable , hence updated the reference.log to remove the warnings from the package that are fixed .
fixes a part of issue #2758


infotroph (approved):
Thanks! Good cleanup.

PR #3447, opened 2025-02-19
Adding the feature of ERA5 download function using the CDS API

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -0,0 +1,142 @@
...
+#' @param variables List: a list contains variables to be downloaded (e.g., list("2m_temperature","surface_pressure")).
+#' @param outdir Character: physical path where the ERA5 data are stored.
+#' @param auto.create.key Boolean: decide if we want to generate the CDS RC file if it doesn't exist, the default is TRUE.
+#'
+#' @return A vector containing file paths to the downloaded files.
+#' @export
+#' 
+#' @examples
+#' @author Dongchen Zhang
+ERA5_cds_annual_download <- function(years, months, days, times, area, variables, outdir, auto.create.key = T) {
```
mdietze: recommend adjusting the function arguments to conform to existing patterns for met download functions. This would mean changing the name to `download.ERA5_cds_annual` and changing the order of the arguments to start with `outfolder, start_date, end_date`, then the spatial information, then any function-specific arguments, with required arguments coming before optional ones that have defaults.

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -0,0 +1,142 @@
+#' @description
+#' This function helps to download the yearly ERA5 data based on the prescribed features using the CDS API.
+#' @title ERA5_cds_annual_download
+#' 
+#' @param years Numeric: a series of years to be downloaded (e.g., 2012:2021).
+#' @param months List: a list contains months to be downloaded (e.g., list("01", "02") to download files in Jan and Feb).
+#' @param days List: a list contains days to be downloaded (e.g., list("01", "02") to download files in the first and second days).
+#' @param times List: a list contains times to be downloaded (e.g., list('00:00','03:00') to download files for the times 12:00 and 3:00 am UTC).
+#' @param area Character: a string contains the bounding box (formatted as "North/West/South/East") to be downloaded (e.g., "85/-179/7/-20").
```
mdietze: This is a really odd format. Even if this is what the internal API expects, I think it would be a better design to take the bounding box as a vector of numbers and then format this string inside your function.

modules/data.atmosphere/R/ERA5_download.R
mdietze: Similar to my comment on the bounding box, I'd recommend that you internally construct the API's odd format for months, days, time, etc from the start and end times

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -0,0 +1,142 @@
...
+#' @param outdir Character: physical path where the ERA5 data are stored.
+#' @param auto.create.key Boolean: decide if we want to generate the CDS RC file if it doesn't exist, the default is TRUE.
+#'
+#' @return A vector containing file paths to the downloaded files.
+#' @export
+#' 
+#' @examples
+#' @author Dongchen Zhang
+ERA5_cds_annual_download <- function(years, months, days, times, area, variables, outdir, auto.create.key = T) {
+  options(timeout=360000)
```
mdietze: I'd make this an optional argument to the function with a default, rather than hard coding.

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -0,0 +1,142 @@
...
+  getnetrc <- function (dl_dir) {
+    netrc <- file.path(dl_dir, ".cdsapirc")
+    if (file.exists(netrc) == FALSE ||
+        any(grepl("https://cds.climate.copernicus.eu/api/v2",
+                  readLines(netrc))) == FALSE) {
+      netrc_conn <- file(netrc)
+      writeLines(c(
+        sprintf(
+          "url: %s",
+          getPass::getPass(msg = "Enter URL from the following link \n (https://cds.climate.copernicus.eu/api-how-to#install-the-cds-api-key):")
```
mdietze, line 61: since getPass is in the suggests, you might add a step in the function that returns an informative error message if it isn't installed

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -0,0 +1,142 @@
...
+        'month' = months,
+        'year' = as.character(y),
+        "area" = area,
+        'variable' = variables
+      ),
+      fname
+    )
+    # convert grib to nc file.
+    nc.path <- gsub(".grib", ".nc", fname, fixed = T)
+    cmd <- paste("grib_to_netcdf", fname, "-o", nc.path)
```
mdietze, line 124: what ensures that this command line utility exists in the user's system? it's not in the DESCRIPTION file, nor do you seem to provide any check that the command exists or instructions on how to install it. If you add a check, I'd do it early in the function -- would be a pain to have this function fail AFTER you've done the file download.

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -0,0 +1,142 @@
...
+        PEcAn.logger::logger.info("Unknown variable format. Please check the variable mannually!")
+        return(NA)
+      }
+    }
+    # store the path.
+    nc.paths <- c(nc.paths, nc.path)
+    # remove previous grib file.
+    unlink(fname)
+  }
+  return(nc.paths)
```
mdietze: In keeping with PEcAn met patterns, I recommend returning a more complete `results` object. Look at other met download functions to see how these are structured. You may be using this function outside of the PEcAn met workflow, but that doesn't me we want to write it so that it **can't** be used within the PEcAn met workflow.

mdietze (changes requested):
Update bookdown documentation as needed

modules/data.atmosphere/R/ERA5_download.R
DongchenZ: Fixed.

modules/data.atmosphere/R/ERA5_download.R
DongchenZ: Fixed.

modules/data.atmosphere/R/ERA5_download.R
DongchenZ: Fixed.

modules/data.atmosphere/R/ERA5_download.R
DongchenZ: Fixed.

modules/data.atmosphere/R/ERA5_download.R
DongchenZ: Fixed.

modules/data.atmosphere/R/ERA5_download.R
DongchenZ: Fixed.

mdietze (approved):

mdietze (approved):

PR #3449, opened 2025-02-21
Update Variables for `leaf_phenology` data extraction

### LINKED ISSUE #3411: investigate commented-out phenology in do_conversions
Code in workflow::do_conversions:
```
if (input.tag == "leaf_phenology" && is.null(input$path)) {
      # settings$run$inputs[[i]]$path <- PEcAn.data.remote::extract_phenology_MODIS(
      #   site_info,
      #   start_date,
      #   end_date,
      #   outdir,
      #   run_parallel = TRUE,
      #   ncores = NULL
      # )
      needsave <- TRUE
    }
```

@mdietze in https://github.com/PecanProject/pecan/pull/3407#discussion_r1905594188 :
> Not sure why this bit is commented out -- seems like something we need to be operational

Sweetdevil144:
## Description
Fixes #3411 

Aim is trying to integrate this feature while verifying all current builds pass.

## Review Time Estimate
- [x] When possible

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)  #3411


base/workflow/R/do_conversions.R
```diff
@@ -76,7 +76,22 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
 
     # Phenology data extraction
     if(input.tag == "leaf_phenology" && is.null(input$path)){
-      #settings$run$inputs[[i]]$path <- PEcAn.data.remote::extract_phenology_MODIS(site_info,start_date,end_date,outdir,run_parallel = TRUE,ncores = NULL)
+      # Use settings$run$site if available, otherwise query site
+      site <- if(!is.null(settings$run$site)) {
+        settings$run$site
+      } else {
+        query.site(site.id = settings$run$site$site.id, con = NULL)
```
mdietze: If settings$run$site is NULL, then settings$run$site$site.id is also NULL

base/workflow/R/do_conversions.R
```diff
@@ -76,7 +76,22 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
...
+      }
+      
+      settings$run$inputs[[i]]$path <-
```
mdietze, line 86: has this code been tested to verify that it runs correctly when you pass settings$run$site in as the site_info or are you just 🤞 ?

Sweetdevil144:
`PEcAn.data.remote` is added as an Import in `NAMESPACE`. Although we would only be using this package import for `extract_phenology_MODIS` only.

Sweetdevil144:
How to update `pecan_package_dependencies.csv`?

infotroph:
> How to update pecan_package_dependencies.csv?

`./scripts/generate_dependencies.R`

base/workflow/R/do_conversions.R
Sweetdevil144: Just 🤞 considering site id is `compulsory` within settings.

mdietze: So my concern isn't that site id will be missing, as much as wanting confirmation that you're not submitting code that you've never run to see whether it works or not. I'm also specifically concerned about whether settings$run$site is organized/formatted in the way that extract_phenology_MODIS is expecting.

Sweetdevil144: I didn't face any error running this. Additionally, all pre existing test cases pass. As far as it goes for structure of settings$run$site, we have info of `site$id`, `site$lat` and `site$lon` which are the only info being required for `leaf_phenology`. Can you confirm this?

mdietze (approved):

PR #3450, opened 2025-02-21
Add the downscale functions that works for North America SDA outputs.

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/assim.sequential/NAMESPACE
```diff
@@ -1,6 +1,8 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(stack,covariates.2.geotiff)
```
mdietze: I suspect this one isn't intended to be a S3method. What about renaming the function with underscores instead of periods?

modules/assim.sequential/R/SDA_NA_downscale.R
```diff
@@ -0,0 +1,459 @@
...
+    } else {
+      start <- as.Date(paste0(lubridate::year(dates[i]), "-01-01"))
+    }
+    if (i == length(dates)) {
+      end <- end.date
+    } else {
+      end <- as.Date(paste0(lubridate::year(dates[i]), "-12-31"))
+    }
+    # loop over years.
+    for (j in seq_along(dates)) {
```
mdietze: not following why there's a nested loop over years within the loop oveer years, especially since I don't see where `j` is used

modules/assim.sequential/R/SDA_NA_downscale.R
```diff
@@ -0,0 +1,459 @@
...
+  # dewpoint.
+  dewpoint <- apply(dewpoint.all, c(1, 2), mean)
+  dewpoint <- PEcAn.utils::ud_convert(dewpoint, "K", "degC")
+  # convert dew point to relative humidity.
+  beta <- (112 - (0.1 * temp) + dewpoint) / (112 + (0.9 * temp))
+  relative.humidity <- beta ^ 8
+  VPD <- PEcAn.data.atmosphere::get.vpd(100*relative.humidity, temp)
+  # combine together.
+  PEcAn.logger::logger.info("Aggregate maps.")
+  met.rast <- c(terra::rast(matrix(temp, nrow = dim(temp)[2], ncol = dim(temp)[1], byrow = T)),
```
mdietze: If I understand correctly, the goal seems to be to generate a single map of mean climate over a specific time period. If you wanted the downscaling to use year-specific meteorology, would you call this function in a loop, iteratively calculating from July 15 of one year to July 14 of the next year, or would it make more sense to have this function generate multiple met layers at different resolutions? Similarly, if/when the SDA drops to a monthly or seasonal timestep, could this function handle calculating monthly mean weather?

modules/assim.sequential/R/SDA_NA_downscale.R
```diff
@@ -0,0 +1,459 @@
...
+  cov.vecs <- vecs %>% purrr::map(function(v){
+    return(v$vec[-na.inds])
+  }) %>% dplyr::bind_cols() %>% `colnames<-`(layer.names) %>% as.data.frame()
+  non.na.inds <- seq_along(matrix(all.rast[[1]]))[-na.inds]
+  return(list(df = cov.vecs, non.na.inds = non.na.inds))
+}
+
+#' @description
+#' This function helps to create the training dataset of specific variable type and locations for downscaling.
+#' TODO: There will be a ratio argument between training and testing samples to testify the ML regression accuracy.
```
mdietze: not following what the TODO is about

modules/assim.sequential/R/SDA_NA_downscale.R
```diff
@@ -0,0 +1,459 @@
...
+#' @title prepare.train.dat
+#' 
+#' @param settings character: physical path that points to the pecan settings XML file.
+#' @param analysis numeric: data frame (rows: ensemble member; columns: site*state_variables) of updated ensemble analysis results from the `sda_enkf` function.
+#' @param covariates.dir character: path to the exported covariates GeoTIFF file.
+#' @param variable character: name of state variable. It should match up with the column names of the analysis data frame. 
+#'
+#' @return matrix within which the first sets of columns contain values of state variables for each ensemble mebers of every site, and the rest columns contain the corresponding covariates.
```
mdietze: 1. clarify the expected dimensions of the matrix to make how things are organized more clear. 2. fix typos

modules/assim.sequential/R/SDA_NA_downscale.R
```diff
@@ -0,0 +1,459 @@
...
+#' 
+#' @author Dongchen Zhang
+prepare.train.dat <- function(settings, analysis, covariates.dir, variable) {
+  # convert settings into geospatial points.
+  pts <- pecan.settings.2.pts(settings)
```
mdietze: If the settings object is only used to get the points, why not just have the points be the function argument? That gives you more flexibility to have pulled the points from the csv or shapefile

modules/assim.sequential/R/SDA_NA_downscale.R
```diff
@@ -0,0 +1,459 @@
+#' @description
```
mdietze: filename includes "NA" which is confusing because 1. for most R users that means Not Available and 2. there's nothing North America specific to any of these functions. The NA specific bit is in the inst/anchor script

mdietze (changes requested):
So at a high level, these new fixes will create duplications between Josh's old code, @dlebauer new code, and @DongchenZ new code in this PR. I'm Ok with that in short term, but highly encourage @DongchenZ @dlebauer and Yang (who wants to downscale fluxes using XGBoost) to work together to come up with an agreed-upon set of functions, arguments, and workflows, and to get these into the bookdown documentation.  I also might suggest that assim.sequential might not be the best long-term home for this code, as it's not part of the SDA itself and can be used to downscale any ensemble runs, not just SDA -- maybe modules/data.mining (which itself could get an updated name, to something like modules/ML?)

modules/assim.sequential/NAMESPACE
DongchenZ: Fixed.

modules/assim.sequential/R/SDA_NA_downscale.R
DongchenZ: Fixed.

modules/assim.sequential/R/SDA_NA_downscale.R
DongchenZ: Fixed. The temporal resolution depends on the start and end dates.

modules/assim.sequential/R/SDA_NA_downscale.R
DongchenZ: Fixed.

modules/assim.sequential/R/SDA_NA_downscale.R
DongchenZ: Fixed.

modules/assim.sequential/R/SDA_NA_downscale.R
DongchenZ: Fixed.

modules/assim.sequential/R/SDA_NA_downscale.R
DongchenZ: Fixed.

mdietze (approved):

mdietze:
@DongchenZ there are a few small errors blocking this PR:

```
 R check of modules/assim.sequential reports the following new problems. Please fix these and resubmit:
  '::' or ':::' imports not declared from:
    ‘abind’ ‘doSNOW’ ‘foreach’ ‘itertools’ ‘PEcAn.data.atmosphere’
  checking Rd \usage sections ... WARNING
  Undocumented arguments in Rd file 'downscale_rf_main.Rd'
    ‘time’ ‘variable’
  Documented arguments not in \usage in Rd file 'downscale_rf_main.Rd':
    ‘time:’ ‘variable:’
  
  Undocumented arguments in Rd file 'parallel_prediction.Rd'
    ‘cov.vecs’
  Documented arguments not in \usage in Rd file 'parallel_prediction.Rd':
    ‘cov.vecs:’
```

PR #3451, opened 2025-02-21
Improve the observation preparation functions for North America SDA experiments.

DongchenZ:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.remote/R/NASA_DAAC_download.R
```diff
@@ -209,4 +241,24 @@ NASA_CMR_finder <- function(doi) {
...
+#' credentials from the environment. 
+#'
+#' @author Dongchen Zhang
+DAAC_Set_Credential <- function(replace = FALSE) {
+  if (replace) {
+    PEcAn.logger::logger.info("Replace previous stored NASA DAAC credentials.")
+  }
+  if (replace | nchar(Sys.getenv("ed_un")) == 0 | nchar(Sys.getenv("ed_un")) == 0) {
+    Sys.setenv(ed_un = sprintf(
+      getPass::getPass(msg = "Enter NASA Earthdata Login Username \n (or create an account at urs.earthdata.nasa.gov) :")
```
mdietze, line 321: The new version seems to require that the system is run interactively. Would there be value in retaining the ability to load authentication from the netrc file? Or could there be instructions on how to run the function prior to execution to set the environment variables? What if the execution is remote (e.g., in the cloud or on a HPC node) where there are different environment variables? Is there a security issue with having passwords stored unencrypted as part of the environment variables? Curious what @robkooper thinks.

mdietze (approved):

PR #3455, opened 2025-02-24
feat: refactor SoilGrids functions to handle multiple properties 

### LINKED ISSUE #3416: Refactor SoilGrids functions
## Description

This is motivated by 
- a discussion in #3406 that included
- need to add coarse fraction to estimation of carbon stocks
- [Results of benchmark testing](https://github.com/PecanProject/pecan/pull/3406#discussion_r1919314617) that show downloading first then extracting points is 100x faster than querying the API for each point. 
- Generalizing functions to work with any soilgrids variable.  

Changes will: 
- Improve the efficiency and scalability of SoilGrids queries, especially for large datasets.
- Simplify and generalize the workflow to handle multiple SoilGrids variables.

## Proposed Solution

Refactor soilgirds_soilC_extract and soilgrids_texture_extraction into three modular functions to increase functionality and flexibility, generalized to work with any variable and depth layer. 

- download_soilgrids_raster a function to download tiff files
- query_soilgrids a function to query data from points or polygons from remote VRTs or from local rasters
- soilgrids_uncertainty to estimate uncertainties
- soilgrids_carbonstocks to calculate OCS to any depth from SOC (to generalize soilgrids 0-30 OCS data product to multiple depths). Need to correct for coarse fraction when converting from SOC (%) to OCS (mass/area).

## Alternatives Considered

Status Quo

FYI @mdietze @DongchenZ @Qianyuxuan @infotroph

AritraDey-Dev:
fixes #3416 

Refactored the SoilGrids functions in `modules/data.land/R/soilgrids_utils.R` to enhance efficiency and scalability. The changes include:
- A new function `download_soilgrids_raster` to handle multiple soil properties.
- Generalization of `query_soilgrids` to query data for multiple soil properties.
- An update to `soilgrids_carbonstocks` to include an optional coarse fraction in carbon stock calculations.


AritraDey-Dev:
I made changes to refactor SoilGrids functions for improved efficiency and scalability; please take a look,i think this works for you.

AritraDey-Dev:
@DongchenZ @Qianyuxuan Can you review this PR and I think it solves the mentioned issue.Just i want you to go through it once and let me know if you feel any changes required.

modules/data.land/R/soilgrids_utils.R
```diff
@@ -0,0 +1,99 @@
...
+#'
+#' @param soil_property The soil property to download (e.g., "ocd", "clay", "sand", etc.)
+#' @param depth The depth layer to download (e.g., "0-5cm", "5-15cm", etc.)
+#' @param quantile The quantile to download (e.g., "mean", "Q0.05", "Q0.5", "Q0.95")
+#' @param extent The extent to crop the raster to (terra::ext object)
+#' @param outdir The directory to save the downloaded raster
+#' @param overwrite Whether to overwrite existing files (default: FALSE)
+#' @return The file path of the downloaded raster
+#' @export
+download_soilgrids_raster <- function(soil_property, depth, quantile, extent, outdir, overwrite = FALSE) {
```
DongchenZ: Normally, we input coordinates (either `extent` or `point`) in the `WGS1984` projection (in lat/lon degrees). I wonder if it could be beneficial to have an extra argument that sets the projection from SoilGrids to the target CRS (WGS1984, for example.)

modules/data.land/R/soilgrids_utils.R
```diff
@@ -0,0 +1,99 @@
...
+#' Query SoilGrids data
+#'
+#' @param points A spatial object (e.g., sf or terra::vect) containing the points to query
+#' @param soil_property The soil property to query (e.g., "ocd", "clay", "sand", etc.)
+#' @param depth The depth layer to query (e.g., "0-5cm", "5-15cm", etc.)
+#' @param quantile The quantile to query (e.g., "mean", "Q0.05", "Q0.5", "Q0.95")
+#' @param local_raster Optional. The file path of a local raster to query. If NULL, queries the remote VRT.
+#' @return A data frame containing the queried data
+#' @export
+query_soilgrids <- function(points, soil_property, depth, quantile, local_raster = NULL) {
```
DongchenZ: See my previous comments for projecting the points from lat/lon degrees to the SoilGrids coordinate system.

modules/data.land/R/soilgrids_utils.R
```diff
@@ -0,0 +1,99 @@
...
+#' @param quantiles A vector of quantiles to estimate uncertainties (default: c(0.05, 0.5, 0.95))
+#' @return A data frame containing the uncertainties
+#' @export
+soilgrids_uncertainty <- function(data, quantiles = c(0.05, 0.5, 0.95)) {
+  uncertainties <- data %>%
+    dplyr::group_by(site_id) %>%
+    dplyr::summarize(
+      mean = mean(value, na.rm = TRUE),
+      sd = sd(value, na.rm = TRUE),
+      quantile_05 = quantile(value, probs = quantiles[1], na.rm = TRUE),
```
DongchenZ: I would not have those hardcoded names for different quantiles. Instead, you might consider calculating them first and naming them afterward based on the quantile we choose.

modules/data.land/R/soilgrids_utils.R
```diff
@@ -0,0 +1,99 @@
...
+  return(do.call(rbind, data_list))
+}
+
+#' Estimate SoilGrids uncertainties
+#'
+#' @param data A data frame containing the SoilGrids data
+#' @param quantiles A vector of quantiles to estimate uncertainties (default: c(0.05, 0.5, 0.95))
+#' @return A data frame containing the uncertainties
+#' @export
+soilgrids_uncertainty <- function(data, quantiles = c(0.05, 0.5, 0.95)) {
```
DongchenZ, line 69: I wonder if the calculation of uncertainty here is correct or not. Our current calculations will first simulate the uncertainty of soil properties across the vertical profile using the `mean` and different quantiles from SoilGrids directly and then summarize the uncertainty once we have a good fit on that (See https://github.com/PecanProject/pecan/blob/72e09bcccb78e27acaa1c6fafc4b5faa9453c84d/modules/data.land/R/soilgrids_soc_extraction.R#L123).

DongchenZ: I think @Qianyuxuan might have better opinions on that.

modules/data.land/R/soilgrids_utils.R
```diff
@@ -0,0 +1,99 @@
...
+#' @param coarse_fraction Optional. A vector of coarse fractions to include in the calculation
+#' @return A data frame containing the calculated carbon stocks
+#' @export
+soilgrids_carbonstocks <- function(soc_data, depths, thickness, coarse_fraction = NULL) {
+  carbonstocks <- soc_data %>%
+    dplyr::group_by(site_id) %>%
+    dplyr::summarize(
+      total_soc = sum(value * thickness, na.rm = TRUE) + 
+                  if (!is.null(coarse_fraction)) sum(coarse_fraction * thickness, na.rm = TRUE) else 0,
+      total_soc_0_30 = sum(value[depths <= 30] * thickness[depths <= 30], na.rm = TRUE)
```
DongchenZ: Why are total carbon and the top 30cm of carbon recorded? Would it be better to have the SOC calculated for any prescribed soil segments (e.g., 0-30 cm, 30-70cm, 70-150cm, etc.)?

DongchenZ (commented):
This PR improves the ability to extract soil properties other than SOC. However, I worry that those functions (which heavily rely on the `terra` package) will be limited to broader spatial extents (e.g., North America, Asia, Africa, etc.) where the computation demand and the memory usage will be unrealistic high. Given that, I think it will be helpful to have a prior check on the volume of data requested (e.g., if the size of the data is too large, then never try it and just return an error message).
For larger spatial extents, it would be more efficient to download the tiled SoilGrids data directly (I usually use the `wget` command) and merge the tiled images using the `gdalwarp` command.

modules/data.land/R/soilgrids_utils.R
AritraDey-Dev: That's a great suggestion! Adding an argument for setting the target CRS will enhance flexibility. I'll implement that for you.

modules/data.land/R/soilgrids_utils.R
AritraDey-Dev: yes, I'll adjust the code to calculate and name the quantiles dynamically

modules/data.land/R/soilgrids_utils.R
AritraDey-Dev: yes, Implementing SOC calculations for customizable soil segments would enhance the analysis and provide more relevant insights.

AritraDey-Dev:
> This PR improves the ability to extract soil properties other than SOC. However, I worry that those functions (which heavily rely on the `terra` package) will be limited to broader spatial extents (e.g., North America, Asia, Africa, etc.) where the computation demand and the memory usage will be unrealistic high. Given that, I think it will be helpful to have a prior check on the volume of data requested (e.g., if the size of the data is too large, then never try it and just return an error message). For larger spatial extents, it would be more efficient to download the tiled SoilGrids data directly (I usually use the `wget` command) and merge the tiled images using the `gdalwarp` command.


That's an important point! Implementing a data volume check could help manage resource usage effectively. Your approach of using tiled downloads and merging with `gdalwarp` is a practical solution for larger extents.

modules/data.land/R/soilgrids_utils.R
Qianyuxuan: I am confused about how you calculate sd here: "sd = sd(value, na.rm = TRUE)" with known quantiles and means. Usually, we estimate means and sds based on the best fits. For organic carbon density (ocd), it was found to follow gamma distribution best based on reported means and quantiles. But for soil texture data (sand, clay, and silt fractions), they follow Dirichlet distribution and estimate of corresponding parameters can be found in my pending PR here: https://github.com/PecanProject/pecan/pull/3406/files#diff-8b26bae8174687f7e2bf631c5188b979a5b8e4bdacc5c604da84e09a2abc0058

modules/data.land/R/soilgrids_utils.R
```diff
@@ -0,0 +1,111 @@
...
+#' @export
+soilgrids_carbonstocks <- function(soc_data, depths, thickness, segments, coarse_fraction = NULL) {
+  carbonstocks <- data.frame(site_id = unique(soc_data$site_id))
+  
+  for (segment in segments) {
+    lower_bound <- segment[1]
+    upper_bound <- segment[2]
+    
+    total_soc <- soc_data %>%
+      dplyr::filter(depths >= lower_bound & depths < upper_bound) %>%
```
Qianyuxuan, line 100: Is “soc_data”  soil organic carbon content (SOC, in g/kg) or organic carbon density (OCD, in kg/m³)? I guess it is OCD if you multiply it by thickness (see https://www.isric.org/explore/soilgrids/faq-soilgrids). For total_SOC, it should be something like "total_SOC=sum(OCD*thickness*(1-coarse_fraction))" but can you explain your calculation here?

AritraDey-Dev: The `soc_data` refers to organic carbon density (OCD) in kg/m³. The calculation for `total_SOC` is performed as follows: 

\[
\text{total\_SOC} = \sum(\text{OCD} \times \text{thickness} \times (1 - \text{coarse\_fraction}))
\]

This accounts for the volume of soil represented by the thickness and adjusts for the coarse fraction to provide an accurate estimate of total SOC.

AritraDey-Dev: Screenshot from 2025-02-26 23-51-42

AritraDey-Dev:
@Qianyuxuan @DongchenZ any insights on this changes?

modules/data.land/R/soilgrids_utils.R
mdietze: If I'm reading the current code correctly, it's grouping by site, and thus calculating a mean and sd over depth, which is wrong on many levels (pun intended). Sampling from rnorm is also wrong. This code is in no way a "refactor" of the existing code

modules/data.land/R/soilgrids_utils.R
mdietze: 1. your math seems right, but that math isn't what's implemented in the code
2. I'm not sure the "if" will work as intended for vector data
3. The "if" assumes that if the coarse fraction is NULL then it's 0. Setting aside that this might not be a valid assumption, it would be far simpler to do that substitution earlier in the pipe

PR #3458, opened 2025-02-24
Improve pkgdown build process and fix documentation generation

### LINKED ISSUE #18: pkgdown docs

divine7022:
## Description

*   **Correct Documentation Generation:** Fixed `script/build_pkdown.R` to ensure `pkgdown::build_site()` correctly generates documentation within the correct package directories. This resolves issues where documentation was not being created for some packages.
*   **Symlink Handling:** Ensured symlinks are preserved during the copy process from `docs` to `book_source/pkgdocs`. This prevents broken assets (CSS, JavaScript, images) in the final documentation.
*   **Robustness and Error Handling:** Improved error handling in `build_pkdown.R` to prevent failures for individual packages from halting the entire build process. The script now includes a `tryCatch` block with a `finally` clause for proper working directory management, and it stops execution on critical errors.
*   **Package Structure Updates:** Updated the package structure for `base`, `models`, `modules`, and `book_source` to align with the new build process. 
*   **File Inclusion:** Included the necessary files: `scripts/build_pkgdown.R`, `Makefile`, `book.yml`, `docker/base/dockerfile`, and `docker-compose.yml`.
*   **Book Integration:** Ensured `pkgdown` output is correctly copied to `book_source/pkgdocs` for seamless integration into the bookdown site.

## Motivation and Context

This change is required to ensure that the `pkgdown` documentation for all PEcAn packages is generated correctly and consistently. Previously, issues with `symlink` handling and incorrect working directory management led to broken documentation and missing files. This PR addresses these problems, providing a more robust and reliable documentation build process. It also automates the process of copying the generated documentation to the `bookdown` source directory, simplifying the integration of `pkgdown` output into the book.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

fixes https://github.com/PecanProject/PecanProject.github.io/issues/18


scripts/build_pkgdown.R
```diff
@@ -0,0 +1,58 @@
...
+    } else { 
+      temp_dest <- paste0(dest, "_temp")
+      dir.create(temp_dest, recursive = TRUE, showWarnings = FALSE)
+      file.copy(from = source_docs, to = temp_dest, recursive = TRUE, overwrite = TRUE)
+      file.rename(temp_dest, dest)
+    }
+    message(paste("✅ Successfully copied docs from", pkg, "to", dest))
+  }, error = function(e) {
+    warning(paste("❌ Error building pkgdown site for", pkg, ":", e$message))
+  }, finally = {
```
infotroph, line 49: Probably good to handle the warning case too:
```suggestion
  }, warning = function(w) {
    warning(paste("⚠️ Warning building pkgdown site for", pkg, ":", w$message))
  }, finally = {
```

scripts/build_pkgdown.R
```diff
@@ -0,0 +1,58 @@
...
+      dir.create(dest, recursive = TRUE, showWarnings = FALSE)
+    }
+    if (getRversion() >= "4.2.0") {
+      dir.copy(from = source_docs, to = dest, recursive = TRUE, overwrite = TRUE)
+    } else { 
+      temp_dest <- paste0(dest, "_temp")
+      dir.create(temp_dest, recursive = TRUE, showWarnings = FALSE)
+      file.copy(from = source_docs, to = temp_dest, recursive = TRUE, overwrite = TRUE)
+      file.rename(temp_dest, dest)
+    }
```
infotroph: This section failed for me because `dir.copy` isn't found, and sure enough I'm not aware of a `dir.copy` command having existed in any version of R, and can't find any references to it in the R 4.2 release notes as I'd expect if it were new then. Can you explain more about what you're trying to work around?

Since `file.copy` exists in any R version and works on directories with `recursive = TRUE`, am I missing a reason this can't simplify to...?
```suggestion
    file.copy(from = source_docs, to = dest, recursive = TRUE, overwrite = TRUE)

```

docker/base/Dockerfile
```diff
@@ -3,6 +3,9 @@ ARG IMAGE_VERSION="latest"
 ARG PARENT_IMAGE="pecan/depends"
 FROM ${PARENT_IMAGE}:${IMAGE_VERSION}
 
+# Install R package for pkgdown
+RUN R -e "install.packages(c('pkgdown', 'remotes'))"
+
```
infotroph, line 8: documentation lives in the `docs` image (see `docker/docs/Dockerfile`), not base; I recommend putting the pkgdown docs there are well

docker-compose.yml
```diff
@@ -135,6 +135,7 @@ services:
       - USERID=${UID:-1001}
       - GROUPID=${GID:-1001}
     volumes:
+      - ./:/pecan
```
infotroph, line 138: I think we want to add the compiled pages to the `docs` image at build time rather than sharing them as a volume -- see my comment below. But if @robkooper contradicts me, listen to him instead :)

Makefile
```diff
@@ -115,6 +115,11 @@ check_modules: $(BASE_I) $(MODULES_C)
 
 document: $(ALL_PKGS_D) .doc/base/all
 
+pkgdocs:
+	Rscript scripts/build_pkgdown.R $(ALL_PKGS) || exit 1
+	
+.PHONY: all install check test document shiny pkgdocs
```
infotroph: 1. I'm on the fence about whether pkgdown build should happen here or in the `book_source/` build (which is currently separate from this Makefile). The fact this is where we already have a list of all the packages certainly helps make the case for doing it here.
2. If building here, add `base/all` to the target list (confusingly, it's not part of `$ALL_PKGS`) 
3. there's already a .PHONY declaration on line 106, so please add `pkgdocs` to that instead of making a new one here
```suggestion
pkgdocs:
	Rscript scripts/build_pkgdown.R $(ALL_PKGS) base/all || exit 1
	
```

scripts/build_pkgdown.R
```diff
@@ -0,0 +1,58 @@
...
+    setwd(pkg) 
+    pkgdown::build_site() 
+    setwd(current_wd) 
+    source_docs <- file.path(pkg, "docs")
+    if (!dir.exists(source_docs)) {
+      warning(paste("No docs folder created for:", pkg))
+      next 
+    }
+    dest <- file.path("book_source/pkgdocs", pkg)
```
infotroph, line 38: This moves files in the right direction, but note that at build time we compile all the Markdown in `book_source` into `book_source/_book` and then deploy that. So as currently written, the pkgdown files will be left behind and not deployed. 

I'm not immediately sure whether that means the pkgdown script should keep writing to `book_source/pkgdocs` and then the book build should copy it into the right place inside `_book`, or if the pkgdown build should be a subprocess of the book build, or what. I'm open to suggestions here.

infotroph (changes requested):
Thanks! The build_pkgdown script works great for me other than one unknown function (see my inline comments). I see more decisions to make on integrating it into our existing deployment by picking where to put the built files and when to call the build process from our existing build scripts.

scripts/build_pkgdown.R
divine7022: @infotroph Thanks for your review !
currently, `errors` are handled explicitly, but `warnings` are not caught separately they just print to the console. Adding a warning block can be useful for better error handling and debugging.

scripts/build_pkgdown.R
divine7022: Thank you for pointing this out! You're absolutely correct .I had intended to implement a workaround for older R versions (< 4.2.0) using a temporary directory and `file.rename()`. However, I realize this is unnecessary because `file.copy()` with `recursive = TRUE` has been available in all versions.

scripts/build_pkgdown.R
divine7022: I was looking into the issue with `pkgdocs` not being included in `_book`, and I see two possible solutions:
- Modify the script to place `pkgdown` files directly in `_book` by adjusting the destination path
`dest <- file.path("book_source/_book/pkgdocs", pkg)`. This ensures the files are placed correctly during the build process.

- Modify the book build process to copy `pkgdocs` into `_book` by adding a step to the `Makefile` or deployment script to copy the files after the build is complete.
`cp -r book_source/pkgdocs book_source/_book/`
This simple fix keeps the pkgdown script independent but ensures the files are included in the final output.

dlebauer: Hi @divine7022 great work on improving the pkgdown build process - thank you for taking this on! 

At today's PEcAn meeting we decided that the generated pkgdown docs should be pushed to a separate repository - similar to the book being pushed to pecan-documentation, but kept in a separate repository. A key concern was making sure that the two builds (docs and package docs) don't depend on each other. I created the package-documentation repository at https://github.com/PecanProject/package-documentation. Could you please update the build process to deploy the pkgdown output directly to that repository?

divine7022: @dlebauer Hello sir!
Thanks for the update! that's really a great approach, That makes sense to keep the builds independent. I'll modify the deployment workflow to generate the pkgdown documentation and push the output directly to the package-documentation repository. Thanks for setting it up!

dlebauer: Thanks! Let me know if you need me to configure the repository or anything else.

PR #3459, opened 2025-02-24
Lpjguess mac compile

infotroph:
## Description

Removes templated `max` and `min` macros and replaces their calls with c++ builtins `fmax` and `fmin`. Unless I've missed something, all the variables passed to these functions (at least for the uses of the in the existing PEcAn.LJPGUESS code -- I have not inspected their calls elsewhere in the LPJ-GUESS model code) are already declared with type `double`, so the type flexibility of the templates is not needed.

## Motivation and Context

Compilation of PEcAn.LPJGUESS was failing on my Mac with `unknown type name '_GLIBCXX14_CONSTEXPR'`. @yinghaoSunn did some great detective work and diagnosed this as one of several macros provided by GCC's glibc that are not present in the libc Apple provides with their version of Clang. Although it would theoretically be possible to declare a SystemRequirements on GNU glibc, in practice that's not automatically enforced and would not prevent build failures on Macs. 

Instead, I propose to use a simpler equivalent that compiles and passes package tests on all systems I've tested it on. __To state the obvious though__, this is introducing more divergence between the LPJ-GUESS model and the code copied into the PEcAn package. I'm opening this PR to start discussion, but will defer fully to actual users of the model on whether compilation across OSes is more or less important than keeping model and PEcAn code in sync.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze:
Assigned to @yinghaoSunn and @istfer but also tagging @HannahOgrady and @MagicForrest to get their opinions

MagicForrest:
Yes, this looks like a good idea to me.  In this case, I don't think you need to worry about the divergence between the LPJ-GUESS code and the code in PEcAn.  There is really no reason for not using `fmin` and `fmax` in LPJ-GUESS, just "historical reasons". So go for it.

mdietze (approved):

PR #3464, opened 2025-02-28
Avoid bottleneck from duplicating settings during sublist extraction

infotroph:
Previous code created extra copies of items on every subsetting operation, even if the extracted item was not modified.

Profiling with a 98-site MultiSettings showed ~90% of runtime spent here even for tasks like "check if this setting exists for one specific site", where the values of other nodes are irrelevant. 

I switch here to comparing x[[1]] in place, which completely eliminates GC pauses and the runtime of the script that sent me down this road goes from 4.5 minutes down to 1.5 seconds. Folks working with only a few sites will probably not see as big a speedup, but this should be either an efficiency win or at worst a no-change for everyone.

All package tests pass and I can't find any differences in output from the settings-manipulation scripts I have handy, but more tests appreciated since this is a component that's been stable in lots of workflows for a while now.

Complementary to the path insertion changes in #3456, but submitting separately because it's an improvement whether or not those go in.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/settings/R/MultiSettings.R
+++ b/base/settings/R/MultiSettings.R
@@ -105,13 +105,7 @@ is.MultiSettings <- function(x) {
 } # "[[.MultiSettings"
 
 .allListElementsEqual <- function(x) {
-  firstElement <- x[[1]]
-  replicatedFirstElement <- replicate(
-    length(x),
-    firstElement,
-    simplify = FALSE)
-  return(isTRUE(
-    all.equal(replicatedFirstElement, x, check.attributes = FALSE)))
+  all(sapply(x, identical, x[[1]]))
 } # .allListElementsEqual
 
 #' @export


infotroph:
The profiling results are very impressve!

## Before
Screenshot 2025-02-27 at 3 29 01 PM

## After
Screenshot 2025-02-27 at 3 31 18 PM

mdietze:
@rykelly I don't know if you remember the logic behind the original design here to decide whether @infotroph change is compatible?

rykelly:
I can't say that I remember how I landed on that logic (or much of the context, for that matter), but revisiting now it certainly seems unnecessarily expensive copy so much data for such a simple task. 

That said, even for my early years the formulation seems surprisingly obtuse, and I wonder if it had something to do with that `check.attributes = FALSE`... Perhaps I also intended to allow the numeric tolerance of `all.equal()`, but I don't remember enough about the contents of settings to know if that makes sense. 

On the other hand, I find it entirely plausible that I just hacked my way to this solution because it worked at the time, and none of the details of `all.equal()`'s behavior are important :) I think my use case involved model ensembles where the inefficiency in mucking around with settings wouldn't have caught my attention anyway. (Not to mention present-day me is loathe to trust anything written in _CamelCase_ by that schmuck from the past. Ugh.)

As fuzzy as my memory is, I think you guys are in a much better position to know whether the nuances of `all.equal()` vs. `identical()` are relevant. If the new version works the same in practice then multiple orders of magnitude speedup sure seems like a win. (And even if you do need aspects of the old behavior, I'm sure my original strategy isn't the most efficient way to get them.)

Sorry, I can't be more help, but thanks for the blast from the past! I hope y'all are doing well.

mdietze (approved):
Approving on @rykelly recommendation, though will ask that you keep an eye out for odd behaviors / new bugs.

PR #3474, opened 2025-03-06
met2model.SIPNET: Refuse to write NAs

infotroph:
## Description

Updates met2model.SIPNET to check for missing values after all conversions are finished and before writing the clim file, and if any values are missing to drop those rows with a warning.

Discussion point: Dropping rows does create the possibility of irregular time step sizes in the simulation. I've included here some reporting to help users decide how much to worry about it, but I can also see an argument for going one step pickier and making missing values cause met2model to stop with an error and write no clim file at all. Opinions, anyone?

Further changes while I was in the file:

* Reduced the number of digits written into the clim file from unspecified (15 decimal places unless the user has set `options("digits")` to 4, which cuts filesize in ~half and eases diagnostics later
* Eliminated an if statement that had no effect
* Documentation polishing

## Motivation and Context

Clim files are required to contain no missing values, and if any are present they can cause an [infinite loop](https://github.com/PecanProject/sipnet/issues/38) in Sipnet. Fortunately most users can avoid this by gapfilling their data before conversion or by using a reanalysis met product without gaps, but there are still enough ways for values to be missing that it seemed worth a check at write time.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/NEWS.md
+++ b/models/sipnet/NEWS.md
@@ -5,8 +5,8 @@
 
 ## Fixed
 
-* `met2model.SIPNET()` now checks for missing values (which are not allowed in SIPNET inputs; see https://github.com/PecanProject/sipnet/issues/38) and removes rows that contain them, with a warning.
-* `write.config.SIPNET()` now consults PFT trait definitions to decide whether to initialize LAI in the leaf-on or leaf-off state, instead of the previous hard-coded defaults. Specifically:
+* `met2model.SIPNET()` now stops with an error if the result contains missing values, which [are not allowed](https://github.com/PecanProject/sipnet/issues/38) in SIPNET inputs (#3474).
+* `write.config.SIPNET()` now consults PFT trait definitions to decide whether to initialize LAI in the leaf-on or leaf-off state, instead of the previous hard-coded defaults (#3419). Specifically:
 	- a PFT with `fracLeafFall` > 0.5 will be treated as deciduous (previously hardcoded to deciduous for anything other than boreal conifers)
 	- deciduous PFTs will get laiInit=0 if the simulation start date is not between `leafOnDay` and `leafOffDay` (previously hardcoded to May through September)
 * The generated Sipnet run script (job.sh) now works correctly, including across machines, when met/input/output files are specified relative to the working directory (#3418). Absolute paths continue to work as always.
--- a/models/sipnet/NEWS.md
+++ b/models/sipnet/NEWS.md
@@ -5,6 +5,7 @@
 
 ## Fixed
 
+* `met2model.SIPNET()` now checks for missing values (which are not allowed in SIPNET inputs; see https://github.com/PecanProject/sipnet/issues/38) and removes rows that contain them, with a warning.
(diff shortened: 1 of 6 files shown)


models/sipnet/R/met2model.SIPNET.R
```diff
@@ -297,22 +324,62 @@ met2model.SIPNET <- function(in.path, in.prefix, outfolder, start_date, end_date
...
 
-    
-    if (is.null(out)) {
-      out <- tmp
-    } else {
-      out <- rbind(out, tmp)
-    }
-    
+
+    out <- rbind(out, tmp)
```
infotroph, line 327: rbind(NULL, x) returns x unchanged, so there was no need for a special case here.

mdietze: Just note that if you try to rbind an undefined variable it throws an error.

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -297,22 +324,62 @@ met2model.SIPNET <- function(in.path, in.prefix, outfolder, start_date, end_date
...
   if (!is.null(out)) {
-    
+
+    # Sipnet does not know how to handle missing values in clim files.
+    # -- No, say it louder: Missing values send Sipnet into _infinite loops_.
+    # Let's remove these with warnings, hmm?
+    if (anyNA(out)) {
+      n_all <- nrow(out)
+
+      out <- na.omit(out)
```
mdietze: This isn't valid. If there are any missing met values the run should fail and the user should get a useful error message. If this is happening regularly with reanalysis met then either there's a bug in our workflow or we should be running our existing gapfilling function on reanalysis met, not just site-level observational met (which is the workflow default)

mdietze (changes requested):

models/sipnet/R/met2model.SIPNET.R
infotroph: Fair point. I was thinking SIPNET doesn't care about the uneven timesteps, but skipped over how that would mean needing to rescale all the cumulative met vars for the obs that now includes the gap time.

I'll convert this to an error

models/sipnet/R/met2model.SIPNET.R
infotroph: Sure, as does evaluating is.null on an undefined variable ;) 
(It's defined on line 107 above, so we're good either way)

models/sipnet/NEWS.md
```diff
@@ -5,7 +5,8 @@
 
 ## Fixed
 
-* `write.config.SIPNET()` now consults PFT trait definitions to decide whether to initialize LAI in the leaf-on or leaf-off state, instead of the previous hard-coded defaults. Specifically:
+* `met2model.SIPNET()` now stops with an error if the result contains missing values, which [are not allowed](https://github.com/PecanProject/sipnet/issues/38) in SIPNET inputs (#3474).
```
```suggestion
* `met2model.SIPNET()` now stops with an error if the result contains missing values, which [are not allowed](https://github.com/PecanProject/sipnet/issues/38#issuecomment-2701749926) in SIPNET inputs (#3474).
```

mdietze (approved):

PR #1733, opened 2017-10-20
Added new Rmd template for basic PEcAn workflow to PEcAn.all package

dlebauer:
## Description

Added a template for a basic PEcAn workflow.

## Motivation and Context

Will help users make the transition from Web to interactive R session by adding an Rmarkdown template that users can use to create a new workflow from within the Rstudio interface.

image

I didn't know where to put the documentation, but it would be 

In Rstudio with the PEcAn.all package loaded 

```r
library(PEcAn.all)
```

* File --> new file --> Rmarkdown --> From Template
* Choose the PEcAn workflow template.
* You can start with a pecan.xml file from the web interface or the tests/ directory and modify as needed

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #2784, opened 2021-03-08
Add API endpoint to kill a workflow


ashiklom:
Sometimes, we make boo boos (e.g., requesting 100 years of met download -- oops!). In those cases, it would be nice to have a shortcut to cancel a currently running workflow. Something like:

```r
rpecanapi::kill_run(server, workflowid)
# Candidate API endpoint: <IP>/api/workflows/cancel?workflowid=XXXX
```

Currently, to cancel a run, you need to do some combination of:

- Terminate the R process in Portainer
- Purge the RabbitMQ queue
- Restart RabbitMQ

Would be nice to do this in a more convenient way.

AritraDey-Dev:
Is this issue still on? 
After going through docs and `apps/api/R/workflows.R` it seems to be not done yet.
Should I work on this?

mdietze:
AFAIK no one has worked on this. That said, if you want to dive into the API I'd argue that an easier and more urgent need would be to break up the monolithic "workflow" (which calls web/workflow.R) into additional endpoints that breaks things up into smaller modules for each major step. This would allow much greater flexibility in PEcAn usage. For example, if one had a pecan settings that had all of the inputs already configured, one could in theory call an API that just calls the run.write.configs module and the start.model.runs module (or even break these into two separate steps) and then the use could use other existing APIs to grab the outputs. To get to that state where you start the model runs there could be separate API end points for preparing the meteorology, soils, initial conditions, phenology, parameter files, etc. Such an effort might dovetail nicely with the GSOC project about notebook-based execution, as well as the one about parallelization, but to be 100% honest I think this idea (both in it's original form and in my suggestions) is far less urgent a priority than diving directly into the core tasks of those projects.

AritraDey-Dev:
Thanks for the explanation! Given the priorities, do you think it's worth starting some initial work on breaking up the workflow, or should I wait until the GSOC projects progress further?

mdietze:
I'd say that breaking up the workflow via direct execution of R code (e.g. R Markdown) is a higher priority than breaking up the workflow via API (though I'd expect the latter could/should also include an example/template notebook). In terms of "waiting until GSOC project progress further" that would depend on whether you're talking about waiting for someone else to complete tasks you need, or waiting to start those tasks yourself. If the latter I don't see a reason to wait.

AritraDey-Dev:
Got it! I'll start working on breaking up the workflow via direct R execution. Here’s my planned approach:

[...]

PR #3486, opened 2025-03-14
Rmd template for running workflow

### LINKED ISSUE #1866: Add Rmd template for basic PEcAn workflow

AritraDey-Dev:
## Description
This is for the issue #1866  and following up on the discussion by @mdietze in the issue #2784 

This pull request introduces a new R Markdown file for the PEcAn modular workflow, which includes loading necessary packages, reading settings, and running various analyses. The key changes are summarized below:

New R Markdown file for PEcAn modular workflow:

* Added file `web/workflow_modular.Rmd` with metadata including title, author, date, and output format.
* Loaded PEcAn packages and settings files to prepare for the workflow execution.
* Implemented trait analysis to fetch plant trait data and prior distributions.
* Performed meta-analysis to derive probabilistic distributions for model parameters.
* Generated model configuration files, executed model simulations, and retrieved results for further analysis.

## Motivation and Context

This PR fixes #1866  

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


web/workflow_modular.R
```diff
@@ -0,0 +1,115 @@
...
+      PEcAn.utils::status.end()
+    }
+  }
+  
+  return(settings)
+}
+
+run_model_execution <- function(settings_path, debug = FALSE) {
+  # Load settings
+  settings <- PEcAn.settings::read.settings(settings_path)
```
mdietze: I don't think you want to re-load the settings object each time. Just pass the settings object, not the settings path

web/workflow_modular.R
```diff
@@ -0,0 +1,115 @@
...
+  
+  # Write configs
+  if (PEcAn.utils::status.check("CONFIG") == 0) {
```
mdietze: status.check doesn't do much outside of the web interface, I think these bits can be dropped

web/workflow_modular.R
```diff
@@ -0,0 +1,115 @@
...
+    if (debug) cat("Writing model configurations...\n")
+    
+    PEcAn.utils::status.start("CONFIG")
```
mdietze: Same with the status.start and status.end

mdietze (changes requested):
First, I think the idea is to work with a Rmd, not a R script. Second, if code is in a Rmd code block, I don't see the advantage of also putting it in a function.

AritraDey-Dev:
thanks for reviewing ! i will make those changes in the new commits.

web/workflow_modular.R
AritraDey-Dev: You're right! `PEcAn.utils::status.check() ` is  primarily useful for the web interface, and for a standalone script, it doesn't add much value. We can safely remove those checks and simplify the script while ensuring proper execution

AritraDey-Dev:
@mdietze I think we should keep this file in `base/all/inst/` directory instead of keeping it in web directory?
Curious to know your thought on the changes.

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,57 @@
+---
+title: "PEcAn Modular Workflow"
+author: "Aritra Dey"
+date: "`r Sys.Date()`"
+output: html_document
+---
+
+```{r libraries}
```
robkooper: Since this is a RMarkdown file, can you maybe add some text about what each section does. This will help for novice users to understand why this specific function is needed.

robkooper: Maybe add a short description and a link to the documentation.

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,57 @@
...
+runModule.get.results(settings)
+```
+
+```{r model-analysis}
+run.sensitivity.analysis()      # Run sensitivity analysis and variance decomposition on model output
+run.ensemble.analysis()  	      # Run ensemble analysis on model output. 
+run.ensemble.analysis(plot.timeseries=TRUE) 
+```
+
+```{r finish}
```
robkooper: Not sure this is needed. You know when it is done, when the last call returns.

robkooper (changes requested):

web/workflow_modular.Rmd
AritraDey-Dev: The `finish` section isn't strictly necessary since the workflow naturally ends when the last function call completes. I included it as a simple confirmation message, especially useful when `settings$debug ` is enabled.

AritraDey-Dev: Let me know if you prefer it removed!

AritraDey-Dev:
@mdietze @robkooper  I am curious about  your input on this PR—I’ve added descriptions for each section for clarity.Could you please review this once?

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,79 @@
...
+---
+
+# Load PEcAn packages
+```{r libraries}
+library("PEcAn.all")
+```
+# Load PEcAn settings files.
+
+Open and read in settings file for PEcAn run.
+To create a pecan.xml, you can download one generated in the PEcAn web interface or one of the `pecan.<modelname>.xml` files in the tests/ directory of the PEcAn repository (github.com/pecanproject/pecan).
```
mdietze: 1. This is a good place to start, but I think in the longer term (not this PR) we'll want the Rmd to help build the settings and deprecate the web interface
2. There might be useful info you can pull from the tutorials (e.g. Demo 1, Demo 2, etc) to help populate the text between code blocks. Similar to (1), the long term goal (not this PR) is also to update those tutorials around the notebook-based  interface

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,79 @@
...
+```{r load-settings}
+
+settings_path <- "settings.xml" 
+settings <- PEcAn.settings::read.settings(settings_path)
+settings <- PEcAn.utils::do_conversions(settings)
```
mdietze: do.conversions 100% needs to be in a separate block. 

Also, you're missing the steps that check/update the settings. e.g. PEcAn.settings::prepare.settings

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,79 @@
...
+
+if (is.null(settings$debug)) {
+  settings$debug <- FALSE  # Default to FALSE if not defined
+}
+```
+
+# Trait Analysis
```
mdietze: I'm not sure if we want the Trait Analysis and Meta Analysis in the default workflow. If we do they should be combined into one block since they're run together by default. If the block is retained, the text need to be expanded to describe this as OPTIONAL and what you need to provide if you elect not to run this (i.e. specify a posterior file in the settings object).

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,79 @@
...
+settings <- runModule.run.write.configs(settings)
+```
+
+
+# Run Model and Retrieve Results
+Executes the model simulations and retrieves the results for further analysis.
+
+```{r run-model}
+runModule.start.model.runs(settings)
+runModule.get.results(settings)
```
mdietze: move get.results to the Model Analyses block. This function is specific to those analyses

web/workflow_modular.Rmd
```diff
@@ -0,0 +1,79 @@
...
+```
+
+# Model Analyses
+Performs various analyses on model outputs, including sensitivity analysis, variance decomposition, and ensemble analysis.
+
+```{r model-analysis}
+run.sensitivity.analysis()      # Run sensitivity analysis and variance decomposition on model output
+run.ensemble.analysis()  	      # Run ensemble analysis on model output. 
+run.ensemble.analysis(plot.timeseries=TRUE) 
+```
```
mdietze: Could be beyond this PR, but I think either this block, or a block after, would be a great place to visualize the results of these analyses.

web/workflow_modular.Rmd
mdietze: Also, could be beyond this PR, but I think either this block, or a block after, would be a great place to show how to visualize the model outputs (e.g. a simple time series plot, a simple bivariate scatter plot). These would replace the interactive visualizations in the old web portal. I'd put this as #1 priority for the next PR.

web/workflow_modular.Rmd
mdietze: I'd prefer to have this block removed

mdietze (changes requested):

web/workflow_modular.Rmd
AritraDey-Dev: I see your point. I'll merge them into one block .

AritraDey-Dev: we have to make sure to clarify that if users skip the Trait & Meta Analysis step, they must manually specify a posterior file in the settings to ensure the model has the necessary input data for further analysis.

mdietze: Correct. In practice, specifying a posterior file is the default way most runs are done, but it's also true that PEcAn doesn't ship with default posterior files for any models, which argues for keeping these tasks in the workflow. That said, it would be useful to add text/code to show users how to grab the posterior from this analysis so that they can re-use it in their future analyses. It's also worth noting, in the text, that this chunk is one of the few where a connection to the BETY database is currently no optional. A workflow that can run front-to-back with BETY being optional is desirable since it makes installation much simpler and also makes it possible to run in a HPC environment. One thing we could also do is to take the posterior files associated with published analyses (e.g. Fer et al 2018) and make sure they are somewhere publically archived and machine readable so that users could do a demo that doesn't require BETY.

web/workflow_modular.Rmd
AritraDey-Dev: would something simple like this work?
`PEcAn.visualization::plot_netcdf(datafile, yvar, xvar, width, height, filename, year)` (after taking the inputs)

mdietze: Seems reasonable.  Once you have a fully working workflow it would be nice if you could post a copy of the knit report so we all can see what the workflow looks like once run.

A few other things to note:

1. Please make sure to update changelogs
2. Please make sure to update the overall Documentation to reference this workflow
3. We should think about what sort of test need to be added to ensure this workflow continues to function (e.g., no new PRs break the workflow). My gut instinct is that this would be an integration test that runs the full workflow for any PR similar to the existing SIPNET Github Action. That said, adding an entirely new GH Action may be beyond an initial PR but is the sort of thing we should follow up on quickly.

AritraDey-Dev:
Screenshot from 2025-03-23 21-25-36
@mdietze, I've been stuck on this issue for a while despite multiple debugging attempts—any insights would be really helpful !

mdietze:
For that bug, have you identified what line of code is throwing the error and what the values are of the arguments being passed? If you can do that it's usually clear which argument is invalid, and then you can traceback to figure out where that input got misspecified or corrupted

AritraDey-Dev:
> Screenshot from 2025-03-23 21-25-36 @mdietze, I've been stuck on this issue for a while despite multiple debugging attempts—any insights would be really helpful !

yes i tried to log them.It looks something like this...

Screenshot from 2025-03-23 21-45-55

But not sure why the values are NULL.

AritraDey-Dev:
[test.pdf](https://github.com/user-attachments/files/19410798/test.pdf)

@mdietze I am sharing the knit report for the workflow, with only the last step (running the workflow) removed.

From my investigation, the issue seems to stem from logging into RStudio with a different user, which prevents the correct configuration of the path to `job.sh`(As sometime it take values from pecan directory). I believe the workflow should function correctly in a properly configured environment. The  changelogs and documentation i will add soon.

AritraDey-Dev:
I believe this PR is ready now. @mdietze @robkooper  When you have a moment, could you please take a quick look and let me know your feedback? I have also shared the knit report above.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
+# Introduction
+
+This tutorial provides a step-by-step guide to running the **PEcAn Modular Workflow**. The PEcAn (Predictive Ecosystem Analyzer) system automates ecological modeling, helping researchers analyze plant functional traits, run model simulations, and perform sensitivity analyses.
+
+After setting up PEcAn locally, open **RStudio** in your browser and log in with:
```
mdietze: These instructions are specific to the Docker stack.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+
+---
+
+# Prerequisites
+
+Before running this workflow, ensure you have:
+
+- Installed **PEcAn** and its dependencies.
+- An XML settings file (`settings.xml`) configured for your use case.
+- A model binary (e.g., **SIPNET** or **ED2**) specified in your settings.
```
mdietze: 1. A couple lines up refers to pecan.xml but here we refer to settings.xml, this will be confusing to new users
2. Can we add an example pecan.xml? For now it could be something configured to run specifically in the default Docker stack, but in the future we'll want to add a section to the Rmd itself to build/update the settings object.
3. I don't think you need an "or" in the e.g., especially if the default starting point will be SIPNET

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+
+---
```
mdietze: From here down appears to mostly duplicate the Rmd itself, which is unnessisary and redundant, meaning that it will also be hard to maintain as any changes there will have to be duplicated here. How to run the Rmd should be self-documenting.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
+# Introduction
+
+This tutorial provides a step-by-step guide to running the **PEcAn Modular Workflow**. The PEcAn (Predictive Ecosystem Analyzer) system automates ecological modeling, helping researchers analyze plant functional traits, run model simulations, and perform sensitivity analyses.
```
mdietze: Rather than a step-by-step guide to running the Rmd, I'd recommend you instead document what the Rmd workflow is and why it exists. Essentially, you're currently providing a guide to run something whose existence/purpose you haven't yet explained.

This information may belong somewhere else in the overall documentation. For example, section 5.1 says there are 2 ways to run PEcAn (web interface and directly within R) and recommends the web interface as the starting point, but we want to get to point where the Rmd becomes the recommended option (we aren't there yet). But we're at the point where it should be mentioned in section 6 (aside, I think section 7 can be joined with 6) as an alternative to the basic web workflow. It's also unclear to me where/if this doc is referenced in 5.2 (pecan demos table) but we shouldl try to get to the point where we might have Basic Run: RStudio and Basic Run: web browser as the first two items (where the latter is the current basic run)

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+- An XML settings file (`settings.xml`) configured for your use case.
+- A model binary (e.g., **SIPNET** or **ED2**) specified in your settings.
+
+---
+
+# Load PEcAn Packages
+
+We first load the necessary PEcAn modules required for running the workflow.
+
+```{r libraries, message=FALSE, warning=FALSE}
```
mdietze: add an echo=FALSE so this output doesn't show up in the knit doc. That said, not sure why we want warning=FALSE

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+
+---
+
+# Load PEcAn Settings
+
+We read the PEcAn settings from an XML file. This file contains configuration details such as input datasets, model selection, and output directories.To create a pecan.xml, you can download one generated in the PEcAn web interface or one of the `pecan.<modelname>.xml` files in the tests/ directory of 
+the PEcAn repository (github.com/pecanproject/pecan).
+
+```{r load-settings}
+settings_path <- "settings.xml"  # Path to the settings file
```
mdietze: As noted before, the difference between pecan.xml and settings.xml will be confusing. I'd also recomend building the tutorial around one fore the test examples, such that this setting is something like "pecan/tests/pecan.sipnet.xml" or some other sipnet example that we know will run out-of-the-box (e.g. the one associated with the GH Actions integration test)

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+```
+
+---
+
+# Prepare and Validate Settings
+
+PEcAn provides utilities to process and validate settings before execution. This step ensures that required fields are properly set.
+
+```{r prepare-settings}
+settings$outdir <- "/home/carya/pecan_runs"
```
mdietze: This line isn't being explained but is important.  Also, if outdir needs to be updated before prepare.settings this needs to be explained.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+settings <- PEcAn.settings::prepare.settings(settings)
+```
+
+---
+
+# Convert Settings to Required Units
```
mdietze: This text is completely wrong. do.conversions doesn't convert the units of the settings, it builds all the input files and converts the input files (not the setting object) into the model's required *format* (i.e.. it's a format converter, not a units converter)

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+
+PEcAn requires that settings be converted into the correct units before running model simulations.
+
+```{r convert-settings}
+if (!is.list(settings$host) || length(settings$host) > 1) {
```
mdietze: This bit isn't being explained. Conceptually, it belongs in the prepare.settings block (and probably in prepare.settings itself)

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+  settings$host <- list(name = as.character(settings$host)[1])  # Ensure single-element list
+}
+settings <- PEcAn.workflow::do_conversions(settings)
+```
+
+---
+
+# Trait and Meta Analysis
+
+PEcAn retrieves plant trait data and performs meta-analysis to derive parameter distributions for the model.
```
mdietze: still need to explain this is optional and what the alternatives are

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+```
+
+---
+
+# Write Model Configuration Files
+
+Model-specific configuration files are generated before running simulations.
+
+```{r run.write.configs}
+settings$model$binary <- "~/pecan/models/sipnet/"  # Update the path to your model
```
mdietze: Updates to the settings need to be done higher up (e.g. around where you update outdir) and you need to explain what specifically you are doing in a way that a novice user would be able to update. 

Also, this path is very misleading as 1. it points to a folder, not a binary and 2. no one should be installing the model binary inside the model coupler folder (or anywhere else in the PEcAn code itself)

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+```
+
+---
+
+# Run Model Simulations
+
+We execute the model using PEcAn’s workflow module.
+
+```{r run-model}
+if (!is.list(settings$host) || length(settings$host) > 1) {
```
mdietze: didn't you already do this?
remvoe the status.start

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
```diff
@@ -0,0 +1,124 @@
...
+  settings$host <- list(name = as.character(settings$host)[1])
+}
+PEcAn.utils::status.start("MODEL")
+PEcAn.workflow::runModule_start_model_runs(settings)
+```
+
+---
+
+# Model Analyses
```
mdietze: missing the code block that visualizes the output, which should come before the SA and EA. Also, this text should explain that it is run if those bit of the settings are configured (which they are not in a default run) and point the reader to where they would learn about how to configure them. Because this won't run by default it should also be described as optional

web/workflow.Rmd
```diff
@@ -0,0 +1,84 @@
+---
+title: "PEcAn Modular Workflow"
+author: "Aritra Dey"
+date: "`r Sys.Date()`"
+output: html_document
+---
+
+# Load PEcAn packages
+```{r libraries}
```
mdietze: add echo=FALSE

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: yeah this should be in general.Will do the changes.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: added the steps for pecan.xml.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: I have tried to do this.But in some of the step as it is neccesary to do some of the changes in each step i have added the steps.I think it will be better for novice user.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: yes changed the documentation for better understanding of user.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: thanks for pointing out! will update this.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: i have added this step as there is some error in new version R which is @infotroph currently working on.This issue was not there before.I have added explanation for that.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: actually this will point to model path but in setting as the name of the variable is `modelbinary`,on the first look it seems to be a binary,but it should be model path.I have tried to write the doc in that it will be clear.

AritraDey-Dev:
@mdietze could you review this once more? 
I  have made some changes in documentation and like you mention it's better not to keep entire code.I completely agree with that.I only put some small piece of code which could help user for a understandable explanation.Once this is done successfully,i can start working on the gh action for this.

dlebauer:
I think the description may reference the wrong issue (#2784) could you please check?

AritraDey-Dev:
> I think the description may reference the wrong issue (#2784) could you please check?

The discussion regarding this with @mdietze started in the issue #2784,So it's there in the description.I will point the issue to #1866 .Thanks for the suggestion !

AritraDey-Dev:
@mdietze whenever you have a moment please take a look at the changes once.

book_source/02_demos_tutorials_workflows/02_user_demos/06_Workflow_using_Rmd_template.Rmd/workflow.Rmd
AritraDey-Dev: This is just required for now. @infotroph already raised a Pr for this #3492  .in rstudio this is a solution for the issue described in #3492 .

infotroph: 1. In general, don't put temporary workarounds for unrelated issues into a feature PR. In cases like this where there's already a permanent fix proposed, I usually apply the fix for local testing but do not commit it into the feature branch.
2. For this issue specifically, this is the wrong fix as well. The issue in #3492 wasn't the format of the host block, it was how `met.process` handled it internally, and it's expected for settings$host to contain other items besides `name`. Removing those other items will break later parts of the workflow that need to use them.

AritraDey-Dev: Apologies for that. As the issue is solved now, this can be safely removed.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,303 @@
...
+Then, navigate to `web/workflow.Rmd` and open it to run the workflow. Once `pecan.xml` is correctly configured, this approach eliminates the need for executing workflows via the web interface.
+
+
+## Why Use the Modular Workflow?
+
+The current PEcAn workflow (workflow.R) is monolithic, making it less flexible for different user needs. The modular workflow approach breaks it into smaller, independent components, enabling greater adaptability. 
+
+Instead of running the entire workflow at once, users can:
+- Execute specific workflow steps using API endpoints.
+- Run only the necessary components, such as **run.write.configs** or **start.model.runs**.
```
divine7022:
```suggestion
- Run only the necessary components, such as **run.write.configs** or **start_model_runs**.
```

Replaced the deprecated `start.model.runs` with `start_model_runs` as the former is no longer available.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,303 @@
...
+
+# Prepare and Validate Settings
+After reading the settings, a directory named `pecan_runs` will be created in the working directory. This directory will store all output files generated during the PEcAn run.  
+
+PEcAn provides utilities to process and validate settings before execution, ensuring that all required fields are correctly configured.  
+
+- If you are using Docker to set up PEcAn, there is no need to modify the `settings$outdir` path.  
+- If you are running RStudio in a PEcAn virtual machine (VM) or on a local system, update the `settings$outdir` path to the `pecan_runs` directory on your system and place it in your working directory.
+
+```{r prepare-settings,, message=FALSE,echo=FALSE, eval=FALSE}
```
divine7022:
```suggestion
```{r prepare-settings, message=FALSE,echo=FALSE, eval=FALSE}
```
incorrect chunk header syntax

AritraDey-Dev:
Hi @mdietze,
Just checking in on this PR—your review would be helpful to move things along when you have a moment.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+
+The **PEcAn Modular Workflow** is an alternative approach to running PEcAn models directly in R, rather than using the web interface. PEcAn (Predictive Ecosystem Analyzer) automates ecological modeling, allowing researchers to analyze plant functional traits, execute model simulations, and perform sensitivity analyses. 
+
+This modular workflow provides more flexibility by allowing users to run specific steps of the workflow independently, rather than executing the entire process at once. It is particularly useful for advanced users who want more control over individual components of the modeling pipeline.
+
+Instead of relying on the web interface, users can configure their workflow using a `pecan.xml` file and execute it directly in RStudio. This enables reproducibility and efficient debugging of individual workflow steps.
+
+To use the modular workflow, open **RStudio** in your local machine.
+
+Then, navigate to `web/workflow.Rmd` and open it to run the workflow. Once `pecan.xml` is correctly configured, this approach eliminates the need for executing workflows via the web interface.
```
mdietze, line 11: I wonder if there's a better place for this workflow to live, especially as it's not called by any of the web tools. maybe `scripts`?

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+Instead of running the entire workflow at once, users can:
+- Execute specific workflow steps using API endpoints.
+- Run only the necessary components, such as **run.write.configs** or **start_model_runs**.
+- Process meteorology, soils, initial conditions, phenology, and parameter files separately.
+
+This modular approach improves usability, debugging, and automation, making PEcAn workflows more efficient.
+
+## How to Generate a `pecan.xml` File
+
+To obtain a `pecan.xml` file while running a workflow in the PEcAn web interface, follow these steps:
```
mdietze, line 27: This seems like a catch-22. The stated goal is to be able to run PEcan without the web interface, but the only way you're providing to be able to generate a pecan.xml is through the web interface. Rather than pointing folks to the web interface I'd recommend starting from an example that's in the repository (either an existing `tests` or a new pecan.xml) and then also cross-referencing the relevant parts of the documentation.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+
+For a detailed step-by-step tutorial, refer to the official documentation:  
+[PEcAn Workflow Tutorial](https://pecanproject.github.io/pecan-documentation/tutorials/Demo01.html).
+
+
+### Example pecan.xml File
+
+If you follow the steps outlined in the PEcAn Basic Workflow section, you will generate a `pecan.xml` file similar to the example below:
+
+```xml
```
mdietze, line 41: this example would be more useful if it was in a format that could be passed to PEcAn (e.g., have it be in xml, load it into R, then display it in the Rmd)

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+  </host>
+</pecan>
+```
+
+
+# Prerequisites
+Before running this workflow, ensure you have:
+
+- Installed **PEcAn** and its dependencies.
+- An XML settings file (`pecan.xml`) configured for your use case.
```
mdietze: I'd drop the "configured for your use case". The first PR might just run an existing pecan.xml, but you're going to want to pretty quickly have the Rmd show how to reconfigure the settings in code.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+- A model binary (e.g., **SIPNET**) specified in your settings.
+- A pecan repository cloned from GitHub(https://github.com/pecanproject/pecan) in rstudiio.If you are using docker to set up pecan no need to clone pecan repo.
```
mdietze, line 125: How is this different from "Installed PEcAn and its depdendencies"

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+    </rabbitmq>
+  </host>
+</pecan>
+```
+
+
+# Prerequisites
+Before running this workflow, ensure you have:
+
+- Installed **PEcAn** and its dependencies.
```
mdietze: crossreference the instructions for this

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+- An XML settings file (`pecan.xml`) configured for your use case.
+- A model binary (e.g., **SIPNET**) specified in your settings.
+- A pecan repository cloned from GitHub(https://github.com/pecanproject/pecan) in rstudiio.If you are using docker to set up pecan no need to clone pecan repo.
+
+# Load PEcAn Packages
```
mdietze, line 127: From here forward there's a LOT of duplicative content between the Documentation and the Rmd itself. The Rmd needs to be sufficiently self documented that you don't need to have a separate document explaining it, you just need to point to it. Plus, you're setting this up so that the code below has to be run on every PR, since it's part of documentation that needs to be updated.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+
+
+# **Model Analyses in PEcAn**  
+
+Once the model run is complete, PEcAn provides tools to **extract, analyze, and visualize** the results. These analyses help in understanding the model's performance, uncertainty, and sensitivity to input parameters.  
+
+## **Understanding the Code**  
+
+### **1. Extracting Model Results**  
+- `PEcAn.workflow::runModule.get.results(settings)`
```
mdietze, line 276: As noted earlier, this is specific to the Sensitivity and Ensemble analyses, and should not be discussed as a separate step. This code doesn't provide any useful tools for more general analyses of model output files.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
```diff
@@ -0,0 +1,300 @@
...
+  - To enable sensitivity analysis, see [PEcAn Sensitivity Analysis Documentation](https://pecanproject.github.io/pecan-documentation/master/demo-2.html).  
+
+### **3. Ensemble Analysis (Optional)**  
+- `PEcAn.workflow::run.ensemble.analysis(settings)`  
+  - Runs multiple simulations with different parameter sets to estimate model uncertainty.  
+  - This step **requires ensemble settings to be configured** in `pecan.xml`.  
+  - Learn more about ensemble analysis settings [here](https://pecanproject.github.io/pecan-documentation/master/demo-2.html).  
+
+### **4. Visualizing Ensemble Outputs (Optional)**  
+- `PEcAn.workflow::run.ensemble.analysis(settings, plot.timeseries=TRUE)`
```
mdietze, line 292: This bit here is clearly part of the previous bullet point on ensemble analysis. Within the workflow.Rmd, I strongly recommend that you do some more basic visualizations (e.g. timeseries and x-y plots) to replace the interactive visualizations lost through the web interface.

web/workflow.Rmd
```diff
@@ -0,0 +1,78 @@
...
+library("PEcAn.all")
+```
+# Load PEcAn settings files.
+
+Open and read in settings file for PEcAn run.
+To create a pecan.xml, you can download one generated in the PEcAn web interface or one of the `pecan.<modelname>.xml` files in the tests/ directory of 
+the PEcAn repository (github.com/pecanproject/pecan).
+```{r load-settings}
+
+settings_path <- "settings.xml"
```
mdietze: not clear what specific file this is pointing to.

web/workflow.Rmd
```diff
@@ -0,0 +1,78 @@
...
+author: "Aritra Dey"
+date: "`r Sys.Date()`"
+output: html_document
+---
+
+# Load PEcAn packages
+```{r libraries}
+library("PEcAn.all")
+```
+# Load PEcAn settings files.
```
mdietze: throughout this Rmd doc, make sure to pull in the relevant additional bits of information that I'm asking you to take out of the bookdown.

web/workflow.Rmd
```diff
@@ -0,0 +1,78 @@
...
+```{r load-settings}
+
+settings_path <- "settings.xml" 
+settings <- PEcAn.settings::read.settings(settings_path)
+```
+
+
+```{r prepare-settings}
+# Prepare and check settings
+settings$outdir <- "/home/carya/pecan_runs"
```
mdietze: This tag needs to be explained more so users know that they probably need to change it, and what they should change it to. Not all installs will have a /home/carya!

web/workflow.Rmd
```diff
@@ -0,0 +1,78 @@
...
+settings <- PEcAn.settings::prepare.settings(settings)
+```
+
+```{r convert-settings}
+# Convert settings to required units
```
mdietze: I think I've commented before that this comment is incredibly misleading and doesn't explain what this function is actually doing. This is much better explained in the bookdown, and that information needs to be here instead

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
mdietze: That said, such visuals could be in the next PR

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
AritraDey-Dev: I thought it should be added in the `apps/api `folder, but yes, adding it to the `scripts` folder makes more sense

infotroph: I haven't been following the rest of this thread closely, but what was discussed in #1733 and #1866 was putting workflow Rmds into the `PEcAn.workflow` package. I recommend `base/workflows/inst/rmarkdown/templates/`, to align with [where Rstudio looks for templates](https://rstudio.github.io/rstudio-extensions/rmarkdown_templates.html).

They should definitely not go into `apps/api`, which runs server-side and is unrelated to local execution of notebooks.

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
AritraDey-Dev: Yeah, basically the idea is that someone who already has a `pecan.xml` file configured from the web can use it directly, or use the xml file from the tests folder.I will change it accordingly

book_source/02_demos_tutorials_workflows/04_modular_workflow.Rmd
AritraDey-Dev: yes i will remove ` Installed **PEcAn** and its dependencies.` this line.There is no need of this.As if someone a PEcAn repo from github,there is no need to install dependencies separately.Thanks for pointing this out

web/workflow.Rmd
AritraDey-Dev: will update the comments accordingly.

web/workflow.Rmd
AritraDey-Dev: @mdietze Now in the new commits i have incorporated the changes.Please have a look.

mdietze:
@AritraDey-Dev should this PR be pulled in or has it been superseded by the (already merged) Demo 1 PR (and this should be closed)

PR #3492, opened 2025-03-22
met.process: ensure host arg is passed on as a list

infotroph:
## Description

The immediate bug, reported by @AritraDey-Dev in Slack: When passing a whole settings$host block in to `met.process`, R >=4.3 fails because we put a condition with length >1 into a scalar `||` operator. Fixed by refactoring to only look at the element we care about.

Potential issue addressed while I was here: When `host` isn't specified, `met.process` defaults to the bare string "localhost", which is fine in this file but is then passed along to some downstream functions that are documented to expect a named list. Fixed by wrapping bare strings as `host <- list(name = host)` before passing it along.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/met.process.R
+++ b/modules/data.atmosphere/R/met.process.R
@@ -92,7 +92,16 @@ met.process <- function(site, input_met, start_date, end_date, model,
 
   on.exit(PEcAn.DB::db.close(con), add = TRUE)
   username <- ifelse(is.null(input_met$username), "pecan", input_met$username)
-  machine.host <- ifelse(host == "localhost" || host$name == "localhost", PEcAn.remote::fqdn(), host$name)
+
+  if (length(host) == 1 && !hasName(host, "name")) {
+    # We were passed a bare hostname; wrap in a list as downstream fns expect
+    host <- list(name = host)
+  }
+  if (host$name == "localhost") {
+    machine.host <- PEcAn.remote::fqdn()
+  } else {
+    machine.host <- host$name
+  }
   machine <- PEcAn.DB::db.query(paste0("SELECT * from machines where hostname = '", machine.host, "'"), con)
 
 
--- a/modules/data.atmosphere/R/met.process.R
+++ b/modules/data.atmosphere/R/met.process.R
@@ -93,7 +93,7 @@ met.process <- function(site, input_met, start_date, end_date, model,
   on.exit(PEcAn.DB::db.close(con), add = TRUE)
   username <- ifelse(is.null(input_met$username), "pecan", input_met$username)
 
-  if (length(host) == 1 && !hasName(host, "name")) {
+  if (length(host) == 1 && !utils::hasName(host, "name")) {
     # We were passed a bare hostname; wrap in a list as downstream fns expect
     host <- list(name = host)
   }
--- a/docker/depends/pecan_package_dependencies.csv
+++ b/docker/depends/pecan_package_dependencies.csv
@@ -655,6 +655,7 @@
 "utils","*","modules/assim.batch","Imports",FALSE
 "utils","*","modules/assim.sequential","Suggests",FALSE
 "utils","*","modules/benchmark","Imports",FALSE
+"utils","*","modules/data.atmosphere","Imports",FALSE
 "utils","*","modules/data.remote","Suggests",FALSE
 "utils","*","modules/photosynthesis","Imports",FALSE
 "vdiffr",">= 1.0.2","base/qaqc","Suggests",FALSE
--- a/modules/data.atmosphere/DESCRIPTION
+++ b/modules/data.atmosphere/DESCRIPTION
@@ -54,8 +54,8 @@ Imports:
     tidyr,
     tidyselect,
     truncnorm,
-    utils,
     units,
+    utils,
     XML (>= 3.98-1.4),
     xts,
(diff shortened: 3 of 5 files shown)


modules/data.atmosphere/R/met.process.R
```diff
@@ -92,7 +92,16 @@ met.process <- function(site, input_met, start_date, end_date, model,
...
+
+  if (length(host) == 1 && !hasName(host, "name")) {
+    # We were passed a bare hostname; wrap in a list as downstream fns expect
+    host <- list(name = host)
+  }
+  if (host$name == "localhost") {
+    machine.host <- PEcAn.remote::fqdn()
+  } else {
+    machine.host <- host$name
+  }
```
Sweetdevil144, line 104: Sometimes I try wrapping if statements within variable as below :

```r
machine.host <- if (host$name == "localhost") {
  PEcAn.remote::fqdn()
} else {
  host$name
}
```

Nevertheless, the code behavior is still gonna be same as yours :)

Sweetdevil144 (approved):

AritraDey-Dev:
Thanks for looking into this. However, after applying both of the changes suggested by @infotroph and @Sweetdevil144, the issue still persists.

few **Observations** from my side ---
If you check the **Rmd template** I am currently working on (#3486), I encountered the same issue while running `PEcAn.workflow::do_conversions`.  
I have attached some images below to illustrate the error.

Screenshot from 2025-03-23 19-23-37

While executing:
```r
settings <- PEcAn.workflow::do_conversions(settings)
```
I encountered the same issue again. Upon logging `str(settings$host)`, it displayed this output (see the attached image).

Screenshot from 2025-03-23 19-43-28

### Temporary Fix in RStudio
To resolve this in **RStudio**, I applied the following fix:
```r
if (!is.list(settings$host)) {
  settings$host <- list(name = settings$host)
}
```

So I think adding this if statement will fix it.
`if (!is.list(host)) {
  host <- list(name = host)
}`

@infotroph any insights on this will be helpful.

infotroph:
@AritraDey-Dev Thanks for testing! I see that `ic_process` and `soil_process`, both of which can be called by do.conversions, contained copy-pasted versions of the same code, and have now pushed fixes for those too. 

But! Processing of soil and IC ("initial conditions") data don't usually happen in simple workflows -- they require a fair amount of extra configuration. If you don't already know you're using one of those, I would instead check whether your test run really used the patched version of met.process.

infotroph:
Oh, and regarding `is.list`: Which file did you add it to? The `!hasName` on line 96 of `met.process` *ought* to handle the list case, so if it doesn't then we should revisit. (I used hasName instead of is.list to avoid needing to care whether `host` is a list or a named vector when the thing we actually care about is "Will it be an error to write `host$name` later?").

AritraDey-Dev:
> @AritraDey-Dev Thanks for testing! I see that `ic_process` and `soil_process`, both of which can be called by do.conversions, contained copy-pasted versions of the same code, and have now pushed fixes for those too.
> 
> But! Processing of soil and IC ("initial conditions") data don't usually happen in simple workflows -- they require a fair amount of extra configuration. If you don't already know you're using one of those, I would instead check whether your test run really used the patched version of met.process.

As there were multiple instances, I already changed it in both places.  that doesn't work too.
Since you mentioned that a simple workflow doesn’t require it, it also makes sense that changing it in both places didn’t fix the issue either.

AritraDey-Dev:
> Oh, and regarding `is.list`: Which file did you add it to? The `!hasName` on line 96 of `met.process` _ought_ to handle the list case, so if it doesn't then we should revisit. (I used hasName instead of is.list to avoid needing to care whether `host` is a list or a named vector when the thing we actually care about is "Will it be an error to write `host$name` later?").

it didn't fix the issue for running workflow in web...but while working on this PR #3486 it did fix it(for a rmd template).
But i am pretty sure this is causing because of while **PEcAn.workflow::do_conversions**.

modules/data.atmosphere/R/met.process.R
```diff
@@ -92,7 +92,16 @@ met.process <- function(site, input_met, start_date, end_date, model,
 
   on.exit(PEcAn.DB::db.close(con), add = TRUE)
   username <- ifelse(is.null(input_met$username), "pecan", input_met$username)
-  machine.host <- ifelse(host == "localhost" || host$name == "localhost", PEcAn.remote::fqdn(), host$name)
+
+  if (length(host) == 1 && !hasName(host, "name")) {
```
Sweetdevil144:
```suggestion
  if (length(host) == 1 && !("name" %in% names(host))) {
```

Would help bypass the `make check` errors without changing the `NAMESPACE`

modules/data.atmosphere/R/met.process.R
infotroph: Yes, both of these are valid styles, and the Tidyverse style guide doesn't express an opinion here. Personally I tend to find assignment outside the if clearer in a one-liner (`x <- if (y) z else q`) and assignment inside clearer in a multiline `if`... but, well, between PEcAn's long package names and `pkg::` spelling convention, I find that not many `if`s actually fit well on one line! 😉

dlebauer (approved):

PR #3494, opened 2025-03-23
Fix #2970: Update scripts/README.md to reflect current file list

kutumia:
This PR updates the scripts/README.md file to reflect the current contents of the scripts/ directory.

Removed references to outdated/nonexistent scripts (e.g., load.bety.sh)

Added all current scripts grouped by functionality (Build, Workflow, Data, etc.)

Resolves #2970.

PR #3495, opened 2025-03-23
Fix #1820: Automate directory structure documentation via script and Makefile integration

### LINKED ISSUE #1820: Automate documentation of directory structure.
When new directories are made we should make it easy for the documentation of directory structures to be updated by automating the process. 

## Possible Implementation
Create a script to list the main directories in a visually appealing format that gets executed by make when documentation is created.

kutumia:
This PR addresses Issue #1820 by automating the generation of PEcAn's directory structure documentation.

1. Changes Made:
Added a new script: [scripts/generate_dir_structure.sh](https://github.com/PecanProject/pecan/compare/scripts/generate_dir_structure.sh)

2. Recursively lists all folders in the project
Outputs results to documentation/directory_structure.md in a clean Markdown format

3. Updated scripts/README.md
Included description of the new script under Documentation and Dependencies

4. Updated the root Makefile
Introduced a new target: generate-dir-doc
This target runs the script automatically as part of the documentation workflow

How to Use:
To regenerate the directory documentation manually, run in bash: make generate-dir-doc

This will update documentation/directory_structure.md with the latest folder structure.


Makefile
```diff
@@ -225,3 +225,10 @@ $(ALL_PKGS_I) $(ALL_PKGS_C) $(ALL_PKGS_T) $(ALL_PKGS_D): | .install/devtools .in
 .shiny_depends/%: $$(call files_in_dir, %) | $$(@D)
 	Rscript scripts/install_shiny_deps.R $(subst .shiny_depends/,shiny/,$@)
 	echo `date` > $@
+
+# Generates directory structure documentation for the PEcAn project.
+# Writes output to documentation/directory_structure.md
+.PHONY: generate-dir-doc
+
+generate-dir-doc:
+	@bash scripts/generate_dir_structure.sh
```
dlebauer, line 234: @infotroph should this be added to the doc target instead of a separate target?

infotroph:
Thank you, @kutumia, for the work on this, and my apologies for the long wait for a review. I had thought this PR contained useful updates to scripts/README.md that we should consider adopting, but I now see those were already merged as part of #3494.

The core functionality proposed here does indeed implement what #1820 specifies, but speaking for myself I don't need or want the feature that #1820 requested. I'll wait for other team members to weigh in before taking any action, but my sense is that an automatically compiled list of directories will only be useful if it provides some additional explanation/metadata/context that can't be gotten from simply looking at the folders. Since the `directory_structure.md` created here shows nothing but folder names, personally I'd rather look at that in my file browser than in a Markdown document.

infotroph:
Closing with another round of apologies that we didn't close the issue earlier!

PR #3498, opened 2025-03-27
Fix input sampling validation and SIPNET IC handling

blesson-07:
Issue Title
Fix Input Sampling Validation and Deterministic IC Handling

Problem
PEcAn improperly passed multiple input paths (e.g., initial conditions) to model configuration functions without upstream sampling. Models like SIPNET handled this by randomly selecting a path, violating PEcAn’s design where sampling should occur at the ensemble level. This led to:

Non-reproducible workflows (random IC selection).

Lack of validation for unsampled multi-path inputs.

Solution
Upstream Validation:
write.ensemble.configs now errors if an input has multiple paths but no sampling method (<samplingspace> in XML).
Ensures only one path per input is passed to model configs.


blesson-07:
@mdietze   @infotroph 
Could you review the input validation changes?

base/all/tests/testthat/ensemble-test.R
```diff
@@ -0,0 +1,88 @@
+library(testthat)
+library(PEcAn.logger)
+library(PEcAn.DB)
+
+
+source("modules/uncertainty/R/ensemble.R")
```
infotroph: Thank you for including tests along with the PR! 

This file should live in modules/uncertainty rather than base/all on principle (to keep tests together with the unit of code they're testing), but happily also because it makes things simpler: Once moved, you can remove this `source` line and testthat will take care of making sure the package is loaded correctly during testing.

base/all/tests/testthat/ensemble-test.R
```diff
@@ -0,0 +1,88 @@
...
+dummy_binary_path <- file.path(tempdir(), "sipnet")
+file.create(dummy_binary_path)
+# Mock SIPNET writer
+if (!exists("write.config.SIPNET")) {
+  write.config.SIPNET <- function(...) {
+    PEcAn.logger::logger.info("Mock SIPNET writer called")
+    return(invisible(TRUE))
+  }
+}
```
infotroph: If all tests work with this version, could consider formally mocking out write.configs.SIPNET for simplicity

modules/uncertainty/R/ensemble.R
```diff
@@ -213,6 +213,28 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
 write.ensemble.configs <- function(defaults, ensemble.samples, settings, model, 
                                    clean = FALSE, write.to.db = TRUE, restart = NULL, rename = FALSE) {
   
+  
+  # --- START OF YOUR NEW CODE ---
+  # Check if there are NO inputs
```
```suggestion
```

modules/uncertainty/R/ensemble.R
```diff
@@ -213,6 +213,28 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
+    
+    # Case: Multiple inputs + no samples → Error
+    if (length(input_paths) > 1 && is.null(ensemble.samples)) {
+      stop(PEcAn.logger::logger.severe(paste(
+        "Multiple", input_type, "inputs found but no sampling method specified.",
+        "Add a sampling method to pecan.xml (e.g., <method>uniform</method>)"
+      )))
+    }
+  }
+  # --- END OF YOUR NEW CODE ---
```
```suggestion
```

modules/uncertainty/R/ensemble.R
```diff
@@ -213,6 +213,28 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
+    stop(PEcAn.logger::logger.severe("No inputs provided - cannot configure ensemble"))
+  }
+  
+  # Check each input type (e.g., soil, veg)
+  for (input_type in names(defaults$inputs)) {
+    input_paths <- defaults$inputs[[input_type]]$path
+    
+    # Case: Multiple inputs + no samples → Error
+    if (length(input_paths) > 1 && is.null(ensemble.samples)) {
+      stop(PEcAn.logger::logger.severe(paste(
```
infotroph: logger.severe calls stop for you -- no need for the wrapper.
```suggestion
      PEcAn.logger::logger.severe(paste(
```

modules/uncertainty/R/ensemble.R
```diff
@@ -213,6 +213,28 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
+        "Multiple", input_type, "inputs found but no sampling method specified.",
+        "Add a sampling method to pecan.xml (e.g., <method>uniform</method>)"
+      )))
```
```suggestion
        "Add a sampling method to the ensemble$samplingspace$", input_type,
        "section of pecan.xml (e.g., <method>uniform</method>)"
      ))
```

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +422,41 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
       #changing the structure of input tag to what the models are expecting
-      for(input_i in seq_along(settings$run$inputs)){
+      for (input_i in seq_along(settings$run$inputs)) {
         input_tag <- names(settings$run$inputs)[[input_i]]
-        if (!is.null(samples[[input_tag]]))
-          settings$run$inputs[[input_tag]][["path"]] <-
-            samples[[input_tag]][["samples"]][[i]]
+        input <- settings$run$inputs[[input_tag]]
+        
+        # --- Start of changes ---
```
```suggestion
```

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +422,41 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+        if (is.null(input$path) || length(input$path) == 0) {
+          PEcAn.logger::logger.severe("Input '%s' has no paths specified", input_tag)
+        }
+        
+        # Check unsampled inputs first
+        if (!input_tag %in% names(samples)) {
+          if (length(input$path) > 1) {
+            PEcAn.logger::logger.severe(
+              paste("Input '%s' has %d paths but no sampling method.",
+                    "Add <samplingspace> for this input in pecan.xml"),
```
infotroph: I don't immediately see why we need the same check both here and at the top of the function -- Am I missing a reason they'll activate under different circumstances?

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +422,41 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+              input_tag, length(input$path)
+            )
+          }
+          if (!file.exists(input$path[[1]])) {  # New: Verify file exists
+            PEcAn.logger::logger.severe(
+              "Input '%s' path '%s' not found", 
+              input_tag, input$path[[1]]
+            )
+          }
```
infotroph: I'd prefer _not_ to check for file existence here -- there are cases (like local writing of configs that will be copied to a cluster before running) where I want to write configurations that refer to files whose paths are known but do not exist at write time.

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +422,41 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+          next  # Valid single path, no sampling needed
+        }
+        # --- End of changes ---
```
```suggestion
```

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +422,41 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+        
+        # Remaining original code for sampled inputs
+        input_paths <- samples[[input_tag]][["samples"]][[i]]
+        # ... (keep existing validation for sampled inputs) ...
```
infotroph: I'm not sure what this comment means

infotroph (changes requested):
Thank you! Overall this is looking very good. I'll leave some notes inline, but the main change I see needed is to put the test file in modules/uncertainty/tests/testthat rather than base/all/[...]. I also see what looks to me like the same check in two places -- is that a true duplicate or am I not seeing the reason both are needed?

base/all/tests/testthat/ensemble-test.R
mdietze: I was thinking that too, but write.config.SIPNET doesn't actually return anything AFAIK. Is there a way to set up a test to see if the mocked write.config.SIPNET received the inputs that it was supposed to (which in this case is that it never receives vector input paths)?

modules/uncertainty/R/ensemble.R
mdietze: I could see keeping the second line, but agree the first has to go

modules/uncertainty/R/ensemble.R
```diff
@@ -213,6 +213,28 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
 write.ensemble.configs <- function(defaults, ensemble.samples, settings, model, 
                                    clean = FALSE, write.to.db = TRUE, restart = NULL, rename = FALSE) {
   
+  
+  # --- START OF YOUR NEW CODE ---
+  # Check if there are NO inputs
+  if (is.null(defaults$inputs)) {
+    stop(PEcAn.logger::logger.severe("No inputs provided - cannot configure ensemble"))
```
mdietze: I don't think logger.severe should to be called from within stop. That might lead to unintended behaviors.

modules/uncertainty/R/ensemble.R
mdietze: I also think its worth considering making these logger.error instead of logger.severe. That way users could more easily toggle if they want to workflow to stop here or not. PEcAn is designed to do large ensembles of runs (e.g. Dongchen is currently doing 640000 run ensembles) and you don't necessarily want the entire workflow to come to a halt if there's something off in one of those 640000 runs, which you might not encounter until you're a week into the compute.

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +422,41 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+            PEcAn.logger::logger.severe(
+              "Input '%s' path '%s' not found", 
+              input_tag, input$path[[1]]
+            )
+          }
+          next  # Valid single path, no sampling needed
+        }
+        # --- End of changes ---
+        
+        # Remaining original code for sampled inputs
```
mdietze: comment is uninformative. Reader won't know what "original code" refers to and instead will just want to know what this step is doing.

base/all/tests/testthat/ensemble-test.R
infotroph: That sounds like a use case for `mockery::expect_call()`!

modules/uncertainty/R/ensemble.R
infotroph: Answer: I was missing that the top of the function checks `defaults` and this line checks `inputs`. Yes, just slightly different circumstances 🙈

infotroph:
@blesson07asd Will you be able to return to this soon? Seems like it's in pretty good shape overall.

The biggest issue I see remaining is that `ensemble-test.R` is now removed entirely -- thanks for taking it out of base/all as requested, but can you please add it back in modules/uncertainty?

blesson-07:
@infotroph    Apologies for the delay, I was in the middle of my semester exams, which are now over. I’ll resume work on this and get the remaining changes pushed ASAP. Thanks for your patience and support!

modules/uncertainty/R/ensemble.R
```diff
@@ -215,22 +215,24 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
-        "Multiple", input_type, "inputs found but no sampling method specified.",
-        "Add a sampling method to pecan.xml (e.g., <method>uniform</method>)"
-      ))
-    }
+  # Check for unsampled multi-path inputs
+  if (length(input_paths) > 1 && 
+     !(input_tag %in% names(settings$ensemble$samplingspace))) {
+    PEcAn.logger::logger.error(
+      "Input '%s' has %d paths but no sampling method. Add <samplingspace> for this input in pecan.xml",
+      input_tag, length(input_paths))
```
```suggestion
      "Input", sQuote(input_tag), "has", length(input_paths), "paths but no sampling method.",
      "Add <samplingspace> for this input in pecan.xml")
```

modules/uncertainty/R/ensemble.R
```diff
@@ -215,22 +215,24 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
-  if (is.null(defaults$inputs)) {
-    PEcAn.logger::logger.severe("No inputs provided - cannot configure ensemble")
+ 
+for (input_tag in names(settings$run$inputs)) {
+  input <- settings$run$inputs[[input_tag]]
+  input_paths <- input$path
+  
+  # Check for required paths
+  if (is.null(input_paths) || length(input_paths) == 0) {
+    PEcAn.logger::logger.error("Input '%s' has no paths specified", input_tag)
```
infotroph: Looks like you're trying to use sprintf style here, but logger.error isn't that smart -- it just pastes all its arguments together.
```suggestion
    PEcAn.logger::logger.error("Input", sQuote(input_tag), "has no paths specified")
```

modules/uncertainty/R/ensemble.R
```diff
@@ -400,14 +423,22 @@ write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
...
+  
+  if (!input_tag %in% names(samples)) {
+    # Use first path (already validated as single path)
+    settings$run$inputs[[input_tag]]$path <- input$path[1]
+  } else {
+    # Use sampled path
+    settings$run$inputs[[input_tag]]$path <- samples[[input_tag]][["samples"]][[i]]
+  }
 
+}
```
infotroph: I think this loop should be indented to match the lines around it

modules/uncertainty/tests/testthat/test_ensemble.R
```diff
@@ -0,0 +1,105 @@
+context("Ensemble Configuration Tests")
+
+# Mock the core functions we need to test
+write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,
```
infotroph: This won't do what you want. Defining a new write.ensemble.configs here makes it so all the tests below use this function and leave the real run.write.configs untested. If you say more about your goal here I can likely suggest approaches.

I remember thinking that the test file you had in an earlier version (the one that was in base/all) looked good other than being in the wrong place -- would it be easier to bring that back but at this path?

blesson-07:
Sorry for the confusion in the PR — my branch was based on an outdated version of develop, and rebasing to update it unintentionally pulled in unrelated commits.

modules/uncertainty/R/ensemble.R
blesson-07: Thanks for pointing that out! Could you clarify exactly what I should adjust?

infotroph (approved):

modules/uncertainty/tests/testthat/test_ensemble.R
```diff
@@ -0,0 +1,138 @@
+library(testthat)
+library(mockery)
+
+# Source the ensemble config function
+source("modules/uncertainty/R/ensemble.R")
```
mdietze:
```suggestion
```

modules/data.remote/DESCRIPTION
```diff
@@ -44,6 +44,7 @@ Suggests:
     lubridate,
     raster,
     reshape,
+    rhdf5,
```
mdietze:
```suggestion
```

docker/depends/pecan_package_dependencies.csv
```diff
@@ -471,6 +471,7 @@
 "reticulate","*","modules/data.atmosphere","Suggests",FALSE
 "reticulate","*","modules/data.land","Suggests",FALSE
 "reticulate","*","modules/data.remote","Imports",FALSE
+"rhdf5","*","modules/data.remote","Suggests",FALSE
```
mdietze:
```suggestion
```

mdietze (commented):
Not sure why you added rhdf5 to the libraries in data.remote as you don't seem to use it anywhere in your code

modules/uncertainty/tests/testthat/test_ensemble.R
```diff
@@ -0,0 +1,137 @@
+library(testthat)
+library(mockery)
```
mdietze: 1. Both these library calls need to be removed from the test file
2. mockery needs to be added to the DESCRIPTION file under Suggests
3. after that you need to run the dependencies script and add the updated docker/depends/pecan_package_dependencies.csv

PR #3499, opened 2025-03-27
[rtm] avoid hangs during vignette build

### LINKED ISSUE #3491: CI checks of RTM time out
### Bug Description

The CI `check` job is intermittently timing out (6 hours!), always while checking `base/rtm`. I have not yet been able to reproduce it locally.

Up until yesterday I had only seen it with R 4.1 and was holding out hope that a version upgrade would resolve it for us, but here it is timing out under R 4.4: https://github.com/PecanProject/pecan/actions/runs/13943588816/job/39025557837?pr=3490

Help wanted, but here potentially be dragons: It has the potential to involve any of (or maybe even all three of!) the GitHub Actions machinery, R package check behavior, and compiled Fortran code. Whoooo wants a challenge?

infotroph:
## Description

Fixes #3491 + some housekeeping while I was in the file.

- fixed spelling of options `do.lsq` and `ngibbs.max`, removed others that have no effect
- fixed some typos in messages
- added :: throughout invert.custom and invert.auto
- bring imports up to date + sort for easier reading

## Motivation and Context

Best I can tell the main issue causing #3491  was the typo setting `do.lsq.first` instead of `do.lsq`, presumably meaning the least-squares initialization was being skipped and the MC parameters often started out far from reasonable values. 

Setting `ngibbs.max` is basically a backup timeout precaution -- if convergence isn't reached by then, vignette building will fail (with an obtuse error about "wrong number of parameters", but refining that can be a separate task). The limit of 50k iterations can be changed -- I choose it to be comfortably higher than the iterations I saw needed in local testing, which were typically in the 10-20k range and I never saw it go higher than 30k.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/rtm/vignettes/pecanrtm.vignette.Rmd
+++ b/modules/rtm/vignettes/pecanrtm.vignette.Rmd
@@ -118,11 +118,9 @@ invert.options$model
 The recommended way to set settings is to start with a default object and modify it. For example, in the following block, we reduce the length of the run to make it go a little faster. `threshold = ...` sets the maximum value of the multivariate Gelman Diagnostic used to assess convergence -- it defaults to 1.1, but we set it higher here for demonstrative purposes).
 
 ```{r}
-invert.options$n.tries <- 1      # Number of attempts
 invert.options$nchains <- 2      # Number of MCMC chains
-invert.options$ngibbs <- 5000    # Number of iterations per chain
-invert.options$burnin <- 1000    # Length of burnin period
-invert.options$do.lsq.first <- TRUE # Initialize with results from a fast least-squares optimization algorithm
+invert.options$ngibbs.max <- 5e4 # Maximum number of iterations per chain (fails if no convergence by then)
+invert.options$do.lsq <- TRUE    # Initialize with results from a fast least-squares optimization algorithm
 invert.options$threshold <- 1.3  # Maximum value for Gelman diagnostic
 ```
 
--- a/modules/rtm/vignettes/pecanrtm.vignette.Rmd
+++ b/modules/rtm/vignettes/pecanrtm.vignette.Rmd
@@ -8,7 +8,7 @@ vignette: >
 ---
 
 ```{r, setup, include=FALSE}
-knitr::opts_chunk$set(fig.width = 6, fig.height = 4.5, out.width = '100%', dpi=60)
+knitr::opts_chunk$set(fig.width = 6, fig.height = 4.5, out.width = '80%', dpi=72, fig.retina = 1)
 ```
 
 # Introduction
--- a/docker/depends/pecan_package_dependencies.csv
+++ b/docker/depends/pecan_package_dependencies.csv
@@ -165,7 +165,7 @@
 "lubridate",">= 1.6.0","modules/assim.sequential","Imports",FALSE
 "lubridate",">= 1.6.0","modules/benchmark","Imports",FALSE
 "lubridate",">= 1.6.0","modules/data.atmosphere","Imports",FALSE
-"lubridate",">= 1.6.0","modules/rtm","Imports",FALSE
+"lubridate",">= 1.6.0","modules/rtm","Suggests",FALSE
 "lubridate",">= 1.7.0","models/biocro","Imports",FALSE
 "Maeswrap","*","models/maespa","Suggests",FALSE
 "magic",">= 1.5.0","modules/assim.sequential","Suggests",FALSE
@@ -565,6 +565,7 @@
 "stats","*","modules/assim.batch","Imports",FALSE
 "stats","*","modules/assim.sequential","Suggests",FALSE
 "stats","*","modules/photosynthesis","Imports",FALSE
+"stats","*","modules/rtm","Imports",FALSE
 "SticsRFiles","*","models/stics","Suggests",FALSE
 "stringi","*","base/logger","Imports",FALSE
 "stringi","*","base/utils","Imports",FALSE
--- a/modules/rtm/.Rbuildignore
+++ b/modules/rtm/.Rbuildignore
@@ -2,3 +2,4 @@
 ^\.Rproj\.user$
 ^dev
 ^tests
+vignettes/.*.rds
--- a/modules/rtm/DESCRIPTION
+++ b/modules/rtm/DESCRIPTION
@@ -15,22 +15,23 @@ Description: Functions for performing forward runs and inversions of radiative
     Underlying numerical analyses are optimized for speed using Fortran code.
 Depends: R (>= 2.10)
 Imports:
-    PEcAn.logger,
-    MASS,
+    BayesianTools,
     coda,
-    lubridate (>= 1.6.0),
+    MASS,
     PEcAn.assim.batch,
-    BayesianTools
+    PEcAn.logger,
+    stats
 Suggests:
-    minpack.lm,
+    knitr (>= 1.42),
+    lubridate (>= 1.6.0),
     mclust,
-    PEcAn.utils,
+    minpack.lm,
     PEcAn.ED2,
-    XML,
+    PEcAn.utils,
+    pwr,
     rmarkdown,
     testthat (>= 1.0.2),
-    knitr (>= 1.42),
-    pwr
+    XML
 OS_type: unix
 License: BSD_3_clause + file LICENSE
 Copyright: Authors
--- a/modules/rtm/R/invert.auto.R
+++ b/modules/rtm/R/invert.auto.R
@@ -37,7 +37,7 @@
 #' inputs list containing runID, initial values, and resume (NULL) as an
 #' argument.
 #'
-#' * `calculate.burnin` -- If `TRUE`, use `PEcAn.assim.batch::autoburin`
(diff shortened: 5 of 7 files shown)


infotroph:
attn @AritraDey-Dev -- not sure why it's not letting me tag you as a reviewer, but please have a look.

modules/rtm/vignettes/pecanrtm.vignette.Rmd
```diff
@@ -118,11 +118,9 @@ invert.options$model
 The recommended way to set settings is to start with a default object and modify it. For example, in the following block, we reduce the length of the run to make it go a little faster. `threshold = ...` sets the maximum value of the multivariate Gelman Diagnostic used to assess convergence -- it defaults to 1.1, but we set it higher here for demonstrative purposes).
 
 ```{r}
-invert.options$n.tries <- 1      # Number of attempts
 invert.options$nchains <- 2      # Number of MCMC chains
-invert.options$ngibbs <- 5000    # Number of iterations per chain
-invert.options$burnin <- 1000    # Length of burnin period
-invert.options$do.lsq.first <- TRUE # Initialize with results from a fast least-squares optimization algorithm
+invert.options$ngibbs.max <- 5e4 # Maximum number of iterations per chain (fails if no convergence by then)
+invert.options$do.lsq <- TRUE    # Initialize with results from a fast least-squares optimization algorithm
```
infotroph, line 123: Here's the real fix! All the changes to other files are just cleanup.

AritraDey-Dev: Increasing ngibbs (the number of iterations per MCMC chain) can improve the accuracy and convergence of the Bayesian inversion,
Also I have tested it on my side by manually running the workflows, and everything seems to be working fine. One important observation is that `check modules/rtm` now completes in around 4 minutes, whereas previously, it took more than 5 minutes.  
I request you not to merge the PR immediately, as I want to run some of the CI checks manually again as in one workflow i found some issue about timeout.Once that is successfully  done, we can proceed.

AritraDey-Dev (commented):
@infotroph  any specific reason behind removing `invert.options$n.tries <- 1` ?

modules/rtm/vignettes/pecanrtm.vignette.Rmd
AritraDey-Dev, line 1: any specific reason behind removing `invert.options$n.tries <- 1`?

infotroph:
I removed it because I can't find anywhere in the code that it would have an effect! If I missed one, we can and should add it back.

AritraDey-Dev:
Also, the CI check is failing for 4.4. From the logs, the issue seems to be related to adding `vignettes/.*.rds` in .Rbuildignore.

AritraDey-Dev:
@infotroph I have tested your code, but the CI checks for ` modules/rtm` are still occasionally timing out.

AritraDey-Dev:
@infotroph if you look at the line [here](https://github.com/PecanProject/pecan/blob/develop/modules/rtm/vignettes/pecanrtm.vignette.Rmd#L143
) 
The comment about `parallel=TRUE` refers to parallel execution for each chain. While this may work locally, but i think  `gh action`cannot do parallel execution for` invert.auto` for each chains.

Screenshot from 2025-03-28 23-39-43
I think The error occurs because the script attempts parallel execution in a restricted **gh action** environment.

infotroph:
* Failure on 4.4 is because package size increased marginally (and was close to `R CMD check`'s 5 MB limit already), not because of the .Rbuildignore change (which was me hoping I could solve the size issue by jettisoning cache files, but looks like those weren't what was causing it). I think fixing it will require somme tedious fiddling to reduce vignette filesize without making the images unreadable -- I'll come back to it when I have a chance; suggestions welcome in the meantime.
* Can you give more detail on your setup for the checks that are timing out? If they were on GHA, can you link to the failing runs?
* Yeah, leave `parallel = FALSE` in the executed code. The line you linked to is great advice for human readers doing local runs for their own use (i.e. the audience this vignette was written for!), but on CI infrastructure the list of ways parallel execution can go wrong is longer than the list of times it will help.

AritraDey-Dev:
- The [workflow](https://github.com/AritraDey-Dev/pecan/actions/runs/14115759956/job/39545465291) caused a timeout.      However, setting `ngibbs.max` appears to be a notable factor, as it is not timing out as frequently as before.
   Previously, I ran multiple workflows together (4-5 in one go), and only one had an issue.
  I believe setting `ngibbs.max` to **1e4** should work fine.

- In the data/ folder of the tar file, sensor.rsr.Rda is the largest file.

Screenshot from 2025-03-29 14-59-56

I guess reducing the size of this data or compressing  may help.

I also tried preventing the last block of code in vignette from being evaluated by setting `eval=FALSE,`  i.e.,

```
```{r eval=FALSE}
par(mfrow=c(2,2))
plot(inversion.output$samples, auto.layout=FALSE)

par(mfrow=c(1,1))
samples.mat <- as.matrix(inversion.output$samples)[-(2000:0),1:5]
colnames(samples.mat) <- params.prospect5
pairs(samples.mat, pch=".")

means <- unlist(inversion.output$results[grep("mu", names(inversion.output$results))])[1:5]
prospect.sim <- prospect(means, 5)[,1]  # reflectance

plot(wl, observed, type='l', col=1, xlab="wavelength (nm)", ylab="reflectance")
lines(wl, prospect.sim, type='l', col=2)
legend("topright", c("observed", "predicted"), lty=1, col=1:2)
```



With this change, the CI checks for version 4.4 were successful.

AritraDey-Dev:
@infotroph one interesting thing -- setting `do.lsq` true is what causing issue in 4.4.I have not changed that part of code (do.lsq) and only changed `ngibbs.max` as a lower value and run gha.In that case ci checks for 4.4 are working fine.

infotroph:
@AritraDey-Dev I'm having trouble tracking which of your comments are evaluating _this PR's proposed changes_ and which are you considering  _alternate fixes_. Let's try to focus this thread mostly on the former, and only move on to discussing alternates if the proposed patch isn't sufficient.

Which is not to say don't share other ideas -- "This works but I see a simpler way" is valid feedback! I just don't want confusion between problems with the proposed solution and problems with alternate solutions that are not problems with the proposed one.

AritraDey-Dev:
> @AritraDey-Dev I'm having trouble tracking which of your comments are evaluating _this PR's proposed changes_ and which are you considering _alternate fixes_. Let's try to focus this thread mostly on the former, and only move on to discussing alternates if the proposed patch isn't sufficient.
> 
> Which is not to say don't share other ideas -- "This works but I see a simpler way" is valid feedback! I just don't want confusion between problems with the proposed solution and problems with alternate solutions that are not problems with the proposed one.

**@infotroph** Apologies for any confusion. What I meant is that the first part of this code block can be safely dropped:

https://github.com/PecanProject/pecan/blob/develop/modules/rtm/vignettes/pecanrtm.vignette.Rmd#L159-L174

```r
par(mfrow=c(2,2))
plot(inversion.output$samples, auto.layout=FALSE)

par(mfrow=c(1,1))
samples.mat <- as.matrix(inversion.output$samples)[-(2000:0),1:5]
colnames(samples.mat) <- params.prospect5
pairs(samples.mat, pch=".")
```
  
This part is just for MCMC diagnostic plots, which visualize the sampling process but don’t modify any variables used later. Since the mean calculations start fresh from `inversion.output$results`, removing these diagnostics won’t affect the final model comparison.  

You can either remove this section or set `eval=FALSE` to prevent execution.  

However, the second part **must be kept** because it generates the actual model comparison plot:  

```r
means <- unlist(inversion.output$results[grep("mu", names(inversion.output$results))])[1:5]
prospect.sim <- prospect(means, 5)[,1]  # reflectance

plot(wl, observed, type='l', col=1, xlab="wavelength (nm)", ylab="reflectance")
lines(wl, prospect.sim, type='l', col=2)
legend("topright", c("observed", "predicted"), lty=1, col=1:2)
```

This part is essential as it compares the fitted model output to observed data, confirming that the inversion worked correctly.


- The motive behind this is to reduce the data folder size. The changes you made in this [commit](https://github.com/PecanProject/pecan/pull/3499/commits/8d8b858ccbe9f428df245c5dc5ed99b59508677e) also help decrease the size of the doc folder.

AritraDey-Dev:
@infotroph Your changes to `fig.retina=2` and the adjustment of `dpi` are no longer causing any timeouts. Since these modifications are working as expected, there is no need to make change on the previously mentioned changes.

AritraDey-Dev (approved):

infotroph:
Thanks for checking this! I'll merge now.

(It's true that skipping the diagnostic plots would reduce vignette size further, but they're pedagogically important -- I'm already on team "always check your chain plots", but a demo where we took shortcuts to reduce MCMC run time is one where they're extra-useful for noticing quality issues!)

AritraDey-Dev:
That makes sense—especially in cases where we take shortcuts, having clear diagnostic plots helps ensure we’re not missing any issues. Definitely worth keeping them in!

dlebauer (approved):

PR #3500, opened 2025-03-29
Fix `Deprecated: htmlspecialchars()` for null warning

Sweetdevil144:
## Description

This PR fixes a small error : `Deprecated: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/pecan/08-finished.php on line 45` which arises when we try to open a pre-existing workflow run from "http://pecan.localhost/bety". For example, if I chose to open Results for the workflow run of `/home/carya/output/PEcAn_5000000241`, we got the above error because PEcAn expected `/home/carya/output/PEcAn_5000000241` directory to exist. If the data directory isn't present within the user's system, taking me as an example, we would have encountered the above error. 

Another Issue I guess we can actually raise here is : _"Should the actual path of data outputs be `/home/carya/output/<x_path>` or should it have been like the current data directory we use, for eg: `output folder = /data/workflows/<x_workflow_executed>` ?"_ .

**Note that this error only arises when I tried opening a pre-existing workflow by @dlebauer  to run from BETYdb (while I tried experimenting with the database)**.

Currently in BETYdb, I can see the folder path is `/home/dlebauer/.pecan/biocrotest`.

## Review Time Estimate
- [x] When possible

## Types of changes
## Checklist:

### DIFF
--- a/web/08-finished.php
+++ b/web/08-finished.php
@@ -42,7 +42,7 @@ $stmt->closeCursor();
 $start = substr($workflow['start_date'], 0, 4);
 $end = substr($workflow['end_date'], 0, 4);
 $folder = $workflow['folder'];
-$notes = htmlspecialchars($workflow['notes']);
+$notes = htmlspecialchars($workflow['notes'] ?? '');
 if ($workflow['value'] != '') {
   $params = json_decode($workflow['value'], true);
 } else {


infotroph (approved):
Thanks -- TIL that PHP has a null-coalescing operator!

This patch is a nice immediate improvement. It does look to me like `$workflow['notes]` will only be null when the query returned zero results, so a nice follow-up change would be to check for zero results immediately after executing the query and handle the no-result case there rather than going through checking for nulls as we examine each field.

The question about how to handle nonexistent paths is a good one, but not related to this particular warning -- let's move it to another thread.

infotroph:
Correction: I was too hasty in my diagnosis here, but @Sweetdevil144's fix is still good. 

Apparently the notes field _can_ be null even when the rest of the record exists (I had convinced myself it'd be an empty string in this case). Maybe someone who knows their PHP better will speak up, but if handling nulls is as simple as `?? ''`, then handling them one field at a time isn't too bad.

PR #3506, opened 2025-04-03
Fix "external pointer is not valid" error and improve output handling in soilgrids_soilC_extract()

divine7022:
## Description
This PR introduces multiple reliability improvements to the `soilgrids_soilC_extract()` function, focusing on handling edge cases, preventing errors, and ensuring robust execution across diverse geographic locations. These PR address issues, such as failures caused by **`external pointer is not valid`** issue and dimension errors during gamma fitting.

### Fixes and Improvements:
1. Fixed "external pointer is not valid" error:
  - Extracted coordinates `(terra::crds())` for safe transfer in parallel execution.
  - Recreated SpatVector objects `(terra::vect())` inside parallel workers to prevent invalid external pointer errors when working across processes.
  - Added fallback to sequential processing when parallel execution fails.
 2. Addressed Gamma Fitting Failures:
  - Fixed cases where missing or insufficient data caused dimension errors in `bestPar[1, ]`.
  - Added explicit validation to prevent gamma fitting when data contains all NA values.
  - Ensured gamma fitting handles edge cases (e.g., sites with insufficient quantiles).
 3. Critical Validation Checks:
  - Early return implemented when no valid sites remain after preprocessing:
```
        if (nrow(internal_site_info) == 0) {
          warning("No valid sites with extractable data.")
          return(NULL)
        }
```
  - Skipped gamma fitting when less than two quantiles are available for a site:
 ```
       if (length(unique(ocd_df$Quantile)) < 2) {
              warning("Insufficient quantiles for gamma fitting.")
              return(NULL)
            }
```
 4. Added directory creation with `recursive = TRUE` to automatically create full directory paths when needed.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### CODE REVIEW

**Comment by divine7022** [comment]
Hello sir,
@dlebauer @infotroph 
Could you please review this PR when you have a moment, containing important error fixes and enhancements for the SoilGrids data extraction.

**Review by dlebauer — CHANGES_REQUESTED** [review]
Looks good to me - this should make the function more robust. 

Could you please update the package NEWS and CONTRIBUTING? Then it will be ready to merge.

**Comment by divine7022** [comment]
@dlebauer thanks for ur feedback! addressed the suggestion

File: modules/data.land/R/soilgrids_soc_extraction.R [hunk 1]
```diff
@@ -66,6 +66,18 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
...
   internal_site_info <- site_info[, c("site_id", "site_name", "lat", "lon")]
+
+  # Early return if no valid sites (after processing internal_site_info)
+  if (nrow(internal_site_info) == 0) {
+    if (verbose) {
+      PEcAn.logger::logger.severe(
+        "No valid sites remaining after NA check. ",
+        "All sites had missing SoilGrids data for the first depth layer."
+      )
+    }
```
infotroph, line 77: 1. Message seems out of place -- no NA check has happened yet at this point. for nrows to be zero it has to be passed that way by the user, so it could definitely happen for any number of reasons that aren't the one the message says. 
2. anything bad enough to warrant logger.severe (which will stop execution of the entire workflow) should not be suppressible by the verbose flag.

```suggestion
    if (verbose) {
      PEcAn.logger::logger.info("No sites specified")
    }
```

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -116,6 +138,19 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
   ocd_df$Value<-as.numeric(ocd_df$Value)
   f1<-factor(ocd_df$Siteid,levels=unique(ocd_df$Siteid))
   f2<-factor(ocd_df$Depth,levels=unique(ocd_df$Depth))
+
+  # Skip if not enough quantiles (before gamma fitting)
+  if (length(unique(ocd_df$Quantile)) < 2) {
```
infotroph, line 143: Did you find cases where this was actually happening? Unless I'm misreading this should be the 4 quantiles defined on line 131, always.

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -78,17 +90,27 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
...
   p_reproj <- terra::project(p, newcrs) # Transform the point vector to data with Homolosine projection
+
+  # Extract coordinates for safe parallel transfer
+  p_coords <- terra::crds(p_reproj)
+
   data_tag <- c("_mean.vrt", "_Q0.05.vrt", "_Q0.5.vrt", "_Q0.95.vrt")
   name_tag <- expand.grid(depths, data_tag, stringsAsFactors = F)#find the combinations between data and depth tags.
   L <- split(as.data.frame(name_tag), seq(nrow(as.data.frame(name_tag))))#convert tags into lists.
 
   get_layer <- function(l) {
```
infotroph, line 101: I know this is the existing design and not your change, but for the record defining a function inside a function like this has the result that all the local variables inside the enclosing function get packaged up into the environment part of the inner function's closure.

Which means p_reproj is still being *sent* to all the workers (along with any other large objects in the function!) even if we don't reference it in the code.

A good future refactor would be to define `get_layer` outside of `soilgrids_soilC_extract`, with explicit arguments for all the variables it uses. No promises it would make a detectable speed difference, but a definite win for clarity of encapsulation.

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -184,11 +226,17 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
...
-    utils::write.csv(soilgrids_soilC_data,file=file.path(outdir,"soilgrids_soilC_data.csv"),row.names = FALSE)
-  }
-  else {
-    PEcAn.logger::logger.error("No output directory found.")
+    # Ensure the directory exists; create if not
+    if (!dir.exists(outdir)) {
+      dir.create(outdir, recursive = TRUE)
+      PEcAn.logger::logger.info(paste0("Created output directory: ", outdir))
+    }
+    PEcAn.logger::logger.info(paste0("Storing results in: ", file.path(outdir, "soilgrids_soilC_data.csv")))
```
infotroph, line 234: - Logger functions do the pasting for you
- No need to separately call out creating the directory -- reporting it in the result path is enough
```suggestion
      dir.create(outdir, recursive = TRUE)
    }
    PEcAn.logger::logger.info("Storing results in: ", file.path(outdir, "soilgrids_soilC_data.csv"))
```

dlebauer (approved):

modules/data.land/R/soilgrids_soc_extraction.R
```diff
@@ -132,22 +167,29 @@ soilgrids_soilC_extract <- function (site_info, outdir=NULL, verbose=TRUE) {
...
+    }
+    theta <- c(10, 10)
+    fit <- tryCatch(
+      stats::optim(theta, cgamma, val = val, stat = stat),
+      error = function(e) NULL
+    )
+    if (is.null(fit)) {
+      return(list(par = c(NA, NA), SS = NA))
+    }
+    return(list(par = fit$par, SS = fit$value))
```
infotroph, line 184: What are the possible lengths of x here? If it's always 1 I'm confused why line 173 is needed, if it can be more than 1 then I don't think list(par = par$fit,...) and list(par = sapply(fit, \(f)x$par), ...) are equivalent

infotroph (changes requested):

PR #3508, opened 2025-04-11
Add IC_SOILGRID_Utilities for SoilGrids Data Processing

divine7022:
## Description
This pull request introduces a new utility script, IC_SOILGRID_Utilities.R, designed to facilitate the processing of SoilGrids data for generating soil carbon initial condition (IC) files. The new script provides essential functions for extracting, processing, and generating ensemble members from SoilGrids250m data. Key changes include the implementation of general **`get.site.info`** function, which extracts site information from various input formats.
#### 1) Implemented general and reusable `get.site.info` function to standardize site information handling:
- Extracts data from Settings, MultiSettings, or flexible enough with CSV inputs too.
- Validates required fields (site_id, lat, lon) and data types.
- Supports both single sites and vectorized site data.
- Includes coordinate validation with strict/lenient modes.
- Returns consistent data frame output (site_id, name, lat, lon, str_id).
- Testing:
  - Developed a comprehensive test suite for the `get.site.info` function, covering various scenarios such as settings objects, CSV files, MultiSettings objects, and vectorized site information.
  - Ensured that all tests pass successfully, validating the correctness of the new functionality.
#### 2) `soilgrids_ic_process`:
- A comprehensive function that processes SoilGrids data to create initial condition files. It extracts soil carbon data, handles missing values, generates ensemble members, and writes output to NetCDF files (End-to-end workflow from input to NetCDF).
- Supports input from both PEcAn settings lists and optional CSV files for site information.
- Includes parameters for output directory management, file overwriting, and verbosity for detailed logging.
#### 3) `preprocess_soilgrids_data`:
- A helper function that preprocesses the raw soil carbon data, handling missing values and ensuring data integrity.
- Implements logic for scaling and defaulting values where necessary.
#### 4) `generate_soilgrids_ensemble`:
- A function that generates ensemble members for a given site based on processed soil carbon data, ensuring reproducibility through random seed setting.
- Uses truncated normal distribution (no negative values).
- Seed set by site_id for deterministic results.
- Configurable ensemble size.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


divine7022:
Hi @dlebauer @infotroph 
Could you please review this PR when you have a moment?

Thank you!

divine7022:
I'll add tests for `IC_SOILGRID_Utilities` after addressing your feedback. Currently, this PR only includes tests for `get.site.info`.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
+#' SoilGrids Initial Conditions (IC) Utilities
+#' 
+#' @author Akash
+#' @description Functions for generating soil carbon IC files from SoilGrids250m data
+#' @details This module provides functions for extracting, processing, and generating
+#'          ensemble members for soil carbon initial conditions using SoilGrids data.
+#'          All soil carbon values are in kg/m².
+
+# Required package
+library(truncnorm)
```
mdietze: Remove library call, reference functions by namespace

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+            processed$`Total_soilC_0-30cm`[i], processed$Site_ID[i]
+          ))
+        }
+      }
+    }
+    
+    # Sites still with missing data - use default value
+    still_missing <- is.na(processed$`Total_soilC_0-30cm`)
+    if (any(still_missing)) {
+      processed$`Total_soilC_0-30cm`[still_missing] <- default_soilC
```
mdietze: Missing data should  not be replaced with a default value

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+  
+  # Handle missing values in Std_soilC_0-30cm
+  na_count <- sum(is.na(processed$`Std_soilC_0-30cm`))
+  if (na_count > 0) {
+    PEcAn.logger::logger.warn(sprintf("Found %d missing values in soil carbon uncertainty", na_count))
+    
+    # Sites with missing 0-30cm but available 0-200cm uncertainty data
+    has_200cm_data <- is.na(processed$`Std_soilC_0-30cm`) & !is.na(processed$`Std_soilC_0-200cm`)
+    if (any(has_200cm_data)) {
+      processed$`Std_soilC_0-30cm`[has_200cm_data] <- processed$`Std_soilC_0-200cm`[has_200cm_data] * 0.15
```
mdietze: First, it's not clear this is a valid substitution. Second, it's not clear, either from the units or from basic knowledge of soil carbon depth distributions, that you can multiply the SD by 0.15.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+          ))
+        }
+      }
+    }
+    
+    # Sites still with missing uncertainty - use default percentage of mean
+    still_missing <- is.na(processed$`Std_soilC_0-30cm`)
+    if (any(still_missing)) {
+      processed$`Std_soilC_0-30cm`[still_missing] <- 
+        processed$`Total_soilC_0-30cm`[still_missing] * default_uncertainty
```
mdietze: Not clear this is valid. Generally bad to assume any constant default value, especially if the goal is ensemble error propagation. If missing data has to be filled, integrating over uncertainties would be the better choice.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+  
+  # Ensure mean is non-negative 
+  neg_mean_count <- sum(processed$`Total_soilC_0-30cm` < 0, na.rm = TRUE)
+  if (neg_mean_count > 0) {
+    PEcAn.logger::logger.warn(sprintf("Found %d negative mean values", neg_mean_count))
+    processed$`Total_soilC_0-30cm` <- pmax(processed$`Total_soilC_0-30cm`, 0, na.rm = TRUE)
+  }
+  
+  # Add minimum standard deviation to avoid zero uncertainty 
+  min_sd <- 0.1 * processed$`Total_soilC_0-30cm` # 10% of mean as minimum SD
```
mdietze: Again, not clear that you want to make such assumptions without evidence behind it. Citation needed. Also, if there is such thing as a minimum uncertainty, it's logically inconsistent to set sites with 0% uncertainty to 10% but leave those with 1% at 1%

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+  }
+  
+  # Get site row from processed data
+  site_row <- which(processed_data$Site_ID == site_id)
+  if (length(site_row) == 0) {
+    PEcAn.logger::logger.severe(sprintf("Site %s not found in processed data", site_id))
+  }
+  
+  # Set random seed for reproducibility
+  set.seed(as.numeric(site_id))
```
mdietze: PLEASE DON'T DO THIS. You're going to mess with any seed setting earlier in the workflow. If you need to ensure reproducibility do so prior to calling functions, not silently inside them. Also, this will fail miserably with non-numeric site_ids

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+  
+  # Generate all ensemble members at once 
+  soil_c_values <- truncnorm::rtruncnorm(
```
mdietze: Unclear why a truncated normal is being assumed when all prior analyses conducted by @Qianyuxuan suggested a gamma

mdietze (dismissed):

mdietze:
1. Unclear where these functions fit in the larger workflow, as I don't see any changes to any of the ic process functions that would be calling this (and merging it with aboveground data). Would be good to check that any function, and function argument, naming conventions used here are consistent with conventions assumed within the larger ic process workflow to ensure interoperability and polymorphism.
2. I'd really like @Qianyuxuan and @DongchenZ to review this before it is approved as a lot of this feels redundant with all the existing SoilGrids code they have developed. It's not obvious to me how this work fits with prior SoilGrids code or with prior SOC IC code.
3. bookdown documentation not updated

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: Thanks for your valuable feedback, and I apologize for the delay.

library(truncnorm) call was included by mistake, as I'm already referencing the functions using the namespace. I'll remove it.

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: I'm considering the following options:

1) Identify sites with missing data, log a warning about missing data, and filter out sites with missing data completely. 
2) Keep missing data as **NA** and let downstream processes handle it. 
3) Implementing a hierarchical approach where we try to find data from **similar sites** or **ecoregions** when direct measurements are unavailable. 

I'd appreciate your thoughts on these approaches.

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: I appreciate your feedback about integrating over uncertainties. I've implemented a statistical approach to address this:
```r
# Sites still with missing uncertainty - integrate over uncertainty distribution
still_missing <- is.na(processed$`Std_soilC_0-30cm`)
if (any(still_missing)) {
  # Calculate coefficient of variation (CV) from sites with data
  sites_with_data <- !is.na(processed$`Std_soilC_0-30cm`) & !is.na(processed$`Total_soilC_0-30cm`)
  
  if (sum(sites_with_data) >= 5) {
    # Calculate CV values from sites with data
    cv_values <- processed$`Std_soilC_0-30cm`[sites_with_data] / 
                 processed$`Total_soilC_0-30cm`[sites_with_data]
    
    # Remove extreme values
    cv_values <- cv_values[cv_values > 0 & cv_values < 1]
    
    # Fit gamma distribution to CV values
    gamma_fit <- try(suppressWarnings(MASS::fitdistr(cv_values, "gamma")), silent = TRUE)
    # ...
   ```
   Then in the ensemble generation function:
```r
   # For each ensemble member, sample a different CV value
if (processed_data$cv_distribution$type == "gamma") {
  cv_samples <- stats::rgamma(
    ensemble_size, 
    shape = processed_data$cv_distribution$shape,
    rate = processed_data$cv_distribution$rate
  )
}
# ...
# Generate ensemble members with different SDs
for (i in 1:ensemble_size) {
  sd_value <- mean_value * cv_samples[i]
  shape <- (mean_value^2) / (sd_value^2)
  rate <- mean_value / (sd_value^2)
  soil_c_values[i] <- stats::rgamma(1, shape = shape, rate = rate)
}
```

This approach:
1) Fits a gamma distribution to the coefficient of variation values from available data
2) For each ensemble member, samples a different CV value from this distribution
3) Uses these sampled CVs to generate ensemble members with varying uncertainties

Or would you suggest a different statistical approach? I'm particularly interested in whether the gamma distribution is appropriate for modeling CV uncertainty,
or since we already have mean and standard deviation then direct propagation through sampling might be more straightforward and interpretable than modeling CV via gamma.

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: Thank you for pointing out the issues with setting the seed inside the function.
```r
generate_soilgrids_ensemble <- function(..., seed = NULL) {
  # Only modify seed state if a seed is provided
  if (!is.null(seed)) {
    # Save current seed state if it exists
    old_seed <- if (exists(".Random.seed", envir = .GlobalEnv)) .Random.seed else NULL
    # Set the provided seed
    set.seed(seed)
    # Ensure restoration of previous state
    on.exit({
      if (!is.null(old_seed)) {
        assign(".Random.seed", old_seed, envir = .GlobalEnv)
      }
    })
  }
  
  # ........
}
```
This approach:
1) Makes seed setting optional via a parameter
2) Only modifies the random state when explicitly requested
3) Properly restores the previous state when the function exits

Would you prefer this approach ? let me know your thoughts .

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: The 0.15 factor was intended as a simple depth ratio (30cm/200cm), assuming uniform carbon distribution with depth. However, I realize I should have considered that soil carbon typically decreases non-linearly with depth and varies by ecosystem type.

Would you recommend: 
1) keeping the missing standard deviation values as `NA`? 
2) Implementing a more sophisticated approach based on literature about soil carbon depth distributions? 
3) Using a different method to estimate uncertainty for these sites?

If you recommend a literature-based approach, are there specific models or papers that have been used successfully in other parts that I should reference?

divine7022:
Hello sir, I’ve addressed the previous concerns in my comments and would value your thoughts on the updated approach. Could you please take a look whenever you have a moment.  
Thank you!

base/settings/R/get.site.info.R
```diff
@@ -0,0 +1,205 @@
...
+#' @examples
+#' \dontrun{
+#' # From settings object
+#' settings <- PEcAn.settings::read.settings("pecan.xml")
+#' site_info <- PEcAn.settings::get.site.info(settings)
+#'
+#' # From CSV file
+#' site_info <- PEcAn.settings::get.site.info(csv_path = "sites.csv")
+#' }
+get.site.info <- function(settings = NULL, csv_path = NULL, strict_checking = TRUE) {
```
DongchenZ: You can reduce the number of arguments by simply using the `path` instead of `settings` + `csv_path`. To do so, you just need to detect if the path extension is `.xml` or `.csv`. And read the settings inside the function if it ends with `.xml`.

DongchenZ: In general, I found this function is too complicated. In my previous code, I usually just use a single chunck of code to do the conversion, which I think might be useful in your case: 
```suggestion
site_info <- settings$run %>% 
                                 purrr::map('site')%>% 
                                 purrr::map(function(site.list){
                                   #conversion from string to number
                                   site.list$lat <- as.numeric(site.list$lat)
                                   site.list$lon <- as.numeric(site.list$lon)
                                   list(site.id=site.list$id, lat=site.list$lat, lon=site.list$lon, site_name=site.list$name)
                                 }) %>% 
                                 dplyr::bind_rows() %>% 
                                 as.list())
```

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+  # Start timing
+  start_time <- proc.time()
+  
+  # Extract site information using PEcAn.settings::get.site.info
+  site_info <- PEcAn.settings::get.site.info(settings = settings, csv_path = csv_path)
+  
+  # Get optional parameters from settings if available
+  ensemble_size <- ifelse(is.null(settings$ensemble$size), 1, settings$ensemble$size)
+  default_soilC <- ifelse(is.null(settings$soil$default_soilC), 5.0, settings$soil$default_soilC)
+  default_uncertainty <- ifelse(is.null(settings$soil$default_uncertainty), 0.2, settings$soil$default_uncertainty)
```
DongchenZ: I don't understand why you are trying to set the default value if there are no pre-settings to the soil mean and uncertainty. You can just set it to emplty `NULL` or `NA` if there is no soilgrids observation.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,373 @@
...
+  
+  # Create a data folder for intermediate outputs
+  data_dir <- file.path(dir, "SoilGrids_data")
+  if (!dir.exists(data_dir)) {
+    dir.create(data_dir, recursive = TRUE)
+  }
+  
+  # Log the number of sites being processed
+  n_sites <- nrow(site_info)
+  PEcAn.logger::logger.info(sprintf("Processing %d site(s)", n_sites))
```
DongchenZ: These functions, including the previous `get_site_info` function, contain a lot of messaging commands, such as `logger.info`. I think it could be useful to have a `verbose` argument to decide if the user want to print out the progresses.

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: Thank you for your feedback and for taking the time to review.

However, based on recent discussions, 
In the updated approach:

1.  **For missing `Total_soilC_0-30cm` :** We no longer fill these missing values with `default_soilC`. Instead, sites where the mean 0-30cm SoilGrids data is missing are now **omitted entirely** from the processed data used for ensemble generation.
2.  **For missing `Std_soilC_0-30cm` :** We also no longer use `default_uncertainty` for direct filling. The updated strategy involves calculating the CV from sites where both mean and standard deviation *are* available, fitting a **gamma distribution** to these CVs, and then sampling from this **gamma distribution** in the `generate_soilgrids_ensemble` function for sites with missing or zero SD.

but i didn't committed the this updated code yet. 
Let me know your thoughts on this approach.

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: Thanks for pointing out,

I have indeed added a `verbose` parameter to all the functions argument but missed in `get.site.info`.

However, you are correct that some of the progress messages (like "Creating output directory", "Extracting SoilGrids data", "Preprocessing soil carbon data", etc.) are currently using `logger.info` directly and are not conditional on the `verbose` flag.

I will update the code to wrap these `logger.info` calls within `if (verbose) { ... }` checks as well.

dlebauer:
@divine7022 could you please resolve conflicts?

@mdietze and @DongchenZ have all of your suggestions / requested changes been addressed?

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+  
+  site_info <- settings$run$site
+  if (is.list(site_info) && !is.null(site_info$id)) {
+    site_info <- list(site_info)
+  }
+  site_info <- site_info %>% 
+    purrr::map(function(site) {
+      site$lat <- as.numeric(site$lat)
+      site$lon <- as.numeric(site$lon)
+      str_id <- if (isTRUE(site$id > 1e9)) {
```
infotroph: Will fail with nonnumeric IDs: Notice that `"definitely not a number" > 1e9` evaluates to TRUE.

infotroph: More generally this function should use IDs exactly as they're given -- the user will be counting on them coming back identical to what they provided.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+        dims = list(
+          lat = current_site$lat,
+          lon = current_site$lon,
+          time = 1,
+          depth = c(0.3, 2.0) 
+        ),
+        vals = list(
+          soil_organic_carbon_content = c(ens_data_30cm[ens], ens_data_200cm[ens]),
+          wood_carbon_content = 0,
+          litter_carbon_content = 0
```
infotroph: Please don't write zeroes for the other pools -- better to save the soil C data by itself and assemble the IC file as a separate step once all the needed pools are processed.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+    "30cm" = list(mean_col = "Total_soilC_0-30cm", std_col = "Std_soilC_0-30cm"),
+    "200cm" = list(mean_col = "Total_soilC_0-200cm", std_col = "Std_soilC_0-200cm")
+  )
+  
+  cv_dist <- lapply(depths, function(depth_info) {
+    valid_cv <- processed[[depth_info$mean_col]] > 0 & 
+               !is.na(processed[[depth_info$std_col]]) & 
+               processed[[depth_info$std_col]] > 0
+    
+    if (sum(valid_cv) < 5) {
```
infotroph: why 5?

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+    
+    cv_values <- processed[[depth_info$std_col]][valid_cv] / processed[[depth_info$mean_col]][valid_cv]
+    cv_bounds <- stats::quantile(cv_values, probs = c(0.05, 0.95), na.rm = TRUE)
+    cv_filtered <- cv_values[cv_values >= cv_bounds[1] & cv_values <= cv_bounds[2]]
+    
+    if (length(cv_filtered) < 5) {
+      return(list(type = "none"))
+    }
+    
+    gamma_fit <- try(MASS::fitdistr(cv_filtered, "gamma"), silent = TRUE)
```
infotroph: Aren't the values in soil_data already taken from a fitted gamma? Why do we need to re-derive it here?

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+    valid <- !is.na(sd_values) & sd_values > 0
+    
+    if (any(valid)) {
+      shape_vec <- (mean_c^2) / (sd_values[valid]^2)
+      rate_vec <- mean_c / (sd_values[valid]^2)
+      soil_c_values[valid] <- pmax(stats::rgamma(sum(valid), shape_vec, rate_vec), 0)
+      soil_c_values[!valid] <- mean_c
+    } else {
+      soil_c_values <- rep(mean_c, size)
+    }
```
infotroph: If the user is asking for a sampling method, signal an error if it fails rather than quietly falling back to deterministic output.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+    
+    sd_values <- mean_c * cv_samples
+    valid <- !is.na(sd_values) & sd_values > 0
+    
+    if (any(valid)) {
+      shape_vec <- (mean_c^2) / (sd_values[valid]^2)
+      rate_vec <- mean_c / (sd_values[valid]^2)
+      soil_c_values[valid] <- pmax(stats::rgamma(sum(valid), shape_vec, rate_vec), 0)
+      soil_c_values[!valid] <- mean_c
```
infotroph: Please give errors or NAs _everywhere_ instead of filling in defaults. Deciding how to fill missing values always needs explicit user involvement.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+  soil_c_values <- numeric(size)
+  
+  # Use site-specific uncertainty
+  if (!is.na(original_sd) && original_sd > 0) {
+    shape <- (mean_c^2) / (original_sd^2)
+    rate <- mean_c / (original_sd^2)
+    if (is.finite(shape) && is.finite(rate) && shape > 0 && rate > 0) {
+      soil_c_values <- pmax(stats::rgamma(size, shape, rate), 0)
+    } else {
+      soil_c_values <- rep(mean_c, size)
```
infotroph: Same comment as below: Do not fill means, return NA or throw an error. Silent fallbacks like this can badly distort the sampled distribution and leave the user unaware of potentially severe data problems.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+    PEcAn.logger::logger.info(sprintf("Generating %d ensemble members for site %s (%s)",size, site_id, depth_layer))
+  }
+  
+  if (!is.null(seed)) {
+    if (exists(".Random.seed", envir = .GlobalEnv)) {
+      old_seed <- .Random.seed
+      on.exit(assign(".Random.seed", old_seed, envir = .GlobalEnv))
+    }
+    set.seed(seed)
+  }
```
infotroph: I strongly discourage trying to control RNG state at the individual function level and I see getting different results when you rerun ensemble generation as a desirable feature, not a problem. 

Instead I usually recommend setting the seed at the top level of a workflow and letting all subsequent steps inherit it from there.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+#' Generate soil carbon ensemble members for specific depth
+#'
+#' @param processed_data Output from preprocess_soilgrids_data()
+#' @param site_id Target site ID
+#' @param size Number of ensemble members to generate
+#' @param depth_layer Depth layer ("0-30cm" or "0-200cm")
+#' @param verbose Logical, print detailed progress information
+#' @param seed Optional random seed for reproducibility
+#' 
+#' @return Vector of soil carbon values with proper uncertainty handling
```
infotroph: Please say explicitly what handling is done that makes this "proper" (it might be best in the body of the function description instead of in the return value, though)

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,348 @@
...
+  }
+  if (verbose) {
+    PEcAn.logger::logger.info("Preprocessing soil carbon data following PEcAn standards")
+  }
+
+  # Only process sites with complete mean data for both depths
+  complete_sites <- !is.na(soil_data$`Total_soilC_0-30cm`) & 
+                   soil_data$`Total_soilC_0-30cm` > 0 &
+                   !is.na(soil_data$`Total_soilC_0-200cm`) & 
+                   soil_data$`Total_soilC_0-200cm` > 0
```
infotroph: I typically would want to pick one depth and use it throughout the data processing rather than carry both through to the IC file.

infotroph (dismissed):
As requested in previous comments, please do not automatically fill the mean or other fixed default for sites with missing data. Users might choose to do that, but they need to know it's happening and not have it hidden inside the sampling process.

I'm also leery of the design of writing NC files with zeroes for wood and leaf C and would strongly prefer a workflow that prepares only the soil C data, with assembly saved for a later step when all needed datasets are available.

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: @infotroph if I correctly understood what you mean -- soilgrids provides summary statistics (mean and standard deviation), not parametric distribution fits for stochastic sampling. Soilgrids soil carbon values are not pre-fitted to gamma distributions. Instead, it outputs summary statistics generated by ML ensemble methods (random forest + gradient boosting). These are empirical predictions, not parametric fits. (Ref: https://www.isric.org/sites/default/files/soilgrids250m_global_gridded_preprint.pdf)
The workflow here is:
1. calculate cv from sites with both mean and std dev available
2. fit gamma distribution to cv values across sites.
3. use fitted cv distribution to estimate uncertainty for sites with missing std dev
4. convert sampled cv values to site-specific standard deviations (sd = mean × cv)

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: thanks for raising this. would you recommend raising to 10 or 15 to strengthen parameter reliability?

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: fixed

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: removed

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: fixed - removed deterministic fallbacks. The function now throws explicit errors when stochastic sampling fails

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: I've changed this to use `NA` instead

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: fixed

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: removed 👍

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: I've updated description to describe the uncertainty handling

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: good suggestion! I've updated the code to use a flexible depth parameter `depth = c(0.3, 2.0)`

modules/data.land/R/IC_SOILGRID_Utilities.R
infotroph: At the moment I can't discern the intention behind this line, so it's hard to recommend alternatives. Can you say more about what this test is supposed to accomplish?

modules/data.land/NEWS.md
```diff
@@ -23,7 +23,11 @@
 * Included all relevant carbon pools (`ROOT_BIOMASS`, `AG_BIOMASS`, `SOIL_STOCK`, `LIT_BIOMASS`) in BADM-based IC extraction; excluded non-pool variables like `SOIL_CHEM`.
 * Added explicit support for `LIT_BIOMASS` to fully utilize **BADM** biomass capabilities.
 * Added `test-IC_BADM_Utilities.R` to validate BADM initial condition extraction and processing
-
+* New utility script `IC_SOILGRID_Utilities.R` for processing SoilGrids data to generate soil carbon initial condition (IC) files. This includes (#3508):
+  - **`soilgrids_ic_process`**: A function to extract, process, and generate ensemble members from SoilGrids250m data.
+  - **`preprocess_soilgrids_data`**: A helper function to handle missing values and ensure data integrity during preprocessing. 
+  - **`generate_soilgrids_ensemble`**: A function to create ensemble members for a site based on processed soil carbon data. 
+
```
infotroph, line 29: Please move to the `unreleased` section at top of file

modules/data.land/R/IC_SOILGRID_Utilities.R
infotroph: I know what SoilGrids provides. My point is that preprocess_soilgrids_data reads data written by  soilgrids_soilC_extract(), in which _we_ fitted the distribution.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,362 @@
...
+  mean_c <- processed_data$data[[mean_col]][site_row]
+  original_sd <- processed_data$data[[std_col]][site_row]
+  cv_dist <- processed_data$cv_distributions[[depth_layer]]
+  
+  if (is.na(mean_c) || mean_c <= 0) {
+    PEcAn.logger::logger.severe(sprintf("Invalid mean soil carbon value for site %s (%s)", 
+                                        site_id, depth_layer))
+  }
+  
+  soil_c_values <- numeric(size)
```
infotroph: No need to preassign here since there's no loop below -- the assignment on line 318 binds the name to a new object and won't reuse the memory from this one.

modules/data.land/R/IC_SOILGRID_Utilities.R
```diff
@@ -0,0 +1,362 @@
...
+      soil_c_values <- pmax(stats::rgamma(size, shape, rate), 0)
+    } else {
+      PEcAn.logger::logger.severe("Cannot generate an ensemble, invalid gamma params")
+    }
+  } else if (cv_dist$type != "none") {
+    # Integrate over uncertainty using CV distribution
+    if (cv_dist$type == "gamma") {
+      cv_samples <- stats::rgamma(size, cv_dist$shape, cv_dist$rate)
+      if (all(is.finite(cv_dist$bounds))) {
+        cv_samples <- pmax(pmin(cv_samples, cv_dist$bounds[2]), cv_dist$bounds[1])
```
infotroph: This will set all samples outside bounds to exactly the boundary value, which can create weird spikes at the edge of  the distribution if there an appreciable number of outliers. Is that going to be an issue here?

modules/data.land/R/ic_process.R
```diff
@@ -104,18 +104,47 @@ ic_process <- function(settings, input, dir, overwrite = FALSE){
...
+      site_files <- soilgrids_result[[str_ns]]
+      if (!is.null(site_files) && length(site_files) > 0) {
+        newfile <- as.list(unlist(site_files, use.names = FALSE))
+        names(newfile) <- rep("path", length(newfile))
+      } else {
+        PEcAn.logger::logger.severe(sprintf("No SoilGrids files generated for site %s", str_ns))
+      }
+    }
+    
+    settings$run$inputs[['poolinitcond']]$path <- newfile
```
infotroph: Will this result in PEcAn thinking IC setup is complete with files that *only* contain soil data and no vegetation? Or are these updated somewhere else I'm not seeing?

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: This test determines whether we have sufficient sites to fit a population model of coefficient of variation (CV) for spatial uncertainty borrowing.
When a site has mean soil carbon but missing/unreliable standard deviation from soilgrids, we need to estimate its uncertainty by cv pattern from sites with complete data.
1. calculate cv = std_dev/mean for sites with reliable uncertainty data 
2. if sufficient sites exist, fit gamma distribution to cv values 
3. for sites missing std_dev -- sample cv from fitted distribution, then  reconstruct std_dev = mean × sampled_cv
- `sum(valid_cv) < 5` --> too few sites --> `cv_distributions = "none"` -->  If a site with missing uncertainty is encountered -->  `logger.severe()` throws error → process terminates
- `sum(valid_cv) ≥ 5` --> fit gamma to cv values --> sites with missing uncertainty get ensemble estimates using cv pattern

here we are just _integrating over uncertainty_ when missing

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: the output from soilgrids_soilC_extract() can still contain NA values;
the CV distribution serves as a fallback mechanism; however it only activates when site-specific uncertainty estimation fails:

```r
if (!is.na(original_sd) && original_sd > 0) {
# use site-specific uncertainty (primary path)
shape <- (mean_c^2) / (original_sd^2)
...
} else if (cv_dist$type != "none") {
# fallback(integrate over uncertinity) -- spatial uncertainty borrowing via CV patterns
cv_samples <- stats::rgamma(size, cv_dist$shape, cv_dist$rate)
...
}
```

would like to hear your advice!

modules/data.land/R/ic_process.R
divine7022: since, SoilGrids only provides soil carbon data --
should SoilGrids use a different input category (e.g `soilinitcond`)?
```r
settings$run$inputs[['soilinitcond']]$path <- newfile
```

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: you're right about the edge artifacts from clamping. I've removed the bounds entirely;
fitted parameters already capture the empirical distribution shape, so additional bounds just add bias without benefit. 
removed 👍

modules/data.land/R/IC_SOILGRID_Utilities.R
divine7022: removed from here and moved where pre-allocation needed 👍

modules/data.land/NEWS.md
divine7022: done 👍

dlebauer (dismissed):
After a reading through again, the remaining changes requested are related to handling missing data:

- for soilgrids, don't fill missing data with mean or other fixed defaults
- for IC not covered by soilgrids, don't add 0s. Best fix is to exclude these variables, just write out the ones provided by soilgrids (and defer to #3603)

modules/data.land/R/ic_process.R
divine7022: any thoughts here ?

infotroph: It's correct for soil C to be stored in the `poolinitcond` category -- it's designed to contain the initial states of all pools, whatever soil/plant/water/atmosphere compartment they are part of), but the key point is that all input categories might be populated with inputs from more than one source. So my question is whether this change to ic_process will play nice with cases where, say, soil C comes from soilGrids but initial wood pools are pulled from a BADM file.

divine7022: The current implementation only processes **one source** per run when it detects an input source (such as BADM or Soilgrids), it processes that source and immediately `returns` the result. This is sufficient for BADM, since it provides a complete set of initial condition variables (both vegetation and soil pools). But for Soilgrids, which only provides soil carbon data, we end up with incomplete initial conditions, lacking vegetation and other non-soil pools.

Would it be appropriate, in cases where Soilgrids is used for soil carbon, to also extract vegetation data from BADM (or another appropriate source) and merge the two, so that we will be having both high-resolution soil carbon and vegetation variables? Is this approach consistent with how you'd recommend structuring the workflow?

Do you think it's correct to extend the logic so the workflow combines Soilgrids soil carbon with vegetation pools from BADM (or another source), for complete set of initial conditions? Or would you recommend handling this integration differently? 
Your advice will help clarify the path forward.

divine7022: hey @dlebauer and @infotroph , i think we can remove the `ic_process` modifications from this PR and have another dedicated PR for that, where we can think about all the edge cases -- like adding soilgrid and pulling out veg and non-soil pools from other sources (like BADM or any other appropriate one) rather than silently filling veg /non-soil pools without users knowing how/where it gets filled, and we could introduce a concept of multi-source settings since currently pecan supports single source; and other improvements that can be done. 
this way we can focus workflow integration separately.

how about that, does it make sense?

dlebauer: I've created a feature request (#3603) to update ic_process to support multiple sources. To merge this PR, please remove changes to `ic_process()`.

divine7022: Thanks for creating @dlebauer i will remove it now

just a thought, instead of calling **BADM** here for filling `veg/non-soil pools` (since it provides a **complete** initial conditions data set), we should just extract the vegetation and other non-soil pools from BADM and merge those with the SoilGrids SOC data. I don't think this would be an ideal approach.

if `SoilGrids` is the passed as source in settings, another option would be to call `Landtrendr_AGB_prep()` to get aboveground biomass `medians and standard deviations`, and use `extract_SM_CDS()` for `soil moisture`. 
That way, we end up with variables like `soil_organic_carbon_content`, `SoilMoistFrac`, `LAI`, and `AbvGrndBiomass`. Then from those, we can calculate further pools like `AbvGrndWood`, `leaf_carbon_content`, and `wood_carbon_content`, as needed.

i think then it make sence to pass to this file as a `poolinitcond` in `ic_process`
```r
settings$run$inputs[['poolinitcond']]$path <- newfile
```

dlebauer (approved):
Yay! Thanks for the heroic effort of addressing all feedback and pushing this through!

PR #3510, opened 2025-04-14
fix: "/pkgdoc/package_documentation/pkgdocs": not found

divine7022:
## Description
This PR fixes an issue with the Docker build process that was failing with the error:

image

#### Changes Made:
1) Added a new line to the Dockerfile to create the required directory structure before building the documentation.
2) This ensures that the `/pkgdoc/package_documentation` directory exists before the `make pkgdocs` command runs, allowing the build process to correctly generate and store documentation in `pkgdocs` directory.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


infotroph:
Is that screenshot from run [CI run 40368329071 ](https://github.com/PecanProject/pecan/actions/runs/14394515442/job/40368329071)? Hard to say why that one failed, but it looks like other runs in the develop branch, including [the most recent](https://github.com/PecanProject/pecan/actions/runs/14436444259/job/40479911054), are succeeding. That makes me hopeful the error in 40368329071 was a transient one-off.

I also don't think what you have here would be the right fix even if the error persists -- it would copy any existing `package_documentation` directory in from the host system, whereas the intended design is for the Docker build to compile all pkgdown pages fresh inside the container. The package_documentation dir doesn't exist at the start of the build, but is created as part of the call to `make pkgdocs`.

divine7022:
> Is that screenshot from run [CI run 40368329071 ](https://github.com/PecanProject/pecan/actions/runs/14394515442/job/40368329071)? Hard to say why that one failed, but it looks like other runs in the develop branch, including [the most recent](https://github.com/PecanProject/pecan/actions/runs/14436444259/job/40479911054), are succeeding. That makes me hopeful the error in 40368329071 was a transient one-off.
> 
> I also don't think what you have here would be the right fix even if the error persists -- it would copy any existing `package_documentation` directory in from the host system, whereas the intended design is for the Docker build to compile all pkgdown pages fresh inside the container. The package_documentation dir doesn't exist at the start of the build, but is created as part of the call to `make pkgdocs`.

Thanks for feedback! I’ll take a closer look now.

divine7022:
> > Is that screenshot from run [CI run 40368329071 ](https://github.com/PecanProject/pecan/actions/runs/14394515442/job/40368329071)? Hard to say why that one failed, but it looks like other runs in the develop branch, including [the most recent](https://github.com/PecanProject/pecan/actions/runs/14436444259/job/40479911054), are succeeding. That makes me hopeful the error in 40368329071 was a transient one-off.
> > I also don't think what you have here would be the right fix even if the error persists -- it would copy any existing `package_documentation` directory in from the host system, whereas the intended design is for the Docker build to compile all pkgdown pages fresh inside the container. The package_documentation dir doesn't exist at the start of the build, but is created as part of the call to `make pkgdocs`.
> 
> Thanks for feedback! I’ll take a closer look now.

divine7022:
**Key improvements from the recent PR:**
1. Added function to generate `_pkgdown.yml`:
Introduced dynamic generation of `_pkgdown.yml` for all packages to customize the pkgdown site appearance. Upgraded the default Bootstrap version from 3 (**deprecated**) to Bootstrap 5, ensuring modern UI compatibility. Also added a persistent link to return to the main landing page (`_pkgdown_docs/index.html`).
2. Structured pkgdown output directories by source:
Refactored the _pkgdown_docs output layout by categorizing packages under their respective source directories (e.g., `_pkgdown_docs/base/PEcAn.logger` instead of `_pkgdown_docs/PEcAn.logger`). This improves clarity and reflects the actual package organization within the PEcAn source tree, rather then all the packages in the same level.
3. Improved automation of the main documentation landing page(`_pkgdown_docs/index.html`)  by grouping and linking top-level directories (base/, models/, modules/). This provides a cleaner and more navigable structure for accessing the documentation site.

**Further Improvements:**
While the current `_pkgdown.yml` setup is consistent across all packages, future improvements could include package-specific customization for finer control over individual documentation sites.

which looks like:

image

scripts/build_pkgdown.R
```diff
@@ -1,15 +1,16 @@
...
-
+library(yaml)
 args <- commandArgs(trailingOnly = TRUE)
 if (length(args) == 0) {
   stop("No package names provided. Please pass package names as arguments.")
 }
-
 packages <- args
-
+output_dir <- "_pkgdown_docs"
```
infotroph, line 11: 👍 to adopting this name for the outdir, but I note you're still using `package_documentation` in the GHA and Docker configs above

scripts/build_pkgdown.R
```diff
@@ -18,15 +19,81 @@ if (requireNamespace("PEcAn.logger", quietly = TRUE)) {
   }
 }
 
-logger("Building pkgdown docs for:", paste(packages, collapse = ", "))
+pkg_config <- function(pkg) {
+  pkgname <- desc::desc_get("Package", pkg)
+
+  list(
```
infotroph: Looks like you're implementing an entire custom layout, which seems more involved than we need or want to maintain. I was picturing something more like [adding one link to the navbar](https://pkgdown.r-lib.org/articles/customise.html#additional-html-and-files) while keeping all other defaults:
```
template:
  bootstrap: 5.0
  includes:
    before_navbar: <a href='../../index.html'>All Packages</a>
```

...but it can't be exactly that because the number of levels up will vary depending if we're looking at `<pkgname>/index.html`, `<pkgname>/reference/<function>.html`, etc. Figuring out how to handle that effectively is the crux of the task.

scripts/build_pkgdown.R
```diff
@@ -18,15 +19,81 @@ if (requireNamespace("PEcAn.logger", quietly = TRUE)) {
...
+    pkg_config <- pkg_config(pkg)
+    # If _pkgdown.yml exists, merge with our config, otherwise create new
+    if (file.exists(pkg_config_path)) {
+      exist_config <- yaml::read_yaml(pkg_config_path)
+      # Merge configurations, preserving existing settings
+      merged_config <- modifyList(exist_config, pkg_config)
+      yaml::write_yaml(merged_config, pkg_config_path)
+    } else {
+      yaml::write_yaml(pkg_config, pkg_config_path)
+    }
```
infotroph: It's almost always a bad idea to allow hand-edits and autogenerated entries in the same file. Better to commit to the principle that _pkgdown.yml is generated by this script and regenerated as needed.

```suggestion
    yaml::write_yaml(pkg_config, pkg_config_path)
```

scripts/build_pkgdown.R
```diff
@@ -18,15 +19,81 @@ if (requireNamespace("PEcAn.logger", quietly = TRUE)) {
...
+logger("Building pkgdown docs for:", paste(packages, collapse = ", "))
 for (pkg in packages) {
   logger("Building pkgdown site for:", pkg)
   current_wd <- getwd()  
   tryCatch({
     if (!dir.exists(pkg)) {
       stop(paste("Package directory does not exist:", pkg))
     }
+    pkg_config_path <- file.path(pkg, "_pkgdown.yml")
+    pkg_config <- pkg_config(pkg)
```
infotroph: This overwrites the function definition and makes the next loop iteration throw an error!

scripts/build_pkgdown.R
```diff
@@ -18,15 +19,81 @@ if (requireNamespace("PEcAn.logger", quietly = TRUE)) {
...
 
+logger("Building pkgdown docs for:", paste(packages, collapse = ", "))
 for (pkg in packages) {
   logger("Building pkgdown site for:", pkg)
   current_wd <- getwd()  
   tryCatch({
     if (!dir.exists(pkg)) {
       stop(paste("Package directory does not exist:", pkg))
     }
+    pkg_config_path <- file.path(pkg, "_pkgdown.yml")
```
infotroph: Is _pkgdown.yml allowed to live inside the docs/ directory, or otherwise not clutter the root of the package directory?

scripts/build_pkgdown.R
```diff
@@ -35,11 +102,17 @@ for (pkg in packages) {
       warning(paste("No docs folder created for:", pkg))
       next 
     }
-    dest <- file.path("package_documentation/pkgdocs", pkg)
-    if (!dir.exists(dest)) {  
+    pkgname <- desc::desc_get("Package", pkg)
+    dest <- file.path(output_dir, strsplit(pkg, "/")[[1]][1], pkgname)
```
infotroph: I do not think the base / models / modules  arrangement needs to be reflected in the package index -- these folders are mostly for developer convenience and they don't really map to any groupings a PEcAn user will care about. Better just to keep this one flat directory.
```suggestion
    dest <- file.path(output_dir, pkgname)
```

infotroph (changes requested):
Thank you for tackling links back to the all-packages index! But that should be a separate PR rather than added on to this one, which is related only by involving pkgdown.

In addition, these changes to build_pkgdown don't run for me; there are at least two syntax errors from stray commas in the list returned by pkg_config, and then inside the loop the pkg_config function gets overwritten by its own output, causing an error when trying to build the second and subsequent packages. I'll review the high-level approach inline and may have more requests once I can see it running.

scripts/build_pkgdown.R
divine7022: **Thanks for the feedback!**
The reason I opted for custom settings is because just adding one element (like a link to “All Packages”) while trying to retain the default UI doesn’t fully work as expected.

Here’s why:
1) The default layout gives us basic elements , but it doesn’t include a modernized UI elements or features like the light/dark mode toggle (`light-switch`).
2) When we try to include only the additional components (like `light-switch`) without explicitly redefining the navbar structure, the defaults like **reference/news** are not preserved; they get replaced instead of merged.
3) To get a modern landing page under **`docs/dev/index.html`** , we need to explicitly configure the `template` and `navbar` sections. Without these, we fall back to the less modern default under **`docs/index.html`**.

So the custom config that gives us both a usable UI and required features looks like this:
```r
  list(
    url = "https://pecanproject.github.io/",
    home = list(
      title = sprintf("%s Functions for PEcAn", pkgname)
    ),
    template = list(
      bootstrap = 5,
      bslib = list(
        primary = "#0054AD", 
        `border-radius` = "0.5rem",
        `btn-border-radius` = "0.25rem"
      ),
      `light-switch` = TRUE
    ),
    navbar = list(
      structure = list(
        left = c("home", "pecan_home", "reference", "news"),
        right = c("search", "github", "lightswitch")
      ),
      components = list(
        home = list(
          icon = "fa fa-home",
          href = "index.html",
          `aria-label` = "PEcAn Documentation Home"
        ),
        pecan_home = list(
          text = "All Packages",
          href = "../../index.html",
          `aria-label` = "All Packages"
        ),
        reference = list(
          text = "Reference",
          href = "reference/index.html"
        ),
        github = list(
          icon = "fab fa-github",
          href = "https://github.com/PecanProject/pecan",
          `aria-label` = "GitHub"
        )
      )
    ),
    reference = list(
      list(
        title = "All Functions",
        desc = "All functions exported by this package",
        contents = list("matches('.*')")
      )
    ),
    news = list(
      text = "News",
      href = "news/index.html"
    ),
    development = list(
      mode = "auto"
    )
  )
  ```
  
  
  
#### So by trying all the possible ways to simplify the settings; even i felt, through this minimum settings which gives only necessary defaults required for us by also including the persistent button, but not include the features like light-switch etc...
``` r
list(
    url = "https://pecanproject.github.io/",
    template = list(
      bootstrap = 5,
      includes = list(
        before_navbar = paste0(
          "<style>",
          ".custom-back-button{position: absolute; top: 10px;left:10px; z-index: 1000; background-color: #f5f5f5; color: #000;border:none }.navbar-brand{ margin-left: 50px }",
          "</style>\n",
          "<button onclick='window.location.href=\"../index.html\"' ",
          "class='custom-back-button btn btn-primary'>",
          "← All Packages</button>"
        )
      )
    )
  )
  ```

- And in above approach added a button in navbar to get back to "All Packages" , which works for navigating from any subpage (`<pkgname>/reference/index.html`, `<pkgname>/reference/<function>.html`, etc.) to `<pkgname>/index.html`, and then from there to the top-level `_pkgdown_docs/index.html`.
  
Let me know your thoughts on these approach.

File: scripts/build_pkgdown.R (hunk 5 above)
Comment (divine7022) [reply scripts/build_pkgdown.R]: Well thought! , but `_pkgdown.yml` file needs to be in the root directory of the package, not in the `docs/` directory. This is because `pkgdown` looks for this configuration file specifically at the package **root** level when building the **site**.

If we put the configuration file(`_pkgdown.yml`) inside `docs/` or any other subdirectory, pkgdown won't find it and will either use default settings (bootstrap 3) or fail to properly configure the site.

pkgdown looks for the configuration file. according to its documentation:
1) First, it looks for `_pkgdown.yml` or `_pkgdown.yaml` in the package root directory.
2) If not found, it falls back to default configuration.

If you're concerned about cluttering the package root directory, there are a couple of approaches we could consider:

1) Using a temporary configuration file, by creating the file temporarily during the build process and then deleting it afterward.
2) Using pkgdown configuration; instead of writing a yml file, we can pass configuration directly to the `build_site( )` function in `build_pkgdown.R` script:
```r
pkgdown::build_site(
  pkg = ".",
  override = list(
    url = "https://pecanproject.github.io/",
    template = list(
      bootstrap = 5,
      includes = list(
        before_navbar = paste0(
          "<style>",
          ".custom-back-button{position: absolute; top: 10px;left:10px; z-index: 1000; background-color: #f5f5f5; color: #000;border:none }.navbar-brand{ margin-left: 50px }",
          "</style>\n",
          "<button onclick='window.location.href=\"../index.html\"' ",
          "class='custom-back-button btn btn-primary'>",
          "← All Packages</button>"
        )
      )
    )
  )
)
```

divine7022:
This commit includes:
- A button to navigate back to All Packages.
- I’ve also tried to address all the feedback mentioned above.
- Added `URL` and `BugReports` fields in all package `DESCRIPTION` files. This was done for a couple of reasons:
  1) To avoid the warning `URLs not ok` while generating the documentation site.
  2) Without these fields, the GitHub icon will not appear on the documentation site when using default settings.
  3) And to maintain consistency acorss all the packages.


Please review and try it out. Let me know if it works for you or if any further improvements are needed

divine7022:
For testing the `pkgdown` docs build locally, i have pushed the container to Docker registry, you can access it using these commands:
```
docker pull akash7022/pkg:latest
docker run -d -p 8080:80 akash7022/pkg:latest
```
Once the container is running, go to:
http://localhost:8080

divine7022:
Tested the `pkgdown` docs build locally, and i have pushed the container to Docker registry, you can access it using these commands:
```
docker pull akash7022/pkg:latest
docker run -d -p 8080:80 akash7022/pkg:latest
```

Once the container is running, go to:
http://localhost:8080

scripts/build_pkgdown.R
infotroph: Ah, cool! Yes, passing the configuration directly in `override` seems like the way to go.

.github/workflows/pkgdown.yml
```diff
@@ -24,9 +24,9 @@ jobs:
     # Checkout source code
     - uses: actions/checkout@v4
 
-    # Install pkgdown
+    # Install dependencies
     - name: Install dependencies
-      run: Rscript -e 'install.packages("pkgdown")'
+      run: Rscript -e 'install.packages(c("pkgdown", "yaml", "desc"))'
```
infotroph: Are you sure desc is needed here? The depends image already has it installed -- it's used in the script that installs everything else.

base/all/DESCRIPTION
```diff
@@ -43,6 +43,8 @@ Description: The Predictive Ecosystem Carbon Analyzer
     PEcAn is to streamline the interaction between data and
     models, and to improve the efficacy of scientific
     investigation.
+URL: https://pecanproject.github.io, https://github.com/PecanProject/pecan
```
infotroph: 1. I don't think we need both of these links -- it's easy enough to get to the repo from either the BugReports URL or the links from the github.io page
2. How specifically does pkgdown use the URL from DESCRIPTION in the compiled pages, and how does that differ from how it uses  the url from `_pkgdown.yml`? In particular, will this mean there are places where following links from the locally-served Docker docs could take people to the github.io page instead of the local file equivalent?
3. Will it be a problem that all packages are pointing to the all-of-PEcAn site instead of to package-specific resources?
```suggestion
URL: https://pecanproject.github.io
```

scripts/build_pkgdown.R
infotroph: > The default layout gives us basic elements , but it doesn’t include a modernized UI elements or features like the light/dark mode toggle

Adding a mode toggle is _definitely_ outside the requirements for this task. Please implement one thing at a time.

> the defaults like reference/news are not preserved; they get replaced instead

Yes, that's why I want to keep defaults everywhere we can -- if there's really no way to add one link without losing all the pieces pkgdown was providing for free, that's a cue to rethink the approach before building a whole new layout.

> To get a modern landing page under docs/dev/index.html

- I don't know where you're getting the `/dev/` part
- What specifically do you mean by "modern" and why do we need it?

scripts/build_pkgdown.R
```diff
@@ -18,15 +20,37 @@ if (requireNamespace("PEcAn.logger", quietly = TRUE)) {
   }
 }
 
-logger("Building pkgdown docs for:", paste(packages, collapse = ", "))
+pkg_config <- function() {
 
+  list(
+    url = "https://pecanproject.github.io/",
```
infotroph: I found a partial answer to my question above about the difference between `URL:` in the DESCRIPTION and `url` here: When we specify a url here, _all internal links_ are changed from relative paths to absolute URLS.

It's possible that's desirable for some cases, but for our local Docker deployment it means we'll either have locally hosted pages that are filled with links out to the Github-hosted equivalents, or that we'll need to rewrite the url field when compiling pkgdocs at Docker build time.

I think for this PR my recommendation is to remove this url from the yml so that the local pages keep working locally; we could rethink this later as needed.
```suggestion
```

scripts/build_pkgdown.R
infotroph: Does using an override also mean we don't need to add the {yaml} package dependency?

scripts/build_pkgdown.R
```diff
@@ -50,4 +80,36 @@ for (pkg in packages) {
...
+before_text <- c(
+  '<!DOCTYPE html>',
+  '<html lang="en">',
+  '<head>',
+  '  <title>Package-specific documentation for the PEcAn R packages</title>',
+  '</head>',
+  '<body>',
+  '<h1>PEcAn package documentation</h1>',
+  '<p>Function documentation and articles for each PEcAn package,',
+  '   generated from the package source using <a href="https://pkgdown.r-lib.org/" target="_blank">pkgdown</a> package.</p>',
```
infotroph: Nit: this sentence has too many copies of the word "package" in it, and this one's the easiest to remove 😉 
```suggestion
  '   generated from the package source using <a href="https://pkgdown.r-lib.org/" target="_blank">pkgdown</a>.</p>',
```

scripts/build_pkgdown.R
divine7022: > I don't know where you're getting the /dev/ part

* The `<package>/docs/dev/index.html` file and other directories like `references` etc.. are generated under `/dev/` when we add custom configurations. If we rely on the defaults, this directory isn't created at all.

> What specifically do you mean by "modern" and why do we need it?

* Ha i do agree that we really don't need any modern ui elements, rather then focusing on the core functionality and the purpose of it at this point.

infotroph: > generated under /dev/ when we add custom configurations

Ah, OK. In that case I think the `/dev/` can be dropped -- recall that the whole collection is deployed into a separate directory for each PEcAn version, so we'll have annotation of dev vs release vs tag versions already built into the URL.

scripts/build_pkgdown.R
divine7022: > Ah, cool! Yes, passing the configuration directly in `override` seems like the way to go.

I'll test it again to make sure everything works as intended.

> Does using an override also mean we don't need to add the {yaml} package dependency?

hopefully not, but I’ll take a closer look to confirm.

scripts/build_pkgdown.R
divine7022: Really good catch! , and a strong reason to conclude that we don’t need to generate a separate `/dev/` directory just for custom configurations.

.github/workflows/pkgdown.yml
divine7022: I didn't test them separately, but initially when i tried to test the docker build it was just bypassing the packages, instead of actually generating the docs for each one.
i will recheck the build.

base/all/DESCRIPTION
divine7022: Thanks for the suggestion! , having both links in the **`DESCRIPTION`** isn't strictly necessary. The `URL` field is primarily used as metadata by R and doesn't impact how the site is built. Since it's already easy to navigate to the GitHub repo via the BugReports field or from other links on the site, I'm going ahead to remove `https://github.com/PecanProject/pecan`

> Will it be a problem that all packages are pointing to the all-of-PEcAn site instead of to package-specific resources?

Regarding pacage-specific resource, I took a closer look, and the issue doesn’t seem to be with the `URL `field in the `DESCRIPTION` or `_pkgdown.yml`, but rather how pkgdown generates source code links by default.

By default, pkgdown creates links like, eg:
```
https://github.com/PecanProject/pecan/blob/HEAD/R/logger.R
```
These aren't correct for accessing the source of individual packages . To fix this, we can override it when calling `pkgdown::build_site()` to generate correct links for each package, like:
```r
pkgdown::build_site(
  pkg = ".",
  override = list(
    repo = list(
      url = list(
        source = paste0("https://github.com/PecanProject/pecan/blob/develop/", pkg)
      )
    ),
    template = list(
      bootstrap = 5,
      includes = list(
        before_navbar = paste0(
          "<style>",
          ".custom-back-button{position: absolute; top: 10px;left:10px; z-index: 1000; background-color: #f5f5f5; color: #000;border:none }.navbar-brand{ margin-left: 50px }",
          "</style>\n",
          "<button onclick='window.location.href=\"../index.html\"' ",
          "class='custom-back-button btn btn-primary'>",
          "← All Packages</button>"
        )
      )
    )
  )
)
```
This ensures the source links point to the correct location eg: 
```
https://github.com/PecanProject/pecan/blob/develop/base/logger/R/logger.R 
```
can see like **Source: `R/logger.R`** label on the site.

image

but one con of this approach is that we won't get the github icon by default. However, I believe accessing the repository from other links on the site is easy enough, so I'm neglecting that at this point.

let me know your thoughts on this.

.github/workflows/pkgdown.yml
divine7022: I’ve verified that both `yaml` and `desc` are already installed in the **`pecan/depends:develop`** image.

divine7022:
Hello sir, as we say in recent deployment to `package-documentation` , `index.html` page currently resides inside the `develop/`. Do you think it would be more convenient to view the documentation directly via the deployed gh page if the `index.html` were placed in the root (main)?
So, should we remove `${version}` from `rsync` path of `pkgdown.yml` ,then the output goes directly to the root for easier access ?

Or would you suggest any alternative.

infotroph:
That's a good thought, but putting it in a subirectory was intentional to facilitate serving multiple versions of the docs. You only see `/develop/` today, but it will eventually parallel the long-form documentation that has https://pecanproject.github.io/documentation/master/, https://pecanproject.github.io/documentation/develop/, https://pecanproject.github.io/documentation/version171/, and so on.

divine7022:
```bash
rsync -a --delete ../_pkgdown_docs/ ${VERSION}/
        
        if [ "$VERSION" = "develop" ]; then
          echo '<!DOCTYPE html>
          <html>
          <head>
            <meta http-equiv="refresh" content="0; url=./develop/" />
            <title>Redirecting to Documentation</title>
          </head>
          <body>
            <p>Redirecting to <a href="./develop/">latest documentation</a>...</p>
          </body>
          </html>' > index.html
        fi
        
        git add --all *
```

so an alternative, if branch is `develop` , it creates a blank webpage named `index.html` , which immediately (0-second delay) send the visitor to the `./develop/ ` dir.
currently if we visit `https://github.com/PecanProject/package-documentation/` we see only `readme`, we must manually type `/develop` to see docs. so in the above approach - if we visit `https://github.com/PecanProject/package-documentation/` the page automatically loads `https://github.com/PecanProject/package-documentation/develop/`

 Open to suggestions if there's a cleaner or more standard way to handle this - happy to adjust!

base/all/DESCRIPTION
divine7022: hello sir, if you have a moment could you please review the changes.

divine7022:
> ```shell
> rsync -a --delete ../_pkgdown_docs/ ${VERSION}/
>         
>         if [ "$VERSION" = "develop" ]; then
>           echo '<!DOCTYPE html>
>           <html>
>           <head>
>             <meta http-equiv="refresh" content="0; url=./develop/" />
>             <title>Redirecting to Documentation</title>
>           </head>
>           <body>
>             <p>Redirecting to <a href="./develop/">latest documentation</a>...</p>
>           </body>
>           </html>' > index.html
>         fi
>         
>         git add --all *
> ```
> 
> so an alternative, if branch is `develop` , it creates a blank webpage named `index.html` , which immediately (0-second delay) send the visitor to the `./develop/ ` dir. currently if we visit `https://github.com/PecanProject/package-documentation/` we see only `readme`, we must manually type `/develop` to see docs. so in the above approach - if we visit `https://github.com/PecanProject/package-documentation/` the page automatically loads `https://github.com/PecanProject/package-documentation/develop/`
> 
> Open to suggestions if there's a cleaner or more standard way to handle this - happy to adjust!

hello sir, any insights on this approach?

scripts/build_pkgdown.R
```diff
@@ -29,7 +30,29 @@ for (pkg in packages) {
       stop(paste("Package directory does not exist:", pkg))
     }
     setwd(pkg) 
-    pkgdown::build_site() 
+    pkgdown::build_site(
+      pkg = ".",
+      override = list(
+        repo = list(
+          url = list(
+            source = paste0("https://github.com/PecanProject/pecan/blob/develop/", pkg)
```
robkooper: is this safe to be hardcoded to develop?

robkooper (dismissed):
looks like there is a hardcoded devlop, see comment.

divine7022:
@robkooper Thanks for your feedback, and sorry for the delay, since it's a busy week with assessments, so i couldn't address this sooner
> looks like there is a hardcoded devlop, see comment.

this is a good catch! ,this should be made dynamic to detect the current branch.
We could make the configuration like:
```r
branch <- Sys.getenv("PECAN_GIT_BRANCH", unset = "develop")  

source = paste0("https://github.com/PecanProject/pecan/blob/", branch, "/", pkg)
```
in the `build_pkgdown.R` 

And in the GHA workflow (`pkgdown.yml`), we could add :
```yml
env:
  PECAN_GIT_BRANCH: ${{ github.event_name == 'pull_request' && github.base_ref || github.ref_name }}
```
The `github.base_ref ` gives the target branch name during the PR, but if the R script is run on a push requests and other events types, `github.ref_name` will provide the branch or tag name.

This approach ensures that the branch name is always correctly set to the target branch for PRs and the branch name for push request and other groups, which is exactly what we want for generating correct links in the documentation.

Let me know if you have any suggestions or would like me to proceed with these changes.

divine7022:
> ```shell
> rsync -a --delete ../_pkgdown_docs/ ${VERSION}/
>         
>         if [ "$VERSION" = "develop" ]; then
>           echo '<!DOCTYPE html>
>           <html>
>           <head>
>             <meta http-equiv="refresh" content="0; url=./develop/" />
>             <title>Redirecting to Documentation</title>
>           </head>
>           <body>
>             <p>Redirecting to <a href="./develop/">latest documentation</a>...</p>
>           </body>
>           </html>' > index.html
>         fi
>         
>         git add --all *
> ```
> 
> so an alternative, if branch is `develop` , it creates a blank webpage named `index.html` , which immediately (0-second delay) send the visitor to the `./develop/ ` dir. currently if we visit `https://github.com/PecanProject/package-documentation/` we see only `readme`, we must manually type `/develop` to see docs. so in the above approach - if we visit `https://github.com/PecanProject/package-documentation/` the page automatically loads `https://github.com/PecanProject/package-documentation/develop/`
> 
> Open to suggestions if there's a cleaner or more standard way to handle this - happy to adjust!

@robkooper 
If you have a moment, could you please take a look at this following conversion ?

infotroph:
I'm sorry for my long delay answering here. 

> currently if we visit https://github.com/PecanProject/package-documentation/ we see only readme, we must manually type /develop to see docs

I think what we want to see at that URL is a list of available versions, like we currently see when we visit https://pecanproject.github.io/documentation. Deciding how to implement that can be a separate PR -- for this one it seems fine to leave that as it is and remember to give people links that have the `/develop/` already included.

dlebauer:
@divine7022 what are next steps here?

divine7022:
@dlebauer Apologies for taking so long to respond here. 

I think this is ready to merge as-is; 
It addresses the `Bootstrap 3` deprecation warnings we've been getting by upgrading to `Bootstrap 5`, plus adds some nice navigation improvements with the custom back button, dynamic branch-specific links to source code of a package and BugReports, and some other improvements. 

As @infotroph mentioned, we can have a separate PR for the landing page version-specific links listed.

docker/docs/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get update \
         -e 'remotes::install_version("rmarkdown", ">= 2.19", dependencies = TRUE, upgrade = FALSE, repos = repos)' \
         -e 'remotes::install_version("knitr", ">= 1.42", dependencies = TRUE, upgrade = FALSE, repos = repos)' \
         -e 'remotes::install_version("bookdown", ">= 0.31", dependencies = TRUE, upgrade = FALSE, repos = repos)' \
-        -e 'install.packages("pkgdown", repos = repos)' \
+        -e 'install.packages(c("pkgdown", "yaml", "desc"), repos = repos)' \
```
infotroph:
```suggestion
```

These are already on the parent image, no need to spend time reinstalling (I wish install.packages had a "skip if already installed" mode, but it does not)

docker/docs/Dockerfile
```diff
@@ -65,4 +65,4 @@ ARG PECAN_GIT_DATE="unknown"
 ENV PECAN_VERSION=${PECAN_VERSION} \
     PECAN_GIT_BRANCH=${PECAN_GIT_BRANCH} \
     PECAN_GIT_CHECKSUM=${PECAN_GIT_CHECKSUM} \
-    PECAN_GIT_DATE=${PECAN_GIT_DATE}
+    PECAN_GIT_DATE=${PECAN_GIT_DATE}
```
infotroph:
```suggestion
    PECAN_GIT_DATE=${PECAN_GIT_DATE}

```

scripts/build_pkgdown.R
```diff
@@ -29,7 +33,30 @@ for (pkg in packages) {
...
     }
     setwd(pkg) 
-    pkgdown::build_site() 
+    pkgdown::build_site(
+      pkg = ".",
+      override = list(
+        repo = list(
+          url = list(
+            source = paste0("https://github.com/PecanProject/pecan/blob/", 
+                            branch, "/", pkg)
```
infotroph:
```suggestion
                            branch, "/", pkg, "/")
```

scripts/build_pkgdown.R
```diff
@@ -29,7 +33,30 @@ for (pkg in packages) {
...
+          bootstrap = 5,
+          includes = list(
+            before_navbar = paste0(
+              "<style>",
+              ".custom-back-button{position: absolute; top: 10px;left:10px; z-index: 1000; background-color: #f5f5f5; color: #000;border:none }.navbar-brand{ margin-left: 50px }",
+              "</style>\n",
+              "<button onclick='window.location.href=\"../index.html\"' ",
+              "class='custom-back-button btn btn-primary'>",
+              "← All Packages</button>"
+            )
```
infotroph:
```suggestion
            before_navbar = "<a href=\"../index.html\" style=\"padding: 0em 1em\">← Up</a>"
```

Reasons for this change:

1. Using an ordinary link (instead of a button that only acts via onclick) allows standard navigation such as cmd-clicking to open in a new tab
2. In narrow windows the absolute positioning leads to the button covering the package name
3. This link is technically an "up one level", which is not always a direct link to the all-packages page. When you're on a function reference page or a package article, one level up is to the readme for that package. I'm fine with that behavior if it's labeled as such.

infotroph (changes requested):

infotroph (approved):

PR #3511, opened 2025-04-15
Add a new state variable of soil water (in cm) to enable the direct assimilation of volumetric soil moisture

Qianyuxuan:
Add a new state variable of soil water (in cm) to enable the direct assimilation of volumetric soil moisture.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/R/model2netcdf.SIPNET.R
+++ b/models/sipnet/R/model2netcdf.SIPNET.R
@@ -196,7 +196,7 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date, 
      output[["Qle"]] <- (sub.sipnet.output$evapotranspiration * 10 * PEcAn.data.atmosphere::get.lv()) / timestep.s  # Qle W/m2
     }
     output[["Transp"]] <- (sub.sipnet.output$fluxestranspiration * 10) / timestep.s  # Transpiration kgW/m2/s
-    output[["SoilMoist"]] <- (sub.sipnet.output$soilWater * 10)  # Soil moisture kgW/m2
+    output[["SoilMoist"]] <- sub.sipnet.output$soilWater   # Soil moisture cm
     output[["SoilMoistFrac"]] <- (sub.sipnet.output$soilWetnessFrac)  # Fractional soil wetness
     output[["SWE"]] <- (sub.sipnet.output$snow * 10)  # SWE
     output[["litter_carbon_content"]] <- sub.sipnet.output$litter * 0.001  ## litter kgC/m2
--- a/models/sipnet/R/model2netcdf.SIPNET.R
+++ b/models/sipnet/R/model2netcdf.SIPNET.R
@@ -196,7 +196,7 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date, 
      output[["Qle"]] <- (sub.sipnet.output$evapotranspiration * 10 * PEcAn.data.atmosphere::get.lv()) / timestep.s  # Qle W/m2
     }
     output[["Transp"]] <- (sub.sipnet.output$fluxestranspiration * 10) / timestep.s  # Transpiration kgW/m2/s
-    output[["SoilMoist"]] <- sub.sipnet.output$soilWater   # Soil moisture cm
+    output[["SoilMoist"]] <- (sub.sipnet.output$soilWater * 10)  # Soil moisture kgW/m2
     output[["SoilMoistFrac"]] <- (sub.sipnet.output$soilWetnessFrac)  # Fractional soil wetness
     output[["SWE"]] <- (sub.sipnet.output$snow * 10)  # SWE
     output[["litter_carbon_content"]] <- sub.sipnet.output$litter * 0.001  ## litter kgC/m2
--- a/models/sipnet/R/read_restart.SIPNET.R
+++ b/models/sipnet/R/read_restart.SIPNET.R
@@ -15,7 +15,7 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   forecast <- list()
   params$restart <-c() #state.vars not in var.names will be added here
   #SIPNET inital states refer to models/sipnet/inst/template.param
-  state.vars <- c("SWE", "SoilMoistFrac", "AbvGrndWood", "TotSoilCarb", "LAI", 
+  state.vars <- c("SWE", "SoilMoist", "SoilMoistFrac", "AbvGrndWood", "TotSoilCarb", "LAI", 
                   "litter_carbon_content", "fine_root_carbon_content", 
                   "coarse_root_carbon_content", "litter_mass_content_of_water")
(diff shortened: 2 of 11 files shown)


models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -196,7 +196,7 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
      output[["Qle"]] <- (sub.sipnet.output$evapotranspiration * 10 * PEcAn.data.atmosphere::get.lv()) / timestep.s  # Qle W/m2
     }
     output[["Transp"]] <- (sub.sipnet.output$fluxestranspiration * 10) / timestep.s  # Transpiration kgW/m2/s
-    output[["SoilMoist"]] <- (sub.sipnet.output$soilWater * 10)  # Soil moisture kgW/m2
+    output[["SoilMoist"]] <- sub.sipnet.output$soilWater   # Soil moisture cm
```
mdietze: Violates PEcAn output standard. Changing the standard requires changing every coupled model

mdietze (commented):
Not clear why the proposed changes are needed or desirable. The supposed purpose is to enable assimilation of volumetric soil moisture, but we can already assimilate soil moisture, and the standard way to accommodate small misalignments between a model and a state variable is through an observation operator, H. Unclear why such an approach could not be used here.  

It is clear that the proposed changes are breaking and non-backwards-compatible and leave stray unused variables in place.

Qianyuxuan:
> Not clear why the proposed changes are needed or desirable. The supposed purpose is to enable assimilation of volumetric soil moisture, but we can already assimilate soil moisture, and the standard way to accommodate small misalignments between a model and a state variable is through an observation operator, H. Unclear why such an approach could not be used here.
> 
> It is clear that the proposed changes are breaking and non-backwards-compatible and leave stray unused variables in place.

@mdietze @DongchenZ In my latest change, I have removed the variable for initial soil water, but only kept “soilWFracInit” in write.comfigs.SIPNET.R. So in this way, we don’t need to add a new variable “soilWaterInit” in SIPNET source code.

models/sipnet/R/model2netcdf.SIPNET.R
Qianyuxuan: Fixed.

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -15,7 +15,7 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
   forecast <- list()
   params$restart <-c() #state.vars not in var.names will be added here
   #SIPNET inital states refer to models/sipnet/inst/template.param
-  state.vars <- c("SWE", "SoilMoistFrac", "AbvGrndWood", "TotSoilCarb", "LAI", 
+  state.vars <- c("SWE", "SoilMoist", "SoilMoistFrac", "AbvGrndWood", "TotSoilCarb", "LAI",
```
mdietze, line 18: Are you sure "SoilMoist" should be added here? Isn't it calculated from "SoilMoistFrac"?

models/sipnet/R/read_restart.SIPNET.R
```diff
@@ -102,6 +102,13 @@ read_restart.SIPNET <- function(outdir, runid, stop.time, settings, var.names, p
     params$restart["litter_mass_content_of_water"] <- ens$litter_mass_content_of_water[last]
   }
   
+  if ("SoilMoist" %in% var.names) {
+    forecast[[length(forecast) + 1]] <- ens$SoilMoist[last]
```
mdietze, line 106: Is read.output already providing the SoilMoist you need? Based on prior discussion I thought it would need to be manually calculated from SoilMoistFrac

models/sipnet/R/write_restart.SIPNET.R
```diff
@@ -89,6 +89,12 @@ write_restart.SIPNET <- function(outdir, runid, start.time, stop.time, settings,
     names(analysis.save[[length(analysis.save)]]) <- c("litter_mass_content_of_water")
   }
   
+  if ("SoilMoist" %in% variables) {
+    analysis.save[[length(analysis.save) + 1]] <- new.state$SoilMoist
```
mdietze, line 93: I'm surprised that write_restart doesn't include a conversion of SoilMoist back to SoilMoistFrac

modules/data.remote/R/SMAP_SMP_prep.R
```diff
@@ -72,7 +72,7 @@ SMAP_SMP_prep <- function(site_info, start_date, end_date, time_points,
   time_points <- time_points[which(lubridate::year(time_points)>=2015)] #filter out any time points that are before 2015
   #initialize SMAP_Output
   SMAP_Output <- matrix(NA, length(site_info$site_id), 2*length(time_points)+1) %>% 
-    `colnames<-`(c("site_id", paste0(time_points, "_SoilMoistFrac"), paste0(time_points, "_SD"))) %>% as.data.frame()#we need: site_id, LAI, std, target time point.
+    `colnames<-`(c("site_id", paste0(time_points, "_SoilMoist"), paste0(time_points, "_SD"))) %>% as.data.frame()#we need: site_id, LAI, std, target time point.
```
mdietze, line 75: I'm surprised to see changes in the SMAP code as I thought you were working with a different product and because here you're changing the output name without changing any of the calculations. Does this imply that SMAP was already column sums not % and we had the units wrong? That would be surprising.

models/sipnet/R/read_restart.SIPNET.R
Qianyuxuan: Yes. The original SIPNET codes provide SoilMoist in the outputs. So I just read the values from the output file and saved it into forecast variable.

models/sipnet/R/read_restart.SIPNET.R
Qianyuxuan: I guess so. The original SIPNET codes have "SoilMoist "outputs and I found it is a little challenging to read the parameter and do the conversion in "read_restart" and "write_restart" if we want it calculated from "SoilMoistFrac".

models/sipnet/R/write_restart.SIPNET.R
Qianyuxuan: There is "write.configs" function called in "write_restart" function to do the conversion of SoilMoist back to SoilMoistFrac.

modules/data.remote/R/SMAP_SMP_prep.R
Qianyuxuan: The new variable "SoilMoist" was for the latest SMAP version 7 L4 data as there is only volumetric soil moisture but no soil wetness index in this dataset. The corresponding "SMAP.csv" was generated from Dongchen's latest codes here: https://github.com/DongchenZ/pecan/commit/193b2d05706cdd1f180fe0bfb4bc4643ee220ee8. So we might want to wait for his PR to merge later.

mdietze (approved):

PR #3520, opened 2025-05-05
version bump

robkooper:
bumps all versions of DESCRIPTION where the R files have changed (see versionbump)

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


models/cable/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.CABLE
 Type: Package
 Title: PEcAn package for integration of the CABLE model
-Version: 1.7.3.9000
+Version: 1.7.3
```
infotroph: Shouldn't versions always move forward?

models/sibcasa/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.SIBCASA
 Type: Package
 Title: PEcAn Package for Integration of the SiBCASA Model
-Version: 0.0.1.9000
+Version: 1.9.0
```
infotroph: This package is definitely not ready for a >1.0 number yet

infotroph (changes requested):
I'm sorry it took me so long to review this!

My concern here is that this would go back to keeping (almost) all package versions equal to each other, which we moved away from intentionally.  I also see that the packages packages not bumped up to 1.9.0 move backwards to versions we already used (e.g. PEcAn.CABLE 1.7.3.9000 -> 1.7.3), which seems undesirable.

Short of hand-choosing the version for each package -- which I'm still willing to do if you want to wait for me, but I'm clearly slow! -- what I'd suggest would be to increment all versions by a fixed step. Maybe if there are no changes in R/ we bump the patch number and if there are changes in R/ (or maybe if there are new entries in NEWS?) we bump the minor version with the conventional zeroing of patch number?  e.g. 1.8.2 -> 1.9.0; 1.7.3 with no code commits -> 1.7.4; 0.0.1 -> 0.1.0

models/cable/DESCRIPTION
robkooper: Except nothing changed in this package since last release

models/sibcasa/DESCRIPTION
robkooper: What version should it be for n that case?

infotroph: I'd say 0.1.0 -- the thing that changed was the package license, which seems to me like a minor-version change not a patch change

PR #3526, opened 2025-05-23
Release/v1.9.0

robkooper:
long over due, release 1.9.0

need following fixes before we can release
- [x] set version to 1.9.0  #3520 
- [x] fix rstudio login #3527 
- [x] #3406


dlebauer (approved):
￼

PR #3534, opened 2025-06-05
[WIP] gSSURGO Enhancements: SOC Integration, enhanced XML Parsing, and Ensemble improvements

divine7022:
## Description
This PR significantly improves the `extract_soil_gssurgo` function for generating soil property ensembles from gSSURGO data. It introduces robust XML parsing, support for soil organic carbon (SOC) calculation, improved uncertainty handling, dynamic depth handling, and fixes for ensemble generation and NetCDF output.

### Key Enhancements:
- Added Fields: Queries `chorizon.om_r` (organic matter) and `chorizon.dbthirdbar_r` (bulk density).
- SOC Calculation:
   - Uses the Van Bemmelen factor `(SOC = OM / 1.724)`.
   - Computes SOC stock:
   ```r
   horizon_thickness_m * (soc_percent / 100) * bulk_density * 10  
  ```
- Simulates `SOC` ensembles via gamma distribution for positive-only values.
- Avoids memory leaks and malformed XML errors by Temp File handling.
- Extracts mukeys from raw XML text if standard parsing fails.
- Automatically extends `depths` vector if observed soil exceeds user-specified layers.
- Uses `pmax/pmin` to ensure valid depth indices, avoiding `subscript out of bounds`
- Valid `sizein` Handling: Coerces ensemble size to `≥1` to prevent `seq_len()` errors.
- Proper mukey_area Initialization: Creates from simulated data instead of filtering undefined variables.
- Uses `min(x, nrow(.x))` to avoid invalid row access.
- Includes `SOC` by removing [1:4] subsetting.
- `extract_soil_gssurgo` now supports spatial sampling using a user-defined **grid** (configurable size and spacing), enabling quantification of within-site spatial variability and improved representation of soil heterogeneity in extracted properties.
- Soil property extraction and aggregation now employ **area-weighted** means across all map unit components and sampled grid cells, ensuring that ensemble statistics reflect the true landscape composition and spatial structure.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


dlebauer:
@divine7022 Thank you for these improvements!!! 

could you please: 
- [x] resolve conflicts in extract_soil_nc.R?
- [x] Add your name to the function authors and CITATION.cff if you haven't already done so
- [x] Update Changelog
- [x] Add tests for new functionality? These don't have to comprehensively cover cornder cases, mostly aimed at identifying if any essential features are broken in the future.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +149,53 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
+    unique_mukeys <- unique(soilprop.new$mukey)
+    mukey_area <- data.frame(
+      mukey = unique_mukeys,
+      Area = rep(1/length(unique_mukeys), length(unique_mukeys))
+    ) 
     mukey_area <- mukey_area %>%
-      dplyr::filter(mukeys %in% simulated.soil.props$mukey) %>%
-      dplyr::mutate(Area=.data$Area/sum(.data$Area))
-    
+  dplyr::filter(mukey %in% unique(simulated.soil.props$mukey)) %>%
```
dlebauer: please fix indentation

modules/data.land/R/extract_soil_nc.R
```diff
@@ -98,13 +127,13 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
   tryCatch({
     # find the soil depth levels based on the depth argument 
     # if soil profile is deeper than what is specified in the argument then I go as deep as the soil profile.
-    if (max(soilprop.new$soil_depth) > max(depths)) depths <- sort (c(depths, max(max(soilprop.new$soil_depth))))
+    if (max(soilprop.new$soil_depth) > max(depths)) depths <- sort (c(depths, max(soilprop.new$soil_depth)))
```
dlebauer:
```suggestion
    if (max(soilprop.new$soil_depth) > max(depths)) {
      depths <- sort (c(depths, max(soilprop.new$soil_depth)))
    }
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -46,49 +46,78 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
-        function(x) x / 100))
-
-  soilprop.new <- soilprop.new[ stats::complete.cases(soilprop.new) , ]
+    dplyr::mutate(
+      dplyr::across(c(dplyr::starts_with("fraction_of"), "soil_depth", "soil_depth_bottom"), 
+                    ~ as.numeric(.) / 100),
+      horizon_thickness_m = soil_depth_bottom - soil_depth,
+      # Van Bemmelen factor conversion: OM to SOC
+      soc_percent = organic_matter_pct / 1.724,  
+      soil_organic_carbon_stock = horizon_thickness_m * (soc_percent / 100) * bulk_density * 10
```
dlebauer: what does the `* 10` represent? For unit conversions please use `units::ud_convert(x, from, to)` i.e. instead of 10, could do `|> units::ud_convert('Mg ha-1', 'kg m-2')`

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +149,53 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
           alpha <- dir.model$alpha
           alpha <- matrix(alpha, nrow= size, ncol=length(alpha), byrow=TRUE )
           simulated.soil <- sirt::dirichlet.simul(alpha)
-          # # using the simulated sand/silt/clay to generate soil ensemble
+          
+          # Simulate SOC uncertainty using Gamma distribution to ensure positive values
+          soc_mean <- mean(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)
+          soc_sd <- stats::sd(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)
```
dlebauer, line 272: Under what conditions would the data have NAs? I generally wouldn't expect them in gSSURGO, and if there are known exceptions it would be better to handle these separately.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +149,53 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
+            simulated_soc <- pmax(stats::rgamma(size, shape=shape, rate=rate), 0)
+          } else {
+            simulated_soc <- rep(soc_mean, size)
+          }
+          
           simulated.soil<-simulated.soil %>%
             as.data.frame %>%
-            dplyr::mutate(DepthL=rep(DepthL.Data[1,6], size),
-                   mukey=rep(DepthL.Data[1,5], size)) %>%
+            dplyr::mutate(DepthL=rep(DepthL.Data[1,12], size),
```
dlebauer: why have these indices changed & is there a way to make this robust under future changes to the query?

dlebauer (changes requested):

modules/data.land/R/extract_soil_nc.R
```diff
@@ -46,49 +46,78 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
     mukeys,
     c("chorizon.sandtotal_r",
       "chorizon.silttotal_r",
       "chorizon.claytotal_r",
-      "chorizon.hzdept_r"))
-
+      "chorizon.hzdept_r",
+      "chorizon.hzdepb_r",  
+      "chorizon.om_r",      
+      "chorizon.dbthirdbar_r"))
```
dlebauer: The PR description says that this is builk density, but in fact it is 

 

> chorizon.dbthirdbar_r: The oven dry weight of the less than 2 mm soil material per unit volume of soil at a water tension of 1/3 bar.


whereas what we want is dbovendry_r: 

> The oven dry weight of the less than 2 mm soil material per unit volume of soil exclusive of the desication cracks, measured on a coated
clod.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +149,53 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
           alpha <- dir.model$alpha
           alpha <- matrix(alpha, nrow= size, ncol=length(alpha), byrow=TRUE )
           simulated.soil <- sirt::dirichlet.simul(alpha)
-          # # using the simulated sand/silt/clay to generate soil ensemble
+          
+          # Simulate SOC uncertainty using Gamma distribution to ensure positive values
+          soc_mean <- mean(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)
+          soc_sd <- stats::sd(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)
+          if (!is.na(soc_sd) && soc_sd > 0) {
```
dlebauer: under what conditions would soc_sd == 0? Variance (and thus SD) of a single value is undefined. 

Also, see Slack thread re: how we want to estimate uncertainty here.

modules/data.land/R/extract_soil_nc.R
infotroph: dbthirdbar is the one most people are thinking of when they say "bulk density": Measure volume at field capacity, then find its oven-dry weight.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -46,49 +46,78 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
-          "soil_depth"),
-        function(x) x / 100))
-
-  soilprop.new <- soilprop.new[ stats::complete.cases(soilprop.new) , ]
+    dplyr::mutate(
+      dplyr::across(c(dplyr::starts_with("fraction_of"), "soil_depth", "soil_depth_bottom"), 
+                    ~ as.numeric(.) / 100),
+      horizon_thickness_m = soil_depth_bottom - soil_depth,
+      # Van Bemmelen factor conversion: OM to SOC
+      soc_percent = organic_matter_pct / 1.724,
```
infotroph: We're certainly in good company if we stick to the Van Bemmelen factor, but https://www.sciencedirect.com/science/article/abs/pii/S0016706110000388 has me convinced 2.0 may be a better assumption

divine7022: @infotroph @dlebauer Thanks for your valuable comments and sorry for delay;
updated code:
- all above comments have been addressed  
- `extract_soil_gssurgo` now supports spatially sampling using a user-defined grid (configurable size and spacing) but by default use the `3x3` grid of 10m cells; enabling quantification of within-site soil heterogeneity and improved capture of spatial variability in soil properties.
- Soil property extraction and aggregation now employ **area-weighted** means across all mapunit components and sampled grid cells, ensuring that ensemble statistics reflect the true landscape composition and spatial structure.
   - Area-weighted aggregation across horizons and map units.
      for each cell queried:
            If a single component exists, the representative value `_r` is used.
            If multiple components exist, an `area-weighted` mean is computed.
   - Ensemble generation using `Dirichlet` and `Gamma based` stochastic sampling, propagating both spatial and compositional uncertainty.

divine7022:
Alternatively we also had an option : for more spatially accurate and high-throughput workflows, we can extract `MUKEYs` and soil properties directly from a local gSSURGO raster (`.tif`) file (e.g, `MapunitRaster_10m.tif`):
- this ensures perfect alignment with the official **gSSURGO** raster grid provided by **NRCS**.
- it is computationally efficient for large areas or bulk processing.
- however, it requires downloading and storing raster files locally, which can significantly increase disk usage.

I'm currently using a grid based sampling approach to retrieve `MUKEYs` and associated soil properties:
- a grid is generated, centered on the target lat/lon.
- for each grid cell center, lat/lon coordinates are calculated.
- and then retrieve the corresponding `MUKEYs` for each location using **WFS** (Web Feature Service) server requests.
- this method enables spatially flexible, sampling and is suitable when local raster data( if we use it in future) are not available or when disk space is limited

modules/data.land/R/soil2netcdf.R
```diff
@@ -81,7 +81,7 @@ soil2netcdf <- function(soil.data, new.file){
   
   ## add data
   for (i in seq_along(ncvar)) {
-    if(is.null(soil.data[[i]])|is.na(soil.data[[i]])) next
+    if(is.null(soil.data[[i]]) || (length(soil.data[[i]]) == 1 && is.na(soil.data[[i]]))) next
```
infotroph: Would `|| all(is.na(soildata[[i]]))` be clearer? But maybe I'm misunderstanding which edge case this catches

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -0,0 +1,343 @@
...
+    size = 2,
+    grid_size = 3,
+    grid_spacing = 10,
+    depths = c(0.15, 0.30)
+  )
+  end_time <- Sys.time()
+  execution_time <- as.numeric(difftime(end_time, start_time, units = "secs"))
+  # Should complete within reasonable time(5 minutes threshold)
+  expect_lt(execution_time, 300, 
+           info = "Function execution time exceeds 5 minutes")
```
infotroph: This is way too long for a unit test -- is there an alternate way to express this that exercises smaller units and completes faster (ideally seconds or less, not minutes)

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -0,0 +1,343 @@
...
+  unlink(tmp_outdir, recursive = TRUE)
+})
+
+test_that("extract_soil_gssurgo performance is reasonable for typical use cases", {
+  skip_on_cran()
+  skip_on_ci()
+  lat <- 40.0
+  lon <- -88.0
+  tmp_outdir <- tempfile("gssurgo_test_")
+  dir.create(tmp_outdir, recursive = TRUE)
```
infotroph: We can let withr handle creating this and cleaning it up for us when the test finishes:

```suggestion
  tmp_outdir <- withr::local_tempdir("gssurgo_test_")
```

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -0,0 +1,343 @@
...
+  end_time <- Sys.time()
+  execution_time <- as.numeric(difftime(end_time, start_time, units = "secs"))
+  # Should complete within reasonable time(5 minutes threshold)
+  expect_lt(execution_time, 300, 
+           info = "Function execution time exceeds 5 minutes")
+  
+  if (!is.null(res)) {
+    expect_true(all(file.exists(unlist(res))))
+  }
+  unlink(tmp_outdir, recursive = TRUE)
```
infotroph: if you switch to local_tempfile() as recommended above, the outdir will be cleaned up on exit with no further action from us
```suggestion
```

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -0,0 +1,343 @@
...
+  )
+  end_time <- Sys.time()
+  execution_time <- as.numeric(difftime(end_time, start_time, units = "secs"))
+  # Should complete within reasonable time(5 minutes threshold)
+  expect_lt(execution_time, 300, 
+           info = "Function execution time exceeds 5 minutes")
+  
+  if (!is.null(res)) {
+    expect_true(all(file.exists(unlist(res))))
+  }
```
infotroph: An `if` in a unit test is a good sign you're not defining the test conditions narrowly enough -- we should know whether we want the result to be null or not from this particular call, and fail if `res` doesn't match that.

modules/data.land/tests/testthat/test-extract_soil_nc.R
infotroph: (Probably obvious, but: I'm only commenting here but you should use this same pattern in all the test blocks that currently have a `dir.create`)

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+#' @param grid_size Size of the spatial sampling grid (default: 3)
+#' @param grid_spacing Spacing between grid cells in meters (default: 10)
 #' @param depths  Standard set of soil depths in m to create the ensemble of soil profiles with.
 #'
 #' @return It returns the address for the generated soil netcdf file
 #'
 #' @importFrom rlang .data
-#'
+#' @importFrom sf st_as_sf st_transform
+#' @importFrom terra rast crds
```
infotroph: Please call these in `pkg::fun()` form instead of importFrom

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+  half_extent <- (grid_size - 1) / 2 * grid_spacing
+  xmin <- coords_proj[1] - half_extent
+  xmax <- coords_proj[1] + half_extent
+  ymin <- coords_proj[2] - half_extent
+  ymax <- coords_proj[2] + half_extent
+  
+  # Create raster template
+  raster_template <- terra::rast(
+    xmin = xmin, xmax = xmax, ymin = ymin, ymax = ymax,
+    resolution = grid_spacing, crs = proj_crs
```
infotroph: This all seems like a really long way around to me - why this rather than define a buffer around the points and then extract cells that fall inside it?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
-  if (length(mukeys) == 0) {
-    PEcAn.logger::logger.error("No mapunit keys were found for this site.")
+  # Transform grid coordinates back to WGS84 for gSSURGO queries
+  grid_sf <- sf::st_as_sf(data.frame(x = grid_coords[, 1], y = grid_coords[, 2]),
+                          coords = c("x", "y"), crs = proj_crs)
+  grid_wgs84 <- sf::st_transform(grid_sf, wgs84_crs)
+  grid_coords_wgs84 <- sf::st_coordinates(grid_wgs84)
+  
+  mukeys_all <- c()
+  for (i in seq_len(nrow(grid_coords_wgs84))) {
```
infotroph, line 74: If I understand correctly this is changing the function from expecting a single location per call to expecting lat and lon to be vectors. 

- Is that correct? Either way it seems worth documenting the expected usage more clearly.
- Does existing usage of the function already handle this iteration externally? No need to add it here if the input will usually/always be a single point.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+      result <- XML::xpathApply(
+        doc = XML::xmlParse(temp_file),
+        path = "//MapUnitKeyList",
+        fun = XML::xmlValue)
+      if (is.list(result)) result[[1]] else result
+    }, error = function(e) {
+      xml_doc <- XML::xmlParse(temp_file)
+      all_text <- XML::xpathSApply(xml_doc, "//text()", XML::xmlValue)
+      mukey_candidates <- all_text[grepl("^[0-9,]+$", all_text)]
+      if (length(mukey_candidates) > 0) mukey_candidates[1] else NULL
```
infotroph: In general: I recommend against this pattern of reparsing input inside the error handling path -- it gets messy fast and makes it hard to reason about program flow. Better to identify a decision point in the process before the possible failure occurs (maybe checking if a `MapUnitKeyList` element exists?) and branch on that via if/else instead.

For this usage specifically: In what cases does the error path get invoked, and do they happen because of a problem in our request or on the server side? One the response has departed from the expected format, is looking for numbers in the whole page really a robust way of recovering from that?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+      handle = curl::new_handle(ssl_verifypeer = FALSE, ssl_verifyhost = FALSE)
+    )
+    
+    # mukey extraction with error recovery
+    mukey_str <- tryCatch({
+      result <- XML::xpathApply(
+        doc = XML::xmlParse(temp_file),
+        path = "//MapUnitKeyList",
+        fun = XML::xmlValue)
+      if (is.list(result)) result[[1]] else result
```
infotroph: What wrapper is being removed here? This looks to me like when the search area contains multiple map units it's going to only return the first of them (which is not what we want), but I didn't test-run before asking this.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+    dplyr::mutate(
+      sandtotal_r = as.numeric(sandtotal_r),
+      silttotal_r = as.numeric(silttotal_r),
+      claytotal_r = as.numeric(claytotal_r),
+      hzdept_r = as.numeric(hzdept_r),
+      hzdepb_r = as.numeric(hzdepb_r),
+      om_r = as.numeric(om_r),
+      dbthirdbar_r = as.numeric(dbthirdbar_r),
+      comppct_r = as.numeric(comppct_r)
+    ) %>%
```
infotroph: `gSSURGO.Query` converts everything to numeric before returning
```suggestion
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
       mukey = "mukey") %>%
-    dplyr::mutate(dplyr::across(
-        c(dplyr::starts_with("fraction_of"),
-          "soil_depth"),
-        function(x) x / 100))
-
-  soilprop.new <- soilprop.new[ stats::complete.cases(soilprop.new) , ]
+    dplyr::mutate(
+      dplyr::across(c(dplyr::starts_with("fraction_of"), "soil_depth", "soil_depth_bottom"), 
+                    ~ as.numeric(.) / 100),
```
infotroph: As above, we already made sure these are numeric. Coercing them multiple times just makes the pipeline harder to read.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+      horizon_thickness_m = soil_depth_bottom - soil_depth,
+      # Van Bemmelen factor conversion: OM to SOC
+      # Using factor of 2.0 based on recent literature (Pribyl, 2010)
+      soc_percent = organic_matter_pct / 2.0,
```
infotroph: Thanks for updating this after my last comments. With apologies for changing my mind, I went digging more and  I now believe the organic matter values that gSSURGO reports are likely to be mostly derived by calculation **from directly measured SOC**, using the classic Van Bemmelen factor of 1.724. See e.g. discussion in section 5.5 of 
https://www.nrcs.usda.gov/sites/default/files/2022-10/SSIR45.pdf, especially Equations 5.5.6 and 5.5.7 on page 249. If I'm interpreting that correctly, that means switching back to a factor of 1.724 will be the more accurate approach.

To get philosophical while I'm here, this is also a reminder of the limitations of any approach to extracting SOC from SSURGO: 

The organic values gSSURGO reports are considered "representative" for a given soil under "undisturbed conditions", and SOC is one of the soil properties that changes the fastest when a soil is disturbed or the land changes use (a good rule of thumb is that a soil used for agriculture has likely lost at least half the SOC it once had). So the gSSURGO OM values are useful for estimating other properties that covary with OM content (water retention capacity, various fertility measures, erosion susceptibility, etc), should only be taken as very broadly indicative of the actual current SOC stock, with uncertainty that's likely much larger than the difference from dividing by 1.7 vs 2.0.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
+      dplyr::across(c(dplyr::starts_with("fraction_of"), "soil_depth", "soil_depth_bottom"), 
+                    ~ as.numeric(.) / 100),
+      horizon_thickness_m = soil_depth_bottom - soil_depth,
+      # Van Bemmelen factor conversion: OM to SOC
+      # Using factor of 2.0 based on recent literature (Pribyl, 2010)
+      soc_percent = organic_matter_pct / 2.0,  
+      soc_kg_m2 = horizon_thickness_m * (soc_percent / 100) * bulk_density,
+      soil_organic_carbon_stock = units::set_units(soc_kg_m2, "kg/m^2") |> 
+                                  units::set_units("Mg/ha")
+    ) %>%
```
infotroph: I don't have an opinion on which pipe to use, but (1) please be consistent within a file, and (2) if using `|>`,  please check for a `Requires: R (>= 4.1)` in the data.land DESCRIPTION and add it if not there yet.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -4,91 +4,191 @@
...
-    purrr::map(function(var) {
-      soilprop.new[, var]
-    }) %>%
-    stats::setNames(names(soilprop.new)[1:4])
+  soil.data.gssurgo <- list(
+    fraction_of_sand_in_soil = as.numeric(soilprop.new$fraction_of_sand_in_soil),
+    fraction_of_silt_in_soil = as.numeric(soilprop.new$fraction_of_silt_in_soil),
+    fraction_of_clay_in_soil = as.numeric(soilprop.new$fraction_of_clay_in_soil),
+    soil_depth = as.numeric(soilprop.new$soil_depth),
+    soil_organic_carbon_stock = as.numeric(soilprop.new$soil_organic_carbon_stock)
```
infotroph: no as.numeric needed here either

modules/data.land/R/extract_soil_nc.R
infotroph: And are there going to be cases where we'd want to keep the texture data for a layer with missing SOC ?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +228,65 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
+          }
+          # Simulate SOC uncertainty using Gamma distribution
+          soc_mean <- mean(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)
+          soc_sd <- stats::sd(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)
+          
+          # Handle edge cases for SOC simulation
+          if (nrow(DepthL.Data) == 1) {
+            simulated_soc <- rep(soc_mean, size)
+          } else if (is.na(soc_sd) || soc_sd == 0) {
+            simulated_soc <- rep(soc_mean, size)
```
infotroph: As commented elsewhere, filling means where measurements are missing is not the right approach here

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +228,65 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
     
     # estimating the proportion of areas for those mukeys which are modeled
+
+    # defining mukey_area
+    unique_mukeys <- unique(soilprop.new$mukey)
+    mukey_area <- data.frame(
+      mukey = unique_mukeys,
+      Area = rep(1/length(unique_mukeys), length(unique_mukeys))
+    ) 
     mukey_area <- mukey_area %>%
```
infotroph: Was this being used undefined before? 😳  Nice catch.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +228,65 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
-        
       }) 
     
     # estimating the proportion of areas for those mukeys which are modeled
+
+    # defining mukey_area
+    unique_mukeys <- unique(soilprop.new$mukey)
+    mukey_area <- data.frame(
+      mukey = unique_mukeys,
+      Area = rep(1/length(unique_mukeys), length(unique_mukeys))
```
infotroph: Can you explain how this works? by my reading the weighted means near line 150 will mean all of these get equal weight here -- which I assume isn't the indention?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +228,65 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
     mukey_area <- mukey_area %>%
-      dplyr::filter(mukeys %in% simulated.soil.props$mukey) %>%
-      dplyr::mutate(Area=.data$Area/sum(.data$Area))
-    
+      dplyr::filter(mukey %in% unique(simulated.soil.props$mukey)) %>%
+      dplyr::mutate(Area=.data$Area/sum(.data$Area, na.rm = TRUE))
     #--- Mixing the depths
     soil.profiles<-simulated.soil.props %>% 
-      split(.$mukey)%>% 
+      split(.$mukey)%>%
```
infotroph:
```suggestion
      split(.$mukey) %>%
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -120,35 +228,65 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
-    
+      dplyr::filter(mukey %in% unique(simulated.soil.props$mukey)) %>%
+      dplyr::mutate(Area=.data$Area/sum(.data$Area, na.rm = TRUE))
     #--- Mixing the depths
     soil.profiles<-simulated.soil.props %>% 
-      split(.$mukey)%>% 
+      split(.$mukey)%>%   
       purrr::map(function(soiltype.sim){
-        sizein <- (mukey_area$Area[ mukey_area$mukey == soiltype.sim$mukey %>% unique()])*size
+        sizein <- (mukey_area$Area[ mukey_area$mukey == unique(soiltype.sim$mukey)[1]])*size
```
infotroph: Boy was the existing version a weird line... I agree it wasn't right as it was, but I wonder if maybe the intention was more like  `mukey_area$mukey %in% unique(soiltype.sim$mukey)`? Does that make sense to you?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -158,13 +296,13 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
           })
       }) %>%
       purrr::flatten()
-
     #- add them to the list of all the ensembles ready to be converted to .nc file
     all.soil.ens<-soil.profiles %>%
       purrr::map(function(SEns){
+        SEns <- SEns[, setdiff(names(SEns), "mukey")]
```
infotroph: This works but `SEns[, names(SEns) != "mukey"]` would be more idiomatic

infotroph (changes requested):

modules/data.land/R/extract_soil_nc.R
dlebauer: After reviewing the paper, I'd also suggest 2. And it appears the factor >2 in ag soils.

image

Paper attached for reference 
[1-s2.0-S0016706110000388-main.pdf](https://github.com/user-attachments/files/21763978/1-s2.0-S0016706110000388-main.pdf)

infotroph: See my [follow-up comment below](https://github.com/PecanProject/pecan/pull/3534#discussion_r2274765532) -- sorry for both breaking the thread and changing my mind after arguing so persuasively 😉 
Short version is I'm fine with 2.0 but suspect these values were in fact converted from SOC using the classic 1.724 factor, so probably best to use that after all.

dlebauer: However, I am going to defer this to a separate discussion of how to handle uncertain constants like this (raised in Slack) 

@divine7022 thanks for all of these updates!

modules/data.land/R/extract_soil_nc.R
dlebauer: Perhaps something like:

```suggestion
point <- terra::vect(cbind(lon, lat), crs = wgs84_crs) |>
  terra::project(proj_crs)

half_extent <- (grid_size - 1) / 2 * grid_spacing
ext <- terra::ext(point)
ext <- ext + c(-half_extent, half_extent, -half_extent, half_extent)

raster_template <- terra::rast(ext, resolution = grid_spacing, crs = proj_crs)
```

?

dlebauer: @infotroph If I'm understanding correctly you are requesting replacing lines 41-50 with something along the lines of: 

```
  bb <- point_proj |>
    sf::st_buffer(dist = half_extent) |>
    sf::st_bbox()

  raster_template <- terra::rast(
    xmin = bb["xmin"], xmax = bb["xmax"],
    ymin = bb["ymin"], ymax = bb["ymax"],
    resolution = grid_spacing, crs = proj_crs
  )
```

But this seems like more of a stylistic than a functional change: it doesn't reduce lines of code or noticeably clarify intent. Am I am missing something?

modules/data.land/R/soil2netcdf.R
divine7022: yeah, I was actually concerned about one edge case that I was trying to fix in the case of Dirichlet model failure, but we are also supposed to handle cause where we get all NA's.
Thanks fixed 👍

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: now completes in less then a minute.

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: fixed 👍

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: done 👍

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: now used a explicit test instead of condition check

modules/data.land/R/extract_soil_nc.R
divine7022: removed

modules/data.land/R/extract_soil_nc.R
divine7022: After a digging into this grid vs buffer, I am convinced to proceed with current grid based approach due few reasons:
The main trade-off is precision vs simplicity - the current method guarantees exact 10m spacing in projected coordinates, while a buffer approach would be less code but with approximate spacing.  However, the grid approach uses multiple simple WFS point queries (`DWithin` filters) that are server-optimized and fault-tolerant, while buffer approach requires a single complex polygon query that's computationally expensive and fails catastrophically.

modules/data.land/R/extract_soil_nc.R
divine7022: The function expects single lat/lon values as input (not vectors). The iteration you're seeing is internal spatial sampling - the function creates a `grid_size × grid_size` sampling grid around the center point to capture local soil variability. Each grid point is then queried separately. However thanks for questioning I've added clearer documentation to make this sampling approach more explicit.

modules/data.land/R/extract_soil_nc.R
divine7022: yeah, right - that pattern was unnecessarily complex. After debugging, I found the main `XPath` approach works reliably, so the error recovery was likely dead code anyway here. 
Replaced it with simpler approach that checks for the expected structure upfront rather than using exceptions for control flow. Now much cleaner

modules/data.land/R/extract_soil_nc.R
divine7022: looking at the gSSURGO API responses during debugging, it turns out they return a single `MapUnitKeyList` element containing comma-separated mukeys (like `<MapUnitKeyList>192018,123456,789012</MapUnitKeyList>`), rather than multiple MapUnitKeyList elements. So the `result[[1]]` was extracting the single element, then we split on commas later to get individual mukeys.
However now uses `getNodeSet()` makes this clearer - we explicitly get the first (and only) `MapUnitKeyList` node, then extract its comma-separated content. This handles same but with much clearer intent.

modules/data.land/R/extract_soil_nc.R
divine7022: removed 👍

modules/data.land/R/extract_soil_nc.R
divine7022: 👍

modules/data.land/R/extract_soil_nc.R
divine7022: Thanks for detailed insights and reference to SSIR45. 
I went deeper into this to ensure data quality
After reviewing the SSIR45 documentation you cited, I agree that to maintain consistency with gSSURGO's internal calculations for avoiding bias in our SOC estimates.
While 
https://www.sciencedirect.com/science/article/abs/pii/S0016706110000388
( abstract: A consideration of the possible variation in organic matter composition predicts a range of factor values between 1.4 and 2.5 )
and other references have demonstrated that conversion factors can range from 1.4-2.5 across different soil types, introducing a different factor than what gSSURGO uses internally, would create a conversion artifact that **increases uncertainty rather than reducing it**.
So i think probably using 1.724 to maintain consistency with gSSURGO's methodology would be more appropriate for now.

modules/data.land/R/extract_soil_nc.R
divine7022: yup, now consistently used `%>%`

modules/data.land/R/extract_soil_nc.R
divine7022: 👍

modules/data.land/R/extract_soil_nc.R
divine7022: now used `NA` instead 👍, now no more deterministic

modules/data.land/R/extract_soil_nc.R
divine7022: 👍

modules/data.land/R/extract_soil_nc.R
divine7022: Both approaches work cuz call to `split(.$mukey)` splits by unique `mukey` values, so each element of the list contains identical mukeys (e.g., `c("191967", "191967", "191967")`). Extracting the first element (`[1]`) or using `unique()` will both return a single value.
then `mukey_area$Area[c(TRUE, FALSE, FALSE)]`
That’s why `[1]` and `%in% `approach behave identically. Using `[1]` is just a bit more explicit.

ex:
```r
simulated.soil.props$mukey = c( 
191967, 191967, 191967, 191967, 191967, 
192111, 192111, 192111, 192111, 192111, 
3332575, 3332575, 3332575, 3332575, 3332575 
# … etc 
)
```
`after split(.$mukey)`
```r
$`191967` 
mukey = c(191967, 191967, 191967, …) 
$`192111` 
mukey = c(192111, 192111, 192111, …) 
$`3332575` 
mukey = c(3332575, 3332575, 3332575, …)
```
If you prefer to use `%in%`, I am happy to update it accordingly.

modules/data.land/R/extract_soil_nc.R
divine7022: changed Thanks!

modules/data.land/R/extract_soil_nc.R
divine7022: thanks for your question!
here it accounts to equal weight(i,e equal areas to all mukey), since we are not capturing the mukey frequency before filtering `mukeys_all` which contains only unique mukeys; when i was testing with the small sampling size and extent, since it was majorly looking to unique mukeys later, Area didn't mattered in calculating ens size(`sizein`);
but now identified how much it matter with large extent and sample size.
So now, before filtering unique mukeys_all, we calculate their frequency across all grid points to assign proportional weights.

```r
    mukey_area <- data.frame(
      mukey = names(mukey_counts),
      Area = as.numeric(mukey_counts) / sum(mukey_counts)
    ) %>%
      dplyr::filter(mukey %in% unique(simulated.soil.props$mukey)) %>%
      dplyr::mutate(Area = Area / sum(Area, na.rm = TRUE))
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -113,61 +229,89 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
+    ) %>%
+      dplyr::filter(mukey %in% unique(simulated.soil.props$mukey)) %>%
+      dplyr::mutate(Area = Area / sum(Area, na.rm = TRUE))
     #--- Mixing the depths
     soil.profiles<-simulated.soil.props %>% 
-      split(.$mukey)%>% 
+      split(.$mukey) %>%   
       purrr::map(function(soiltype.sim){
-        sizein <- (mukey_area$Area[ mukey_area$mukey == soiltype.sim$mukey %>% unique()])*size
+        sizein <- (mukey_area$Area[ mukey_area$mukey == unique(soiltype.sim$mukey)[1]])*size
```
dlebauer: Both the `%in%` and `== unique()[1]` obscure the fact that the `split` on `mukey` enforces the fact that `unique(soiltype.sim$mukey)` will be a single value. I'd suggest:

```suggestion
        sizein <- mukey_area$Area[mukey_area$mukey == unique(soiltype.sim$mukey) * size
```

modules/data.land/R/extract_soil_nc.R
dlebauer: That is a useful suggestion, though I think modeling soc = f(texture) to impute missing values is outside the scope of this PR.

divine7022: If `om_r` is missing then `soc_percent` becomes `NA` -- then `soil_organic_carbon_stock` becomes `NA`.

The `dplyr::filter(stats::complete.cases(.))` filter _removes this entire record_. So if SOC is missing, the texture data for that same horizon is also removed.

However debugging and testing across multiple geographic regions i couldn't find any sites having incomplete records in gSSURGOS, but **Nevada desert location Nevada (36.0, -115.5)** among **37 total records** i found **27 texture records and 27 OM records** -- so this reflected me when texture is missing `OM` is missing too.. ( just based on this only site that i found with missing ) .

Anyways function filters records with only complete case(`dplyr::filter(stats::complete.cases(.))​`) we wouldn't get case where texture is missing and soc present. 
but wrote `na.rm` here just as a standard practice 
`soc_sd <- stats::sd(DepthL.Data$soil_organic_carbon_stock, na.rm = TRUE)`

divine7022:
heads up; After investigating the gSSURGO unit specifications, I've confirmed the correct units from 
https://www.nrcs.usda.gov/sites/default/files/2022-08/SSURGO-Metadata-Tables-and-Columns-Report.pdf

soil_depth = "hzdept_r"   `(cm)`
soil_depth_bottom = "hzdepb_r"  `(cm)`
bulk_density = "dbthirdbar_r" `(g/cm3)`

Updated SOC calculation
```r
      horizon_thickness_cm = soil_depth_bottom - soil_depth,
      # Van Bemmelen factor conversion: OM to SOC
      soc_percent = organic_matter_pct / 1.724,
      # SOC(kg/m2) =  SOC(frac) × bulk density(kg/m3) × depth(m)
      soc_kg_m2 =
        PEcAn.utils::ud_convert(horizon_thickness_cm, "cm", "m") *
        (soc_percent / 100) *
        PEcAn.utils::ud_convert(bulk_density, "g cm-3", "kg m-3"),
      soil_organic_carbon_stock = soc_kg_m2
 ```

File: modules/data.land/R/extract_soil_nc.R (hunk 30 above)
Comment (divine7022) [reply modules/data.land/R/extract_soil_nc.R]: thanks! done 👍

Comment (divine7022) [reply modules/data.land/R/extract_soil_nc.R]: done 👍

**Comment by divine7022** [comment]
@dlebauer @infotroph thanks for all your valuable suggestions
i hope all the concerns have been addressed;
can you merge this ?

File: modules/data.land/R/extract_soil_nc.R [hunk 31]
```diff
@@ -174,13 +174,14 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
-      soc_percent = organic_matter_pct / 1.724,  
-      soc_kg_m2 = horizon_thickness_m * (soc_percent / 100) * bulk_density,
-      soil_organic_carbon_stock = units::set_units(soc_kg_m2, "kg/m^2") %>% 
-                                  units::set_units("Mg/ha")
+      soc_percent = organic_matter_pct / 1.724,
+      # SOC(kg/m2) =  SOC(frac) × bulk density(kg/m3) × depth(m)
+      soil_organic_carbon_stock =
+        PEcAn.utils::ud_convert(horizon_thickness_cm, "cm", "m") *
+        (soc_percent / 100) *
+        PEcAn.utils::ud_convert(bulk_density, "g cm-3", "kg m-3")
```
dlebauer: This calculation doesn't account for the coarse fraction - as [described here](https://github.com/PecanProject/pecan/pull/3455/files#r1972122344). Not accounting for the coarse fraction can lead to significant biases by inflating the calculated stock. 

But if we are going to keep it here, we should use a common helper function like the [soilgrids_carbonstocks](https://github.com/PecanProject/pecan/pull/3455/files#diff-0a54c504b21907e71ee01454452a9de6120dc99de20b510bd0a11060acf98956R83-R92) function in #3455. @AritraDey-Dev could you please submit that as a separate pull request? And rename it something generic? soc2ocs [which would be more consistent with the naming conventions in [metutils.R](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/metutils.R)?

modules/data.land/R/extract_soil_nc.R
dlebauer: @divine7022 were you going to change this? Maybe to 


```suggestion
                    ~ . / 100),
```

?

modules/data.land/R/extract_soil_nc.R
dlebauer: as mentioned elsewhere, I think that the conversion from ocs --> soc should be in a separate helper function https://github.com/PecanProject/pecan/pull/3534#discussion_r2286258610

modules/data.land/R/extract_soil_nc.R
divine7022: @dlebauer Thanks for that catch! now updated the SOC calculation to include coarse fragment correction.
As shown in the **USDA-NRCS Soil Organic Carbon Stock Monitoring** ref
https://www.nrcs.usda.gov/sites/default/files/2024-02/FY24_CEMA_221_Soil%20Carbon%20Stock%20Monitoring_10-2023_0.pdf ( see page 6, point 7), 
correct formula is:
`SOC stock = Bulk Density × SOC content × Thickness × (1 – coarse fragment)`

Now this prevents overestimation of soil carbon stocks due to inert rock fragments that don't contain organic carbon. 
Used helper functions `soc2ocs` and `om2soc`

could you please take a look ?

modules/data.land/R/extract_soil_nc.R
infotroph: I actually liked the xpathApply part of your first version! Given it's expected to be a 1-element list, to me the cleanest approach would be to edit `xpathApply` to `xpathSApply` so that it collapses the result to a vector for us. Then `result` would be a single string, empty if no mapunits were found. 

Your current getNodeSet approach looks like it should work too, so no need to revert. Just seems less clear to me that the "set" is expected to consist of exactly one node.

modules/data.land/R/gSSURGO_Query.R
```diff
@@ -43,12 +43,12 @@ gSSURGO.Query <- function(mukeys,
                <RunQuery xmlns="http://SDMDataAccess.nrcs.usda.gov/Tabular/SDMTabularService.asmx">
                <Query>
                SELECT ',
-                 paste(fixed_fields, collapse = ", "),
-                 paste(qry_fields, collapse = ", "),
+                 paste(c(fixed_fields, qry_fields), collapse = ", "),
```
infotroph, line 73: Ouch, I think this one was my fault (and clearly it's been broken for a while). Nice catch.

modules/data.land/R/gSSURGO_Query.R
```diff
@@ -43,12 +43,12 @@ gSSURGO.Query <- function(mukeys,
...
                  ' from mapunit
                join muaggatt on mapunit.mukey=muaggatt.mukey
                join component on mapunit.mukey=component.mukey
                join chorizon on component.cokey=chorizon.cokey
+               left join chfrags on chorizon.chkey=chfrags.chkey
```
infotroph, line 78: Add chfrags to list of queryable tables on line 14 above (which should probably become its own paragraph)

infotroph: Also be aware this will change the number of rows returned: One per fragment size class per component horizon (was one per component horizon), even if no variables from chfrags are requested in the output. Is that the intended result?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -1,94 +1,205 @@
...
+      "chorizon.hzdept_r",
+      "chorizon.hzdepb_r",  
+      "chorizon.om_r",      
+      "chorizon.dbthirdbar_r",  # bulk density at 1/3 bar (field capacity);which is the standard field capacity bulk density measurement
+      "chfrags.fragvol_r",
+      "component.comppct_r")) 
+  
+  # Area-weighted aggregation by mukey and horizon depth
+  soilprop.weighted <- soilprop %>%
+    dplyr::group_by(mukey, hzdept_r, hzdepb_r) %>%
```
infotroph: I think what you're trying to do here is get a weighted horizon mean for each component in the map unit, but beware that horizon boundary depths vary between components, even inside the same map unit. So this will return somewhere between 1 and n_components sets of horizons per mukey.

modules/data.land/R/extract_soil_nc.R
infotroph: Right, I'm not suggesting we impute missing values. I'm saying complete.cases() is a reasonable QC filter when only retrieving texture, but it may not be the right call when some users might want every SOC number they can get even in cases where texture is NA, while others might want all the textures even if SOC is missing, and so on.

modules/data.land/R/extract_soil_nc.R
infotroph: So this won't work like you want, because what's reported from each grid point are the keys for all mapunits that intersect the grid buffer, with no weighting for their relative abundance inside that radius. I'll say more in a separate thread below.

infotroph:
tl;dr: The proposed sampling will not work and needs a change of approach.

I'm sorry that it took me so long to start reviewing and then so long after that to understand the limitations of your proposed method. Now that I've wrapped my head around it, let me lay out the context:

The first thing to know is that this function is secretly misnamed: gSSURGO is "gridded SSURGO", but the existing gSSURGO.Query function does not perform any raster operations at all. It submits (unprojected) point coordinates and a buffer radius (in meters) to the NRCS WFS server, which looks it up in the _non-gridded_ SSURGO soil map and returns all the polygons that intersect the circular buffer. If you knew this already then great; if you were expecting the sampling process to be operating on a raster map then this is an important difference. Notice especially that circular buffers with radius `grid_spacing` placed `grid_spacing` apart (as you're doing here) produces heavily overlapping queries, so small changes in grid placement can make a given feature be counted in anywhere between 1 and 4 grid points.

The second thing to know is that [I broke the existing code](https://github.com/PecanProject/pecan/pull/2964) three years ago!

The original pre-2022 design was to submit a WFS request returning a GML file that contains all the _full polygons_ of every map unit intersecting the buffer, then parse those into spatial objects and intersect each of them with the buffer to get their they contribute to the sample area. 

In #2964 I apparently misunderstood all of that (though given my comment, apparently my spidey sense did tingle a little) and thought that all the local spatial operations did nothing but extract the mukeys, so I deleted all of them and changed the WFS query to have it return the mukeys directly instead of GML. At some point after that was merged, I noticed that this left `mukey_area` undefined, felt embarrassed but was glad that it would break loudly rather than return incorrect results, and put it on my list to fix eventually. 

I'm really, really glad you've stepped up to fix it! All my comments here are trying to be sure the fix returns valid results. 

I see three options to get to valid results:

* Minimal change from current state of the PR: reduce query radius to  `grid_spacing / 2` or less, so that queries don't overlap, and possibly increase the number of grid points. My gut is you need a _lot_ more than 9 points to get a reliable estimate of variability, but I don't like the prospect of needing >9x as many queries to do something the WFS server can do in one request.
* Closest to original intention: Undo a bunch of my 2022 deletions; go back to requesting a GML result containing the full polygons; process them locally to determine relative area. Note that this will be more complicated than just reverting my commit, because Hamze's code used the `rgdal` package that is now retired.
* Biggest change but biggest long-term improvement, *should be a different PR*: Delegate all the sSSURGO details to the `soilDB` R package, which is very good and is written and maintained by folks who know SSURGO inside-out. 
     - I think the core query would be something like `soilDB::mukey.wcs(list(c(xmin, ymin, xmax, ymax), crs = "EPSG:4326"))`, which returns a list of mukeys for each raster pixel and can then be aggregated the same as your grid-sampled results
     - property retrieval for each desired layer would be something like `soilDB::get_SDA_property(mukeys = keys, property = c("om_r", "dbthirdbar_r", ...), top_depth=layer_top, bottom_depth = layer_bottom, method= "Weighted Average", ...)`.  If we don't like their aggregation methods, we could set `method = "None"` and process the individual component horizons as desired.

infotroph (dismissed):
Posting some inline comments that I queued up before realizing the problems I posted in a separate comment above; some of these may be obsoleted by fixing the grid sampling issues, others will need separate resolution,

modules/data.land/R/extract_soil_nc.R
```diff
@@ -1,94 +1,205 @@
...
-
-  if (length(mukeys) == 0) {
-    PEcAn.logger::logger.error("No mapunit keys were found for this site.")
+  # Transform grid coordinates back to WGS84 for gSSURGO queries
+  grid_sf <- sf::st_as_sf(data.frame(x = grid_coords[, 1], y = grid_coords[, 2]),
+                          coords = c("x", "y"), crs = proj_crs)
+  grid_wgs84 <- sf::st_transform(grid_sf, wgs84_crs)
+  grid_coords_wgs84 <- sf::st_coordinates(grid_wgs84)
+  
+  # Query gSSURGO for each grid point to capture spatial variability
```
dlebauer, line 67:
```suggestion
  # Query gSSURGO for each grid point to capture spatial variability
  buffer_radius <- grid_spacing / 2
  PEcAn.logger::logger.warn(
    "Buffer radius set to grid_spacing/2 to avoid overlap",
    "results may be biased due to lack of area weighting and incomplete spatial coverage."
    )
  
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -1,94 +1,205 @@
...
+      "&VERSION=1.1.0",
+      "&REQUEST=GetFeature&TYPENAME=MapunitPoly",
+      "&FILTER=",
+        "<Filter>",
+          "<DWithin>",
+            "<PropertyName>Geometry</PropertyName>",
+            "<gml:Point>",
+              "<gml:coordinates>", this_lon, ",", this_lat, "</gml:coordinates>",
+            "</gml:Point>",
+            "<Distance%20units=%27m%27>", grid_spacing, "</Distance>",
```
dlebauer:
```suggestion
            "<Distance%20units=%27m%27>", buffer_radius, "</Distance>",
```

dlebauer (dismissed):
I made a pair of suggested changes that get to the MVP requirements to merge as suggested by @infotroph in https://github.com/PecanProject/pecan/pull/3534#issuecomment-3225086247.

Please also update the function docs and comments to reflect this behavior.

I moved the implement of more accurate weighting fix to a separate bug report #3609

modules/data.land/R/extract_soil_nc.R
divine7022: updated docs acknowledging about current limitation

modules/data.land/R/extract_soil_nc.R
divine7022: done. Thanks!

modules/data.land/R/extract_soil_nc.R
```diff
@@ -113,61 +243,89 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
+          
+          # Handle edge cases for SOC simulation
+          if (nrow(DepthL.Data) == 1) {
+            simulated_soc <- rep(NA_real_, size)
+          } else if (is.na(soc_sd) || soc_sd == 0) {
+            simulated_soc <- rep(NA_real_, size)
+          } else {
+            shape <- (soc_mean^2) / (soc_sd^2)
+            rate <- soc_mean / (soc_sd^2)
+            simulated_soc <- pmax(stats::rgamma(size, shape=shape, rate=rate), 0)
```
infotroph: Why the pmax? Gamma deviates are >=0 by definition.

modules/data.land/R/soil_process.R
```diff
@@ -56,9 +56,8 @@ soil_process <- function(settings, input, dbfiles, overwrite = FALSE,run.local=T
     names(newfile) <- rep("path", length(newfile))
 
     if(length(newfile)==0){
-      radiusL <- ifelse(is.null(settings$run$input$soil$radius), 500, as.numeric(settings$run$input$soil$radius))
```
infotroph, line 59: Why are you removing this? Given the sampling radius is a key parameter to tune the sampling to the purpose (e.g. folks working on specific small plots of one field will want it much smaller than folks working on a whole landscape), being able to set sampling radius in the settings file seems important to keep.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -113,61 +243,89 @@ extract_soil_gssurgo<-function(outdir, lat, lon, size=1, radius=500, depths=c(0.
...
-      dplyr::filter(mukeys %in% simulated.soil.props$mukey) %>%
-      dplyr::mutate(Area=.data$Area/sum(.data$Area))
-    
+
+    # defining mukey_area
+    mukey_area <- data.frame(
+      mukey = names(mukey_counts),
+      Area = as.numeric(mukey_counts) / sum(mukey_counts)
+    ) %>%
+      dplyr::filter(mukey %in% unique(simulated.soil.props$mukey)) %>%
```
infotroph: Didn't simulate.soil.props just operate on all the mukeys? When would this filter remove any?

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -0,0 +1,103 @@
+context("extract_soil_gssurgo")
+
+test_that("extract_soil_gssurgo returns valid NetCDF files for valid US coordinates", {
+  skip_on_cran()
+  skip_on_ci()
+  lat <- 40.1164
+  lon <- -88.2434
+  tmp_outdir <- withr::local_tempdir("gssurgo_test_")
+  
+  res <- extract_soil_gssurgo(
```
infotroph, line 10: On my machine this is failing with `unable to find an inherited method for function ‘crs’ for signature ‘x = "crs"’`

modules/data.land/R/soil_process.R
divine7022: can i pass this `radiusL` as  `grid_spacing` either ?
```r
newfile <- extract_soil_gssurgo( ..., grid_spacing = radiusL )
```

modules/data.land/R/extract_soil_nc.R
divine7022: because the dirichlet modeling step can fail for some mukeys due to insufficient or invalid soil texture data. The filter make sure we only calculate area weights for mukeys that successfully simulated soil properties

```r
simulated.soil.props <- soilprop.new.grouped %>%
    .....
    tryCatch({
      # dirichlet modeling
      return(simulated.soil) 
    },
    error = function(e) {
      PEcAn.logger::logger.warn(conditionMessage(e))
      return(NULL)  # <-- returns NULL on failure
    })
  })

File: modules/data.land/tests/testthat/test-extract_soil_nc.R [hunk 41]
```diff
@@ -0,0 +1,103 @@
...
+  # Validate files exist
+  file_paths <- unlist(res)
+  expect_true(all(file.exists(file_paths)))
+  
+  # Validate NetCDF content
+  if (requireNamespace("ncdf4", quietly = TRUE)) {
+    expected_vars <- c("fraction_of_sand_in_soil", "fraction_of_silt_in_soil", 
+                      "fraction_of_clay_in_soil", "soil_organic_carbon_stock")
+    
+    nc <- ncdf4::nc_open(file_paths[2]) # 2nd file because 1st file contains original gSSURGO data
```
infotroph: > 1st file contains original gSSURGO data

Can you rephrase this? I eventually figured out that it means the first ensemble member always uses the reported values without sampling from the estimated distributions, but it took me a while to figure that out.

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -0,0 +1,103 @@
...
+    size = 3,
+    grid_size = 3,
+    grid_spacing = 100,
+    depths = c(0.15, 0.30)
+  )
+  
+  expect_false(is.null(res))
+  
+  expect_type(res, "list")
+  expect_gte(length(res), 1)
```
infotroph: Why >=1 rather than exactly 3?

modules/data.land/R/extract_soil_nc.R
divine7022: though the `purrr::map_df()` drops NULL results from individual attempts, the filter ensures area weighting only includes mukeys that successfully completed the entire data processing and modeling pipeline

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: could you please check the package version might be deprecated ?
packageVersion("sf") 
packageVersion("terra")

```
> packageVersion("sf")
[1] ‘1.0.19’
> packageVersion("terra")
[1] ‘1.8.29’
```

divine7022:
```r
Imports: 
    sf (>= 1.0.0),
    terra (>= 1.8.0)
 ```
 https://cran.r-project.org/web/packages/terra/index.html

Comment (infotroph) [reply modules/data.land/tests/testthat/test-extract_soil_nc.R:10]: > packageVersion("sf")
[1] ‘1.0.20’
> packageVersion("terra")
[1] ‘1.8.42’

File: modules/data.land/R/soil_process.R (hunk 38 above)
Comment (infotroph) [reply modules/data.land/R/soil_process.R:59]: The basic idea of using radiusL to choose grid spacing seems sound. Seems like the meanings of `radius` and `grid_spacing` are different enough to need some conversion math, though -- maybe choose a grid that will cover the same area as a circle with the requested radius?

File: modules/data.land/tests/testthat/test-extract_soil_nc.R (hunk 42 above)
Comment (divine7022) [reply modules/data.land/tests/testthat/test-extract_soil_nc.R]: This ensures that we have at least one soil ensemble in case the modeling part failed
`all.soil.ens <-c(all.soil.ens, list(soil.data.gssurgo))`
failed ensemble generations are removed, potentially reducing the count. The function doesn't guarantee exactly size ensembles due to its area-weighted sampling 
```
sizein <- mukey_area$Area[mukey_area$mukey == unique(soiltype.sim$mukey)] * size
        1:ceiling(sizein)
```
generates ensembles based on area proportion, `ceiling(sizein)` can create more ensembles than requested under good coverage condition

File: modules/data.land/tests/testthat/test-extract_soil_nc.R (hunk 41 above)
Comment (divine7022) [reply modules/data.land/tests/testthat/test-extract_soil_nc.R]: yup, will update comment to make it more clear and explicit

File: modules/data.land/R/extract_soil_nc.R (hunk 37 above)
Comment (divine7022) [reply modules/data.land/R/extract_soil_nc.R]: yup, it is redundant. removed  👍

File: modules/data.land/R/soil_process.R (hunk 38 above)
Comment (divine7022) [reply modules/data.land/R/soil_process.R:59]:
```r
buffer_radius <- ifelse(is.null(settings$run$input$soil$radius), 100, 
                         as.numeric(settings$run$input$soil$radius))
  grid_size <- ifelse(is.null(settings$run$input$soil$grid_size), 3,
                     as.numeric(settings$run$input$soil$grid_size))

  grid_extent <- buffer_radius * sqrt(pi)
  grid_spacing <- grid_extent / (grid_size - 1)
```

does this works ? i am reducing the buffer radius from 500 to 100 , since it takes more time, for computational efficiency, it is not ideal as a default.

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: The `crs = proj_crs to crs = proj_crs$wkt` change should resolve issue, ensures it passed as CRS string. Using `$wkt` passes the complete text specification, ensures proper method dispatch and prevents the `'unable to find inherited method'` error.

modules/data.land/R/soil_utils.R
```diff
@@ -342,4 +342,47 @@ mpot2smoist <- function(mpot,soil_water_potential_at_saturation,soil_hydraulic_b
...
+#'
+#' @param soc_percent soil organic carbon concentration (percent, 0-100)
+#' @param bulk_density bulk density (g/cm3)
+#' @param thickness layer thickness (cm)
+#' @param coarse_fraction coarse fragment volume fraction (0-1, default = 0)
+#' @return organic carbon stock (kg/m2)
+#' @export
+#' @author Akash
+#' @examples
+#' soc2ocs(2.5, 1.3, 30, 0.15)
```
dlebauer, line 357: awesome, thanks for these new functions!

modules/data.land/R/soil_process.R
dlebauer: - I would set min grid_size = 3 to avoid dividing by zero when <=1
- 100m default for gssurgo should be plenty large.

modules/data.land/R/soil_utils.R
divine7022: thanks, resolving the conflicts that newly arised

modules/data.land/R/extract_soil_nc.R
dlebauer: looks great, thanks for creating these functions!

modules/data.land/tests/testthat/test-extract_soil_nc.R
infotroph: >  The function doesn't guarantee exactly size ensembles 

I had not realized this before and it seems very undesirable! Was that the existing behavior or is it newly added?

divine7022: It was an existing behavior. And works as i have explained above

divine7022: works as 
suppose `size=3` and two type covering 70% and 30% of the area respectively:

1st : `ceiling(0.7 × 3)` = 3 ens
2nd : `ceiling(0.3 × 3)` = 1 ens
Total: 4 + 1(reported values) = 5 ens

modules/data.land/R/gSSURGO_Query.R
dlebauer: added chfrags to docs https://github.com/PecanProject/pecan/pull/3534/commits/cf051e246b75290ec26c06e89d8cd83dc360c406

divine7022: thanks! that would be really helpful for the folks to quickly look up units, etc.. here rather then having to going [SSURGO-Metadata-Tables-and-Columns](https://www.nrcs.usda.gov/sites/default/files/2022-08/[SSURGO-Metadata-Tables-and-Columns-Report.pdf](https://www.nrcs.usda.gov/sites/default/files/2022-08/SSURGO-Metadata-Tables-and-Columns-Report.pdf))

modules/data.land/R/soil_process.R
divine7022: done 👍

modules/data.land/R/gSSURGO_Query.R
```diff
@@ -4,15 +4,41 @@
...
+#' | `sandtotal_r` | Total sand (<2 mm fraction)               | %            |
+#' | `silttotal_r` | Total silt (<2 mm fraction)               | %            |
+#' | `claytotal_r` | Total clay (<0.002 mm fraction)           | %            |
+#' | `om_r`        | Organic matter (<2 mm soil)               | %            |
+#' | `hzdept_r`    | Horizon top depth                         | cm           |
+#' | `frag3to10_r` | Rock fragments 3–10 in                    | % (by weight)|
+#' | `dbovendry_r` | Bulk density, oven dry                    | g/cm³        |
+#' | `ph1to1h2o_r` | Soil pH (1:1 H2O)                         | pH (unitless)|
+#' | `cokey`       | Component key (identifier)                | —            |
+#' | `chkey`       | Horizon key (identifier)                  | —            |
```
infotroph:
```suggestion
#' | `chorizon.cec7_r`      | Cation exchange capacity at pH 7          | cmol(+)/kg   |
#' | `chorizon.sandtotal_r` | Total sand (<2 mm fraction)               | %            |
#' | `chorizon.silttotal_r` | Total silt (<2 mm fraction)               | %            |
#' | `chorizon.claytotal_r` | Total clay (<0.002 mm fraction)           | %            |
#' | `chorizon.om_r`        | Organic matter (<2 mm soil)               | %            |
#' | `chorizon.hzdept_r`    | Horizon top depth                         | cm           |
#' | `chfrags.fragvol_r` | Rock fragments                            | % (by volume)|
#' | `chorizon.dbthirdbar_r` | Bulk density at field capacity          | g/cm³        |
#' | `chorizon.ph1to1h2o_r` | Soil pH (1:1 H2O)                         | pH (unitless)|
#' | `chorizon.cokey`       | Component key (identifier)                | —            |
#' | `chorizon.chkey`       | Horizon key (identifier)                  | —            |
```

divine7022: updated thanks!

modules/data.land/R/gSSURGO_Query.R
infotroph: Thanks for the documentation; the table of common variables was a good call.

The number-of-rows issue still stands, though: the chfrags table reports one row from each of between 1 and 10 fragment size classes per horizon, so these need to be summed to get expected total coarse fraction. in `extract_soil_nc` you currently take the weighted mean of these, so e.g. a soil with 30% 2-20 mm fragments and 5% 20-100 mm fragments would get reported as 17.5% coarse when it ought to be 35%.

As an extra complication, fragvol_r _can_ sum to greater than 100%, though it's rare (48 of ~2.6M horizons in SSURGO). For this purpose it's probably enough to just do min(sum(fragvol_r), 100).

modules/data.land/tests/testthat/test-extract_soil_nc.R
infotroph: Seeing the math makes me more rather than less sure this behavior is undesirable. Note especially that this will give at least 1 ensemble member for every map unit, even if n map units >>  size and some of them are very rare.

I won't make you change the sizein calculation since it was existing code and less of an issue at typical ensemble sizes (dozens to hundreds, so the rounding doesn't change totals as much). But more relevant to the line I'm commenting on, I still recommend testing for an exact number instead of a greater-than:

1. For unit testing purposes where we control the inputs, we should be able to predict and test against the number of ensemble members we'll get from _this specific call_, even if it might differ in the general case.
2. A behavior that is surprising to the user is one that is more rather than less important to test carefully.

modules/data.land/R/extract_soil_nc.R
infotroph: Closing this thread because what's here works and I think the next refactor should completely drop the grid sampling approach, but for the record I was thinking all of lines 43 - 59 could be reduced to basically `data.frame(lon=lon, lat=lat) |> terra::vect(crs = "EPSG:4326") |> terra::buffer(radius) |> terra::rast(nrow = grid_size, ncol = grid_size) |> terra::crds()`

modules/data.land/R/extract_soil_nc.R
infotroph: First-order correction applied by avoiding overlapping queries. This is still going to be a crude count (especially at the default of 9 grid points) that will tend to overcount rare map units (by not having enough cells that count them as absent)

modules/data.land/R/gSSURGO_Query.R
divine7022: hey @infotroph thanks for catching this behavior!

```r
  soilprop.weighted <- soilprop %>%
    dplyr::group_by(cokey, hzdept_r, hzdepb_r) %>%
    dplyr::mutate(fragvol_r = pmin(sum(fragvol_r, na.rm = TRUE), 100)) %>%
    dplyr::slice(1) %>% 
    dplyr::ungroup() %>%
    dplyr::group_by(mukey, hzdept_r, hzdepb_r) %>%
    dplyr::summarise(
      sandtotal_r = weighted.mean(sandtotal_r, comppct_r, na.rm = TRUE),
      silttotal_r = weighted.mean(silttotal_r, comppct_r, na.rm = TRUE),
      claytotal_r = weighted.mean(claytotal_r, comppct_r, na.rm = TRUE),
      om_r = weighted.mean(om_r, comppct_r, na.rm = TRUE),
      dbthirdbar_r = weighted.mean(dbthirdbar_r, comppct_r, na.rm = TRUE),
      fragvol_r = weighted.mean(fragvol_r, comppct_r, na.rm = TRUE),
      .groups = "drop"
    )
 ```
This now properly sums fragment volumes within each horizon before component weighting, so your 30% + 5% example would correctly give 35% rather than 17.5%.
Does this approach addresses your concern ?

Comment (infotroph) [reply modules/data.land/R/gSSURGO_Query.R:78]: Yes, that looks right to me! For extra clarity to the next reader (AKA me after I forget this conversation 😉 ) you could 
* change `pmin` to `min` -- the result is identical here since both inputs have length 1, but to me makes it clearer we expect a single output value
* maybe change `slice(1)` to `distinct()` to emphasize that the rows we're dropping are duplicates? I'm not sure if that's clearer to everyone or just me
* add a comment explaining the reason for the operation -- maybe something like 
```
dplyr::group_by(cokey, hzdept_r, hzdepb_r) %>%
# Each horizon has as many rows as fragment size classes.
# Need to sum frags & remove duplicates of the other variables
dplyr::mutate(fragvol_r = min(sum(fragvol_r, na.rm = TRUE), 100)) %>%
...
```

divine7022: Thanks for the clarity suggestions! Added comments for future-me (and anyone else)

modules/data.land/R/extract_soil_nc.R
infotroph: Ah, gotcha. Conceptually I think it'd be cleaner to remove unusable components once (either right after retrieval or at the end after letting NAs accumulate) rather than dropping pieces out as they fail at each step, but that can be a future refactor.

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: updated test to expect the exact size number. 👍

dlebauer (approved):
Great work, we made it!

modules/data.land/R/extract_soil_nc.R
divine7022: Currently separates components with different horizon structures into distinct groups.
But this can be efficiently addressed in next iteration using **SoilDB's** `get_SDA_property` function with `method = "Weighted Average"`, which implements the two-step SSURGO aggregation protocol -- depth weighted horizon aggregation within components, followed by area-weighted component aggregation within map units.

PR #3535, opened 2025-06-06
joint ensemble sampling mechanism to the PEcAn workflow

blesson-07:
Summary
This pull request introduces a joint ensemble sampling mechanism to the PEcAn workflow. The change ensures that multi‑site ensemble runs use a consistent set of input samples (parameters, meteorology, etc.) across all sites, thereby eliminating unwanted site‑to‑site variability caused by independent sampling. As a beneficial side effect, the workflow now performs input sampling only once, significantly reducing runtime for large multi‑site analyses.

Motivation
Previously, when running ensembles across multiple sites, each site independently generated its own input samples. This approach could introduce confounding variation between sites, making it difficult to attribute differences in model outputs to site characteristics versus input uncertainty. To enable robust inter‑site comparisons and combined analyses, a method that guarantees identical inputs across sites is required.

Key Changes
Centralised sampling design: Added a new function generate_joint_ensemble_design() that creates the master set of input samples (parameter indices, meteorological realisations, etc.) once, at the start of the workflow.

Refactored core modules:

get.parameter.samples.R now returns sample indices instead of full samples, allowing them to be shared.

runModule.run.write.configs.R calls the joint design function and passes the shared indices to site‑specific routines.

run.write.configs.R and ensemble.R are updated to accept and use these shared indices, removing per‑site sampling code.

Sensitivity analysis compatibility: Verified that the changes do not affect sensitivity analyses, which rely on structured parameter variations.

Performance Improvement
By generating input samples only once, rather than once per site, the workflow eliminates redundant calculations. This optimisation reduces overall runtime, with the benefit becoming more pronounced as the number of sites or the complexity of the sampling (e.g., large ensembles, high‑dimensional parameter spaces) increases.

Testing
Successfully tested with ensemble runs to confirm consistent inputs across sites.

Verified that sensitivity analysis workflows remain unchanged.

All continuous integration checks have passed.


modules/uncertainty/R/ensemble.R
```diff
@@ -169,18 +170,23 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
+          sampled.indices[[pft.i]][, trait.i] <- same.i
         }else{
+          # Extract original trait values
+          trait.values <- pft.samples[[pft.i]][[trait.i]]
+          sampled.values <- stats::quantile(trait.values, random.samples[, col.i])
+
           ensemble.samples[[pft.i]][, trait.i] <- stats::quantile(pft.samples[[pft.i]][[trait.i]],
                                                                   random.samples[, col.i])
-        }
+          sampled.indices[[pft.i]][, trait.i] <- sapply(sampled.values, function(val) {which.min(abs(trait.values - val)) })
```
mdietze, line 185: Totally not your fault, but looking at your solution to this problem makes me realize that I have concerns about the way that we're generating random.samples. I think the logic we're following made sense when each parameter's estimates were being generated independently, but no longer makes sense now that the norm is to generate parameter estimates via Bayesian methods that produce joint parameter samples.

How to do this better would be simplest to implement for the case of a "random" sample, which we want to add as a new `method` and make the new default. In that case we use `sample(ensemble.size,nrow(pft.samples[[pft.i]]), replace=TRUE)` to generate the row indices that we use to generate ensemble.samples.

For the existing methods, I'd appreciate input from others on how best to generate correlated samples using the other methods (halton, sobol, lhc, etc). I'm thinking something like a Cholesky Decomposition (e.g. how a multivariate normal is generated from a univariate normal) so that each new variable sampled is done conditional on the previous variables

modules/uncertainty/R/ensemble.R
```diff
@@ -169,18 +170,23 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
         col.i <- col.i + 1
         if(names(pft.samples[[pft.i]])[trait.i] %in% param.names[[pft.i]]){ # keeping samples
           ensemble.samples[[pft.i]][, trait.i] <- pft.samples[[pft.i]][[trait.i]][same.i]
+          sampled.indices[[pft.i]][, trait.i] <- same.i
         }else{
+          # Extract original trait values
+          trait.values <- pft.samples[[pft.i]][[trait.i]]
+          sampled.values <- stats::quantile(trait.values, random.samples[, col.i])
+
           ensemble.samples[[pft.i]][, trait.i] <- stats::quantile(pft.samples[[pft.i]][[trait.i]],
```
mdietze, line 183: If we're returning the ensemble.samples matrices, and those matrices are what we're using to drive the ensemble runs, then the indices we want are just in order (i.e. ensemble member 1 is row 1, ensemble member 2 is row 2, etc). So for parameters (and parameters alone) the sampled.indices are 1:ensemble.size. 

When there's more than 1 PFT we may need to think carefully about how to generate the ensemble samples to make sure we're doing the correct joint sampling -- I suspect what we're doing now (generate one random.samples, use it for all PFTs) is correct (though see the note below about needing to update how we generate correlated samples)

base/workflow/R/run.write.configs.R
```diff
@@ -22,9 +22,10 @@
 #' @export
 #'
 #' @author David LeBauer, Shawn Serbin, Ryan Kelly, Mike Dietze
-run.write.configs <- function(settings, write = TRUE, ens.sample.method = "uniform", 
+run.write.configs <- function(settings, write = TRUE,  
                               posterior.files = rep(NA, length(settings$pfts)), 
-                              overwrite = TRUE) {
+                              overwrite = TRUE,
+                              input_design) {
```
mdietze: a couple things, one small, one big

small: it's generally best practice in R to put required parameters before those with defaults

BIG: is this a required parameter? or do we want to retain the capability for run.write.configs to generate designs? If input_design = NULL then what design to you generate? My gut says to make this required, and that the design of run.write.configs will ultimately be a lot simpler and cleaner for the separation of design generation from config writing.

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -7,6 +7,12 @@
 #' @export
 runModule.run.write.configs <- function(settings, overwrite = TRUE) {
 
+ensemble_size <- settings[[1]]$ensemble$size
```
mdietze:
```suggestion
ensemble_size <- settings$ensemble$size
```

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -7,6 +7,12 @@
...
+input_design <- PEcAn.uncertainty::generate_joint_ensemble_design(settings=settings[[1]],ensemble_size=ensemble_size)
```
mdietze: are you sure the [[1]] will work as expected if settings is just a single site, not a MultiSettings?

modules/uncertainty/R/ensemble.R
```diff
@@ -149,44 +149,44 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
         } else if (method == "lhc") {
-          same.i <- round(c(PEcAn.emulator::lhc(t(matrix(0:1, ncol = 1, nrow = 2)), ensemble.size) * length(pft.samples[[pft.i]][[1]])))
+          same.i <- floor(c(PEcAn.emulator::lhc(t(matrix(0:1, ncol = 1, nrow = 2)), ensemble.size) * length(pft.samples[[pft.i]][[1]])))+1
         } else if (method == "uniform") {
           same.i <- sample.int(length(pft.samples[[pft.i]][[1]]), ensemble.size)
-        } else {
-          PEcAn.logger::logger.info("Method ", method, " has not been implemented yet, using uniform random sampling")
-          # uniform random
-          same.i <- sample.int(length(pft.samples[[pft.i]][[1]]), ensemble.size)
+        } else if (method == "random") {
```
mdietze, line 162: I think there needs to be a final "else" after this chunk so the code doesn't keep running if the method prescribed doesn't match any of the methods we support. It probably makes sense to end with an error rather than assume random.

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -1,6 +1,49 @@
-generate_joint_ensemble_design <- function() {
- inputsamples<-input.ens.gen()
- paramsample<-get.parameter.samples()
+generate_joint_ensemble_design <- function(settings, ensemble_size, 
+                                           posterior.files = rep(NA, length(settings$pfts)), 
+                                           ens.sample.method = "uniform") {
```
mdietze: Does this have to be an argument? can't this come from the settings like it it is for input.ens.gen?

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -201,10 +201,10 @@ get.parameter.samples <- function(settings,
         quantiles = 0.5
       )
     } else if (settings$ensemble$size > 1) {
-      
+      ensemble.method<-"random"
       ## subset the trait.samples to ensemble size using Halton sequence
       ensemble.samples <- get.ensemble.samples(settings$ensemble$size, trait.samples, 
-                                               env.samples, ens.sample.method, param.names)
+                                               env.samples,ensemble.method, param.names)
```
mdietze: This bit of code shouldn't be changed -- we shouldn't hard code the sampling method using a new variable. It  should still (ultimately) come from the settings

base/workflow/R/run.write.configs.R
blesson-07: Thanks! I agree I've updated the function to stop with an error if input_design is NULL.

base/workflow/R/run.write.configs.R
```diff
@@ -22,10 +22,14 @@
 #' @export
 #'
 #' @author David LeBauer, Shawn Serbin, Ryan Kelly, Mike Dietze
-run.write.configs <- function(settings, write = TRUE, ens.sample.method = "uniform", 
+run.write.configs <- function(settings,input_design,write = TRUE,
```
infotroph: Nit:  Tidyverse style calls for a space after every comma
```suggestion
run.write.configs <- function(settings, input_design, write = TRUE,  
```

base/workflow/R/run.write.configs.R
```diff
@@ -22,10 +22,14 @@
...
-run.write.configs <- function(settings, write = TRUE, ens.sample.method = "uniform", 
+run.write.configs <- function(settings,input_design,write = TRUE,  
                               posterior.files = rep(NA, length(settings$pfts)), 
                               overwrite = TRUE) {
-  ## Skip database connection if settings$database is NULL or write is False
+  
+  if(is.null(input_design)) {
+    stop("`input_design` is required and must not be NULL.")
+     }
+     ## Skip database connection if settings$database is NULL or write is False
```
```suggestion
    stop("`input_design` is required and must not be NULL.")
  }

  ## Skip database connection if settings$database is NULL or write is False
```

base/workflow/R/run.write.configs.R
```diff
@@ -93,13 +97,22 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
...
+    ensemble.samples <- list()
+    for (pft in names(trait.samples)) {
+        pft_traits <- trait.samples[[pft]]
+        ensemble.samples[[pft]] <- as.data.frame(
+        lapply(names(pft_traits), function(trait) {
+          pft_traits[[trait]][trait_sample_indices]
+        })
+  )
+  names(ensemble.samples[[pft]]) <- names(pft_traits)
+}
```
infotroph: I had a hard time following the nesting here -- please keep closing brackets and parens indented the same amount as the line that opened them

```suggestion
    for (pft in names(trait.samples)) {
        pft_traits <- trait.samples[[pft]]
        ensemble.samples[[pft]] <- as.data.frame(
          lapply(
            names(pft_traits),
            function(trait) pft_traits[[trait]][trait_sample_indices]
          )
        )
      names(ensemble.samples[[pft]]) <- names(pft_traits)
    }
```

infotroph: Also: What you wrote here works, but if I'm following right you could also simplify this block to

```
    for (pft in names(trait.samples)) {
        pft_traits <- trait.samples[[pft]]
        ensemble.samples[[pft]] <- purrr::map_dfc(
            pft_traits,
            \(trait) trait[trait_sample_indices]
        )
    }
```

base/workflow/R/run.write.configs.R
```diff
@@ -167,7 +180,8 @@ run.write.configs <- function(settings, write = TRUE, ens.sample.method = "unifo
                                                           ensemble.samples = ensemble.samples, 
                                                           settings = settings,
                                                           model = model, 
-                                                          write.to.db = write)
+                                                          write.to.db = write,
+                                                          input_design= input_design)
```
infotroph: Nit 1: spaces on both sides of the equal sign
Nit 2: function calls are a little easier to read if the arguments are written in the same order as the formal signature (R doesn't care at all, this is just for the humans)
```suggestion
                                                          input_design = input_design,
                                                          write.to.db = write)
```

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -5,14 +5,19 @@
 #' @return A modified settings object, invisibly
 #' @importFrom dplyr %>%
 #' @export
-runModule.run.write.configs <- function(settings, overwrite = TRUE) {
+runModule.run.write.configs <- function(settings, overwrite = TRUE,input_design=NULL) {
```
infotroph: spacing 
```suggestion
runModule.run.write.configs <- function(settings, overwrite = TRUE, input_design = NULL) {
```

(I won't comment on the rest of these individually, but please use the same spacing throughout the PR)

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -5,14 +5,19 @@
...
+
 
   if (PEcAn.settings::is.MultiSettings(settings)) {
     if (overwrite && file.exists(file.path(settings$rundir, "runs.txt"))) {
       PEcAn.logger::logger.warn("Existing runs.txt file will be removed.")
       unlink(file.path(settings$rundir, "runs.txt"))
     }
-    return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = FALSE))
+    if(is.null(input_design)){
```
infotroph: Tidyverse style calls for spaces around the parens of an if statement
(This is one we're very inconsistent about in the existing codebase! But it's still worth improving)

```suggestion
    if (is.null(input_design)) {
```

modules/uncertainty/R/ensemble.R
```diff
@@ -80,11 +80,11 @@ read.ensemble.output <- function(ensemble.size, pecandir, outdir, start.year, en
 ##' @export
 ##' @author David LeBauer, Istem Fer
 get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples, 
-                                 method = "uniform", param.names = NULL, ...) {
+                                 method = "random", param.names = NULL, ...) {
```
infotroph: I assume you and @mdietze already discussed and made a considered decision about this, so I'll only add that this feels like a big enough change in behavior that it's worth making sure it's well-advertised -- there should be a NEWS entry that calls this out specifically as a change in defaults.

modules/uncertainty/R/ensemble.R
```diff
@@ -80,11 +80,11 @@ read.ensemble.output <- function(ensemble.size, pecandir, outdir, start.year, en
...
   
   if (is.null(method)) {
     PEcAn.logger::logger.info("No sampling method supplied, defaulting to uniform random sampling")
```
infotroph: Two questions:
1. Since `method` has a default value, it will only be null if someone explicitly passes `method = NULL`. Under what circumstances will that happen, and will users who do it be expecting/wanting that NULL to result in the same behavior as if they had left it unspecified?
2. Is "uniform random sampling" still the right description to use here?

modules/uncertainty/R/ensemble.R
```diff
@@ -314,16 +322,19 @@ for (input_tag in names(settings$run$inputs)) {
...
+      samples <- list()
+      input_tags <- names(settings$run$inputs)
+
+      for (input_tag in input_tags) {
+           if (input_tag %in% colnames(input_design)) {
+                  input_paths <- settings$run$inputs[[input_tag]]$path
+                  input_indices <- input_design[[input_tag]]
+
+                 samples[[input_tag]] <- list(
+                 samples = lapply(input_indices, function(idx) input_paths[[idx]])
```
```suggestion
                   samples = lapply(input_indices, function(idx) input_paths[[idx]])
```

modules/uncertainty/R/generate_joint_ensemble_design.R
infotroph: I have the same question! I see this comment is marked as resolved, but can't tell what the resolution was.

infotroph (commented):
Thanks for working on  this! I commented on some comparatively minor style points (consistent spacing, etc) and before merging I would like to see more documentation of the changes -- definitely in the form of  changelog/NEWS entries that tell users how behavior has changed, and possibly also in the long-form manual.

modules/uncertainty/R/ensemble.R
blesson-07: Currently, if NULL is explicitly passed, the function logs a message and defaults to "random" sampling. Would you prefer that we change this behavior to throw an error instead, requiring users to explicitly provide a valid  sampling method?

mdietze: @blesson07asd, what @infotroph is saying is that this check, as defined, seems unnecessary since someone would have to deliberately write in "method=NULL" for this case to ever be invoked, which seems exceedingly unlikely. Much more likely would be people passing in methods that don't exactly match the subset of valid methods. I think it's worth having a check for that but not for NULL.

If this check is removed, the message is also removed, which renders the second question moot, but if you DO retain a similar message you should drop "uniform" from the "random sample"

docker/depends/run_pecan_depends.sh.e9269185
mdietze, line 1: This file and the other .e and .o files appear to be build logs, and should not have been committed. Please remove

docker/depends/run_pecan_depends.sh
mdietze, line 1: This is a personal helper script that is specific to your installation and which contains lots of local paths. Please remove

base/workflow/R/run.write.configs.R
```diff
@@ -22,9 +22,14 @@
 #' @export
 #'
 #' @author David LeBauer, Shawn Serbin, Ryan Kelly, Mike Dietze
-run.write.configs <- function(settings, write = TRUE, ens.sample.method = "uniform", 
+run.write.configs <- function(settings, input_design, write = TRUE,
```
mdietze, line 27: new argument `input_design` needs to be documented in above with a new `@param` and then the documentation needs to be recompiled

base/workflow/R/run.write.configs.R
```diff
@@ -22,9 +22,14 @@
 #' @export
 #'
 #' @author David LeBauer, Shawn Serbin, Ryan Kelly, Mike Dietze
-run.write.configs <- function(settings, write = TRUE, ens.sample.method = "uniform", 
+run.write.configs <- function(settings, input_design, write = TRUE,  
                               posterior.files = rep(NA, length(settings$pfts)), 
                               overwrite = TRUE) {
+  
+  if(is.null(input_design)) {
```
mdietze: Unless you default input_design to NULL I don't think this check will ever be true. It's exceedingly unlikely that anyone will manually pass input_design = NULL into the function. As such I think this check can be dropped

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -5,14 +5,19 @@
 #' @return A modified settings object, invisibly
 #' @importFrom dplyr %>%
 #' @export
-runModule.run.write.configs <- function(settings, overwrite = TRUE) {
+runModule.run.write.configs <- function(settings, overwrite = TRUE, input_design = NULL) {
```
mdietze: As above, input_design needs to be documented

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -5,14 +5,19 @@
...
 
   if (PEcAn.settings::is.MultiSettings(settings)) {
     if (overwrite && file.exists(file.path(settings$rundir, "runs.txt"))) {
       PEcAn.logger::logger.warn("Existing runs.txt file will be removed.")
       unlink(file.path(settings$rundir, "runs.txt"))
     }
-    return(PEcAn.settings::papply(settings, runModule.run.write.configs, overwrite = FALSE))
+    if (is.null(input_design)) {
+       ensemble_size <- settings[[1]]$ensemble$size
+       input_design <- PEcAn.uncertainty::generate_joint_ensemble_design(settings=settings[[1]],ensemble_size=ensemble_size)
```
mdietze: This may be more clear later, but why does this function specifically take just the first site within the settings as an argument rather than the overall settings object? None of the ensemble or sensitivity analysis configuration information is in the site-specific part of the settings

modules/uncertainty/R/ensemble.R
```diff
@@ -139,48 +140,58 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
...
-        } else {
-          PEcAn.logger::logger.info("Method ", method, " has not been implemented yet, using uniform random sampling")
-          # uniform random
-          same.i <- sample.int(length(pft.samples[[pft.i]][[1]]), ensemble.size)
+        } else if (method == "random") {
+            PEcAn.logger::logger.info("Using random row sampling for MCMC draws")
+           same.i <- sample(nrow(pft.samples[[pft.i]][[1]]), ensemble.size, replace = TRUE)
+        }
+        else {
+            stop("Sampling method ", method, "is not recognized")
```
mdietze: Replace with logger.error or logger.severe

modules/uncertainty/R/ensemble.R
mdietze: @blesson07asd could you explain what's been done to address this comment?

modules/uncertainty/R/ensemble.R
```diff
@@ -211,7 +222,7 @@ get.ensemble.samples <- function(ensemble.size, pft.samples, env.samples,
 ##' @importFrom rlang .data
 ##' @export
 ##' @author David LeBauer, Carl Davidson, Hamze Dokoohaki
-write.ensemble.configs <- function(defaults, ensemble.samples, settings, model, 
+write.ensemble.configs <- function(defaults, ensemble.samples, settings, model,input_design,
```
mdietze: New argument needs documenting

web/workflow.R
```diff
@@ -215,4 +215,4 @@ if (PEcAn.utils::status.check("FINISHED") == 0) {
 }
 
 db.print.connections()
-print("---------- PEcAn Workflow Complete ----------")
+print("---------- PEcAn Workflow Complete ----------")
```
mdietze, line 218: workflow.R hasn't been updated for the new functions and function arguments, and thus will fail to run. This is probably why your SIPNET test is failing.

mdietze (commented):
In addition to the ensemble test run you've performed, it would be useful to perform a sensitivity analysis test run to verify that the way we've changed the ensemble sampling hasn't broken the sensitivity analysis code

modules/uncertainty/R/ensemble.R
blesson-07: Adding “random” as the default and sampling parameters by full posterior row indices per PFT is implemented in the shown code and addresses the correlation concern.

mdietze (approved):

PR #3540, opened 2025-06-10
Update the NASA_DAAC_download function to write the setup credentials.

DongchenZ:
Previously, the `NASA_DAAC_download` function set the credentials into the environment as default. For some users, this approach may fail to correctly set up the credentials, resulting in an HTTP 404 error when attempting to download the files. Therefore, I updated the function so the credentials will be set correctly anywhere.

The new version accepts downloading the `nc` files.

The user can now input multiple bands and download images given vector of band names.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.remote/R/NASA_DAAC_download.R
```diff
@@ -11,9 +11,9 @@
 #'   "yyyy-mm-dd".
 #' @param outdir Character: path of the directory in which to save the
 #'   downloaded files. Default is the current work directory(getwd()).
-#' @param band Character: the band name of data to be requested.
+#' @param band Character: the band name (or vector of band names) of data to be requested.
 #' @param credential.folder Character: physical path to the folder that contains 
-#' the credential file. The default is NULL.
+#' the credential file. The default is the current working directory.
```
infotroph: I think we've discussed this before but I forget details, so apologies if this is a repeat, but: I'm strongly convinced the current working directory is _not_ the right default location. It makes it too easy to leave plaintext credentials lying around in directories that are likely to be archived and shared.

I suggest defaulting to `~/.netrc` with user consent before writing (which I think you're doing already) and with care not to overwrite existing contents (see below)

modules/data.remote/R/NASA_DAAC_download.R
```diff
@@ -298,30 +305,34 @@ NASA_CMR_finder <- function(doi) {
...
-  if (replace | nchar(Sys.getenv("ed_un")) == 0 | nchar(Sys.getenv("ed_un")) == 0) {
-    Sys.setenv(ed_un = sprintf(
-      getPass::getPass(msg = "Enter NASA Earthdata Login Username \n (or create an account at urs.earthdata.nasa.gov) :")
-    ), 
-    ed_pw = sprintf(
-      getPass::getPass(msg = "Enter NASA Earthdata Login Password:")
-    ))
+getnetrc <- function (dl_dir) {
+  netrc <- file.path(dl_dir, ".netrc")
+  netrc <- gsub("~", Sys.getenv("HOME"), netrc)
```
infotroph: 1. I suggest allowing arbitrary filenames in case users want separate netrcs per host/profile (I guess this means renaming `credential_folder` to `credential_path` above too)
2. technically only a leading tilde should expand to HOME; path.expand knows the details so we don't have to.
```suggestion
getnetrc <- function (dl_path) {
  netrc <- path.expand(dl_path)
```

modules/data.remote/R/NASA_DAAC_download.R
```diff
@@ -298,30 +305,34 @@ NASA_CMR_finder <- function(doi) {
...
+  if (file.exists(netrc) == FALSE ||
+      any(grepl("urs.earthdata.nasa.gov",
+                readLines(netrc))) == FALSE) {
+    netrc_conn <- file(netrc)
```
infotroph: As written this will overwrite any existing content in the file. Should append instead:
```suggestion
    netrc_conn <- file(netrc, open = "at")
```

infotroph (changes requested):

modules/data.remote/R/NASA_DAAC_download.R
DongchenZ: Fixed. Thanks for the suggestion.

modules/data.remote/R/NASA_DAAC_download.R
DongchenZ: Fixed.

modules/data.remote/R/NASA_DAAC_download.R
DongchenZ: Fixed.

infotroph (approved):
Thanks for updating. Looks good!

modules/data.remote/R/NASA_DAAC_download.R
```diff
@@ -94,37 +94,31 @@ NASA_DAAC_download <- function(ul_lat,
...
-      if (doi == "10.5067/MEaSUREs/GLanCE/GLanCE30.001") {
-        granules_href <- c(granules_href, sapply(granules, function(x) {
-          links <- c()
-          for (j in seq_along(x$links)) {
-            links <- c(links, x$links[[j]]$href)
-          }
-          return(links)
-        }))
+      # GLANCE and HLS products have special data archive.
+      if (doi %in% c("10.5067/MEaSUREs/GLanCE/GLanCE30.001", "10.5067/HLS/HLSS30.002")) {
```
mdietze: This approach of manually listing exempted DOIs is not sustainable as the number of exempted DOIs will just keep going up (e.g. you're already missing the landsat half of the HLS band data) and diagnosing which case you are in isn't obvious to users. Is there any reason to not make this version the default? Are there cases where grabbing links beyond links[[1]] will lead to errors or the selection of unwanted files that aren't being filtered out by the later checks (file type, access type, etc.

mdietze (approved):

PR #3531, opened 2025-05-26
DEMO-1 quarto notebook (run workflow from pecan.xml)

AritraDey-Dev:
## Description
This PR introduces the initial version of a` Quarto notebook` as part of the Quarto `PEcAn notebooks-based workflow GSoC project`. The notebook enables users to run a complete `PEcAn` workflow using a` pre-generated pecan.xml` . It replicates the web-based `Demo 1 `workflow starting after the configuration step, allowing users to execute model runs, perform analysis, and visualize results directly from within the notebook interface.

## Motivation and Context

This PR targets `PEcAn gsoc project running worflows from quarto notebook`.It is decided to keep two separate notebooks for replicating demo 1 workflow running via the web.
- first notebook ->     One for interactively generating the pecan.xml file based on user inputs (model, site, met selection, etc.).
- second notebook ->     Another that takes a pre-generated pecan.xml as input and handles the model run, analysis, and visualization, replicating the web interface flow post-setup.

This particular PR targets the second notebook, which uses a pre-generated pecan.xml file (either configured via the PEcAn web interface or to be created by the first notebook, which will be introduced in a follow-up PR).

#### Directory Structure
Few things to point at this point is about the directory structure -
i planned to keep the quarto notebook in this directory `pecan/base/inst/quarto-notebooks/_extensions/`
then two folder demo 1 and demo 2 
demo 1 folder will consist of two notebooks 1)generate_xml.qmd 2)run_pecan.qmd
demo2  folder will consist of single notebook which will do `ensemble`,`sensivity analysis` basically replicating demo 2 workflow.
```
_extensions/
├── demo1/
│   ├── generate_xml.qmd     # Interactive XML configuration notebook
│   └── run_pecan.qmd        # Run PEcAn workflow using pre-generated XML
└── demo2/
    └── demo2_workflow.qmd   # Notebook for uncertainty & ensemble analysis

```

Any feedback on this directory structure is appreciated.

Knit report: [PEcAn_workflow.pdf](https://github.com/user-attachments/files/20439217/Untitled-1.pdf)

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
[...]

PR #3541, opened 2025-06-11
added multisite-workflow-example.qmd

### LINKED ISSUE #3539: Enable Flat-File Multisite Workflow via createMultiSettings() with Site Metadata Injection
This enhancement documents and demonstrates how to run PEcAn multisite workflows entirely from a flat file (site_info.csv), without querying the BETY database.

Instead of using createSitegroupMultiSettings() (which depends on sitegroup_id and DB queries), users can now:

Filter site metadata directly from a CSV (e.g., by site group)

Use createMultiSettings() to generate settings per site, with all site-level metadata already embedded

<pre> ```r # --- Multisite Flat-File Workflow Example --- 
# Use this when working with site_info.csv and no BETY database # # library(dplyr) 
# site_info <- read.csv("config/site_info.csv") |> # filter(site_group == "NEON") 
# # settings <- createMultiSettings(templateSettings = settings, siteInfo = site_info) 
# # This populates a full settings list for each site with all metadata pre-injected. # No further site metadata DB lookups will be needed in the workflow. ``` </pre>

Documentation plan 

1. Add this example block to workflow.R as a comment

2. Include a note in the user guide/dev README on using createMultiSettings() with site_info.csv

3. Clarify that site_info.csv must contain fields like site_id, lat, lon, city, etc.

harshagr70:
This PR introduces a Quarto notebook (multisite-workflow-example.qmd) that demonstrates how to run a PEcAn multisite workflow using a flat file (site_info.csv) instead of querying the BETY database.

It complements the ongoing GSoC Quarto-based notebook efforts (e.g., [PR #3531](https://github.com/PecanProject/pecan/pull/3531)) by providing a clear, functional alternative for users who wish to run simulations across multiple sites without database dependencies.

fixes : #3539


documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
...
+```
+
+---
+
+## Add the following lines to your workflow code .
+
+```{r}
+site_info <- read.csv("site_info.csv")
+# Filter for NEON sites (change "NEON" to your desired group)
+selected_sites <- site_info %>% filter(site_group == "NEON")
```
AritraDey-Dev, line 72: I also worked on this part to retrieve site info based on the site_group while generating the pecan.xml from Quarto, although I used a direct DB query to do this.

A few doubts about your implementation:
I'm currently retrieving site info based on both the `model` and the `site_group`, which is also how it's done via the web interface. But how do you plan to print all the site info based only on `site_group`, given that the model is also required in that case? Are you assuming a default model in such scenarios?

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
...
+---
+
+## Add the following lines to your workflow code .
+
+```{r}
+site_info <- read.csv("site_info.csv")
+# Filter for NEON sites (change "NEON" to your desired group)
+selected_sites <- site_info %>% filter(site_group == "NEON")  
+print(selected_sites)           ## prints demo output (not required in main code) 
+settings <- createMultiSiteSettings(settings, selected_sites) ## populates settings with metadata for each site in the sitegroup
```
AritraDey-Dev: How will settings object look like after this step?Could you please give an example

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
harshagr70: @AritraDey-Dev . when i looked in the code in the step particularly where  createSitegroupMultiSettings( ) is called i dont see the database being queried for the sites on the basis of models as well , maybe in the further steps this might happen , can you pinpoint the part where this thing happens particularly !!

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
harshagr70: Adding these lines will just simply skip the db call for the sites thats it , the format of the settings object will remain same as it is being now handled by the createMultiSiteSettings() function .

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
AritraDey-Dev: Yes, you're right — the goal was to avoid the DB query.So your approach definitely makes it cleaner to retrieve the site info.
That said, I might be mistaken, but I had the impression that the site info was originally being derived based on both the model and the sitegroup .I need to double check this thing

infotroph: The filtering by model isn't inherent to the concept of sitegroups -- the web workflow does it to support the "conversion" checkbox, which toggles the map between showing all sites in a group or only those that already have inputs (which are model-specific) prepared. I now run quite a few workflows that start from the list of sites and then use it to prepare the missing inputs, so for those cases I wouldn't want any sites filtered out of the group.

In a DB-free notebook context, filtering by input status would probably look something like `site_info$inputs_ready <- some_maybe_expensive_lookup_of_input_files_for_each_site(site_info); selected_sites <- site_info |> filter(site_group == "NEON" & inputs_ready)`. My first thought is that this notebook doesn't need to go into that level of detail, but am curious what y'all think.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
infotroph: I agree it's worth showing a very simple before and after example, even though it'll work the same way it does from a DB query. @harshagr70 let me know if you want help constructing this.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
...
+
+```
+
+---
+
+## (Optional) Integrate with workflow 
+
+```{r}
+settings$site_info <- selected_sites
```
infotroph: This isn't quite right -- after CreateMultiSiteSettings the site info is already integrated into the settings object, but inside each site rather than as its own object. For example a downstream function looking for site latitude will look at `settings$run[[siteID]]$site$lat`.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
+---
+title: "PEcAn Multisite Flat-File Workflow Example"
+format: html
+editor: visual
+---
+
+## Introduction
+
+This notebook demonstrates how to use a flat-file (`site_info.csv`) to set up a multisite workflow in PEcAn, without querying the database.  
+You will learn how to read site metadata, filter by group, and create a MultiSettings object for your runs.
```
infotroph:
```suggestion
This notebook demonstrates how to set up a multisite workflow in PEcAn that uses a subset of the sites defined in a flat-file (`site_info.csv`), without querying the database.  
You will learn how to read site metadata, filter by group, and create a MultiSettings object for your runs.

Sites stored in the BETY database can be queried by `sitegroup` to retrieve predefined sets of sites. Inside the database a sitegroup is nothing but a name given to a list of siteIDs, so in a flat file context we can do the same thing by adding one or more grouping columns.
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
...
+Suppose you have a CSV file like this: 
+
+| id  | lat   | lon    | site_group |
+|-----|-------|--------|------------|
+| 101 | 40.1  | -88.2  | NEON       |
+| 102 | 41.2  | -87.9  | NEON       |
+| 201 | 39.9  | -90.0  | Ameriflux  |
+| 202 | 38.5  | -89.5  | Ameriflux  |
+
+Save this as `site_info.csv` in your working directory.
```
infotroph: You can turn this into a self-contained example by showing the file as a tribble:
~~~suggestion
```{r}
# For this demo:
site_info <- tibble::tribble(
	~id, ~lat, ~lon,  ~elev, ~site_group,
	101, 40.1, -88.2, 10,    "NEON",
	102, 41.2, -87.9, 1980,  "NEON",
	201, 39.9, -90.0, 925,   "Ameriflux",
	202, 38.5, -89.5, 1108,  "Ameriflux"
)
# In a real workflow:
# site_info <- read.csv("site_info.csv")
```
~~~

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
...
+# library(PEcAn.settings) # Uncomment if needed
+# library(PEcAn.all)      # Uncomment if needed
+```
+
+---
+
+## Add the following lines to your workflow code .
+
+```{r}
+site_info <- read.csv("site_info.csv")
```
infotroph:
```suggestion
## Filter by one group

```{r}
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,69 @@
...
+# Filter for NEON sites (change "NEON" to your desired group)
+selected_sites <- site_info %>% filter(site_group == "NEON")  
+print(selected_sites)           ## prints demo output (not required in main code) 
+settings <- createMultiSiteSettings(settings, selected_sites) ## populates settings with metadata for each site in the sitegroup
+
+```
+
```
infotroph, line 78:
~~~suggestion

## Group by multiple characteristics

Say you're developing a new parameterization for the conifer PFT and you want to test it specifically at high elevation. By filtering and reassigning, this can be done without any edits to the saved `site_info.csv`:

```{r}
settings$pfts <- settings$pfts |>
  append(list(
    name="new_conifer",
    posterior.files = "/path/to/new_conifer/post.distns.Rdata"))
neon_hi_elev <- site_info |>
  filter(
    site_group == "NEON",
    elev > 1000) |>
  mutate(site.pft = "new_conifer")
settings <- createMultiSiteSettings(settings, neon_hi_elev)
```
~~~

infotroph (changes requested):
Thanks for creating this! This is a good start that could use a couple more steps to show the key concepts. I've added some suggestions inline and am happy to discuss more.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
harshagr70: True !! , this one seems to be not correct here , will remove it !

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,142 @@
...
+  run = list(
+    start.date = "2025-01-01",
+    end.date = "2025-12-31",
+    inputs = list()
+    # No site info yet!
+  )
+)
+
+str(settings)  
+
```
infotroph: This can be simplified further by keeping the settings object we created above and displaying that -- we usually show it in XML form just to emphasize the connection between the settings object in your R session and the file written out.
```suggestion
listToXml(settings, tag = "pecan")
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,142 @@
...
+settings$pfts <- settings$pfts |>
+  append(list(
+    name="new_conifer",
+    posterior.files = "/path/to/new_conifer/post.distns.Rdata"))
+neon_hi_elev <- site_info |>
+  filter(
+    site_group == "NEON",
+    elev > 1000) |>
+  mutate(site.pft = "new_conifer")
+settings <- createMultiSiteSettings(settings, neon_hi_elev)
```
infotroph:
```suggestion
settings_neon <- createMultiSiteSettings(settings, neon_hi_elev)
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,142 @@
...
+    list(
+      site = list(id = 202, lat = 38.5, lon = -89.5, elev = 1108, site_group = "Ameriflux", met.start = "2025-01-01", met.end = "2025-12-31"),
+      start.date = "2025-01-01",
+      end.date = "2025-12-31",
+      inputs = list()
+    )
+  )
+)
+
+str(settings)
```
infotroph:
```suggestion
listToXml(settings_neon, tag = "pecan")
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
infotroph: This also makes me realize the `tag = "pecan"` is boilerplate and I'll submit a PR to let us skip it.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,130 @@
...
+## Load Required Libraries
+
+```{r}
+library(dplyr)
+library(PEcAn.settings)
+```
+
+
+> **NOTE:** You must provide a 'settings' object (as an R list) before running the code below.
+> Or load from file: settings <- xmlToList(xmlParse("settings.xml"))
```
infotroph:
```suggestion
> Or load from file: `settings <- read.settings("settings.xml")`
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,130 @@
...
+###(sample settings object for running the notebook for multisitesettings)
+
+```{r}
+settings <- list(
+  outdir = "pecan_runs",
+  database = list(
+    bety = list(
+      user = "bety",
+      password = "bety",
+      host = "localhost",
```
AritraDey-Dev: host should be `postgres`.please refer `pecan.xml `generated via web as if you refer `tests` directory there in some of the settings file it is mentioned as `localhost` so it is obvious to make mistake.

AritraDey-Dev: for your information what database config should be
```
  <database>
    <bety>
      <user>bety</user>
      <password>bety</password>
      <host>postgres</host>
      <port>5432</port>
      <dbname>bety</dbname>
      <driver>PostgreSQL</driver>
    </bety>
  </database>
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,130 @@
...
+    )
+  ),
+  model = list(
+    id = 1000000022,
+    type = "SIPNET"
+  ),
+  pfts = list(
+  )
+)
+class(settings) <- "Settings"
```
AritraDey-Dev: for this entire block of code why we cann't use a existing pecan.xml ?and then just do `read.settings`

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,130 @@
...
+# For this demo:
+site_info <- tibble::tribble(
+	~id, ~lat, ~lon,  ~elev, ~site_group,
+	101, 40.1, -88.2, 10,    "NEON",
+	102, 41.2, -87.9, 1980,  "NEON",
+	201, 39.9, -90.0, 925,   "Ameriflux",
+	202, 38.5, -89.5, 1108,  "Ameriflux"
+)
+# In a real workflow:
+# site_info <- read.csv("site_info.csv")
```
AritraDey-Dev, line 26: it will be very helpful if you can provide a csv file that you have prepared already or may be a link that is available in pecan repo that a user can use.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
infotroph: For this example we should be able to remove the `bety` block entirely (and probably the enclosing `database` too, but check me on that)

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,131 @@
...
+  outdir = "pecan_runs",
+  database = list(
+    bety = list(
+      user = "bety",
+      password = "bety",
+      host = "postgres",
+      dbname = "bety",
+      driver = "PostgreSQL"
+    )
+  ),
```
infotroph: Can skip the db section for simplicity:
```suggestion
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
infotroph: I agree we should use the settings constructors rather than set class directly, but I do like that this approach shows the contents inline rather than reading opaquely from a file. 

For demo purposes it would be nice if the inline example were XML text, but that takes enough parsing steps to be distracting. @AritraDey-Dev what do you think about keeping the list format as it is, but changing this line to:
```suggestion
settings <- as.Settings(settings)
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,131 @@
...
+
+## Introduction
+
+This notebook demonstrates how to set up a multisite workflow in PEcAn that uses a subset of the sites defined in a flat-file (`site_info.csv`), without querying the database.  
+You will learn how to read site metadata, filter by group, and create a MultiSettings object for your runs.
+
+Sites stored in the BETY database can be queried by `sitegroup` to retrieve predefined sets of sites. Inside the database a sitegroup is nothing but a name given to a list of siteIDs, so in a flat file context we can do the same thing by adding one or more grouping columns.
+
+## Example `site_info.csv`
+## You can either use this tibble for running demo , or can also use the site_info_example.csv in the document/tutorial/multisite-workflow/site_info_example.csv
```
infotroph: This is a thoughtful detail but don't think we want to include it -- the point of defining this table inline is to avoid needing multiple files for the demo, and I'm OK with assuming someone who's ready to run a PEcAn demo is familiar enough with R to be able to figure out how to read a dataframe from a CSV.

documentation/tutorials/multisite-workflow/site_info_example.csv
```diff
@@ -0,0 +1,11 @@
+id,lat,lon,elev,site_group
```
infotroph: As commented above, I recommend deleting this file -- we show the whole contents inline in enough detail for users to construct their own.

infotroph (commented):
Thanks! I don't think we need the example CSV and left recommendations on resolving @AritraDey-Dev comments; otherwise this is looking great.

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
infotroph:
```suggestion
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,120 @@
...
+```
+
+
+## Filter by one group
+
+```{r}
+# Filter for NEON sites (change "NEON" to your desired group)
+selected_sites <- site_info %>% filter(site_group == "NEON")  
+print(selected_sites)           ## prints demo output (not required in main code) 
+settings <- createMultiSiteSettings(settings, selected_sites) ## populates settings with metadata for each site in the sitegroup
```
infotroph: One last tweak: Renaming to keep the template settings object around for use in the next example. Sorry I missed this before -- @harshagr70 I'll apply this right away and merge.

```suggestion
settings_neon <- createMultiSiteSettings(settings, selected_sites) ## populates settings with metadata for each site in the sitegroup
listToXml(settings_neon, tag = "pecan")
```

documentation/tutorials/multisite-workflow/multisite-workflow-example.qmd
```diff
@@ -0,0 +1,120 @@
...
+listToXml(settings_neon, tag = "pecan")
+```
+
+
+## Shows settings object after "createMultiSiteSettings(settings, selected_sites)"
+
+```{r}
+
+listToXml(settings, tag = "pecan")
+
```
infotroph:
~~~suggestion
## Shows settings object before and after "createMultiSiteSettings(settings, selected_sites)"

```{r}
listToXml(settings, tag = "pecan")
listToXml(settings_neon, tag = "pecan")
```
~~~

infotroph (approved):
Thanks! I see two more tweaks I missed on my last read -- will apply those and merge.

infotroph (approved):

PR #3544, opened 2025-06-12
Add the parallel mode for the North American multi-site SDA workflow.

DongchenZ:
Add the parallel mode for the North American multi-site SDA workflow.

The workflow contains functions to split site lists into various chunks and run SDA for each group of sites.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/assim.sequential/R/sda.enkf_parallel.R
```diff
@@ -0,0 +1,608 @@
...
+    future::plan(future::multisession)
+  }
+  # Tweak outdir if it's specified from outside.
+  if (!is.null(outdir)) {
+    settings$outdir <- outdir
+    settings$rundir <- file.path(outdir, "run")
+    settings$modeloutdir <- file.path(outdir, "out")
+    settings$host$folder <- file.path(outdir, "out")
+    settings$host$outdir <- file.path(outdir, "out")
+    settings$host$rundir <- file.path(outdir, "run")
```
mdietze, line 62: This chunk is silently making very strong assumptions that various folders are always in their default locations, which seems dangerous if they are not. Also, setting the local and host as identical won't work if the host isn't the localhost.

modules/assim.sequential/R/sda.enkf_parallel.R
```diff
@@ -0,0 +1,608 @@
...
+               "module load R/4.1.2", 
+               "echo \"require (PEcAnAssimSequential)", 
+               "      require (PEcAn.uncertainty)",
+               "      require (foreach)", 
+               "      qsub_sda_batch('@FOLDER_PATH@')", 
+               "    \" | R --no-save")
+    jobsh <- gsub("@FOLDER_PATH@", folder.path, jobsh)
+    writeLines(jobsh, con = file.path(folder.path, "job.sh"))
+    # qsub command.
+    qsub <- "qsub -l h_rt=10:00:00 -l mem_per_core=4G -l buyin -pe omp @CORES@ -V -N @NAME@ -o @STDOUT@ -e @STDERR@ -S /bin/bash"
```
mdietze: These bits are hard coded to BU settings. Anything that's BU specific needs to come from the settings file.

modules/assim.sequential/R/sda.enkf_parallel.R
```diff
@@ -0,0 +1,608 @@
...
+#' `OutlierDetection` decide if we want to execute the outlier detection each time after the model forecasting;
+#' `send_email` contains lists for sending email to report the SDA progress;
+#' `keepNC` decide if we want to keep the NetCDF files inside the out directory;
+#' `forceRun` decide if we want to proceed the Bayesian MCMC sampling without observations;
+#' `MCMC.args` include lists for controling the MCMC sampling process (iteration, nchains, burnin, and nthin.).
+#' 
+#' @return NONE
+#' @export
+#' 
+sda.enkf_local <- function(settings,
```
mdietze, line 25: At a high level, there's too much code internal to this function that is identical to that in the SDA multisite. That's going to make both too hard to maintain -- any change in one will need to be made in the other and tested. Much better approach would be to refactor so as to put the shared parts in shared functions that are called by both.

mdietze (changes requested):

modules/assim.sequential/R/sda.enkf_parallel.R
DongchenZ: Fixed the first part by adding a couple of  `logger.info`. For the second part, the current function is called `sda.enkf_local`, which means it can only be run locally. I also added the messaging part to warn the user if they would like to run it through the remote server.

modules/assim.sequential/R/sda.enkf_parallel.R
DongchenZ: Fixed.

modules/assim.sequential/R/sda.enkf_parallel.R
DongchenZ: For this PR, I would like to keep this version for now (so I can reference the exact version of pecan in Chapter 2, and this PR will not affect other people's SDA workflow). And you are right, in the longer term, we will definitely merge those two SDA functions. However, the code-cleaning and merging are non-trivial tasks. I will open another PR that incorporates both functions with a cleaner format, comments, logics, etc.

DongchenZ:
@mdietze @infotroph @dlebauer @Qianyuxuan @cami-webb 
This PR is for the continental SDA workflow, which also made some changes to the current SIPNET SDA functions. Please review the changes and let me know if you have any comments or suggestions.

Specifically, the changes are mostly due to the edge cases across the continent:
1. In `models/sipnet/R/read_restart.SIPNET.R` script.
    1) Remove the `print(run.id)` for clearner log output.

2. In `models/sipnet/R/write.configs.SIPNET.R` script.
    1) resolve the edge cases where locations may have a `leafoff` date earlier than a `leafon` date (See Line 477 - 487).
    2) resolve an issue where SIPNET takes plantWoodInit as all woods (including roots) when t=1, which will lead to smaller wood than expected (See Line 641 - 644).

3. In `models/sipnet/R/write_restart.SIPNET.R` script.
    1) Add an `if-control` for printing out the message (See Line 40 - 42).

4. In `modules/assim.sequential/R/Analysis_sda_block.R` script.
    1) Grabbing `cores` from settings to standardize the parallel computation throughout the SDA workflow (See Line 21 - 28).
    2) Replacing `furrr::map` with `foreach` to silence the MCMC messages (See Line 69 - 89).
    3) Returning the `analysis` output per block to avoid heavy computation load when we sample across the continent (See Line 106 - 107, 668 - 673, and 683).
    4) Resolving edge cases when there are limited observations (<= 1) for a specific site (See Line 243 - 246) when building blocks for the MCMC sampling (See Line 245 - 248).
    5) Resolving edge cases where the process error (in precision) could be very small, causing the singularity problem during the MCMC calculations (this was hard-coded long ago; I will test if the problem still exists, and I think this will not affect anyone) (See Line 411 - 415).
    6) Resolving the edge cases where we have limited observations (<= 1) during the calculations of `mufa` and `pfa` after the MCMC sampling (See Line 542 - 552).

5. In `modules/assim.sequential/R/Nimble_codes.R` script.
    1) Resolving edge cases where we have only one observation for a specific site (See Line 214).

6. In `modules/uncertainty/R/ensemble.R` script.
    1) Adding `unlist` to the path in case the path is in the `list` format instead of a vector (mine is in the `list` format, which causes an error when trying to read the IC file).

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -315,6 +344,7 @@ build.block.xy <- function(settings, block.list.all, X, obs.mean, obs.cov, t) {
   #if it's Wishart Q, we need to replace any NA Y with corresponding muf, and r with Pf.
   #also, if length of observation is 1, the Wishart Q is not suitable for the MCMC.
   #we will then need to change the Q type to 3, which is the vector Q.
+  #the wishart-MCMC is still under development, so I commanded them out for now.
```
infotroph:
```suggestion
  #the wishart-MCMC is still under development, so I commented them out for now.
```

modules/assim.sequential/R/sda.enkf_parallel.R
```diff
@@ -0,0 +1,644 @@
...
+    gc()
+    ###-------------------------------------------------------------------###
+    ###  preparing OBS                                                    ###
+    ###-------------------------------------------------------------------###---- 
+    #To trigger the analysis function with free run, you need to first specify the control$forceRun as TRUE,
+    #Then specify the settings$state.data.assimilation$scalef as 0, and settings$state.data.assimilation$free.run as TRUE.
+    if (!is.null(obs.mean[[t]][[1]]) | (as.logical(settings$state.data.assimilation$free.run) & control$forceRun)) {
+      #decide if we want the block analysis function or multi-site analysis function.
+      #initialize block.list.all.
+      if (t == 1 | !exists("block.list.all")) {
```
infotroph:
```suggestion
    if (!is.null(obs.mean[[t]][[1]]) || (as.logical(settings$state.data.assimilation$free.run) & control$forceRun)) {
      #decide if we want the block analysis function or multi-site analysis function.
      #initialize block.list.all.
      if (t == 1 || !exists("block.list.all")) {
```

modules/assim.sequential/R/sda.enkf_parallel.R
```diff
@@ -0,0 +1,644 @@
...
+##' @author Dongchen Zhang.
+##' @export
+sda_assemble <- function (batch.folder, outdir) {
+  # find folder paths to each SDA job.
+  folders <- list.files(batch.folder, full.names = T)
+  fail.folders <- (folders %>% furrr::future_map(function(f){
+    if(!file.exists(file.path(f, "sda.all.forecast.analysis.Rdata"))) {
+      return(f)
+    }
+  }) %>% unlist)
```
infotroph: This looks like it works fine, but FTR since file.exists accepts vectors I think you could skip the `map` entirely:

```suggestion
fail.folders <- folders[!file.exists(file.path(folders, "sda.all.forecast.analysis.Rdata"))]
```

modules/uncertainty/R/ensemble.R
```diff
@@ -464,7 +464,7 @@ for (input_tag in names(settings$run$inputs)) {
   
      if (!input_tag %in% names(samples)) {
         # Use first path (already validated as single path)
-        settings$run$inputs[[input_tag]]$path <- input$path[1]  } 
+        settings$run$inputs[[input_tag]]$path <- unlist(input$path[1])}
```
infotroph, line 462: Just to confirm: This has been tested after #3561, correct? I think they should be compatible (That was just for the one-input case whereas this will also work on each member of a larger ensemble), but wanted to check.

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ section for the next release.
  - Support for inspecting and plotting NetCDF output variables within the notebook workflow.
 - added support for soil temperature, relative humidity, soil moisture, and PPFD downscaling to `met_temporal_downscale.Gaussian_ensemble`
 
+- Added parallel mode for the entire SDA workflow.
+
```
infotroph, line 59: This is now in an already-released section -- Move up above line 15

infotroph (approved):
Looks good to me! Glad to be converging on a unified parallel strategy. I left a couple very minor comments; only the one about changelog line placement is essential.

Note that I focused on the mechanics of the parallel changes and am leaving all consideration of how it will affect the SDA results to you and @mdietze.

modules/uncertainty/R/ensemble.R
DongchenZ: Yes. I tested using the newest version of pecan.

CHANGELOG.md
DongchenZ: Fixed.

modules/assim.sequential/R/sda.enkf_parallel.R
DongchenZ: Thanks for the suggestion!

modules/assim.sequential/R/sda.enkf_parallel.R
DongchenZ: Fixed.

modules/assim.sequential/R/Analysis_sda_block.R

infotroph:
@mdietze Have your requested changes been addressed here?

mdietze (approved):
Approving, but would like to see an Issue opened about pulling together (and probably refactoring) all the different SDA code variants. It makes me nervous to see the number of variants keep going up instead of going down!

PR #3545, opened 2025-06-13
write.config.SIPNET: check if variables exist in IC file before trying to read them

infotroph:
As discussed in Slack: Checks whether each variable exists in `names(IC.nc$var)` before trying to read it from the file, so missing variables will now be skipped silently. This should reduce the log noise from ncdf4 "cant find var (or dimvar) in file" messages about variables that we intentionally did not include in the IC files. 

Reviewers, **please flag** if any of these variables shouldn't be treated as optional -- if any are mandatory we should stop execution on discovering they're missing.

Also: Added missing NEWS entries

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/NEWS.md
+++ b/models/sipnet/NEWS.md
@@ -7,7 +7,7 @@
 ## Fixed
 
 * `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables. (#3545)
-* When passed a vector of input paths, `write.config.SIPNET` was choosing one at random; it now uses only the first (Blesson Thomas, #3298). Note that a single input path per call has always been the intended usage; being passed many was a second bug in PEcAn.uncertainty that is also now fixed.
+* When passed a vector of multiple input paths, `write.config.SIPNET` was choosing one at random; it now throws an error (Blesson Thomas, #3298). Note that a single input path per call has always been the intended usage; being passed many was a second bug in PEcAn.uncertainty that is also now fixed.
 
 
 # PEcAn.SIPNET 1.9.0
--- a/models/sipnet/NEWS.md
+++ b/models/sipnet/NEWS.md
@@ -6,7 +6,7 @@
 
 ## Fixed
 
-* `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables.
+* `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables. (#3545)
 * When passed a vector of input paths, `write.config.SIPNET` was choosing one at random; it now uses only the first (Blesson Thomas, #3298). Note that a single input path per call has always been the intended usage; being passed many was a second bug in PEcAn.uncertainty that is also now fixed.
 
 
--- a/models/sipnet/NEWS.md
+++ b/models/sipnet/NEWS.md
@@ -1,3 +1,15 @@
+# PEcAn.SIPNET 1.9.1
+
+## Changed
+
+* Breaking: Renamed the setting used to pass soil and hydrology parameters. `write.config.SIPNET` previously read these from `settings$run$inputs$soilinitcond`, now `settings$run$inputs$soil_physics` to better reflect that these are state factors applicable to the whole run rather than initial conditions. (Quianyu Xuan, #3406)
+
+## Fixed
+
+* `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables.
+* When passed a vector of input paths, `write.config.SIPNET` was choosing one at random; it now uses only the first (Blesson Thomas, #3298). Note that a single input path per call has always been the intended usage; being passed many was a second bug in PEcAn.uncertainty that is also now fixed.
+
+
 # PEcAn.SIPNET 1.9.0
 
 ## License change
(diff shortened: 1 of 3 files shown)


models/sipnet/NEWS.md
```diff
@@ -1,3 +1,15 @@
+# PEcAn.SIPNET 1.9.1
+
+## Changed
+
+* Breaking: Renamed the setting used to pass soil and hydrology parameters. `write.config.SIPNET` previously read these from `settings$run$inputs$soilinitcond`, now `settings$run$inputs$soil_physics` to better reflect that these are state factors applicable to the whole run rather than initial conditions. (Quianyu Xuan, #3406)
+
+## Fixed
+
+* `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables.
```
```suggestion
* `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables. (#3545)
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -598,24 +598,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    if (!is.null(IC.pools)) {
       IC.nc <- ncdf4::nc_open(IC.path) #for additional variables specific to SIPNET
+      ic_ncvars_wanted <- c(
+        "nee",
+        "SoilMoistFrac",
+        "SWE",
+        "date_of_budburst",
+        "date_of_senescence",
+        "Microbial Biomass C"
+      )
```
DongchenZ, line 628: This is a great fix.  I would like the `ic_ncvars_wanted` as the optional function input (which could be set as `NULL` as default), and if it's `NULL`, we will use the default name list as you defined here.

infotroph: I like this request in concept -- it would be nice to be able to call write.configs saying "please treat these variables as mandatory, these ones as optional, and ignore these ones" -- but I think that implies a bigger refactor. Each variable here has its own `if` block below, so we'd at least need to define what happens when the user requests an unsupported variable and decide whether unlisted means "optional" or "ignored".

@DongchenZ is exposing this necessary for your use cases or nice to have?

DongchenZ: I agree. It's just something nice to have. You can leave it as it is.

infotroph: Filed this for later as https://github.com/PecanProject/pecan/issues/3554

models/sipnet/NEWS.md
```diff
@@ -1,3 +1,15 @@
+# PEcAn.SIPNET 1.9.1
+
+## Changed
+
+* Breaking: Renamed the setting used to pass soil and hydrology parameters. `write.config.SIPNET` previously read these from `settings$run$inputs$soilinitcond`, now `settings$run$inputs$soil_physics` to better reflect that these are state factors applicable to the whole run rather than initial conditions. (Quianyu Xuan, #3406)
+
+## Fixed
+
+* `write.config.SIPNET` now checks more carefully whether an optional variable exists in an initial condition file before trying to read it, therefore printing fewer messages about (expectedly) missing variables. (#3545)
+* When passed a vector of input paths, `write.config.SIPNET` was choosing one at random; it now uses only the first (Blesson Thomas, #3298). Note that a single input path per call has always been the intended usage; being passed many was a second bug in PEcAn.uncertainty that is also now fixed.
```
mdietze: If more than one path is passed in it should be an error, not a silent decision to keep running with an arbitrary choice of input, which will silently lead to undesirable behavior. If someone is mistakingly passing in a whole ensemble, but every run in the ensemble always runs with ensemble member 1, then they will unknowingly collapse the input uncertainty.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -598,24 +598,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+  } else if (length(settings$run$inputs$poolinitcond$path) > 0) {
     ICs_num <- length(settings$run$inputs$poolinitcond$path)
     IC.path <- settings$run$inputs$poolinitcond$path[[1]] 
     
     IC.pools <- PEcAn.data.land::prepare_pools(IC.path, constants = list(sla = SLA))
     
-    if(!is.null(IC.pools)){
+    if (!is.null(IC.pools)) {
       IC.nc <- ncdf4::nc_open(IC.path) #for additional variables specific to SIPNET
+      ic_ncvars_wanted <- c(
```
mdietze: would be good to comment this, and possibly consider a name change since this seems to be the list of variables you DON'T need, not the variables you want.

models/sipnet/NEWS.md
infotroph: This is a good point, though I *think* #3298 created a guarantee  that run.write.configs will only pass one member. Are you saying write.config.SIPNET ought to check for itself anyway?

models/sipnet/R/write.configs.SIPNET.R
infotroph: It's a list of variables we want to use if they're available but not complain about if they're missing, but you're right the name isn't great.

models/sipnet/NEWS.md
mdietze: While the fixes *should* guarantee exactly one value is passed, I'm saying that because write.config.SIPNET is already doing a check that there are >0 values, that it's negligibly more difficult, but safer, to check that there's exactly 1

infotroph: Done in 222d150f4

models/sipnet/R/write.configs.SIPNET.R
infotroph: Commented and renamed in 222d150f4

mdietze (approved):

PR #3546, opened 2025-06-13
Correct XML error when writing MultiSettings with only one site

infotroph:
Removes assumption that length > 1 from `listToXml.MultiSettings` -- the rest of the MultiSettings class seems to already be fine with "multi" happening to be only one site.

Yes, this is a corner case, but a useful one: I've been using and liking `createMultiSiteSettings(settings, site_info)` as a single-step way of injecting site-specific information into my run settings, and I'd like to be able to use the same pipelines with only one location in site_info.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


dlebauer (approved):
thanks for adding a test!

PR #3547, opened 2025-06-14
Enhance ERA5 Download Function: Direct NetCDF, Flexible Parameters, and Improved Docs

divine7022:
## Description
This PR refactors the `download.ERA5_cds` function to:
- Download ERA5 data directly in `netcdf` format (removes `grib_to_netcdf` conversion dependency).
- Parameters are now passed in the correct vector format to ensure proper API requests.
- Added support for flexible selection of `product type`, `dataset`, and `time steps` (no longer hardcoded).
- Fixed the `extent` argument to ensure correct spatial subsetting as required by the API.
- Provided clearer, beginner-friendly documentation and example usage.
- Removed the **Python** dependency( `cdsapi` via `reticulate` ) and migrating to the native R `ecmwfr` package
- added a unit tests for the extraction function(`test.download.ERA5_cds.R`)
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,55 @@
...
 #' @author Dongchen Zhang
-download.ERA5_cds <- function(outfolder, start_date, end_date, extent, variables, auto.create.key = T, timeout = 36000) {
-  # check shell environments.
-  if ("try-error" %in% class(try(system("grib_to_netcdf"), silent = T))) {
-    PEcAn.logger::logger.info("The grib_to_netcdf function is not detected in shell command.")
-    return(NA)
-  }
+download.ERA5_cds <- function(outfolder, start_date, end_date, 
+                              extent, variables, time = NULL, dataset = "reanalysis-era5-single-levels",
+                              product_type = "ensemble_members", auto.create.key = T, timeout = 36000) {
```
dlebauer:
```suggestion
                              product_type = "ensemble_members", auto.create.key = TRUE, timeout = 36000) {
```

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,55 @@
...
+#'   \item{host}{Host name where the file was downloaded.}
+#'   \item{startdate}{Start date and time of the data in the file.}
+#'   \item{enddate}{End date and time of the data in the file.}
+#'   \item{mimetype}{MIME type of the file ("application/x-netcdf").}
+#'   \item{formatname}{Format name ("ERA5_year.nc").}
+#'
+#' @examples
+#' \dontrun{
+#' era5_files <- download.ERA5_cds(
+#'   outfolder = "D:/working/era5_func_test",
```
dlebauer:
```suggestion
#'   outfolder = with::with_tempdir(),
```

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -99,34 +134,30 @@ download.ERA5_cds <- function(outfolder, start_date, end_date, extent, variables
...
   nc.paths <- c()
   for (y in years) {
-    fname <- file.path(outfolder, paste0("ERA5_", y, ".grib"))
+    fname <- file.path(outfolder, paste0("ERA5_", y, ".nc"))
     # start retrieving data.
     # you need to have an account for downloaing the files
     # Read the documantion for how to setup your account and settings before trying this
     # https://confluence.ecmwf.int/display/CKB/How+to+download+ERA5#HowtodownloadERA5-3-DownloadERA5datathroughtheCDSAPI
     c$retrieve(
       'reanalysis-era5-single-levels',
```
infotroph: Should this now be `dataset` instead of hard-coded?

divine7022: Thanks for your patience; since I got caught up with other tasks and just got a chance to circle back.

The function is now explicitly designed to request and download ERA5 data directly in NetCDF format, so sticking with the `.nc` extension is both intentional and aligned with the scope of this
PR:
> Download ERA5 data directly in netcdf format (removes grib_to_netcdf conversion dependency).

Given that, it makes sense to treat `.nc` as the default and expected output moving forward. That said, if we anticipate needing GRIB support again in the future (e.g, for legacy compatibility), we could make flexible with something like:
```r
if (!is.null(request$data_format) && request$data_format == "netcdf") {
  file_ext <- ".nc"
} else {
  file_ext <- ".grib"
}
```
But unless that flexibility is needed right now, I’d lean toward keeping it simple and fully optimized for NetCDF.
Let me know which direction we’d prefer, I’m happy to adjust accordingly.

infotroph: grib vs nc is a valid question that I think we discussed in another thread, but this comment wasn't about that -- I was talking about line 143 and asking if it should be `c$retrieve(dataset...` instead of `c$retrieve('reanalysis-era5-single-levels'...`

divine7022: the `dataset` parameter is now being used properly in the request. I addressed that in a previous commits 

Also, I’ve removed the `python` dependency and now uses `ecmwfr` package, and it worked well on my end.
If you get a chance, please review the PR. Let me know if anything else stands out!

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,81 @@
...
+#'   \item{file}{File path to the downloaded NetCDF file.}
+#'   \item{host}{Host name where the file was downloaded.}
+#'   \item{startdate}{Start date and time of the data in the file.}
+#'   \item{enddate}{End date and time of the data in the file.}
+#'   \item{mimetype}{MIME type of the file ("application/x-netcdf").}
+#'   \item{formatname}{Format name ("ERA5_year.nc" or "ERA5.rea_year.nc").}
+#'
+#' @examples
+#' \dontrun{
+#' # Download ERA5 reanalysis data for 2020-2021
```
dlebauer: how long do we expect this example to take? What is the approx. size of the download?

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,81 @@
...
+#'   variables = "surface_solar_radiation_downwards",
+#'   time = c("00:00", "12:00"),
+#'   user = "your_cds_user_id",
+#'   key = "your_cds_api_key"
+#' )
+#' }
 #' @export
 #' 
 #' @importFrom purrr %>%
 #' @author Dongchen Zhang
```
dlebauer:
```suggestion
#' @author Dongchen Zhang, Akash BV
```

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,81 @@
...
-#' 
-#' @param outfolder Character: physical path where the ERA5 data are stored.
+#' Download ERA5 climate data from the Copernicus Climate Data Store (CDS) API as NetCDF files, year by year, according to user-specified parameters.
+#' The function saves one NetCDF file per year in the specified output directory.
+#'
+#' @details
+#' This function requires a valid CDS API key and the \code{ecmwfr} package for accessing the Copernicus Climate Data Store.
+#' To get a Copernicus CDS API key, register at \url{https://cds.climate.copernicus.eu/profile}.
+#' You must provide both \code{user} (UID) and \code{key} parameters from your CDS profile.
+#'
```
dlebauer, line 11:
```suggestion
#'
#' You must also accept the terms of use (CC-BY) for the dataset you are downloading by
#' checking "CC-BY" under the ['licenses' tab of your profile page](https://cds.climate.copernicus.eu/profile?tab=licences).
```

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,81 @@
...
+#'   \item{startdate}{Start date and time of the data in the file.}
+#'   \item{enddate}{End date and time of the data in the file.}
+#'   \item{mimetype}{MIME type of the file ("application/x-netcdf").}
+#'   \item{formatname}{Format name ("ERA5_year.nc" or "ERA5.rea_year.nc").}
+#'
+#' @examples
+#' \dontrun{
+#' # Download ERA5 reanalysis data for 2020-2021
+#' era5_files <- download.ERA5_cds(
+#'   outfolder = "~/output/era5_data",
```
dlebauer:
```suggestion
#' output_dir <- withr::local_tempdir()
#' era5_files <- download.ERA5_cds(
#'   outfolder = output_dir,
```

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,81 @@
...
+#'   extent = c(-72.2215, -72.1215, 42.4878, 42.5878),
+#'   variables = c("2m_temperature", "surface_pressure"),
+#'   product_type = "reanalysis",
+#'   user = "your_cds_user_id",
+#'   key = "your_cds_api_key"
+#' )
+#' 
+#' # Download ensemble data for specificed hours only
+#' era5_files <- download.ERA5_cds(
+#'   outfolder = "~/output/era5_data",
```
dlebauer:
```suggestion
#'   outfolder = output_dir,
```

dlebauer (changes requested):
Looks great - thanks for these tests.

I'm running one of the examples, and it has been running for ~20 minutes. How long do you expect the examples to take? This seems long for an example / proof of concept. 

I've made a few minor suggestions, including one that generally applies to your work - please add your name to the author field when you have made non-trivial changes to functions!

modules/data.atmosphere/R/ERA5_download.R
divine7022: thanks for the feedback! you're right about the runtime concern. The current example downloads 2 years of data (2020-2021) and for larger extent, which is indeed excessive for documentation purposes.
so i went ahead and reduced the reduced the `extent` and `year` for documentation purpose;
it takes time around 10 - 15 mins and file size are in few kbs; depends on how large input we are requesting.

modules/data.atmosphere/R/ERA5_download.R
divine7022: 👍

modules/data.atmosphere/R/ERA5_download.R
divine7022: 👍

modules/data.atmosphere/R/ERA5_download.R
divine7022: thanks for noticing -- added it in!

modules/data.atmosphere/R/ERA5_download.R
divine7022: done 👍

modules/data.atmosphere/R/ERA5_download.R
dlebauer: As commented elsewhere, we use `pkg::fn` notation which makes `@importFrom` not necessary

modules/data.atmosphere/R/ERA5_download.R
dlebauer: Note that using "T" or "F" is discouraged; use "TRUE" or "FALSE" instead

dlebauer (changes requested):
Hi @divine7022 thanks again for this major update to the ERA5 code, it looks great - and is very close! 

Mostly there are just a few minor changes to implement, including a number of suggested changes that you 👍🏻 but were not committed. These are now marked 'outdated' but not 'resolved'.

modules/data.atmosphere/R/ERA5_download.R
divine7022:
```r
download.ERA5_cds <- function(outfolder, start_date, end_date, 
                              extent, variables, time = NULL, dataset = "reanalysis-era5-single-levels",
                              product_type = "ensemble_members", user = NULL, key = NULL, timeout = 36000) {
```             
this is the current function signature;
 `auto.create.key = TRUE` argument is outdated -- it was used when the function relied on Python via `reticulate`.
Now that we’ve switched to using `ecmwfr`.

modules/data.atmosphere/R/ERA5_download.R
divine7022: This has already been fixed

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,26 +1,83 @@
...
-download.ERA5_cds <- function(outfolder, start_date, end_date, extent, variables, auto.create.key = T, timeout = 36000) {
-  # check shell environments.
-  if ("try-error" %in% class(try(system("grib_to_netcdf"), silent = T))) {
-    PEcAn.logger::logger.info("The grib_to_netcdf function is not detected in shell command.")
-    return(NA)
+#' @author Dongchen Zhang, Akash
+
+download.ERA5_cds <- function(outfolder, start_date, end_date, 
+                              extent, variables, time = NULL, dataset = "reanalysis-era5-single-levels",
+                              product_type = "ensemble_members", user = NULL, key = NULL, timeout = 36000) {
```
divine7022: this is the current function signature;
`auto.create.key = TRUE` argument is outdated -- it was used when the function relied on Python via reticulate.
Now that we’ve switched to using ecmwfr.

divine7022:
> As commented elsewhere, we use `pkg::fn` notation which makes @importFrom not necessary

@dlebauer thanks! for the catch - now uses base R pipe `|>` instead of `%>%`
i hope all the suggestions have been addressed.

infotroph:
> now uses base R pipe |> instead of %>%

Thank you for that change! Note that this means the package now requires a version of R new enough to contain the base pipe, but I don't think the package checks detect this yet. Can you please add `Depends: R (>= 4.1)`  at line 22 of the DESCRIPTION file (between the Description: and Imports: sections)?

modules/data.atmosphere/NEWS.md
```diff
@@ -14,6 +14,7 @@
 
 * Functions that take argument `site_id` now accept strings as well as BETYdb numeric IDs. 
 * `download.ERA5.old` renamed to `download.ERA5_cds`
+* `download.ERA5_cds` now uses the R package ecmwfr (replacing python dependency of cdsapi via reticulate), enabling direct NetCDF downloads; and made flexible for both reanalysis and ensemble data product.
```
infotroph: 1.9.0 is the previous release -- please move this up to the 1.9.1 section

divine7022:
thanks @infotroph now `Depends: R (>= 4.1.0)` added to DESCRIPTION

modules/data.atmosphere/NEWS.md
divine7022: fixed

modules/data.atmosphere/R/ERA5_download.R
infotroph: Are user and key optional or required here? A NULL default typically means more or less "the function will figure this out if not provided",  which would be aligned with the previous getnetrc approach. But on a quick skim below I see an error if user or key are null and didn't see anything that would fill them in before that -- if they're now required arguments that's fine but might as well remove the default value to make that clearer.

divine7022: thanks! @infotroph now removed NULL defaults for `user/key` parameters

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,133 +1,147 @@
...
-download.ERA5_cds <- function(outfolder, start_date, end_date, extent, variables, auto.create.key = T, timeout = 36000) {
-  # check shell environments.
-  if ("try-error" %in% class(try(system("grib_to_netcdf"), silent = T))) {
-    PEcAn.logger::logger.info("The grib_to_netcdf function is not detected in shell command.")
-    return(NA)
+#' @author Dongchen Zhang, Akash
+
+download.ERA5_cds <- function(outfolder, start_date, end_date, 
+                              extent, variables, time = NULL, dataset = "reanalysis-era5-single-levels",
+                              product_type = "ensemble_members", user, key, timeout = 36000) {
```
infotroph:
```suggestion
                              extent, variables, user, key, time = NULL,
                              dataset = "reanalysis-era5-single-levels",
                              product_type = "ensemble_members",
                              timeout = 36000) {
```

modules/data.atmosphere/R/ERA5_download.R
```diff
@@ -1,133 +1,147 @@
...
-  #grab the client object.
-  tryCatch({
-    c <- cdsapi$Client()
-  }, error = function(e) {
-    PEcAn.logger::logger.severe(
-      "The following error was thrown by `cdsapi$Client()`: ",
-      conditionMessage(e)
-    )
-  })
+  ecmwfr::wf_set_key(user = user, key = key)
```
infotroph, line 109: I've pulled this branch for testing and am getting an error at this step:
```
Error in ecmwfr::wf_set_key(user = user, key = key) : 
  Please provide a service for which
         to set your API key ('webapi' or 'cds')
```

Should this call include `service = "cds"`?

infotroph: No, my copy of ecmwfr was out of date

divine7022: thanks for updating; i was suspecting why it was working for me -- that explains why it was working fine for me - looks like it was a version-specific requirement in an intermediate ecmwfr release. The current version handles the service detection automatically. Classic dependency version issue!
https://bluegreenlabs.org/post/ecmwfr-2.0.0-release-and-changes/

infotroph: But testing this does suggest a way to make user and key optional in a future iteration:

- give key and user default NULL in the function signature
- if key provided, use it as here (if user is NULL, ecmwfr uses "ecmwfr" as a default)
- if key not provided, try a `wf_get_key(user = user)` (NULL uses "ecmwfr" same as above)
- If not authed yet, check for a `~/.cdsapirc` and read from there before giving up

But like I said, future iteration

modules/data.atmosphere/DESCRIPTION
```diff
@@ -61,6 +62,7 @@ Imports:
     zoo
 Suggests:
     doParallel,
+    ecmwfr,
```
infotroph:
```suggestion
    ecmwfr (>= 2.0.0),
```

modules/data.atmosphere/R/ERA5_download.R
divine7022: great suggestion on the `.cdsapirc` credential file approach!; the `.cdsapirc` pattern follows the established precedent from the python cdsapi library; 
i will add this update as u said in next iteration.

modules/data.atmosphere/DESCRIPTION
divine7022: fixed!

dlebauer (approved):
Thanks for making these changes - looks like all requested changes have been made. 

Apologies some of my earlier requests were based outdated code.

divine7022:
Thanks for approving @dlebauer

PR #3548, opened 2025-06-14
fix(ci): docker stack ci by using latest tag

AritraDey-Dev:
## Description

This PR updates the docker sipnet CI workflow to address a runtime issue when executing the sipnet model via Docker.

## Motivation and Context

While running the workflow with sipnet, the following error is encountered:
```
$ cat logfile.txt
ERROR in readNamelistInputs: Read unexpected input item: CHANGE_PARAM
Please fix sipnet.in and re-run
mv: cannot stat '/data/docker_stack_tests/sipnet/run/99000000001/sipnet.out': No such file or directory
```
The error in the first line which prevents the model from executing successfully. As a result, sipnet.out is not generated, leading to the failure seen in the third line.

This issue was originally addressed by @Alomir in [PecanProject/sipnet#94](https://github.com/PecanProject/sipnet/pull/94). However, the Docker image currently used in the workflow pulls the develop tag, which may not include this fix.

This PR resolves the issue by explicitly setting the Docker image tag to latest in the env.
Alternatively, the issue could also be resolved by omitting the develop tag from the` .env `file, which would cause docker compose to use the latest image by default.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/docker-stack-sipnet.yml
+++ b/.github/workflows/docker-stack-sipnet.yml
@@ -39,12 +39,12 @@ jobs:
       run: |
           cp docker-compose.dev.yml docker-compose.override.yml
           sed -i'' '/R_library/d' docker-compose.override.yml
-
+      #test
     - name: Build Containers
       run: |
         cp docker/env.example .env
         echo "COMPOSE_PROJECT_NAME=pecan" >> .env
-        echo "PECAN_VERSION=latest" >> .env
+        echo "PECAN_VERSION=develop" >> .env
         echo "UID=$(id -u)" >> .env
         echo "GID=$(id -g)" >> .env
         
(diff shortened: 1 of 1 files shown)


infotroph:
Thanks for noticing this and for proposing, but this isn't the right fix -- The point of CI is to test the develop branch, so switching to another one defeats the point.

The good news is that I think this issue has already been fixed in Sipnet, so all we should need on the PEcAn side is to cause a rebuild of the Docker image so that it picks up the updated version. I've just kicked one off manually and will check back after it finishes,

AritraDey-Dev:
Thanks for updating the image.`sipnet` with develop tag is now available. I'll go ahead and close this PR. pushed an empty commit to verify the update.

Would it make sense to mark this GitHub Action as a required check for CI? Given its relevance, it seems valuable to catch any issues early, especially those that could impact local development setups as well.

PR #3550, opened 2025-06-24
fix: rendered the target link for the update bety documentation

harshagr70:
fixes - https://github.com/PecanProject/PecanProject.github.io/issues/110
The new target link is now set to open the page to the introduction sub bar directly , fixing the issue of page not found .


infotroph (approved):

harshagr70:
@infotroph Thanks for the approval !! 😄

Issue #3553, opened 2025-06-26
Refactor all query.site usage to check for site metadata in settings


harshagr70:
## Description

The PEcAn codebase currently calls the query.site function in several places to fetch site metadata (e.g., latitude and longitude) from the database. However, users now have the option to supply site metadata (such as lat/lon) directly in their settings, for example by using a flat file (currently being a part of a feature in PR #3541 )(like site_info.csv). In these cases, re-querying the database for site metadata is unnecessary. 

## Proposed Solution

- Locate all uses of query.site in the codebase.
- At each location, add a conditional check:
     If settings$run[[siteID]]$site$lat and settings$run[[siteID]]$site$lon are already present, use these values 
     directly and skip the database query. Only call query.site if the required site metadata is missing from the 
     settings.

This update will make workflows more efficient for users who provide site metadata in their settings, while maintaining backward compatibility for users who rely on the database.

PR #3555, opened 2025-06-28
refactor: query.site use in ic_process.R

harshagr70:
This PR updates `ic_process.R` to use site latitude and longitude from the user-provided settings if available, falling back to the BETY database only when needed. This change supports workflows that do not require a database connection for site metadata and improves flexibility.
Refactors (checks) the unnecessary call to query.site , if metadata is available .
fixes a part of #3553


infotroph (approved):
Thanks! LGTM

(When we discussed this live I wasn't sure if this would work with multi-site settings files. For the record it won't, but it's not a problem because the rest of the function already won't either)

harshagr70:
@infotroph 
Just a heads-up: if `<lat> or <lon>` are present in the XML but empty (`<lat></lat>`), they get parsed as "", not NULL. So checks like `!is.null(settings$run$site$lat) `will return `TRUE `even when the values are effectively missing. Might be safer to also check for` != "".`
or are we assuming that the tags are never passed empty here in the settings.xml

infotroph:
@harshagr70 ooh, good observation! my gut says empty lat/lon tags are a rare enough corner case that it's OK to leave them unhandled, but I don't object to adding the empty string check.

harshagr70:
Same goes for this PR as well https://github.com/PecanProject/pecan/pull/3568
i will add this extra check for both the cases here , so that i doesn't break anything later

infotroph (approved):

PR #3559, opened 2025-07-01
added functions for converting fertilizer type and amounts to applica…

dlebauer:
## Description

This PR contains a function that will estimate the amount of N and C from the fertilizer type and amount. 

- It leverages the fertilizer table from the SWAT model
- Adds C:N ratio
- Adds a row for ammonium nitrate, which was surprisingly missing.

## Motivation and Context

This is to support workflows that take agronomic management events and translate them into model inputs.

This is a DRAFT intended to get feedback on the approach
 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.land/NAMESPACE
```diff
@@ -61,6 +64,8 @@ export(to.Tag)
 export(to.TreeCode)
 export(write_ic)
 export(write_veg)
+import(dplyr)
+import(tidyverse)
```
mdietze: pretty sure this shouldn't be here

modules/data.land/NAMESPACE
```diff
@@ -14,6 +14,7 @@ export(cohort2pool)
 export(dataone_download)
 export(download.SM_CDS)
 export(download_package_rm)
+export(download_soilgrids_raster)
```
mdietze: not seeing this anywhere else in the PR

modules/data.land/NAMESPACE
```diff
@@ -46,6 +48,7 @@ export(pool_ic_list2netcdf)
 export(pool_ic_netcdf2list)
 export(prepare_pools)
 export(put_veg_module)
+export(query_soilgrids)
```
mdietze: not seeing this anywhere else in the PR

modules/data.land/data-raw/create_fertilizer_data.R
```diff
@@ -0,0 +1,29 @@
+library(readr)
+library(dplyr)
+library(stringr)
```
mdietze: are there supposed to be library calls here? seems like it would be better to use namespaces like the rest of the code

mdietze (commented):
Seems potentially useful. Will need to think more about how this intersects with the information we're compiling about the fertilizer application rates for different crops (I'm not sure they're specifying specific fertilizer types as much as specific nutrient recommendations). Also, are there cases where we should be tracking the water content of fertilizers as well (e.g. liquid manure).

modules/data.land/NAMESPACE
dlebauer:
```suggestion
```

Sorry - I had some extra files sitting around

modules/data.land/data-raw/create_fertilizer_data.R
dlebauer: fixed

dlebauer (commented):
fixed exports in NAMESPACE and use of namespaces in data_raw code.

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,110 @@
...
+#'
+#' This function calculates the different forms of nitrogen (NO3-N, NH4-N, organic N) and organic carbon (C_org) in a fertilizer application.
+#' It can determine fertilizer nitrogen and carbon content using either a lookup table based on 
+#' the SWAT model's [`fertilizer.frt`](https://github.com/swat-model/swatplus/blob/main/data/Osu_1hru/fertilizer.frt)
+#' file or determine the fertilizer's 
+#' nutrient content based on NN-PP-KK format.
+#'
+#' Consistent with assumptions (TODO: confirm this statement) in DayCent, DSSAT, and other models, urea is treated as NH3 because the transformation typically occurs within a day.
+#' 
+#' @param fertilizer_type Character string specifying the type of fertilizer. Can be one of "urea", "ammonium_nitrate", "compost", "manure", etc., or a string in the format "NN-PP-KK" (e.g., "45-5-10").
```
infotroph: "etc" implies it accepts other values not mentioned here; not clear where one would find the authoritative list of accepted values

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,110 @@
...
+#'
+#' @examples
+#' fertilizer_components("45-00-00", 200)
+#' fertilizer_components("compost", 1000)
+#' fertilizer_components("manure", 1000, fraction_organic_n = 0.02, fraction_organic_c = 0.08)
+#'
+#' @import dplyr
+#' @import tidyverse
+#' @export
+fertilizer_components <- function(
```
infotroph: Nit: Tidyverse style recommends verbs for function names and nouns for variable names. Would `look_up_fertilizer_components` be more clear?

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,110 @@
...
+      NH4_N <- 0
+      N_org <- 0
+      C_org <- 0
+      return(list(
+        fertilizer_type = fertilizer_type,
+        NO3_N = NO3_N,
+        NH4_N = NH4_N,
+        N_org = N_org,
+        C_org = C_org
+      ))
```
infotroph:
```suggestion
      return(list(
        fertilizer_type = fertilizer_type,
        NO3_N = (fertilizer_amount * fraction_organic_n),
        NH4_N = 0,
        N_org = 0,
        C_org = 0
      ))
```

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,110 @@
...
+#' fertilizer_components("45-00-00", 200)
+#' fertilizer_components("compost", 1000)
+#' fertilizer_components("manure", 1000, fraction_organic_n = 0.02, fraction_organic_c = 0.08)
+#'
+#' @import dplyr
+#' @import tidyverse
+#' @export
+fertilizer_components <- function(
+  fertilizer_type, 
+  fertilizer_amount,
```
infotroph: And now that I started nitpicking: Would just "type" and "amount" be clear enough here since the whole function is about fertilizer? Having 4/5 arguments start with F seems like it makes tab-complete harder than it needs to be.

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,110 @@
...
+    } else {
+      PEcAn.logger::logger.error(
+        "Invalid fertilizer type. Please choose a valid fertilizer_type or provide NN-PP-KK format.\n",
+        "valid fertilizer_types include:", paste0(fertilizer_data$fertilizer_name, sep = ", ")
+      )
+    }
+  } else {
+    # If fractions are provided for a fertilizer type that is in the database, warn the user
+    if (!is.null(fraction_organic_n) && !is.null(fraction_organic_c)) {
+      PEcAn.logger::logger.warn("Provided fraction_organic_n and fraction_organic_c values are being ignored as the fertilizer type is found in the database.")
```
infotroph: Yes to warning the user, but why ignore the value they provided rather than use it to override the DB value? Being able to do things like "treat this as if it were ammonium nitrate but with 10% organic N" seems pretty useful for custom formulations.

modules/data.land/data-raw/create_fertilizer_data.R
```diff
@@ -0,0 +1,25 @@
...
+      swat_id = dplyr::row_number(),
+      fraction_no3_n = fraction_organic_n - fraction_nh3_n
+    ) |>
+    dplyr::select(swat_id, fertilizer_name, fraction_mineral_n, fraction_nh3_n, 
+           fraction_no3_n, fraction_organic_n, description)
+}
+
+fertilizer_nutrient_data <- get_swat_fert_table()
+
+usethis::use_data(fertilizer_nutrient_data, overwrite = TRUE)
```
infotroph:
```suggestion
usethis::use_data(fertilizer_nutrient_data, overwrite = TRUE)

```

modules/data.land/data-raw/create_fertilizer_data.R
```diff
@@ -0,0 +1,25 @@
+get_swat_fert_table <- function() {
```
infotroph: Thanks for including the data creation function! Can you please also commit the resulting `data/fertilizer_nutrient_data.rda` and its Roxygen documentation block?

modules/data.land/tests/testthat/test.fertilizer_components.R
```diff
@@ -0,0 +1,51 @@
...
+                    C_org = 0)
+  )
+})
+
+test_that("Create fertilizer based on specified components", {
+  result <- fertilizer_components(
+    fertilizer_type   = "compost", 
+    fertilizer_amount = 1000, 
+    fraction_organic_n        = 0.02, 
+    fraction_organic_c        = 0.08)
```
infotroph: Odd alignment
```suggestion
    fertilizer_type = "compost", 
    fertilizer_amount = 1000, 
    fraction_organic_n = 0.02, 
    fraction_organic_c = 0.08)
```

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,110 @@
...
+#'   - `fertilizer_type`: The type of fertilizer used.
+#'   - `NO3_N`: The amount of nitrate nitrogen (NO3-N) in kg/ha.
+#'   - `NH4_N`: The amount of ammonium nitrogen (NH4-N) in kg/ha.
+#'   - `N_org`: The amount of organic nitrogen in kg/ha.
+#'   - `C_org`: The amount of organic carbon in kg/ha.
+#'
+#' @examples
+#' fertilizer_components("45-00-00", 200)
+#' fertilizer_components("compost", 1000)
+#' fertilizer_components("manure", 1000, fraction_organic_n = 0.02, fraction_organic_c = 0.08)
```
infotroph: I don't see either "compost" or "manure" as names in the SWAT table -- will these examples actually work as written?

infotroph (commented):
Thanks for putting this together! I commented on a few nits, plus three substantive changes needed:

* Overriding user-specified organic C and N when it conflicts with the database seems backwards to me -- is there a design reason for this that I'm not seeing?
* Need to commit the fertilizer data file as well as the data creation function
* You mentioned adding a row for ammonium nitrate, but I can't find where in the code it happens. Can you point me to it?

PS: Nice work on the tests!

modules/data.land/R/fertilizer_components.R
dlebauer: Fixed

modules/data.land/R/fertilizer_components.R
dlebauer: done

modules/data.land/R/fertilizer_components.R
dlebauer: If there is ammonium nitrate _and_ 10% organic N, then the user can create a new name for the type of fertilizer. Changed to severe, and asks user to supply a different name.

dlebauer:
@mdietze regarding 

> are there cases where we should be tracking the water content of fertilizers as well (e.g. liquid manure).

This source of water can be important, but at this point out of scope for this function. The function assumes dry weight. Improvements to this function are certainly welcome, but the SIPNET implementation would require this to be treated as a separate irrigation event.

dlebauer:
@infotroph I updated the dataset and function following your suggestions.

infotroph:
@dlebauer I see lots of comments referring to changes that I don't see in the code -- did you push the right branch?

modules/data.land/NAMESPACE
```diff
@@ -61,6 +62,8 @@ export(to.Tag)
 export(to.TreeCode)
 export(write_ic)
 export(write_veg)
+import(dplyr)
+import(tidyverse)
```
dlebauer, line 66:
```suggestion
```

modules/data.land/NAMESPACE
dlebauer:
```suggestion
```

dlebauer:
@infotroph sorry - I got mixed up because `gh pr checkout 3559` creates a branch named `pr/dlebauer/3559` for review, but the intended workflow is to make changes on the original feature branch.

modules/data.land/tests/testthat/test.fertilizer_components.R
```diff
@@ -0,0 +1,84 @@
...
+  expect_equal(result,
+               list(type = "dairy_fr",
+                    NO3_N = 0, NH4_N = 7,
+                    N_org = 31, C_org = 391)
+  )
+})
+
+test_that("Invalid fertilizer type returns NULL", {
+  # It generates PEcAn.logger::logger.severe, returns NULL
+  # Temporarily disable logging to avoid cluttering test output
```
infotroph: This is OK; FYI if you want to verify directly that logger.severe fired I think `expect_error(expr, "SEVERE")` should work
(And since I have it fresh in my head from fighting it last night: if you want to check logger messages via `expect_output`, you want `logger.setUseConsole(TRUE, FALSE)` beforehand)

modules/data.land/R/fertilizer_components.R
```diff
@@ -0,0 +1,116 @@
...
+#' @param fraction_organic_n Optional numeric value specifying the fraction of the organic matter that is nitrogen. 
+#' Used to define organic matter additions if not provided in the dataset.
+#' @param fraction_organic_c Optional numeric value specifying the fraction of the organic matter that is carbon. 
+#' Used to define organic matter additions if not provided in the dataset.
+#'
+#' @md
+#' @note The following is a list of valid fertilizer names:
+#' - Mineral fertilizers: ammonium_nitrate, anhydrous_ammonia, urea
+#' - Fresh manures: beef_fr, broil_fr, dairy_fr, duck_fr, goat_fr, horse_fr, 
+#'      layer_fr, sheep_fr, swine_fr, trkey_fr, veal_fr
```
infotroph: Should the generic `manure` you added below appear here as well?

infotroph (approved):

modules/data.land/R/fertilizer_components.R
dlebauer: good catch; fixed

modules/data.land/tests/testthat/test.fertilizer_components.R
dlebauer: well, I noticed that yesterday and thought the behavior strange. I think there is also a bug in logger.getLevelNumber that prevented the error from being passed (fixed in https://github.com/infotroph/pecan/pull/11). That would also be a good place to document these use cases ... lots to keep track of!

infotroph: If it's the same issue I saw, the problem isn't print levels but that expect_output captures stdout while logger prints to stderr by default (but doesn't issue classed conditions for any level below severe, hence why expect_error doesn't catch those)

mdietze:
@sarahkanee make sure to take this new bit of code into account (either using it or building upon it)

PR #3567, opened 2025-07-12
Refactor ic_process.R to get modelname from settings.xml

harshagr70:
- **Removed redundant `is.null(model)` check**  
`model` was always initialised as a list just before the conditional, so `is.null(model)` would always return `FALSE`. This check was unnecessary and has been removed to simplify the logic.

- **Added support for retrieving `model$name` directly from `settings.xml`**  
If `settings$model$name` is present, it will now be used to populate `model$name` without requiring a database query.

- **Retained DB lookup as a fallback**  
If `settings$model$name` is not provided, the function will fall back to querying BETYdb for the model name using the type and the id , maintaining backward compatibility.

---

This update improves efficiency, removes dead code, and respects user-defined values when available.


harshagr70:
Closing this — created from the wrong base. Will open a clean PR from the correct branch. Apologies

Issue #2274, opened 2019-02-04
`load.cfmet` end date is too strict?


ashiklom:
### Bug Description

Based on [this conversation](https://github.com/PecanProject/pecan/pull/2273#discussion_r253212143) on PR #2273.

Basically, `data.atmosphere::load.cfmet` throws an error like the following when given CRUNCEP input (this case is thrown from `met2model.BIOCRO`).

```
test.met2model.R:15: error: Met conversion runs without error
run end date 2000-12-31 23:59:59 after met data ends 2000-12-31 21:00:00
1: met2model.BIOCRO(in.path, in.prefix, outfolder, lat = 45.25, lon = -84.75, start_date = start_date, 
       end_date = end_date) at /Users/shik544/Box Sync/Projects/pecan_project/pecan/models/biocro/tests/testthat/test.met2model.R:15
2: PEcAn.data.atmosphere::load.cfmet(met.nc, lat = lat, lon = lon, start.date = yrstart, 
       end.date = yrend) at /Users/shik544/Box Sync/Projects/pecan_project/pecan/models/biocro/R/met2model.BIOCRO.R:71
3: PEcAn.logger::logger.severe("run end date", end.date, "after met data ends", max(all.dates$date))
4: stop(paste(msg, ...))
```

The issue is that [this code block (`load.cfmet`)](https://github.com/PecanProject/pecan/blob/develop/modules/data.atmosphere/R/load.cfmet.R#L52-L61) expects the end of the year to be 23:59 on December 31, but the last data point for 3-hourly met like CRUNCEP is at 21:00.

### To Reproduce

Remove the `skip` statement from `test.met2model.R` for BIOCRO (once #2273 is merged); then, run `devtools::test("models/biocro")`.

ashiklom:
Possibly related to #1343, which was closed by @dlebauer without comment.

infotroph:
@ashiklom I mean, @dlebauer closed it by merging my PR that claimed it was fixed, so I don't think this is too heavily on him.

infotroph:
Proposed resolution: 

```
delta <- stats::median(diff(all.dates$date), na.rm = TRUE)
...
if (end.date > (max(all.dates$date) + delta)) { ...
```
...our met files _are_ all regularly spaced modulo missing data, yeah?

infotroph:
Closing as (hopefully!) fixed by #3578

PR #3578, opened 2025-07-20
fix(biocro): handle met data end time in met2model.BIOCRO to prevent run end date error

AritraDey-Dev:
## Description
This PR addresses a runtime error encountered when running the BIOCRO model via the PEcAn web interface. The error is as follows:
```
remotefunc <- function() {PEcAn.BIOCRO::met2model.BIOCRO(lst=-8, lat=41.0310397553517, lon=-116.357721712155, spin_nyear=NULL, spin_nsample=NULL, spin_resample=NULL, dbfile.id=99000000167, overwrite=FALSE, in.path='/data/dbfiles/CRUNCEP_site_2-1087', in.prefix='CRUNCEP', outfolder='/data/dbfiles/CRUNCEP_BIOCRO_site_2-1087/', start_date='2004/01/01', end_date='2004/12/31')}
> remoteout <- remotefunc()
2025-07-20 12:50:19.723249 SEVERE [PEcAn.data.atmosphere::load.cfmet] :
run end date 1104537599 after met data ends 1104526800
Error in PEcAn.logger::logger.severe("run end date", end.date, "after met data ends", :
run end date 2004-12-31 23:59:59 after met data ends 2004-12-31 21:00:00
Calls: remotefunc -> -> ->
Execution halted
```
## Motivation and Context

may be related to [Issue #2274](https://github.com/PecanProject/pecan/issues/2274).
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/load.cfmet.R
+++ b/modules/data.atmosphere/R/load.cfmet.R
@@ -51,10 +51,11 @@ load.cfmet <- function(met.nc, lat, lon, start.date, end.date) {
 
   all.dates <- data.frame(index = seq_along(time.idx), date = date)
   
-  if (nrow(all.dates) > 1) {
-    delta <- stats::median(diff(all.dates$date), na.rm = TRUE)
-  } else {
-    delta <- as.difftime(1, units = "hours")  
+  delta <- stats::median(diff(all.dates$date), na.rm = TRUE)
+  if (is.na(delta)) {
+    # probably only happens with a one-line met file
+    # fall back to requiring exact match
+    delta <- 0
   }
   
   if (start.date < (min(all.dates$date) - delta)) {
--- a/modules/data.atmosphere/R/load.cfmet.R
+++ b/modules/data.atmosphere/R/load.cfmet.R
@@ -51,11 +51,21 @@ load.cfmet <- function(met.nc, lat, lon, start.date, end.date) {
 
   all.dates <- data.frame(index = seq_along(time.idx), date = date)
   
-  if (start.date + lubridate::days(1) < min(all.dates$date)) {
-   PEcAn.logger::logger.severe("run start date", start.date, "before met data starts", min(all.dates$date))
+  # Calculate the typical timestep of the met data using median for robustness
+  if (nrow(all.dates) > 1) {
+    delta <- stats::median(diff(all.dates$date), na.rm = TRUE)
+  } else {
+    delta <- as.difftime(1, units = "hours")  # default to 1 hour if only one timestamp
   }
-  if (end.date > max(all.dates$date)) {
-   PEcAn.logger::logger.severe("run end date", end.date, "after met data ends", max(all.dates$date))
+  
+  # Check if start date is more than one timestep before the first available data
+  if (start.date < (min(all.dates$date) - delta)) {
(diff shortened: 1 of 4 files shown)


infotroph (changes requested):
Thanks for working on this -- I agree it's caused by #2274 and the code in question is at least partly my fault.

Unfortunately this patch won't fix the underlying issue. Consider a file with hourly observations whose last timestamp is `2024-12-31 23:00:00`, for which an end time of 23:59:59 would still complain. Being able to pass an earlier end time is a useful workaround for this bug, but we should make sure bare dates without times do still work.

I think the place to fix this is in the same block of `load.cfmet` that @ashiklom suspected in #2274, probably by identifying the file timestep (e,g, `tstep <- mean(diff(all.dates$date))` and only signaling an error if the requested start/end is more than a full timestep outside the first/last timestamp in the file. I suspect once that's changed, the existing `met2model.BIOCRO` code may perform as intended without changes.

modules/data.atmosphere/R/load.cfmet.R
```diff
@@ -51,11 +51,18 @@ load.cfmet <- function(met.nc, lat, lon, start.date, end.date) {
 
   all.dates <- data.frame(index = seq_along(time.idx), date = date)
   
-  if (start.date + lubridate::days(1) < min(all.dates$date)) {
-   PEcAn.logger::logger.severe("run start date", start.date, "before met data starts", min(all.dates$date))
+  if (nrow(all.dates) > 1) {
+    delta <- stats::median(diff(all.dates$date), na.rm = TRUE)
+  } else {
+    delta <- as.difftime(1, units = "hours")  
   }
```
infotroph: I think one-timepoint files will be very rare; instead of testing for them as part of the happy path, handling them as part of a more general exception feels clearer to me:

```suggestion
  delta <- stats::median(diff(all.dates$date), na.rm = TRUE)
  if (is.na(delta)) {
    # probably only happens with a one-line met file
    # fall back to requiring exact match
    delta <- 0
  }
```

infotroph (commented):
Thanks for the quick pivot! I see one more block that was confusing to me -- does my suggested edit work for the case you were thinking of?

modules/data.atmosphere/R/load.cfmet.R
AritraDey-Dev: Would hardcoding 0 as the fallback for delta be correct in this context? Since the timestep is likely meant to represent a duration, would it be more appropriate to default to as.difftime(1, units = "hours") instead?

infotroph: I agree that delta represents a duration, but if we can't discern it from the file then we don't even know what unit to use -- it could be seconds, days, months, etc.

Falling back to 0 will give essentially the current behavior: It will fail if the last timestamp in the file isn't >= the requested end time. In the one-timestep case that means the user can get past this by passing identical start and end times, which seems reasonable to me -- and maybe even a useful check on whether they realized they were asking for only one timepoint.

AritraDey-Dev: > but if we can't discern it from the file then we don't even know what unit to use -- it could be seconds, days, months, etc.

yes this makes sense

infotroph (approved):

PR #3579, opened 2025-07-22
checks: file-path for met.proces in do_conversions

harshagr70:
This PR adds a simple check in do_conversions() to skip calling met.process() if a valid path is already supplied in the settings.

✅ Avoids unnecessary processing step
✅ Improves performance and clarity 
✅ Keeps met.process() callable directly if needed

This helps streamline workflows where users provide pre-processed met data.

### DIFF
--- a/base/workflow/R/do_conversions.R
+++ b/base/workflow/R/do_conversions.R
@@ -104,7 +104,6 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
       if (!is.null(settings$run$inputs$met$path)) {
       PEcAn.logger::logger.info("Skipping met.process: model-ready path provided")
       settings$run$inputs[[i]] <- settings$run$inputs$met
-      needsave <- TRUE
 
       }else if ( (PEcAn.utils::status.check(name) == 0)) { ## previously is.null(input$path) && 
         PEcAn.logger::logger.info("calling met.process: ",settings$run$inputs[[i]][['path']])
--- a/base/workflow/R/do_conversions.R
+++ b/base/workflow/R/do_conversions.R
@@ -98,8 +98,15 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     # met conversion
     
     if (input.tag == "met") {
+
       name <- "MET Process"
-      if ( (PEcAn.utils::status.check(name) == 0)) { ## previously is.null(input$path) && 
+
+      if (!is.null(settings$run$inputs$met$path)) {
+      PEcAn.logger::logger.info("Skipping met.process: model-ready path provided")
+      settings$run$inputs[[i]] <- settings$run$inputs$met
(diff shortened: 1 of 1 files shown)


base/workflow/R/do_conversions.R
```diff
@@ -98,8 +98,15 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
     # met conversion
     
     if (input.tag == "met") {
+
       name <- "MET Process"
-      if ( (PEcAn.utils::status.check(name) == 0)) { ## previously is.null(input$path) && 
+
+      if (!is.null(settings$run$inputs$met$path)) {
+      PEcAn.logger::logger.info("Skipping met.process: model-ready path provided")
+      settings$run$inputs[[i]] <- settings$run$inputs$met
```
mdietze, line 106: Is this line needed? seems like you're just overwriting settings$run$inputs$met with settings$run$inputs$met
```suggestion
```

base/workflow/R/do_conversions.R
```diff
@@ -98,8 +98,15 @@ do_conversions <- function(settings, overwrite.met = FALSE, overwrite.fia = FALS
...
+      needsave <- TRUE
```
mdietze: Pretty sure that you don't need to save an updated xml if you haven't changed anything

```suggestion
```

mdietze (changes requested):

infotroph:
Notes from live discussion where we decided to close this:

* When a path is provided, met.process will already exit quickly (though the check is indirect -- this might deserve some further attention)
* do_conversions is mainly concerned with checking which input types exist at all, and lets each conversion function decide what conversions are needed. Calling met.process and letting it be a near-no-op seems more in line with that design.
* We could imagine a design where do_conversions is responsible for more extensive checks of every input before passing them off to the input-specific function, but that would be a bigger redesign and should be done all at once instead of one input at a time.

@harshagr70 will dig further into met.process and
 - see if the check for a valid path needs refinement
 - consider if it can return anything useful when no db params are provided.

mdietze:
Personally, I think a generic do.conversions check that skips input processing if files are already provided seems like a good idea and doesn't seem like it would require a substantial redesign to be able to generalize beyond just met. This seems safer than requiring every single processing function to independently check whether files need to be processed or not. I could also see adding a "force" tag that would ignore this check and trigger the input conversion functions to run even if paths are provided.

mdietze:
reopen?

infotroph:
Can't reopen because the branch is already deleted, but I see your point. @harshagr70, how much work would it be to add this check to each input type (presumably at the top of the loop)? Will "path exists" be a sufficient condition to check for every input type?

harshagr70:
@infotroph it wont be much painful to make the changes again , or centrally for all the tags , but the bigger question that i still have in my mind is , will this path be enough !!?

PR #3584, opened 2025-07-26
Enhanced ERA5 data processing: generalize ERA5 extraction and CF conversion to support both ensemble and reanalysis data

divine7022:
## Description
This PR enhances the functionality of `extract.nc.ERA5` and `met2CF.ERA5` to support both ERA5 ensemble and reanalysis datasets. Originally, the functions were tailored only for ensemble data. And refactor the code to fix several issues.
- previously, the `met2CF.ERA5` function made assumptions about the presence and order of ERA5 variables (e.g, `ssrd, tp, d2m, sp`), and a fixed 3-hour timestep. So now made timestep (`1h, 3h, etc.`) to detected dynamically not just 3 hours.
- the function attempted to calculate specific humidity even if required variables (`t2m, d2m, sp`) were missing, causing errors. now only calculated if all required variables are present.
- reanalysis data is typically stored as a 3D , dimensions `[longitude, latitude, time]`. Works correctly with `raster::brick()` which expects this format.
- ensemble data uses a 4D , dimensions `[longitude, latitude, time, ensemble_member]` . `raster::brick()` fails or extracts only a slice unless explicitly adapted. this PR introduces conditional logic to detect and correctly extract the ensemble dimension.
- dynamically extracts origin from units attribute (`hours since...` or `seconds since...`) to construct correct time sequences.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -45,21 +57,41 @@ extract.nc.ERA5 <-
...
            ...) {
+    
+
+    years <- seq(lubridate::year(start_date),
+                 lubridate::year(end_date),
+                 1
+    )
+    sample_file <- file.path(in.path, paste0(in.prefix, years[1], ".nc"))
+    if (!file.exists(sample_file)) {
+      PEcAn.logger::logger.severe(paste0("Sample file not found: ", sample_file))
```
mdietze: I find referring to the original met file as the "sample file" to be confusing. If I saw this error message I wouldn't know what it was talking about.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -45,21 +57,41 @@ extract.nc.ERA5 <-
...
+    }
+    
+    # Determine data type (ensemble vs reanalysis)
+    nc_test <- ncdf4::nc_open(sample_file)
+    is_ensemble <- "number" %in% names(nc_test$dim) || any(sapply(nc_test$var, function(v) v$ndims == 4))
+    ncdf4::nc_close(nc_test)
+    
+    if (is_ensemble) {
+      ensemblesN <- seq(1, 10)
```
mdietze: might not hurt to detect the ensemble size from the nc file rather than assuming it will always be 10

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -85,18 +150,33 @@ extract.nc.ERA5 <-
       all.data.point <- 
         foreach::foreach(vname = vars, 
                          .packages=c("Kendall", "ncdf4")) %dopar% {
+                           nc_data <- ncdf4::nc_open(ncfile)
+                           on.exit(ncdf4::nc_close(nc_data), add = TRUE)
```
mdietze, line 196: will this close the connection at the end of the foreach or at the end of the overall function? If the latter, it would be easy to accidentally open too many connections and throw an error.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -130,22 +210,24 @@ extract.nc.ERA5 <-
...
-                           )
-                           out %>% purrr::map(~.x[['file']]) %>% unlist
-                         }
+                         .packages=c("Kendall", "ncdf4", "PEcAn.data.atmosphere", "purrr", "xts", "lubridate"),
+                         .options.snow=opts,
+                         .export = c("met2CF.ERA5")) %dopar% {
+                                       # Calling the met2CF inside extract bc in met process met2CF comes before extract !
+                                       out <- met2CF.ERA5(
+                                         slat[s.ind],
+                                         slon[s.ind],
```
mdietze: can met2CF.ERA5 take in a vector of locations, because that would be much more efficiently that k copies of the function each trying to separately open the same file in parallel

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -130,22 +210,24 @@ extract.nc.ERA5 <-
...
+                                         paste0(year,"-01-01"),
+                                         paste0(year,"-12-31"),
```
mdietze: Is there a reason we're assuming all years are complete and that we need entire years? This assumption was usually valid when met.process was first written, but these days its pretty common to do short-term nowcasts and forecasts.

modules/data.atmosphere/R/extract_ERA5.R
divine7022: yes, the `on.exit()` ensures the netcdf connection is closed when each `%dopar%` iteration completes, regardless of how it exits. each `%dopar%` iteration runs in a _separate environment/process_, so, 
process isolation prevents sharing file handles between workers and automatic cleanup occurs when each worker process completes.

modules/data.atmosphere/R/extract_ERA5.R
divine7022: thank you for the suggestion about vectorizing `met2CF.ERA5` to handle multiple locations. I've analyzed this approach carefully, and I believe we should retain the current parallel-by-site processing for the following reasons,
in current **parallel approach** for (eg, `~125 seconds for 100 sites (8 cores) `) and **vectorized sequential approach** for (eg, `~1000 seconds for 100 sites` ) .so the parallel approach is ~8x faster, which far outweighs any function call overhead.

however, if you still suggest removing the **parallel processing** approach and moving to **sequential processing** by passing entire vectors, please let me know . I'm open to understanding if there are other considerations

modules/data.atmosphere/R/extract_ERA5.R
divine7022: fixed 👍

modules/data.atmosphere/R/extract_ERA5.R
divine7022: fixed - now detects actual ensemble size dynamically

modules/data.atmosphere/R/extract_ERA5.R
divine7022: fixed - now supports partial years and custom date ranges

divine7022:
@dlebauer could you please take a look ? i have added ERA5 entry to `pecan_standard_met_table` and updated `met2CF.ERA5` to use `pecan_standard_met_table`

dlebauer (approved):
Looks great!

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -45,38 +56,128 @@ extract.nc.ERA5 <-
...
+          } else {
+            if (verbose) PEcAn.logger::logger.info("processing ERA5 reanalysis data")
+          }
+        }, error = function(e) {
+          if (verbose) PEcAn.logger::logger.info("processing ERA5 reanalysis data")
+        })
+      } else {
+        if (verbose) PEcAn.logger::logger.info("processing ERA5 reanalysis data")
+      }
+    }
```
infotroph: Having a chain of three different conditions that all emit the same message seems pretty confusing to the user

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -45,38 +56,128 @@ extract.nc.ERA5 <-
...
-      # set the vars
+      
+      # handle different time units: 'time' uses hours, 'valid_time' uses seconds
+      if (time_var == "time") {
+        # traditional format: "hours since YYYY-MM-DD HH:MM:SS"
+        timestamp <- as.POSIXct(t * 3600, tz = "UTC", origin = tustr[[1]][3])
+      } else {
+        # new format: "seconds since YYYY-MM-DD HH:MM:SS" (typically 1970-01-01)
+        timestamp <- as.POSIXct(t, tz = "UTC", origin = tustr[[1]][3])
+      }
```
infotroph:
```suggestion
PEcAn.utils::ud_convert(t, tunits, paste("seconds since", tustr[[1][3]])
```

infotroph: Actually both current and suggested will ignore the HH:MM:SS component of tunits. Maybe `tustr <- strsplit(tunits, " since ")`?

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -85,18 +186,33 @@ extract.nc.ERA5 <-
...
+                                 xmx = max(nc_data$dim$longitude$vals),
+                                 ymn = min(nc_data$dim$latitude$vals),
+                                 ymx = max(nc_data$dim$latitude$vals),
+                                 crs = "+proj=longlat +datum=WGS84"
+                               )
+                               raster::setZ(brick.tmp, timestamp)  
+                             } else {
+                               brick.tmp <- raster::brick(ncfile, varname = vname)
+                               raster::setZ(brick.tmp, timestamp)
+                             }
```
infotroph, line 226: Does this change the output format compared to the previous version? I can't immediately tell by reading this. If so, this should be loudly advertised as a breaking change -- multiple people have downstream code that would be affected by it.

modules/data.atmosphere/R/extract_ERA5.R
```diff
@@ -85,18 +186,33 @@ extract.nc.ERA5 <-
...
                              nn <-
                                raster::extract(brick.tmp,
                                                sp::SpatialPoints(cbind(slon, slat)),
                                                method = 'simple')
                              # replacing the missing/filled values with NA
                              nn[nn == nc_data$var[[vname]]$missval] <- NA
                              # send out the extracted var as a new col
-                             ens.out[[ens]] <- t(nn)
+                             ens.out[[ens]] <- t(nn)
```
infotroph:
```suggestion
                             ens.out[[ens]] <- t(nn)
```

modules/data.atmosphere/R/extract_ERA5.R
infotroph: This might be a case where the fastest answer depends a lot on the conditions of the test, e.g. a small-to-medium file on fast local disk vs. a cluster environment where the a big source file is accessed through a network filesystem or is otherwise disk-bound rather than CPU-bound. 

Not asking you to try to test every possible setup, but the latter example is one where we've seen trouble before and is worth at least thinking through.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -6,97 +6,174 @@
...
+                       long,
+                       start_date,
+                       end_date,
+                       sitename,
+                       outfolder,
+                       out.xts,
+                       overwrite = FALSE,
+                       verbose = TRUE,
+                       is_ensemble = TRUE,
+                       ens_size = 10) {
```
infotroph: Note that you could get the same result with fewer parameters by setting ens_size = 1 for the reanalysis case and letting everything else iterate over "each of the 1 ensemble members".

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -6,97 +6,174 @@
...
+    era5_tbl <- pecan_standard_met_table %>%
+      dplyr::filter(!is.na(.data$era5) & nzchar(.data$era5))
+    
+    if (nrow(era5_tbl) > 0) {
+      era5_to_cf <- stats::setNames(era5_tbl$cf_standard_name, era5_tbl$era5)
+      cf_units_map <- stats::setNames(era5_tbl$units, era5_tbl$cf_standard_name)
+    }
+  }
+  
+  # fallback mappings if table is missing or empty
```
infotroph: The table is built into the same package, so it seems pretty safe to assume it's available any time this function is. Putting fallbacks here seems more confusing than useful.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -6,97 +6,174 @@
...
+        }
+        # attach timestep as attribute for downstream use
+        attr(ens_cf, "timestep_hours") <- timestep_hours
+        return(ens_cf)
+        
       },
       error = function(e) {
         PEcAn.logger::logger.severe("Something went wrong during the unit conversion in met2cf ERA5.",
                                     conditionMessage(e))
+        return(NULL)
```
infotroph: logger.severe always throws an error, so this line won't be reached.
```suggestion
```

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -142,25 +215,20 @@ met2CF.ERA5<- function(lat,
...
-          #Each ensemble gets its own file.
-          time_dim = ncdf4::ncdim_def(
+          time_vals <- as.numeric(zoo::index(data.for.this.year.ens))
+          time_dim <- ncdf4::ncdim_def(
             name = "time",
-            paste(units = "hours since", format(start_date, "%Y-%m-%dT%H:%M")),
-            seq(0, (length(zoo::index(
-              data.for.this.year.ens
-            )) * 3) - 1 , length.out = length(zoo::index(data.for.this.year.ens))),
+            units = "seconds since 1970-01-01 00:00:00",
```
infotroph, line 191: Why the change from hours since start_date?

infotroph (commented):
Thanks for taking this on. I have some questions about compatibility with previously existing met2cf outputs -- if the format has changed that's potentially a big deal, because lots of folks have stashes of existing CF output that they'd like to keep using and add to. See inline for specifics.

modules/data.atmosphere/R/extract_ERA5.R
dlebauer: This is an interesting discussion and good points, but this level of optimization (vectorized vs. parallel processing) seems outside the scope of this quite valuable PR. To keep scope clear and minimize delay, I’d suggest saving this for future iterations - lerhaps creating a separate feature request labeled as “good first issue” that includes exploring tradeoffs in different environments.

modules/data.atmosphere/R/met2cf.ERA5.R
```diff
@@ -6,97 +6,174 @@
...
+      "air_pressure" = "Pa",
+      "precipitation_flux" = "kg m-2 s-1",
+      "eastward_wind" = "m s-1",
+      "northward_wind" = "m s-1",
+      "surface_downwelling_shortwave_flux_in_air" = "W m-2",
+      "surface_downwelling_longwave_flux_in_air" = "W m-2",
+      "specific_humidity" = "1",
+      "volume_fraction_of_condensed_water_in_soil" = "1"
+    )
+  }
```
dlebauer: Can this condense to:
```suggestion
era5_tbl <- pecan_standard_met_table %>%
  dplyr::filter(!is.na(.data$era5) & nzchar(.data$era5))

era5_to_cf <- stats::setNames(era5_tbl$cf_standard_name, era5_tbl$era5)
cf_units_map <- stats::setNames(era5_tbl$units, era5_tbl$cf_standard_name)
```

dlebauer (commented):
Getting closer!

modules/data.atmosphere/R/extract_ERA5.R
divine7022: yeah simplified 👍

modules/data.atmosphere/R/extract_ERA5.R
divine7022: good catch! now splits by `" since "`  and capture `YYYY-MM-DD HH:MM:SS` 👍

modules/data.atmosphere/R/extract_ERA5.R
divine7022: this is a valid concern but
this is a work around where raster support for 3D(`[lat ,lon, time*ens ]`) and doesn't support of 4D (new ens formate `[lat, lon, valid_time, number]`
Currently fixed the edge case for old ens format `[lat ,lon, time*ens ]` and it work reliably for legacy format too..

modules/data.atmosphere/R/extract_ERA5.R
divine7022: 👍

modules/data.atmosphere/R/met2cf.ERA5.R
divine7022: yeah removed! Now use case where
`if (ens_size > 1) { #ens } else { #rean }` in `met2cf.ERA5` but preserves in `extract.nc.ERA5` for robust flexible format detections.

modules/data.atmosphere/R/met2cf.ERA5.R
divine7022: removed 👍

modules/data.atmosphere/R/met2cf.ERA5.R
divine7022: removed 👍

modules/data.atmosphere/R/met2cf.ERA5.R
divine7022: The change from `"hours since start_date"` to `"seconds since 1970-01-01"` was made to address fundamental compatibility issues with ERA5 format diversity
so i fixed 3-hour interval assumption `seq(0, (n*3)-1, length.out=n)` assumes exactly 3-hour spacing but ERA5 data can be 1-hourly, 3-hourly, 6-hourly, or daily
```
old ERA5 NetCDF Header:
dimensions: longitude, latitude, time
time:units = "hours since 1900-01-01 00:00:00.0"

new ERA5 NetCDF Header:  
dimensions: longitude, latitude, valid_time [, number]
valid_time:units = "seconds since 1970-01-01"
```

Current approach handles both formats by converting to a standardized Unix timestamp format

`as.numeric(zoo::index(data))` preserves exact timing from processed data and works with any ERA5 format (old/new) and any time interval 
and ECMWF documentation confirms ERA5 time format changed from "hours since 1900-01-01" to "seconds since 1970-01-01", making Unix timestamps the most compatible choice.
https://confluence.ecmwf.int/display/CKB/GRIB+to+netCDF+conversion+on+new+CDS+and+ADS+systems

modules/data.atmosphere/R/extract_ERA5.R
infotroph: Point taken about scope. 

But re labels, a detailed optimization question that requires evaluating performance tradeoffs across multiple use cases and hardware configurations, and therefore also needs an understanding of the the whole workflow to know which cases to prioritize, is a great example of something that is _not_ a good first issue.

PR #3585, opened 2025-07-26
Add the `gdal_conversion` function for remote sensing image conversion using GDAL utility.

DongchenZ:
This function provides conversions of any GDAL-supported format.
Currently tested H5, NetCDF, HDF4, and GeoTIFF formats. 
This function should be ready for any GDAL-supported image format.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


mdietze:
Function looks good, but you need to add the gdal util package to the DESCRIPTION under Suggests and also update the logs, version number, etc. There's also a warning about needing to break the Example up over multiple lines

```
 Execution halted
  R check of modules/data.remote reports the following new problems. Please fix these and resubmit:
  '::' or ':::' import not declared from: ‘gdalUtils’
  checking Rd line widths ... NOTE
  Rd file 'gdal_conversion.Rd':
    \examples lines wider than 100 characters:
       band_names <- gdal_conversion(in_path = in_path, outfolder = outfolder, band_name = NULL, just_band_name = T)
       f <- gdal_conversion(in_path = in_path, outfolder = outfolder, band_name = band_names[1], just_band_name = F, target_format = ".tif")
```

modules/data.remote/R/merge_image_tiles.R
```diff
@@ -102,4 +102,75 @@ merge_image_tiles <- function(folder.path,
...
+#' f <- 
+#'   gdal_conversion(in_path = in_path, 
+#'     outfolder = outfolder, 
+#'     band_name = band_names[1], 
+#'     just_band_name = F, 
+#'     target_format = ".tif")
+#' }
+gdal_conversion <- function(in_path, outfolder = NULL, band_name = NULL, just_band_name = T, target_format = ".tif") {
+  # check if package has been installed.
+  if (!"gdalUtils" %in% installed.packages()) {
```
infotroph: Thanks for including a check for availability! But `?installed.packages` recommends against this usage:

> It will be slow when thousands of packages are installed, so do not use it to find out if a named package is installed (use [find.package](http://localhost:31789/library/utils/help/find.package) or [system.file](http://localhost:31789/library/utils/help/system.file)) nor to find out if a package is usable (call [requireNamespace](http://localhost:31789/library/utils/help/requireNamespace) or [require](http://localhost:31789/library/utils/help/require) and check the return value) nor to find details of a small number of packages (use [packageDescription](http://localhost:31789/library/utils/help/packageDescription)).

I usually use requireNamespace for this:

```suggestion
  if (!requireNamespace("gdalUtils", quietly = TRUE)) {
```

modules/data.remote/R/merge_image_tiles.R
```diff
@@ -102,4 +102,75 @@ merge_image_tiles <- function(folder.path,
...
+#'     just_band_name = T)
+#' # try convert the first band of the available band names to GeoTIFF file.
+#' f <- 
+#'   gdal_conversion(in_path = in_path, 
+#'     outfolder = outfolder, 
+#'     band_name = band_names[1], 
+#'     just_band_name = F, 
+#'     target_format = ".tif")
+#' }
+gdal_conversion <- function(in_path, outfolder = NULL, band_name = NULL, just_band_name = T, target_format = ".tif") {
```
infotroph:
```suggestion
gdal_conversion <- function(in_path, outfolder = NULL, band_name = NULL, just_band_name = TRUE, target_format = ".tif") {
```

infotroph (commented):
Thanks. Your code looks good, but I fear getting `gdalUtils` into the PEcAn stack is going to be a challenge -- gdalUtils has been archived on CRAN for three years and the GitHub version of it still imports `rgdal`, which is itself retired. Is there a way to do these conversions using sf or terra?

modules/data.remote/R/merge_image_tiles.R
DongchenZ: Fixed.

modules/data.remote/R/merge_image_tiles.R
DongchenZ: Fixed.

DongchenZ:
> Thanks. Your code looks good, but I fear getting `gdalUtils` into the PEcAn stack is going to be a challenge -- gdalUtils has been archived on CRAN for three years and the GitHub version of it still imports `rgdal`, which is itself retired. Is there a way to do these conversions using sf or terra?

modules/data.remote/R/merge_image_tiles.R
DongchenZ: @infotroph Thanks for the comments. I just revised the PR so that the `gdalUtils` functions are directly added to the `data.remote` package instead of being called from the `gdalUtils` package.

infotroph: From a dependency management perspective this would be a great direction, but it raises a new blocker: gdalUtils has a GPL2 license, which does not allow incorporating the code into BSD-licensed projects like PEcAn. I think that to do this we'd be required to relicense the entire data.remote package as GPL.

DongchenZ: Got it! I updated the PR such that we are not relying on any functions from the `gdalUtils` package. And it's cleaner too.

infotroph: Nice! I like this. Thanks for all the quick edits here.

infotroph (approved):

PR #3586, opened 2025-07-29
Add AmeriFlux meteorological data processing and ensemble generation with ERA5 fallback

divine7022:
## Description

This PR introduces the `AmeriFlux_met_process()` function, a comprehensive meteorological data processing pipeline with **ERA5 fallback capabilities** for missing **radiation** and **soil moisture** data, including automated gap-filling and ensemble generation.

- **complete processing pipeline**: **`downloads -> extraction(state drivers) -> CF conversion -> gap-filling -> ensemble generation`**
- detects insufficient data coverage and supplements with **ERA5 reanalysis** data

- ERA5 Fallback Logic :
The ERA5 fallback addresses critical limitations in AmeriFlux data processing:
   - **Radiation fallback** - `metgapfill` function requires at least one of **PAR** (`surface_downwelling_photosynthetic_photon_flux_in_air`) or **shortwave radiation (Rg)** (`surface_downwelling_shortwave_flux_in_air`) to successfully gap-fill other meteorological drivers. Fallback is triggered when **both PAR and Rg have < 50% coverage** or when both variables are missing entirely.

- **Soil moisture fallback** -  since `metgapfill` only processes atmospheric variables and cannot handle the complex spatial and physical dynamics of soil moisture (`volume_fraction_of_condensed_water_in_soil`), ERA5 fallback is triggered for **any missing values** in this variable

- Enhancements to `met_temporal_downscale.Gaussian_ensemble()`:
    - fixed critical netcdf writing bug, which corrected variable writing from index-based to name-based access, resolving incorrect column detection that caused data loss
    - changed `specific_humidity` units from "g/g" to "kg/kg" a CF standards and ensure proper statistical processing
    - Improved statistical processing 
    - resolved `specific_humidity` NA issue - variables now properly maintain temporal variability instead of returning all NA values

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


divine7022:
@dlebauer @infotroph if you’ve got a moment, I'd really appreciate a quick look at this PR

modules/data.atmosphere/DESCRIPTION
```diff
@@ -25,6 +25,7 @@ Imports:
     assertthat,
     arrow,
     curl,
+    data.table,
```
infotroph: Please undo and use dplyr or base R instead -- data.table is a good package, but it works differently enough from dplyr and base R that we have been working to remove it so that maintainers don't have to keep as many APIs in their heads while editing.

modules/data.atmosphere/R/Amerfilux_met_process.R
```diff
@@ -0,0 +1,430 @@
+#' @title Process AmeriFlux meteorological data
+#' @description Downloads AmeriFlux data, applies ERA5 fallback for missing radiation and soil moisture,
+#' performs gap-filling, and generates ensembles.
+#' This function provides a complete pipeline from raw AmeriFlux data to CF-compliant
+#' ensemble meteorological files.
```
infotroph: I strongly encourage _not_ putting this entire pipeline into one function. It's very common for folks to want to download raw data products separately from the processing steps, experiment with different strategies for the scaling and gapfilling steps, etc.

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
```diff
@@ -40,7 +40,7 @@ met_temporal_downscale.Gaussian_ensemble <- function(in.path, in.prefix, outfold
                                  "surface_downwelling_longwave_flux_in_air", "air_pressure", "surface_downwelling_shortwave_flux_in_air", 
                                  "eastward_wind", "northward_wind", "specific_humidity", "precipitation_flux"), 
                     units <- c("Kelvin", "Kelvin", "Kelvin", "W/m2", "Pascal", "W/m2", "m/s", 
-                               "m/s", "g/g", "kg/m2/s"))
+                               "m/s", "kg/kg", "kg/m2/s"))
```
infotroph: Why this unit change? The [CF canonical unit for specific_humidity is "1"](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html), so these both seem fully equivalent

modules/data.atmosphere/R/Amerfilux_met_process.R
infotroph: I also suspect it will be _easier_ to build this pipeline by composition of existing tools than as a single function, but if keeping the one-function design then I suggest a name that makes it more clear this is for creating one specific set of outputs rather than a generalized workflow for all Ameriflux work.

modules/data.atmosphere/R/Amerfilux_met_process.R
```diff
@@ -0,0 +1,480 @@
...
+  # input validation
+  if (!grepl("@", useremail)) {
+    PEcAn.logger::logger.severe("useremail must contain '@' for AmeriFlux authentication")
+  }
+  
+  # validate ERA5 credentials if provided
+  has_era5 <- !is.null(era5_user) && !is.null(era5_key)
+  if (!has_era5 && verbose) {
+    PEcAn.logger::logger.info("ERA5 credentials not provided. ERA5 fallback disabled.")
+  }
```
infotroph: Notice that this means not only needing to configure two different set of credentials before calling, but also means the call will fail if either of two separate authentication services is down at run time.

modules/data.atmosphere/R/Amerfilux_met_process.R
```diff
@@ -0,0 +1,480 @@
...
+  dirs <- list(
+    downloads = file.path(outfolder, "downloads"),
+    extracted = file.path(outfolder, "extracted"),
+    cf_format = file.path(outfolder, "cf_format"),
+    gapfilled = file.path(outfolder, "gapfilled"),
+    ensembles = file.path(outfolder, "ensembles")
+  )
+  if (has_era5) {
+    dirs$era5_downloads <- file.path(outfolder, "era5_downloads")
+    dirs$era5_processed <- file.path(outfolder, "era5_processed")
```
infotroph: Seems like this should be configurable so folks can use files they already have on disk

modules/data.atmosphere/R/Amerfilux_met_process.R
```diff
@@ -0,0 +1,480 @@
+#' @title Process AmeriFlux meteorological data
```
infotroph: typo in the filename

infotroph (commented):
My biggest overall comment is that this seems like it rebuilds a lot of pieces already found in the existing Ameriflux and ERA5 pipelines, so it may be more efficient to leverage those for all steps other than the new gapfilling functionality -- I suspect that approach will be more maintainable and more likely to be extendible to other data sources.

I assume the decision to use ERA5 specifically was a considered one that I don't need to revisit, but for the record I was surprised to see a relatively coarse-scale gridded product chosen for fallback of inherently site-specific tower data. Did you also consider finer-scale products (e.g Terraclimate) before settling on ERA5?

modules/data.atmosphere/R/met_temporal_downscale.Gaussian_ensemble.R
dlebauer: addressed in #3590; file removed from this PR

modules/data.atmosphere/R/Amerfilux_met_process.R
dlebauer: This is a great idea for future development

dlebauer:
Summarized (and elaborated) suggestions from @infotroph about refactoring in #3605

modules/data.atmosphere/R/Ameriflux_met_ensemble.R
```diff
@@ -0,0 +1,486 @@
+#' @title Generate AmeriFlux meteorological ensembles
+#' @description Downloads AmeriFlux data, applies ERA5 fallback for missing radiation and soil moisture,
+#' performs gap-filling, and generates ensembles.
+#' This function provides a complete pipeline from raw AmeriFlux data to CF-compliant
+#' ensemble meteorological files.
```
infotroph:
```suggestion
#' Generate AmeriFlux meteorological ensembles
#'
#' Downloads AmeriFlux data, applies ERA5 fallback for missing radiation and soil moisture,
#' performs gap-filling, and generates ensembles.
#' This function provides a complete pipeline from raw AmeriFlux data to CF-compliant
#' ensemble meteorological files.
#'
```

modules/data.atmosphere/R/Ameriflux_met_ensemble.R
```diff
@@ -0,0 +1,486 @@
...
+#'
+#' @author Akash
+#' @export
+
+AmeriFlux_met_ensemble <- function(site_id,
+                                   start_date,
+                                   end_date, 
+                                   outfolder,
+                                   username = "pecan", 
+                                   useremail = "@",
```
infotroph:
```suggestion
                                   ameriflux_username = "pecan", 
                                   ameriflux_useremail = "@",
```

modules/data.atmosphere/R/Ameriflux_met_ensemble.R
divine7022: done 👍

modules/data.atmosphere/R/Ameriflux_met_ensemble.R
divine7022: 👍

dlebauer (approved):
Nice, its all coming together now!

PR #3592, opened 2025-08-12
Debias branch

Sr31bu:
Description
This PR integrates a residual-based debiasing step into the State Data Assimilation (SDA) workflow, applied between the PEcAn model’s forecast and analysis phases. The correction uses prior-year observations, covariates, and raw model forecasts to train a two-model ensemble (KNN + Extra Trees) with optimal weighting. Predictions from this model are applied to adjust the forecast mean before assimilation, reducing systematic bias in the SDA cycle.

Key updates include:

Added debias.py implementing the residual correction pipeline.

Integrated Python model training and prediction into the SDA R loop via reticulate.

Implemented per-variable rolling training buffers so the model updates at each timestep with all available historical data.

Adjusted forecast means using predicted residuals before analysis step.

Preserved ensemble spread by applying correction only to the mean (not individual members).

Motivation and Context
The SDA process was showing persistent bias in forecasts, particularly when initialized from limited historical data. By learning residual patterns from previous timesteps and covariate information, this approach dynamically corrects the forecast mean, improving agreement with observations without distorting ensemble spread.


modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -394,8 +397,52 @@ sda.enkf.multisite <- function(settings,
     }
   }
   
-      
-
+  get_covariates_for_date <- function(settings, obs_date) {
+    yr <- lubridate::year(obs_date)
+    settings$covariates_df %>%
```
mdietze: This bit envisions the covariate dataframe as being attached to the settings object, but I don't see any place in the larger PR where the covariate file is opened, loaded, and attached to the settings object. Furthermore, I'd recommend passing that object as a separate function argument, not attaching it to the settings object, because the existing read/write functions for settings won't be able to handle it and because it makes the function more transparent about what it actually requires to run

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -394,8 +397,52 @@ sda.enkf.multisite <- function(settings,
     }
   }
   
-      
-
+  get_covariates_for_date <- function(settings, obs_date) {
```
mdietze: Add Roxygen documentation tags to this function

mdietze: Also, I'd strongly recommend against defining R functions within other R functions unless there's a real compelling reason to restrict their scope in such a way. Doing so makes the code shorter, more modular, and more readable

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -394,8 +397,52 @@ sda.enkf.multisite <- function(settings,
...
+    yr <- lubridate::year(obs_date)
+    settings$covariates_df %>%
+      dplyr::filter(year == yr) %>%
+      dplyr::right_join(dplyr::select(settings$site_coords, site), by = "site") %>%
+      dplyr::arrange(site)    # don't drop 'site' or 'year' here
+  }
+  reticulate::source_python("/projectnb/dietzelab/Shashank-PECAN/debias.py")  # your module with train_full_model/predict_residual
```
mdietze: 1. This file really needs to be part of the PR, not living separately from PEcAn on a local filesystem
2. once in pecan, use system.file to look up the system-independent file path e.g. system.file("python/debias.py",package="PEcanAssimSeq")

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -394,8 +397,52 @@ sda.enkf.multisite <- function(settings,
...
+  
+  obs_vec_for_time <- function(t_idx, site_index, col_vars, obs.mean, name_map = NULL) {
```
mdietze: as before, move ancillary functions to be external to the SDA function and document using roxygen

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -394,8 +397,52 @@ sda.enkf.multisite <- function(settings,
...
+    out
+  }
+  
+  cov_by_columns <- function(obs_date, site_index) {
+    df <- get_covariates_for_date(settings, obs_date)
+    idx <- match(site_index, df$site)  # repeat each site's row for each column of X
+    as.matrix(df[idx, setdiff(names(df), c("site","year")), drop = FALSE])
+  }
+  
+  name_map <- c(AGB = "AbvGrndWood",
```
mdietze: This sort of info needs to be extracted directly from the settings since the variables in any particular SDA could change from run to run and one shouldn't have to magically know to update this internal variable list and recompile pecan any time you add a new variable to the SDA.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -353,7 +355,8 @@ sda.enkf.multisite <- function(settings,
         load(file.path(settings$outdir, "samples.Rdata"))
       }
       #reformatting params
-      new.params <- sda_matchparam(settings, ensemble.samples, site.ids, nens)
+      new.params <- PEcAnAssimSequential:::sda_matchparam(settings, ensemble.samples, site.ids, nens)
```
mdietze: Namespaces aren't required for internal functions

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -486,7 +533,7 @@ sda.enkf.multisite <- function(settings,
         #put building of X into a function that gets called
         max_t <- 0
         while("try-error" %in% class(
-          try(reads <- build_X(out.configs = out.configs, 
+          try(reads <- PEcAnAssimSequential:::build_X(out.configs = out.configs,
```
mdietze: Namespaces aren't required for internal functions

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -495,7 +542,7 @@ sda.enkf.multisite <- function(settings,
                                t = t, 
                                var.names = var.names, 
                                my.read_restart = my.read_restart,
-                               restart_flag = restart_flag), silent = T))
+                               restart_flag = restart_flag), silent = F))
```
mdietze: Are you recommending this be a general change, or did you just turn this on for debugging and forget to switch it back for the PR?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
         }
         
       }  ## end else from restart & t==1
+      
+      raw_mean_t <- colMeans(X)
```
mdietze, line 653: Here you're adding a sizable chunk of code without documenting what its purpose is. Also, this seems like a place where it would be advantageous to put all of this inside a function, which would keep it more self-contained and also keep the main SDA code more readable.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+          y_v_all   <- obs_prev_vec[cols_v] - as.numeric(raw_prev[cols_v])
+          Xprev_all <- cbind(cov_prev_mat[cols_v, , drop = FALSE],
+                             raw = as.numeric(raw_prev[cols_v]))
+          mask <- !is.na(y_v_all) & complete.cases(Xprev_all)
+          if (any(mask)) {
+            # always (re)train this step using all rows seen so far for this var
+            # keep a per-var growing buffer if you want; or just train on Xprev_all[mask,]
+            # Here: simple “use everything we have so far” buffer
+            rec <- if (exists(v, train_buf, inherits = FALSE)) get(v, train_buf)
+            else list(X = NULL, y = NULL)
```
mdietze: recommend using pecan standard formatting (which is based on tidyverse) across the board, which includes things like this hard-to-read if/else

https://pecanproject.github.io/pecan-documentation/develop/coding-practices.html#developer-codestyle

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: Also, if you create a new settings argument (which I think should be the file path to this info) you need to update the general pecan documentation in the book_source folder

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+            rec$X <- rbind(rec$X, Xprev_all[mask, , drop = FALSE])
+            rec$y <- c(rec$y, y_v_all[mask])
+            assign(v, rec, train_buf)
+            train_full_model(name = as.character(v),X = as.matrix(rec$X), y= as.numeric(rec$y))
```
mdietze: Undefined function

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+          
+            }
+          
+          # predict residuals at t (only where features are complete)
+          if (has_model(v)) {
+            Xt_v <- cbind(cov_t_mat[cols_v, , drop = FALSE],
+                          raw = as.numeric(raw_mean_t[cols_v]))
+            ok <- complete.cases(Xt_v)  # <-- vectorized
+            if (any(ok)) {
+              pred_resid[cols_v[ok]] <- predict_residual(v, Xt_v[ok, , drop = FALSE])
```
mdietze: what happens to the cases where features are incomplete? Does it really make sense to assume there's no bias in those cases?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+          post = as.numeric(post_mean),
+          obs  = as.numeric(obs_t_vec),
+          stringsAsFactors = FALSE
+        )
+        
+        # order for readability
+        comp_df <- comp_df[order(comp_df$var, comp_df$site), ]
+        
+        # print a quick peek
+        cat("\n--- Debias diagnostics for", obs.t, "---\n")
```
mdietze: Messages should be sent through pecan logger functions. Also, I don't think this info should be shared at all times and sites by default

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+        utils::head(comp_df, 20) |> print()
+        
+        # print full tables per variable (handy in browser())
+        for (vv in unique(comp_df$var)) {
+          cat("\n=== Variable:", vv, "===\n")
```
mdietze: same points about logger and not showing debug info by default

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+        rmse_tbl <- do.call(rbind, lapply(split(comp_df, comp_df$var), function(d){
+          data.frame(
+            var       = d$var[1],
+            rmse_pre  = rmse(d$pre,  d$obs),
+            rmse_post = rmse(d$post, d$obs),
+            stringsAsFactors = FALSE
+          )
+        }))
+        cat("\n--- RMSE by variable (pre vs post) ---\n")
+        print(rmse_tbl)
```
mdietze: same points about logger and debug messages. Where are these diagnostic tables being stored for later analysis?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -571,7 +714,7 @@ sda.enkf.multisite <- function(settings,
             MCMC.args <- control$MCMC.args
           }
           #running analysis function.
-          enkf.params[[obs.t]] <- analysis_sda_block(settings, block.list.all, X, obs.mean, obs.cov, t, nt, MCMC.args, pre_enkf_params)
+          enkf.params[[obs.t]] <- PEcAnAssimSequential:::analysis_sda_block(settings, block.list.all, X, obs.mean, obs.cov, t, nt, MCMC.args, pre_enkf_params)
```
mdietze: revert

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +582,103 @@ sda.enkf.multisite <- function(settings,
...
+        cat("\n--- RMSE by variable (pre vs post) ---\n")
+        print(rmse_tbl)
+        
+        # stash for later inspection/export if you want
+        if (!exists("DIAG", inherits = FALSE)) DIAG <- list()
+        DIAG[[obs.t]] <- list(comp = comp_df, rmse = rmse_tbl)
+        
+        offsets   <- sweep(X, 2, raw_mean_t, FUN = "-")
+        corrected <- raw_mean_t + pred_resid
+        X         <- sweep(offsets, 2, corrected, FUN = "+")
```
mdietze: seems like there should be a flag on whether one wants to turn-on or -off the ML bias correction features, rather than assuming they are 'always on'

mdietze (changes requested):
Make sure to increase the amount of documentation, including roxygen, changelogs, and overall pecan book_source documentation

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,93 @@
...
+#' cov_df  <- generate_covariates_df(settings, cov_dir)
+#' dplyr::glimpse(cov_df)
+#' }
+#'
+#' @importFrom purrr map_df map2_dfr
+#' @importFrom dplyr mutate select filter arrange
+#' @importFrom tibble tibble as_tibble
+#' @importFrom stringr str_extract
+#' @importFrom terra vect rast extract
+#' @importFrom rlang .data
```
DongchenZ: Please use the namespace in the script instead of importing from the package in the roxygen. For example: `purrr::map_df`, `dplyr::mutate`, etc.

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,93 @@
...
+#' @export
```
DongchenZ: Probably not export to the namespace if this function is only used for debiasing.

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,93 @@
...
+generate_covariates_df <- function(settings, cov_dir, crs = "EPSG:4326") {
```
DongchenZ: Instead of inputting the `settings` object. I recommend putting the lat/lon dataframe as the input argument (`colnames <- c("lon", "lat")`).

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,93 @@
...
+  # --- checks ---
+  if (!dir.exists(cov_dir)) stop("`cov_dir` does not exist: ", cov_dir)
+  if (is.null(settings$run) || length(settings$run) == 0) {
+    stop("`settings$run` is missing or empty; cannot build site coordinates.")
+  }
```
DongchenZ, line 26: If you are importing the lat/lon dataframe, make sure to update this `if-control`.

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,93 @@
...
+  
+  # 1) Build site_coords and coerce lon/lat to numeric
+  site_coords <- purrr::map_df(settings$run, ~ tibble::tibble(
+    site = as.character(.x$site$id),
+    lat  = suppressWarnings(as.numeric(.x$site$lat)),
+    lon  = suppressWarnings(as.numeric(.x$site$lon))
+  ))
```
DongchenZ: If you are importing the lat/lon dataframe, make sure to remove this `site_coords` creation.

modules/assim.sequential/R/covariates_loader.R
DongchenZ: Probably replacing `settings` with `site_coords`.

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,93 @@
...
+    stop("Found non-numeric lat/lon for sites: ", paste(bad, collapse = ", "))
+  }
+  
+  # 2) Create the SpatVector
+  coords_mat <- as.matrix(site_coords[, c("lon", "lat")])
+  pts        <- terra::vect(coords_mat, type = "points", crs = crs)
+  pts$site   <- site_coords$site
+  
+  # 3) Discover years from filenames
+  tif_files <- list.files(cov_dir, pattern = "^covariates_\\d{4}\\.tiff$", full.names = TRUE)
```
DongchenZ: Probably wrap this covariate file pattern as the `prefix` argument of the function.

modules/assim.sequential/R/covariates_loader.R
DongchenZ: And I saw you already added the namespace for those functions. So you just need to delete those `@importfrom` from the roxygen.

modules/assim.sequential/R/debias_py.R
```diff
@@ -0,0 +1,76 @@
...
+#' The imported module is cached across calls. Pass `reload = TRUE` to force
+#' re-import (e.g., after editing the Python code in development).
+#'
+#' @param reload Logical; if `TRUE`, force re-import even if a cached module exists.
+#'
+#' @return A reticulate Python module object. Errors if no suitable path is found
+#'   or if the import returns a null Python pointer.
+#'
+#' @keywords internal
+#' @importFrom reticulate import_from_path py_is_null_xptr
```
DongchenZ: See my comments above for removing the `@importfrom` from the roxygen.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +599,40 @@ sda.enkf.multisite <- function(settings,
...
         
       }  ## end else from restart & t==1
+      
+      raw_mean_t <- colMeans(X)
+      site_index <- attr(X, "Site")
+      col_vars   <- colnames(X)
       FORECAST[[obs.t]] <- X
+      name_map <- debias_name_map
+      #DEBIAS STEP 
+      if (t > 1 && isTRUE(debias_mode)) {
```
DongchenZ: We should also be able to prescribe the starting year for the debiasing feature.

modules/assim.sequential/R/preprocessing_debias.R
```diff
@@ -0,0 +1,130 @@
...
+    dplyr::filter(.data$year == !!yr)
+  
+  # enforce presence & order of the sites we’re using this step
+  df_year <- df_year |>
+    dplyr::right_join(tibble::tibble(site = sites_used), by = "site") |>
+    dplyr::arrange(.data$site)
+  
+  # sanity check
+  if (any(is.na(df_year$year))) {
+    missing_sites <- df_year$site[is.na(df_year$year)]
```
DongchenZ: Here, we should not only skip sites that have missing covariates, but also sites with inconsistent observations between time points. For example, we should skip the site if we have the AGB observations in 2012 but no AGB observations in 2013.

modules/assim.sequential/R/preprocessing_debias.R
```diff
@@ -0,0 +1,130 @@
...
+  yr <- lubridate::year(obs_date)
+  sites_used <- unique(as.character(site_index))
+  
+  if (is.null(covariates_df)) {
+    stop("covariates_df is NULL. Provide a table with columns: site, year, <features...>.")
+  }
+  
+  df_year <- covariates_df |>
+    dplyr::filter(.data$year == !!yr)
+
```
DongchenZ: Need more descriptions above.

modules/assim.sequential/R/preprocessing_debias.R
```diff
@@ -0,0 +1,130 @@
...
+      data.frame(
+        var       = d$var[1],
+        rmse_pre  = rmse(d$pre,  d$obs),
+        rmse_post = rmse(d$post, d$obs),
+        stringsAsFactors = FALSE
+      )
+    })
+  )
+}
+
```
DongchenZ: Need more descriptions/comments for the above functions.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -393,9 +407,59 @@ sda.enkf.multisite <- function(settings,
...
+  #   time      : time label
+  #   var       : variable name
+  #   rmse_pre  : RMSE before debias (raw ensemble mean vs obs)
+  #   rmse_post : RMSE after debias (corrected mean vs obs)
+  DIAG <- list()  # per-time: DIAG[[time]] = list(comp=..., rmse=...)
+  RMSE_DF <- data.frame(
+    time      = character(),
+    var       = character(),
+    rmse_pre  = numeric(),
+    rmse_post = numeric(),
```
DongchenZ: Is it possible to report not just the RMSE, but also more metrics like R2 and bias?

DongchenZ (changes requested):
It's great to see the modular SDA debaing feature finally. The key gap here is the need not just to skip sites that have missing covariates, but also sites with inconsistent observations between time points. For example, we should skip the site if we have the AGB observations in 2012 but no AGB observations in 2013. And we will only train/predict/correct model forecasts when we have consistent observations.

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,464 @@
...
+    prev_union <- unique(unlist(lapply(seq_len(t_idx - 1L), observed_vars_at, site_id = s), use.names = FALSE))
+    if (length(prev_union) == 0) next
+    cur_vars <- observed_vars_at(t_idx, s)
+    if (length(setdiff(prev_union, cur_vars)) > 0) {
+      inconsistent <- c(inconsistent, s)
+    }
+  }
+  unique(inconsistent)
+}
+
```
DongchenZ, line 156: @Sr31bu could you add more descriptions for the above functions?

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,464 @@
...
+  
+  list(
+    X = X_new,
+    weights_entry   = if (length(weights_entry)) weights_entry else NULL,
+    weights_df_rows = weights_df_rows,
+    diag            = list(comp = comp_df, rmse = diag_metrics),
+    rmse_rows       = rmse_rows
+  )
+}
+
```
DongchenZ: Please add descriptions to the codes throughout this script.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -535,7 +626,44 @@ sda.enkf.multisite <- function(settings,
...
+          drop_incomplete_covariates = debias_drop_incomplete_covariates,   # <- fixed
+          enforce_consistent_obs     = debias_enforce_consistent_obs,       # <- fixed
+          require_obs_at_t_for_predict = debias_require_obs_at_t_for_predict # <- fixed
+        )
+        X <- out$X
+        if (!is.null(out$weights_entry)) DEBIAS_WEIGHTS[[obs.t]] <- out$weights_entry
+        if (nrow(out$weights_df_rows)) DEBIAS_WEIGHTS_DF <- rbind(DEBIAS_WEIGHTS_DF, out$weights_df_rows)
+        DIAG[[obs.t]] <- out$diag
+        if (nrow(out$rmse_rows)) RMSE_DF <- rbind(RMSE_DF, out$rmse_rows)
+
```
DongchenZ, line 692:
```suggestion
      # remove negative values after the bias-correction.
      offsets   <- sweep(X, 2, raw_mean_t, FUN = "-")
      corrected <- raw_mean_t + pred_resid
      X         <- sweep(offsets, 2, corrected, FUN = "+")
      for(i in 1:ncol(X)){
        int.save <- state.interval[which(startsWith(colnames(X)[i], var.names)),]
        X[X[,i] < int.save[1],i] <- int.save[1]
        X[X[,i] > int.save[2],i] <- int.save[2]
      }
      # add attributes to the X table.
      X <- X %>%
        `colnames<-`(c(rep(var.names, length(site.ids)))) %>%
        `attr<-`('Site',c(rep(site.ids, each=length(var.names))))
```

DongchenZ: @Sr31bu, please add this chunk of code to remove any negative values after the debiasing.

DongchenZ: And you also need to apply that to the continental SDA script: https://github.com/PecanProject/pecan/blob/90bd4d29d58ced64b6cce9fe8a1b4b52a6d546b0/modules/assim.sequential/R/sda.enkf_parallel.R#L25.

DongchenZ (approved):

modules/assim.sequential/R/debias_py.R
mdietze: Agree with removing importfrom and instead use namespaces instead unless there's some super compelling reason that wouldn't work (e.g., previous uses of importFrom were used for pipe operators like %>%)

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,654 @@
...
+# ------------------------------------------------------------------------------
+
+#' Map observation names -> state names
+#'
+#' This mapping is applied whenever observations are merged into the state layout.
+#' If your upstream naming changes, **edit here** to keep the rest of the code stable.
+#'
+#' @rdname debias_helpers
+#' @keywords internal
+debias_name_map <- c(
```
mdietze, line 46: Is this strictly needed? Can we just use the actual variable names throughout the code? Or can this be gleaned from the settings object? Having hard-coded mappings like this seems dangerous as future uses won't remember to update this (e.g. Cami has already added GWBI and has no idea this bit of code exists)

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,654 @@
...
+  SoilC = "TotSoilCarb"
+)
+
+# ------------------------------------------------------------------------------
+# (2) Site filtering utilities
+# ------------------------------------------------------------------------------
+
+#' Sites with complete covariates in a given year
+#'
+#' Returns the subset of `candidate_sites` whose covariate row at `year` has **no NA**
```
mdietze, line 78: How is this different from `na.omit`?

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,654 @@
...
+  
+  ok_mask <- rowSums(is.na(df_year[, cov_cols, drop = FALSE])) == 0L
+  df_year$site[ok_mask]
+}
+
+#' Sites that become inconsistent in observed variables over time
+#'
+#' A site is flagged **inconsistent at `t_idx`** if it is missing *any* variable at time
+#' `t_idx` that it had reported in **any earlier time** (1..`t_idx-1`). This prevents
+#' training/evaluation on sites that drop variables mid-series.
```
mdietze, line 113: I don't fully understand what it means for a site to become "inconsistent" or why this is a problem for the debiasing when there's already another function that removes sites if they are missing covariates. Why would missing data at an earlier time point make it such that you can't calculate or model residuals at this time step? Why would missing one variable affect your ability to calculate or model residuals for other variables?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -31,12 +31,21 @@
 #' `forceRun` decide if we want to proceed the Bayesian MCMC sampling without observations;
 #' `run_parallel` decide if we want to run the SDA under parallel mode for the `future_map` function;
 #' `MCMC.args` include lists for controling the MCMC sampling process (iteration, nchains, burnin, and nthin.).
+#' @param cov_dir Directory containing yearly covariate stacks named like "covariates_YYYY.tiff".
+#' @param debias_mode Logical; if TRUE, run the residual debias step each cycle (t > 1).
+#' @param debias_start_year Integer year (e.g., 2015). If set, debiasing starts at this year (inclusive).
```
mdietze: 1. Feels like having both debias_mode and debias_start_year is redundant, you could just set the start year to NULL to turn off debiasing
2. Alternatively/additionally, all the debias variables could be combined into a single object (e.g. named list). This is common for functions with lots of options, and we're already doing this with the `control` list arg.
3. debias_start_year should be generalized to debias_start_date since multiple people are running the SDA workflow with a sub-annual assimilation time step. [note: this could be left to a follow up PR if the change would be nontrivial to implement and test, but shouldn't be forgotten!]

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -122,6 +136,23 @@ sda.enkf.multisite <- function(settings,
...
   enkf.params <- list()
   restart.list <- NULL
+  covariates_df <- NULL
+  if (!is.null(cov_dir)) {
+    site_coords <- purrr::map_df(settings$run, ~ tibble::tibble(
+      site = as.character(.x$site$id),
+      lon  = suppressWarnings(as.numeric(.x$site$lon)),
+      lat  = suppressWarnings(as.numeric(.x$site$lat))
+    ))
+    covariates_df <- generate_covariates_df(
```
mdietze: For the future, it might be advantageous to load the covariates within the loop over time so that we only load the covariates that we need at any point in time rather than keeping them all in memory.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -393,15 +425,74 @@ sda.enkf.multisite <- function(settings,
...
+  #   learner : model id / learner name ("KNN", "TREE", etc.)
+  #   weight  : numeric weight for that learner at that time/variable
+  DEBIAS_WEIGHTS_DF <- data.frame(
+    time    = character(),
+    var     = character(),
+    learner = character(),
+    weight  = numeric(),
+    stringsAsFactors = FALSE
+  )
+  # Small helper to append one row per learner into DEBIAS_WEIGHTS_DF.
```
mdietze, line 471: Any reason this helper and the next can't be moved to the file that has all the other debias helper functions?

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: Don't resolve conversations without either answering the question or making the change.

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: @DongchenZ **ONLY** variables that the settings says are zero bound should be right censored. Also, the censoring has to be ZERO based (or based on the threshold value in the settings), not based on an arbitrary small number (e.g. 0.01). Similarly, only variables that are listed as having an upper bound should be right censored. 

Finally, your request that Shashank copy-and-paste all his changes into sda.enkf_parallel reinforces my point in your earlier PR that we really need to refactor and reconcile what has now become at least 4 different versions of the SDA into a SINGLE operational version. Copy-and-pasting code across versions is the wrong solution.

modules/assim.sequential/R/debias_step.R
Sr31bu: na.omit() drops rows from a data frame you already subsetted; here we need a per-year, per-candidate-sites check that returns the list of site ids to use consistently across both train (t−1) and predict (t).

the function:

filters covariates_df to year == Y and candidate_sites

tests completeness over feature columns only (excludes site/year)

returns the site ids that are complete

this is used to construct aligned matrices with a guaranteed, consistent site set across steps. na.omit() can’t directly produce that site-id set or enforce symmetry across t−1 and t.

mdietze:
@Sr31bu looks like there are just a few small issues that need to be resolved

R check of modules/assim.sequential reports the following new problems. Please fix these and resubmit:
  '::' or ':::' import not declared from: ‘tibble’
  checking package dependencies ... NOTE
  Imports includes 21 non-default packages.
  Importing from so many packages makes the package vulnerable to any of
  them becoming unavailable.  Move as many as possible to Suggests and
  use conditionally.
  checking Rd \usage sections ... NOTE
  Duplicated \argument entries in Rd file 'debias_helpers.Rd':
    ‘drop_incomplete_covariates’
  
  Functions with \usage entries need to have the appropriate \alias
  entries, and all their arguments documented.
  The \usage entries must correspond to syntactically valid R code.
  See chapter ‘Writing R documentation files’ in the ‘Writing R
  Extensions’ manual.
  Execution halted

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -152,6 +164,44 @@ sda.enkf.multisite <- function(settings,
     site.ids <- as.character(settings$run$site$id)
   }
   
+  # Build site coords once (used by covariate extraction)
+  site_coords <- purrr::map_df(settings$run, ~ tibble::tibble(
```
infotroph: We can avoid needing to add another package to our (already long) import list by using `dplyr::tibble` -- it's the same function, just re-exported by dplyr for exactly this reason.
```suggestion
  site_coords <- purrr::map_df(settings$run, ~ dplyr::tibble(
```

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,85 @@
...
+  # per-year extractor
+  extract_year <- function(tif_path, year) {
+    r    <- terra::rast(tif_path)
+    vals <- terra::extract(r, pts)
+    
+    if ("ID" %in% names(vals)) {
+      vals <- vals[, setdiff(names(vals), "ID"), drop = FALSE]
+    }
+    
+    out <- tibble::as_tibble(vals)
```
infotroph:
```suggestion
    out <- dplyr::as_tibble(vals)
```

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,691 @@
...
+#' Converts the per-site covariate data into a matrix aligned with the **columns of `X`**.
+#' For any column whose site was filtered out, the function inserts a row of `NA`
+#' features to preserve shape and ordering.
+#'
+#' @param covariates_df See `debias_get_covariates_for_date()`.
+#' @param obs_date Date for which to fetch covariates (year is used).
+#' @param site_index Site id per column of `X`.
+#' @param obs.mean Observation structure used if enforcing consistency.
+#' @param t_idx Time index associated with `obs_date`.
+#' @param drop_incomplete_covariates, enforce_consistent_obs See above.
```
mdietze, line 254: I'm not entirely sure what's happening with the documentation for these helper functions, but right now they're all being pushed into one big shared documentation file, which is then complaining that this specific @param is multiply defined (because its documented for a later function)

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,691 @@
...
+    return(matrix(numeric(0), nrow = length(site_index), ncol = 0))
+  }
+  
+  feat_cols <- setdiff(names(df_year), c("site", "year"))
+  idx <- match(as.character(site_index), df_year$site)
+  
+  # Filler row of NA to keep alignment when a site is missing
+  na_row <- as.list(rep(NA_real_, length(feat_cols)))
+  names(na_row) <- feat_cols
+  filler <- tibble::as_tibble_row(na_row)
```
mdietze:
```suggestion
  filler <- dplyr::as_tibble_row(na_row)
```

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,691 @@
...
+      stop("obs.mean and t_idx must be provided when enforce_consistent_obs = TRUE.")
+    }
+    inconsistent_sites <- debias_sites_inconsistent_obs(obs.mean, t_idx, name_map = debias_name_map)
+    eligible_sites <- setdiff(complete_sites, inconsistent_sites)
+  } else {
+    eligible_sites <- complete_sites
+  }
+  
+  if (length(eligible_sites) == 0) {
+    return(tibble::tibble(site = character(0), year = integer(0)))
```
mdietze:
```suggestion
    return(dplyr::tibble(site = character(0), year = integer(0)))
```

modules/assim.sequential/R/debias_step.R
infotroph: Turns out dplyr doesn't export as_tibble_row -- would just as_tibble(na_row) work? Looks like the main thing as_tibble_row gives you is guaranteed complaint if the inputs aren’t all exactly length one, but given you just constructed na_row  I think that’s covered in this particular case
```suggestion
  filler <- dplyr::as_tibble(na_row)
```

modules/assim.sequential/DESCRIPTION
```diff
@@ -23,13 +23,15 @@ Imports:
     Matrix,
     ncdf4,
     nimble,
+    tibble,
```
mdietze:
```suggestion
```

modules/assim.sequential/NAMESPACE
```diff
@@ -73,3 +73,6 @@ importFrom(foreach,"%dopar%")
 importFrom(lubridate,"%m+%")
 importFrom(magrittr,"%>%")
 importFrom(rlang,.data)
+importFrom(tibble,as_tibble)
+importFrom(tibble,as_tibble_row)
+importFrom(tibble,tibble)
```
mdietze:
```suggestion
```

modules/assim.sequential/R/debias_step.R
```diff
@@ -0,0 +1,696 @@
+# at the very top of that file
+#' @importFrom tibble tibble as_tibble as_tibble_row
```
mdietze:
```suggestion
```

mdietze (approved):

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,85 @@
...
+    if ("ID" %in% names(vals)) {
+      vals <- vals[, setdiff(names(vals), "ID"), drop = FALSE]
+    }
+    
+    out <- dplyr::as_tibble(vals)
+    if (nrow(out) != nrow(site_coords)) {
+      stop("Row mismatch for year ", year, ": expected ", nrow(site_coords),
+           " but got ", nrow(out), ". Check CRS/coordinates or raster extent.")
+    }
+
```
DongchenZ, line 79:
```suggestion
    site <- NULL
```

DongchenZ: Add this to avoid the namespace checking for the `site` variable.

modules/assim.sequential/R/debias_py.R
```diff
@@ -0,0 +1,75 @@
...
+      # Installed package roots
+      roots <- Filter(nzchar, c(
+        system.file("python",        package = pkg),
+        system.file("python_models", package = pkg)
+      ))
+      
+      # Dev fallbacks (if running from source)
+      if (!length(roots)) {
+        ns_path <- tryCatch(getNamespaceInfo(pkg, "path"), error = function(e) NA_character_)
+        roots <- unique(na.omit(c(
```
DongchenZ:
```suggestion
        roots <- unique(stats::na.omit(c(
```

modules/assim.sequential/R/covariates_loader.R
```diff
@@ -0,0 +1,85 @@
...
+    }
+    
+    out <- dplyr::as_tibble(vals)
+    if (nrow(out) != nrow(site_coords)) {
+      stop("Row mismatch for year ", year, ": expected ", nrow(site_coords),
+           " but got ", nrow(out), ". Check CRS/coordinates or raster extent.")
+    }
+    
+    dplyr::mutate(out, site = site_coords$site, year = as.integer(year)) |>
+      dplyr::select(site, year, dplyr::everything())
```
mdietze:
```suggestion
      dplyr::select("site", "year", dplyr::everything())
```

PR #3590, opened 2025-08-07
enhance met downscaling with adding a support for soil temp, RH, soil moisture, PPFD and refactored to address other critical issues

divine7022:
This PR improves the met downscaling process by introducing new variable support, correcting key bugs, and tightening up physical consistency checks.
## Description
- added `soil temp, RH, soil moisture, and PPFD` downscaling
- proper `Tmax/Tmin` gaussian ensemble with bounds checking
- physical bounds enforcement across all variables
- fixed critical netcdf writing bug, which corrected variable writing from index-based to name-based access, resolving incorrect column detection that caused data loss
- changed specific_humidity units from "g/g" to "kg/kg" , a CF standards and ensure proper statistical processing
- resolved `specific_humidity` NA issue - variables now properly maintain temporal variability instead of returning all NA values
- removed step that incorrectly overwrote `air_temperature_max/min` to NA
- and refactored to address other critical issues

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [ ] I agree that PEcAn Project may distribute my contribution under any or all of
	- the same license as the existing code,
	- and/or the BSD 3-clause license.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3597, opened 2025-08-14
Consolidate methods and standardize units for calculating saturated vapor pressure deficit

dlebauer:
consolidate get.es, SatVapPres functions and a new method proposed in #3590 into a single saturated_vapor_pressure function with args

❓ is there a preferred method to use as default?

```
#' @param method c("Magnus","ClausiusClapeyron","GoffGratch")
#' @param temp_units input temperature units ("degC","K","degF"), default "degC"
#' @param out_units output pressure units ("kPa","hPa","Pa","mb"), default "kPa"
```

## Review Time Estimate

- [x] Immediately: so it can be used in #3590
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.atmosphere/R/metutils.R
```diff
@@ -69,31 +69,104 @@ get.vpd <- function(rh, temp) {
...
+      8.1328e-3 * (10^(-3.49149 * (Tst / Tk - 1)) - 1) +
+      lg10(est)
+    es_hPa <- 10^log10_es
+    return(units::ud_convert(es_hPa, "hPa", out_units))
+  }
+
+  PEcAn.logger::logger.severe(
+    method,
+    "for converting temperature to saturated vapor pressure not supported"
+  )
```
infotroph: FTR match.arg will handle this case for you by erroring on no match, so this logger call will never fire.

infotroph (approved):
Needs Roxygen run and I see one logger call that will never fire, but otherwise looks good

modules/data.atmosphere/R/metutils.R
dlebauer: Cool, thanks!

PR #3598, opened 2025-08-15
Quick update to #3597: export and document t2es alias; add venv/ to .Rbuildignore

dlebauer:
Quick update to #3597

- export and document t2es alias
- add venv/ to .Rbuildignore


modules/data.land/.Rbuildignore
```diff
@@ -1,3 +1,4 @@
 contrib
 data-raw
 ^docs$
+.*venv.*
```
dlebauer:
```suggestion
.*venv/
```

infotroph (approved):
Not sure why there's more Rd changes showing here, but 👍 whenever CI passes

PR #3608, opened 2025-08-26
Fixed two issues in pecan-to-sipnet parameter conversion.

arob5:
## Description

Two small bug fixes in `write.configs.SIPNET.R`. Both were parameter translations involving `SLA`. 

1. The forward and inverse conversions between `leafCSpWt` and `leafC` were inconsistent. This was likely due to a unit conversion error. 
2. The forward and inverse conversions between `amax` (SIPNET) and `Amax` (PEcAn) were inconsistent. This was a simple arithmetic error.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/R/write.configs.SIPNET.R
+++ b/models/sipnet/R/write.configs.SIPNET.R
@@ -191,31 +191,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     pft.traits <- as.numeric(pft.traits)
     
     # Leaf carbon concentration
-    leafC <- 0.48  #0.5
+    leafC <- 0.48
     if ("leafC" %in% pft.names) {
       leafC <- pft.traits[which(pft.names == "leafC")]
       id <- which(param[, 1] == "cFracLeaf")
-      param[id, 2] <- leafC * 0.01  # convert to percentage from 0 to 1
+      param[id, 2] <- leafC * 0.01  # convert from percentage to fraction
     }
     
     # Specific leaf area converted to SLW
+    # leafCSpWt [gC/m2 leaf], SLA [m2 leaf/kg C], leafC [percentage C]
     SLA <- NA
     id <- which(param[, 1] == "leafCSpWt")
     if ("SLA" %in% pft.names) {
       SLA <- pft.traits[which(pft.names == "SLA")]
-      param[id, 2] <- 1000 * leafC * 0.01 / SLA
+      param[id, 2] <- 10 * leafC / SLA
     } else {
-      SLA <- 1000 * leafC / param[id, 2]
+      SLA <- 10 * leafC / param[id, 2]
     }
     
     # Maximum photosynthesis
+    # amax [nmol CO2 * g^{-1} leaf * sec^{-1}]
+    # Amax [umol CO2 * m^2 leaf * sec^{-1}]
     Amax <- NA
     id <- which(param[, 1] == "aMax")
     if ("Amax" %in% pft.names) {
       Amax <- pft.traits[which(pft.names == "Amax")]
       param[id, 2] <- Amax * SLA
     } else {
-      Amax <- param[id, 2] * SLA
+      Amax <- param[id, 2] / SLA
     }
     # Daily fraction of maximum photosynthesis
     if ("AmaxFrac" %in% pft.names) {
--- a/models/sipnet/R/write.configs.SIPNET.R
+++ b/models/sipnet/R/write.configs.SIPNET.R
@@ -221,7 +221,7 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
       param[id, 2] <- PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol", "nmol") # [nmol/g/s]
     } else {
       amax_mass <- param[id, 2] # [nmol/g/s]
-      Amax <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol", "umol") # [umol/m2/s]
+      Amax_area <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol", "umol") # [umol/m2/s]
     }
     
     # Daily fraction of maximum photosynthesis
--- a/models/sipnet/R/write.configs.SIPNET.R
+++ b/models/sipnet/R/write.configs.SIPNET.R
@@ -165,7 +165,7 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
   }
   for (pft in seq_along(trait.values)) {
     pft.traits <- unlist(trait.values[[pft]])
-    pft.names <- names(pft.traits)
+    pft.trait.names <- names(pft.traits)
     
     ## Append/replace params specified as constants
     constant.traits <- unlist(defaults[[1]]$constants)
@@ -173,10 +173,10 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     
     # Replace matches
     for (i in seq_along(constant.traits)) {
-      ind <- match(constant.names[i], pft.names)
+      ind <- match(constant.names[i], pft.trait.names)
       if (is.na(ind)) {
         # Add to list
-        pft.names <- c(pft.names, constant.names[i])
+        pft.trait.names <- c(pft.trait.names, constant.names[i])
         pft.traits <- c(pft.traits, constant.traits[i])
       } else {
         # Replace existing value
@@ -186,24 +186,26 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     
     # Remove NAs. Constants may be specified as NA to request template defaults. Note that it is 'NA'
     # (character) not actual NA due to being read in as XML
-    pft.names <- pft.names[pft.traits != "NA" & !is.na(pft.traits)]
+    pft.trait.names <- pft.trait.names[pft.traits != "NA" & !is.na(pft.traits)]
     pft.traits <- pft.traits[pft.traits != "NA" & !is.na(pft.traits)]
     pft.traits <- as.numeric(pft.traits)
     
-    # Leaf carbon concentration
-    leafC <- 0.48
-    if ("leafC" %in% pft.names) {
-      leafC <- pft.traits[which(pft.names == "leafC")]
+    # Leaf carbon concentration    
+    leafC <- NA
+    if ("leafC" %in% pft.trait.names) {
+      leafC <- pft.traits[which(pft.trait.names == "leafC")]
       id <- which(param[, 1] == "cFracLeaf")
-      param[id, 2] <- leafC * 0.01  # convert from percentage to fraction
+      param[id, 2] <- PEcAn.utils::ud_convert(leafC, "percent", "1") # Convert from percentage to fraction
+    } else {
+      leafC <- 0.48 # Fixed value if not available, because it is used in downstream calculations
     }
     
     # Specific leaf area converted to SLW
     # leafCSpWt [gC/m2 leaf], SLA [m2 leaf/kg C], leafC [percentage C]
     SLA <- NA
     id <- which(param[, 1] == "leafCSpWt")
-    if ("SLA" %in% pft.names) {
-      SLA <- pft.traits[which(pft.names == "SLA")]
+    if ("SLA" %in% pft.trait.names) {
+      SLA <- pft.traits[which(pft.trait.names == "SLA")]
       param[id, 2] <- PEcAn.utils::ud_convert(leafC / SLA, "kg/m2", "g/m2")
     } else {
       SLA <- PEcAn.utils::ud_convert(leafC / param[id, 2], "kg", "g")
@@ -213,8 +215,8 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     # PEcAn:  Amax [umol CO2 / m^2 leaf / sec]
     id <- which(param[, 1] == "aMax")
     SLA_g <- PEcAn.utils::ud_convert(SLA, "1/kg", "1/g") 
-    if ("Amax" %in% pft.names) {
-      Amax_area <- pft.traits[which(pft.names == "Amax")] # [µmol/m2/s]
+    if ("Amax" %in% pft.trait.names) {
+      Amax_area <- pft.traits[which(pft.trait.names == "Amax")] # [µmol/m2/s]
       param[id, 2] <- PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol/m2/s", "nmol/g/s")
     } else {
       amax_mass <- param[id, 2] # [nmol/g/s]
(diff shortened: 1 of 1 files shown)


arob5:
@mdietze

mdietze (approved):
All this looks good to me, but I want @DongchenZ @Qianyuxuan etc to be aware of this fix as it could substantially change parameter values in their model runs (e.g. Amax, SLA). These are core parameters driving LAI and GPP, and thus everthing else in the model

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -191,31 +191,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
     }
     
     # Specific leaf area converted to SLW
+    # leafCSpWt [gC/m2 leaf], SLA [m2 leaf/kg C], leafC [percentage C]
     SLA <- NA
     id <- which(param[, 1] == "leafCSpWt")
     if ("SLA" %in% pft.names) {
       SLA <- pft.traits[which(pft.names == "SLA")]
-      param[id, 2] <- 1000 * leafC * 0.01 / SLA
+      param[id, 2] <- 10 * leafC / SLA
```
dlebauer: I could easily have gotten confused, but based on the units specified in the comments above, I _**think**_

Lets use "C" and "biomass" to make units more explicit:

- SLA [m2 leaf / kg _biomass_] 
- leafC [fraction C] == [kg C / kg biomass] 

(_note:_ comments say leaf C is `percentage C` but if leafC = 0.48, units are fraction C, as above). 

Then the conversion would be

```
                                          leafC [kg C / kg biomass]
leafCSpWt [gC/m2 leaf] = 1000 [g / kg] x ---------------------------
                                          SLA [m2 leaf / kg biomass]
```

and 

[updated: invert first term on RHS per [comment below](https://github.com/PecanProject/pecan/pull/3608#discussion_r2308175886)] 

```
                                                  leafC [kg C / kg biomass]
SLA [m2 leaf / kg biomass] = 1000 [g C / kg C ] x ---------------------------
                                                  leafCSpWt [g C / m2 leaf]
```

If that is the case, we can do:

```suggestion
      param[id, 2] <- PEcAn.utils::ud_convert(leafC / SLA, 'kg/m2', 'g/m2')
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -191,31 +191,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
     } else {
-      SLA <- 1000 * leafC / param[id, 2]
+      SLA <- 10 * leafC / param[id, 2]
```
dlebauer: and following from above:

```suggestion
      SLA <- PEcAn.utils::ud_convert(leafC / param[id, 2], 'kg', 'g')
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -191,31 +191,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     pft.traits <- as.numeric(pft.traits)
     
     # Leaf carbon concentration
-    leafC <- 0.48  #0.5
+    leafC <- 0.48
```
dlebauer: if leafC _is_ a fraction here; is it necessary to convert `leafC * 0.01` on line 198?

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -191,31 +191,34 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-      param[id, 2] <- 1000 * leafC * 0.01 / SLA
+      param[id, 2] <- 10 * leafC / SLA
     } else {
-      SLA <- 1000 * leafC / param[id, 2]
+      SLA <- 10 * leafC / param[id, 2]
     }
     
     # Maximum photosynthesis
+    # amax [nmol CO2 * g^{-1} leaf * sec^{-1}]
+    # Amax [umol CO2 * m^2 leaf * sec^{-1}]
```
dlebauer: These are units from documentation.

```suggestion
    # SIPNET: aMax [nmol CO2 / g   leaf / sec]
    # PEcAn:  Amax [umol CO2 / m^2 leaf / sec]
```

PEcAn: https://github.com/PecanProject/pecan/blob/975dcc2fd30c0d12715df9c2e96ba6351ae2ee12/base/utils/data/trait.dictionary.csv#L34

[SIPNET docs](https://github.com/PecanProject/sipnet/blob/8b629c9ca7654b451d7335e641e131757abfab5c/docs/parameters.md?plain=1#L132)
image

In which case 

aMax (per g)

```
aMax_mass [nmol CO2 / g biomass / s] =
    Amax [umol CO2 / m2 leaf / s]  x 1000 [nmol / umol]
  x SLA [m2 leaf / kg biomass] x [1 kg / 1000 g]

```

Amax (per area):
```
Amax_area [umol CO2 / m2 leaf / s] =
    aMax [nmol CO2 / g biomass / s]    x 0.001 [umol / nmol]
  x (1 / SLA) [kg biomass / m2 leaf]   x [1 kg / 1000 g]
  #             ^  
  #             kg/m2 correspond to 1/SLA
```

And code could be 

```
SLA_g <- PEcAn.utils::ud_convert(SLA, '1/kg', '1/g') 
if ("Amax" %in% pft.names) {
  Amax_area <- pft.traits[which(pft.names == "Amax")] # [µmol/m2/s]
  param[id, 2] <- PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol/m2/s", "nmol/g/s")
} else {
  amax_mass <- param[id, 2] # [nmol/g/s]
  Amax <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol/g/s", "umol/m2/s")
}
```

dlebauer (changes requested):
Thanks for identifying these errors! Units are tricky - especially when it requires looking up and trusting documentation.

If my calculations are correct, there are still a few order of magnitude errors. But I also wouldn't be surprised if I've overlooked something. Please check the calculations below.

I've also suggested some changes to use `PEcAn.utils::ud_convert`. This should help make the intention more clear and reduce unit conversion errors generally (hard coded conversions are widespread so I'm just taking this chance to clean up as we go).

So my two requests are:
- check if my calculations are correct and update if necessary
- use the PEcAn.units::ud_convert.

models/sipnet/R/write.configs.SIPNET.R
infotroph: I can confirm that the Sipnet PFT files I've looked at all _do_ all seem to specify leafC as a percentage, so on line 198 I think the 0.01 is needed to keep using existing PFTs.

infotroph: (For the record I'd be OK with trying to have it both ways by assuming values read from the PFT are percent if > 1 and fraction if between 0 and 1, but I respect the counterargument for enforcing one standard )

dlebauer: We should definitely have a standard; slight preference for fraction but definitely not 0.48*0.01=0.0048

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: The idea of specifying "biomass" and "C" is great! But I guess the scalar "0.001 kg/g" should be "1000g/kg" in your SLA calculation above to cancel off "kg C" and "g C" (if we could assume 1000 g C/kg C for 1000g/kg).

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: I also noticed that if there is no "leafC" in the parameter file, leafC is the fraction according to "leafC <- 0.48". However, it will cause inconsistency in the following Line 207, in which leafC is assumed to be a percentage.

models/sipnet/R/write.configs.SIPNET.R
DongchenZ: I agreed that an official unit conversion would be better than just multiplying by 10.

models/sipnet/R/write.configs.SIPNET.R
Qianyuxuan: But shouldn't the unit of SLA be "[m2 leaf / kg biomass]" according to the above codes?

models/sipnet/R/write.configs.SIPNET.R
DongchenZ: Based on the fact that all existing PFTs have a prescribed percentage instead of a proportion definition for the `leafC`, I'd recommend using percentage everywhere throughout the script.

Qianyuxuan:
> All this looks good to me, but I want @DongchenZ @Qianyuxuan etc to be aware of this fix as it could substantially change parameter values in their model runs (e.g. Amax, SLA). These are core parameters driving LAI and GPP, and thus everthing else in the model

I checked and found the related change would not influence our current results as "SLA" and "Amax" could be found in our parameter file so code would not go to the "else" part.

models/sipnet/R/write.configs.SIPNET.R
dlebauer: @Qianyuxuan good catch on the SLA calculation - I've updated it. 

Looks like I had the eqn on line 207 correct, but 209 incorrect (now corrected).

models/sipnet/R/write.configs.SIPNET.R
dlebauer: @Qianyuxuan's - you are correct that SIPNET does not have a leafC parameter, since it only tracks C (and H2O). 

@DongchenZ you are also correct that "existing PFTs have a prescribed percentage", but it isn't used. 

It would be possible replace the hard coded value of leafC with a draw from both SLA and leafC - and perhaps that was the intent. If the priors for both are already in the environment, it could be done.

models/sipnet/R/write.configs.SIPNET.R
dlebauer: @Qianyuxuan thanks! I updated it by adding a SLA_g variable to cancel out the 'g' units of amax_mass

arob5:
Apologies for the delay in addressing this. I made the changes as specified in the above comments. I am unsure if the `leafC` question was resolved, so have not yet updated its current hard-coded value.

models/sipnet/R/write.configs.SIPNET.R
dlebauer: It seems the original implementation (that appears in this PR) is correct, but I've suggested some changes for clarity in https://github.com/arob5/pecan-fork/pull/4.

I apologize because I missed the fact that leafC was first initialized as 0.48 and then this value was only used if `!"leafC" %in% pft.names`. I wouldn't have noticed if @arob5 hadn't clarified the comments.

dlebauer (dismissed):
My requested change was misguided. Apologies. While I was at it I took the chance to clarify the code just a little bit. Please merge https://github.com/arob5/pecan-fork/pull/4, and then this can be merged.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -186,78 +186,84 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+      SLA <- pft.traits[which(pft.trait.names == "SLA")]
+      param[id, 2] <- PEcAn.utils::ud_convert(leafC / SLA, "kg/m2", "g/m2")
     } else {
-      SLA <- 1000 * leafC / param[id, 2]
+      SLA <- PEcAn.utils::ud_convert(leafC / param[id, 2], "kg", "g")
     }
     
-    # Maximum photosynthesis
-    Amax <- NA
+    # SIPNET: aMax [nmol CO2 / g   leaf / sec]
```
infotroph, line 215:
```suggestion
    # Maximum photosynthesis
    # SIPNET: aMax [nmol CO2 / g   leaf / sec]
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -186,78 +186,84 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-    pft.names <- pft.names[pft.traits != "NA" & !is.na(pft.traits)]
+    pft.trait.names <- pft.trait.names[pft.traits != "NA" & !is.na(pft.traits)]
     pft.traits <- pft.traits[pft.traits != "NA" & !is.na(pft.traits)]
     pft.traits <- as.numeric(pft.traits)
     
-    # Leaf carbon concentration
-    leafC <- 0.48  #0.5
-    if ("leafC" %in% pft.names) {
-      leafC <- pft.traits[which(pft.names == "leafC")]
+    # Leaf carbon concentration
```
```suggestion
    # Leaf carbon concentration
```

infotroph (approved):

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -186,78 +186,84 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+    # SIPNET: aMax [nmol CO2 / g   leaf / sec]
+    # PEcAn:  Amax [umol CO2 / m^2 leaf / sec]
     id <- which(param[, 1] == "aMax")
-    if ("Amax" %in% pft.names) {
-      Amax <- pft.traits[which(pft.names == "Amax")]
-      param[id, 2] <- Amax * SLA
+    SLA_g <- PEcAn.utils::ud_convert(SLA, "1/kg", "1/g") 
+    if ("Amax" %in% pft.trait.names) {
+      Amax_area <- pft.traits[which(pft.trait.names == "Amax")] # [µmol/m2/s]
+      param[id, 2] <- PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol/m2/s", "nmol/g/s")
```
infotroph: Sipnet integration test fails at this line:
```
Error in PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol/m2/s", "nmol/g/s") : 
  units::ud_are_convertible(u1, u2) is not TRUE
```

infotroph: Does this work?
```suggestion
      param[id, 2] <- PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol", "nmol") # [nmol/g/s]
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -186,78 +186,84 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
     } else {
-      Amax <- param[id, 2] * SLA
+      amax_mass <- param[id, 2] # [nmol/g/s]
+      Amax <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol/g/s", "umol/m2/s")
```
```suggestion
      Amax <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol", "umol") # [umol/m2/s]
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -186,78 +186,85 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
-      Amax <- pft.traits[which(pft.names == "Amax")]
-      param[id, 2] <- Amax * SLA
+    SLA_g <- PEcAn.utils::ud_convert(SLA, "1/kg", "1/g") 
+    if ("Amax" %in% pft.trait.names) {
+      Amax_area <- pft.traits[which(pft.trait.names == "Amax")] # [µmol/m2/s]
+      param[id, 2] <- PEcAn.utils::ud_convert(Amax_area * SLA_g, "umol", "nmol") # [nmol/g/s]
     } else {
-      Amax <- param[id, 2] * SLA
+      amax_mass <- param[id, 2] # [nmol/g/s]
+      Amax <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol", "umol") # [umol/m2/s]
```
infotroph: Needed for use below in response calculation, right? 
```suggestion
      Amax_area <- PEcAn.utils::ud_convert(amax_mass / SLA_g, "nmol", "umol") # [umol/m2/s]
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -186,78 +186,85 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
+      param[which(param[, 1] == "attenuation"), 2] <- pft.traits[which(pft.trait.names == "extinction_coefficient")]
     }
     
     # Leaf respiration rate converted to baseFolRespFrac
-    if ("leaf_respiration_rate_m2" %in% pft.names) {
-      Rd <- pft.traits[which(pft.names == "leaf_respiration_rate_m2")]
+    if ("leaf_respiration_rate_m2" %in% pft.trait.names) {
+      Rd <- pft.traits[which(pft.trait.names == "leaf_respiration_rate_m2")]
       id <- which(param[, 1] == "baseFolRespFrac")
       param[id, 2] <- max(min(Rd/Amax, 1), 0)
```
infotroph: ?
```suggestion
      param[id, 2] <- max(min(Rd / Amax_area, 1), 0)
```

infotroph:
Thank you, @arob5 !

models/sipnet/R/write.configs.SIPNET.R
dlebauer: LGTM!

PR #3611, opened 2025-08-30
Move *.nc.var creation out of model2netcdf

infotroph:
## Description

* Adds a new function `PEcAn.utils::nc_write_vars()`, which creates variable lookup files for a directory full of netcdf files. It has two run modes: paired mode writes a `*.nc.var` for every `*.nc` in the directory; combined mode writes a single `nc_vars.txt` containing all the variables that appear in any nc in the directory.

* Removes the `*.nc.var` writing that was previously hardcoded into every `model2netcdf.<modelname>()` function.

* Input requested: This PR does not yet call `nc_write_vars()` in any default workflows. Do we want to add that? If so where -- the end of web/workflow.R? As an optional side-effect of some other function (get.results()?)

## Motivation and Context

All models currently write a [file].nc.var alongside every [file].nc, containing a list of the variables that netCDF contains. This is redundant in at least two ways:
*  It creates many (hundreds of thousands!) of tiny identical files because each model (curretly) writes the same set of variables every time without any variation from year to year or run to run.
* It means every model implements essentially the same code to write var files.

The main current use for these files is for the PHP code to use them for listing and plotting outputs in the web workflow. Having them created as the model runs is nice, but in practice we don't think many people are using the web interface for long-running workflows. It may be enough for web/workflow.R to run `nc_write_vars()` once after model runs are completed.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (approved):
These changes look good.

What if we add this to web/workflow.R and also add a new settings flag to control it? Set off by default but have the php turn it on?

Note also that, in addition to the web portal, sda.rewind.R and the api's runs.R both look for nc.var files. The former is one of Raiho's helper functions for cleaning up -- I don't think anyone is using it and I don't think that not finding the files will break anything. The latter would be a breaking change, so we should make sure that API-based model runs also turn on the the nc.var changes by default as well. AFAIK both the web and API based runs are currently low volume so generating these files by default in that context should be fine.

dlebauer (approved):

PR #3612, opened 2025-08-31
Sobol sensitivity analysis 

blesson-07:
Implement Sobol Sensitivity Analysis 

Overview

This PR introduces Sobol sensitivity analysis to PEcAn's uncertainty .Sobol indices are a variance-based global sensitivity analysis method that quantifies how much of the output variance can be attributed to each input parameter and their interactions. 

Key Features
* Variance-based Sensitivity Analysis: Implements Sobol' method to compute first-order (main effects) and total-order 
  (including interactions) sensitivity indices

 * Seamless Integration: Extends the existing ensemble framework to support Sobol sampling designs

 * Automated Index Calculation: New compute_sobol_indices() function processes model outputs and calculates Sobol  
   indices automatically

Technical Implementation
  - Sampling Design: Leverages the joint ensemble framework (from PR #3535) to generate appropriate Sobol sample   
    matrices (A, B, and AB designs)

 - Workflow Integration: Updates run.write.configs and runModule.run.write.configs to handle Sobol-specific configuration    
    writing

 - Output Processing: New utility function compute_sobol_indices() reads ensemble outputs and computes sensitivity 
   indices using the sensitivity package


base/workflow/R/run.write.configs.R
```diff
@@ -7,8 +7,9 @@
 #'
 #'
 #' @param settings a PEcAn settings list
+#' @param all.param.samples description
```
mdietze: Insufficient documentation. Also, parameters should be documented in the oder that they are listed in the function

base/workflow/R/run.write.configs.R
```diff
@@ -98,29 +99,11 @@ run.write.configs <- function(settings, input_design, write = TRUE,
...
-        )
-      names(ensemble.samples[[pft]]) <- names(pft_traits)
-    }
-    sa.samples <- samples$sa.samples
-    runs.samples <- samples$runs.samples
-    ## env.samples <- samples$env.samples
-  } else {
-    PEcAn.logger::logger.error(samples.file, "not found, this file is required by the run.write.configs function")
-  }
+  trait.samples     <- all.param.samples$trait.samples
```
mdietze: What is the reason for passing the contents of samples.Rdata as a list into the function rather than retaining the default behavior? This doesn't seem like something that's required to implement Sobol

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -25,23 +25,65 @@ runModule.run.write.configs <- function(settings, overwrite = TRUE, input_design
...
     }
+
       if (is.null(input_design)) {
       ensemble_size <- settings$ensemble$size
       input_design <- PEcAn.uncertainty::generate_joint_ensemble_design( settings = settings, ensemble_size = ensemble_size )
     }
     
 
 
+
```
mdietze: There's already excessive white space here, why are you adding more?

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -25,23 +25,65 @@ runModule.run.write.configs <- function(settings, overwrite = TRUE, input_design
...
     #check to see if there are posterior.files tags under pft
     posterior.files <-   settings$pfts %>%
       purrr::map_chr("posterior.files", .default = NA_character_)
-
+      
+    
+    
+    samples.file <- file.path(settings$outdir, "samples.Rdata")
```
mdietze: Per earlier comment, I'm not seeing the rational for moving the loading of samples.Rdata out of run.write.configs and into runModule.run.write.configs. This doesn't seem required in any way to get Sobol to work and was never discussed. Please revert

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
+#loading libs
```
mdietze: "sobol" is not a valid subdirectory for an R package, please git mv this to the inst folder. That said, doing so should be a temporary solution, as the goal here is to integrate the Sobol analysis as something that can be run as part of the workflow, not something that's implemented in a on-off script

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+library(assertthat)
+library(lubridate)
+library(sensitivity)
+library(PEcAn.SIPNET)
+
+#reading XML 
+
+args <- list(continue = FALSE)
+
+settings <- PEcAn.settings::read.settings("/projectnb/dietzelab/bthomas/pecan_runs/sipnet_test/pecan_updated.xml")
```
mdietze: This fill will not be accessible to other users unless it's included in the PR and accessed based on relative paths (e.g. system.file)

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+settings<-settings[1]
```
mdietze: Unclear why this step is included

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+
+if ("benchmarking" %in% names(settings)) {
```
mdietze: unused, delete

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+  library(PEcAn.benchmark)
+  settings <- papply(settings, read_settings_BRR)
+}
+
+if ("sitegroup" %in% names(settings)) {
```
mdietze: unused, delete

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+
+# Update/fix/check settings.
+# Will only run the first time it's called, unless force=TRUE
+settings <-
+  PEcAn.settings::prepare.settings(settings, force = FALSE)
+
+# Write pecan.CHECKED.xml
+PEcAn.settings::write.settings(settings, outputfile = "pecan.CHECKED.xml")
+# start from scratch if no continue is passed in
+status_file <- file.path(settings$outdir, "STATUS")
```
mdietze: unused, delete

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+input_design <- PEcAn.uncertainty::generate_joint_ensemble_design(settings=settings,ensemble_size=ensemble_size)
+input_design
+
+
+
+#convverting input indices to input
+
+
+
+# Sample inputs
```
mdietze: Sampling of inputs should have occurred earlier on the generate_joint_ensemble_design step. Please improve the documentation of the intent of this chunk of code -- I have no idea what you're trying to do here so I can't tell if it is correct or not.  Based on my understanding of both Sobol and how your joint ensemble design code works, I can't see why pulling all the file paths out into a list would be necessary, since all the info needed for the sobol is already present in input_design.

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+}
+#input file path to inputs
+
+
+
+
+# extracting parameter 
+
+
+samples.file <- file.path(settings$outdir, "samples.Rdata")
```
mdietze: Same comment as last code block -- intent here is insufficiently documented and seems superfluous.

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+  sa.samples <- samples$sa.samples
+  runs.samples <- samples$runs.samples
+  ## env.samples <- samples$env.samples
+  
+} else {
+  PEcAn.logger::logger.error(samples.file, "not found, this file is required by the run.write.configs function")
+}
+
+ensemble.samples
+all_params <- ensemble.samples$temperate.deciduous.HPDA
```
mdietze: you definitely can't hard code a specific PFT into a general purpose script

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+#X2_small <- all_params[(N_small + 1):(2 * N_small), ]  # Next N rows
+#sobol_obj <- soboljansen(model = NULL, X1 = X1_small, X2 = X2_small)
+#U <- sobol_obj$X
+#length(U)
+
+
+
+
+X1 <- all_params[1:25, ]
+X2 <- all_params[26:50, ]
```
mdietze: 1. This bit of code is based on a hard-coded sample size rather than the sample size parameters.
2. If you're going to generate both the X1 and X2 ensembles from a single call to generate_joint_ensemble_design then you're going to need to double the ensemble size when that function is called relative to what the user is asking for

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+sobol_obj <- soboljansen(model = NULL, X1 = X1, X2 = X2)
+U <- sobol_obj$X
```
mdietze: step seems superfluous

modules/uncertainty/sobol/sobol.R
mdietze: 3. Sobol design needs to be over the entire input space using the row indices. Right now you're both only sampling over the parameter space and are mistakenly breaking up the parameter vectors.

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+ensemble.samples$temperate.deciduous.HPDA <-U
```
mdietze: sobol sampling needs to modify the input_design, not ensemble.samples

modules/uncertainty/sobol/sobol.R
```diff
@@ -0,0 +1,232 @@
...
+                                 )
+
+
+
+
+#running the model 
+PEcAn.workflow::runModule_start_model_runs(settings, stop.on.error = stop_on_error)
+ 
+
+#reading output
```
mdietze: Strongly suggest encapsulating the code that postprocesses the output and generates the sobol indices into a stand-alone function.

mdietze (commented):
just realized that the comments I wrote days ago may not have posted

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -3,12 +3,155 @@
...
+    }
+  } else {
+    PEcAn.logger::logger.error(samples.file, "not found, this file is required")
+  }
+  design_matrix<- data.frame(design_list)
+  
+  
+  
+  
+ if(sobol){
```
mdietze: My suggestion by Slack, which I will repeat here, is that the if(sobol) should come **AFTER THE design_matrix IS GENERATED**.  There's no way to write code that generates a Sobol design based on the design_matrix BEFORE the design_matrix is generated.

base/workflow/R/run.write.configs.R
blesson-07: That was part of one of my earlier approaches, but it turned out to be a wrong one. I've reverted the changes back to how it was originally.

modules/uncertainty/sobol/sobol.R
blesson-07: That line was included as part of a practice run or test I was doing earlier.

modules/uncertainty/sobol/sobol.R
blesson-07: That was because of my misunderstanding of how things were supposed to be.

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -202,13 +60,15 @@ generate_joint_ensemble_design <- function(settings, ensemble_size, sobol = FALS
...
-  
-  
-  
-  
+  if(sobol){
+    half<-floor(ensemble_size / 2)
+    X1 <- design_matrix[1:half, ]    
+    X2 <- design_matrix[(half + 1):ensemble_size, ] 
+    sobol_obj <- soboljansen(model = NULL, X1 = X1, X2 = X2)
+    return(sobol_obj)
```
mdietze, line 69: If you return the full sobol_obj here, but just the design_matrix in the non-sobol version, then the downstream code is definitely going to break. I suggest that you change the `return(design_matrix)` to `return(list(X=design_matrix))` and adapt the downstream code as needed. That way, in both modes the design itself is an object within the list returned by the joint ensemble design function.

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -192,18 +193,18 @@ get.parameter.samples <- function(settings,
   }
   if ("ensemble" %in% names(settings)) {
     #if it's not there it's one probably
-    if (is.null(settings$ensemble$size)) settings$ensemble$size <- 1
-    if (settings$ensemble$size == 1) {
+    if (is.null(ensemble.size)) ensemble.size <- 1
```
mdietze: I'm not seeing how ensemble.size would ever be NULL unless you set that as the default. This line of code suggests that the function definition line should set ensemble.size = 1 as the default.

modules/uncertainty/inst/sobol/sobol_analysis.R
```diff
@@ -0,0 +1,63 @@
...
+    posterior.files = posterior.files,
+    overwrite = TRUE ,
+    input_design = input_design
+  )
+  
+  #running the model 
+  PEcAn.workflow::runModule_start_model_runs(settings, stop.on.error = stop_on_error)
+  
+  
+  #reading output
```
mdietze: I think you misunderstood my previous suggestion to make the output processing a function. That would be everything from this line down, but not the code above that generates the input design and writes the configs, and runs that model -- that you should leave in a script.  Also, once in a function, the post-processing code should move to the /R folder instead of the /inst folder

modules/uncertainty/R/get.parameter.samples.R
blesson-07: I’ll remove the line inside the function and instead set ensemble.size = 1 as the default value right in the function definition.

modules/uncertainty/inst/sobol/sobol_analysis.R
blesson-07: Sure, now I get it! I’ll create a new function called processModelOutput() to handle the output processing.

modules/uncertainty/R/processModelOutput.R
```diff
@@ -0,0 +1,30 @@
+#' process model output and return the results as a list
+#'
+#' @return an list of model outputs
+#' @export
+
+
+processModelOutput <- function(){
```
mdietze: 1. This function name is way too generic and not particularly precise. Remember that PEcAn is a much larger system than just your Sobol analysis and we process model output in tons of different places. I'd recommend something more specific that makes it clear that what you're supposed to be writing here is a function that specifically loads the outputs generated as part of a Sobol uncertainty analysis and generates the Sobol indices associated with those analyses

2. This function has no arguments? R is going to throw warnings about that because you pretty immediately start accessing undefined global variables that weren't passed into the function. At a minimum it seems like this function needs the output directory and the sobol_obj

modules/uncertainty/R/processModelOutput.R
```diff
@@ -0,0 +1,30 @@
...
+ for (i in run_ids) { 
+   run_specific_outdir <- file.path(settings$outdir, i)  
+  
+   # Read output 
+   model_out <- read.output(runid = i, 
+                           outdir = run_specific_outdir)
+  
+   all_model_out[[i]] <- model_out
+ }
+  return(all_model_out)
```
mdietze: Function is returning too soon. Where is the calculation of sobol indices occurring?

modules/uncertainty/inst/sobol/sobol_analysis.R
```diff
@@ -1,63 +1,52 @@
...
+                                    posterior.files = posterior.files,
+                                    overwrite = TRUE ,
+                                    input_design = input_design)
   
   #running the model 
   PEcAn.workflow::runModule_start_model_runs(settings, stop.on.error = stop_on_error)
   
+ 
+
+ sobol_obj <- PEcAn.uncertainty::compute_sobol_indices(outdir = settings$outdir,
```
mdietze: Not sure why you're overwriting the previous sobol_obj with new content that's completely different. I'd recommend using a different variable name

modules/uncertainty/inst/sobol/sobol_analysis.R
```diff
@@ -1,63 +1,52 @@
...
 
 #' Run Sobol Sensitivity Analysis
 #' @param settings a PEcAn Settings or MultiSettings object
 #' @export
+#' @return sobol analysis result
+#' 
+
+
+
 sobol_analysis <- function( settings ) {
```
mdietze: I don't think this script should be a function. That just makes it harder to analyze multiple outputs posthoc

modules/uncertainty/inst/sobol/sobol_analysis.R
```diff
@@ -1,63 +1,52 @@
...
+  }else{
+        sobol_obj <- PEcAn.uncertainty::generate_joint_ensemble_design(settings = settings, ensemble_size = ensemble_size, sobol = TRUE) }
+  
+  
+  input_design <- sobol_obj$X
+  ensemble_size <- nrow(input_design)
+  
+  sobol_results<-list()
+  for ( site in seq_along(settings)) {
+    settings <- settings[site]
```
mdietze: This will only work the first iteration, after which setting only include the first site. Also, site access should be double [[ brackets

modules/uncertainty/inst/sobol/sobol_analysis.R
```diff
@@ -1,63 +1,52 @@
...
+  #check to see if there are posterior.files tags under pft
+  posterior.files <-   settings$pfts %>%
+    purrr::map_chr("posterior.files", .default = NA_character_)
+  PEcAn.workflow::run.write.configs(settings = settings,
```
mdietze: Rather than looping over sites and calling run.write.configs, why not just call runModule.run.write.configs and pass it the MultiSettings object?

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -57,6 +59,16 @@ generate_joint_ensemble_design <- function(settings, ensemble_size ) {
...
   design_matrix<- data.frame(design_list)
-  return(design_matrix)
-}
-
+  
+  if(sobol){
+    half<-floor(ensemble_size / 2)
+    X1 <- design_matrix[1:half, ]    
+    X2 <- design_matrix[(half + 1):ensemble_size, ] 
+    sobol_obj <- soboljansen(model = NULL, X1 = X1, X2 = X2)
```
mdietze: namepace required on all non-base functions

mdietze (approved):

PR #3615, opened 2025-09-01
Add Matt Kim's R scripts to data.remote/inst

sarahkanee:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.remote/inst/download_updated.R
```diff
@@ -0,0 +1,34 @@
+#DAAC_Set_Credential(replace = TRUE)
```
infotroph, line 1: I suggest putting this in a subdirectory named `magic`, as done for the files in meta.analysis -- even though there's only one file here it should help us keep track of where things came from. 

Even better but obviously more effort would be to add a short README to each subdirectory -- future viewers will be grateful for even a couple sentences explaining whose scripts they are, what project they were part of / how they're used, and how finished they are.

infotroph (approved):

PR #3617, opened 2025-09-04
Improve the `merge_image_tiles` function.

DongchenZ:
## Description

In this PR, I added a new argument called `ignore.conversion`, which can be set to TRUE if you already have the GeoTIFF format. I am still working on specifying the band names of the merged image.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.remote/R/merge_image_tiles.R
```diff
@@ -8,7 +8,9 @@
 #' @param out.path  character: physical path to the folder that contains converted and merged images.
 #' @param band.name character: band name of the image. Default is NULL.
 #' @param just.band.name boolean: if we just want the band names of the image file. Default is TRUE.
-#' @param keep.files Boolean: if we want to keep the image tiles at the end.
+#' @param keep.files boolean: if we want to keep the image tiles at the end.
+#' @param ignore.conversion boolean: if we want to ignore the image conversion.
```
infotroph: Tiny nit: in R the standard nomenclature for TRUE/FALSE variables is "logical" rather than "boolean".

Less of a nit, but still small: It's not obvious to me from this description what "ignore" means in this context, nor what is being converted to what. am I interpreting right that setting this TRUE leads to not performing a call to `gdal_conversion`, and if so would "skip conversion" be a clearer way to phrase it?

infotroph (approved):

modules/data.remote/R/merge_image_tiles.R
DongchenZ: Fixed. I also fixed the way to check the shell module as you suggested in another PR.

PR #3619, opened 2025-09-04
Build the Docker image stack for fork PRs without pushing

### LINKED ISSUE #3618: Fix CI workflow for fork branches

allgandalf:
## Description

Fork PRs get a read-only `GITHUB_TOKEN` and cannot push to `ghcr.io/pecanproject`, so the Docker GHA workflow failed at the push step on every external PR (`denied: installation not allowed to Write organization package`). This does two things:

- Skips the push-based `docker.yml` for pull requests opened from forks. The gate is on the `rversion` job that every build job depends on, so the whole stack is skipped for fork PRs. Non-fork events (push, merge_group, schedule, workflow_dispatch) and same-repo PRs are unchanged and still push to ghcr.
- Adds a new build-only workflow (`docker-build-pr.yml`) that runs on fork PRs and builds the core image chain (depends, base, models, sipnet) in a single job. Plain `docker build` keeps each image in the runner's local store so the next `FROM pecan/<parent>:latest` resolves locally. No registry, no push, no secrets.

This replaces the earlier artifact-based attempt on this branch. The single-job build is the sequential approach @AritraDey-Dev suggested, and it sidesteps the security concern @infotroph raised: the fork build never has access to secrets and never pushes anywhere, so it is the same trust level as running the R tests on a fork PR.

Scope is the core linear chain only. On this PR it built in about 13 minutes and used roughly 4 GB of runner disk, so there is plenty of headroom. The other model binaries, baseplus (docs, executor, api) and extras images are left out for now and can be added later behind a free-disk-space step if we want the full matrix on forks.

## Motivation and Context

Fixes #3618. Every external contributor's PR showed red Docker checks because the build tried to push to the org registry with a fork's read-only token. This gives fork PRs a real "does the stack still build" signal without needing registry write access.

## Review Time Estimate
- [x] Within one week

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue) #3618

## Checklist:


.github/workflows/docker-build-image.yml
```diff
@@ -123,6 +123,8 @@ jobs:
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
+        # Check if we are pushing to GHCR (i.e., not a PR from a fork)
+        if: ${{ github.repository == 'pecanproject/pecan' && (github.event_name == 'push' || github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule') }}
```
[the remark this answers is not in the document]
robkooper: best if we do a case insensitive compare.

robkooper (commented):
the other thing we need to fix is that it uses the right base image later, see error at: https://github.com/PecanProject/pecan/actions/runs/17475953611/job/49637278181?pr=3619#step:12:254

.github/workflows/docker-build-image.yml
```diff
@@ -123,6 +123,8 @@ jobs:
```
allgandalf, line 123: Putting comments from slack here for future work since slack messages aren't visible after sometime:

<details>

Rohan Sasne
  [Sep 5th at 2:12 AM](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757018544901639?thread_ts=1757018467.288169&cid=C8RD9BRLK)
my understanding here was that we do not need to push docker builds from PR's which are raised from fork repositories to GHCR, but please correct me if i am wrong!
[2:13](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757018587416769?thread_ts=1757018467.288169&cid=C8RD9BRLK)
we only need to build the docker file


infotroph
  [Sep 5th at 2:20 AM](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757019053191979?thread_ts=1757018467.288169&cid=C8RD9BRLK)
The part that gets thorny is that the build is a whole stack of images that build on each other, so if we’re not pushing one we have to decide whether that means not building the ones that depend on it, building them on top of the last-pushed version and hoping it’s close enough, etc.
I believe there are some complications around Docker build cache as well, but I don’t have a firm enough grasp on how that works to know what changes are needed.
All solveable, but I’ll defer to [@kooper](https://pecanproject.slack.com/team/U4QGS0LA1) on how we should approach it.
:+1:
1



Rohan Sasne
  [Sep 5th at 2:28 AM](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757019497788869?thread_ts=1757018467.288169&cid=C8RD9BRLK)
so if we’re not pushing one
Just to be clear are you concerned that we will not push the changes to GHCR when we merge the pull request with my current changes?
[2:28](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757019536274479?thread_ts=1757018467.288169&cid=C8RD9BRLK)
or do you mean that during building on the new commits in the current pull request which is yet to get merged? (edited) 


infotroph
  [Sep 5th at 2:35 AM](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757019950952369?thread_ts=1757018467.288169&cid=C8RD9BRLK)
The second — we want to check that all images build successfully from this particular commit, and the way everything is setup right now that requires pushing the freshly built images as we go so that we can pull the ones that are dependencies back when building each downstream one on top of it.
[2:38](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757020093494329?thread_ts=1757018467.288169&cid=C8RD9BRLK)
Last time we discussed this I think [@kooper](https://pecanproject.slack.com/team/U4QGS0LA1) mentioned we might be able to work around that by configuring a local-to-the-runner Docker registry, but then immediately added something to the effect “but then we’d run into all kinds of disk space issues”. I’ll let Rob tell us whether that was a possiblity worth pursuing or one he was immediately talking himself out of. (edited) 
Also sent to the channel


kooper
  [Sep 11th at 7:31 AM](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757556117654409?thread_ts=1757018467.288169&cid=C8RD9BRLK)
Yes if you look at the workflow for the docker images they are building on top of each other:
depends -> base -> pecan ->models -> sipnet
[7:33](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757556235300909?thread_ts=1757018467.288169&cid=C8RD9BRLK)
so if you want to test the  build we need each of the intermediate images, and to get those in the next step of the workflow we have to store them somewhere. As far as I have seen there are 2 options, you save them and write them to storage, every github has some small storage, or we leverage a docker registry. We can use a local registry, but the more I looked into that the harder it will be since we have different jobs, and I don't think we can keep the registr alive between jobs, esp since we use a matrix, where the jobs will run on different servers.
[7:35](https://pecanproject.slack.com/archives/C8RD9BRLK/p1757556343302479?thread_ts=1757018467.288169&cid=C8RD9BRLK)
or, what we started to do, we push the images to a registry, like dockerhub or github. What we need to do is when we specify the image we want to use, we don't just specify the name, but also include the project and repository, so image arg should be more like [ghcr.io/pecanproject/depends:pr-966](http://ghcr.io/pecanproject/depends:pr-966) this way we have the exact image that we want to use.

</details>

AritraDey-Dev (commented):
For this PR, if I'm not wrong, it looks like the PR description is mismatched with the code changes done in the PR. It does build the image and upload the changes as artifacts, till here it's fine. But then I'm not sure why that workflow is not triggered in the CI for this PR.

Given the stack of images that we should run in CI one after another, as @infotroph also mentioned in Slack, I feel we should do it like: after building those artifacts, it should trigger another workflow that runs in the main repo context with access to secrets, downloads the artifacts, and pushes them to DockerHub. So that we should build those images one after another, probably something like first `depends` builds with the tag of the commit SHA(This can be the unique tag), and then the next one (i.e., `base` Docker build) will wait for it until depends is pushed,then pull it, and like that it will continue. Probably that will take some time in CI, but I think it's worth doing.

Another idea I have is to build the images sequentially in a single workflow, like depends → base → models → sipnet, so that every build gets the context of the previous build. This would be a more monolithic approach, but it helps to avoid the complexity of artifacts  and pushing them.

infotroph:
> after building those artifacts, it should trigger another workflow that runs in the main repo context 

I think this would still have the same basic problem that it lets someone else run code they control in a context where our secrets live. As annoying as the restriction against forked workflows using main repo secrets is for this case, it does has a solid security justification: Granting automated access to untrusted code is very hard to do safely. As soon as a workflow from an external fork can automatically make something happen in a context we think of as under our control (uses our secrets / runs on hardware billed to us / pushes to our registries), the number of ways malicious (or just incautious) actors could cause trouble for us expands greatly.

I'm now leaning more towards saying this isn't worth _either_ the hassle to push to a registry from a fork _or_ to try to build the stack without pushing. I wonder if instead the entire Docker GHA job:
* shouldn't run at all in external PRs before merge
* should run in the merge queue and we should treat that as the decisive test whether a change builds successfully
* For PRs that need more detailed testing (ones with Docker-specific changes?), encourage those with write access to push their branches to origin?

...but I might be off-base, and obviously waiting for merge queue failures means we're clicking merge on things before knowing the build will work. Does this idea make any sense?

allgandalf:
Pushed a new take that i think threads the concerns here.

Two changes:
- `docker.yml` now skips the whole push-based stack for fork PRs (gated on the `rversion` job that everything depends on). This matches @infotroph's point that the pushy workflow shouldn't run on external PRs before merge. Non-fork events and same-repo PRs are unchanged.
- A new `docker-build-pr.yml` builds the core chain (depends, base, models, sipnet) in one job for fork PRs, using plain `docker build` so each image stays in the runner's local store and the next `FROM` resolves locally. No registry, no push, no secrets.

@infotroph on the security worry: the build-only job never gets secrets and never pushes anywhere, so it is not the "fork code running where our secrets live" problem. It is the same trust level as running the R tests on a fork PR, just building images. @AritraDey-Dev this is your second idea, the sequential single-workflow build, which avoids the artifact-then-privileged-workflow hop.

You can see it on this PR: Docker GHA is skipped and the new "Docker build (fork PRs)" check went green, built in about 13 minutes using around 4 GB of runner disk. Kept scope to the core chain to keep it simple and fast. Happy to extend to the other models, baseplus and extras behind a free-disk-space step if we want the full build matrix on forks.

.github/workflows/docker-build-image.yml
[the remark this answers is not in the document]
allgandalf: Reworked this so the base-image resolution isn't a concern anymore. Instead of pulling each parent from ghcr (which needed the exact `ghcr.io/pecanproject/<parent>:tag` you flagged), the new `docker-build-pr.yml` builds the whole chain in one job and each `FROM pecan/<parent>:latest` resolves from the local image store. No registry pull, so no tag or case to get wrong. The fork check in `docker.yml` also compares two GitHub-provided values (`github.event.pull_request.head.repo.full_name` vs `github.repository`) instead of a hardcoded owner string, so the case-insensitive point is moot too. (46f7d18)

.github/workflows/docker-build-pr.yml
```diff
@@ -0,0 +1,73 @@
...
+# the runner's local image store and the next `FROM pecan/<parent>:latest`
+# resolves locally. No registry, no push, no secrets.
+#
+# Scope is the core linear chain only (depends -> base -> models -> sipnet). It
+# fits the default runner disk (~19 GB free; the resident stack is ~5-6 GB). The
+# other model binaries, baseplus (docs/executor/api) and extras images are not
+# built here to keep the job within disk and time budget.
+
+on:
+  pull_request:
```
infotroph, line 23:
```suggestion
  pull_request:
  workflow_dispatch:
```

(If we want to get fancy later it might be nice to allow choosing the R version at manual dispatch time, but let's wait on that until we need it -- maybe those are cases where we'd trigger docker.yml instead.)

.github/workflows/docker-build-pr.yml
```diff
@@ -0,0 +1,73 @@
+name: Docker build (fork PRs)
+
+# Pull requests opened from forks get a read-only GITHUB_TOKEN and cannot push
```
infotroph:
```suggestion
# A fast check for Docker build errors introduced by the current PR.
# Implemented separately from the full-stack build in docker.yml because 
# pull requests opened from forks get a read-only GITHUB_TOKEN and cannot push
```

infotroph (commented):
Thanks for sticking with this! I like the overall approach and the only reason I'm not clicking approve yet is that I want to make sure we get @robkooper signoff before merging.

I agree this fully removes the authorization concerns 👍 

The remaining piece I'm cautious about is runtime -- adding another 13-minute job to the build matrix for every PR is probably OK but could definitely add up at times when a lot of PRs are being updated (or a few are being updated very quickly). If this turns out to be a problem, one possibility would be to revisit the idea of skipping the depends build (and instead retagging the one from `pecan/depends:develop-${R_VERSION}` unless the PR touches the dependency list.

Looks like at the moment the build time is roughly half spent building depends and half building base; if others downstream of base are fast to build I support adding (in future PRs) as many of them as disk space will reliably allow.

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - Added `PEcAn.data.land::event_parquet_to_json` for generating PEcAn `event.json` files from well-formatted event parquet files, with support for ensembles of events.
 
 ### Fixed
+- Docker GHA workflow no longer fails on pull requests opened from forks. Forks get a read-only token and cannot push to `ghcr.io/pecanproject`, so the push-based build is now skipped for them; a new build-only workflow builds the core image stack (depends, base, models, sipnet) without pushing so fork PRs still get a build signal (#3618).
```
```suggestion
- Docker GHA workflow no longer fails on pull requests opened from forks (#3618).
```

infotroph (approved):

PR #3620, opened 2025-09-04
Adding functions to merge netCDF files from pecan model outputs.

DongchenZ:
## Description

@infotroph @dlebauer @cami-webb @mdietze @Qianyuxuan @arob5
The current continental-scale SDA presents challenges for file management, which could result in millions of files being generated per year. This can also pose challenges to data publication, making it extremely difficult for people to use our data. Therefore, I proposed this PR to combine model netCDF files into a single file for each time step. 

The computational costs are at a medium level: merging netCDF files for half a year (3-hour interval, ~1500 steps), 100 locations, 100 ensembles, and 23 variables will take ~ 8 min for a workbench that has 28 CPUs.

The file size, however, could be a problem if we want to merge across all sites. For example, the file size could be ~ 380 GB per year for 8,000 locations and 100 ensembles. Therefore, we might consider using a tiling system similar to those used by global remote sensing products.

Let me know if you have any questions!
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+#'
+#' @return character: file paths to the merged netCDF files.
+#' @export
+#' 
+#' @author Dongchen Zhang
+#' @importFrom purrr %>%
+#' @importFrom foreach %dopar%
+all_site_nc_merge_by_year <- function (settings.dir, nc.outdir) {
+  # check shell environments.
+  if ("try-error" %in% class(try(temp <- system("which cdo", intern = T), silent = T))) {
```
infotroph: This doesn't work on my machine --  when cdo isn't found, `which` returns a warning not an error, and therefore isn't caught by `try`. Maybe...?

```suggestion
  if (suppressWarnings(system2("which", "cdo", stdout = FALSE)) != 0 ) {
```

DongchenZ: Fixed. In addition, I also fixed the checks for the `merge_image_tiles` functions.

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+    return(NA)
+  }
+  # read settings.
+  settings <- PEcAn.settings::read.settings(settings.dir)
+  # grab model outdir.
+  model.outdir <- settings$modeloutdir
+  # grab ensemble size.
+  ens.num <- settings$ensemble$size %>% as.numeric
+  # grab number of CPUs for parallel computation.
+  cores <- as.numeric(settings$state.data.assimilation$batch.settings$general.job$cores)
```
infotroph: Two competing thoughts here:

1. If reading core count from settings, seems like it should be stored someplace that's not SDA-specific. `settings$host$ncpus`, maybe? I've experimented with local approaches to run Sipnet in parallel that would be able to benefit from this setting too.
2. But higher level, I'd encourage structuring this function in a way that can work when called manually on an existing set of outputs, which probably implies ability to take more (all?) settings as function arguments rather than needing to read them from a monolithic settings. 
I'm picturing an interface more like `merge_nc_by_year(nc_dir, start_year, end_year, out_dir, cores)`, which would then find all nc files from the specified time window that are  (recursively) present in nc_dir and consolidate them to yearly files in out_dir. Does that make sense to others?

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+  site_info <- settings %>% purrr::map(~.x[["run"]]) %>% 
+    purrr::map("site") %>% purrr::map(function(s) {
+      temp <- as.numeric(c(s$id, s$lon, s$lat))
+      names(temp) <- c("site_id", "lon", "lat")
+      temp
+    }) %>% dplyr::bind_rows() %>% as.data.frame()
+  # grab time points.
+  time.points <- lubridate::year(seq(lubridate::date(settings$state.data.assimilation$start.date), 
+                                     lubridate::date(settings$state.data.assimilation$end.date), 
+                                     paste0("1 ", settings$state.data.assimilation$forecast.time.step)))
```
infotroph: As above, I'd definitely like to be able to use this on non-SDA runs and encourage making the time window a function argument instead of reading it from any one subsection of the settings (especially since different sections can have different date ranges)

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+#' @details
+#' The function is only tested for SIPNET model runs.
+#' Please make sure you have the same netCDF formats if you want to proceed with different models.
+#' 
+#' @param model.outdir character: physical path to the model output folder.
+#' @param nc.outdir  character: physical path to the folder that contains the merged netCDF files.
+#' @param ens.num numeric: ensemble size.
+#' @param site.id numeric: identification number of the site.
+#' @param lat numeric: latitude of the site.
+#' @param lon numeric: longitude of the site.
```
infotroph: These are already written into the existing nc files (see eg [model2netcdf.SIPNET](https://github.com/PecanProject/pecan/blob/develop/models/sipnet/R/model2netcdf.SIPNET.R#L225-L228)) -- I recommend using those rather than overriding here

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+  for (i in seq_along(nc.vars)) {
+    # grab the variable name.
+    var <- nc.vars[[i]]$name
+    # skip if it's time related variable.
+    if (grepl("time", var, fixed = T)) next
+    # loop over ensembles.
+    var.mat <- matrix(NA, time.values$len, ens.num)
+    for (ens in 1:ens.num) {
+      folder.name <- paste0(prefix, sprintf("%05d", ens), "-", site.id)
+      nc <- ncdf4::nc_open(file.path(model.outdir, folder.name, paste0(time, ".nc")))
```
infotroph, line 158: Seems wise to do some minimal validation here to confirm that all files are at least the same shape and use the same units for all vars (including time!)

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+#' @param lon numeric: longitude of the site.
+#' @return character: file path to the merged netCDF file.
+#' @export
+#' 
+#' @author Dongchen Zhang
+single_site_nc_merge <- function (model.outdir, nc.outdir, ens.num, site.id, lat, lon, time) {
+  # grab basic formats from the first nc file of the site.
+  # create the folder name associated with first ensemble and first site.
+  prefix <- "ENS-"
+  folder.name <- paste0(prefix, sprintf("%05d", 1), "-", 1)
```
infotroph: Shouldn't the folder name have the site id in it?
```suggestion
  folder.name <- paste0(prefix, sprintf("%05d", 1), "-", site.id)
```

infotroph: Or: rather than construct names from ensemble numbers, consider something like `folders <- list.files(out.dir, pattern = site.id, recursive = FALSE)` and then iterate over those?

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+    # loop over sites.
+    nc.files <- 
+      foreach::foreach(s = seq_along(site_info$site_id), 
+                       .packages = c("Kendall", "purrr", "ncdf4"), 
+                       .options.snow=opts) %dopar% {
+                         single_site_nc_merge(model.outdir = model.outdir, 
+                                              nc.outdir = nc.outdir, 
+                                              ens.num = ens.num, 
+                                              # cdo collgrid only works for numeric data type.
+                                              site.id = as.numeric(site_info$site_id[s]),
```
infotroph: ⚠️ This will have confusing results (involving NAs) for sites with character ID strings, which will be increasingly many of them as we do more BETY-free runs. I think this is worth getting right from the beginning.

Possibilities I see: 
1. Work through the needed steps to record character site_ids in the nc output. Pro: easiest to use downstream, minimal information loss. Con: I gather character data in netCDFs is kind of a pain.
2. Write a site number into the file and retain the number <> ID mapping elsewhere (a sidecar file of some kind?). Pro: easier to write, allows retaining assumption all NC dimensions are numeric. Con: now we have to keep track of the mapping.
3. Same as 2 but treating lat/lon as the sole site identifiers. Pro: uses dimensions we already have, so one less variable to write. Con: the rest of PEcAn is fine with multiple sites sharing a physical location, and I already have output on disk that relies on that (e.g. running the same under different IDs for different PFTs).

infotroph (changes requested):

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+  progress <- function(n) utils::setTxtProgressBar(pb, n)
+  opts <- list(progress=progress)
+  # record nc paths.
+  nc.paths <- c()
+  for (t in seq_along(time.points)) {
+    time <- time.points[t] # grab the current time point.
+    # loop over sites.
+    nc.files <- 
+      foreach::foreach(s = seq_along(site_info$site_id), 
+                       .packages = c("Kendall", "purrr", "ncdf4"),
```
dlebauer: is the Kendall package being used?

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,167 @@
...
+                                              nc.outdir = nc.outdir, 
+                                              ens.num = ens.num, 
+                                              # cdo collgrid only works for numeric data type.
+                                              site.id = as.numeric(site_info$site_id[s]), 
+                                              lat = site_info$lat[s], 
+                                              lon = site_info$lon[s], 
+                                              time)
+                       } %>% unlist
+    # merge across sites using CDO command.
+    cmd <- "cdo -P @CORES@ collgrid @NC.OUTDIR@/*.nc @OUTFILE@"
```
dlebauer: in subsequent iterations, will this match previously merged <year>.nc files and thus ingest its own outputs?

also, I'm not 100% sure how `collgrid` works, but does this and lines 112-128 guarantee that - for a given run, across years - the sites will always be the same order along the site dimension?

base/utils/R/combine_model_netcdf.R
dlebauer: If helpful, I used the runs.txt file to build the output directories:

https://github.com/ccmmf/downscaling/blob/dd4235cd5b6dd8d59d46f128a560b2fbf240981e/scripts/030_extract_sipnet_output.R#L32-L43

```
# Read the runs.txt file to get ensemble and site and directory information
runs_file <- file.path(pecan_outdir, "run", "runs.txt")
## Expected outputs based on runs.txt
expected_ens_dirs <- tibble::tibble(
    dir = readr::read_lines(runs_file)
) |>
    tidyr::separate(
        col = "dir",
        into = c("prefix", "ens", "site_id"),
        sep = "-",
        remove = FALSE
    )
```

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed. I added a separate `ancillary.inputs` argument in the function, which will be used when `settings.dir` is empty and will throw an error message if you forget to specify any inputs.

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed.

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed.

base/utils/R/combine_model_netcdf.R
DongchenZ: I will leave it to the `TODO` list.

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed. If the `site_ids` are in `character` format, we will add them to the nc file later.

base/utils/R/combine_model_netcdf.R
DongchenZ: I prefer to create the `folder.name` from `ens` and `site_id` rather than do it inversely, because that's what we have from the initial. And I also trust the naming convention more than `list.files`.

DongchenZ:
@infotroph, if everything goes well. I would also like to know your thoughts on the naming of functions and where to put the script.

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,218 @@
...
+#' `cores` number of CPUs for the parallel computation;
+#' `start.date` start date of the model run;
+#' `end.date` end date of the model run.
+#' `time.step` time step of the model run (e.g., 1 year).
+#'
+#' @return character: file paths to the merged netCDF files.
+#' @export
+#' 
+#' @author Dongchen Zhang
+#' @importFrom purrr %>%
```
infotroph: Since other functions in this package are already importing %>% from magrittr, let's use that one 

(Sorry @DongchenZ, I know I've made you switch this the other direction in other PRs 😅. My reasoning is just that we want to minimize new imports at any given moment )
```suggestion
#' @importFrom magrittr %>%
```

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,218 @@
+#' Merge model outputted netCDF files by time steps specified by pecan settings file.
+#' @details
+#' The function is only tested for SIPNET model runs.
+#' Please make sure you have the same netCDF formats if you want to proceed with different models.
+#' We could also have more functions that deal with different dimensions (e.g., by site instead of by year).
```
infotroph:
```suggestion
#' Combine many netCDFs into one file per year
#'
#' Merges model outputted netCDF files by the time steps specified in a pecan settings file.
#'
#' The function is only tested for SIPNET model runs that were run with state data assimilation enabled.
#' Please make sure you have the same netCDF formats if you want to proceed with different models.
#' We could also have more functions that deal with different dimensions (e.g., by site instead of by year).
```

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,218 @@
...
+                                              time)
+                       } %>% unlist
+    # merge across sites using CDO command.
+    cmd <- "cdo -P @CORES@ collgrid @NC.OUTDIR@/*.nc @OUTFILE@"
+    cmd <- gsub("@CORES@", cores, cmd)
+    cmd <- gsub("@NC.OUTDIR@", nc.outdir, cmd)
+    cmd <- gsub("@OUTFILE@", file.path(nc.outdir, paste0(time, ".nc")), cmd)
+    out <- system(cmd, intern = TRUE, ignore.stdout = TRUE, ignore.stderr = TRUE)
+    # if we have site ids in character format.
+    if (all(is.character(site.ids))) {
```
infotroph, line 84: Do I understand right that what happens here is `cdo collgrid` writes a numeric site var into YYYY.nc and then we overwrite that with character if needed? That seems OK with the disclosure that I haven't actually worked through how the dimensions propagate here so I'm taking your word for that.

base/utils/R/combine_model_netcdf.R
infotroph: Those are fair points, but do make sure you actually use `site.id` instead of the current hardcoded `1`.

(I may want to revisit this later as part of an effort to move the behavior from "enforce a predetermined run design" more toward "here, merge everything in this directory unless the dimensions are incompatible", but that can come later)

base/utils/R/combine_model_netcdf.R
infotroph: Thanks!

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,218 @@
...
+
+#' Merge model outputted netCDF files across ensembles for a single site.
+#' @details
+#' The function is only tested for SIPNET model runs.
+#' Please make sure you have the same netCDF formats if you want to proceed with different models.
+#' 
+#' @param model.outdir character: physical path to the model output folder.
+#' @param nc.outdir  character: physical path to the folder that contains the merged netCDF files.
+#' @param ens.num numeric: ensemble size.
+#' @param site.id numeric: identification number of the site.
```
infotroph, line 117:
```suggestion
#' @param site.id numeric: identification number of the site.
#'   See details for use with non-numeric siteIDs
```

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,218 @@
...
+  parallel::stopCluster(cl)
+  foreach::registerDoSEQ()
+  # return nc paths.
+  return(nc.paths)
+}
+
+#' Merge model outputted netCDF files across ensembles for a single site.
+#' @details
+#' The function is only tested for SIPNET model runs.
+#' Please make sure you have the same netCDF formats if you want to proceed with different models.
```
infotroph, line 106:
```suggestion
#' Please make sure you have the same netCDF formats if you want to proceed with different models.
#'
#' This function requires `site.id` to be an integer. If your sites have non-numeric IDs, one possible workaround is to pass a dummy value and then edit the output file afterward to replace its `site_id` variable with character data. If you do this, do be aware many legacy netCDF tools have poor support for netCDFs containing character data.  
```

base/utils/R/combine_model_netcdf.R
infotroph: That works

base/utils/R/combine_model_netcdf.R
infotroph: Thank you! I think the updated design will still be awkward for my workflows, but let's not let that hold up something that works for your current needs.

CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ section for the next release.
 * Add CH4 and N2O to standard_vars in PEcAn.utils
 * New function `sat_vapor_pressure()` added for computing saturation vapor pressure from temperature using various methods.
 * Added `AmeriFlux_met_ensemble()` function with ERA5 fallback for AmeriFlux meteorological data processing and ensemble generation
-
+* Added `all_site_nc_merge_by_year()` and `single_site_nc_merge()` functions to merge netCDF files across ensembles and sites from pecan model netCDF outputs.
```
infotroph, line 16: Just maintaining whitespace between release sections
```suggestion
* Added `all_site_nc_merge_by_year()` and `single_site_nc_merge()` functions to merge netCDF files across ensembles and sites from pecan model netCDF outputs.


```

infotroph (commented):
Thanks for the quick updates! I do think I'll want to revisit this design before I start using it, but let's get this merged with the version that works for you. 

I've left a few more Roxygen suggestions to make the existing assumptions clear -- see what you think.

You asked about function names and locations -- putting them in utils seems right; for naming I weakly lean toward `nc_merge_single_site` and `nc_merge_all_sites_by_year`, so that folks can find all the netcdf-related PEcAn utils functions together in the `nc_*` section of the function index. Does that make sense to the rest of the team? 

Do we expect to use the single-site function directly, or only as a helper to the all-sites one? If the latter we could consider leaving it unexported until we have  a clear indication someone wants to use it outside the utils package.

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed.

base/utils/R/combine_model_netcdf.R
DongchenZ: Thanks for the notes.

base/utils/R/combine_model_netcdf.R
DongchenZ: The cdo will skip the `site_id` var if it's in the `character` format. So we can add the variable afterwards.

base/utils/R/combine_model_netcdf.R
DongchenZ: Thanks!

DongchenZ:
@infotroph Thanks for all the comments! We should definitely discuss the design that works best for both of us later. I will give it a try for the 8,000 SDA outputs, and after that, we can worry about the code design, format, etc.

CHANGELOG.md
DongchenZ: Fixed.

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed.

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,227 @@
...
+                                                                time.step = NULL)) {
+  # check shell environments.
+  if (suppressWarnings(system2("which", "cdo", stdout = FALSE)) != 0) {
+    PEcAn.logger::logger.info("The cdo function is not detected in shell command.")
+    return(NA)
+  }
+  # if we specify settings.dir.
+  if (!is.null(settings.dir)) {
+    # read settings.
+    settings <- PEcAn.settings::read.settings(settings.dir)
```
infotroph: Sorry I missed before that this creates a circular dependency -- PEcAn.settings already imports PEcAn.settings, so we don't want to import settings back into utils. Would it work for your usage pattern to switch the first argument from `settings.dir` (a file that we read  in)  to `settings` (a settings object that was already read in the user's session)?

DongchenZ: I am not sure. I experienced situations where the `settings` object was deprecated when trying to import `settings` into a function.

DongchenZ: So I think it's safer to read the settings inside the function rather than importing the whole list.

DongchenZ: How about this, I can remove the entire `settings.dir` arguments and just use the arguments inside the `ancillary.inputs` list.

DongchenZ: Suppose you feel comfortable with it. What arguments' design would be helpful to you? So I can tweak the function entirely.

infotroph: > I experienced situations where the settings object was deprecated when trying to import settings into a function

I think this wouldn't be an issue here because (1) the settings namespace will be loaded by the external parsing, and (2) everything inside nc_merge* is just treating `settings` as an ordinary list. But I maybe forgetting something, and I do like your proposal to switch to individual arguments.

infotroph: For a quick switchover, using the existing ancillary arguments seems totally fine. Maybe the signature can be ...?
```
nc_merge_all_sites_by_year <- function (model.outdir,
                                        nc.outdir,
                                        ens.num,
                                        site.ids,
                                        start.date,
                                        end.date,
                                        time.step = "year",
                                        cores = 1)) {
```

DongchenZ: Just one more question: do you think we should have `cores` defaulted to be `NULL` or just `1`?

DongchenZ: I will just leave it as `cores = 1` so we don't have to struggle about assigning default `cores` later in the function.

infotroph: I like 1 but am not going to hold you up over it

DongchenZ: Fixed. The current version should look much cleaner. Let me know your thoughts @infotroph !

infotroph (approved):
Thanks for being fast and flexible on this! Looking forward to seeing how it does in daily use.

dlebauer (changes requested):
Hi @DongchenZ - this is nice work, it will help not only with file counts, but also with simplifying (and perhaps speeding up?) code that handles ensemble outputs  thanks for implementing this solution.

I left a few thoughts inline. To make sure this new format is stable, could you please write some minimal tests, including: 
- outputs have
  - same dimensions as inputs
  - same length(ensemble), length(time) and length(site_id) = n_files?
  - same values

base/utils/R/combine_model_netcdf.R
```diff
@@ -0,0 +1,194 @@
...
+                                        start.date, 
+                                        end.date, 
+                                        time.step = "1 year", 
+                                        cores = 1) {
+  # check shell environments.
+  if (suppressWarnings(system2("which", "cdo", stdout = FALSE)) != 0) {
+    PEcAn.logger::logger.info("The cdo function is not detected in shell command.")
+    return(NA)
+  }
+  # load arguments from the ancillary.inputs.
```
dlebauer: where do ancillary inputs come from? do they come from the global environment? what are they?

base/utils/R/combine_model_netcdf.R
DongchenZ: Yes. I tested. It will be the same order as what the `site.ids` that you input.

base/utils/R/combine_model_netcdf.R
DongchenZ: Fixed. I forgot to remove those lines.

base/utils/R/combine_model_netcdf.R
DongchenZ: Removed.

DongchenZ:
> Hi @DongchenZ - this is nice work, it will help not only with file counts, but also with simplifying (and perhaps speeding up?) code that handles ensemble outputs thanks for implementing this solution.
> 
> I left a few thoughts inline. To make sure this new format is stable, could you please write some minimal tests, including:
> 
> * outputs have
>   
>   * same dimensions as inputs
>   * same length(ensemble), length(time) and length(site_id) = n_files?
>   * same values

Thanks for the comments! Actually, I think there is no need to add additional checks to the code because the current workflow is based on an iterative merging strategy that loops over variables, ensembles, sites, and time points. The error will be thrown automatically if there is any dimension mismatch.

base/utils/R/combine_model_netcdf.R
dlebauer: > Yes. I tested

Thanks for confirming that these were tested. 
Did you also check:
- what happens if the input site dimensions have different orders? 
- whether if re-running the function causes `@NC.OUTDIR@/*.nc` to match (and merge) outputs from previous merge?

These are the kind of cases where a small unit test would helps insure that the behavior is as expected and will be stable. I drafted a test for the second one in https://github.com/DongchenZ/pecan/pull/2

dlebauer:
Thanks @DongchenZ - it seems you've clearly thought about the edge cases, and the function looks solid.

In case I wasn't clear (since Chris was asking about runtime assertions), What I was referring to is unit tests. The goal of unit tests is to capture the types of manual checks that you are already doing now, in order to catch future regression or changes in intended functionality. I push a small PR to this PR with some test files and basic checks e.g. of name, order, and values in the input and output files. The tests don't currently pass, but they should illustrate the idea and be easy to build on. https://github.com/DongchenZ/pecan/pull/2

DongchenZ:
> Thanks @DongchenZ - it seems you've clearly thought about the edge cases, and the function looks solid.
> 
> In case I wasn't clear (since Chris was asking about runtime assertions), What I was referring to is unit tests. The goal of unit tests is to capture the types of manual checks that you are already doing now, in order to catch future regression or changes in intended functionality. I push a small PR to this PR with some test files and basic checks e.g. of name, order, and values in the input and output files. The tests don't currently pass, but they should illustrate the idea and be easy to build on. [DongchenZ#2](https://github.com/DongchenZ/pecan/pull/2)

Thanks for the effort.  I just merged the PR, and I am also running a full 8,000-site outputs merging. I hope it will be finished on time, and I will keep you posted.

DongchenZ:
@infotroph @dlebauer @mdietze 
I have finished merging the single-year NC data across 8,000 sites and 2,920 time steps. The output file is about 380GB. During the merging, the memory usage is about 200MB. And the memory issue didn't happen when reading the NC file (because the reading process is based on a pointer instead of reading everything into the memory). I also included a new function `extract_nc_sda` to efficiently extract the merged NC file by `site.id`, `var.name`, and `time window`. You can iterate over sites/times if you want to extract the NC file across multiple sites/times (parallel computation could save a lot of time). 
Regarding time consumption, the full-year merging will take ~25 minutes, which will result in about 5 to 6 hours from 2012 to 2024.

dlebauer (commented):
Thanks @DongchenZ this looks good

Given that the time is manageable but non-trivial, I'm dropping an [nco-based script]([this script](https://github.com/PecanProject/pecan/pull/3620#issuecomment-3276304136)) I wrote a while ago for the same goal of concatenating ensemble directories [a rough draft never used at scale]. But if we ever want to revisit performance (e.g. read/write speed, chunking, compression).

dlebauer (approved):
Thanks @DongchenZ this looks good

Given that the time is manageable but non-trivial, I'm dropping an [nco-based script](https://gist.github.com/dlebauer/a093e6a13aa6adb5eef36421fe5efa28) that I wrote a while ago for the same goal of concatenating ensemble directories [a rough draft never used at scale]. But if we ever want to revisit performance (e.g. read/write speed, chunking, compression).

PR #3621, opened 2025-09-09
events.json schema + event generation script for california design points

### LINKED ISSUE #143: integrate events into workflow
- [x] Determine events spec 
  - https://github.com/PecanProject/pecan/discussions/3551
  - Specification: https://gist.github.com/dlebauer/5a2e505bca6afe66ca6aa44d9dd163f7
- [ ] Ensure compatibility with inventory workflows and downscaling

Focus on minimal viable product to avoid premature optimization

dlebauer:
## MVP for events based on #3551  

**use case** the primary goal is having an events.json that we can use to start building out write.events.SIPNET and doing runs for the CCMMF project. Closes https://github.com/ccmmf/organization/issues/143

Includes two files:

- Script builds MVP JSON file from site attributes. 
- MVP JSON schema for events

It generates fixed values for each site (planting, harvest, pruning, tillage, irrigation) according to the PEcAn events schema. 

Output: `events.json` file.
- There is one on geo.bu.edu: `/projectnb2/dietzelab/ccmmf/data/events/events.json`
- attached here 
[events.json.gz](https://github.com/user-attachments/files/22264674/events.json.gz)
- and a [three site version as a gist](https://gist.github.com/dlebauer/d7da24aa2a316a2dbd2f4d72d8331634).

This script uses dummy data and fixed values in order to support development of downstream pipeline. There is much work to do.

- Currently running for design points
- Later will need to generate event summaries in a matrix that can be used as inputs to site selection and downscaling.

## Discussion ?s:

- should these go in a new package focused on agronomic data?
- These quickly get very large - with 10 sites

## TODO

- [ ] add optional fraction_area_0to1 to specification
- [ ] make crop required for planting and harvest
- [ ] populate provenance / source information [not for MVP, but should be done when these are created)
- [ ] Add examples with ground cover added to orchards 
- [ ] Find a home for JSON schema (inst/extdata? utils?)


modules/data.land/data/events_schema_v0.1.0.json
```diff
@@ -0,0 +1,73 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://pecanproject.org/schema/events-mvp-0-1-0.json",
+  "type": "object",
+  "required": ["pecan_events_version", "site_id", "events"],
+  "properties": {
+    "pecan_events_version": { "type": "string", "const": "0.1.0" },
+    "site_id": { "type": "string", "minLength": 1 },
+    "ensemble_id": { "type": ["integer", "null"], "minimum": 0 },
```
infotroph: Can we spec this as string instead? I have at least one local branch where ensemble IDs are hashes.

modules/data.land/inst/events_schema_v0.1.0.json
```diff
@@ -0,0 +1,73 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://pecanproject.org/schema/events-mvp-0-1-0.json",
+  "type": "object",
+  "required": ["pecan_events_version", "site_id", "events"],
+  "properties": {
+    "pecan_events_version": { "type": "string", "const": "0.1.0" },
+    "site_id": { "type": "string", "minLength": 1 },
+    "ensemble_id": { "type": ["integer", "null"], "minimum": 0 },
+    "geometry_uri": { "type": ["string", "null"], "format": "uri" },
+    "provenance": { "type": "object", "additionalProperties": true },
```
infotroph, line 11: How carefully considered is the decision to represent geometry as a uri rather than an object? Might there be cases where geometry wants to be embedded directly like the provenance is?

modules/data.land/inst/events_schema_v0.1.0.json
```diff
@@ -0,0 +1,73 @@
...
+          "fraction_area": { "type": "number", "minimum": 0, "maximum": 1, "default": 1.0 },
+          "source": { "type": "string" },
+
+          "leaf_c_kg_m2": { "type": "number", "minimum": 0 },
+          "wood_c_kg_m2": { "type": "number", "minimum": 0 },
+          "fine_root_c_kg_m2": { "type": "number", "minimum": 0 },
+          "coarse_root_c_kg_m2": { "type": "number", "minimum": 0 },
+          "cultivar": { "type": "string" },
+          "crop_code": { "type": "string" },
+          "crop_display": { "type": "string" },
```
infotroph, line 32: I ask this without looking at the rules for JSON schema specs, so apologies if the answer is "this is how it has to work". BUT:

One list of all the possible properties of every event type feels like it will get unmanageable really fast as we add more event types. Is there a way to enumerate the possible event types here without any of their internal details, and then elsewhere in the schema have each event type get its own separate definition of only the properties accepted by that type?

modules/data.land/inst/generate_events.R
```diff
@@ -0,0 +1,254 @@
...
+# - Input:  data/ca_field_attributes.csv (columns: site_id, year, pft, ...)
+# - Output: data/mvp_events.json following data/pecan_events_schema_v0.1.0.json
+# - Events (minimal):
+#     * planting:  annual crops -> every site-year; woody perennials -> first observed year only
+#     * harvest:   all site-years
+#   Each event includes only the schema-required fields per event_type.
+
+# --- Config ---
+data_dir <- "/projectnb2/dietzelab/ccmmf/data"
+input_csv <- file.path(data_dir, "ca_field_attributes.csv")
```
infotroph:
```suggestion
field_attr_csv <- file.path(data_dir, "ca_field_attributes.csv")
```

modules/data.land/inst/generate_events.R
```diff
@@ -0,0 +1,254 @@
...
+DESIGN_POINTS <- TRUE
+
+PRODUCTION <- FALSE # set TRUE for all sites, not needed if DESIGN_POINTS is TRUE
+if (PRODUCTION) {
+  stop("This could be very slow; consider profiling and writing to db or arrow etc")
+}
+set.seed(123)
+
+ca_field_attributes <- vroom::vroom(
+  input_csv,
```
infotroph:
```suggestion
field_attr_csv,
```

modules/data.land/inst/generate_events.R
```diff
@@ -0,0 +1,254 @@
...
+    dplyr::filter(site_id %in% design_points$site_id)
+} else if (!PRODUCTION) {
+  ca_field_attributes <- ca_field_attributes |>
+    dplyr::slice_sample(n = 1000)
+}
+
+ca_fields <- ca_field_attributes |>
+  dplyr::select(site_id, pft, crop) |>
+  dplyr::distinct() |>
+  tidyr::crossing(year = 2016:2024) |>
```
infotroph, line 54: This repeats the crop listed for 2016 (because that's what's in ca_field_attributes.csv) every year until 2024. Don't we want to be using the mapped crop for each year?

modules/data.land/inst/generate_events.R
```diff
@@ -0,0 +1,254 @@
...
+  dplyr::transmute(
+    event_type         = "tillage",
+    date               = paste0(year, till_suffix),
+    site_id            = site_id,
+    tillage_eff_0to1   = 0.10
+  )
+
+# Irrigation (both pfts): 3 per month for all months
+months <- sprintf("%02d", 1:12)
+days <- c("05", "15", "25")
```
infotroph, line 150: For annuals this seems like it should be limited to the planting-harvest interval

modules/data.land/inst/generate_events.R
```diff
@@ -0,0 +1,254 @@
...
+  dplyr::transmute(
+    event_type = "irrigation",
+    date       = paste0(year, "-", month, "-", day),
+    site_id    = site_id,
+    amount_mm  = 40,
+    method     = "soil"
+  )
+
+# Combine and order by site/date
+events_all <- dplyr::bind_rows(
```
infotroph, line 163: Seems overly complex to squish all event types together into one (pretty sparse) data frame and then pass them into a function full of type-specific logic (line 183 below)  to reconstruct to the set of fields that were already present in the individual dataframes. 

My instinct for heterogeneous collections like this would be to convert from df to list within each type (I think e.g. `tillage |> pmap(list)` would work), then concatenate the individual event objects and do the subsetting/sorting in list form. 

It might also work to put all types in one tibble but nest the type-specific columns into a list-column so that you have something like (`site_id`, `date`, `event_type`, `event_spec`)

infotroph (approved):

infotroph:
Check failure is saying to move the json schema out of `data/` -- I'd be fine with dropping it into `inst/` instead, optionally in an `events/` subdir.

infotroph:
Merging https://github.com/dlebauer/pecan/pull/9 should fix the CI failure

modules/data.land/inst/generate_events.R
infotroph: Saving for later as https://github.com/PecanProject/pecan/issues/3635

modules/data.land/inst/generate_events.R
infotroph:
```suggestion
# Irrigation (both pfts): 3 per month for all months
# TODO: Should annual crops skip irrigation during fallow season?
months <- sprintf("%02d", 1:12)
days <- c("05", "15", "25")
```

modules/data.land/inst/generate_events.R
infotroph: filed as https://github.com/PecanProject/pecan/issues/3636

modules/data.land/inst/events_schema_v0.1.0.json
infotroph: Filed as https://github.com/PecanProject/pecan/issues/3637

PR #3622, opened 2025-09-10
Bug fixes for the date calculations in the `model2netcdf.SIPNET` function

DongchenZ:
## Description

Previously, we assumed there was no change in the `year`, so that we used the constant calculations of `pecan_start_doy` object across time steps, which only holds true if 1) we have a constant time step (e.g., every 1 year or 1 month) and 2) we are not jumping between years (e.g., jumping from 2012-12-31 to 2013-01-01). Otherwise, we will have incorrect calculations of `sub_dates` and `sub_dates_cf` (for example, if we are now calculating `sub_dates` and `sub_dates_cf` for the year 2013, where the  `sub.sipnet.output[["day"]][1] = 1`, then the calculated first date will be `2013-01-02` instead of `2013-01-01`).

Therefore, this PR 1) enables the iterative calculations of the `pecan_start_doy` based on dynamically calculated `last_date` and the year of the current loop; 2) adds another `if condition (Line 161)` that detects the situation where we are jumping between years.
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -138,29 +139,34 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
-    pecan_start_doy <- PEcAn.utils::datetime2cf(start_date, paste0("days since ",lubridate::year(start_date),"-01-01"), 
-                                          tz = "UTC")
+    # The previous code was assuming the constant pecan_start_doy throughout times.
+    # However, when we go to the next year, we will still need to account for the difference in day of the year.
+    pecan_start_doy <- PEcAn.utils::datetime2cf(last_date, paste0("days since ", y, "-01-01"), tz = "UTC")
     tvals <- sub.sipnet.output[["day"]] + sub.sipnet.output[["time"]] / 24
-    if (sub.sipnet.output[["day"]][1]-pecan_start_doy==1) {
+    # the first condition is to detect if we are jumping within a year (e.g., from 2012-07-15 to 2012-07-16)
+    # the second condition is to detect if we are jumping between years (e.g., from 2012-12-31 to 2013-01-01)
+    if (sub.sipnet.output[["day"]][1]-pecan_start_doy==1 | sub.sipnet.output[["day"]][1]+pecan_start_doy==0) {
```
infotroph: The more I look at this the more I think this `if` should be removed entirely. Sipnet's timestamps are straightforward date-times and their interpretation does not change with the simulation start date: `2012 1 12.5` on any line of sipnet.out should become `2012-01-01 12:30:00` in sub_dates, end of story.

I don't know the full history of this code and am unclear what aspect of CF conventions the `force_cf` flag is trying to enforce, but my best guess is it was trying to detect whether the stamped time represents the beginning or end of the timestep, possibly assuming this could differ depending on the input dataset. Sipnet itself doesn't care about that internally (the output will be identical* however you interpret it), but the Sipnet code documents that the input times are assumed to be the start of the interval, so I lean toward doing the same here.

(*Although it is true Sipnet passes its input timestamps through with very little validation; if your input contains a line that claims to be hour -3 of DOY 0,  sipnet.out will say the same. But I don't think this function is the place to catch issues that originate in the clim files)

infotroph: @mdietze do you know any more of the history here?

DongchenZ: I agree, I was thinking about removing the entire if control but I am not sure if the removal will cause new issues.

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -138,29 +139,34 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
-                                                ## hard-coded "step" because 
-                                                ## leap years may not contain 
-                                                ## all "steps", or
-                                                ## if model run doesnt start 
-                                                ## at 00:00:00
+                                                       ## hard-coded "step" because 
+                                                       ## leap years may not contain 
+                                                       ## all "steps", or
+                                                       ## if model run doesnt start 
+                                                       ## at 00:00:00
```
infotroph: Hmm, looks like `step` is never used below -- maybe both the assignment and the comment can be removed now?

DongchenZ: Removed.

models/sipnet/R/model2netcdf.SIPNET.R
DongchenZ: @infotroph do you think we can remove the `if control`, and set the `force_cf` as `FALSE`?

infotroph: Weirdly I think `force_cf = TRUE` is what we want here (causes sipnet2datetime to convert them without any adjustment), but this is all me reading the code without actually test-running anything!

infotroph: The follow-up question is whether we even still need sipnet2datetime, but that can be a separate refactor.

DongchenZ: According to the debug process, the `force_cf = TRUE` will cause the error, where `2013-01-01` will become `2013-01-02`.

infotroph: Ah, right, because it's treating `tval`, which is DOY-plus-fractional-days, as if it was days-elapsed.   Which means we're actually making two entire round-trips from Sipnet timestamp to CF time to POSIXct and then back to CF time with a different reference date!

Here's a possible alternative that eliminates use of `sipnet2datetime` entirely and feels cleaner to me: replace current lines 148 (`# try to determine if...`) through 162 (`sub_dates_cf <- ...`) with:

```
    raw_time <- sub.sipnet.output[["time"]] # decimal hours (eg 13.75 = 1:45 PM)
    doy <- sub.sipnet.output[["day"]] # day of year, not of month
    hr <- floor(raw_time)
    minsec <- PEcAn.utils::ud_convert(raw_time - hr, "hour", "min")
    min <- floor(minsec)
    sec <- PEcAn.utils::ud_convert(minsec - min, "minute", "second")
    sub_dates <- strptime(
      paste(y, doy, hr, min, sec),
      "%Y %j %H %M %S",
      tz = "UTC"
    )
    sub_dates_cf <- PEcAn.utils::datetime2cf(
      sub_dates,
      paste0("days since ", y, "-01-01"),
      tz = "UTC"
    )
```

DongchenZ: Thanks! I just merged the suggestions and tested, and it works.

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -93,35 +58,35 @@ sipnet2datetime <- function(sipnet_tval, base_year, base_month = 1,
 ##' @author Shawn Serbin, Michael Dietze
 model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date, delete.raw = FALSE, revision, prefix = "sipnet.out",
                                 overwrite = FALSE, conflict = FALSE) {
-
+
```
infotroph: nit: Throughout the file let's avoid adding new end-of-line whitespace, please

(The indentation fixes further down are good, though 👍 )

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -138,29 +103,30 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
+      paste(y, doy, hr, min, sec),
+      "%Y %j %H %M %S",
+      tz = "UTC"
+    )
+    sub_dates_cf <- PEcAn.utils::datetime2cf(
+      sub_dates,
+      paste0("days since ", y, "-01-01"),
+      tz = "UTC"
+    )
+
```
infotroph, line 125: If I were a little faster last night I would have realized it's worth defining this as a function even if it's only going to be called in one place, because then unit-testing becomes very easy (e.g. `expect_identical(sipnet2POSIX(2013, 365, 23.5), as.POSIXct("2013-12-31 23:30:00", tz = "UTC"))`) and we could add coverage for all the edge cases you just found. I can do this in a follow-up PR if you'd like.

models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -93,35 +58,35 @@ sipnet2datetime <- function(sipnet_tval, base_year, base_month = 1,
...
-      na.rm = TRUE
-    ) # switched to day 2 in case first day is partial
- 
-
+  
+  out_day <- sum(
+    sipnet_output$year == simulation_years[1] &
+      sipnet_output$day == unique(sipnet_output$day)[1],
+    na.rm = TRUE
+  ) # switched to day 2 in case first day is partial
```
infotroph, line 86: Another one for a separate PR: The comment about "day 2" seems out of sync with the code and makes me wonder if there are more edge cases hiding here in cases of (1) incomplete first day as commented; (2) uneven output time steps.

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ section for the next release.
 * New function `sat_vapor_pressure()` added for computing saturation vapor pressure from temperature using various methods.
 * Added `AmeriFlux_met_ensemble()` function with ERA5 fallback for AmeriFlux meteorological data processing and ensemble generation
 
+### Fixed
+
+* Fixed a bug within the `model2netcdf.SIPNET` function where we assumed the constant calculations of `pecan_start_doy` across years (the calculations should vary depending on the last date from the last loop and the start date of the current loop), which will lead to incorrect calculations of the start `sub_dates` and `sub_dates_cf` if we are jumping between years (e.g., from 2012-12-31 to 2013-01-01).
+
```
infotroph, line 20: Can you also add a bullet to models/sipnet/NEWS.md noting this change and mentioning the removal of `sipnet2datetime`?

infotroph (approved):

models/sipnet/R/model2netcdf.SIPNET.R
DongchenZ: Fixed.

models/sipnet/R/model2netcdf.SIPNET.R
DongchenZ: Yeah. Thanks!

CHANGELOG.md
DongchenZ: added.

models/sipnet/R/model2netcdf.SIPNET.R
infotroph: Filed this as https://github.com/PecanProject/pecan/issues/3624 -- no further action needed in this PR

models/sipnet/R/model2netcdf.SIPNET.R
infotroph: filed as #3625

PR #3623, opened 2025-09-11
Add write.events.SIPNET function

dlebauer:
Add function that converts a generic `events.json` (see #3621 and #3551) into sipnet-specific `events.in` files.

## Description

Includes:

- `write.events.SIPNET`
- documentation w/ example
- tests

## Motivation and Context

Required to use PEcAn to configure SIPNET to simulate agronomic managements.  

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
+        ord <- order(dates)
+        lines <- vector()
+        for (idx in ord) {
+            e <- evs[[idx]]
+            d <- as.Date(e$date)
+            year <- as.integer(format(d, "%Y"))
+            day <- as.integer(strftime(d, "%j"))
```
infotroph: Either format or strftime works for both of these, but pls be consistent

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
+        ord <- order(dates)
+        lines <- vector()
+        for (idx in ord) {
+            e <- evs[[idx]]
```
infotroph: I find this sorting-on-the-fly pretty confusing and it feels like a likely source of sequencing bugs. Could it be done before or after the loop instead? 

Or if `idx` isn't used anywhere else below (I don't think it is, but only looked quickly), maybe this could be `for (e in  evs[ord])`?

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+    mm2cm <- as.numeric(PEcAn.utils::ud_convert(1, "mm", "cm")) # 0.1
+
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
+        ord <- order(dates)
+        lines <- vector()
```
infotroph: Preassigned empty vector will have to grow every time it's assigned to ==> potentially slow. We know up front there will be `length(evs)` lines, yeah?

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+#' )
+#' jsonlite::write_json(list(site), tmp, auto_unbox = TRUE)
+#' outdir <- withr::local_tempdir()
+#' files <- write.events.SIPNET(tmp, outdir)
+#' files
+#'
+#' @export
+write.events.SIPNET <- function(events_json, outdir) {
+    x <- jsonlite::fromJSON(events_json, simplifyVector = FALSE)
+    site_objs <- if (!is.null(x$site_id)) list(x) else x
```
infotroph, line 52: Intention not clear -- is this handling a case where if there's only one site the JSON is structured differently? If so, can we eliminate that edge case at JSON production time? Either way could use a comment here about what's happening.

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+    kg2g <- as.numeric(PEcAn.utils::ud_convert(1, "kg", "g")) # 1000
+    mm2cm <- as.numeric(PEcAn.utils::ud_convert(1, "mm", "cm")) # 0.1
+
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
+        ord <- order(dates)
```
infotroph, line 70: If I'm following the generate_events code in #3621 right, we impose ordering at json build time. Is it necessary to enforce it at both ends / does checking here mean we could save steps upstream by allowing events to be out-of-order until here?

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+        lines <- vector()
+        for (idx in ord) {
+            e <- evs[[idx]]
+            d <- as.Date(e$date)
+            year <- as.integer(format(d, "%Y"))
+            day <- as.integer(strftime(d, "%j"))
+            type <- e$event_type
+            if (type == "tillage") {
+                f <- if (is.null(e$tillage_eff_0to1)) 0 else e$tillage_eff_0to1
```
infotroph, line 78: Isn't this field mandatory in the events schema? Seems like validating up front (or saying validation is required but external) would be cleaner than mixing it into the writing steps.

That said if keeping these checks, they're a perfect use case for the null coalescing operator: `f <-e$tillage_eff_0to1 %||% 0`. And since Sipnet already imports from rlang, the @importFrom is ~free.

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+#'         list(event_type = "planting", date = "2022-02-19", leaf_c_kg_m2 = 0.01)
+#'     )
+#' )
+#' jsonlite::write_json(list(site), tmp, auto_unbox = TRUE)
+#' outdir <- withr::local_tempdir()
+#' files <- write.events.SIPNET(tmp, outdir)
+#' files
+#'
+#' @export
+write.events.SIPNET <- function(events_json, outdir) {
```
infotroph, line 45: I predict we'll want an `overwrite` flag sooner or later, since it seems to have become standard for ~all our format converters

models/sipnet/tests/testthat/test-write.events.SIPNET.R
```diff
@@ -0,0 +1,36 @@
...
+    got <- readLines(files[1])
+    expected <- c(
+        "2022  35  till  0.2",
+        "2022  40  till  0.1",
+        "2022  40  irrig  5 1",
+        "2022  40  fert   0 0 10",
+        "2022  50  plant  10 3 2 5",
+        "2022  250 harv   0.1"
+    )
+    expect_equal(norm(got), norm(expected))
```
infotroph, line 22: What's generating the whitespace differences that norm() is addressing here? Seems like this case should be deterministic enough to chase them down and eliminate them

models/sipnet/tests/testthat/test.met2model.R
```diff
@@ -17,7 +17,8 @@ add_gaps_to_nc <- function(src_nc, gapped_nc,
 
 test_that("Met conversion runs without error", {
   nc_path <- system.file("test-data", "CRUNCEP.2000.nc",
```
infotroph:
```suggestion
  nc_path <- system.file(
    "test-data",
    "CRUNCEP.2000.nc",
```

infotroph (changes requested):

models/sipnet/R/write.events.SIPNET.R
dlebauer:
```suggestion
            day <- as.integer(format(d, "%j"))
```

models/sipnet/tests/testthat/test-write.events.SIPNET.R
dlebauer:
```suggestion
    expect_equal(norm(got), norm(expected))
    # TODO determine What's generating the whitespace differences and eliminate use of norm()
```

models/sipnet/R/write.events.SIPNET.R
dlebauer:
```suggestion
                f <- if (is.null(e$tillage_eff_0to1)) 0 else e$tillage_eff_0to1
                # TODO: consider validating up front against schema rather than here
```

models/sipnet/R/write.events.SIPNET.R
dlebauer:
```suggestion
        lines <- vector(length = length(evs))
```

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,105 @@
...
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
+        ord <- order(dates)
+        lines <- vector(length = length(evs))
+        for (idx in ord) {
+            e <- evs[[idx]]
```
dlebauer:
```suggestion
        for (e in evs) {
```

models/sipnet/R/write.events.SIPNET.R
dlebauer: This function can't control where the events come from. Upstream functions could generate it out of order, but I am not sure the time savings is worth it.

models/sipnet/R/write.events.SIPNET.R
dlebauer:
```suggestion
write.events.SIPNET <- function(events_json, outdir) {
    # TODO add overwrite argument
```

models/sipnet/R/write.events.SIPNET.R
dlebauer:
```suggestion
    # allow a single site events.json that does not have a site_id 
    site_objs <- if (!is.null(x$site_id)) list(x) else x
```

models/sipnet/R/write.events.SIPNET.R
infotroph: @dlebauer Did you mean to delete the `year` assignment here? Looks like the variable is still used below

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,106 @@
...
+    mm2cm <- as.numeric(PEcAn.utils::ud_convert(1, "mm", "cm")) # 0.1
+
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
+        ord <- order(dates)
+        lines <- vector(length = length(evs))
```
infotroph: OK but when you assign below you're still appending to the end of the vector, so now every file contains n lines of `FALSE` followed by n lines of events.

If you're in a hurry, OK to change this back to `lines <- character()` and leave it alone until/unless profiling identifies it as a bottleneck.

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,106 @@
...
+        for (e in evs) {
```
infotroph:
```suggestion
        for (e in evs[ord]) {
```

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,106 @@
...
+            d <- as.Date(e$date)
+            day <- as.integer(format(d, "%j"))
```
infotroph, line 75:
```suggestion
            year <- as.integer(format(d, "%Y"))
            day <- as.integer(format(d, "%j"))
```

models/sipnet/R/write.events.SIPNET.R
infotroph:
```suggestion
        lines <- character()
```

infotroph (changes requested):

infotroph:
Merging https://github.com/dlebauer/pecan/pull/8 should resolve the current test failures

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,118 @@
...
+#'     events = list(
+#'         list(event_type = "tillage", date = "2022-02-04", tillage_eff_0to1 = 0.2),
+#'         list(event_type = "planting", date = "2022-02-19", leaf_c_kg_m2 = 0.01)
+#'     )
+#' )
+#' jsonlite::write_json(list(site), tmp, auto_unbox = TRUE)
+#' outdir <- withr::local_tempdir()
+#' files <- write.events.SIPNET(tmp, outdir)
+#' files
+#'
```
infotroph, line 42:
```suggestion
#'
#' @importFrom rlang %||%
```

models/sipnet/NAMESPACE
```diff
@@ -9,6 +9,7 @@ export(sample.IC.SIPNET)
 export(split_inputs.SIPNET)
 export(veg2model.SIPNET)
 export(write.config.SIPNET)
+export(write.events.SIPNET)
 export(write_restart.SIPNET)
 importFrom(dplyr,"%>%")
```
infotroph, line 14:
```suggestion
importFrom(dplyr,"%>%")
importFrom(rlang,"%||%")
```

infotroph:
```
  Error ('test-write.events.SIPNET.R:9:5'): write.events.SIPNET produces expected lines
  Error in `vapply(evs, function(e) e$date, character(1))`: values must be type 'character',
   but FUN(X[[1]]) result is type 'list'
  Backtrace:
   1. PEcAn.SIPNET::write.events.SIPNET(ev_json1, outdir)
        at test-write.events.SIPNET.R:9:5
   3. base::vapply(evs, function(e) e$date, character(1))
  ```

@dlebauer seems like the proximal issue is the entries in events_site1.json being list instead of scalar, but I don't understand the intention behind that format enough to recommend a fix.

**Review by infotroph — APPROVED** [review]

**Comment by infotroph** [comment]
@dlebauer reminder that this PR is still waiting for your input -- ping me if you want help on the test failures, but I'll need context from you to suggest the right fix.

File: models/sipnet/R/write.events.SIPNET.R [hunk 16]
```diff
@@ -0,0 +1,119 @@
...
+    # Unit conversion helpers
+    kg2g <- as.numeric(PEcAn.utils::ud_convert(1, "kg", "g")) # 1000
+    mm2cm <- as.numeric(PEcAn.utils::ud_convert(1, "mm", "cm")) # 0.1
+
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) e$date, character(1)))
```
dlebauer:
```suggestion
        dates <- as.Date(vapply(evs, function(e) ac.character(e$date), character(1)))
```

models/sipnet/R/write.events.SIPNET.R
```diff
@@ -0,0 +1,119 @@
...
+    # Unit conversion helpers
+    kg2g <- as.numeric(PEcAn.utils::ud_convert(1, "kg", "g")) # 1000
+    mm2cm <- as.numeric(PEcAn.utils::ud_convert(1, "mm", "cm")) # 0.1
+
+    # For each site, build event time series and write file
+    for (site in site_objs) {
+        sid <- site$site_id
+        evs <- site$events
+        # Order by date and build lines
+        dates <- as.Date(vapply(evs, function(e) ac.character(e$date), character(1)))
```
dlebauer:
```suggestion
        dates <- as.Date(vapply(evs, function(e) as.character(e$date), character(1)))
```

dlebauer:
@infotroph if you want to take another look at the changes that I've made (https://github.com/PecanProject/pecan/compare/f800475a3caa88ed75e5b4532590500ae175e777...dlebauer:write_events_sipnet) while trying to get the tests to pass, they include

- move example events.json test fixtures to data.land package
- create a validate_events_json function that validates an events.json against the schema
  - add jsonlite and jsonvalidator to data.land suggests [not sure if this should be Imports?]
  - add jsonvalidator to PEcAn.SIPNET suggests (not sure if this is necessary, 
  - add tests
  - use this function in write.configs.SIPNET()
  - as implemented, if jsonvalidate is not installed, it skips the test and returns validate = TRUE. This was to avoid haveing jsonvalidate in imports for a single function. But perhaps that is where it belongs.

modules/data.land/R/validate_events.R
```diff
@@ -0,0 +1,53 @@
...
+#' @export
+validate_events_json <- function(events_json, verbose = TRUE) {
+  if (!file.exists(events_json)) {
+    PEcAn.logger::logger.error(glue::glue("events_json file does not exist: {events_json}"))
+    return(FALSE)
+  }
+
+  if (!requireNamespace("jsonvalidate", quietly = TRUE)) {
+    PEcAn.logger::logger.warn("Skipping events schema validation: package 'jsonvalidate' not installed.")
+    return(TRUE)
```
infotroph: I see the appeal of allowing work to proceed if validation isn't available, but I'm not wild about returning TRUE -- seems too easy to interpret as "the file has been checked and is OK."  I'd be OK with any of returning FALSE ("The file has not been validated to be OK, in this case because we couldn't validate it"), returning NA  and letting callers decide whether to stop or continue with an unchecked file, or changing design to throw an error if jsonvalidate isn't installed (in which case it would still be fine to leave jsonvalidate in Suggests -- it's optional for anyone who uses data.land for things other than this function)

models/sipnet/DESCRIPTION
```diff
@@ -25,6 +26,7 @@ Imports:
     stats
 Suggests:
     coda,
+    jsonvalidate,
```
infotroph: You noted above this may not be needed. I agree: unless I missed a usage inside PEcAn.SIPNET, data.land takes care of the dependencies and sipnet can just call `PEcAn.data.land::validate_events_json()` without worrying about it.

infotroph:
```suggestion
```

modules/data.land/R/validate_events.R
infotroph: Of these I think the NA approach best, so here's a suggestion for it. 
```suggestion
    PEcAn.logger::logger.warn("Skipping events schema validation: package 'jsonvalidate' not installed.")
    return(NA)
```

infotroph (changes requested):

docker/depends/pecan_package_dependencies.csv
```diff
@@ -131,9 +131,13 @@
 "IDPmisc","*","modules/assim.batch","Imports",FALSE
 "itertools","*","modules/assim.sequential","Suggests",FALSE
 "jsonlite","*","base/remote","Imports",FALSE
+"jsonlite","*","models/sipnet","Imports",FALSE
 "jsonlite","*","models/stics","Imports",FALSE
 "jsonlite","*","modules/data.atmosphere","Imports",FALSE
+"jsonlite","*","modules/data.land","Suggests",FALSE
 "jsonlite","*","modules/data.remote","Suggests",FALSE
+"jsonvalidate","*","models/sipnet","Suggests",FALSE
```
infotroph: And if dropping above, drop here too
```suggestion
```

infotroph (approved):

PR #3631, opened 2025-09-25
[CI] add check for packages needing version increment

infotroph:
## Description

* New script `ensure_version_bump.sh` takes a Git tag (defaulting to most recent) and a path (defaulting to cwd) and complains about any R package under that path that both (1) has any changes since the tag; and (2) has not changed its package version.
* Added ensure_version_bump to the `CI/check` GHA workflow
* Bumped versions of the packages that needed it

## Motivation and Context

Way back when, all PEcAn packages shared a version and we incremented it each time we tagged a PEcAn release.

Then we started versioning each package individually, but with a mass version bump  both before (setting intended release version following semver rules) and after (appending '.9000' to each package to indicate an unreleased development version) tagging each release. This meant it was not possible for a package to keep the same version across two PEcAn releases, even if nothing had changed in the package. It also meant there was no clear way to release the same version of a package both inside and outside of a PEcAn system release (e.g. if we started submitting packages to CRAN).

Starting with PEcAn 1.9.0, we decided to stop mass-adding the .9000 suffix and instead expect a version bump in the first PR that touches each package after release. Then if we are not updating a package there is nothing further to do to it at release time, extraneous version increments are eliminated, and each package can be set to its next release version whenever we finish work on it; no need to wait for a release branch. 

The obvious drawback of expecting version bumps to be done during routine PRs rather than at release time is that it's easy to forget to do them -- see as evidence the 18 packages I'm updating after the fact here! Hence this PR to make the robots remember for us.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


hdpriest-ui:
This seems to work as an approach.

One potential (minor) pitfall is that because this is looking for a change in code, and-not a change in package, if the code is changed, and the package in question has a version which is not the tag we're searching for, that package will not be flagged.

for example:
```sh
hdpriest@NCSA-00086 MINGW64 ~/Projects/pecan (Add-targets-sipnet-image)
$ ./check_bump.sh "v1.8.0"
These packages have changed since v1.8.0 and need a version increment:
     ./shiny/BenchmarkReport ./shiny/Data-Ingest ./shiny/dbsync ./shiny/Elicitation ./shiny/global-sensitivity ./shiny/Pecan.depend ./shiny/ViewMet ./shiny/workflowPlot
Edit the DESCRIPTION file(s) to add '.9000' to the 'Version:' field.

# these files actually don't have a version number at all
hdpriest@NCSA-00086 MINGW64 ~/Projects/pecan (Add-targets-sipnet-image)
$ echo 'Version: v1.7.0' >> ./shiny/BenchmarkReport/DESCRIPTION

# because BenchmarkReport version is not v1.8.0, it will not be shown here
hdpriest@NCSA-00086 MINGW64 ~/Projects/pecan (Add-targets-sipnet-image)
$ ./check_bump.sh "v1.8.0"
These packages have changed since v1.8.0 and need a version increment:
     ./shiny/Data-Ingest ./shiny/dbsync ./shiny/Elicitation ./shiny/global-sensitivity ./shiny/Pecan.depend ./shiny/ViewMet ./shiny/workflowPlot
Edit the DESCRIPTION file(s) to add '.9000' to the 'Version:' field.

# benchmarkreport has changes since v1.8.0, but is escaping because its version is different from v1.8.0
# but its version is 1.7.0, so its not more recent, just different
```

i don't think this is that big of a deal; we will find these as we go along, and as long as we pay attention to the output of this check, we can make sure we dont leave version numbers behind.

hdpriest-ui (approved):

infotroph:
@hdpriest-ui Ooh, good catch on the unversioned Shiny apps. As soon as we merge develop into this branch, we'll see that warning with v1.9.0 for dbsync (since we just updated it in #3658).

Option 1 (my vote): Add versions to each Shiny app when it complains about them, update them by the same rules we use elsewhere.
Option 2: Exempt the entire Shiny directory on the grounds those aren't "real" R packages, just directories that happen to contain DESCRIPTION files

PR #3633, opened 2025-09-30
[sipnet] Initial write.configs support for  event files

infotroph:
## Description

Implements the changes needed inside write.config.SIPNET so that passing `settings$run$inputs$events$path` causes the file at that path to be copied into `[rundir]/events.in`. If paths are not specified Sipnet will run without events; if paths are specified but do not exist then it continues with a warning.

Also removed limitation on which input types can be written by `PEcAn.settings::setEnsemblePaths`. I think I'm the one who imposed that limitation, and best I recall I was just assuming there were only three input types so enumerating them all would make them more easily discoverable. Now it accepts any input name, creating a block for it in `inputs` if  not already present.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


dlebauer (commented):
Looks good. Nice testing. Excited to have events integrated into the pipeline!

dlebauer (approved):
Looks good. Nice testing. Excited to have events integrated into the pipeline!

PR #3634, opened 2025-10-03
Implement the `join_inputs_design` into the current SDA workflows.

DongchenZ:
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


mdietze (approved):

mdietze (approved):

mdietze (commented):
FYI, this provides an immediate fix, but really you should be passing in the new input_design object too, not just the new ensemble.size variable. That change is needed to get the desired result of sampling jointly once at the start of the SDA and then keeping that design as the SDA iterates.

mdietze (changes requested):
Just checked are was reminded that input_design is now the first argument to write.ensemble.configs and it's not optional, so adding that fix is actually required in this PR, not just a good next step for a future PR. 

https://github.com/PecanProject/pecan/blob/ccc5135da7a0007560821cac7640d3d59c89afde/modules/uncertainty/R/ensemble.R#L228

I'd recommend that you add a call to modules/uncertainty/R/generate_joint_ensemble_design once in your code early on, that you save the design to restart, and that you read that design during restart rather than generating a new one (i.e., that if you have an "if restart" option, then the input_design is generated in the 
"else")

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -44,19 +41,22 @@ sda.enkf.multisite <- function(settings,
                                restart = NULL, 
                                pre_enkf_params = NULL,
                                ensemble.samples = NULL,
-                               control=list(trace = TRUE,
-                                            TimeseriesPlot = FALSE,
-                                            debug = FALSE,
-                                            pause = FALSE,
-                                            Profiling = FALSE,
```
DongchenZ, line 59: Removed unnecessary flags.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -44,19 +41,22 @@ sda.enkf.multisite <- function(settings,
...
+                               control=list(TimeseriesPlot = FALSE,
                                             OutlierDetection=FALSE,
-                                            parallel_qsub = TRUE,
+                                            parallel_qsub = FALSE,
```
DongchenZ: Set the `parallel_qsub` as `FALSE` because we can't use both local and remote `qsub` for the model execution.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -44,19 +41,22 @@ sda.enkf.multisite <- function(settings,
...
                                             send_email = NULL,
                                             keepNC = TRUE,
                                             forceRun = TRUE,
                                             run_parallel = TRUE,
-                                            MCMC.args = NULL),
+                                            MCMC.args = NULL,
+                                            local.execution = TRUE),
```
DongchenZ: Add the `local.execution` when we just want to run the jobs locally.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -44,19 +41,22 @@ sda.enkf.multisite <- function(settings,
...
                                ...) {
+  # make sure we only specify one method for the model execution.
+  if (control$parallel_qsub & control$local.execution) {
+    PEcAn.logger::logger.info("Please make sure you have only one of the following control options turning on.")
+    PEcAn.logger::logger.info("parallel_qsub or local.execution.")
+    return(0)
+  }
```
DongchenZ: Make sure we only set one of the two flags as `TRUE` as mentioned earlier.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -65,15 +65,14 @@ sda.enkf.multisite <- function(settings,
   }else{
     restart_flag = FALSE
   }
+  # register parallel nodes.
```
DongchenZ, line 69: Add comment.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -84,50 +83,23 @@ sda.enkf.multisite <- function(settings,
...
-                  settings$state.data.assimilation$nthin %>% 
-                    as.numeric)
-  nburnin<- ifelse(is.null(settings$state.data.assimilation$nburnin), 
-                   1e4, 
-                   settings$state.data.assimilation$nburnin %>% 
-                     as.numeric)
-  censored.data<-ifelse(is.null(settings$state.data.assimilation$censored.data), 
-                        TRUE, 
-                        settings$state.data.assimilation$censored.data %>% 
-                          as.logical)
```
DongchenZ, line 132: Remove the traditional settings, as we have already set up `MCMC.args` in the arguments.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -84,50 +83,23 @@ sda.enkf.multisite <- function(settings,
...
   forecast.time.step <- settings$state.data.assimilation$forecast.time.step  #idea for later generalizing
   nens       <- as.numeric(settings$ensemble$size)
-  processvar <- settings$state.data.assimilation$process.variance
-  if(processvar=="TRUE"){
-    processvar <- TRUE
-  }else{
-    processvar <- FALSE
-  }
-  Localization.FUN <- settings$state.data.assimilation$Localization.FUN # localization function
-  scalef <- settings$state.data.assimilation$scalef %>% as.numeric() # scale factor for localization
```
DongchenZ, line 109: Remove unnecessary variable definitions.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -84,50 +83,23 @@ sda.enkf.multisite <- function(settings,
...
   #--------Initialization
   FORECAST    <- ANALYSIS <- ens_weights <- list()
   enkf.params <- list()
   restart.list <- NULL
   #create SDA folder to store output
   if(!dir.exists(settings$outdir)) dir.create(settings$outdir, showWarnings = FALSE)
-  
-  ##### Creating matrices that describe the bounds of the state variables
-  ##### interval is remade everytime depending on the data at time t
-  ##### state.interval stays constant and converts new.analysis to be within the correct bounds
```
DongchenZ, line 142: Simplify comments.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -136,7 +108,7 @@ sda.enkf.multisite <- function(settings,
   #Here I'm trying to make a temp config list name and put it into map to iterate
   if(multi.site.flag){
     conf.settings<-settings
-    site.ids <- conf.settings$run %>% purrr::map('site') %>% purrr::map('id') %>% base::unlist() %>% base::as.character()
+    site.ids <- conf.settings %>% purrr::map(~.x[['run']] ) %>% purrr::map('site') %>% purrr::map('id') %>% base::unlist() %>% base::as.character()
```
DongchenZ: I should revert this change.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -218,7 +184,7 @@ sda.enkf.multisite <- function(settings,
   if(!file.exists(paste0(settings$outdir, "/Extracted_met/"))){
     dir.create(paste0(settings$outdir, "/Extracted_met/"))
   }
-  
+  PEcAn.logger::logger.info("Splitting mets!")
```
DongchenZ, line 184: Add progress report messaging.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -295,8 +260,15 @@ sda.enkf.multisite <- function(settings,
           # Loading the model package - this is required bc of the furrr
           library(paste0("PEcAn.",settings$model$type), character.only = TRUE)
           # wrtting configs for each settings - this does not make a difference with the old code
+          # if we don't specify the input_design.
+          if (!exists("input_design")) {
+            input_design <- NULL
+          }
+          # wrtting configs for each settings - this does not make a difference with the old code
```
DongchenZ, line 269: Will remove this if-control and move the one that handles error messaging and returns up here (or earlier).

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -343,65 +315,31 @@ sda.enkf.multisite <- function(settings,
...
-    #replicate the same ids for the remaining sites
-    for (s in seq_along(conf.settings)) {
-      if (s!= random_site) {
-        if (is.null(inputs[[s]])) {
-          inputs[[s]] <- list() 
-        }
-        input_path <- conf.settings[[s]]$run$inputs[[tolower(input_tag)]]$path
-        inputs[[s]][[input_tag]]$ids<-inputs[[random_site]][[input_tag]]$ids
-        inputs[[s]][[input_tag]]$samples<- input_path[inputs[[random_site]][[input_tag]]$ids]
-      }
```
DongchenZ, line 447: Remove this whole section because we don't need the inputs/samples, which will be created during the `write.configs` function later.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -343,65 +315,31 @@ sda.enkf.multisite <- function(settings,
...
+      load(file.path(settings$outdir, "samples.Rdata"))
     }
+    #reformatting params
+    new.params <- sda_matchparam(settings, ensemble.samples, site.ids, nens)
+    # if it's not a restart run, we will generate the joint input design.
+    # get the joint input design.
+    input_design <- generate_joint_ensemble_design(settings = settings[[1]], 
+                                                   ensemble_samples = ensemble.samples, 
+                                                   ensemble_size = nens)[[1]]
   }
```
DongchenZ, line 336: Add the function call to create the `input_design`.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -440,100 +377,110 @@ sda.enkf.multisite <- function(settings,
...
+                                                          write.to.db = temp.settings$database$bety$write,
+                                                          restart = restart.arg,
+                                                          # samples=inputs,
+                                                          rename = TRUE
+                                                        )
+                                                        return(temp)
+                                                      } %>% stats::setNames(site.ids)
+        parallel::stopCluster(cl)
+        foreach::registerDoSEQ()
+        # update the file paths of different inputs when t = 1.
```
DongchenZ: Here, we switch from `furrr` to the `foreach`. This is because, for some reason, the `furrr` doesn't return the correct samples returned from the `write.config` function. For example, only the `met` and `parameter` will be returned if we sample all the `met`, `poolinitcond`, and `parameter` inputs.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -440,100 +377,110 @@ sda.enkf.multisite <- function(settings,
...
+        if (t == 1) {
+          inputs <- out.configs %>% purrr::map(~.x$samples)
+        }
```
DongchenZ, line 444: Here, we need to grab the correct inputs from the `write.config` function when `t=1`, and save them for later use (when `t>1`, we will be relying on them).

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -440,100 +377,110 @@ sda.enkf.multisite <- function(settings,
...
+          run.list <- c()
+          for (j in 1:nens) {
+            run.list <- c(run.list, paste0("ENS-", sprintf("%05d", j), "-", i))
+          }
+          return(run.list)}, .progress = F) %>% unlist
+        # create folder paths to each ensemble of each site.
+        runs.tmp <- file.path(rundir, ensemble.ids)
+        # start model runs.
+        PEcAn.logger::logger.info("Running models!")
+        # if we want to submit jobs through the combined job file.
```
DongchenZ: Update the way we create the `run.id` file.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -440,100 +377,110 @@ sda.enkf.multisite <- function(settings,
...
-        }else{
+        } else if (control$local.execution) {
+          # if we want to execute jobs locally.
+          job.files <- file.path(runs.tmp, "job.sh")
+          temp <- job.files %>% furrr::future_map(function(f){
+            cmd <- paste0("cd ", dirname(f), ";./job.sh")
+            system(cmd, intern = F, ignore.stdout = T, ignore.stderr = T)
+          }, .progress = F)
+        } else {
+          # if we want to submit jobs through the regular job submission function.
```
DongchenZ: Add the local execution part.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -440,100 +377,110 @@ sda.enkf.multisite <- function(settings,
...
+        reads <- build_X(out.configs = out.configs, 
+                         settings = settings, 
+                         new.params = new.params, 
+                         nens = nens, 
+                         read_restart_times = read_restart_times, 
+                         outdir = outdir, 
+                         t = t, 
+                         var.names = var.names, 
+                         my.read_restart = my.read_restart,
+                         restart_flag = restart_flag)
```
DongchenZ: Remove the error detections as they are unnecessary now.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -440,100 +377,110 @@ sda.enkf.multisite <- function(settings,
...
-          X <- X %>%
-          purrr::map_dfc(~.x) %>% 
-          as.matrix() %>%
+        # convert from forecast list to data frame.
+        X <- seq_along(X) %>% furrr::future_map(function(i){
+          temp <- do.call(cbind, X[i])
+          colnames(temp) <- paste0(var.names, ".", i)
+          return(temp)
+        }) %>% 
+          dplyr::bind_cols() %>%
```
DongchenZ: Convert the forecasts to a data frame.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -543,33 +490,27 @@ sda.enkf.multisite <- function(settings,
...
+        }
+        #initialize MCMC arguments.
+        if (is.null(control$MCMC.args)) {
+          MCMC.args <- list(niter = 1e5,
+                            nthin = 10,
+                            nchain = 1,
+                            nburnin = 5e4)
+        } else {
+          MCMC.args <- control$MCMC.args
         }
```
DongchenZ, line 532: Grab the `MCMC.args` from the arguments or set it up.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -677,29 +617,6 @@ sda.enkf.multisite <- function(settings,
...
-          print(enkf.params[[obs.t]]$Pf)
-          PEcAn.logger::logger.warn ("\n --------------Analysis mean ----------- \n")
-          print(t(enkf.params[[obs.t]]$mu.a))
-          PEcAn.logger::logger.warn ("\n --------------Analysis Cov ----------- \n")
-          print(enkf.params[[obs.t]]$Pa)
-          PEcAn.logger::logger.warn ("\n ------------------------------------------------------\n")
-        }
-        if (control$debug) browser()
-        if (control$pause) readline(prompt="Press [enter] to continue \n")
       } else {
```
DongchenZ: Remove the trace part as they no longer be informative.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -738,7 +655,12 @@ sda.enkf.multisite <- function(settings,
       if(adjustment == TRUE){
         analysis <-adj.ens(Pf, X, mu.f, mu.a, Pa)
       } else {
-        analysis <- as.data.frame(mvtnorm::rmvnorm(as.numeric(nrow(X)), mu.a, Pa, method = "svd"))
+        # if we don't have the analysis from the analysis function.
+        if (is.null(enkf.params[[obs.t]]$analysis)) {
+          analysis <- as.data.frame(mvtnorm::rmvnorm(as.numeric(nrow(X)), mu.a, Pa, method = "svd"))
+        } else {
+          analysis <- enkf.params[[obs.t]]$analysis
+        }
```
DongchenZ, line 654: The new analysis function provides samples of analysis ensembles, which are calculated across blocks, and it is faster than sampling everything together, so we don't have to redo the job.

modules/uncertainty/R/ensemble.R
```diff
@@ -226,7 +225,7 @@ get.ensemble.samples <- function( ensemble.size, pft.samples, env.samples,
 ##' @export
 ##' @author David LeBauer, Carl Davidson, Hamze Dokoohaki
 write.ensemble.configs <- function(input_design , ensemble.size, defaults, ensemble.samples, settings, model, 
-                                   clean = FALSE, write.to.db = TRUE, restart = NULL, samples = NULL, rename = FALSE) {
+                                   clean = FALSE, write.to.db = TRUE, restart = NULL, rename = FALSE) {
```
DongchenZ, line 229: Remove the `samples` argument as it is no longer needed.

modules/uncertainty/R/ensemble.R
```diff
@@ -320,22 +319,18 @@ for (input_tag in names(settings$run$inputs)) {
...
+    input_tags <- names(settings$run$inputs)
+    for (input_tag in input_tags) {
+      if (input_tag %in% colnames(input_design)) {
+        input_paths <- settings$run$inputs[[input_tag]]$path
+        input_indices <- input_design[[input_tag]]
+        
+        samples[[input_tag]] <- list(
+          samples = lapply(input_indices, function(idx) input_paths[[idx]])
+        )
+      }
```
DongchenZ, line 332: Remove the `samples` argument as it is no longer needed.

modules/uncertainty/R/ensemble.R
```diff
@@ -515,6 +510,8 @@ for (input_tag in names(settings$run$inputs)) {
     for (i in seq_len(ensemble.size)) {
       input_list <- list()
       for (input_tag in names(inputs)) {
+        # if it's the parameter list, skip.
+        if (input_tag == "parameters") next
```
DongchenZ, line 513: Make sure we are not sending parameters for the paths, as the new `inputs_design` does include the `parameter` field.

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -4,19 +4,20 @@
 #' are shared across sites to ensure consistent parameter sampling.
 #'
 #' @param settings A PEcAn settings object containing ensemble configuration
-#' @param sobol for activating sobol
 #' @param ensemble_size Integer specifying the number of ensemble members
+#' @param ensemble_samples list of ensemble parameters across PFTs. The default is NULL.
+#' @param sobol for activating sobol
```
DongchenZ, line 13: Add the `ensemble_samples` argument to ensure we can also pass the parameters directly.

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -51,29 +52,34 @@ generate_joint_ensemble_design <- function(settings,
...
+  # Sample parameters if we don't have it.
+  if (is.null(ensemble_samples)) {
+    PEcAn.uncertainty::get.parameter.samples(
+      settings,
+      ensemble.size = ensemble_size,
+      posterior.files,
+      ens.sample.method
+    )
+    samples.file <- file.path(settings$outdir, "samples.Rdata")
+  }
```
DongchenZ: Add `if-control` for the case where we don't assign the `ensemble_samples` argument.

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -51,29 +52,34 @@ generate_joint_ensemble_design <- function(settings,
...
     }
+  }
+  if (!is.null(samples$ensemble.samples) | !is.null(ensemble_samples)) {
+    # Just a placeholder: extract representative trait index per ensemble member
+    # You may want to flatten or select indices per trait
+    design_list[["param"]] <- seq_len(ensemble_size)
   } else {
-    PEcAn.logger::logger.error(samples.file, "not found, this file is required")
+    PEcAn.logger::logger.warn("ensemble.samples not found in samples.Rdata")
   }
```
DongchenZ, line 67: Add `if-control` for the case where we don't assign the `ensemble_samples` argument.

DongchenZ (commented):
Here, I have added a couple of bug fixes related to the new `joint_inputs_design` feature to the current SDA workflows. In addition, I also made a couple of modifications to the `sda_multisite` script to remove unnecessary/redundant parts, better aligning it with the current workflows. The revised SDA workflow was tested and found to work on the NEON-scale experiments.

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: rather than having both variables, would it be better to define one variable able to take on more than two options e.g. `execution = c("local", "qsub", "parallel_qsub")`

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: How are these variables now unnecessary? How do you control localization if these are deleted?

modules/assim.sequential/R/sda.enkf_MultiSite.R
mdietze: I'd revert, no need to remove useful information

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -289,14 +254,26 @@ sda.enkf.multisite <- function(settings,
       #out.configs object required to build X and restart.list object required for build X
       #TODO: there should be an easier way to do this than to rerun write.ensemble.configs
       restart.list <- vector("list", length(conf.settings))
+      # make sure we have the input_design variable before running the write configuration function.
+      if (!exists("input_design")) {
```
mdietze, line 255: I'm confused by this bit. First, if input_design isn't an argument, and this is the non-restart part of the if/else, then input_design will never exist at this point. Second, if input_design doesn't exist, then why not just generate it instead of throwing an error?

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -343,65 +320,31 @@ sda.enkf.multisite <- function(settings,
...
-        if (is.null(inputs[[s]])) {
-          inputs[[s]] <- list() 
-        }
-        input_path <- conf.settings[[s]]$run$inputs[[tolower(input_tag)]]$path
-        inputs[[s]][[input_tag]]$ids<-inputs[[random_site]][[input_tag]]$ids
-        inputs[[s]][[input_tag]]$samples<- input_path[inputs[[random_site]][[input_tag]]$ids]
-      }
+  #create params object using samples generated from TRAITS functions
+  if(restart_flag){
+    new.params <- new.params
```
mdietze, line 322: This line doesn't do anything

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -343,65 +320,31 @@ sda.enkf.multisite <- function(settings,
...
+  } else {
+    if(!file.exists(file.path(settings$outdir, "samples.Rdata"))) PEcAn.logger::logger.severe("samples.Rdata cannot be found. Make sure you generate samples by running the get.parameter.samples function before running SDA.")
```
mdietze, line 324: `get.parameter.samples` is called WITHIN `generate_joint_input_design`. This shouldn't need to be generated ahead of time. What the input design function needs is to be able to find the posterior files from the settings object.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -543,33 +493,27 @@ sda.enkf.multisite <- function(settings,
...
         if(processvar == FALSE) {
-          an.method<-EnKF
+          an.method <- EnKF
         } else if (processvar == TRUE && settings$state.data.assimilation$q.type %in% c("SINGLE", "SITE")) {
-          an.method<-GEF.MultiSite
+          an.method <- GEF.MultiSite
+        }
+        #initialize MCMC arguments.
+        if (is.null(control$MCMC.args)) {
+          MCMC.args <- list(niter = 1e5,
```
mdietze: drop this here and just make these the defaults directly in the function args

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -756,7 +681,7 @@ sda.enkf.multisite <- function(settings,
       ### save outputs                                                      ###
       ###-------------------------------------------------------------------###---- 
       Viz.output <- list(settings, obs.mean, obs.cov) #keeping obs data and settings for later visualization in Dashboard
-      
+      # save SDA outputs.
```
mdietze: input_design needs to be added to the save!

modules/uncertainty/R/generate_joint_ensemble_design.R
mdietze: Could you clarify the argument for when it makes sense to assume that we **have** ensemble_samples already? In other words, in what cases would you want/need to recycle a previous ensemble_samples rather than generate a new sample from the posteriors?

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -51,29 +52,34 @@ generate_joint_ensemble_design <- function(settings,
...
-  samples.file <- file.path(settings$outdir, "samples.Rdata")
   samples <- new.env()
-  if (file.exists(samples.file)) {
-    load(samples.file, envir = samples)
-    if (!is.null(samples$ensemble.samples)) {
-      # Just a placeholder: extract representative trait index per ensemble member
-      # You may want to flatten or select indices per trait
-      design_list[["param"]] <- seq_len(ensemble_size)
+  # if we don't have the parameters from the outside. 
+  if (is.null(ensemble_samples)) {
```
mdietze: Why have to consecutive if(is.null(ensemble_samples) blocks? why can't these be combined into one block?

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -51,29 +52,34 @@ generate_joint_ensemble_design <- function(settings,
...
+      PEcAn.logger::logger.error(samples.file, "not found, this file is required")
     }
+  }
+  if (!is.null(samples$ensemble.samples) | !is.null(ensemble_samples)) {
+    # Just a placeholder: extract representative trait index per ensemble member
+    # You may want to flatten or select indices per trait
+    design_list[["param"]] <- seq_len(ensemble_size)
   } else {
-    PEcAn.logger::logger.error(samples.file, "not found, this file is required")
+    PEcAn.logger::logger.warn("ensemble.samples not found in samples.Rdata")
```
mdietze: why drop this from error to warn?

modules/assim.sequential/R/sda.enkf_MultiSite.R
DongchenZ: Fixed.

modules/assim.sequential/R/sda.enkf_MultiSite.R
DongchenZ: 1. Those two variables are not used anywhere in the script.
2. The localization will be applied within the `analysis` function instead of being here. eg., `analysis_block` function:https://github.com/PecanProject/pecan/blob/6d7a913dd9c5f6f3f992cbe3f9e3f263cd56bb6f/modules/assim.sequential/R/Analysis_sda_block.R#L144;
and the traditional `analysis` function: https://github.com/PecanProject/pecan/blob/6d7a913dd9c5f6f3f992cbe3f9e3f263cd56bb6f/modules/assim.sequential/R/Analysis_sda_multiSite.R#L150

modules/assim.sequential/R/sda.enkf_MultiSite.R
DongchenZ: fixed.

modules/uncertainty/R/generate_joint_ensemble_design.R
DongchenZ: done

base/logger/R/logger.R
```diff
@@ -146,7 +146,7 @@ logger.message <- function(level, msg, ..., wrap = TRUE) {
       paste(c(msg, args), collapse = " "),
       "latin-ascii"
     )
-    if (nchar(long.msg) > 20 && wrap) {
+    if (wrap) {
```
DongchenZ: The hard-coded length in the if control should be removed. The `wrap` argument should fully define it.

models/stics/NAMESPACE
```diff
@@ -2,5 +2,6 @@
 
 export(met2model.STICS)
 export(model2netcdf.STICS)
+export(pecan2stics)
```
DongchenZ, line 5: This is generated by building the existing pecan files.

modules/assim.sequential/R/Analysis_sda_block.R
```diff
@@ -21,11 +21,12 @@ analysis_sda_block <- function (settings, block.list.all, X, obs.mean, obs.cov,
...
-  if (is.null(cores)) {
-    cores <- parallel::detectCores() - 1
-    # if we only have one CPU.
-    if (cores < 1) cores <- 1
+  if (length(cores) == 0 | is.null(cores)) {
+    cores <- parallel::detectCores()
   }
+  cores <- cores - 1
+  # if we only have one CPU.
+  if (cores < 1) cores <- 1
```
DongchenZ, line 29: Update the parallel registration.

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -17,59 +17,47 @@
...
-#' @param debias_require_obs_at_t_for_predict Logical; only make residual predictions
+#' `merge_nc` determine if we want to merge all netCDF files across sites and ensembles.
+#' If it's set as `TRUE`, we will then combine all netCDF files into the `merged_nc` folder within the `outdir`.
+#' `execution` decide the way we want to execute model 
+#' including `local` ,where we execute the model locally;
+#' `qsub`, where we use the traditional `start_model_runs` function for submission;
+#' `qsub_parallel`, where we first combine jobs and submit them into the SCC.
+#' @param debias List: R list containing the covariance directory and the start year.
+#' covariance directory should include GeoTIFF files named by year.
+#' start year is numeric input which decide when to start the debiasing feature.
```
DongchenZ, line 37: Removed the previous `debias` workflow and replaced it with the new workflow.

DongchenZ (commented):
The newest commits provide the new `debias` workflow (which has already been implemented in the current SDA workflows), and the old script has been moved to the inst folder. This is cleaner and easier to understand, which could also benefit any future development of the debiasing algorithm.

modules/data.remote/R/SMAP_SMP_prep.R
```diff
@@ -88,15 +88,15 @@ SMAP_SMP_prep <- function(site_info, start_date, end_date, time_points,
...
   }, .progress = T)
   for (i in seq_along(time_points)) {
     t <- time_points[i]#otherwise the t will be number instead of date.
-    SMAP_Output[, paste0(t, "_SoilMoist")] <- SMAP.list[[i]][,paste0(t, "_SoilMoist")]
+    SMAP_Output[, paste0(t, "_SoilMoistFrac")] <- SMAP.list[[i]][,paste0(t, "_SoilMoistFrac")]
     SMAP_Output[, paste0(t, "_SD")] <- SMAP.list[[i]][,paste0(t, "_SD")]
   }
   PEcAn.logger::logger.info("SMAP SMP Prep Completed!")
-  list(SMP_Output = SMAP_Output, time_points = time_points, var = "SoilMoist")
+  list(SMP_Output = SMAP_Output, time_points = time_points, var = "SoilMoistFrac")
```
DongchenZ, line 99: Correct the variable name from `SoilMoist` to ` SoilMoistFrac`.

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -85,7 +96,5 @@ generate_joint_ensemble_design <- function(settings,
     return(sobol_obj)
   }
 
-
-
   return(list(X = design_matrix))
```
infotroph, line 85: Is it important that this be a list? All the uses I've seen so far immediately subset the result to only store the X component, so would it make sense to return `design_matrix` unwrapped in the first place?

mdietze: It is important so that the sobol version works correctly. This ensures that regardless of whether the sobol or non-sobol version is called that the output is a list that includes the design as X. In the sobol version the list includes additional info beyond just X that's required by the function that does the sobol index calculations, but not required to do the runs themselves

infotroph: 👍 Can you add this to the function docs, please? I think it'd be enough to list in the @return block the possible items for the Sobol case and mention that in the non-sobol case the list only contains X

modules/assim.sequential/R/sda.enkf_MultiSite.R
```diff
@@ -17,59 +17,47 @@
 #' @param Q         Process covariance matrix given if there is no data to estimate it.
 #' @param restart   Used for iterative updating previous forecasts. Default NULL. List object includes file path to previous runs and start date for SDA.
 #' @param pre_enkf_params Used for passing pre-existing time-series of process error into the current SDA runs to ignore the impact by the differences between process errors.
-#' @param ensemble.samples Pass ensemble.samples from outside to avoid GitHub check issues.
+#' @param ensemble.samples list of ensemble parameters across PFTs. Default is NULL.
 #' @param control   List of flags controlling the behavior of the SDA.
```
infotroph, line 21: This would be easier to read as a bulleted list -- consider making this line be something like ` @param control   List of flags controlling the behavior of the SDA. See details`, then moving all the flag descriptions on lines 22-34 into their own section of the details?

modules/uncertainty/R/generate_joint_ensemble_design.R
DongchenZ: Fixed.

modules/uncertainty/R/generate_joint_ensemble_design.R
mdietze: I'm not following the logic of why we're making ensemble_samples an internal requirement for the input_design. I don't see where/how we're actually using this anywhere. The sampling of parameters seems to be based solely on knowing the ensemble_size

modules/uncertainty/R/generate_joint_ensemble_design.R
infotroph: I was picturing something that appeared in the `@return` section of the Roxygen documentation. Maybe something like...

```
#' @return a list with at least component `X` (a dataframe containing the computed design matrix).
#'   If `sobol = TRUE`, the list will be a `sensitivity::soboljansen()` result and will contain the components documented therein.
```

DongchenZ: Done

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -57,35 +56,10 @@ generate_joint_ensemble_design <- function(settings,
     sampled_inputs[[input_tag]] <- input_result$ids
     design_list[[input_tag]] <- input_result$ids
   }
-  
-  # Sample parameters if we don't have it.
-  if (is.null(ensemble_samples)) {
```
mdietze: It needs to be detected a different way, but this function is still the one responsible for running get.parameter.samples if there's not already a samples.Rdata

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -95,9 +69,5 @@ generate_joint_ensemble_design <- function(settings,
     sobol_obj <- sensitivity::soboljansen(model = NULL, X1 = X1, X2 = X2)
     return(sobol_obj)
   }
-  # This ensures that regardless of whether the sobol or non-sobol version is called
```
mdietze: This comment seems useful, not sure why it was removed

modules/uncertainty/R/generate_joint_ensemble_design.R
DongchenZ: Fixed.

modules/uncertainty/R/generate_joint_ensemble_design.R
DongchenZ: Reverted.

mdietze (approved):

PR #3570, opened 2025-07-13
Demo 2 Quarto notebook tutorial for PEcAn uncertainty analysis

AritraDey-Dev:
## Description
Added a Quarto notebook tutorial for PEcAn uncertainty analysis (uncertainty.qmd) that demonstrates ensemble and sensitivity analysis workflows. The tutorial includes:

- Complete workflow from loading settings to running ensemble/sensitivity analyses
- Parameter customization sections for ensemble size, sampling methods, and sensitivity quantiles
- Comprehensive output interpretation guide explaining variance decomposition and parameter prioritization
- Model result visualization with run ID identification
- Session information for reproducibility

## Motivation and Context

This change introduces Demo2 for the Quarto notebook focused on uncertainty analysis. It serves as a continuation and enhancement of the work initiated in [Demo1](https://github.com/PecanProject/pecan/pull/3531). The goal is to extend the functionality and provide a more comprehensive demonstration of uncertainty analysis within the PEcAn framework.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [ ] I agree that PEcAn Project may distribute my contribution under any or all of
	- the same license as the existing code,
	- and/or the BSD 3-clause license.
- [x] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #3638, opened 2025-10-04
CI:  Demo 2 notebook rendering 

AritraDey-Dev:
## Description
This PR adds CI checks for uncertainty notebook(DEMO 2 notebook) implemented in this PR #3570

Once #3570 is merged this PR can be merged.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


.github/workflows/render-quarto.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and Render Quarto PDF
+name: Render Quarto NoteBooks
```
infotroph:
```suggestion
name: Render Demo Notebooks
```

infotroph (approved):
LGTM other than a suggestion to tweak the job name.

Two things I think it's worth considering eventually, but neither are caused by this PR and don't have to be fixed in it:
* As long as we've generated a rendered artifact, should it be posted somewhere in /accessible from the rest of the rendered documentation?
* Rebuilding the Docker base image just for this run feels kinda boil-the-oceans to me, but I recognize it helps reduce dependencies between CI jobs. Is this the right balance between efficiency and self-sufficiency?

AritraDey-Dev:
@infotroph 
1) Would sharing an artifact in the workflow run help? For example, after the workflow executes successfully, it could upload an artifact.

2)Actually I’ve tried a few different approaches to solve this. Installing pre-built dependencies didn’t help because the workflow needs to test against the newly built package. I also tried running `make install` and building the dependencies manually, but that takes almost 90 minutes, which we definitely can’t rely on.

I have another idea: caching the required packages. I attempted this earlier but wasn’t successful in getting the packages to cache during an example run. If caching seems like a better approach, I can spend some more time working on it—please confirm.

infotroph:
@AritraDey-Dev Immediate question: This looks ready to merge to me. Does it need to wait for #3570 to go in?

tl;dr for the rest of this comment: I already typed longwinded answers to your questions below, so I'll click "comment" now. But we're now definitely discussing things that would go into future PRs, so let's move any further discussion to dedicated issue threads.


> Would sharing an artifact in the workflow run help?

If you're thinking of something that uses `upload_artifact`, I'm not strongly opposed but I doubt many people would brave GHA's opaque artifact URLs to find it. I'm thinking more of something that, say, pushes a rendered HTML version into the appropriate place in the documentation repo so that folks visiting https://pecanproject.github.io/pecan-documentation/develop/demo-2.html can see the most recently rendered output as part of the PEcAn book. I haven't thought through the implementation enough to say whether it should appear as the literal contents of that page or as a separate page linked from there or what.

> Actually I’ve tried a few different approaches [...] another idea: caching the required packages

I should back up to say: The fresh-build-of-the-image approach you've implemented here has at least three big advantages: it's ready now (thanks!), we already know it works, and it's very easy to understand.  As you've found already, there's no guarantee we'd get all three of those for the other approaches we could try.

That said, you're in good company trying different approaches to PEcAn build dependencies! We've tried+discarded a lot of ideas in this area over the past few years. In particular we've tried variants on package caching at least three times and _so far_ all the ways we've tried have come out too complex / fragile / slow to feel like an improvement over what we do now, but I definitely don't think we tried all the possible approaches. If you're excited to try another time (and are OK with the possibility it doesn't turn out to be the right approach here), then I'll be excited to see what you come up with.

Another direction worth considering is to ask whether it's really _essential_ to build from the PR target branch, or whether it would be OK to use, say, the most recently pushed base image. That might imply a switch from thinking of this as a task that runs on every push to one that runs on a schedule (nightly? weekly?). Here too, I haven't thought through all the possible approaches and the right design could be very different from what I'm throwing out here.

AritraDey-Dev:
> @AritraDey-Dev Immediate question: This looks ready to merge to me. Does it need to wait for https://github.com/PecanProject/pecan/pull/3570 to go in?

Yes, we need to wait until #3570 is merged because the Demo 2 notebook hasn’t been added to the repo yet, so the workflow would fail right now.

> If you're thinking of something that uses upload_artifact,[...] as a separate page linked from there or what.

That makes sense. Showing the output in HTML whenever there’s a push to the repo is reasonable, but instead of rendering everything, I think it would be better to provide a link to the artifacts, as the uncertainty analysis output is quite large.

dlebauer (approved):

AritraDey-Dev:
This is ready to merge. The Demo2 (uncertainty notebook) has been merged into the repo, and all CI checks are passing. The notebooks are rendering correctly.

PR #3640, opened 2025-10-06
Handle null status in SIPNET download script For DEMO 1 Notebook

AritraDey-Dev:
## Description

The `download_sipnet.R` script, used in the `Demo 1: Basic Run` tutorial, would crash with an argument is of length zero error if the downloaded sipnet binary failed to execute.The resulting output looks like this:
```
downloaded 146 KB

Error in `if (status == 1) ...`:
! argument is of length zero

Quitting from run_pecan.qmd:98-105 [download-sipnet]
Execution halted
```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/documentation/tutorials/Demo_1_Basic_Run/download_sipnet.R
+++ b/documentation/tutorials/Demo_1_Basic_Run/download_sipnet.R
@@ -39,23 +39,30 @@ download.file(
 Sys.chmod(dest_path, mode = "0755")
 
 ## Now we are run, lets just check that `sipnet -h` works
-
-status <- suppressWarnings(
-    system2(dest_path, "-h", stderr = TRUE, stdout = TRUE) |>
-        attr("status")
-)
-
-if (is.null(status)) {
+tryCatch(
+  {
+    # This block runs if system2 succeeds with exit code 0 (status attribute is NULL).
+    # This is unexpected for `sipnet -h`, so we warn but assume it's OK.
+    system2(dest_path, "-h", stderr = TRUE, stdout = TRUE)
+    PEcAn.logger::logger.warn("SIPNET ran with exit code 0, but expected 1. Assuming installation is OK.")
+  },
+  warning = function(w) {
+    # This block runs if system2 returns a non-zero exit code, which is expected.
+    # We check the warning message for the expected status of 1.
+    if (grepl("had status 1", w$message, fixed = TRUE)) {
+      PEcAn.logger::logger.info("SIPNET has been installed!")
+    } else {
+      PEcAn.logger::logger.error("SIPNET ran but failed with an unexpected status.", "Details:", w$message)
+    }
+  },
+  error = function(e) {
+    # This block runs if system2 fails to execute the command at all.
     PEcAn.logger::logger.error(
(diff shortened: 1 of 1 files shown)


documentation/tutorials/Demo_1_Basic_Run/download_sipnet.R
```diff
@@ -45,7 +45,12 @@ status <- suppressWarnings(
         attr("status")
 )
 
-if(status == 1){
+if (is.null(status)) {
```
infotroph: Annoyingly, when capturing output with `stdout=TRUE` it seems `system2` only sets the status attribute for one of its three possible outcomes:
1. If the command runs with no problems and exits with a status of 0 ("success"), then attr("status") is unset. No, I don't know why they designed it this way instead of setting attr("status") to 0 on success.
2. If the command runs but reports a non-zero exit status, then attr(status) is set.
3. If the command doesn't run at all, system2 stops with an R error before returning (so there's no return value to which to attach a status attribute).

If you don't care about the error message then you can use `system2(..., stderr=FALSE, stdout=FALSE)` to get numeric status as the return value, but then the warning from case 2 gets swallowed. I think your best bet will be to replace the `SuppressWarnings` with a `tryCatch` that handles both the error from case 3 and the warning from case 2.

AritraDey-Dev: I hadn’t realized that `system2() `only sets the status attribute for non-zero exits and actually throws an R error when the command fails to run at all.
I’ll switch to a` tryCatch()` block that handles both warnings and errors as you suggested.

dlebauer (approved):
LGTM!

documentation/tutorials/Demo_1_Basic_Run/download_sipnet.R
```diff
@@ -39,19 +39,31 @@ download.file(
...
-
-status <- suppressWarnings(
-    system2(dest_path, "-h", stderr = TRUE, stdout = TRUE) |>
-        attr("status")
+tryCatch(
+  {
+    # This block runs if system2 succeeds with exit code 0 (status attribute is NULL).
+    # This is unexpected for `sipnet -h`, so we warn but assume it's OK.
+    system2(dest_path, "-h", stderr = TRUE, stdout = TRUE)
+    PEcAn.logger::logger.warn("SIPNET ran with exit code 0, but expected 1. Assuming installation is OK.")
```
infotroph, line 47: Wait, why are you issuing a warning here? exit code 0 = successful execution = yes, the installation works.

infotroph (changes requested):

documentation/tutorials/Demo_1_Basic_Run/download_sipnet.R
AritraDey-Dev: yes it should print a success message instead of warning.

AritraDey-Dev: Done here https://github.com/PecanProject/pecan/pull/3645

PR #3643, opened 2025-10-08
gSSURGO: Fix spatial sampling and improve data aggregation accuracy

### LINKED ISSUE #3609: Restore Area-Weighted Sampling for SSURGO
### Bug Description

Current sampling in `extract_soil_gssurgo()` uses counts, not true area-weighted aggregation. This can bias soil property estimates, especially for heterogeneous polygons.

## Solution

- Use SoilDB package
- Not sure what of the following is handled by that package, but sampling should:
  - Request SSURGO polygon geometries (GML) from WFS for the sampling area.
  - Intersect polygons with sample buffers e.g. using `sf` or `terra`.
  - Compute area fractions
  - Aggregate soil properties using these area weights.

## References

- Chris' comments and related discussion on PR 3534 https://github.com/PecanProject/pecan/pull/3534#issuecomment-3225086247
- Code that previously did area weighting was unintentionally removed in #2964

divine7022:
## Description
Refactored `extract_soil_gssurgo()` to replace point-based WFS queries with raster-based Web Coverage Service (WCS) approach, enabling accurate area-weighted sampling and eliminating spatial coverage gaps.

* Implemented bounding box raster query using `soilDB::mukey.wcs()` to extract map unit keys with complete spatial coverage at 30m resolution. This provides accurate area weighting through pixel counts and eliminates spatial coverage gaps.
* Utilized `soilDB::get_SDA_property()` with "Weighted Average" aggregation method. This retrieves soil properties (sand, silt, clay, organic matter, bulk density) integrated across specified depth ranges with component weighting in a single batch query.
* Integrated `soilDB::fetchSDA()` to obtain complete rock fragment data (`fragvol_r`) representing total volume across all size classes: 2-75mm (pebbles), 75-250mm (cobbles), 250-600mm (stones), and >600mm (boulders). Applied proper depth and component weighting for fragments:

## Motivation and Context
Fixes #3609 

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ section for the next release.
...
 - Fixed "external pointer is not valid" error and addressed key bugs in `soilgrids_soilC_extract()` function (#3506)
 - Fixed a bug within the `model2netcdf.SIPNET` function where we assumed the constant calculations of `pecan_start_doy` across years (the calculations should vary depending on the last date from the last loop and the start date of the current loop), which will lead to incorrect calculations of the start `sub_dates` and `sub_dates_cf` if we are jumping between years (e.g., from 2012-12-31 to 2013-01-01). The `sipnet2datetime` function is no longer used anywhere and therefore has been removed.
+- `extract_soil_gssurgo()` -- replaced point-based WFS queries with raster-based WCS approach using `soilDB::mukey.wcs()` for accurate area-weighted sampling. Integrated `soilDB::get_SDA_property()` for depth-integrated soil property retrieval and `soilDB::fetchSDA()` for comprehensive rock fragment data across all size classes. This eliminates spatial coverage gaps and reduces network requests while maintaining backward compatibility.
```
dlebauer:
```suggestion
- `extract_soil_gssurgo()` #3643
  - replaced point-based WFS queries with raster-based WCS approach using `soilDB::mukey.wcs()` for accurate area-weighted sampling. 
  - Integrated `soilDB::get_SDA_property()` for depth-integrated soil property retrieval and `soilDB::fetchSDA()` for comprehensive rock fragment data across all size classes. This eliminates spatial coverage gaps and reduces network requests while maintaining backward compatibility.
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
+        dplyr::mutate(
+          depth_layer = depths[i],
+          hzdept_r = top_depth,
+          hzdepb_r = bottom_depth
+        )
+    } else {
+      # Keep other soil data, mark fragments as explicitly missing
+      # complete.cases() will filter these out later
+      PEcAn.logger::logger.info(
+        paste("Fragment data unavailable for depth", top_depth, "-", bottom_depth,
```
dlebauer: Note that the logger has its own internal pasting approach. See output of `PEcAn.logger:::logger.message`.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -205,35 +178,34 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
-  #- see if we need to generate soil ensemble and add that to the list of all
+  # Generate modeled ensembles
   tryCatch({
-    # find the soil depth levels based on the depth argument 
-    # if soil profile is deeper than what is specified in the argument then I go as deep as the soil profile.
-    if (max(soilprop.new$soil_depth) > max(depths)) {
-      depths <- sort(c(depths, max(soilprop.new$soil_depth)))
+    # Adjust depth levels if needed
+    if (max(soilprop.new$soil_depth_bottom) > max(depths_cm)) {
+      depths_cm <- sort(c(depths_cm, max(soilprop.new$soil_depth)))
```
dlebauer: I don't understand fully, but the code compares against the deepest bottom soil depth but appends the (soil_depth) should this be:

```suggestion
      depths_cm <- sort(c(depths_cm, max(soilprop.new$soil_depth_bottom)))
```
?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
+            na.rm = TRUE
+          ),
+          .groups = "drop"
+        )
+      
+      # Merge soil properties with fragment data
+      depth_data <- soil_props %>%
+        dplyr::left_join(fragment_data, by = "mukey") %>%
+        dplyr::mutate(
+          depth_layer = depths[i],
```
dlebauer: is depth_layer used downstream? Should it be? Or should it be dropped?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
+  all.soil.ens <- list()
   
-  # Convert single center lat/lon to projected coordinates
-  point_sf <- sf::st_sfc(sf::st_point(c(lon, lat)), crs = wgs84_crs)
-  point_proj <- sf::st_transform(point_sf, proj_crs)
-  coords_proj <- sf::st_coordinates(point_proj)
+  # create spatial bounding box
+  half_extent_m <- (grid_size - 1) / 2 * grid_spacing
+  lat_offset <- half_extent_m / 111000
+  lon_offset <- half_extent_m / (111000 * cos(lat * pi / 180))
```
dlebauer:
```suggestion
  # spherical approximation 
  # error (max 0.3% (3m/km)) << gssurgo 10m grid 
  lon_offset <- half_extent_m / (111320 * cos(lat * pi / 180))
```

dlebauer (changes requested):
This PR implements the features specified in #3609. Please check comments and suggestions, then will be ready to merge, thanks!

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
   )
-  grid_coords <- terra::crds(raster_template)
+  aoi <- sf::st_as_sfc(bbox)
   
-  # Transform grid coordinates back to WGS84 for gSSURGO queries
-  grid_sf <- sf::st_as_sf(data.frame(x = grid_coords[, 1], y = grid_coords[, 2]),
-                          coords = c("x", "y"), crs = proj_crs)
-  grid_wgs84 <- sf::st_transform(grid_sf, wgs84_crs)
-  grid_coords_wgs84 <- sf::st_coordinates(grid_wgs84)
+  PEcAn.logger::logger.info("Querying gSSURGO Web Coverage Service for map unit keys")
```
infotroph: this message seems unnecessary 
```suggestion
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
 #' @export
 #'  
 extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spacing=100, depths=c(0.15,0.30,0.60)){
```
infotroph: With the change back to a set distance instead of a grid, I think it would also make sense to revert the `grid_size` and `grid_spacing` parameters back to something more like the `radius` that was here before #3534.

infotroph: This can also greatly simplify aoi setup -- lines 30 through 42 below could reduce to
```
aoi <- data.frame(lon = lon, lat = lat) |>
  terra::vect(crs = "epsg:4326") |>
  terra::buffer(radius)
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
-          NULL
-        }
-      } else {
-        PEcAn.logger::logger.debug(paste("No MapUnitKeyList found for coordinates", 
-                                         this_lat, ",", this_lon, "skipping grid point"))
-        NULL
-      }
+    # Use fetchSDA instead of get_SDA_property to obtain complete rock fragment data
+    # get_SDA_property only provides frag3to10_r and fraggt10_r 
+    # but fetchSDA returns fragvol_r which represents TOTAL rock fragment volume including
```
infotroph: For the record the smaller size classes _are_ provided, but they're named differently: `sieveno10_r` is the percentage of the less-than-3-inch fraction that passes through a 2-mm sieve, so `(1-sieveno10_r) + frag3to10_r + fraggt10_r` is a defensible total... the catch is that's a percentage by _weight_, whereas fragvol_r is the percentage by _volume_ we want for C stock adjustment.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
+    # all size classes: 2-75mm (pebbles), 75-250mm (cobbles), 250-600mm (stones), and >600mm (boulders).
+    # plus component weighting needed for aggregation 
+    sda_data <- tryCatch({
+      soilDB::fetchSDA(
```
infotroph: Every call to fetchSDA returns all horizons from all components in the requested map units, so this does not need to be inside the loop over depths.

Also, in my local testing it seems `all(c("sandtotal_r", "silttotal_r", "claytotal_r", "om_r", "dbthirdbar_r") %in% names(sda_data))` is TRUE, so it may make sense to skip the `get_SDA_property` call entirely and work just from this result.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -205,35 +178,34 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
-  all.soil.ens <-c(all.soil.ens, list(soil.data.gssurgo))
   
+  all.soil.ens <- c(all.soil.ens, list(soil.data.gssurgo))
   
-  # What I do here is that I put soil data into depth classes and then model each class speparatly
-  #- see if we need to generate soil ensemble and add that to the list of all
+  # Generate modeled ensembles
   tryCatch({
-    # find the soil depth levels based on the depth argument 
-    # if soil profile is deeper than what is specified in the argument then I go as deep as the soil profile.
```
infotroph, line 235: I know this was pre-existing behavior, but I've never understood it -- why add values deeper than the user requested? Seems like especially for the properties that are queried separately for each layer, this is just extra run time and bandwidth. Am I missing a reason this is necessary?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -242,76 +214,65 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
-          # Handle edge cases for SOC simulation
-          if (nrow(DepthL.Data) == 1) {
-            simulated_soc <- rep(NA_real_, size)
-          } else if (is.na(soc_sd) || soc_sd == 0) {
+          # SOC modeling
+          soc_mean <- mukey_group$soil_organic_carbon_stock
+          soc_sd <- stats::sd(soc_mean, na.rm = TRUE)
+          n_depths <- length(soc_mean)
+          
+          if (n_depths == 1 || is.na(soc_sd) || soc_sd == 0) {
```
infotroph: Why NA when n_depths == 1? I would have expected that 1 layer still means one datapoint to be had

modules/data.land/R/extract_soil_nc.R
```diff
@@ -362,19 +319,16 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
     })
-  # removing the nulls or the ones that throw exception in the above trycatch
-  out.ense<- out.ense %>%
+  
+  # remove nulls 
+  out.ense <- out.ense %>%
     purrr::discard(is.null)
   
-  out.ense<-out.ense %>% 
+  out.ense <- out.ense %>%
```
infotroph: An additional refactor that would make this read easier: Getting to the bottom of the function I notice there's a long series (at least lines 307-330) of assignments to `out.ense` that each perform one mutation -- this is a pipeline! Might want to consider writing it that way explicitly by changing the repeated `out.ense <- out.ense...` to `%>%`

infotroph (changes requested):

modules/data.land/R/extract_soil_nc.R
dlebauer: @infotroph are there guidelines or do you have suggestions for when to (or not to) use logger.info?

infotroph: This specific suggestion is just from the principle of "Don't emit a message unless the user needs to read and act on it", which I don't think they do here. Especially since this is called once per site, so every extra line of logs here will turn into a lot of console spam in a multisite workflow.

modules/data.land/R/extract_soil_nc.R
infotroph: Followup thought: I'd also love to have support for passing in a custom area of interest so that I could retrieve e.g. soil properties for one whole field instead of a fixed radius/grid that approximates it. I think the change above would make that pretty easy -- if the user provides a lat/lon then construct AOI as shown here, if they provide a polygon then pass that straight into `mukey.wcs`.

Does the following (!untested!) design make sense to you?

```suggestion
extract_soil_gssurgo <- function(outdir, lat, lon, aoi = NULL, size = 1, radius = 500, depths=c(0.15, 0.30, 0.60)) {
  if (missing(aoi)) {
    aoi <- data.frame(lon = lon, lat = lat) |>
      terra::vect(crs = "epsg:4326") |>
      terra::buffer(radius)
  }
```

... along with deletion of current lines 30 to 42 below?

modules/data.land/R/extract_soil_nc.R
```diff
@@ -31,168 +25,147 @@
...
-  mukeys_all <- unique(mukeys_all)
-  if (length(mukeys_all) == 0) {
-    PEcAn.logger::logger.severe("No mapunit keys were found for this site.")
-    return(NULL)
+    all_soil_data[[i]] <- depth_data
+    # Loop continues to next depth layer regardless
   }
   
-  # calling the query function sending the mapunit keys
-  soilprop <- gSSURGO.Query(
```
infotroph, line 156: Does the approach used here completely replace `gSSURGO.Query`? If so we might be able to deprecate it now for removal in a future release -- as far as I know it is mostly (only?) used inside extract_soil_nc.

modules/data.land/R/extract_soil_nc.R
infotroph: I talked myself out of this upon realizing this message helps give context to the messages printed by soilDB, e.g. `mukey.wcs` always reports "Downloaded [n] bytes" and without this line it's not obvious what was downloaded from where. Finding a way to suppress those message would be a worthy project for another day, but until then I'm OK with this logger line staying.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -242,76 +214,65 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
     }
     
-    soilprop.new.grouped<-soilprop.new %>% 
-      dplyr::mutate(DepthL=depths[depth.levs])
+    soilprop.new.grouped <- soilprop.new %>% 
+      dplyr::mutate(DepthL = depths_cm[depth.levs])
     
-    # let's fit dirichlet for each depth level separately
-    simulated.soil.props<-soilprop.new.grouped %>%
-      split(list(soilprop.new.grouped$DepthL, soilprop.new.grouped$mukey)) %>%
```
infotroph, line 255: @divine7022 can you explain more of the intention here? It looks like removing the split by depth switches this from fitting a texture for each depth layer to fitting one texture across the entire depth profile, which is not what we want.

CHANGELOG.md
divine7022: 👍

modules/data.land/R/extract_soil_nc.R
divine7022: thanks for the info!

modules/data.land/R/extract_soil_nc.R
divine7022: this is not necessary here now -- since aggregation already CLIPS to `bottom_depth` and it always sets to user requested depth

modules/data.land/R/extract_soil_nc.R
divine7022: original purpose `depth_layer` was meant to track which depth interval (0.15, 0.30, 0.60) each record belonged to ; the code uses `soil_depth` and `soil_depth_bottom` (from hzdept_r and hzdepb_r) for depth tracking, later in code DepthL is computed from `findInterval()` to assign depth levels;
So the answer for your question it's an historical Artifact. 
Thanks removed!

modules/data.land/R/extract_soil_nc.R
divine7022: now switched to `radius` and `aoi`, no more gird sampling. see [comment](https://github.com/PecanProject/pecan/pull/3643#discussion_r2480599781)

modules/data.land/R/extract_soil_nc.R
divine7022: i don't really have a strong preference here radius vs grid. Switching to `radius` makes more sense since we are no longer doing grid sampling (single `fetchSDA` call)
implemented changes, thanks!

modules/data.land/R/extract_soil_nc.R
divine7022: this is not necessary to be here now -- since aggregation already CLIPS to `bottom_depth` and it always sets to user requested depth

modules/data.land/R/extract_soil_nc.R
divine7022: I was visioning back then ,if we have one record it's oboes that the sd become NA.
But good catch!, returning NA for a single observation discards valid data unnecessarily
i have update code to :
```
if (is.na(soc_sd) || soc_sd == 0) {
  # No variability - use measured value (all ensemble members identical)
  simulated_soc <- rep(soc_mean, size)
} else {
  # Has variability - sample from gamma distribution
  shape <- (soc_mean^2) / (soc_sd^2)
  rate <- soc_mean / (soc_sd^2)
  simulated_soc <- stats::rgamma(size, shape = shape, rate = rate)
}
```
single component (n=1) `sd = NA` all ensemble members will get measured value. Multiple different components (`sd > 0`) sample from distribution.
make sense now ?

modules/data.land/R/extract_soil_nc.R
divine7022: done 👍

modules/data.land/R/extract_soil_nc.R
divine7022: good question! suspecting this is the point where i was got stuck in with the decision

I reverted back to splitting by both depth AND mukey.
The initial change was an experiment to increase sample size for Dirichlet fitting, and get vertical variability.

Context for why I tried it:
The old code used `get_SDA_property` which returned one pre-aggregated value per (mukey, depth) - essentially 1 observation per group, with n=1 dirichlet parameter estimation is unstable (requires min ~3-5 samples for reasonable MLE convergence).
The new changes uses single `fetchSDA` call with component-level aggregation, which preserves multiple components per (mukey, depth).

modules/data.land/R/extract_soil_nc.R
divine7022: Yes new approach completely replaces `gSSURGO.Query` and uses `fetchSDA`, As far as I can tell, `gSSURGO.Query` is only called within `extract_soil_nc.R` (now removed), so user impact should be minimal.

modules/data.land/R/extract_soil_nc.R
divine7022: I have implemented both suggestions in the recent commit, now `fetchSDA` retrieves all horizons once, then filters by depth in the loop - much more efficient for complex query

modules/data.land/R/extract_soil_nc.R
divine7022: Thanks for the detailed explanation!, 
`(1 - sieveno10_r/100) + frag3to10_r + fraggt10_r` gives total rock fragment weight `%`

modules/data.land/R/extract_soil_nc.R
```diff
@@ -24,27 +24,26 @@
...
+    PEcAn.logger::logger.severe("Must provide either 'aoi' OR both 'lat' and 'lon'")
+  }
   
+  # Create AOI from point + radius if needed
+  if (is.null(aoi)) {
+    aoi <- data.frame(lon = lon, lat = lat) %>%
+      terra::vect(crs = "epsg:4326") %>%
+      terra::project("epsg:5070") %>%
+      terra::buffer(width = radius) %>%
+      terra::project("epsg:4326")
```
infotroph: Reprojecting introduces distortion that can be severe, and isn't needed because terra knows how to accurately buffer in meters from a geographic CRS. Better to just pass the latlongs and let it buffer from those.
```suggestion
      terra::buffer(width = radius)
```

(for a visual demo of the distortion issue, compare the circular result of `data.frame(lat=70, lon = 0) |> vect(crs = "epsg:4326") |> buffer(1000) |> plot()` to the very elongated result of `data.frame(lat=70, lon = 0) |> vect(crs = "epsg:4326") |> project("epsg:5070") |> buffer(1000) |> project("epsg:4326") |> plot()`)

dlebauer (approved):
This looks good to me. Thanks for all of your work on this! 

@infotroph it looks like @divine7022 has addressed your suggestions and answered your questions. This seems to be a substantial improvement. Are there any errors that would prevent this from being merged?

modules/data.land/R/extract_soil_nc.R
infotroph: Thanks for reverting. For future work, changes that fundamentally alter a function's behavior ought to be called out in the PR description, especially if they're experimental -- I didn't notice this until I tried a local test run and wondered why it produced files with all layers identical.

modules/data.land/R/extract_soil_nc.R
```diff
@@ -30,172 +24,121 @@
...
+      dplyr::mutate(
+        hz_top_adj = pmax(hzdept_r, top_depth),
+        hz_bot_adj = pmin(hzdepb_r, bottom_depth),
+        hz_thickness = hz_bot_adj - hz_top_adj
+      ) %>%
+      dplyr::group_by(mukey, cokey, comppct_r) %>%
+      dplyr::summarise(
+        sandtotal_r = stats::weighted.mean(sandtotal_r, hz_thickness, na.rm = TRUE),
+        silttotal_r = stats::weighted.mean(silttotal_r, hz_thickness, na.rm = TRUE),
+        claytotal_r = stats::weighted.mean(claytotal_r, hz_thickness, na.rm = TRUE),
```
infotroph: Could treating these as three independent averages cause any problems for Dirichlet fitting below  in cases where the horizons we're averaging together have strong discontinuities in texture? I haven't worked through the math yet, but maybe you did already.

(By "discontinuity" I mean cases where, say, a horizon with 30% sand and a horizon with 70% sand get combined into a weighted average of 50%. Since soil mappers very often assign horizon boundaries on the basis of a change in texture, this will be a pretty common occurrence in the database.)

infotroph (commented):
A few more comments here. I haven't finished my re-review yet, but I can already say that given the magnitude of the changes it would be very helpful to include some test cases  so we can verify the function is behaving as intended.

modules/data.land/R/extract_soil_nc.R
divine7022: when we calculate three independent weighted means across horizons with texture discontinuities, floating point arithmetic can cause the sum to deviate slightly from 100%. While these deviations are small, they violate the strict constraint required by dirichlet distn.
I have added closure normalization immediately after aggregation
thanks!

modules/data.land/R/extract_soil_nc.R
divine7022: that makes sense, removed EPSG:5070 reprojection and now apply terra::buffer() directly on the EPSG:4326 geometry

modules/data.land/R/extract_soil_nc.R
infotroph: I was more wondering if the result will make sense in cases where different layers have very different amounts of sand/silt/clay, and especially I was worrying about cases like "layer 1 has a lot of sand, a lot of silt, and some clay; layer 2 has a lot of sand, a lot of clay, and next to no silt."

On working through some examples, I _think_ that's not a problem because even though the means of the three components are computed separately, they all use the same set of weights (layer thickness), so any proportions that summed to 1 within layers before averaging will keep their relative proportions after averaging.

divine7022: Exactly, same weights across all components = proportions preserved
Thanks for verifying!

modules/data.land/R/extract_soil_nc.R
```diff
@@ -30,172 +24,127 @@
...
+  hz_data <- aqp::horizons(sda_data)
+  site_data <- aqp::site(sda_data)
+  
+  # Component-level aggregation by depth
+  depths_cm <- depths * 100
+  all_soil_data <- list()
+  
+  for (i in seq_along(depths_cm)) {
+    if (i == 1) {
+      top_depth <- 0
```
mdietze: FYI you could eliminate this `if` by requiring the first depth to be 0 and then shifting the indexing. This isn't unreasonable as it's the exact same behaviour as `hist` when you set a vector of breaks (i.e. there is one more break than there is bins)

divine7022: this makes sense, so here i would go for adding a minimal fix -- rather then adding a new value to default parameter `depths` and break the folks using legacy workflow  ?

simplest fix would be to remove the `if` block and :
```
depths_cm <- c(0, depths * 100) 

for (i in seq_len(length(depths_cm) - 1)) {
  top_depth <- depths_cm[i]
  bottom_depth <- depths_cm[i + 1]
  ```

Comment (mdietze) [reply modules/data.land/R/extract_soil_nc.R]: This is a case where it's fine to break the legacy workflow, just (a) fix the defaults and (b) find the places where the function is being called and update the arguments. But if you do attempt to prepend the vector with 0, it would be good to first check that the first argument isn't already 0

File: modules/data.land/tests/testthat/test-extract_soil_nc.R [hunk 18]
```diff
@@ -90,16 +89,147 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+    lon = -88.2434,
+    size = 1,
+    radius = 1000,
+    depths = c(0, 0.15)
+  )
+  
+  expect_false(is.null(res_small))
+  expect_false(is.null(res_large))
+  expect_type(res_small, "list")
+  expect_type(res_large, "list")
```
infotroph: checking both for null and for list type looks redundant to me -- wouldn't a null result be caught by expect_type as not a list?

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,147 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+})
+
+test_that("extract_soil_gssurgo component-level variability is preserved", {
```
infotroph: Not sure what this means

modules/data.land/R/extract_soil_nc.R
```diff
@@ -205,127 +158,114 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
     
-    # estimating the proportion of areas for those mukeys which are modeled
-
-    # defining mukey_area
+    # Calculate area weights
     mukey_area <- data.frame(
       mukey = names(mukey_counts),
       Area = as.numeric(mukey_counts) / sum(mukey_counts)
     ) %>%
       dplyr::filter(.data$mukey %in% unique(simulated.soil.props$mukey)) %>%
```
infotroph:
```
> PEcAn.data.land::extract_soil_gssurgo(outdir=tempdir(),lat=38.56,lon=-121.9,size=1)
[snip]
Warning messages:
1: There was 1 warning in `dplyr::filter()`.
ℹ In argument: `.data$mukey %in% unique(simulated.soil.props$mukey)`.
Caused by warning:
! Unknown or uninitialised column: `mukey`. 
2: Unknown or uninitialised column: `mukey`. 
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -30,172 +24,131 @@
...
-      bulk_density = "dbthirdbar_r", # g/cm3
-      coarse_fragment_pct = "fragvol_r", # %
-      mukey = "mukey") %>%
+      fraction_of_sand_in_soil = "sandtotal_r",
+      fraction_of_silt_in_soil = "silttotal_r",
+      fraction_of_clay_in_soil = "claytotal_r",
+      soil_depth = "hzdept_r",
+      soil_depth_bottom = "hzdepb_r",
+      organic_matter_pct = "om_r",
+      bulk_density = "dbthirdbar_r",
```
infotroph, line 59: In local testing with commit ae98820864 , `bulk_density` seems to be either `NaN` or `1350` in every layer of every soil I check.

infotroph (changes requested):

modules/data.land/R/extract_soil_nc.R
```diff
@@ -205,127 +158,114 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
-          dir.model <-DepthL.Data[,c(1:3)] %>%
-            as.matrix() %>%
-            sirt::dirichlet.mle(.)
-          # Monte Carlo sampling based on my dirichlet model
-          alpha <- dir.model$alpha
-          alpha <- matrix(alpha, nrow= size, ncol=length(alpha), byrow=TRUE )
+          texture_data <- DepthL.Data[,c("fraction_of_sand_in_soil",
+                                         "fraction_of_silt_in_soil",
+                                         "fraction_of_clay_in_soil")] %>%
+            as.matrix()
```
infotroph: dirichlet.mle is documented to take a dataframe, so we can skip one transformation here
```suggestion
                                         "fraction_of_clay_in_soil")]
```

modules/data.land/R/extract_soil_nc.R
```diff
@@ -205,127 +158,114 @@ extract_soil_gssurgo <- function(outdir, lat, lon, size=1, grid_size=3, grid_spa
...
+          
+          if(nrow(texture_data) == 0) return(NULL)
+          
+          # Fit Dirichlet on component textures
+          dir.model <- sirt::dirichlet.mle(texture_data)
```
infotroph: I'm seeing a lot of MLE fit failures and on investigation it looks like these come from soils with only one component (and therefore only one texture estimate) in the layer. What if we treat that texture as our estimate?

@mdietze I want to be careful not to embed too many assumptions about variability here -- does this choice of dispersion parameter (basically "assume it's about as variable as most soils are") seem acceptable? 

One more difficult alternative would be to start parsing the high and low values (sandtotal_l, sandtotal_h, etc) for each texture component as well as the representative values (sandtotal_r), but that would be a bigger refactor and they are not universally populated, so there would still be soils that needed a fallback approach.

```suggestion
          if (nrow(texture_data) == 0) {
            # No data, can't fit anything
            return(NULL)
          } else if (nrow(texture_data) == 1) {
            # With only one value an MLE fit will fail.
            # Instead use the reported means,
            # setting the dispersion value `alpha0` to give 90% intervals
            # on the order of 20 percentage points each direction from the mean
            # (comparable to typical variation within one soil type).
            alpha0 <- 10
            dir.model <- list(alpha = unlist(texture_data) * alpha0)
          } else {
            # Fit Dirichlet on component textures
            dir.model <- sirt::dirichlet.mle(texture_data)
          }
```

modules/data.land/R/extract_soil_nc.R
infotroph: On consideration having qSSURGO.Query be a separate function is a good design because it separates _fetching_ database values from _generating_ our own ensemble values. In testing this PR I've found myself very often wanting to look at the original gSSURGO results to debug the behavior of the ensemble-generation part, and the unit testing could also be simpler if these concerns were separated.

I now think we should pull out to their own function the pieces of `extract_soil_gssurgo` that perform the query; since whose inputs are AOI or lat/lon/radius, I think it should be a new function rather than replacing the existing `gSSURGO.Query` (which takes mukeys). Maybe something like `gssurgo_fetch_area`?

modules/data.land/R/extract_soil_nc.R
infotroph: I now think this is an issue in `soil_params` rather than this PR, but it made me realize the bulk density in the output files is estimated by PTF and the database value we've captured here does not get passed through to the ensemble. Would it make sense to gamma-fit bulk density the same way we do SOC?

infotroph (changes requested):

modules/data.land/R/extract_soil_nc.R
infotroph: Discussion on Slack is leaning toward making nrow = 1 be a failure case unless size=1 as well, in which case returning means is correct behavior. Rationale: If we don't know the variability, we just don't know it; assuming any particular value would require justification with a more formal analysis (/citation to someone else's analysis).

Bottom line is this should be handled by parsing the low/representative/high values (where provided) instead of by assuming between-components variation is the only important source of heterogeneity.

modules/data.land/tests/testthat/test-extract_soil_nc.R
divine7022: renamed and clarified

modules/data.land/R/extract_soil_nc.R
divine7022: fixed by adding validation before the dplyr::filter() call. When all dirichlet mle fits fail (e.g. all single-component soils with size > 1), purrr::map_df() returns an empty data frame without the mukey column

modules/data.land/R/extract_soil_nc.R
divine7022: The base ensemble member now includes soil_bulk_density converted to proper units
and added component-weighted gamma sampling for bulk density, parallel to the SOC approach
The gamma distribution is appropriate here as bulk density is strictly positive and often right-skewed in heterogeneous soils. When only one component exists (no between component variance), we preserve the database value without artificial inflation of uncertainty.

modules/data.land/R/extract_soil_nc.R
divine7022: per Mike's suggestion,
for size = 1 (point estimate), return the database mean values directly (represent best estimate values)

for size > 1 with single-component soils, log a warning and return NULL for that mapunit/depth combination. We don't fabricate uncertainty where we have no data to support it.

future work: as discussed, parsing sandtotal_l/sandtotal_h (low/high values) from gSSURGO would provide uncertainty bounds, I will open an issue to track this enhancement

modules/data.land/R/gSSURGO_Query.R
```diff
@@ -134,3 +134,134 @@ gSSURGO.Query <- function(mukeys,
...
+      terra::vect(crs = "epsg:4326") %>%
+      terra::buffer(width = radius)
+  }
+  
+  # Validate depths parameter
+  if (depths[1] != 0) {
+    PEcAn.logger::logger.severe(
+      "First depth must be 0. Use depths = c(0, 0.15, 0.30, ...) like hist() breaks."
+    )
+  }
```
infotroph, line 186: Per conversation (much) earlier in this PR, this is clarifying and simplifying behavior that already existed. 👍 to that.

But! This function is the only PEcAn soil tool I can find that expects top-of-layer depths; the others either don't say or expect bottom-of-layer depth: 
* `standard_vars.csv` is [explicit](https://github.com/PecanProject/pecan/blame/develop/base/utils/data/standard_vars.csv#L4) that depth means to bottom
* The SoilGrids tools [use bottom depth](https://github.com/PecanProject/pecan/blob/develop/modules/data.land/R/soil_params_ensemble.R#L137-L140) when one number is needed, but also are good about keeping labels with explicit top and bottom depths in most places.
* `soil.units` [does not say](https://github.com/PecanProject/pecan/blob/develop/modules/data.land/R/extract_soil_nc.R#L530).
* `soil2netcdf` doesn't state either way (nor do any internals care), but its [example uses bottom depth](https://github.com/PecanProject/pecan/blob/develop/modules/data.land/R/soil2netcdf.R#L26-L27)
* `write.config.SIPNET` [explicitly treats](https://github.com/PecanProject/pecan/blob/develop/models/sipnet/R/write.configs.SIPNET.R#L570) depths as bottom of layer
* My currently open PR #3788 [will make](https://github.com/PecanProject/pecan/pull/3788/files#diff-a67d543ff1cf7c6744f8c07d81a3f1ebf3bbebb7144e026e884ddc62ec9db802R43) RothC assume depths as bottom of layer if nobody stops me first 😈. Working on that is what made me notice the difference from how this here gSSURGO function works. 

See also #3775

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,174 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+    outdir = tmp_outdir,
+    lat = 40.1164,
+    lon = -88.2434,
+    size = 1,
+    radius = 500,
+    depths = c(0, 0.15, 0.30)  # Correct format
+  )
+  
+  expect_false(is.null(res))
+})
```
infotroph, line 236: Depth validation happens before any API calls, so it's safe to remove the skips from this. Avoiding boilerplate also makes it much easier to see what's actually happening in this test:
```suggestion
test_that("extract_soil_gssurgo requires depths to start with 0", {
  expect_error(
    extract_soil_gssurgo(
      outdir = withr::local_tempdir(),
      lat = 40,
      lon = -88,
      depths = c(0.15, 0.30)  # Missing 0 at start
    ),
    regexp = "First depth must be 0"
  )
})
```

(Personally I'd also move this to the tests for gssurgo_fetch_area and keep this file focused on testing the ensemble fitting portions, possibly using mocked results from gssurgo_fetch_area to do it)

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,174 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+    outdir = tmp_outdir,
+    lat = 40.1164,
+    lon = -88.2434,
+    size = 1,
+    radius = 1000,
+    depths = c(0, 0.15)
+  )
+  
+  expect_type(res_small, "list")
+  expect_type(res_large, "list")
```
infotroph, line 163: Note that this doesn't actually check that the change in buffer had any effect! Am I right we should expect `re_larges$mukey_counts` to sum to more pixels than `res_small$mukey_counts`?

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,174 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+    depths = c(0, 0.15, 0.30)
+  )
+  
+  expect_false(is.null(res))
+  expect_gt(length(res), 2)
+  
+  if (requireNamespace("ncdf4", quietly = TRUE) && length(res) >= 3) {
+    # Compare two different ensemble members (skip first - it's unsampled)
+    nc1 <- ncdf4::nc_open(unlist(res)[2])
+    nc2 <- ncdf4::nc_open(unlist(res)[3])
```
infotroph, line 190: It's possible you could reduce boilerplate by using `soil_ic_netcdf2list()` as a helper to read the nc files

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,174 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+  
+  res <- extract_soil_gssurgo(
+    outdir = tmp_outdir,
+    aoi = aoi,
+    size = 2,
+    depths = c(0, 0.15, 0.30)
+  )
+  
+  expect_false(is.null(res))
+  expect_type(res, "list")
```
infotroph, line 130: I'm fairly sure testing if the result is a list is also an effective check that it's not null. Not a big deal in one place, but across many test blocks this adds up to a lot of extra lines to scroll through without much added information about the behavior of the function.
```suggestion
  expect_type(res, "list")
```

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,174 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+  expect_true("mukey_counts" %in% names(result))
+  expect_true("depths_cm" %in% names(result))
+  
+  # Validate raw data structure
+  expect_s3_class(result$soilprop, "data.frame")
+  expect_true(all(c("sandtotal_r", "silttotal_r", "claytotal_r", 
+                    "mukey", "cokey") %in% names(result$soilprop)))
+  
+  # Values should be in original units (percentages, not fractions)
+  expect_true(all(result$soilprop$sandtotal_r <= 100, na.rm = TRUE))
```
infotroph, line 260: Note that this test won't fail if they _are_ fractions! Would checking that `sandtotal+silttotal+claytotal == 100` be too strict?

CHANGELOG.md
```diff
@@ -64,6 +64,11 @@ For more information about this file see also [Keep a Changelog](http://keepacha
...
 - Fixed "external pointer is not valid" error and addressed key bugs in `soilgrids_soilC_extract()` function (#3506)
 - Fixed a bug within the `model2netcdf.SIPNET` function where we assumed the constant calculations of `pecan_start_doy` across years (the calculations should vary depending on the last date from the last loop and the start date of the current loop), which will lead to incorrect calculations of the start `sub_dates` and `sub_dates_cf` if we are jumping between years (e.g., from 2012-12-31 to 2013-01-01). The `sipnet2datetime` function is no longer used anywhere and therefore has been removed.
+- `extract_soil_gssurgo()` (#3643)
+  - Replaced point-based WFS queries with raster-based WCS approach using `soilDB::mukey.wcs()`.
+  - Replaced `grid_size`/`grid_spacing` parameters with `radius` (meters) for simpler buffer-based AOI creation.
+  - Switched to single `soilDB::fetchSDA()` call for component-level soil data retrieval, enabling better ensemble uncertainty quantification through within-map-unit variability. Added support for custom AOI polygons. 
+  - This eliminates spatial coverage gaps and reduces network requests while maintaining backward compatibility.
```
infotroph, line 81: Move up to line 25-ish

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -12,11 +12,10 @@ test_that("extract_soil_gssurgo returns valid NetCDF files for valid US coordina
...
     lon = lon,
     size = 2,
-    grid_size = 3,
-    grid_spacing = 100,
-    depths = c(0.15, 0.30)
+    radius = 500,
+    depths = c(0, 0.15, 0.30)
   )
-
+
```
infotroph:
```suggestion

```

modules/data.land/tests/testthat/test-extract_soil_nc.R
infotroph: No, I'm thinking of gssurgo_fetch_area.... But this test would give about the same information if it called gssurgo_test_area directly.

modules/data.land/tests/testthat/test-extract_soil_nc.R
```diff
@@ -90,16 +89,174 @@ test_that("extract_soil_gssurgo handles ensemble generation", {
...
+  expect_true("depths_cm" %in% names(result))
+  
+  # Validate raw data structure
+  expect_s3_class(result$soilprop, "data.frame")
+  expect_true(all(c("sandtotal_r", "silttotal_r", "claytotal_r", 
+                    "mukey", "cokey") %in% names(result$soilprop)))
+  
+  # Values should be in original units (percentages, not fractions)
+  expect_true(all(result$soilprop$sandtotal_r <= 100, na.rm = TRUE))
 })
```
infotroph:
```suggestion
})

```

modules/data.land/R/gSSURGO_Query.R
```diff
@@ -134,3 +134,134 @@ gSSURGO.Query <- function(mukeys,
...
+  
+  soilprop <- do.call(rbind, all_soil_data)
+  
+  return(list(
+    soilprop = soilprop,
+    mukey_counts = mukey_counts,
+    depths_cm = depths_cm
+  ))
+}
```
infotroph:
```suggestion
}

```

infotroph (changes requested):

dlebauer:
What is the status of this PR?

Also, Does this pr #3455 overlap or augment other work that has been/will be done on the gssurgo handling?

PR #3649, opened 2025-10-14
fix(data.land): use year-specific crop data in generate_events

### LINKED ISSUE #3636: generate_events: use mapped from from each year
Comment by @infotroph in https://github.com/PecanProject/pecan/pull/3621#discussion_r2373178088 :

> This repeats the crop listed for 2016 (because that's what's in ca_field_attributes.csv) every year until 2024. Don't we want to be using the mapped crop for each year?

AritraDey-Dev:
## Description

Fix generate_events to use year-specific crop data instead of repeating 2016 crop data across all years

## Motivation and Context

Fixes #3636

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/data.land/inst/generate_events.R
```diff
@@ -49,9 +49,8 @@ if (DESIGN_POINTS) {
 }
 
 ca_fields <- ca_field_attributes |>
-  dplyr::select(site_id, pft, crop) |>
-  dplyr::distinct() |>
-  tidyr::crossing(year = 2016:2024) |>
+  dplyr::select(site_id, year, pft, crop) |>
+  dplyr::filter(year >= 2016 & year <= 2024) |>
```
infotroph, line 53: @AritraDey-Dev this looks right once we have data from multiple years in hand, but do note that the `ca_field_attributes` I know about currently only contains rows for 2016.  Have you tested this patch with a multi-year  field listing?

AritraDey-Dev: @infotroph I didn't test it separately with a multi-year field listing, given your comment on the issue that `ca_field_attributes` currently only contains rows for 2016, I've implemented a fix based on the issue you raised regarding the crop being repeated for every year until 2024.

infotroph: Tagging @dlebauer as the main user of this script so far -- your call whether we should to make this change now (and therefore have output be 2016-only until the field attributes file contains other years), or keep `crossing(2016:2024)` until we have an updated attribute file in hand?

infotroph:
ping @dlebauer -- this should be a quick merge once you weigh in above

dlebauer:
Unless it is being actively used and developed I'd suggest keeping as is until we update the script with multi-year data. Right now it is an artifact of an MVP that can be iteratively improved as we go.

PR #3651, opened 2025-10-18
feat(db): Re-enable PFT cultivar query test

AritraDey-Dev:
## Description

This pull request re-enables the `query.pft_cultivars` test in `base/db/tests/testthat/test-query.pft.R` by removing the `skip()` line.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/db/tests/testthat/test-query.pft.R
+++ b/base/db/tests/testthat/test-query.pft.R
@@ -36,7 +36,6 @@ test_that("nonexistant PFTs and modeltypes return empty dataframes", {
 
 
 test_that("query.pft_cultivars finds cultivars for a PFT", {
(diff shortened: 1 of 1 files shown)


infotroph (approved):
Thanks! 

If you want an easy follow-up, looks like the same skip message appears in `test-get.trait.data.pft.R` -- is it safe to re-enable that test too?

...and if you want a much less easy follow-up, this is a great example of a test that would benefit from being refactored to use mocks instead of relying on having the right state present in a live database. Every time I consider this I decide it isn't the highest priority for me, but I'll keep putting the idea out there 😉

PR #3653, opened 2025-10-20
fix: error in ma.R causing API container crashes

AritraDey-Dev:
## Description

This PR fixes issue in [ma.R](https://github.com/PecanProject/pecan/blob/develop/apps/api/R/ma.R) file that was causing the API container to fail and crash repeatedly. The following Docker logs helped identify the cause of the issue:
```
Error in parse(file, keep.source = FALSE, srcfile = src, encoding = enc) : 
  ma.R:55:5: unexpected '}'
54:      return(list(status = "Meta Analysis ran successfully", data = as.list(e))
55:     }
        ^
Error in parseUTF8(file) : Error sourcing ma.R
Calls: <Anonymous> -> initialize -> parseUTF8
Execution halted
```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/apps/api/R/ma.R
+++ b/apps/api/R/ma.R
@@ -51,7 +51,7 @@ submitWorkflow <- function(req, res){
      filepath <- paste0(settings$pfts$pft$outdir, "/post.distns.Rdata")
      e <- new.env(parent = emptyenv())
      load(filepath, envir = e)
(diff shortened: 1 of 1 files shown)


infotroph (approved):
Nice catch!

PR #3654, opened 2025-10-20
Partial fixes for multisite ensembles

infotroph:
## Description
A couple small fixes and a questionable hack for running ensemble and uncertainty analyses without database access.

- Make sure SA run directories contain site IDs, so that e.g. `SA-median/` isn't overwritten for each site in turn
- Only try to run SA on the PFTs actually present at a given site
- If Bety isn't available and no ensemble ID is present in the settings object, assign a new one by hashing the settings object.

Note that the latter step is done independently for each call to `write.*.configs`, so in a multisite run this will effectively set up a separate ensemble/SA for each site. This was what I wanted today, but I suspect most people will want outputs aggregated across sites, which this PR does not implement.

## Motivation and Context

For the MAGiC project I wanted to quickly evaluate AGB timeseries from many sites, for which the timeseries plots from the ensemble analysis would be perfect except that I'm running with no Bety access and the existing code sets the ensemble ID to `NOENSEMBLEID`, making each site overwrite the outputs from the previous one.

Since the issue applies to both ensemble and sensitivity I tried to implement a fix for both, but note that I focused on avoiding collisions between distinct ensembles -- there are still places where two sites with _the same_ ensemble ID will overwrite each other. 

I'm pasting my working notes below -- @divine7022 and @dlebauer will likely want to consider the unresolved issues in their work on multisite sensitivity.

* write.configs fails if SA is requested in a settings with ensemble size > 1
	- Workaround: run SA and full ensemble in separate settings files
	=> unresolved

* (minor): README.txt does not specify which met/IC/soil/event inputs were used
	=> unresolved

* rundir `SA-<pft>-<var>-<quantile>` contents are overwritten by each site in turn
	=> Resolved by adding site id to the get.run.id call

* rundir SA-median-<siteid> tries to run analysis for "ALL PFT", fails on NAs from pfts not present at that site
	=> Resolved by having run.sensitivity.analysis subset PFTs to those in run$site$site.pft. PFT doesn't show up in rundir names, but since only one per site it works.

* each site's call to run.write.configs overwrites `sensitivity.samples`
	- Since run.write.configs only sees one site at a time, need to choose one of:
		- write separate samples file for each site, combine later
		- append samples to existing samples file
		- move entire SA sample generation to a step not wrapped in papply
		- ?stop saving sensitivity.samples if not strictly needed after write.sa.configs is finished
			(But I think this is where run IDs are taken from)
	=> Unresolved

* runModule.run.sensitivity.analysis overwrites outputs as it runs for each site
	- Affects ensemble analysis too
	- The correct fix for this will probably parallel the fix for run.write.configs
	- Hacky workaround: Pass each site a different ensemble id to get n_sites separate outputs, then manually combine posthoc
		=> This workaround implemented by setting null ensemble.ids to `rlang::hash(settings)`,
			but if we might consider settings$run$site$id instead. Are there cases where a multiSettings might contain multiple entries from the same site? or where site ID would be unset?

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ section for the next release.
 
 ### Changed
 
+- Ensemble and sensitivity analyses run with now assign an ensemble ID if one is not specified in the XML, even when running with no DB (#TBD).
```
```suggestion
- Ensemble and sensitivity analyses run with now assign an ensemble ID if one is not specified in the XML, even when running with no DB (#3654).
```

infotroph:
@DongchenZ Will anything I propose here conflict with #3634?

divine7022:
LGTM, as discussed my debug shows the settings will get mutated between two calls of `write.*.configs` , so there won't be conflict between the ensemble ids of multisite settings with ensembles only required.
add this missing hash generation to 
`run.ensemble.analysis()` and `read.ensemble.ts`

infotroph:
> add this missing hash generation to run.ensemble.analysis() and read.ensemble.ts

@divine7022 Good catch that those are potential failure points. On a quick look I think their current behavior (of erroring if no ensemble id) is correct -- recall that the model runs have already executed by the time run.ensemble.analysis and read.ensemble.ts is called, so if we can't tell what ensemble ID was used at write.configs time then I think it's too late to create one in these steps. Does that seem right to you?

divine7022:
sounds correct!, realizing the different ensemble.id validation behaviors for sensitivity and ensemble runs ; and strict ensemble.id requirements for `run.ensemble.analysis`

modules/uncertainty/R/ensemble.R
```diff
@@ -281,14 +278,12 @@ for (input_tag in names(settings$run$inputs)) {
 
   
   # Get the workflow id
-  if (!is.null(settings$workflow$id)) {
-    workflow.id <- settings$workflow$id
-  } else {
-    workflow.id <- -1
-  }
+  workflow.id <- settings$workflow$id %||% -1
```
mdietze, line 283: I've never seen this operator before, and it's not the sort of thing one can easily google, so could you add a comment explaining what this step is doing.

modules/uncertainty/R/run.sensitivity.analysis.R
```diff
@@ -104,18 +107,12 @@ run.sensitivity.analysis <- function(settings,
...
-                                    all.var.yr = TRUE)
-    } else if (!is.null(settings$sensitivity.analysis$ensemble.id)) {
-      ensemble.id <- settings$sensitivity.analysis$ensemble.id
-      fname <- sensitivity.filename(settings, "sensitivity.samples", "Rdata",
-                                    ensemble.id = ensemble.id,
-                                    all.var.yr = TRUE)
-    } else {
-      ensemble.id <- NULL
-    }
+    ensemble.id <- ensemble.id %||%
```
mdietze, line 110: 1. comment doesn't match what the code actually does.
2. Explain code more clearly

modules/uncertainty/R/ensemble.R
infotroph: This is the scalar null coalescing operator, provided by base R since 4.4.0 and by {rlang} before that -- see `?"%||%"` for details. Basically `x %||% y` is "x unless it's null, in which case y"

modules/uncertainty/R/run.sensitivity.analysis.R
```diff
@@ -104,18 +107,12 @@ run.sensitivity.analysis <- function(settings,
 
     # Can specify ensemble ids manually. If not, look in settings.
     # If none there, will use the most recent, which was loaded with samples.Rdata
```
infotroph:
```suggestion
    # Ensemble ID is expected to be specified in function args or settings.
    # If none there, create one specific to this site.
```

modules/uncertainty/R/ensemble.R
```diff
@@ -217,37 +217,34 @@ get.ensemble.samples <- function( ensemble.size, pft.samples, env.samples,
...
 ##' @details The restart functionality is developed using model specific functions by calling write_restart.modelname function. First, you need to make sure that this function is already exist for your desired model.See here \url{https://pecanproject.github.io/pecan-documentation/latest/pecan-models.html}
 ##' new state is a dataframe with a different column for each state variable. The number of the rows in this dataframe needs to be the same as the ensemble size.
-##' State variables that you can use for setting up the intial conditions differs for different models. You may check the documentation of the write_restart.modelname your model.
+##' State variables that you can use for setting up the initial conditions differs for different models. You may check the documentation of the write_restart.modelname your model.
 ##' The units for the state variables need to be in the PEcAn standard units which can be found in \link{standard_vars}.
 ##' new.params also has similar structure to ensemble.samples which is sent as an argument.
 ##'
 ##' @importFrom dplyr %>%
 ##' @importFrom rlang .data
```
```suggestion
##' @importFrom rlang .data %||%
```

modules/uncertainty/R/ensemble.R
dlebauer: @mdietze 👍🏻 to adding a quick comment like "# x %||% y is "x unless it's null, in which case y" when it first appears in a file could be helpful.

Agreed that it's not obvious, but it has been used in a number of [tidyverse](https://github.com/search?q=%22%25%7C%7C%25%22+language%3AR+org%3Atidyverse&type=code) (as well as ropensci) packages.

modules/uncertainty/R/ensemble.R
```diff
@@ -217,37 +217,34 @@ get.ensemble.samples <- function( ensemble.size, pft.samples, env.samples,
 ##' @return list, containing $runs = data frame of runids, $ensemble.id = the ensemble ID for these runs and $samples with ids and samples used for each tag.  Also writes sensitivity analysis configuration files as a side effect
 ##' @details The restart functionality is developed using model specific functions by calling write_restart.modelname function. First, you need to make sure that this function is already exist for your desired model.See here \url{https://pecanproject.github.io/pecan-documentation/latest/pecan-models.html}
 ##' new state is a dataframe with a different column for each state variable. The number of the rows in this dataframe needs to be the same as the ensemble size.
-##' State variables that you can use for setting up the intial conditions differs for different models. You may check the documentation of the write_restart.modelname your model.
+##' State variables that you can use for setting up the initial conditions differs for different models. You may check the documentation of the write_restart.modelname your model.
```
dlebauer:
```suggestion
##' The state variables that you can use for setting up initial conditions are model specific. Check the documentation of the write_restart.<modelname> function for the model you are using.
```

dlebauer (changes requested):
It seems prudent to merge this hack to keep work moving. But before merging, please file one or more follow-up issues to track the open threads that it leaves.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ section for the next release.
 
 ### Changed
 
+- Ensemble and sensitivity analyses run with now assign an ensemble ID if one is not specified in the XML, even when running with no DB (#3654).
```
```suggestion
- Ensemble and sensitivity analyses now assign an ensemble ID if one is not specified in the XML, even when running with no DB (#3654).
```

modules/uncertainty/R/run.sensitivity.analysis.R
infotroph: [804c932](https://github.com/PecanProject/pecan/pull/3654/commits/804c93228f1ecc813b77deaa56ddb772c6a810ad)

infotroph:
> please file one or more follow-up issues to track the open threads that it leaves

@dlebauer this PR fixes the problem I was having, so as far as I'm concerned it's not leaving threads open 🤷 

Like I said above, the dump of my working notes is specifically for you to take and run with, because the correct resolution for each "unresolved" behavior (or even the decision whether it's a bug at all) could depend greatly on which direction you decide to take the SA work.

divine7022:
Those fixes are addressed and will be pushed once this PR gets merged

modules/uncertainty/R/ensemble.R
mdietze: I agree it seems like a handy function, but I think how it works if far from common knowledge, so it doesn't hurt to improve the documentation, both of what is occurring here and (more importantly) why (i.e., why do we set unspecified workflow IDs to -1 instead of leaving them NULL)

infotroph: > why do we set unspecified workflow IDs to -1

This is a good question. Looks like @dlebauer added this line in 2013 or earlier -- David, will you suggest the right explanatory comment and/or change to the default?

dlebauer: I don't recall the rationale other than to represent a database id value as an integer that will not occur in the database. 

But I think for consistency with below it would make sense to use the hash
  
```suggestion
  # if workflow$id is null, set to -1
  # to avoid collision w/ database ids
  workflow.id <- settings$workflow$id %||% -1
```

infotroph: Maybe? The issue I see is that using hash will set different workflow IDs for every site in the workflow. For ensemble ID I see a reasonable interpretation of that ("We're going to analyze all runs from this site as one ensemble, ignoring any from other sites"), but I'm not thinking of an equivalent justification for workflow ID -- I lean toward thinking it's better to leave it clearly undefined at this per-site level if nobody set it at the whole-workflow level. Especially since it looks like printing workflow.id into README.txt is the only thing DB-free runs of write.ensemble.configs do with it.

dlebauer (approved):
LGTM

PR #3655, opened 2025-10-20
[docker] Add ARM support where available from upstream

infotroph:
## Description

Allows declaration of Docker platform support to vary by R version as well as by image, and uses it to add ARM support when R version is 4.5.

## Motivation and Context

The Rocker project is now providing an arm64 build of the tidyverse image, thus far for R 4.5 only. I'd love to use it on my machine!

One relevant question is how much complexity is needed here and how long we'll need to support a mix of platforms instead of assuming dual amd64/arm64 everywhere. I think the answer is "a while" -- Rocker's general approach is to _not_ backport changes like this to older images, so I expect the R 4.2 through 4.4 images to stay AMD-only, and it's not clear if/when they plan to add ARM support to the R-devel image.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


.github/workflows/docker.yml
```diff
@@ -62,10 +62,20 @@ jobs:
       - id: default
         if: github.event_name != 'schedule'
         run: echo "R_VERSION=${{ github.event.inputs.r_version || env.DEFAULT_R_VERSION }}" >> "$GITHUB_OUTPUT"
+      - id: platform
+        # upstream rocker/tidyverse image is only on ARM for R 4.5.* (so far)
+        shell: bash
+        run: |
+          if [[ "${{ join(steps.*.outputs.R_VERSION, '') }}" == "4.5" ]]; then
```
robkooper, line 69: more complex, but do we want to use explicit if 4.1 the x86, etc, then else arm. That way develop also compiles with arm. Once we disable 4.2 and add 4.6 it does not need any updates, except removal of 4.2

robkooper (approved):
looks good, left a comment, feel free to take it.

Not tested yet since we need images build :(

.github/workflows/docker.yml
infotroph: I implemented it that way at first, then realized `rocker/tidyverse:devel` doesn't' provide an arm version yet. I'll merge this and plan to add the complexity back whenever the list of x86-only versions gets shorter than the list of versions with arm.

PR #3656, opened 2025-10-21
Simpler str_ns handling

infotroph:
## Description

This is a condensed implementation of the core ideas from #3324: 
- If site settings already contain lat and lon, use those and do not attempt to look site location up from the db
- For met.process only, allow site id to be missing. If so, create an ID by pasting lat & lon together
- Always treat site id as character not numeric, but continue condensing Bety IDs (`1000000772` => `1-772`)  as before

@Sweetdevil144 please review -- if you agree this is a sufficient replacement, merging this will close #3324.

## Motivation and Context

We learned a lot about DB-free runs since starting work on #3324. It's now clear that for most cases we want to require the user to provide site IDs rather than generate them ourselves, meaning much of the complexity proposed in the existing  PR is unneeded. Rather than try to unwind all the changes there, I've cherry-picked out the core commits here and updated the format of the latlon-based IDs for reduced ambiguity.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/met.process.R
+++ b/modules/data.atmosphere/R/met.process.R
@@ -142,14 +142,14 @@ met.process <- function(site, input_met, start_date, end_date, model,
   
   
   # setup site database number, lat, lon and name and copy for format.vars if new input
-  if(is.null(site$lat) | is.null(site.lon)) {
+  if(is.null(site$lat) | is.null(site$lon)) {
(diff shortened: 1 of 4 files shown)


Sweetdevil144 (approved):
LGTM

PR #3660, opened 2025-11-10
Multisite SA support with input coordination and README tracking

divine7022:
## Description
* enable SA + ensemble size > 1 through input coordination
  workaround: add `input_design` parameter to `write.sa.config` to coordinate multiple inputs (met, IC, soil, etc.) across SA runs.
* enhanced README.txt files to track all inputs used per run
* fix overwriting issue of `sample.Rdata` in case of multisite setting; 
  workaround : implemented logic to append each site's data to existing `samples.Rdata` instead of overwriting

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/workflow/R/run.write.configs.R
```diff
@@ -51,7 +51,7 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
       }
     )
   }
-
+
```
infotroph: Please don't add end-of-line whitespace

modules/uncertainty/R/sensitivity.R
```diff
@@ -176,15 +177,15 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
       "INSERT INTO runs ",
       "(model_id, site_id, start_time, finish_time, outdir, ensemble_id, parameter_list) ",
       "values ('",
-        settings$model$id, "', '",
-        settings$run$site$id, "', '",
-        settings$run$start.date, "', '",
-        settings$run$end.date, "', '",
-        settings$run$outdir, "', ",
-        ensemble.id, ", '",
-        paramlist, "') ",
```
infotroph, line 176: I found this indentation really helpful for understanding the nesting of the query it's writing and would rather not remove it. This may be a place automated linter recommendations go wrong.

modules/uncertainty/R/sensitivity.R
```diff
@@ -214,15 +215,57 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
...
-
+  
+  # I check to make sure the path under the met is a list.
+  # if it's specified what met needs to be used in 'met.id' under sensitivity
+  # analysis of pecan xml we used that otherwise, I use the first met.
+  if (is.list(settings$run$inputs$met$path)) {
+    # This checks for met.id tag in the settings under sensitivity analysis -
+    # if it's not there it creates it. Then it's gonna use what it created.
+    if (is.null(settings$sensitivity.analysis$met.id)) {
+      settings$sensitivity.analysis$met.id <- 1
```
infotroph, line 228: 👍 to the overall goal of picking one met (/other input) path to use across the whole SA. 

But calling this "met.id" is confusing to me -- looks like this is basically an index into the ensemble, whereas I would usually expect "met ID" to mean an identifier for which _source_ (e.g. CRUNCEP ERA5, etc) the met data came from. 

Could it instead be something like `settings$sensitivity.analysis$met_path` and contain the actual filepath to be used by the SA instead of an index that needs to be extracted from the ensemble? That'd be clearer to me, and would also allow a user to pass an arbitrary path to the SA even if it doesn't exist in the ensemble (not that I expect most people to want that, but it could come in handy sometimes)

base/workflow/R/run.write.configs.R
```diff
@@ -195,29 +196,113 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+  if (file.exists(samples.file)) {
+    existing_env <- new.env()
+    load(samples.file, envir = existing_env)
+    if (!is.null(existing_env$ensemble.samples)) {
+      for (pft_name in names(ensemble.samples)) {
+        if (pft_name %in% names(existing_env$ensemble.samples)) {
+          # combine parameter samples for same PFT across sites
+          ensemble.samples[[pft_name]] <- rbind(
+            existing_env$ensemble.samples[[pft_name]],
+            ensemble.samples[[pft_name]]
```
infotroph, line 237: I'm not seeing how one figures out after the fact which samples were used for which site

modules/uncertainty/R/sensitivity.R
```diff
@@ -343,46 +373,77 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
...
             unlink(file.path(settings$rundir, run.id))
             unlink(file.path(settings$modeloutdir, run.id))
           }
           dir.create(file.path(settings$rundir, run.id), recursive = TRUE)
           dir.create(file.path(settings$modeloutdir, run.id), recursive = TRUE)
-
-          # write run information to disk
+          
+          # Apply input design coordination for SA runs
+          settings_copy <- settings
```
infotroph, line 388: It's confusing that below here some values are taken from `settings` and others from `settings_copy`. Is there actually a need for two separate objects or could these mutations be applied straight to `settings`?

modules/uncertainty/R/sensitivity.R
```diff
@@ -236,55 +279,42 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
...
   )
-
+  
+  run_index <- 1
+  
   ## loop over pfts
   runs <- list()
-  for (i in seq_along(names(quantile.samples))) {
-    pftname <- names(quantile.samples)[i]
+  for (pft_idx in seq_along(names(quantile.samples))) {
```
infotroph, line 301: Nice clarification!

base/workflow/R/run.write.configs.R
```diff
@@ -195,29 +196,113 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
-
+  
   PEcAn.logger::logger.info("###### Finished writing model run config files #####")
   PEcAn.logger::logger.info("config files samples in ", file.path(settings$outdir, "run"))
+  
+  samples.file <- file.path(settings$outdir, "samples.Rdata")
+  
+  if (file.exists(samples.file)) {
+    existing_env <- new.env()
+    load(samples.file, envir = existing_env)
```
infotroph: This was already loaded above (line 109) -- why the reload?

modules/uncertainty/R/sensitivity.R
infotroph: ...But wait! Doesn't `input_design` handle this for us just below?

base/workflow/R/run.write.configs.R
```diff
@@ -195,29 +196,113 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+      }
+    }
+    
+    if (!is.null(existing_env$trait.samples)) {
+      trait.samples <- modifyList(existing_env$trait.samples, trait.samples)
+    }
+    
+    if (!is.null(existing_env$sa.samples)) {
+      sa.samples <- modifyList(existing_env$sa.samples, sa.samples)
+    }
```
infotroph, line 256: If I'm following right, none of these are changed from when they were read in from samples.Rdata. Is this going to wind up just repeating the same values n times?

infotroph (commented):
Thanks for working on this! Big picture, I don't follow how appending to samples.Rdata will work for the processes that need to read from it downstream and I wonder how much of the information appended is already captured by the input design. I left some more specific comments inline, but it's worth being sure the overall design is right before worrying about most of them.

base/workflow/R/run.write.configs.R
divine7022: 👍

modules/uncertainty/R/sensitivity.R
divine7022: 👍

modules/uncertainty/R/sensitivity.R
divine7022: good question, it duplicates with `input_design`( new approach )
* and also one of the reason is to be consistent with the ensemble pattern, so SA doing different (path) would be inconsistent
* and works when `input_design` is NOT provided (single-site or legacy workflows)
* The met.id resolution is indeed redundant when input_design is present. The met.id logic serves as a backward compatibility fallback for workflows that don't use `input_design`. However, when `input_design` exists it **supersedes** met.id and handles ALL inputs (met, IC, soil, etc.) . 
* Alternatively we could skip met.id resolution entirely when `input_design` exists


short-term fix (This PR):
keep current approach, add clarifying comment:
```
# select single met path for SA runs without input_design (backward compatibility)
# when input_design is provided (recommended), it supersedes this selection 
# and allows different met files per SA run 
if (is.list(settings$run$inputs$met$path)) { 
if (is.null(settings$sensitivity.analysis$met.id)) { 
settings$sensitivity.analysis$met.id <- 1 
} 
settings$run$inputs$met$path <- settings$run$inputs$met$path[[settings$sensitivity.analysis$met.id]] 
}
```
Long-term (Further iteration): 
Document `input_design` as recommended approach , consider deprecating met.id in favor of `input_design`.

base/workflow/R/run.write.configs.R
divine7022: The site information is preserved through the run id's; Each run id contains the site id
(e.g, `ENS-00001-1000004925` --- > site 1000004925), and this becomes the output folder name. Downstream functions don't parse the site id, they just iterate through all run id's and read from folders named with those id's.

```
# site 1 : 
runs.samples$ensemble <- data.frame( 
id = c("ENS-00001-1000004925", # site id
"ENS-00002-1000004925") 
) 
ensemble.samples$temperate <- data.frame( 
growth_resp = c(0.28, 0.31) # rows 1-2 
) 

# after site 2 merging: 
runs.samples$ensemble <- data.frame( 
id = c("ENS-00001-1000004925", # site 1, run 1 
"ENS-00002-1000004925", # site 1, run 2 
"ENS-00001-1000004928", # site 2, run 1 <-- Different site id
"ENS-00002-1000004928") # site 2, run 2 
) 
ensemble.samples$temperate <- data.frame( 
growth_resp = c(0.28, 0.31, 0.33, 0.30) # rows match run id's 
)
```

modules/uncertainty/R/sensitivity.R
divine7022: SA loop without settings_copy:
```
for (trait in traits) {
  for (quantile in quantiles) {
    settings$run$inputs$met$path <- input_paths[[run_index]]  # mutates
    ...
    # write config
    write.config.SIPNET(settings, ...)
    # next iteration uses MUTATED settings; 
    # quantile 2 sees met path from quantile 1
  }
}

```
* so by having `settings_copy` it modify copy only (`settings_copy$run$inputs$met$path <- input_paths[[run_index]]`) and `write.config.* ` gets _clean_ settings each iteration.
* If we mutated `settings` directly, run 2's config would inherit run 1's modified input paths, causing incorrect configs.

The answer for question: why `settings` is used for some values ?
`cat("model id : ", settings$model$id, "\n")` # never changes across runs
`cat("site id : ", settings$run$site$id, "\n")` # same for all SA runs

BUT inputs must come from `settings_copy`:
`do.call(my.write.config, settings = settings_copy)` # has resolved input paths

modules/uncertainty/R/sensitivity.R
divine7022: 👍

base/workflow/R/run.write.configs.R
divine7022: good catch! i added the merge without consolidating the loads
now i have consolidated to a single load

base/workflow/R/run.write.configs.R
divine7022: No, these are NOT repeating values - there's **new data generated between the loads**

lines 106-110: load existing `samples.Rdata`
- contains site 1's data from previous run
- `trait.samples` site 1's MCMC samples
- `ensemble.samples` site 1's ens data

lines 111-121: generate NEW `ensemble.samples` for current site
- uses current site's `trait.samples` (loaded above)
- indexes specific samples via `input_design[["param"]]`
- creates _fresh_ `ensemble.samples` for current site
- ex: if site 2, generates site 2's NEW ensemble data

lines 154-171: generate NEW run id's for current site
- `write.sa.configs()` creates new SA run id's
- `write.ensemble.configs()` creates new ensemble run id's  
- updates `runs.samples$sa` and `runs.samples$ensemble` with current site's runs

lines 220-246: merge existing (all previous sites) + new (current site)

Big picture concerns
> how appending to samples.Rdata works for downstream?

The good news is that downstream functions are actually **designed** to handle accumulated multisite data, downstream analysis functions load samples.Rdata and loop through **ALL** runs in `runs.samples$ensemble` and `runs.samples$sa`
> if your merging duplicates what input_design already provides ?

 NO - input_design controls input selection during config generation (for selecting which inputs to use during config generation, which contains indices only for current site and used during `write.*.configs` )
while samples.Rdata stores the results (run id's + parameter values) needed for analysis
> how much of the information appended is already captured by the input design ?

input_design and samples.Rdata serve DIFFERENT purposes
```
input_design <- data.frame(
param = c(1, 2, 3, 4), # which parameter sample to use
met = c(1, 3, 2, 1), # which met file to use
poolinitcond = c(1, 1, 2, 2) # which ic file to use
)
```
and sample.Rdata , which is archive of actual run id's and parameter values for analysis

modules/uncertainty/R/sensitivity.R
infotroph: Re met path: You've talked me out of the idea, especially since it would need to be site-specific -- so my proposal wouldn't need just one `settings$sensitivity.analysis$met_path`, it'd be at least a `settings$sensitivity.analysis$<site>$met_path` for every site. I do still suggest naming it something like `met.index` instead of `met.id`.

infotroph: Re input design: I'll defer to @dlebauer but my sense is that since we've bought into making `input_design` mandatory for other parts of the workflow and it's specifically designed to be where we keep all the pieces of the sampling design together in one place, it may be sensible to make it mandatory here too rather than duplicate its functionality.

divine7022: > I do still suggest naming it something like `met.index` instead of `met.id`

that's renaming to `met.index` would be nice clarity, but i am afraid of breaking backward compatibility for existing workflows may have `<met.id>` , and a direct rename would silently ignore those settings ?

dlebauer: I'm in favor of having a consistent `input_design` object. It is hard to see an advantage of different approaches if one generalize solution is available.

base/workflow/R/run.write.configs.R
infotroph: In that case I don't think ensemble.samples should grow at all -- The point of passing in an input design is to enforce that every run with a given ensemble member uses the same value of `growth_resp` (and every other parameter too), so ensemble.samples needs to stay n_ens lines long. I guess one _could_ repeat the same set of values for each site, but that seems VERY inefficient.

base/workflow/R/run.write.configs.R
```diff
@@ -211,6 +215,80 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+      # preserves existing PFTs -- add PFTs that exist in file but not current site
+      for (existing_pft in names(existing_data$ensemble.samples)) {
+        if (!existing_pft %in% names(ensemble.samples)) {
+          ensemble.samples[[existing_pft]] <- existing_data$ensemble.samples[[existing_pft]]
+        }
+      }
+    }
+    # Merge trait.samples
+    if (!is.null(existing_data$trait.samples)) {
+      trait.samples <- utils::modifyList(existing_data$trait.samples, trait.samples)
```
infotroph, line 251: Trait.samples ought to be generated exactly once at the beginning of setup (to define the set of posteriors that `ensemble.samples` is selected from) and not differ from site to site. Since it may contain MCMC results with potentially as many as millions of samples per trait, I _really_ don't think we want to be duplicating it here.

modules/uncertainty/R/sensitivity.R
infotroph: Right, but the path is updated earlier in the loop than the write.config call, so by the time write.config gets called for run 2 the path from run 1 has already been overwritten by the path from run 2. As long as you're sure that every iteration will update the same fields of the settings object, it doesn't matter whether it started from a "clean" object or not.

Similarly, if you _are_ using settings_copy then it also contains all the fields from `settings` that _aren't_ updated each iteration, so there's definitely no reason to interleave uses of both.

modules/uncertainty/R/sensitivity.R
dlebauer: I am curious - where is met.id used? I can't find the string in either the PEcAn repository or in xml settings files used by CCMMF or Dongchen's workflows.

mdietze: @dlebauer it's only used immediately below to set the met$path.

I agree with the consensus to drop the met.id solution proposed here and just use input_design. If one wants to only vary parameters and hold everything else fixed it's really easy to generate an input_design that's all 1's for everything other than parameters. The met.id "solution" isn't really a solution anyways as it only fixes met ensemble members and not the other ensembled inputs (phenology, soil physics, initial conditions, etc.)

modules/uncertainty/R/sensitivity.R
```diff
@@ -343,46 +373,77 @@ write.sa.configs <- function(defaults, quantile.samples, settings, model,
...
               "site        : ", settings$run$site$name, "\n",
               "site  id    : ", settings$run$site$id, "\n",
-              "met data    : ", settings$run$site$met, "\n",
+              sa_input_info,
               "start date  : ", settings$run$start.date, "\n",
               "end date    : ", settings$run$end.date, "\n",
               "hostname    : ", settings$host$name, "\n",
               "rundir      : ", file.path(settings$host$rundir, run.id), "\n",
               "outdir      : ", file.path(settings$host$outdir, run.id), "\n",
               file = file.path(settings$rundir, run.id, "README.txt"),
```
dlebauer, line 432: This should address the suggestion in https://github.com/PecanProject/pecan/pull/3660#discussion_r2555267550 to avoid mixing use of settings and settings_copy once the copy is made.  
```suggestion
              "workflow id : ", workflow.id, "\n",
              "ensemble id : ", ensemble.id, "\n",
              "pft name    : ", names(trait.samples)[pft_idx], "\n",
              "quantile    : ", quantile.str, "\n",
              "trait       : ", trait, "\n",
              "run id      : ", run.id, "\n",
              "model       : ", model, "\n",
              "model id    : ", settings_copy$model$id, "\n",
              "site        : ", settings_copy$run$site$name, "\n",
              "site  id    : ", settings_copy$run$site$id, "\n",
              sa_input_info,
              "start date  : ", settings_copy$run$start.date, "\n",
              "end date    : ", settings_copy$run$end.date, "\n",
              "hostname    : ", settings_copy$host$name, "\n",
              "rundir      : ", file.path(settings_copy$host$rundir, run.id), "\n",
              "outdir      : ", file.path(settings_copy$host$outdir, run.id), "\n",
              file = file.path(settings_copy$rundir, run.id, "README.txt"),
```

modules/uncertainty/R/sensitivity.R
dlebauer: I agree that it makes sense to use settings_copy instead of settings. It seems this only concerns the README block that pulls fields from the original settings. If we just switch those `cat()` calls to read from `settings_copy`, it’ll be obvious that each iteration uses one clean settings object end-to-end.

I've proposed this change below in https://github.com/PecanProject/pecan/pull/3660/files/f075e4d7d0f9ad85ee340bbe01f7dca464988677#r2561591261

modules/uncertainty/R/sensitivity.R
dlebauer, line 443:
```suggestion
            file = file.path(settings_copy$rundir, "runs.txt"),
```

base/workflow/R/run.write.configs.R
divine7022: @infotroph @mdietze @dlebauer want to make sure that i got right logic with the merging approach ?

Explained in comments

```r
  PEcAn.logger::logger.info("###### Finished writing model run config files #####")
  PEcAn.logger::logger.info("config files samples in ", file.path(settings$outdir, "run"))
  
  # -----------------------------------------------------------------------
  # multisite merging logic
  # -----------------------------------------------------------------------
  # In multisite ensemble analysis, we maintain parameter
  # consistency across sites while allowing run tracking to grow. 
  # This ensures that the same parameter samples are applied across all sites,
  # enabling valid comparison of model performance under different environmental
  # conditions while tracking site-specific run execution.
  # -----------------------------------------------------------------------
  
  samples.file <- file.path(settings$outdir, "samples.Rdata")
  
  if (file.exists(samples.file)) {
    PEcAn.logger::logger.info("Merging with existing samples.Rdata to support multisite run...")
    
    # load existing data
    existing_env <- new.env()
    load(samples.file, envir = existing_env)
    
    # Ensemble parameters must remain identical across sites to answer:
    # How do the SAME parameter sets perform under DIFFERENT conditions ?
    
    if (!is.null(existing_env$ensemble.samples)) {
      # does the existing list have names ?
      # this validates we have a properly structured ensemble.samples object
      if (length(names(existing_env$ensemble.samples)) > 0) {
        # use existing parameter set to maintain ensemble consistency
        # all sites use identical parameter samples for ensemble analysis
        ensemble.samples <- existing_env$ensemble.samples
      } else {
        # existing parameter structure is invalid - use current site's parameters
        PEcAn.logger::logger.warn("Existing ensemble.samples structure is invalid. Using current site parameters as new baseline.")
      }
    }
    
    # maintain consistency in trait and sensitivity analysis samples
    if (!is.null(existing_env$trait.samples)) {
      if (length(names(existing_env$trait.samples)) > 0) {
        trait.samples <- existing_env$trait.samples
      }
    }
    if (!is.null(existing_env$sa.samples)) {
      if (length(names(existing_env$sa.samples)) > 0) {
        sa.samples <- existing_env$sa.samples
      }
    }
    
    # Run IDs append (this SHOULD grow)
    # ---------------------------------------------------------------------
    # Run IDs accumulate to track which parameter sets execute at which sites
    # eg. with 2 ens members across 2 sites:
    # - site 1: ENS-00001-1000004925, ENS-00002-1000004925
    # - site 2: ENS-00001-1000004928, ENS-00002-1000004928  
    # - final: 4 run IDs tracking same 2 parameter sets across 2 sites
    # ---------------------------------------------------------------------
    
    if (!is.null(existing_env$runs.samples)) {
      
      # ensemble runs tracking (data frame -> rbind)
      # rbind here because run IDs must accumulate across sites
      if (!is.null(runs.samples$ensemble)) {
        if (!is.null(existing_env$runs.samples$ensemble)) {
          # check we are not rbinding garbage. Check for 'id' column.
          if("id" %in% names(existing_env$runs.samples$ensemble)) {
            runs.samples$ensemble <- rbind(existing_env$runs.samples$ensemble, runs.samples$ensemble)
            # run IDs grow to track which sites use which ensemble members
          }
        }
      } else if (!is.null(existing_env$runs.samples$ensemble)) {
        runs.samples$ensemble <- existing_env$runs.samples$ensemble
      }
      
      # SA runs tracking (list structure -> modifyList)
      if (!is.null(runs.samples$sa)) {
        if (!is.null(existing_env$runs.samples$sa)) {
          runs.samples$sa <- utils::modifyList(existing_env$runs.samples$sa, runs.samples$sa)
        }
      } else if (!is.null(existing_env$runs.samples$sa)) {
        runs.samples$sa <- existing_env$runs.samples$sa
      }
    }
    
    # metadata (pft names, trait names) should merge to create 
    # coverage across all sites while avoiding duplication
    # ---------------------------------------------------------------------
    
    if (!is.null(existing_env$pft.names)) {
      pft.names <- unique(c(existing_env$pft.names, pft.names))
    }
    if (!is.null(existing_env$trait.names)) {
      if (length(names(existing_env$trait.names)) > 0) {
        trait.names <- utils::modifyList(existing_env$trait.names, trait.names)
      }
    }
  }
  
  save(ensemble.samples, trait.samples, sa.samples, runs.samples, pft.names, trait.names,
       file = samples.file
  )
  
  PEcAn.logger::logger.info("parameter values for runs in ", samples.file)
  options(scipen = scipen)
  invisible(settings)
  return(settings)
}
```

mdietze: This seems overly complicated. get.param.samples should generate one matrix of parameters per PFT and currently there is not within-PFT variability in parameter samples across sites (though this may change in the future). For a particular site, the PFT and input_design should determine which parameter matrix you use, and which row in that matrix you select, respectively. There should be no need to rbind matrices.

infotroph: I agree with Mike here and am leaning toward an even stronger claim that run.write.configs _should not be rewriting samples.Rdata at all_. Its contents are all either generated once by get.parameter.samples before run.write.configs starts (ensemble.samples, sa.samples, trait.samples) or are not actually samples and can be taken from runs.txt (runs.samples)

divine7022: > not actually samples and can be taken from runs.txt (runs.samples)

does the downstream analysis expect runs.samples to be in samples.Rdata ?

divine7022: I looked into read.ensemble.output and read.sa.output to verify if we can stop writing samples.Rdata, but those functions explicitly load samples.Rdata to find the run ids : 
```ens.run.ids <- samples$runs.samples$ensemble```
```sa.run.ids <- samples$runs.samples$sa```

they do not currently have logic to parse runs.txt

so here we can ---
1 ) reload the existing parameters (avoid duplication/growth) but I must append the run ids to runs.samples and save the file.  (`run.write.cofigs`)
2 ) or refactor downstream to read runs.txt   ( does this feel like to be a separate PR  ? )

mdietze: Don't know what you mean by "stop writing samples.Rdata". No one is saying this isn't needed, we're saying it should already exist before run.write.configs is called (it is now generated by the new input design function, which should be called before the write configs functions) and it shouldn't need to be modified. 

Only start model runs should rely on runs.txt. There's not enough information in there for anything else downstream

divine7022: > Don't know what you mean by "stop writing samples.Rdata".

I was responding specifically to chris's suggestion that runs.samples might be redundant cuz that info 
> "can be taken from runs.txt". 

I was verifying if `run.write.configs` could stop appending even the new run ids(runs.samples) to samples.Rdata ( relaying on runs.txt instead) , but downstream (read.ensemble.output , .. ) depends on finding those ids in samples.Rdata.

But it's clear now 
> Only start model runs should rely on runs.txt.

divine7022:
extended in -- https://github.com/PecanProject/pecan/pull/3708

Issue #3662, opened 2025-11-14
Avoid overloading `pft$outdir` as an input path


infotroph:
### Bug Description

Context: `get.parameter.samples` expects two paths for each PFT: 
* `pfts$<pft>$posterior.files` is a path including filename (which is typically `post.distns.Rdata` inside a PFT-specific directory) for an Rdata file containing the parameters characterized by named distributions
* `pfts$<pft>$outdir` is a path to a directory to look in for MCMC results (with the filename hard-coded as `trait.mcmc.Rdata`) from the parameters whose distribution is not easily named.  

Bug: Both of these paths are _inputs_ to get.parameter.samples, so needing to specify the MCMC location in `outdir` is confusing  and (best I can tell) undocumented. It could make sense in the context of a workflow that always runs meta-analysis and therefore always recreates these files, but given my posteriors are generated in a separate offline calibration process I want to store them in a path that is separate from the easily-recreated workflow outputs.  Since `pfts$<pft>$outdir` is also used by `run.sensitivity.analysis` (and possibly others) to write their own outputs, as far as I can tell there's currently no way to pass MCMC files from a readonly location -- my [gollum voice] preciousssss input artifacts (or copies/links to them) have to be mixed in with outputs.

### Possible fixes

My preference would be to have separate settings for each input and output, and to specify them in the block for each step rather than together in the pft block. Where a file is updated by multiple steps it may make more sense to have a setting that identifies it by purpose in a way that makes sense both for steps that take it in and put it out: 

* `get.parameter.samples` reads the MCMC file from `pfts$<pft>$posterior.mcmc.file` and the distribution file from `pfts$<pft>$posterior.distribution.file` (or possibly from `pfts$<pft>$posterior.files$mcmc` and `pfts$<pft>$posterior.files$distribution` respectively)
* Meta-analysis takes its output directory from a setting in the `meta.analysis` block rather than the `pfts` block, possibly _updating_ the `pfts` block to point to the relevant files after writing them
* sensitivity analysis takes its output directory from a setting in the `sensitivity.analysis` block, separating PFTs as subdirs within the specified path. It does not update the `pfts` block because none of its outputs change how downstream steps use a given PFT (right?)

mdietze:
[...]

PR #3663, opened 2025-11-17
Fix(uncertainty): allow explicit MCMC and distribution paths; avoid o…

shbhmexe:
Title: fix(uncertainty): allow explicit MCMC and distribution paths; avoid overloading pft$outdir as an input

What was fixed or enhanced:
- `get.parameter.samples()` implicitly treated `pft$outdir` as an input source for both `post.distns.Rdata` and `trait.mcmc.Rdata`. This forced read-only calibration artifacts to be placed/mixed in output directories. The function now supports explicit, per‑PFT input paths and only falls back to `pft$outdir` when nothing is provided.

Why the change was necessary:
- To separate immutable input artifacts (posterior distributions and MCMC samples) from workflow outputs and to align with the intent that `pft$outdir` is primarily for outputs. This reduces confusion and accidental overwrites and enables using externally curated calibration results without copying them into PEcAn’s output tree.

What exact changes were made:
- Updated `modules/uncertainty/R/get.parameter.samples.R` to:
  - Accept `posterior.files` as either a vector of distribution files (backward compatible) or a list of lists per PFT with named entries `distribution` and/or `mcmc`.
  - If `posterior.files[[i]]$mcmc` is provided, load that MCMC directly and skip any `pft$outdir`/DB fallback; preserve PDA correlation handling.
  - If `posterior.files[[i]]$distribution` is provided, load it directly; otherwise, fall back to the most recent `post.distns.Rdata` in `pft$outdir`, then `prior.distns.Rdata`.
  - Added roxygen comments documenting the new accepted structure while keeping the old behavior working.

How it improves the project:
- Decouples input data locations from output directories, reducing accidental mixing of inputs/outputs and enabling cleaner workflows, especially when posteriors are generated offline.

Verification steps:
1. Provide explicit paths for a single PFT:
```r
settings <- PEcAn.settings::read.settings('pecan.xml')
posterior.files <- list(list(
  distribution = '/path/to/post.distns.Rdata',
  mcmc         = '/path/to/trait.mcmc.Rdata'
))
PEcAn.uncertainty::get.parameter.samples(settings, ensemble.size = 3, posterior.files = posterior.files)
```
2. Confirm that `samples.Rdata` is written under `settings$outdir` and that no files are read from `pft$outdir` when explicit paths are provided.
3. Backward compatibility: call with `posterior.files = NA` and confirm previous fallback behavior still works.

Next steps (optional, not included in this PR):
- Consider moving MA/SA output directories to `settings$meta.analysis` and `settings$sensitivity.analysis` blocks respectively, to fully eliminate the input/output overloading risk.

### Modified Files
- `modules/uncertainty/R/get.parameter.samples.R` ✅ push

No config or environment files were added.


infotroph:
See #3662 for problem description and design discussion that has continued after this PR was started

PR #3671, opened 2025-11-23
Correct clean rule in Makefile

AritraDey-Dev:
## Description

After running `make build` or `make deploy` in the `book_source` directory, two directories—`_book` and `extfiles`—are generated as output. However, the `clean` target in the `Makefile` incorrectly specifies `../book`, which does not remove the generated directories.

This PR fixes the `clean` target by correcting the paths so that the appropriate output directories are deleted.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


infotroph:
Thanks! While we're looking at it, can you please also delete the `deploy` rule and its accompanying `deploy.sh`? Those are unused now that deployment is now handled by `.github/book.yml`.

AritraDey-Dev:
Could we also add a `serve` rule to the Makefile to serve the built site?

infotroph:
> Could we also add a serve rule to the Makefile to serve the built site?

I don't immediately see the need, but it would definitely be a separate PR so I'll merge this now and we can follow up in a fresh thread.

infotroph (approved):

PR #3672, opened 2025-11-23
test(db): fix fancy_scientific test by comparing deparsed expressions

AritraDey-Dev:
## Description

Fix `fancy_scientific() ` Test Failures in `test.query.dplyr.R`
```
Failure (test.query.dplyr.R:9:3): `fancy_scientific()` converts numbers to scientific expressions with proper formatting
`result` not equal to expression("1" %*% 10^-20).
Attributes: < Modes: list, NULL >
Attributes: < Lengths: 3, 0 >
Attributes: < names for target but not for current >
Attributes: < current is not list-like >
target is not list-like
```
The tests were failing because `expect_equal()` checks every attribute of the parsed expressions, and those attributes can differ even when the expressions themselves mean the same thing.
To avoid these false failures, the tests now compare the `deparsed` string versions of the expressions using `deparse()`, instead of comparing the full expression objects.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/db/tests/testthat/test.query.dplyr.R
+++ b/base/db/tests/testthat/test.query.dplyr.R
@@ -1,12 +1,15 @@
 test_that("`fancy_scientific()` converts numbers to scientific expressions with proper formatting", {
   result <- fancy_scientific(1234567890)
-  expect_equal(result, expression("1.234568" %*% 10^+9))
+  expected <- expression("1.234568" %*% 10^+9)
+  expect_equal(deparse(result[[1]]), deparse(expected[[1]]))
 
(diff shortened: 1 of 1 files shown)


infotroph (approved):
Thanks. This fix looks reasonable, but FTR I haven't been able to reproduce the failure on my machines. What OS and R version are you seeing it on?

AritraDey-Dev:
I'm on Ubuntu (24.04) with` R 4.5.2`. The failure may be specific to newer R versions.

PR #3673, opened 2025-11-23
R bugfixes across dbfiles, remote helpers, and met alignment

shbhmexe:
## What
Small, targeted fixes to prevent runtime errors and subtle misbehavior in DB utilities, remote job helpers, and meteorology alignment/debiasing routines. No new features, no API changes.

## Why
- dbfiles: wrong get.id argument order prevented lookups; file-type guard always evaluated TRUE.
- align_met: incorrect `rep(..., each=stamps.hr)` used a vector as a scalar, leading to length mismatch.
- debias_met_regression: ensemble-size checks and sampling counts were incorrect; could error or oversample.
- insert.format.vars: invalid indexing and collect() usage, wrong type check for `skip`, and unsafe unit checks could raise errors.

## Scope/Risk
- Changes are confined to the listed functions; default behavior is preserved except to correct the bugs.
- No interfaces or schemas changed; CI and existing callers should remain unaffected.

## Testing hints
- Run existing R tests for affected packages (PEcAn.DB, PEcAn.data.atmosphere, PEcAn.remote) if available.
- Smoke test `align.met()` with coarse source and finer training data.
- Call `insert.format.vars()` with a tibble for `formats_variables` and confirm validations and DB writes.
- Exercise `dbfile.move()` with supported types ("nc"/"clim").

### DIFF
--- a/base/db/R/insert.format.vars.R
+++ b/base/db/R/insert.format.vars.R
@@ -48,18 +48,20 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
   }
 
   # Test if format name already exists
-  name_test <- dplyr::tbl(con, "formats") %>% dplyr::select("id", "name") %>% dplyr::filter(.data$name %in% !!format_name) %>% dplyr::collect()
-  name_test_df <- as.data.frame(name_test)
-  if(!is.null(name_test_df[1,1])){
+  name_test <- dplyr::tbl(con, "formats") %>%
+    dplyr::select(.data$id, .data$name) %>%
+    dplyr::filter(.data$name %in% !!format_name) %>%
+    dplyr::collect()
+  if(nrow(name_test) > 0){
     PEcAn.logger::logger.error(
       "Name already exists"
     )
   }
 
-   #Test if skip is an integer
-  if(!is.character(skip)){
-  PEcAn.logger::logger.error(
-    "Skip must be of type character"
+   # Test if skip is numeric
+  if(!is.numeric(skip)){
+    PEcAn.logger::logger.error(
+      "Skip must be of type numeric"
     )
   }
 
@@ -71,7 +73,7 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
   }
 
   # Test if notes are a character string
-  if(!is.character(notes)&!is.null(notes)){
+  if(!is.character(notes) & !is.null(notes)){
     PEcAn.logger::logger.error(
       "Notes must be of type character"
     )
@@ -79,39 +81,43 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
 
   ######## Formats-Variables tests ###############
   if(!is.null(formats_variables)){
-    for(i in 1:nrow(formats_variables)){
-      if(!is.numeric(formats_variables[[i,"variable_id"]])){
+    for(i in seq_len(nrow(formats_variables))){
+      row_i <- formats_variables[i, ]
+      if(!is.numeric(row_i[["variable_id"]])){
         PEcAn.logger::logger.error(
           "variable_id must be an integer"
         )
       }
 
-      if(suppress == FALSE){
-        ## Test if variable_id already exists ##
-        var_id_test <- dplyr::tbl(con, "variables") %>% dplyr::select("id") %>% dplyr::filter(.data$id %in% !!formats_variables[[i, "variable_id"]]) %>% dplyr::collect(.data$id)
-        if(!is.null(var_id_test[1,1])){
+      if(identical(suppress, FALSE)){
+        ## Test if variable_id exists in variables table ##
+        var_id_test <- dplyr::tbl(con, "variables") %>%
+          dplyr::select(.data$id) %>%
+          dplyr::filter(.data$id %in% !!row_i[["variable_id"]]) %>%
+          dplyr::collect()
+        if(nrow(var_id_test) == 0){
           PEcAn.logger::logger.error(
-            "variable_id already exists"
+            "variable_id does not exist in 'variables' table"
           )
         }
       }
 
-      if(!is.character(formats_variables[[i, "name"]])&!is.na(formats_variables[[i, "name"]])){
+      if(!is.na(row_i[["name"]]) && !is.character(row_i[["name"]])){
         PEcAn.logger::logger.error(
           "Variable name must be of type character or NA"
         )
       }
-      if(!is.character(formats_variables[[i, "unit"]])&!is.na(formats_variables[[i, "unit"]])){
+      if(!is.na(row_i[["unit"]]) && !is.character(row_i[["unit"]])){
         PEcAn.logger::logger.error(
           "Units must be of type character or NA"
         )
       }
-      if(!is.character(formats_variables[[i, "storage_type"]])&!is.na(formats_variables[[i, "storage_type"]])){
+      if(!is.na(row_i[["storage_type"]]) && !is.character(row_i[["storage_type"]])){
         PEcAn.logger::logger.error(
           "storage_type must be of type character or NA"
         )
       }
-      if(!is.numeric(formats_variables[[i, "column_number"]])&!is.na(formats_variables[[i, "column_number"]])){
+      if(!is.na(row_i[["column_number"]]) && !is.numeric(row_i[["column_number"]])){
         PEcAn.logger::logger.error(
           "column_number must be of type numeric or NA"
         )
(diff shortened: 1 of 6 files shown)


base/db/R/insert.format.vars.R
```diff
@@ -48,18 +48,20 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
   }
 
   # Test if format name already exists
-  name_test <- dplyr::tbl(con, "formats") %>% dplyr::select("id", "name") %>% dplyr::filter(.data$name %in% !!format_name) %>% dplyr::collect()
-  name_test_df <- as.data.frame(name_test)
-  if(!is.null(name_test_df[1,1])){
+  name_test <- dplyr::tbl(con, "formats") %>%
+    dplyr::select(.data$id, .data$name) %>%
```
infotroph, line 52: Use of .data inside select() is [deprecated](https://github.com/r-lib/tidyselect/issues/169) -- this trips me up too; the distinction is between "data-masking contexts" and "tidy selection contexts". But long story short, use quoted strings inside select now.
```suggestion
    dplyr::select("id", "name") %>%
```

base/db/R/insert.format.vars.R
```diff
@@ -48,18 +48,20 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
   }
 
-   #Test if skip is an integer
-  if(!is.character(skip)){
-  PEcAn.logger::logger.error(
-    "Skip must be of type character"
+   # Test if skip is numeric
+  if(!is.numeric(skip)){
+    PEcAn.logger::logger.error(
+      "Skip must be of type numeric"
```
infotroph, line 64: I see that enforcing character here was in clear violation of both the comment and the Roxygen documentation saying this should be an integer, but at the same time I think it was serving a good purpose -- below it's inserted into a database column of type character.  Maybe the right move is to enforce numeric here and then coerce it to character in `formats_df` below?

base/db/R/insert.format.vars.R
```diff
@@ -71,47 +73,51 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
   }
 
   # Test if notes are a character string
-  if(!is.character(notes)&!is.null(notes)){
+  if(!is.character(notes) & !is.null(notes)){
```
infotroph, line 76:
```suggestion
  if(!is.character(notes) && !is.null(notes)){
```

base/db/R/insert.format.vars.R
```diff
@@ -71,47 +73,51 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
         PEcAn.logger::logger.error(
           "variable_id must be an integer"
         )
       }
 
-      if(suppress == FALSE){
-        ## Test if variable_id already exists ##
-        var_id_test <- dplyr::tbl(con, "variables") %>% dplyr::select("id") %>% dplyr::filter(.data$id %in% !!formats_variables[[i, "variable_id"]]) %>% dplyr::collect(.data$id)
-        if(!is.null(var_id_test[1,1])){
+      if(identical(suppress, FALSE)){
```
infotroph, line 92: am I missing a reason this can't just be
```suggestion
      if (!suppress) {
```
?

base/db/R/insert.format.vars.R
```diff
@@ -71,47 +73,51 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
+        ## Test if variable_id exists in variables table ##
+        var_id_test <- dplyr::tbl(con, "variables") %>%
+          dplyr::select(.data$id) %>%
```
infotroph, line 95: As above
```suggestion
          dplyr::select("id") %>%
```

base/db/R/insert.format.vars.R
```diff
@@ -71,47 +73,51 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
+          dplyr::filter(.data$id %in% !!row_i[["variable_id"]]) %>%
+          dplyr::collect()
+        if(nrow(var_id_test) == 0){
```
infotroph, line 98: The whole function is attempting to add entries to the table, so emit a message when there _is_ an existing entry really is the desired behavior -- if there _isn't_ one, we'll add it.
```suggestion
        if (nrow(var_id_test) > 0) {
```

base/db/R/insert.format.vars.R
```diff
@@ -71,47 +73,51 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
           PEcAn.logger::logger.error(
-            "variable_id already exists"
+            "variable_id does not exist in 'variables' table"
```
infotroph, line 100:
```suggestion
            "variable_id already exists"
```

base/db/R/insert.format.vars.R
```diff
@@ -71,47 +73,51 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
           )
         }
       }
 
-      if(!is.character(formats_variables[[i, "name"]])&!is.na(formats_variables[[i, "name"]])){
+      if(!is.na(row_i[["name"]]) && !is.character(row_i[["name"]])){
```
infotroph, line 105: Out of curiosity, why the flipped order?

(Also I think both the old and new versions of this line will fail unexpectedly if passed an `NA_character_`, for which both `is.character` and `is.na` are true. But I haven't checked whether that's possible for these inputs)

base/db/R/insert.format.vars.R
```diff
@@ -122,32 +128,39 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
-      if(!units::ud_are_convertible(u1, u2)){
-        PEcAn.logger::logger.error(
-          "Units are not convertable."
-        )
+    for(i in seq_len(nrow(formats_variables))){
+      u1 <- formats_variables[["unit"]][i]
+      vid <- formats_variables[["variable_id"]][i]
+      if(!is.na(u1) && u1 != ""){
+        u2 <- dplyr::tbl(con, "variables") %>%
+          dplyr::select(.data$id, .data$units) %>%
```
infotroph, line 136:
```suggestion
          dplyr::select("id", "units") %>%
```

base/db/R/insert.format.vars.R
```diff
@@ -122,32 +128,39 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
-      }
-      # Grab the bety units and
-      if(!units::ud_are_convertible(u1, u2)){
-        PEcAn.logger::logger.error(
-          "Units are not convertable."
-        )
+    for(i in seq_len(nrow(formats_variables))){
+      u1 <- formats_variables[["unit"]][i]
+      vid <- formats_variables[["variable_id"]][i]
+      if(!is.na(u1) && u1 != ""){
```
infotroph, line 134: Why skip the rows with empty units? If both u1 and u2 are empty then there's no problem; if only one is empty it ought to emit a warning (as they will if the parseable/convertible tests below are applied to them), not be ignored.

(And there's also no need to check for NAs here because they were all filled as "" on line 128)

base/db/R/insert.format.vars.R
```diff
@@ -122,32 +128,39 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
+          dplyr::pull(.data$units)
+
+        if(!PEcAn.utils::unit_is_parseable(u1)){
+          PEcAn.logger::logger.error(
+            "Units '", u1,  "' not parseable.",
+            "Please provide a unit that is parseable by the udunits library."
+          )
+        }
+        # Grab the bety units and compare only if we have both units
+        if(length(u2) > 0 && !is.na(u2)){
```
infotroph, line 147: An empty result seems worth checking for and warning (erroring?) about rather than skipping over. Not sure how u2 would be NA here -- units are constrained to be not null in the variables table.

base/db/R/insert.format.vars.R
```diff
@@ -122,32 +128,39 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
+          }
+        }
       }
     }
   }
 
     formats_df <- tibble::tibble(
-      header = as.character(header),
+      header = header,
       skip = skip,
```
infotroph, line 160: Both `header` and `skip` need to be of type character when they get inserted into the `formats` table of BetyDB. As commented above, it _was_ a bit weird that `skip` was converted way up at the top of the function while `header` stayed logical until here. Let's coerce both in the same place:
```suggestion
      header = as.character(header),
      skip = as.character(skip),
```

base/db/R/insert.format.vars.R
```diff
@@ -122,32 +128,39 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
...
       mimetype_id = mimetype_id,
       notes = notes,
-      name = format_name,
-      stringsAsFactors = FALSE
```
infotroph, line 150: Nice catch!

infotroph (changes requested):
Thanks! A bunch of nice fixes here. 

I left comments on a few places where the existing behavior was preferable -- in fairness, all of them were confusing and I see why it would look like the change was needed. If you have ideas on how to further clarify those parts of the code I'd love to hear them.

infotroph:
@shbhmexe Why did you close this PR? You were on the right track and that it only needed minor changes to be  mergeable.

PR #3680, opened 2025-11-26
Update phenological parameters when there are NAs for specific site-years using across-year means

Qianyuxuan:
Update phenological parameters when there are NAs for specific site-years using across-year means

## Description
Update phenological parameters when there are NAs for specific site-years using across-year means.

## Motivation and Context
Resolve the issue that default leaf-on date of 144 and leaf-off date of 285 were used when missing or poor quality phenological data were presented in the parameter file derived from MODIS Land Cover Dynamics (MCD12Q2) Version 6.1 data product. There is also a need to rerun "PEcAn.data.remote::extract_phenology_MODIS" with start date as "2001-01-01" to calculate the site-specific across-year means.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/models/sipnet/R/write.configs.SIPNET.R
+++ b/models/sipnet/R/write.configs.SIPNET.R
@@ -487,18 +487,42 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
       leaf_pheno_path <- settings$run$inputs$leaf_phenology$path
       if (!is.null(leaf_pheno_path)) {
         ##read data
-        leafphdata <- utils::read.csv(leaf_pheno_path)
+        leafphdata <- utils::read.csv(leaf_pheno_path) #leaf phenology data starting from 2001-01-01 to current
         leafOnDay <- leafphdata$leafonday[leafphdata$year == obs_year_start
                                           & leafphdata$site_id == settings$run$site$id]
         leafOffDay <- leafphdata$leafoffday[leafphdata$year == obs_year_start
                                             & leafphdata$site_id == settings$run$site$id]
-        # when we have NAs for phenology.
-        if (is.na(leafOnDay)) {
-          leafOnDay <- param[which(param[, 1] == "leafOnDay"), 2]
+        # when we have NAs for phenology (or missing years)
+        if (length(leafOnDay) == 0 || is.na(leafOnDay)) {
+          # 1. Try to calculate the mean across all available years for this site
+          site_phenology_on <- leafphdata$leafonday[leafphdata$site_id == settings$run$site$id]
+          mean_on <- mean(site_phenology_on, na.rm = TRUE)
+          
+          if (!is.nan(mean_on) && !is.na(mean_on)) {
+            leafOnDay <- round(mean_on)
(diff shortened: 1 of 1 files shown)


DongchenZ (commented):
In general, the PR looks good. However, under certain circumstances, we should not average across phenological DOYs if they vary too much (e.g., averaging between 30 and 150 should not be practical). So, it's good to set a threshold to filter out noisy time series.

dlebauer:
@Qianyuxuan this looks good to me. Only wondering if this should be applied upstream rather than in model specific fn, though we can defer that question to future work. 

Please cite Wheeler et al 2023 in the function roxygen docs. We found "historical means are difficult to outperform (in predicting greenup across a wide range of temperate sites)". https://doi.org/10.1016/j.agrformet.2023.109810

Also please update news + changelog.

@DongchenZ Is the scenario you are referring to where the historical values are incorrect, or is there a specific real world example where a wide range would bias the estimate? I could imagine checking for outliers and having a fallback (median?) but mean is simple and effective.

mdietze (approved):
Agree with @dlebauer that a future PR should pull this logic out of write.configs and either move it into the code that generates the phenology files or into some intermediate step, because this sort of gap-filling isn't model specific

@infotroph the GHA fail on this PR is a new one for me. Any guess what needs to be tweaked?

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -487,18 +487,42 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
...
         }
-        if (is.na(leafOffDay)) {
-          leafOffDay <- param[which(param[, 1] == "leafOffDay"), 2]
+        
+        if (length(leafOffDay) == 0 || is.na(leafOffDay)) {
+          # 1. Try to calculate the mean across all available years for this site
+          site_phenology_off <- leafphdata$leafoffday[leafphdata$site_id == settings$run$site$id]
+          mean_off <- mean(site_phenology_off, na.rm = TRUE)
+          
+          if (!is.nan(mean_off) && !is.na(mean_off)) {
```
infotroph, line 516: This works; if you want to play code golf it would be OK to cut it down to `if (!is.na(mean_off)) {`, because it turns out `is.na(NaN)` => TRUE. 
(But `is.nan(NA)` => FALSE, to keep us on our toes and keep the behavior from being easy to remember 😜).

infotroph:
The CI failure is coming from in or around `/scripts/ensure_version_bump.sh`, which is my fault, but the weird part is the git hash referenced in the error message doesn't seem to exist in history. I'll keep digging.

infotroph:
Best I can tell this was driven by a temporary failure in Github's releases API (which we query to find the most recent tag and pass it to `ensure_version_bump.sh`), compounded by some commit history not being present in GHA's default shallow clones. 

Rerunning the job seems to have worked; I'll merge this and work on a most robust fix for the test as a separate PR.

PR #3684, opened 2025-11-27
Integrate Quarto notebooks into documentation site

### LINKED ISSUE #3681: Add Rendered Demo Output to PEcAn Documentation Site
@infotroph in https://github.com/PecanProject/pecan/pull/3638#pullrequestreview-3445276018

>As long as we've generated a rendered artifact, should it be posted somewhere in /accessible from the rest of the rendered documentation?

@infotroph in https://github.com/PecanProject/pecan/pull/3638#issuecomment-3564717478

>If you're thinking of something that uses upload_artifact, I'm not strongly opposed but I doubt many people would brave GHA's opaque artifact URLs to find it. I'm thinking more of something that, say, pushes a rendered HTML version into the appropriate place in the documentation repo so that folks visiting https://pecanproject.github.io/pecan-documentation/develop/demo-2.html can see the most recently rendered output as part of the PEcAn book. I haven't thought through the implementation enough to say whether it should appear as the literal contents of that page or as a separate page linked from there or what.

AritraDey-Dev:
## Description

This PR integrates the Quarto tutorials into the documentation build. It updates `render-quarto.yml` to generate HTML artifacts in CI and modifies `book.yml` to include those rendered tutorials in the docs. 

## Motivation and Context

Fixes #3681 
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -12,7 +12,11 @@ on:
   workflow_dispatch:
 
 jobs:
+  call-render-quarto:
+    uses: ./.github/workflows/render-quarto.yml
+
   bookdown:
+    needs: call-render-quarto
     runs-on: ubuntu-latest  
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
@@ -45,6 +49,11 @@ jobs:
       run: |
         cd book_source
         Rscript -e 'options(bookdown.render.file_scope=FALSE); bookdown::render_book("index.Rmd", "bookdown::gitbook")'
+    # download quarto artifacts
+    - uses: actions/download-artifact@v4
+      with:
+        name: quarto-html
+        path: book_source/_book/
     # save artifact
     - uses: actions/upload-artifact@v4
       with:
--- a/.github/workflows/render-quarto.yml
+++ b/.github/workflows/render-quarto.yml
@@ -4,6 +4,7 @@ on:
   push:
   pull_request:
   workflow_dispatch:
+  workflow_call:
 
 jobs:
   render-notebook:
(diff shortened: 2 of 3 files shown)


dlebauer:
@AritraDey-Dev this seems like a good time to update references in the PEcAn README and anywhere else in docs that point to the web interface as a good place to start. We can demote refs to the web interface, but note that it will be deprecated in PEcAn v2.

dlebauer (approved):
Awesome, thanks for doing this!

infotroph:
@dlebauer Please revert.

As I think y'all discussed previously somewhere in this chain of PRs, book.yml is strictly doing Markdown conversion and _should not_ take on either the computation cost or the dpendencies of running any PEcAn code. What I was suggesting in #3681 was to have the demo rendering process upload its output to the documentation site _independently of_ the bookdown compilation, each at paths that won't conflict with each other.

AritraDey-Dev:
@infotroph The way I set up the workflow, we are not doing any heavy or computational rendering inside the bookdown workflow. Like you suggested, the rendering step is done separately. After rendering, the workflow uploads the generated HTML artifacts, and then the bookdown workflow simply downloads them here:
https://github.com/PecanProject/pecan/blob/827613ffaf3beaa1d7de10d79cedf7a756bc58a1/.github/workflows/book.yml#L52-L61
So the only extra step we’re doing in bookdown is downloading those generated artifacts so they can be included on the website. If you still think this counts as computational work, let me know — I can revert it and try another approach. 
But I think we still need some kind of integration with the bookdown workflow so the rendered HTML ends up in the site.

[ One possible optimization: right now we upload and download the artifacts for every PR. Maybe we shouldn’t do that. We could do that step only on push, instead of every PR event. ]

infotroph:
@AritraDey-Dev Sure, the computation lives in a separate job, and you're right that means dependencies aren't an issue. But the 2-minute bookdown job still has to wait for the quarto rendering job to complete (currently about 9 minutes, presumably will get longer as more notebooks added), and it means nobody can reason about one of these actions without also understanding the other one.

I recommend the Quarto workflow implement its own copy of what's done by current lines 62-94 of book.yml: 
* check out the pecan-documentation repo
* add the rendered files -- I think they should go in the existing `02_demos_tutorials_workflows/` folder, but I don't think it matters a lot as long as the link from the demos table is correct
* push the result if the action ran on a push event (i.e. if we're merging to develop/main)

If you want to avoid duplicating code, this chunk of book.yml could probably become its own callable workflow that then gets invoked separately by the Bookdown job and by the Quarto job (or potentially the array of Quarto jobs, if that makes sense to do later)

AritraDey-Dev:
@infotroph that approach makes sense since the Quarto workflow will commit to the documentation repo separately. I won’t revert this PR now; instead, I’ll create a new one with the fix and remove the changes made in this PR.

PR #3685, opened 2025-11-28
build(docs): minor cleanups in book_source/Makefile

AritraDey-Dev:
## Description

A minor follow-up cleanup after #3671.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/book_source/Makefile
+++ b/book_source/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all bkdcheck build clean serve
+.PHONY: all bkdcheck build clean
 
 all: bkdcheck build
 
@@ -32,9 +32,6 @@ CLEAN_DIRS = _book extfiles
 
 clean:
 	rm -rf $(CLEAN_DIRS)
-
-serve: build
-	Rscript -e 'servr::httd("_book", browser = TRUE)'
 
 pdf: bkdcheck
 	Rscript -e 'options(bookdown.render.file_scope=FALSE); bookdown::render_book("index.Rmd", "bookdown::pdf_book")'
--- a/book_source/Makefile
+++ b/book_source/Makefile
@@ -1,6 +1,6 @@
-.PHONY: all bkdcheck build clean deploy
+.PHONY: all bkdcheck build clean serve
 
-all: bkdcheck build deploy
+all: bkdcheck build
 
 bkdcheck:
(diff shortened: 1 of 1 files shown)


infotroph:
How is this better than opening the built index.html directly?

AritraDey-Dev:
It’s much easier than finding and opening the generated HTML manually every time. With `make serve`, you just run one command and it automatically serves the book in your browser. Another small benefit is that opening the file manually uses a `file://` URL, while make serve uses `http://`, which gives you a preview closer to a real deployment. 
It also helps someone new who doesn’t have experience with how `bookdown` generates output—they can simply run `make serve` and get the results.

infotroph:
> ... easier than finding and opening the generated HTML ...
> ... helps someone new who doesn’t have experience with how bookdown generates output ...

I'd be more sympathetic to this argument if the last line of console output from the build step weren't `Output created: _book/index.html`. Every part of PEcAn requires its developers to think about local filepaths printed in log files, while we try hard _not_ to assume that everyone knows what "Serving the directory at http://127.0.0.1:4321" means. I don't see adding the second kind of complexity as a desirable way of reducing the first.

> a preview closer to a real deployment

The point of building to static HTML is that the compiled files _are_ real. I promise there's no extra magic added by delivering them over HTTP to a different place on the same computer.


Another reason I don't think a `make serve` makes sense is that processes which need interactive input or spawn persistent background processes are generally a bad fit for Make -- it's intended to run with one simple invocation, do what's needed to create a set of files, and exit when it's done. Looks to me like `servr::httd()` blocks until interrupted, meaning `make serve` would never exit until you killed it with ctrl-C.

AritraDey-Dev:
Thanks for the clarification. The reason behind doing this is that most of the time I use the `serve::httd` command to open the build in a browser.

>  I'd be more sympathetic to this argument if the last line of console output from the build step weren't Output created: _book/index.html.

But this definitely makes more sense since we have output line saying `Output created: _book/index.html.`

I will remove serve rule from the PR and just keep the cleanup part that is required after #3671.

infotroph (approved):
Thanks for quick adaptation to feedback! This now looks good.

PR #3694, opened 2025-12-03
Refactor Quarto workflow: HTML rendering and independent deployment

AritraDey-Dev:
## Description

Refactors the Quarto workflow to independently render and deploy HTML tutorials to the pecan-documentation repo, decoupling it from the main book build.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


.github/workflows/render-quarto.yml
```diff
@@ -60,4 +60,46 @@ jobs:
...
+          
+          # Determine version/branch
+          export VERSION=$(echo $GITHUB_REF | sed 's,.*/,,' )
+          if [ "$VERSION" = "main" ]; then
+            export VERSION=latest
+          fi
+          
+          # Create target directory
+          TARGET_DIR="pecan-documentation/${VERSION}/"
+          mkdir -p $TARGET_DIR
```
AritraDey-Dev, line 96: I’m not entirely sure whether this is the correct path, but looking at the https://github.com/PecanProject/pecan-documentation/tree/main/develop structure, it seems to be the right one.

infotroph: Thanks for checking -- the wrinkle I didn't think of earlier is that since book.yml (correctly) uses rsync with `--delete` so that it can remove old files as well as add/update them, we need to either write to a path the bookdown script doesn't manage or tell the two scripts how to coordinate their files.

I lean toward writing the Quarto outputs to `pecan-documentation/${VERSION}/rendered-demo-notebooks/` and then updating the rsync command in book.yml with `--exclude=rendered-demo-notebooks/` -- thoughts?

infotroph:
```suggestion
          TARGET_DIR="pecan-documentation/${VERSION}/rendered-demo-notebooks"
          mkdir -p $TARGET_DIR
```

.github/workflows/render-quarto.yml
```diff
@@ -60,4 +60,46 @@ jobs:
...
+          
+          # Copy HTML files and support folders
+          cp -r quarto_html_build/* $TARGET_DIR/
```
infotroph:
```suggestion
          rsync -a --delete quarto_html_build/* $TARGET_DIR/
          echo "Autogenerated by https://github.com/PecanProject/pecan/actions/workflows/render-quarto.yml" > "$TARGET_DIR"/README.txt
```

.github/workflows/render-quarto.yml
infotroph: If we go this route obviously the links to these notebooks in the table of demos will need to be updated as well -- your call whether to do that in this PR or a separate one

infotroph (changes requested):

.github/workflows/render-quarto.yml
AritraDey-Dev: >pecan-documentation/${VERSION}/rendered-demo-notebooks/ and then updating the rsync command in book.yml with --exclude=rendered-demo-notebooks/ -- thoughts?

Yes, that makes sense. After some digging, I think `--exclude` should work fine. Unfortunately, I don't know a way to test this part of the workflow. For now, using` --exclude `seems like the correct approach.

AritraDey-Dev: > If we go this route obviously the links to these notebooks in the table of demos will need to be updated as well -- your call whether to do that in this PR or a separate one

yes need to update that table in that case,it should be `rendered-demo-notebooks/run_pecan.html` or `rendered-demo-notebooks/uncertainty.html`

infotroph: >  I don't know a way to test this part

Yep, but you're in good company. This is the part of GHA development where the median approach seems to be "push it to prod and hope" 😓

infotroph (approved):
Thanks! This all looks right to me now.

AritraDey-Dev:
@infotroph it worked, the HTML links are redirecting to the correct place.
But I'm not sure why another directory is being created here: https://github.com/PecanProject/pecan-documentation/tree/main/pr-3694-381c450d92f0921c3f3391bf6c10786ed7f35c16/rendered-demo-notebooks.

It looks like when we add a PR to the merge queue, it creates another branch and deploys to that branch as well.

AritraDey-Dev:
Understood the issue. Basically, it is committing changes for pushes to any branch, so the merge queue creates another directory as well.
We just need to fix the workflow triggers.

Issue #3679, opened 2025-11-25
Bump package versions for 1.10.0 release


infotroph:
Similar to #3529

* [x] All packages bumped away from dev version
* [x] version now > version in 1.9.0 unless package is actually unchanged
* [x] versions agree between NEWS and DESCRIPTION 
* [x] `pecan_releases.csv` and `pecan_version_history.csv` updated with correct date (it's annoying that we need to do this by hand, but oh well)
* [ ] Document all of this for next time

PR #3704, opened 2025-12-06
version bump for model packages

infotroph:
## Description

Part of #3679: Went through the models/ folder and checked versions and NEWS entries. Similar treatment for base/ and modules/ coming soon.

Note on models/template: I broke the cardinal rule here and _reduced_ the package version to 0.0.1 on the grounds it isn't a real package whose history needs to be traced carefully, it's a template that needs to give people reasonable starting values for their brand-new package... which definitely shouldn't start at > v1.0.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


dlebauer (approved):
LGTM. Makes sense to revert template to 0.0.1

PR #3707, opened 2025-12-07
Meta Analysis Quarto NoteBook

AritraDey-Dev:
## Description

This PR implements a Quarto notebook for the meta-analysis. It uses the posterior files and `trait.data.Rdata` to run the Meta Analysis demo. A pecan.xml file is also included to run the workflow.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+    fig-width: 10
+    fig-height: 6
+    fig-dpi: 300
+---
+
+# Introduction {#introduction}
+
+Meta-analysis in PEcAn is a hierarchical Bayesian statistical approach that synthesizes plant trait data from literature to constrain ecosystem model parameters. The **PEcAn.MA** module implements this functionality to combine prior information with observational data, generating posterior distributions for model parameters.
+
+In a standard PEcAn workflow, this step queries the BETYdb database for trait data and priors. For this demonstration, we will use pre-generated data files to simulate the workflow without requiring an active database connection during the notebook execution.
```
mdietze: I'd rephase this. The MA runs on tabular data in a specific format. One way to easily get data in that format is to query BETYdb, but you can also generate that format manually if you have other trait data. Indeed, a great new Issue for first time PEcAn developers would be create a helper function(s) that reformats trait data from common trait databases (e.g. TRY) into the tabular format this module is expecting. Given that no one is actively updating BETY, this approach is probably going to be the de facto norm for most users, and in the future there should be an update to this demo once we have functions that enable this

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+1.  Load and configure PEcAn settings for meta-analysis.
+2.  Understand the `meta.analysis` settings in `pecan.xml`.
+3.  Run the meta-analysis module using pre-generated trait data.
+4.  Analyze and visualize the posterior distributions and MCMC diagnostics.
+
+## The Scenario
+
+**Context & modeling scenario:**
+
+We simulate plant and ecosystem carbon balance (Net Primary Productivity and Net Ecosystem Exchange) at the AmeriFlux Niwot Ridge Forest site ([US‑NR1](https://ameriflux.lbl.gov/sites/siteinfo/US-NR1)) during the year 2004. We use SIPNET parameterized as a temperate conifer PFT and driven by AmeriFlux meteorology following the analysis in [Moore et al. (2007)](https://doi.org/10.1016/j.agrformet.2008.04.013). This notebook also provides a compact template that can be extended to more years, locations, and PFTs.
```
mdietze: language here is a bit rough. Is this something you're telling the user they want to do? An example? A reference back to Demo 1 and Demo 2?  Also note that one can use the MA without then feeding the posteriors into a model (though we definitely want to highlight the latter)

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+
+In this specific demo, we are performing a meta-analysis for the **temperate coniferous** Plant Functional Type (PFT). The goal is to estimate the probability distributions for key model parameters (e.g., SLA, leaf turnover rate) by combining:
+*   **Priors**: Existing knowledge about the parameters.
+*   **Data**: Observed trait data (simulated for this demo).
```
mdietze: Is the demo really using simulated data? Why? We've got plenty of real data

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+
+*   **PEcAn packages installed**: You can install all PEcAn packages and their dependencies by running the following command in the root of your PEcAn repository:
+
+```r
+# Enable repository from pecanproject
+options(repos = c(
+  pecanproject = 'https://pecanproject.r-universe.dev',
+  CRAN = 'https://cloud.r-project.org'))
+# Download and install PEcAn.all in R
+install.packages('PEcAn.all')
```
mdietze: I'd recommend a bit more nuance here. If you just want to run a trait meta-analysis you don't need to install PEcAn.all, but rather just a subset of packages (and it would be good to show which subset). If you want to then run a model using those posteriors you probably will end up installing PEcAn.all.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+```
+
+*   **A valid `pecan.xml` configuration file**: Start with the example at `pecan/documentation/tutorials/Demo_03_Meta_Analysis/pecan.xml`.
```
mdietze: Issue to open: one should be able to run the MA itself without a full settings object

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+*   **SIPNET binary**: While not strictly used for the meta-analysis calculation itself, it is part of the broader workflow context.
+*   **Pre-generated Data**: This demo relies on `trait.data.Rdata` and `prior.distns.Rdata` files which are included in the `pft/temperate.coniferous` directory.
+
+## Install SIPNET and Meteorological Data
```
mdietze: If this was covered in Demo 1 or Demo 2 send users there rather than duplicating text. If you duplicate then there's twice as much text to keep up-to-date if anything changes

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+# Read the settings from the pecan.xml file
+settings <- PEcAn.settings::read.settings(settings_path)
+
+# Prepare and validate the settings
+settings <- PEcAn.settings::prepare.settings(settings)
+```
+
+See Demo 1 Section 6 for details on what these functions do. Briefly, they read the XML file, convert it into an R list object that PEcAn can use, check that settings are valid, fill in defaults, and create the output directory.
+
+## Explore the Settings Object
```
mdietze: not needed except for the parts related to the MA, which you've already covered

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+
+```{r explore-settings}
+str(settings)
+```
+
+
+# Run Meta Analysis
+
+We now run the meta-analysis. The `runModule.run.meta.analysis` function will:
+1.  Read the `trait.data.Rdata` and `prior.distns.Rdata` from the PFT output directory.
```
mdietze: Issue to open: would be great to be able to pass data into the MA directly, rather than it having to come from a file with a very specific file name

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+# for (pft in settings$pfts) {
+#   print(paste("Posterior Distributions for PFT:", pft$name))
+#   load(file.path(pft$outdir, "post.distns.MA.Rdata"))
+#   print(post.distns)
+# }
+```
+
+# Visualize Meta Analysis Results {#sec-visualize}
+
+It is important to check the MCMC chains for convergence. We can visualize the trace plots and density plots for each trait.
```
mdietze: might be nice to provide additional explanation about what these figures are showing and how to interpret them

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+#
+#   # Iterate through each trait and plot the MCMC chains
+#   for (trait in names(trait.mcmc)) {
+#     print(paste("Plotting", trait))
+#     plot(trait.mcmc[[trait]], main = paste("MCMC Plots for", trait, "(", pft$name, ")"))
+#   }
+# }
+```
+
+# Conclusion
```
mdietze, line 203: 1. Notebook doesn't go on to use these posteriors in a set of model runs so there's definitely no need to include the text and code for installing the model, installing all PEcAn packages, etc.
2. Notebook doesn't explain HOW one goes on to use these posteriors in a set of model runs. Could be as simple as "now edit your pecan.xml to point <posterior.files> to these outputs and rerun Demo 02. How did your results change in terms of the width of the CI and the uncertainty analyses?"

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,277 @@
...
+# Further Exploration
+
+The [next set of tutorials](#demo-table) will focus on the process of data assimilation and parameter estimation. The next two steps are in “.Rmd” files which can be viewed online.
+
+**Assimilation 'by hand'**
+
+[Explore](https://github.com/PecanProject/pecan/blob/main/documentation/tutorials/sensitivity/PEcAn_sensitivity_tutorial_v1.0.Rmd) how model error changes as a function of parameter value (i.e. data assimilation ‘by hand’)
+
+
+**MCMC Concepts**
```
mdietze, line 276: Both these two modules are fairly pedagogical about calibration concepts, but neither really show how to run the PEcAn calibration or SDA code. I'd include those modules too.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Will add them.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Yes, for this meta-analysis the posterior files are not needed.will remove this block.

AritraDey-Dev: A follow-up question: in this case, the model block in the settings configuration file (`pecan.xml`) isn’t needed, and `run$input` isn’t needed either. Basically, this whole section:
Basically this part
```
 <model>
  <type>SIPNET</type>
  <revision>git</revision>
  <delete.raw>FALSE</delete.raw>
  <binary>demo_outdir/sipnet</binary>
 </model>
 <run>
  <site>
   <met.start>2004/01/01</met.start>
   <met.end>2004/12/31</met.end>
   <name>Niwot Ridge Forest/LTER NWT1 (US-NR1)</name>
   <lat>40.0329</lat>
   <lon>-105.546</lon>
  </site>
  <inputs>
   <met>
    <source>AmerifluxLBL</source>
    <output>SIPNET</output>
    <username>Aritra_2004</username>
    <path>
     <path1>dbfiles/AMF_US-NR1_BASE_HH_23-5.2004-01-01.2004-12-31.clim</path1>
    </path>
   </met>
  </inputs>
  <start.date>2004/01/01</start.date>
  <end.date>2004/12/31</end.date>
 </run>
```
Should we remove this, or keep it so that users still get a clear idea of what the settings file normally looks like(Demo 1 and demo 2 does it though)?

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev:
```
<?xml version="1.0" encoding="UTF-8"?>
<pecan>
 <pfts>
  <pft>
   <name>temperate.coniferous</name>
   <posterior.files>pft/temperate.coniferous</posterior.files>
   <outdir>pft/temperate.coniferous</outdir>
  </pft>
 </pfts> 
 <meta.analysis>
    <iter>3000</iter>
    <random.effects>
      <on>FALSE</on>
      <use_ghs>TRUE</use_ghs>
    </random.effects><threshold>1.2</threshold>
  </meta.analysis>
</pecan>

```
This configuration alone is enough to run the meta-analysis; no additional model or run blocks are required.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Yes, I just copied this part from Demo 1 and Demo 2.fixed it now.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Yes, I believe this needs a separate function to implement it. We just need to pass `trait.data.Rdata` and `prior.distns.Rdata` to make it work.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Not **simulated**,it should be pre-generated.I just query trait data with the settings and save it as from the db(local postgres setup) `trait.data.Rdata `particularly for this demo.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Instead, can we do this within this notebook by adding a section where the user only needs to provide the paths to `trait.data.Rdata` and `prior.distns.Rdata` **?** If a user already has these two files, they can run the meta-analysis directly.

mdietze: follow up: I'm not requesting a change to this PR to implement what I'm suggesting. I'm instead asking that you open a new Issue to improve what we're doing in the future. Specifically, I'd recommend that the MA module take in the trait dataframe and prior dataframe as arguments to the function itself, rather than relying on the functions knowing to load those specific files from paths provided within an overly complex settings object. This will push a tiny bit of work into the demo (load the example files, look at them to see how they are formatted, pass them into the MA function) but IMHO will greatly increase the usability of the MA module as a stand-alone tool.  Right now, it's functionally easy to to use the MA outside the PEcAn workflow, but its CONCEPTUALLY hard to do so because there's so much mystery in what it's doing. Right now, no one can actually run this as a stand alone module in practice without a whole lot of diving into the code to see what the module actually does and what it actually needs to work. Actually getting this working and documented might be a good place for a new GSOC student.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
mdietze: I'd vote against keeping anything that you don't need. It's fine, in practice, to say that the settings for using these posteriors for forward model simulation is more complicated (see Demo1 and Demo2) and that in practice you can write one settings file that contains both the model run and MA settings and run it all as a single workflow.  Making the settings here minimal is an asset for making the module more accessible and for ultimately moving towards dropping `settings` as an argument and instead passing the function just the info it needs

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
mdietze: I'd hoped that was the case. Just make sure the text matches what you actually did.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
mdietze: Cool. So I think it would be fine in the Issue I'm recommending to recommend that the MA module be refactored to take in the following as arguments:
1. trait dataframe
2. prior dataframe
3. output directory
4. list containing MA configs (iter, random.effects, use_ghs, theshold, etc.) with some sensible defaults

Here in the demo one could elect to grab those things from a settings object, but one could also build a demo based on just specifying those things.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: created here https://github.com/PecanProject/pecan/issues/3717

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Done here #3718

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Actually,This was taken from Demo 1 and Demo 2 to show the scenario we were considering for this notebook. However, in our case this scenario is incorrect because we are not using site information or met data in the settings file.need to update the scenario to match the minimal settings file used in this demo i.e. https://github.com/PecanProject/pecan/pull/3707#discussion_r2602936922.

AritraDey-Dev: Adjusted this in https://github.com/PecanProject/pecan/pull/3707/commits/ac4ece1d6c3b33d9539a2032f2a8824564ea159b..

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: added in https://github.com/PecanProject/pecan/pull/3707/commits/eb4c2f40e763526207a16ccc28cc93b6f67d865d

AritraDey-Dev:
@mdietze Thanks for the review and the detailed comments. I’ve tried to address and fix the suggested items in the subsequent commits.

Mayanknishad9:
@mdietze This is a great suggestion. My `format_try_for_ma()` function (PR #3720 ) creates the `trait_data` dataframe from TRY exports that could feed directly into such a simplified MA function. 

If the MA module is refactored as you suggest, users could:
1. Use `format_try_for_ma()` to convert TRY data to proper format
2. Pass that dataframe directly to a simplified `run_meta_analysis()` function
3. Bypass the complex settings object entirely

This would make PEcAn's meta-analysis much more accessible to researchers with external data sources.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
Mayanknishad9: @AritraDey-Dev Thanks for creating issue #3717! I've implemented the `format_try_for_ma()` function in PR #3720. 

This connects directly to @mdietze's suggestion about using external data. Once PR #3720 (TRY formatter) and issue #3718 (MA refactoring) are both implemented, this tutorial could show how to use TRY data with PEcAn's meta-analysis.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
mdietze: update: If this demo is updated to use the new `meta_analysis_standalone` function then I think we'll be able to get away from a pecan.xml and instead just have a code block that defines the inputs and configurations that the  MA needs. That should make the use of the MA package more transparent. 

That said, at the end of the demo you could show how to do the same thing with a pecan.xml and `runModule.run.meta.analysis`, which could reuse a lot of the text and code you've already written, and could point out that the "module" version is handy if you're running the MA as part of a larger modeling workflow

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,278 @@
...
+*   **Pre-generated Data**: This demo relies on `trait.data.Rdata` and `prior.distns.Rdata` files which are included in the `pft/temperate.coniferous` directory.
+
+# Load PEcAn Packages
+
+First, we need to load the PEcAn R packages. These packages provide all the functions we'll use to run the workflow.
+
+```{r libraries}
+# Load necessary PEcAn packages
+# library("PEcAn.all") # Alternatively, load all packages if installed
+library("PEcAn.settings")
```
mdietze, line 66: If you switch to meta_analysis_standalone you might only need PEcAn.MA and could push the other library loads to the end when you show the module version.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
mdietze: per earlier comments, would be good to update to meta_analysis_standalone in this PR. Then #3720 could include an update to the tutorial mentioning the new TRY conversion function. This PR shouldn't wait for that -- because we don't have permissions to redistribute TRY data we can't include a fully worked example in the Demo.

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
AritraDey-Dev: Updated in https://github.com/PecanProject/pecan/pull/3707/changes/c40a0e05e72a2153dc31bb869b604880ef37a32a

documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd
```diff
@@ -0,0 +1,309 @@
...
+4.  Analyze and visualize the posterior distributions and MCMC diagnostics.
+5.  (Optional) Run meta-analysis using `pecan.xml` configuration (Workflow approach).
+
+## The Scenario
+
+**Context & modeling scenario:**
+
+This demo supports the modeling scenario introduced in [Demo 01](../Demo_1_Basic_Run/run_pecan.qmd) and [Demo 02](../Demo_02_Uncertainty_Analysis/uncertainty.qmd), which simulated ecosystem carbon balance at the AmeriFlux Niwot Ridge Forest site ([US‑NR1](https://ameriflux.lbl.gov/sites/siteinfo/US-NR1)) using the SIPNET model.
+
+While those demos focused on running the ecosystem model, this notebook zooms in on the **Meta-Analysis** step. We will generate informative posterior distributions for the **temperate coniferous** Plant Functional Type (PFT) used in those simulations. These posteriors can then be used to constrain the parameters of the SIPNET model (or other models) to improve prediction accuracy.
```
mdietze: This paragraph and the next are a bit repetitive. Recommend merging

mdietze (approved):

PR #3708, opened 2025-12-07
Refactor workflow to stateless run manifest architecture and SA input design coordination

### LINKED ISSUE #3692: Replace stateful samples.Rdata logging with structured manifest file (runs_manifest.csv)
Currently `run.write.configs` acts as a "stateful" logger, appending run information to `samples.Rdata`. As we scale to hundreds or thousands of sites, this approach becomes inefficient and tightly couples the configuration step with the analysis step.

This issue proposes refactoring the downstream analysis functions ( e.g. `read.ensemble.output`, `read.sa.output` , etc) to adopt a stateless design that reads from a manifest, removing the dependency on runtime mutation of samples.Rdata.

`run.write.configs` generates run ids (e.g. ENS-0001-siteID) and physically saves them into the `runs.samples` list within samples.Rdata.
The samples.Rdata file grows linearly with the number of sites. The analysis modules depend on this file being constantly updated/mutated by the write step.

Proposed workaround : 

Refactor the workflow to decouple "parameter definition" from "execution logging" by introducing a lightweight, Manifest file.

1. samples.Rdata becomes static treat samples.Rdata strictly as a "Master parameter definition" file (generated upstream by `get.parameter.samples`). It should be immutable during the run.write.configs step.

2. Introduce runs_manifest.csv instead of modifying the RData file, run.write.configs will generate a structured CSV file in the output directory. This file explicitly maps run ids to their design parameters.

proposed structure (`runs_manifest.csv`):

| run_id            | site_id | pft_name | trait | quantile | type         |
|-------------------|---------|----------|-------|----------|--------------|
| ENS-0001-siteA     | siteA   | NA       | NA    | NA       | Ensemble     |
| SA-median-siteA    | siteA   | grass     | NA    | 0.5      | Sensitivity  |
| SA-pft_name-T1-Q1-siteA     | siteA   | grass     | SLA   | 0.158    | Sensitivity  |

3. update downstream analysis functions (read.ensemble.output, read.sa.output, etc..) to read this CSV manifest.
logic: look up the `run_id` where `site_id == X` and `trait == Y`

### LINKED ISSUE #3693: Vectorize variable reading in `read.sa.output` to reduce disk I/O
While reviewing downstream dependencies for Multisite SA support, we identified a severe I/O bottleneck in `read.sa.output`

The current implementation uses a triple-nested loop structure. `read.output` was being called repeatedly inside the loops, but `get.results` was wrapping it in yet another loop. That is definitely a performance overhead and extremely inefficient for file access.
This means if a user requests 5 variables, it opens and closes the exact same netcdf file 5 separate times. For a large _sensitivity analysis or multisite run_, this results in thousands of redundant disk I/O operations, significantly slowing down the analysis step

Solution : refactor `read.sa.output` (and potentially `get.results`) to use vectorized reading: 
1. open the target netcdf file once per run.
2. read all requested variables into memory in a single pass.
3. close the file.

Identify the places where these inefficiency exists in downstream and refactor it.

divine7022:
## Description
fixes #3692
- updated `run.write.configs` to generate a `runs_manifest.csv` file containing run metadata (run_id, site_id, pft_name, trait, quantile, type) instead of appending to samples.Rdata.
- samples.Rdata is now treated as a read-only "Master parameter definition" file.
- updated write.sa.configs and write.ensemble.configs to return manifest dataframes.

fixes #3693
- refactored `read.sa.output` and `read.ensemble.output` to pass the full vector of requested variables to read.output in a single call.
- eliminated the inefficient inner loop that opened/closed netcdf files for every individual variable.

Additionally this PR separates input design generation for SA and ensemble runs, fixing dimension mismatch errors in _multisite_ workflows.
previously both SA and ensemble used a single `input_design` matrix, causing failures when:
- Ensemble needed N runs (e.g. 50)
- SA needed 1 (median) + (traits * quantiles) runs (e.g. 145)

causes in dimension mismatch errors in input design coordination.

`runModule.run.write.configs` generates separate `ens_input_design` and `sa_input_design` matrices.
`run.write.configs` accepts both `input_design_ens` and `input_design_sa` parameters.
Old `input_design` parameter still works with deprecation warning.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/workflow/R/run.write.configs.R
```diff
@@ -8,11 +8,13 @@
...
 #' @param settings a PEcAn settings list
 #' @param ensemble.size number of ensemble runs
-#' @param input_design input indices for samples
 #' @param write should the runs be written to the database?
 #' @param posterior.files Filenames for posteriors for drawing samples for ensemble and sensitivity
 #'    analysis (e.g. post.distns.Rdata, or prior.distns.Rdata)
 #' @param overwrite logical: Replace output files that already exist?
+#' @param input_design DEPRECATED. Use input_design_ens and input_design_sa instead
+#' @param input_design_ens Input design matrix for ensemble analysis
+#' @param input_design_sa Input design matrix for sensitivity analysis
```
infotroph: I'm not a fan of having this be three separate arguments. What if we kept `input_design`, expecting a list with entries `ensemble` and `sensitivity` and also accepting a single design object (interpreting it as being for the ensemble)?

(We discussed live that it'd be nice to encode SA and ensemble setup into the design such that run.write.configs doesn't even need to know which one it's working with, but that's still well in the future)

base/workflow/R/run.write.configs.R
```diff
@@ -211,13 +228,26 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+                       append = file.exists(manifest.file))
+
+    PEcAn.logger::logger.info("Run manifest written to ", manifest.file)
+  }
+
   options(scipen = scipen)
   invisible(settings)
   return(settings)
-}
+}
```
infotroph: Nit: Please don't remove terminating newlines. These days the rule is mostly just to avoid needless extra changes, but there do exist some (mostly ancient 😅 ) tools that complain if it isn't there.

base/workflow/R/run.write.configs.R
```diff
@@ -211,13 +228,26 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
-  # A lot of this is duplicate with the ensemble/sa specific output above, but kept for backwards compatibility.
-  save(ensemble.samples, trait.samples, sa.samples, runs.samples, pft.names, trait.names,
-    file = file.path(settings$outdir, "samples.Rdata")
-  )
-  PEcAn.logger::logger.info("parameter values for runs in ", file.path(settings$outdir, "samples.RData"))
+  # write runs manifest
+
+  manifest.file <- file.path(settings$outdir, "runs_manifest.csv")
+
+  if (nrow(run_manifest_df) > 0) {
```
infotroph: Is it OK for the manifest to be empty? If so I'd favor writing a zero-row file so it's easy to tell that afterwards; if not OK then should we throw an error here instead of skipping?

base/workflow/R/run.write.configs.R
```diff
@@ -211,13 +228,26 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+    if (overwrite && file.exists(manifest.file)) {
+      unlink(manifest.file)
+    }
+
+    utils::write.table(run_manifest_df,
+                       file = manifest.file,
+                       sep = ",",
+                       row.names = FALSE,
+                       col.names = !file.exists(manifest.file),
+                       append = file.exists(manifest.file))
```
infotroph: write.table removes existing files unless append=TRUE, so this unlink isn't needed
```suggestion
    utils::write.table(run_manifest_df,
                       file = manifest.file,
                       sep = ",",
                       row.names = FALSE,
                       col.names = !file.exists(manifest.file),
                       append = !overwrite)
```

base/workflow/R/run.write.configs.R
```diff
@@ -211,13 +228,26 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+
+    PEcAn.logger::logger.info("Run manifest written to ", manifest.file)
+  }
+
   options(scipen = scipen)
   invisible(settings)
   return(settings)
```
infotroph: Not your fault but fixing this while I see it -- `invisible()` was doing nothing here and usually only makes sense as a return value
```suggestion
  return(invisible(settings))
```

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,46 +2,127 @@
...
+#' @importFrom rlang %||%
 #' @export
 
 
 runModule.run.write.configs <- function(settings,
                                         overwrite = TRUE,
-                                        input_design = NULL) {
+                                        input_design = NULL,
+                                        ens_input_design = NULL,
+                                        sa_input_design = NULL) {
```
infotroph: Same comment about 3 params as for run.write.configs

modules/uncertainty/R/ensemble.R
```diff
@@ -17,36 +17,42 @@
...
+      outdir = file.path(outdir, run.id), 
+      start.year = start.year, 
+      end.year = end.year, 
+      variables = variables
+    )
+    
+    # Assign loaded variables to local environment for evaluation
+    for (var in names(out.tmp)) {
+      assign(var, out.tmp[[var]])
     }
```
infotroph, line 49: Is the evaluation feature used regularly in practice? I'm having trouble picturing where it would be needed and this (plus several other functions!) could be substantially simplified without it.

base/workflow/NEWS.md
```diff
@@ -8,6 +8,8 @@
 - **New Default**: The sampling method default has changed from "uniform" to "random"
 - **Impact**: This ensures consistency across sites in ensemble runs but may produce different results compared to previous versions
 
+* Sensitivity analysis and ensemble runs now generate separate input design matrices with appropriate dimensions, fixing dimension mismatch errors in multisite workflows. (#3708)
+* Generated runs are now stored in a `runs_manifest.csv` file in the output directory instead of modifying `samples.Rdata` (#3708)
```
infotroph, line 5: Once 1.10.0 is released this will need moving to go under the next unreleased heading, but need to wait until then

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,46 +2,127 @@
...
+        
+        num_sa_runs <- num_sa_runs + (n_traits * n_quantiles)
+      }
+
+      # Generate SA-specific input design
+      design_result_sa <- PEcAn.uncertainty::generate_joint_ensemble_design(
+        settings = settings[1],
+        ensemble_size = num_sa_runs
+      )
+      sa_input_design <- design_result_sa$X
```
infotroph: This looks to me as if it will assign a random input file to each row of the SA rather than (the desired behavior of) keeping met/IC/etc constant across OAT steps. Am I misunderstanding something here?

modules/uncertainty/R/ensemble.R
```diff
@@ -17,36 +17,42 @@
...
+  
+  for (i in seq_len(nrow(ens.run.ids))) {
+    
+    # Handle column name difference (manifest has 'run_id', legacy 'ens.run.ids' might have 'id')
+    run.id <- if ("run_id" %in% names(ens.run.ids)) ens.run.ids$run_id[i] else ens.run.ids$id[i]
 
-    for(var in seq_along(variables)){
-      out.tmp <- PEcAn.utils::read.output(run.id, file.path(outdir, run.id), start.year, end.year, variables[var])
-      assign(variables[var], out.tmp[[variables[var]]])
+    # We pass the whole 'variables' vector once, avoiding repeated file open/close.
```
infotroph, line 37: 🎉

modules/uncertainty/R/ensemble.R
```diff
@@ -17,36 +17,42 @@
...
   
   ensemble.output <- list()
-  for (row in rownames(ens.run.ids)) {
-    run.id <- ens.run.ids[row, "id"]
-    PEcAn.logger::logger.info("reading ensemble output from run id: ", format(run.id, scientific = FALSE))
+  
+  for (i in seq_len(nrow(ens.run.ids))) {
+    
+    # Handle column name difference (manifest has 'run_id', legacy 'ens.run.ids' might have 'id')
+    run.id <- if ("run_id" %in% names(ens.run.ids)) ens.run.ids$run_id[i] else ens.run.ids$id[i]
```
infotroph: Would it work to iterate over the IDs directly instead of by index?
```suggestion
  # Handle column name difference (manifest has 'run_id', legacy 'ens.run.ids' might have 'id')
  run_ids <- ens.run.ids$run_id %||% ens.run.ids$id
  for (run_id in run_ids) {
```

modules/uncertainty/R/sensitivity.R
```diff
@@ -45,34 +37,58 @@ read.sa.output <- function(traits, quantiles, pecandir, outdir, pft.name = "",
...
+        manifest$type == "Sensitivity" & 
+        manifest$pft_name == pft.name &
+        manifest$trait == trait &
+        as.character(manifest$quantile) == as.character(quantile), 
+      ]
+
+      if (nrow(subset_df) == 1) {
+         run.id <- subset_df$run_id
+      } else if (nrow(subset_df) > 1) {
+         PEcAn.logger::logger.warn("Multiple runs found for", trait, quantile, "- using the last one.")
```
infotroph, line 51: When would this happen and should it be an error instead of a warning?

modules/uncertainty/R/sensitivity.R
```diff
@@ -45,34 +37,58 @@ read.sa.output <- function(traits, quantiles, pecandir, outdir, pft.name = "",
...
+         PEcAn.logger::logger.warn("No run found in manifest for", trait, quantile)
+         next # Skip this quantile
+      }
+
+      if (is.null(run.id) || is.na(run.id)) {
+         PEcAn.logger::logger.warn("Run ID invalid or missing for", trait, quantile)
+         next
+      }
+
+      pass_pft <- switch(per.pft + 1, NULL, pft.name)
```
infotroph: Weirdly phrased -- I see this was existing code but I wonder if I'm missing a reason it wasn't `pass_pft <- if(isTRUE(per.pft)) pft.name else NULL`, which I'd find  easier to read.

modules/uncertainty/R/sensitivity.R
```diff
@@ -45,34 +37,58 @@ read.sa.output <- function(traits, quantiles, pecandir, outdir, pft.name = "",
...
+
+      # Pass ALL variables at once to avoid repeated file opening. And call read.output once
+      out.tmp <- PEcAn.utils::read.output(
```
infotroph, line 69: This might be for a different PR, but since this is getting converted to data frame a few lines down it might be worth benchmarking the existing approach against passing `dataframe = TRUE` into read.output

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,46 +2,127 @@
...
+      ensemble_size <- settings$ensemble$size %||% 1
       design_result <- PEcAn.uncertainty::generate_joint_ensemble_design(
         settings = settings,
         ensemble_size = ensemble_size
       )
-      input_design <- design_result$X
+      ens_input_design <- design_result$X
+    }
+    
+    # handle SA design - check if already provided before generating
```
infotroph: This is now turning into a fair amount of code repeated in both legs of the settings/multisettings if/else. Could we do some of this unconditionally outside the if? Or define a helper function to encapsulate it?

base/workflow/R/run.write.configs.R
mdietze: I 100% agree with @infotroph that we shouldn't have multiple input_design arguments. Where I disagree is that I don't think it makes sense to have `ensemble` and `sensitivity` elements. I think we should move to a place where a single call to run.write.configs runs a single input_design. If you want to run both an ensemble projection and a SA then you should call run.write.configs twice with different input_designs (one for SA one for ensemble). In other words, I think the SA/ensemble distinction should be deprecated at this point w.r.t. run.write.configs. For example, a Sobol SA is already actually executed via the "ensemble" pathway.

base/workflow/R/run.write.configs.R
mdietze: I should know this, but what's actually been changed in the returned settings object. I'm wondering whether the settings is still the right thing to return at this point (e.g., are we moving to a place where the manifest might be the thing a user actually wants back?)

base/workflow/R/runModule.run.write.configs.R
mdietze: I suspect that what we actually need is a deeper change, either to enable `generate_joint_ensemble_design` to generate a full OAT design (i.e. the parameter sampling should be IN the function not a prerequisite) or to produce an analogous `generate_OAT_SA_design` function. I actually favor the latter in concept unless we find a ton of code redunancies with the joint ensemble function.

base/workflow/R/run.write.configs.R
infotroph: Changed: at least ensemble IDs and PFT outdirs (if they weren't there in the input settings).

Return value: My gut is settings is still the right thing to return. I'm saying this without checking, but I think at the moment every `runModule.*()` takes a MultiSettings and returns a possibly modified MultiSettings, so changing that would be an even bigger overhaul. But maybe this is an argument for putting (a copy of) the manifest _into_ the returned settings?

base/workflow/R/run.write.configs.R
divine7022: I think here the better approach would be to always write manifest;
distinguishes "workflow completed with no runs" from "workflow crashed before writing", and let the downstream consumers handles it.

base/workflow/R/run.write.configs.R
divine7022: exactly, its defensive that adds complexity without benefit

base/workflow/R/run.write.configs.R
divine7022: > But maybe this is an argument for putting (a copy of) the manifest into the returned settings?

If you are saying (settings$runs_manifest <- run_manifest_df), then manifest written to pecan.CONFIGS.xml via listToXml(), which would serialize every row as nested XML elements. But for for large SDA runs would increase the disk size (may be in giga byte). Slow down read.settings() / write.settings()

modules/uncertainty/R/ensemble.R
divine7022: The eval(parse()) pattern is required for supporting derived variables (e.g. GPP = NPP + Ra). It's used consistently across run.ensemble.analysis.R, get.results.R, and model2netcdf.ED2.R. The apparent complexity exists to support a real feature.

modules/uncertainty/R/ensemble.R
divine7022: The index `i` is needed for output list indexing, but we could simplify by extracting run_ids once before the loop; and run_id mean the actual id's which seems confusing;
you suggested `for (run_id in run_ids)` but opted for the index-based approach because `ensemble.output[[as.character(i)]]` uses position based keys not run ids and `seq_along()` provides automatic increment without having to define another variable for manual `idx <- idx + 1`

modules/uncertainty/R/sensitivity.R
divine7022: i have added TODO comment; defer to future PR if complex filtering needed;
because i suspect keeping dataframe = FALSE make sense, cuz the current aggregation doesn't benefit from data.frame structure. If we add more complex post-processing then we can then benchmark.

base/workflow/R/runModule.run.write.configs.R
divine7022: addressed in PR #3729

base/workflow/R/run.write.configs.R
```diff
@@ -27,10 +26,14 @@
...
-run.write.configs <- function(settings, ensemble.size, write = TRUE,
+run.write.configs <- function(settings, ensemble.size, input_design, write = TRUE,
                               posterior.files = rep(NA, length(settings$pfts)),
-                              overwrite = TRUE, input_design = NULL,
-                              input_design_ens = NULL, input_design_sa = NULL) {
+                              overwrite = TRUE) {
+
+  # extract designs from input_design list
+  input_design_ens <- if (!is.null(input_design)) input_design$ensemble else NULL
+  input_design_sa <- if (!is.null(input_design)) input_design$sensitivity else NULL
```
mdietze: Per earlier comments, I'd recommend only having one input_design, not separate ens and sa designs

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,128 +2,63 @@
 #'
 #' @param settings a PEcAn Settings or MultiSettings object
 #' @param overwrite logical: Replace config files if they already exist?
-#' @param input_design the input indices for samples (DEPRECATED - use NULL)
-#' @param ens_input_design Input design matrix for ensemble (internal use)
-#' @param sa_input_design Input design matrix for SA (internal use)
+#' @param input_design Optional. Input design specification. Can be:
+#'   \itemize{
+#'     \item A list with \code{ensemble} and/or \code{sensitivity} entries
```
mdietze:
```suggestion
```

modules/uncertainty/R/ensemble.R
infotroph: Right, it's real but I wonder if we might want to kill it anyway. If GPP was a primary motivator then I note that most PEcAn models -- including ED2 -- save GPP as its own output and don't need to re-derive it at read time.

base/workflow/R/run.write.configs.R
divine7022: >If you want to run both an ensemble projection and a SA then you should call run.write.configs twice with different input_designs (one for SA one for ensemble)

https://github.com/PecanProject/pecan/pull/3708#discussion_r2613599482 according to the discussion there now run.write.configs call seperatly -- once for SA once for ens, each receiving its own input_design data frame rather than a combined list

base/workflow/R/run.write.configs.R
divine7022: >  think the SA/ensemble distinction should be deprecated at this point w.r.t. run.write.configs. For example, a Sobol SA is already actually executed via the "ensemble" pathway.

i am just curious to know -- are u visioning to deprecating the SA/ensemble distinction inside run.write.configs ( write.sa.configs and write.ensemble.configs ) ? it sounds like a vision for the future iteration; for unificaton -- would the desing could eventually encode everything needed, eliminating the need for run.write.configs to branch based on settings. happy to take this on..

mdietze: Exactly. I'm envisioning deprecating write.sa.configs first, since any SA could be done as an ensemble run with a specific design, and then probably tweaking the SA post-processing code to handle this (e.g., this may require updating exactly how those functions get the info they need and relax some hard-coded file naming conventions). 

I suspect that after we deprecate write.sa.configs that there could be a wholesale refactoring of runModule.run.write.configs, run.write.configs, and write.ensemble.configs with an eye towards simplifying the code (e.g., make input_design and a parameter matrix required inputs rather than things that can be generated automatically if they don't exist yet).

But yes, these are things for future PRs

mdietze (approved):

infotroph:
@divine7022 Can you please resolve the merge conflicts, including making sure to move the CHANGELOG entry up from (what today became) the v1.10.0 section to the new unreleased section above it?

base/workflow/NEWS.md
infotroph: 1.10.0 is released now, so it's time to move this!

infotroph (dismissed):

dlebauer (approved):

base/workflow/R/run.write.configs.R
dlebauer: Implemented

base/workflow/R/run.write.configs.R
dlebauer: implemented

modules/uncertainty/R/ensemble.R
dlebauer: not required for this PR; consider opening new issue.

modules/uncertainty/R/sensitivity.R
dlebauer: Not sure, but not required for this PR

base/workflow/R/run.write.configs.R
dlebauer: implemented

dlebauer:
[failing build](https://github.com/PecanProject/pecan/actions/runs/22413250455/job/64892065476?pr=3708#step:9:18) - looks like base/workflows/NAMESPACE needs to be re-auto-generated

```r
devtools::document('base/workflows')
```

base/workflow/NEWS.md
```diff
@@ -1,8 +1,12 @@
-# PEcAn.workflow 1.10.0.9000
+# Unreleased
```
infotroph:
```suggestion
PEcAn.workflow 1.10.0.9000
```

infotroph (changes requested):

CHANGELOG.md
```diff
@@ -12,16 +12,20 @@ For more information about this file see also [Keep a Changelog](http://keepacha
...
+- Added `generate_OAT_SA_design()` function for sensitivity analysis input design. Unlike ensemble design which randomizes all inputs, SA design holds non-parameter inputs constant to isolate parameter effects (#3729)
 - Added datasets to `PEcAn.data.land` 
   *  `landiq_crop_mapping_codes` dataset mapping LandIQ crop classification codes to human-readable crop names.
   *  `bism_kc_by_crop` dataset containing BISm crop coefficient schedules and stage timing references for use in ET estimation, including columns that map to LandIQ class and subclass.
 
-
 ### Fixed
 
 ### Changed
+- Added `generate_OAT_SA_design()` function for sensitivity analysis input design. Unlike ensemble design which randomizes all inputs, SA design holds non-parameter inputs constant to isolate parameter effects (#3729)
```
```suggestion
```

CHANGELOG.md
```diff
@@ -12,16 +12,20 @@ For more information about this file see also [Keep a Changelog](http://keepacha
...
+- Added Demo 03: Meta Analysis Quarto notebook (`documentation/tutorials/Demo_03_Meta_Analysis/meta_analysis.qmd`) to demonstrate how to perform Bayesian meta-analysis and visualize posterior distributions using pre-generated trait data.
```
infotroph: This is already present above, which seems to be a very common issue with merge-conflicted changelogs -- it's super-important to read ALL the context lines and make sure what's left after the merge makes sense as a while NEWS file, not just in GitHub's diff view.

base/workflow/NEWS.md
```diff
@@ -1,8 +1,12 @@
...
 
-* Removed `tests/Rcheck_reference.log`, which was used to ignore historic check messages that have now been fixed.
+## Changed
+* Sensitivity analysis and ensemble runs now generate separate input design matrices with appropriate dimensions, fixing dimension mismatch errors in multisite workflows. (#3708)
+* Generated runs are now stored in a `runs_manifest.csv` file in the output directory instead of modifying `samples.Rdata` (#3708)
 
+# PEcAn.workflow 1.10.0.9000
 
+* Removed `tests/Rcheck_reference.log`, which was used to ignore historic check messages that have now been fixed.
```
infotroph: .9000 means unreleased, so if a .9000 section exists you should always add entries there rather than start new sections above it.

base/workflow/R/run.write.configs.R
```diff
@@ -227,13 +254,19 @@ run.write.configs <- function(settings, ensemble.size, input_design, write = TRU
...
+                     append = !overwrite)
+  
+  PEcAn.logger::logger.info("Run manifest written to ", manifest.file)
+
   options(scipen = scipen)
-  invisible(settings)
-  return(settings)
-}
+  return(invisible(settings))
+}
```
```suggestion
}

```

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,82 +2,193 @@
...
+#' @details
+#' This function serves as the orchestration layer between PEcAn workflows and
+#' the config-writing machinery. It generates appropriate input designs
+#' (ensemble and/or SA) if not provided. For MultiSettings, it generates designs once
+#' from the first site then shares across all sites for consistent sampling. Finally,
+#' it delegates to \code{\link{run.write.configs}} for actual config generation.
+#' The input design determines how parameter samples and input files (met, soil,
+#' etc.) are coordinated across runs. Ensemble designs typically use random or
+#' quasi-random sampling, while SA designs hold non-parameter inputs constant
+#' (OAT methodology).
```
infotroph: Move up: Our convention for Roxygen order is title, description, details, params, examples, (import/export and author tags float around a bit, but always below params)
```suggestion
```

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -2,82 +2,193 @@
 #'
```
infotroph, line 2:
```suggestion
#'
#' This function serves as the orchestration layer between PEcAn workflows and
#' the config-writing machinery. It generates appropriate input designs
#' (ensemble and/or SA) if not provided. For MultiSettings, it generates designs once
#' from the first site then shares across all sites for consistent sampling. Finally,
#' it delegates to \code{\link{run.write.configs}} for actual config generation.
#' The input design determines how parameter samples and input files (met, soil,
#' etc.) are coordinated across runs. Ensemble designs typically use random or
#' quasi-random sampling, while SA designs hold non-parameter inputs constant
#' (OAT methodology).
```

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -3,17 +3,54 @@
...
+#' @param sobol Logical. If TRUE, returns a \code{sensitivity::soboljansen}
+#'   object for Sobol sensitivity analysis.
+#'
+#' @return A list containing ensemble samples and indices.
+#'   If \code{sobol = FALSE}, returns \code{list(X = design_matrix)}.
+#'   If \code{sobol = TRUE}, returns a \code{sensitivity::soboljansen()}
+#'   result object with the design matrix in \code{$X} plus additional
+#'   components for Sobol index calculations.
+#'
+#' @details
```
infotroph: Move above params

CHANGELOG.md
infotroph:
```suggestion
```

base/workflow/R/runModule.run.write.configs.R
```diff
@@ -1,83 +1,193 @@
...
+#' This function serves as the orchestration layer between PEcAn workflows and
+#' the config-writing machinery. It generates appropriate input designs
+#' (ensemble and/or SA) if not provided. For MultiSettings, it generates designs once
+#' from the first site then shares across all sites for consistent sampling. Finally,
+#' it delegates to \code{\link{run.write.configs}} for actual config generation.
+#' The input design determines how parameter samples and input files (met, soil,
+#' etc.) are coordinated across runs. Ensemble designs typically use random or
+#' quasi-random sampling, while SA designs hold non-parameter inputs constant
+#' (OAT methodology).
```
infotroph, line 11:
```suggestion
#' (OAT methodology).
#'
```

infotroph (approved):

base/workflow/NEWS.md
divine7022: addressed [04816ad](https://github.com/PecanProject/pecan/pull/3708/commits/04816ad0e06eb8fe75d9672f9f6df17e9381b7d8)

Issue #3677, opened 2025-11-25
Document joint input design


infotroph:
* Docs should explain what it should look like, what information it contains, how it is used
* mention in modules/uncertainty/NEWS.md (one sentence with relevant PR numbers)
* Mention in CHANGELOG.md (short but enough detail for users to know why it's important and what they'll need to change to use it in their workflows)

dlebauer:
will be fixed by #3660 once changes from https://github.com/divine7022/pecan/pull/1 are included

infotroph:
Also: Update examples /demos to show an explicit call to generate_join_input_design as part of the workflow script rather than hiding it inside runModule_run_write_configs (OK if this doesn't go into current release)

PR #3715, opened 2025-12-11
bump versions for modules and base

infotroph:
## Description

Part of #3679
* Bumped package versions and edited NEWS entries for all packages in base/ and modules/
* updated pecan_version_history.csv 
* Updated pecan_releases.csv assuming we'll tag the release on Friday. If not, will need to change the date here.

Not yet done but needs doing: document joint ensemble design (#3677), edits to CHANGELOG.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (approved):

base/all/NEWS.md
```diff
@@ -1,3 +1,8 @@
+# PEcAn.all 1.9.1
```
robkooper: should this be 1.10.0?

base/all/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.all
 Type: Package
 Title: PEcAn Functions Used for Ecological Forecasts and Reanalysis
-Version: 1.9.0.9000
+Version: 1.9.1
```
robkooper: do we want pecan.all to just follow the pecan release number?

modules/uncertainty/NEWS.md
```diff
@@ -1,16 +1,26 @@
-# PEcAn.uncertainty 1.8.2
+# PEcAn.uncertainty 1.9.0
```
robkooper, line 1: did we never release 1.8.2 of uncertainty?

robkooper (changes requested):
requestion some clarifications

base/all/DESCRIPTION
infotroph: I debated that and could be persuaded either way. On the one hand I like keeping the package version proportional to the semantic changes in the package itself, on the other hand PEcAn.all does get used as a shorthand for the PEcAn system and certainly always gets updated at release time (since we always have to update the version tables). 

So yes, I'm OK with making this 1.10.0 if you prefer

modules/uncertainty/NEWS.md
infotroph: No, but there's enough accumulated new functionality that it felt to me like it deserved more than a patch increment. Thoughts?

robkooper: I agree, just making sure this was not accidentally removing v1.8.2, but if it never was a version that is ok.

base/all/DESCRIPTION
infotroph: Done in 20b84cffb

base/all/NEWS.md
infotroph: fixed in 20b84cffb

robkooper (approved):

PR #3724, opened 2025-12-18
Fix regression in convert_input.R and check_missing_files.R

AritraDey-Dev:
## Description

Running a workflow that has a db connection and attempts to download met data is failing silently after the meta-analysis step with the following error:
```
 Error in sample.int(length(x), size, replace, prob) : 
  invalid first argument
Calls: <Anonymous> -> <Anonymous> -> <Anonymous> -> sample -> sample.int
```
The same issue occurs in the Docker stack tests.

This PR fixes a regression in `convert_input.R` where met file downloads were silently skipped, causing downstream sample.int errors in ` ensemble.R` when the workflow fell back to missing met data after meta-analysis. The fix relaxes an incorrect NULL check and ensures  `check_missing_files` returns a named list.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


AritraDey-Dev:
Just to add, after digging into the git history, this issue seems to have been introduced in PR https://github.com/PecanProject/pecan/pull/3338.

base/db/R/convert_input.R
```diff
@@ -489,7 +489,7 @@ convert_input <-
   # Get machine information
   machine.info <- get_machine_info(host, input.args = input.args, input.id = input.id, con = con)
 
-  if (any(sapply(machine.info, is.null))) {
```
mdietze, line 492: Not sure about this change. If there's only 1 machine info it seems like the old version should still work, but if there's more than 1 the new version will definitely fail

mdietze (commented):
I'm fine with sorting out the requested fixes, but ultimately they shouldn't be causing the issues you are discussing -- the meta-analysis code should not depend on the met data from convert.inputs in any way.

infotroph (approved):
Thanks for tracking this down!

base/db/R/convert_input.R
AritraDey-Dev: `get_machine_info` returns a single list with three named elements: list(machine = ..., input = ..., dbfile = ...). It does not return a list of multiple machine infos.

For a new input (which is the case failing here),` input.id` is NULL. In this scenario, `get_machine_info` explicitly sets input and dbfile to NULL and returns:
```
list(machine = <data.frame>, input = NULL, dbfile = NULL)
```
This is a valid state for a new file download.

However, the old check `any(sapply(machine.info, is.null)) `iterates over these three elements. Since input and dbfile are NULL, it evaluates to TRUE, incorrectly flagging this valid state as a fatal error and stopping the workflow.

The new check is.null(machine.info) is correct because `get_machine_info` returns NULL (the whole object) only when the machine lookup itself fails, which is the actual error condition we want to catch.

AritraDey-Dev:
>meta-analysis code should not depend on the met data from convert.inputs in any way.

Yes i  agree it is not a meta analysis issue,the thing is `convert_input` fails silently,so no met data is downloaded and after meta analysis is run(since it don't require met data), then it proceeds to ensemble run (here we need met data) and which is empty now,so giving `sample.int ` error.

base/db/R/convert_input.R
infotroph: The old check also behaved badly in the single null case because `sapply(NULL, is.null)` returns an empty list and then `any(list())` returns FALSE.

infotroph:
@robkooper should we pull this into release too?

PR #3725, opened 2025-12-18
Run Docker stack test on set schedule instead of each PR

infotroph:
This job pulls its Docker containers from `develop` and does not rebuild them from the PR branch, so running it for each push implied a closer connection between test output and the changes in the PR than is actually there. 

This patch stops running the job on pull requests and instead bumps up the frequency of scheduled runs -- currently to once every 3 days, but I could see moving it all the way to daily if we find that's still too much lag between introducing changes and being alerted to problems.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/docker-stack-sipnet.yml
+++ b/.github/workflows/docker-stack-sipnet.yml
@@ -5,7 +5,7 @@ on :
 
   schedule:
     # run Thursday 4:30 AM UTC
-  - cron: '30 4 * * */3'
+  - cron: '30 4 * * 1,3,5'
 jobs:
   sipnet:
     runs-on: ubuntu-latest
--- a/.github/workflows/docker-stack-sipnet.yml
+++ b/.github/workflows/docker-stack-sipnet.yml
@@ -1,13 +1,11 @@
 name : Docker Stack Model Test
 on :
-  pull_request:
-
   # allow manual triggering
   workflow_dispatch:
 
   schedule:
     # run Thursday 4:30 AM UTC
-  - cron: '30 4 * * 4'    
+  - cron: '30 4 * * */3'
 jobs:
   sipnet:
     runs-on: ubuntu-latest
--- a/.github/workflows/docker-stack-sipnet.yml
+++ b/.github/workflows/docker-stack-sipnet.yml
@@ -4,7 +4,7 @@ on :
   workflow_dispatch:
 
   schedule:
-    # run Thursday 4:30 AM UTC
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-stack-sipnet.yml
```diff
@@ -1,13 +1,11 @@
...
 on :
-  pull_request:
-
   # allow manual triggering
   workflow_dispatch:
 
   schedule:
     # run Thursday 4:30 AM UTC
-  - cron: '30 4 * * 4'    
+  - cron: '30 4 * * */3'
```
AritraDey-Dev: According to the PR description, “once every 3 days” would correspond to: `30 4 */3 * *`

The current cron: `30 4 * * */3` runs on sun, wed, and sat, which is not always once every 3 days.

Did you intend to schedule it on specific weekdays (Sun/Wed/Sat)?
If so, the current cron is correct; otherwise, the schedule may need to be adjusted.

infotroph: Good catch! Let's stick to specific days of the week but make more of them land on weekdays. How does Sunday/Tuesday/Thursday sound?
```suggestion
  - cron: '30 4 * * 0,2,5'
```

AritraDey-Dev: Yes, that works. One small nit: instead of running it on Sunday(i.e. 0) , what if we use `30 4 * * 1,3,5`, i.e., Monday, Wednesday, and Friday (all weekdays)?

infotroph: Works for me!

(I could come up with reasons to prefer one weekday over another, but we should recognize this as classic [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality). Any schedule that runs "often enough we can figure out what changed since last time" will work here.)
```suggestion
  - cron: '30 4 * * 1,3,5'
```

AritraDey-Dev (approved):

.github/workflows/docker-stack-sipnet.yml
```diff
@@ -1,13 +1,11 @@
 name : Docker Stack Model Test
 on :
-  pull_request:
-
   # allow manual triggering
   workflow_dispatch:
 
   schedule:
     # run Thursday 4:30 AM UTC
```
AritraDey-Dev: nit:
```suggestion
   # run at 4:30 AM UTC on Mon, Wed, Fri
```

PR #3727, opened 2025-12-19
db: fix convert_input and missing site.id 

AritraDey-Dev:
## Description

Encountered a few issues while running the workflow using a database connection. Below are the error logs:

```
Error in PEcAn.DB::convert_input(...) : object 'existing.input' not found
```

and 

```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'statement' in selecting a method for function 'dbGetQuery': object 'site.id' not found
Calls: <Anonymous> ... db.query -> <Anonymous> -> paste0 -> .handleSimpleError -> h
```
I will try to explain these issues with the fixes in more detail below.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/db/R/convert_input.R
+++ b/base/db/R/convert_input.R
@@ -112,6 +112,9 @@ convert_input <-
   # Forecast data sets require their own set of database checking operations, making the following else if and else irrelevant
   # for such data.  Forecast data are different if their start date (and if they are part of an ensemble, their ensemble id) are 
   # different.
+  existing.input <- NULL
+  existing.dbfile <- NULL
+  PEcAn.logger::logger.info("DEBUG: Initialized existing.input/dbfile to NULL")
   if (forecast) {
     #if the data is an ensemble, ensemble will be set equal to the number of ensemble members.
     #However, if the data is not an ensemble, ensemble will be equal to FALSE.  In order to treat ensemble and 
@@ -497,6 +500,9 @@ convert_input <-
   input <- machine.info$input
   dbfile <- machine.info$dbfile
   
+  if (is.data.frame(input)) {
+    input <- list(input)
+  }
   #--------------------------------------------------------------------------------------------------#
   # Perform Conversion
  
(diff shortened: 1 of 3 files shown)


base/db/R/convert_input.R
```diff
@@ -112,6 +112,9 @@ convert_input <-
   # Forecast data sets require their own set of database checking operations, making the following else if and else irrelevant
   # for such data.  Forecast data are different if their start date (and if they are part of an ensemble, their ensemble id) are 
   # different.
+  existing.input <- NULL
```
AritraDey-Dev, line 115:
```
Error in PEcAn.DB::convert_input(...) : object 'existing.input' not found
```
They were only defined inside the `if (forecast)` block. If forecast was FALSE, the variables were never created, causing an "object not found" error when used later in the function.

base/db/R/convert_input.R
```diff
@@ -574,7 +580,7 @@ convert_input <-
                                machine, mimetype, formatname,
                                allow.conflicting.dates, ensemble,
                                ensemble_name, existing.input,
-                               existing.dbfile, input)
+                               existing.dbfile, input, site.id)
```
AritraDey-Dev, line 582:
```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'statement' in selecting a method for function 'dbGetQuery': object 'site.id' not found
Calls: <Anonymous> ... db.query -> <Anonymous> -> paste0 -> .handleSimpleError -> h
```

base/db/R/convert_input.R
```diff
@@ -497,6 +500,9 @@ convert_input <-
   input <- machine.info$input
   dbfile <- machine.info$dbfile
   
+  if (is.data.frame(input)) {
```
AritraDey-Dev, line 502:
```
Error in input[[i]]$id : $ operator is invalid for atomic vectors
```
Downstream code expects a list so `input[[1]] ` returns a data frame. If input is a data frame, `input[[1]]` returns a column (vector), causing $id and nrow() to fail. Fixed it bt wrapping it in a list ensures the code receives a data frame.

mdietze (approved):

Issue #3718, opened 2025-12-11
Allow MA module to accept trait and prior dataframes directly


AritraDey-Dev:
> one should be able to run the MA itself without a full settings object

_Originally posted by @mdietze in https://github.com/PecanProject/pecan/pull/3707#discussion_r2598199827_

For more context https://github.com/PecanProject/pecan/pull/3707#discussion_r2602936922 & https://github.com/PecanProject/pecan/pull/3707#discussion_r2607156453

Mayanknishad9:
This is a great improvement! My `format_try_for_ma()` function (PR #3720) creates exactly the kind of `trait_data` dataframe that this refactored MA module would accept. 

Once both are implemented, users could have a clean workflow:
1. `try_data <- read.delim("TRY_export.txt")`
2. `trait_df <- format_try_for_ma(try_data, trait_mapping)`
3. `run_meta_analysis(trait_df, prior_df, config)`  This new function from #3718

This would make PEcAn's meta-analysis much more accessible to researchers with external data sources like TRY.

ashiklom:
Isn't this already possible? Here's the signature of the `PEcAn.meta.analysis::pecan.ma` function.

https://github.com/PecanProject/pecan/blob/ad821ea74a96af01eba238169de5356bdce5bcdd/modules/meta.analysis/R/meta.analysis.R#L44-L54

That said, there's a lot of extra logic in `run.meta.analysis.pft` that could be cleaned up, but my high-level recommendation would probably be to just not use that function and just use `PEcAn.meta.analysis::pecan.ma` in standalone contexts (like the Quarto notebook cited above).

ashiklom:
A more general observation: A lot of the stuff in [`run.meta.analysis.R`](https://github.com/PecanProject/pecan/blob/develop/modules/meta.analysis/R/run.meta.analysis.R) feels specific to the way full PEcAn workflows are typically executed. As a matter of organization / refactoring, I would suggest moving a lot of that functionality into something like `PEcAn.workflow` and keep the `PEcAn.meta.analysis` package specifically focused on executing meta-analysis _given trait and PFT information as data frames_.

That would also eliminate dependencies on `PEcAn.DB`, `PEcAn.settings`, and `PEcAn.utils`, which would make this much easier to install.

@mdietze -- What do you think?

mdietze:
[...]

PR #3728, opened 2025-12-22
Make it easier to run meta-analysis standalone

ashiklom:
## Description

The `PEcAn.MA::run.meta.analysis.pft` function has a lot of file I/O and some database interactions that make it difficult to run without the full PEcAn infrastructure. At the same time, `PEcAn.MA::pecan.ma` is missing some useful pre- and post- checking and data processing. This introduces an intermediate function --- `meta_analysis_standalone` --- that includes the logic of `run.meta.analysis.pft` but without the extra I/O and database interactions.

`run.meta.analysis.pft` has been refactored to be a thin wrapper around `meta_analysis_standalone` along with some file I/O. A minor breaking change is that *all* file writing now happens _after_ `meta_analysis_standalone` runs (e.g., before, `jagged.data` was written to disk right after being created but before the meta-analysis was executed; now it's saved after the entire meta-analysis runs).

I also added a short new vignette that shows how this function can be used. The Quarto notebook in #3707 can be a more complete version of this.

## Motivation and Context

See discussion in #3718. Also related to open PR #3707.

## Review Time Estimate
- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:
- [X] My name is in the list of CITATION.cff
- [X] I agree that PEcAn Project may distribute my contribution under any or all of
  - The new vignette sort-of functions as a mini integration test.
  - NOTE: I ran the PEcAn.MA tests, but not the full suite of PEcAn tests. I was hoping the CI could do that for me; otherwise, if there's a quick set of tests I can run on my local machine, let me know what the current best practice is for doing that.


modules/meta.analysis/R/assign.treatments.R
```diff
@@ -3,9 +3,14 @@
...
 ##' The algorithm (incorrectly) assumes that each site has a unique set of experimental
-##' treatments. This assumption is required by the data in BETYdb that does not always consistently name treatments or quantity them in the managements table. Also it avoids having the need to estimate treatment by site interactions in the meta analysis model. This model uses data in the control treatment to estimate model parameters so the impact of the assumption is minimal.
-##' @name assign.treatments
-##' @title assign.treatments
+##' treatments. This assumption is required by the data in BETYdb that does not
+##' always consistently name treatments or quantity them in the managements
+##' table. Also it avoids having the need to estimate treatment by site
+##' interactions in the meta analysis model. This model uses data in the control
+##' treatment to estimate model parameters so the impact of the assumption is
+##'
```
mdietze: I'm assuming this blank line is accidental?

mdietze (approved):

modules/meta.analysis/R/assign.treatments.R
ashiklom: Yup, sorry. Error in automatic comment reformatting.

ashiklom:
I introduced at least one bug in here, caught by the CI. So hold off on merging this.

I'll get that fixed and make sure the tests pass.

modules/meta.analysis/NAMESPACE
```diff
@@ -1,7 +1,9 @@
 # Generated by roxygen2: do not edit by hand
 
 export(approx.posterior)
+export(assign.treatments)
```
infotroph: Nit: Would it make sense to rename this `assign_treatments` as long as we're moving it, to avoid any confusion from looking like as S3 method? (Not that `base::assign` seems all that likely to become a generic...)

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -1,4 +1,190 @@
...
+    jagged_data,
+    function(x) mean(x[["Y"]], na.rm = TRUE),
+    numeric(1)
+  )
+
+  ## Set gamma distribution prior
+  prior_variances <- as.data.frame(rep(1, nrow(priors)))
+  row.names(prior_variances) <- row.names(priors)
+  prior_variances[names(trait_average), ] <- 0.001 * trait_average ^ 2
+  prior_variances["seedling_mortality", 1] <- 1
```
infotroph, line 107: Would it be possible to add a few comments here explaining (1) where the constants 0.001 and 1 come from, and (2) why seedling mortality gets special handling and whether that's a hack or a principled design choice?

infotroph: On scrolling down further I see this is pre-existing code moved up from `run.meta.analysis.pft`, so maybe these explanations are lost to time...

modules/meta.analysis/NAMESPACE
ashiklom: Addressed in 4316360.

modules/meta.analysis/R/run.meta.analysis.R
ashiklom: Yeah, this is pre-existing code. The oldest git blame gets me is here:

https://github.com/PecanProject/pecan/blame/bed27a2b37619c2297c3c66becf97b8c16233494/modules/meta.analysis/R/run.meta.analysis.R

...but I'm almost certain this logic predates that, which looks more like refactoring than development.

The more correct thing to do here is probably (1) remove the `seedling_mortality` hard-coded value altogether and replace it with a global default prior of some kind in the database; (2) pass the `0.001` as an (optional) argument called something like `default_prior_variance_fraction` with 0.001 as the default value. But I'm hesitant to tie those changes to this refactor --- probably better to submit a dedicated PR on that. (In general, in my experience, it's much easier to review a larger number of smaller PRs with each PR having a specific objective).

infotroph: Filed for later as #3736

PR #3729, opened 2025-12-22
Add generate_OAT_SA_design() for sensitivity analysis input design

divine7022:
originally discussed -- [comment](https://github.com/PecanProject/pecan/pull/3708/changes#r2621323517) 

## Description
Add `generate_OAT_SA_design()` function to create input design matrices for OAT sensitivity analysis. 
SA requires isolating the effect of each parameter by holding all other inputs (met, IC, soil) constant. The existing `generate_joint_ensemble_design()` randomizes these inputs, which is correct for ensemble runs but invalidates SA variance decomposition.

added tests in `test.input_design.R` to validate the OAT sensitivity design, including checks :
- that the total number of runs is calculated correctly
- all non-parameter inputs (such as met, IC, and soil data) remain constant across OAT steps as required for valid SA
- comparison test showing the key difference from ensemble design

## Review Time Estimate

- [X] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/uncertainty/R/generate_OAT_SA_design.R
```diff
@@ -0,0 +1,123 @@
+#' Generate One-At-a-Time (OAT) input design for sensitivity analysis
+#'
+#' Creates an input design matrix for sensitivity analysis where non-parameter
+#' inputs (met, IC, soil, etc.) are held constant while parameters vary
+#' one-at-a-time across quantiles. This differs from ensemble design where
+#' all inputs vary together.
+#'
+#' @param settings PEcAn settings object
+#' @param sa_samples Optional. Pre-loaded SA samples from samples.Rdata.
```
mdietze: This argument suggests a fundamental design misunderstanding. sa_samples should be *generated by* this function, not read by it. I'm OK with taking this as an optional input if one wants to reuse a previous sa_samples, but if it's not provided it needs to be generated

modules/uncertainty/R/generate_OAT_SA_design.R
```diff
@@ -0,0 +1,123 @@
...
+#' @export
+#' @author Akash B V
+#' @importFrom rlang %||%
+generate_OAT_SA_design <- function(settings, sa_samples = NULL) {
+  
+  # load SA samples if not provided
+  if (is.null(sa_samples)) {
+    samples_file <- file.path(settings$outdir, "samples.Rdata")
+    
+    # generate samples if they don't exist (safety fallback)
```
mdietze: This should be the default, not the safety fallback

modules/uncertainty/R/generate_OAT_SA_design.R
```diff
@@ -0,0 +1,123 @@
...
+    if (!file.exists(samples_file)) {
+      posterior.files <- settings$pfts %>%
+        purrr::map_chr("posterior.files", .default = NA_character_)
+      ens.sample.method <- settings$ensemble$samplingspace$parameters$method %||% "uniform"
+      
+      PEcAn.uncertainty::get.parameter.samples(
+        settings,
+        ensemble.size = 1,  # SA doesn't need ensemble samples
```
mdietze: not needed as this is the default

modules/uncertainty/R/generate_OAT_SA_design.R
```diff
@@ -0,0 +1,123 @@
...
+        posterior.files,
+        ens.sample.method
```
mdietze: also not needed as this isn't an ensemble sample

modules/uncertainty/R/generate_OAT_SA_design.R
```diff
@@ -0,0 +1,123 @@
...
+#' # 1     1   1  1    1   # Median run
+#' # 2     2   1  1    1   # trait1 @ q=2.3%
+#' # 3     3   1  1    1   # trait1 @ q=15.9%
+#' # 4     4   1  1    1   # trait1 @ q=84.1%
+#' # ...
+#' }
+#' @export
+#' @author Akash B V
+#' @importFrom rlang %||%
+generate_OAT_SA_design <- function(settings, sa_samples = NULL) {
```
mdietze, line 77: It might be too much at this PR, but it would be good to move away from `settings` as a dependency unless the underlying number of pieces of information is too large to meaningfully pass to the function. But in that case it would be good to document exactly what part of the settings is required. Here I think it might just be settings$outdir, settings$pfts, settings$sensitivity.analysis, and settings$ensemble (i.e. I wonder if you could get away with a function that has outdir, pfts, sensitivity.analysis, ensemble, and sa_samples as arguments?). Would also be good to better document semi-hidden dependencies (e.g. what does the pft$posterior.files need to point to for the function to actually sample parameters correctly)

mdietze (commented):
A good overall test that this function is working correctly would be to verify that the run.write.configs module runs correctly with the OAT design as an input and continues to generate output that's readable by the SA postprocessing and graphing functions. If it doesn't, then this function is generating inputs that are not serving any useful purpose. Along the way, we may want/get to simplify the run write configs logic

modules/uncertainty/R/generate_OAT_SA_design.R
divine7022: I agree with reducing dependency of settings for input desing by passing specific argument, but after analyzing both functions to keep the architecture consistent, i found that the parameter requirements are not same; `generate_OAT_SA_design` needs outdir, pfts, samplingspace, sensitivity.analysis and `generate_joint_ensemble_design` additionally needs _run$inputs_ (via input.ens.gen() which samples from `settings$run$inputs[[input]]$path`).
However there is a deeper blocker -- both functions call `get.parameter.samples(settings, ...)` which itself uses many settings fields (database$bety, host$name, sensitivity.analysis, etc.). So even with explicit parameters in the design functions, we'd still pass settings through to get.parameter.samples. And then that involves refactoring SDA and sobol callers.
anyways i have documented what setting it uses and semi-hidden dependiences in both desing function. I happy to know ur thoughts.

modules/uncertainty/R/generate_OAT_SA_design.R
divine7022: the parameter now clearly indicates it's optional for reusing existing samples, not the expected input

modules/uncertainty/R/generate_OAT_SA_design.R
divine7022: removed "safety fallback" comment

modules/uncertainty/R/generate_OAT_SA_design.R
divine7022: removed, good catch!

modules/uncertainty/R/generate_OAT_SA_design.R
mdietze: I think it's fine if the parameters requirements are not the same. I also think it's fine to push the refactor of the generate design functions and get.parameter.samples to a future PR

mdietze (approved):

mdietze:
── Failed tests ────────────────────────────────────────────────────────────────
  Error ('test.input_design.R:165:3'): OAT design integrates with write.sa.configs for SA postprocessing
  Error in `PEcAn.uncertainty::write.sa.configs(defaults = settings$pfts, 
      quantile.samples = sa_samples, settings = settings, model = "FAKE", 
      write.to.db = FALSE, input_design = input_design)`: unused argument (input_design = input_design)

divine7022:
>PEcAn.uncertainty::write.sa.configs(defaults = settings$pfts,  quantile.samples = sa_samples, settings = settings, model = "FAKE",  write.to.db = FALSE, input_design = input_design): unused argument (input_design = input_design)

we are passing `input_design` as a new parameter to write.sa.configs() in #3708 -- all tests are passed when i branched that PR. 
This will fix when we merge #3708

divine7022:
https://github.com/PecanProject/pecan/actions/runs/20528083080/job/58974923917?pr=3729

`modules/meta.analysis/vignettes/meta-analysis-demo.Rmd` -- failing because `use()` 
it seems like this PR induced this failure -- #3728 ,
is there any reason to use `use()` here instead of library() ?

modules/uncertainty/R/generate_OAT_SA_design.R
```diff
@@ -0,0 +1,151 @@
+#' Generate One-At-a-Time (OAT) input design for sensitivity analysis
+#'
+#' Creates an input design matrix for sensitivity analysis where non-parameter
+#' inputs (met, IC, soil, etc.) are held constant while parameters vary
+#' one-at-a-time across quantiles. This differs from ensemble design where
+#' all inputs vary together.
+#'
+#' @param settings PEcAn settings object. This function directly uses:
```
infotroph: Consider "see details" and moving the itemized list down? I find this long a `@param` entry makes it hard to skim all the options.

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+}
+
+make_mock_sa_samples <- function() {
+  list(
+    pft1 = structure(
+      matrix(1:9, nrow = 3, ncol = 3),
+      dimnames = list(c("25", "50", "75"), c("trait1", "trait2", "trait3"))
+    )
+  )
+}
```
infotroph: Returns a constant -- why is this a function and not just a list?

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
+make_sa_test_settings <- function() {
+  list(
+    outdir = withr::local_tempdir(),
```
infotroph: This path looks to me as if it will stop existing when make_sa_test_settings returns! Is the outdir used for anything during testing or could this just be `"/fake/output/path/"` and skip the withr usage?

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+test_that("generate_OAT_SA_design keeps non-param columns constant at 1", {
+  settings <- make_sa_test_settings()
+  sa_samples <- make_mock_sa_samples()
+  
+  result <- generate_OAT_SA_design(settings, sa_samples = sa_samples)
+  
+  non_param_cols <- setdiff(names(result$X), "param")
+  for (col in non_param_cols) {
+    expect_true(all(result$X[[col]] == 1),
+      info = paste("Column", col, "should be constant 1 for SA"))
```
infotroph: `?expect_true` discourages using `info` in new code, and the expectation seems pretty clear without it to me. Could unquote for slightly better diagnostic messages:
```suggestion
   # all columns but `param` should be a constant 1
   # (`!!` to get the column name into the failure message)
    expect_true(all(result$X[[!!col]] == 1))
```

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+  }
+})
+
+test_that("generate_OAT_SA_design param column is sequential", {
+  settings <- make_sa_test_settings()
+  sa_samples <- make_mock_sa_samples()
+  
+  result <- generate_OAT_SA_design(settings, sa_samples = sa_samples)
```
infotroph: Style nit: Since generating the design takes several lines of setup and is identical for both these tests, multiple expectations in the same test block feels cleaner to me.
```suggestion
```

(If you accept this, probably want to edit line 38 to something like "...keeps param column sequential and others constant at 1")

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
+make_sa_test_settings <- function() {
```
infotroph: Please match test file names to R file names by naming this file  `test-generate_OAT_SA_design.R`, to avoid needing to wonder which test file to look in to go along with the code you're writing (or vise versa).

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+test_that("OAT design has constant inputs while ensemble design varies them", {
+  settings <- make_sa_test_settings()
+  sa_samples <- make_mock_sa_samples()
+  sa_result <- generate_OAT_SA_design(settings, sa_samples = sa_samples)
+  
+  sa_non_param <- setdiff(names(sa_result$X), "param")
+  for (col in sa_non_param) {
+    expect_equal(length(unique(sa_result$X[[col]])), 1,
+      info = "SA design: non-param columns must be constant")
+  }
```
infotroph: Isn't this the same condition you already tested around line 45?

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+  
+  ## ensemble design - non-param can vary (mocked to show variation)
```
infotroph: I don't follow what "to show variation" means here and why it needs all the mocks

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+  runs_file <- file.path(rundir, "runs.txt")
+  expect_true(file.exists(runs_file))
+  run_ids <- readLines(runs_file)
+  expect_equal(length(run_ids), nrow(input_design))
+  
+  # verify run directories created (required for model output reading)
+  for (run_id in run_ids) {
+    expect_true(dir.exists(file.path(rundir, run_id)))
+  }
+})
```
infotroph:
```suggestion
})

```

modules/uncertainty/tests/testthat/test.input_design.R
```diff
@@ -0,0 +1,194 @@
...
+  ens_result <- generate_joint_ensemble_design(settings, ensemble_size = 5)
+  
+  expect_true(length(unique(ens_result$X$met)) > 1,
+    info = "Ensemble design: non-param columns should vary")
+})
+
+#------------------ tests: OAT design with write.sa.configs -------------------
+# verifies design produces output compatible with SA postprocessing
+
+test_that("OAT design integrates with write.sa.configs for SA postprocessing", {
```
infotroph: The degree of complication needed to set up these tests feels like an indicator that we could do better at designing these functions for testability, but I think that can be a future project.

infotroph (changes requested):

modules/uncertainty/tests/testthat/test.input_design.R
divine7022: thanks, yes never accessed, when i see the settings parameter requirements for generate_OAT_SA_desing; 
used "/fake/output/path/" and commented

modules/uncertainty/tests/testthat/test.input_design.R
divine7022: yup, unified test "generate_OAT_SA_design keeps param _sequential_ and _non-param constant at 1_"

modules/uncertainty/tests/testthat/test.input_design.R
divine7022: yeah, this is intentional, to check in the comparison test of ens and SA design ( now simplified and with clear intent )

modules/uncertainty/tests/testthat/test.input_design.R
divine7022: the mock returns varied indices c(1, 2, 3, 1, 2) to demonstrate that the ensemble design structure passes through whatever input.ens.gen returns, unlike the OAT design which forces all non-param columns to constant 

 and now made the comment much clear

modules/uncertainty/tests/testthat/test.input_design.R
divine7022: the file tests both generate_OAT_SA_design and generate_joint_ensemble_design (plus their comparison, etc), naming it test-generate_OAT_SA_design.R would suggest it only tests SA design. so i kept test.input_design.R to reflect that it covers both design types

divine7022:
@infotroph please take a look, once you approved i will merge this branch with https://github.com/PecanProject/pecan/pull/3708
since both PRs the depended on merging each other

modules/uncertainty/tests/testthat/test.input_design.R
infotroph: Then please move the tests of generate_joint_ensemble_design to their own file, and if there are shared helpers then put them in a file whose name starts with `helpers`.

See https://r-pkgs.org/testing-basics.html for more on recommended organization, but know that the reason I'm insisting on this is that unit testing is built on boring consistency, and inconsistent test file names _will_ confuse future maintainers (including me).

divine7022: thanks for pointing this out, agreed!

divine7022:
heads-up : `generate-sa-design` branch is already merged with `run-manifest` https://github.com/PecanProject/pecan/pull/3708 

leaving this PR open until we merge `run-manifest` branch

PR #3742, opened 2026-01-01
CI: add meta-analysis notebook and modularize rendering workflow

AritraDey-Dev:
## Description

This PR adds the newly introduced meta-analysis notebook to the CI rendering workflow to ensure it compiles successfully after https://github.com/PecanProject/pecan/pull/3707.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


.github/notebooks.yaml
```diff
@@ -0,0 +1,4 @@
+notebooks:
```
infotroph: - Nitpick: Can we rename this `notebooks.yml` instead of mixing "yaml" and "yml" extensions?
- More substantive: Am I right this is only used by reading it into `workflows/render-quarto.yml`? If so, why not define it directly in that file[*]? Feels confusing to have the list of notebooks live in a third place that's neither alongside the files nor alongside the code that uses them.

[*]possibly it could be specified as a matrix that could then be processed in parallel? I haven't looked carefully, so may be missing a reason that wouldn't work

AritraDey-Dev: > If so, why not define it directly in that file[*]? 

Yes, it is used by the render-quarto workflow, so it can definitely be kept in that file itself. will do that.

AritraDey-Dev: > [*]possibly it could be specified as a matrix that could then be processed in parallel? I haven't looked carefully, so may be missing a reason that wouldn't work

I've considered this before.I am slightly against this, particularly for this workflow, since we only use parallel tasks when they are long-running. In this case, rendering demo 1 takes approximately 20 seconds, and the meta-analysis takes about the same time. The uncertainty notebook, however, takes longer which is around 150 seconds.
Considering these durations, if we run them in parallel, we first need to build the base image and then render the notebooks. For each parallel matrix job, additional time is spent pulling the image and setting up the job, which will defintely end up taking more total time than the actual rendering.
Also after rendering the notebooks, we have to merge the artifacts in order to commit the changes to the documentation repository, which again adds more overhead compared to what we are doing atm.So, do you think we should consider this approach?

infotroph: 👍 No need to consider further! "setup and postprocessing make sense to do once and they take longer than rendering" is enough reason by itself to to not bother with parallel yet (or possibly ever).

.github/workflows/render-quarto.yml
```diff
@@ -28,13 +28,32 @@ jobs:
...
+          r_files_raw=$(echo "$paths" | awk 'NF {printf "\\\"/work/%s\\\", ", $0}')
+          r_files=${r_files_raw%, }
+          render_cmd_raw=$(echo "$paths" | awk 'NF {printf "quarto render /work/%s --to html && ", $0}')
+          render_cmd=${render_cmd_raw% && }
+          paths_out=$(echo "$paths" | tr '\n' ' ')
+          
+          echo "r_files=$r_files" >> $GITHUB_OUTPUT
+          echo "render_cmd=$render_cmd" >> $GITHUB_OUTPUT
+          echo "paths=$paths_out" >> $GITHUB_OUTPUT
+
```
infotroph: Will it stay true that we want this job to render and upload every Quarto notebook in the tutorials directory? If so I think you could remove this entire block *and* avoid needing to update this file in the future by using shell wildcards below: `quarto render /work/documentation/tutorials/**/*.qmd --to html`, `for p in documentation/tutorials/**/*.qmd; do`, etc.

infotroph: (Obviously this suggestion is completely incompatible with my retracted parallel rendering idea!)

AritraDey-Dev: Yes, that's correct,we want to render all `.qmd` files in the tutorials directory, assuming future notebooks will also be added there.
Note that if we ever decide to add other quarto notebook in other directories (e.g., inside specific packages) and want to render them in CI, we would need to update the discovery logic, but for now, this covers everything in `documentation/tutorials`.

AritraDey-Dev: >Artifact download URL: https://github.com/PecanProject/pecan/actions/runs/20835708283/artifacts/5070021922

Looking at the artifact, the implementation seems correct.

infotroph (approved):

PR #3750, opened 2026-01-05
CI failure in check_modules( R 4.2)

AritraDey-Dev:
## Description

Fix CI failure in `meta.analysis` vignette by replacing `use()` with `library()`.
https://github.com/PecanProject/pecan/actions/runs/20719742556/job/59479981701

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/meta.analysis/vignettes/meta-analysis-demo.Rmd
+++ b/modules/meta.analysis/vignettes/meta-analysis-demo.Rmd
@@ -8,8 +8,8 @@ vignette: >
 ---
 
 ```{r}
-use("PEcAn.MA")
-use("tibble", c("tribble"))
+library("PEcAn.MA")
+library("tibble")
 
 set.seed(20251222)
 ```
--- a/modules/meta.analysis/vignettes/meta-analysis-demo.Rmd
+++ b/modules/meta.analysis/vignettes/meta-analysis-demo.Rmd
@@ -9,7 +9,7 @@ vignette: >
 
 ```{r}
 library("PEcAn.MA")
-library("tibble")
+library("tibble", include.only = "tribble")
 
 set.seed(20251222)
 ```
--- a/docker/depends/pecan_package_dependencies.csv
+++ b/docker/depends/pecan_package_dependencies.csv
@@ -651,6 +651,7 @@
 "tibble","*","models/lpjguess","Imports",FALSE
 "tibble","*","modules/data.atmosphere","Imports",FALSE
 "tibble","*","modules/data.remote","Suggests",FALSE
+"tibble","*","modules/meta.analysis","Suggests",FALSE
 "tictoc","*","modules/assim.sequential","Suggests",FALSE
 "tidyr","*","base/db","Imports",FALSE
 "tidyr","*","models/ed","Imports",FALSE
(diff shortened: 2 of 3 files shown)


mdietze (approved):

modules/meta.analysis/vignettes/meta-analysis-demo.Rmd
```diff
@@ -8,8 +8,8 @@ vignette: >
 ---
 
 ```{r}
-use("PEcAn.MA")
-use("tibble", c("tribble"))
+library("PEcAn.MA")
+library("tibble")
```
infotroph: `use` does seem useful here! What if we condition on the R version?
cc @ashiklom in case I'm missing gotchas
```suggestion
if (getRversion() < "4.4") {
  use <- function(...) library(..., attach.required = FALSE)
}
use("PEcAn.MA")
use("tibble", include.only = "tribble")
```

AritraDey-Dev: Not sure what the specific benefit of use() would be here, but isn't `library()` the standard way to load pkgs in a vignette? Also, `use()` was only added in newer R versions

divine7022: yeah, had a suspicion and flagged it earlier here --- https://github.com/PecanProject/pecan/pull/3729#issuecomment-3693303874  😄

infotroph: > Not sure what the specific benefit of use() would be

Easier control over the environment and better clarity of intention, especially when you want one function but specifically not others from the same package. This particular vignette _could_ get by without it, but I endorse keeping it because habitually  using the `use(<pkg>, <fns>)` form means you're also providing a nice easy-to-read answer to two very common questions from readers of R code: "Why are we loading `<pkg>` here anyway?" and "which package did `<function>` come from?"

> use() was only added in newer R versions

Which is why I'm suggesting a fallback for older systems. Is this a problem?

ashiklom: Meh, if we want to support older R versions, I would just stick with `library`, possibly `library(..., include.only = c(...))` which goes back at least a few versions (I remember it appearing around 2019 when I was at PNNL). Also, `library` is such a widely used R idiom that the bar for using something else is relatively high.

Moreover, a vignette feels like the wrong place to introduce any kind of complex version checking logic; better to keep vignettes as simple as possible to avoid confusing users.

I added `use` because I saw it recommended... somewhere (possibly an LLM? Can't recall now) and thought "hey, that's neat!", but I didn't realize just how new it was.

Of course, [*real* programmers](https://xkcd.com/378/) use [`box`](https://klmr.me/box/)... 😉

ashiklom: To be more specific, you should be able to replace the second "use" with:

```r
library(tibble, include.only = "tribble")
```

I believe that goes back to R 3.6.0 (2019).

infotroph:
```suggestion
library("PEcAn.MA")
library("tibble", include.only = "tribble")
```

PR #3751, opened 2026-01-05
Add pkgdown links for PEcAn functions

### LINKED ISSUE #3745: Add cross-references to pkgdown package documentation from the book
## Description

The book mentions many PEcAn functions and packages but does not link to the rendered pkgdown documentation where users can find detailed API reference (parameters, return values, examples). The pkgdown documentation exists at https://pecanproject.github.io/package-documentation/develop/ but is underutilized.

The book should focus on **explanation** (concepts, workflows, how things fit together), while **referencing** pkgdown for detailed function documentation. Currently function names appear as plain text instead of clickable links.

**Current state**

**Functions mentioned without links to pkgdown:**

`book_source/03_topical_pages/03_pecan_xml.Rmd`:
```
- `PEcAn.DB::db.open`
- `PEcAn.workflow::start_model_runs`
- `PEcAn.uncertainty::write.ensemble.configs`
- `PEcAn.uncertainty::run.ensemble.analysis`
```
`book_source/03_topical_pages/95_remote_execution.Rmd`:
```
The `PEcAn.remote::remote.execute.cmd`
...see the `PEcAn.remote` package documentation.
```
(Says "see the package documentation" but doesn't provide a link)

`09_rabbitmq.Rmd:`
`PEcAn.remote::rabbitmq_post_message` function

**Vignettes linked to raw github Rmd instead of rendered html:**
`01_introductions_user.Rmd` has a table with vignettes
The "html" column is empty, but pkgdown renders these vignettes to html. These could link to the rendered versions at
e.g:
- https://pecanproject.github.io/package-documentation/develop/PEcAn.allometry/articles/AllomVignette.html
- https://pecanproject.github.io/package-documentation/develop/PEcAn.assim.batch/articles/AssimBatchVignette.html

**"Under construction" sections that could link to packages:**
`03_pecan_xml.Rmd` -- These mention packages but don't link to their pkgdown documentation.

**Proposed approach:**

1. Convert plain text function references to hyperlinks pointing to pkgdown reference pages
    - e.g. : `PEcAn.DB::db.open` --> [PEcAn.DB::db.open](https://pecanproject.github.io/package-documentation/develop/PEcAn.DB/reference/db.open.html)
2. Link "see the package" statements to actual pkgdown pages
   - e.g. : "see the PEcAn.remote package documentation" -- link to the package index
3. Fill in rendered html links for vignettes table
   - the vignette table has an "html" column that's currently empty
   - these should link to the pkgdown rendered versions
4. Replace "under construction" placeholders with package links
   - if detailed content isn't in the book, at least provide a link to pkgdown
5. Audit the book to identify other places where package names or functions are mentioned and add links to the appropriate pkgdown pages where a reference already exists.

man080107:
This PR replaces plain-text references to PEcAn functions and packages in the book with hyperlinks to the existing pkgdown documentation.

The goal is to keep the book focused on concepts and workflows while directing readers to pkgdown for detailed API references.

Changes are limited to documentation only and do not affect code execution.

Fixes #3745

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -178,7 +178,10 @@ In other words, this configuration...
   </database>
 ```
 
-...will be translated (by way of `PEcAn.DB::db.open(settings$database$bety)`) into R code like the following:
+... will be translated (by way of
+[`PEcAn.DB::db.open`](https://pecanproject.github.io/package-documentation/develop/PEcAn.DB/reference/db.open.html)
```
infotroph, line 182: I'm hoping we can find a way to avoid hardcoding the `.../develop/...` here -- ideally the links from the v1.9.0 book should lead to the v1.9.0 package docs, develop book to develop package docs, etc. See https://github.com/PecanProject/pecan/issues/3745#issuecomment-3712541191 for more about my reasoning.

infotroph (changes requested):

man080107:
To avoid hard-coding /develop/, should I use relative links, or is there an existing version-aware approach you’d like me to follow?

infotroph:
In principle I like relative links, but unfortunately I don't think that solves the problem -- the way we're currently nesting tags inside documentation types (`/documentation/[tag]/foo.html`, `/package-documentation/[tag]/[pkgname]/reference/bar.html`) means you need to know the tag whether the link is relative or absolute. I'm not sure what the right fix is; I'll keep thinking but suggestions welcome from anyone.

modules/data.atmosphere/README.md
```diff
@@ -26,7 +26,7 @@ The PEcAn.data.atmosphere package is 'standalone'.
 
 ## Documentation
 
-* [Package Documentation](https://pecanproject.github.io/package-documentation/develop/data.atmosphere/index.html)
+* [Package Documentation](../../../../package-documentation/PEcAn.DB/reference/index.html)
```
infotroph: R package READMEs appear in many contexts other than the pkgdown website (on GitHub, in the user's file browser/text editor, etc), so we can't assume a relative link here. 
```suggestion
* [Package Documentation](https://pecanproject.github.io/package-documentation/develop/data.atmosphere/index.html)
```

shreyannandanwar:
@infotroph Can we generate the pkgdown links at build time via inline R to avoid hard-coded URLs.
This will preserve version-correctness and will be offline-safe documentation.

infotroph:
> generate the pkgdown links at build time via inline R

That would be technically possible, but I'm not aware of any other places in the book that we run inline R code, and to me it feels like more complexity than is warranted. 

I see four possible approaches here:

1. Do nothing at all and leave function names unlinked, trusting that readers can figure out how to look them up in whichever version of the documentation they prefer. This one has my vote so far, but maybe I'm being a stick-in-the-mud.
2. Link to the one specific version of the package docs as the issue originally proposed, trusting that readers can figure out what to do if the page they land on doesn't match the book version they left from. If we do this I'd weakly prefer the links point to the current stable release (`/pkg-documentation/latest/...`) instead of develop.
3. Rewrite links at book compilation time, either via inline R code or via some kind of separate preprocessing step. This would add complexity both in the compilation script and for everyone editing the book ("How do I write a link again? Which magic variables does it need? Why does it have to be different in the book vs the package vignette?") and I don't think it's worth it.
4. Restructure one or both documentation sites so that relative links work, either by changing the folder structure (which would probably break lots of existing links/bookmarks) or by implementing some kind of redirection scheme (for which I believe the options inside GitHub Pages are limited). I suspect this one isn't worth it either, but may be missing something that makes it simpler than I think.

mdietze:
I vote for #2 now but wouldn't object to trying #3. For 2, I'd argue for pointing to develop not main, otherwise there will be issues that crop up (Eg documentation for a new function that's not in "latest" yet would be a dead link)

man080107:
Thanks for the detailed discussion — that clarifies the trade-offs really well.
Based on this, I’ll proceed with option #2, keeping links simple and pointing to the develop package documentation so newly added functions don’t result in dead links. I’ll avoid inline R or preprocessing for now to keep the book easy to maintain, but I’m happy to revisit option #3 later if it turns out to be worthwhile.

infotroph:
> I'd argue for pointing to develop not main, otherwise there will be issues that crop up

I foolishly split this discussion between the issue thread and the PR thread, but in https://github.com/PecanProject/pecan/issues/3745#issuecomment-3712541191 I argued the opposite direction: If a function is removed from develop but still present in main, then the dead link is in the released manual instead of the development one. I recognize we add functions more often than we remove them, but am not sure which one is the worse disruption to the reader. 

Bigger picture I'm still more concerned about the potential confusion from links that _do_ resolve (and therefore move the reader between release contexts with little or no warning), than I am about dead links; There's no fixed branch that can eliminate that but I guess we can minimize it by pointing at the branch we expect the most people to be reading the book from -- Will that be develop?

mdietze:
I expect most folks use Develop. If we want data you could drop a quick poll into the PEcAn Slack GSOC channel, which has a whole pile of new users.

To simplify all of this, I'll add that I also don't think we need to continue to maintain so many archived copies of the documentation. Each version of PEcAn contains sufficient info to completely re-create the bookdown and pkgdown for that specific instant in time.

man080107:
Thanks for expanding on this — that makes the underlying trade-off much clearer. I agree that reader context-switching is the bigger concern than occasional dead links, and that there’s no perfect branch choice that avoids both. Given that, I’m happy to stick with pointing to develop for now if that’s where we expect most readers to be, and adjust if the consensus shifts. Let me know if you’d like me to make any changes based on a poll or a final decision.

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -178,7 +178,10 @@ In other words, this configuration...
...
+(`settings$database$bety`)) into R code ...
```
infotroph:
```suggestion
(`settings$database$bety`)) into R code like the following:
```

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -206,7 +209,9 @@ The default `config.php` settings on any given platform (VM, Docker, etc.) or in
 
 Key R functions using these parameters are as follows:
 
-- `PEcAn.DB::db.open` -- Open a database connection and create a connection object, which is used by many other functions for communicating with the PEcAn database.
+- [`PEcAn.DB::db.open`](https://pecanproject.github.io/package-documentation/develop/PEcAn.DB/reference/db.open.html)
+  —-Open a database connection and create a connection object
+,   which is used by many other functions for communicating with the PEcAn database.
```
infotroph:
```suggestion
- [`PEcAn.DB::db.open`](https://pecanproject.github.io/package-documentation/develop/PEcAn.DB/reference/db.open.html)
  -- Open a database connection and create a connection object,
  which is used by many other functions for communicating with the PEcAn database.
```

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -263,7 +268,9 @@ The PEcAn system requires at least 1 plant functional type (PFT) to be specified
 
 This information is currently used by the following PEcAn workflow function:
 
-- `get.traits` - ??????
+- - [`PEcAn.workflow::get.traits`](https://pecanproject.github.io/package-documentation/develop/PEcAn.workflow/reference/get.traits.html)
+  — Retrieves trait data for selected PFTs from the database or posterior files
```
infotroph: Can you please explain how you decided to use this link, which does not exist? Remember that the task is not just to  add link-shaped text but to verify that the pages you link to are real and appropriate for the context.

I suspect linking to `PEcAn.DB::get.trait.data` will be the most useful option here, but hard to say since it's not very clear what the previous author of this line was trying to say about it.

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -426,7 +435,10 @@ The following are the most common types of inputs, along with their correspondin
 
 ##### `met`: Meteorological inputs {#xml-run-inputs-met}
 
-(Under construction. See the `PEcAn.data.atmosphere` package, located in `modules/data.atmosphere`, for more details.)
+-(Under construction. See the
+-[`PEcAn.data.atmosphere` package documentation](https://pecanproject.github.io/package-documentation/develop/PEcAn.data.atmosphere/)
```
infotroph, line 439: This link also doesn't exist! Checking it made me realize the data.atmosphere package docs are not building at the moment, but this _is_ the URL the docs will live at once fixed. So no change needed here; I'll open a separate issue for the failing package docs build.

book_source/03_topical_pages/03_pecan_xml.Rmd
infotroph:
```suggestion
- - [`PEcAn.DB::get.trait.data`](https://pecanproject.github.io/package-documentation/develop/PEcAn.DB/reference/get.trait.data.html)
  — Retrieves trait data for selected PFTs from the database or posterior files
```

infotroph (changes requested):

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -206,7 +209,9 @@ The default `config.php` settings on any given platform (VM, Docker, etc.) or in
 
 Key R functions using these parameters are as follows:
 
-- `PEcAn.DB::db.open` -- Open a database connection and create a connection object, which is used by many other functions for communicating with the PEcAn database.
+- [`PEcAn.DB::db.open`](https://pecanproject.github.io/package-documentation/develop/PEcAn.DB/reference/db.open.html)
+  —-Open a database connection and create a connection object,
```
infotroph:
```suggestion
  -- Open a database connection and create a connection object,
```

infotroph (approved):

Issue #3762, opened 2026-01-08
Irrigation management inputs: CIMIS ET


mdietze:
## Description

Building on the work already completed in https://github.com/PecanProject/pecan/tree/develop/modules/data.remote/inst/Python we need a variant of the irrigation model that uses the California-specific CIMIS evapotranspiration product in place of OpenET. In doing so we want to retain OpenET support as an alternative because it supports a larger region than just CA. 

## Proposed Solution

1. @infotroph already downloaded a bunch of CIMIS data to the BU server, but it would be good to get code to do this download into the repository (or instructions into the documentation if there's not a simple API) so that this can be replicated on other servers

2. Switching ET products in the daily workflow should be fairly straightforward. In doing so please double check units as there were some lingering concerns about unit issues in the earlier switch from daily to monthly OpenET

3. CIMIS provides crop coefficients to account for crop-by-crop differences in ET relative to their reference crop type (pasture?). Adding this in requires looking up at least our harmonized LandIQ polygon layer and dataframe to track crop history.

4. At the same time, it would be good to also look up the phenology information @sarahkanee is extracting and matching with LandIQ so we know not just what crop is planted but when. In doing so it would be good to add functionality to NOT irrigate during the fallow season. More complex irrigation scheduling may be desirable in the future (e.g., some crops need to dry down before planting and/or harvest to facilitate access of machinery) but isn't MVP.

## Additional Context 

Existing README's todo mentions the need to add site specific water holding capacity and crop specific rooting depth. This will be added in a separate, but related, Issue as that task also requires tracking crop type and phenology. 

A third issue will be opened specifically on rice irrigation

infotroph:
With apologies for the brain dump, here's what I know about the CIMIS files:

# What's on the server
	
`/projectnb/dietzelab/ccmmf/data_raw/cimis/` is a recursive fetch of all the daily spatial CIMIS maps available on September 4, 2025 from https://spatialcimis.water.ca.gov/cimis/, meaning it has data from 2003-02-20 to 2025-09-04 plus one random day in 2000 (which seems to be the only pre-2003 data on the site).
	
It contains about 74 GB, organized in subdirectories by year, month, and day. Each directory contains an `index.html` that I downloaded for completeness but have not used, plus one PNG and one `asc.gz` for each of the variables "ETo", "K", "Rnl", "Rs", "Rso", "Tdew", "Tn", "Tx", and "U2" with exceptions as follows.

[...]

PR #3765, opened 2026-01-09
Add BISm crop coefficient dataset and related documentation to PEcAn.data.land

dlebauer:
### Description

This PR contains two key additions

1. crop coefficients (Kc) from BISm
2. functions to convert a generic ET for grass (ETo) into crop and growth stage-specific ET (ETc) 

### What this PR Adds

1. new datasets 
  - `bism_kc_by_crop`, including CSV exported from the BIS spreadsheet and rda data file.
  - `landiq_crop_mapping_codes.tsv` landIQ crop codes
  - `landiq_bsim_lookup.csv` lookup table
2. new functions
  - eto_to_etc_bsim 
  - `eto_to_etc` helper that implements ETc=Kc * ETo; can get Kc from crop name + date or canopy cover  
- Documentation of functions and data
- Tests
- updated CHANGELOG and NEWS.md

### Out of scope

This PR was originally written to add the Kc. Then, I added basic functionality to illustrate how this data can be used.

Future work (can be turned into separate issue): 
- the crop_name is the name from the BISm table.
- look up LandIQ crop ID
- separate function into `get_bsim_kc(crop, date)` and then run `eto_to_etc`
- allow get_bsim_kc to accept planting_date and/or harvest_date.

### References: 

- Snyder, R., Orang, M., Bali, K., Eching, S., Zaccaria, D., 2014. BISm Basic Irrigation Scheduling Excel program (metric units). https://biomet.ucdavis.edu/basic-irrigation-scheduling-(BIS).html (Accessed Jan 8, 2026) [BISm.xlsx](https://github.com/user-attachments/files/24511292/BISm.xlsx)
- Documentation [BIS.pdf](https://github.com/user-attachments/files/24511291/BIS.pdf)
- Theoretical basis[ISWBM.pdf](https://github.com/user-attachments/files/24511308/ISWBM.pdf)


modules/data.land/R/data-docs.R
```diff
@@ -0,0 +1,80 @@
...
+# Crop coefficient (Kc) schedules extracted from the Basic Irrigation Scheduling
+# (BIS) Excel workbook (Snyder et. al., 2014).
+# The dataset is a direct export of the workbook's `CropRef` worksheet, with columns renamed.
+# This dataset provides the information needed to reconstruct a stage-based daily Kc curve when
+# combined with grass-reference evapotranspiration (ETo), such as that provided
+# by CIMIS (California Department of Water Resources, 2025).
+#
+#' @format A data frame with one row per crop and the following columns:
+#' \describe{
+#'   \item{crop_number}{Numeric crop identifier used internally by BIS.}
```
mdietze: Could you match these against the LandIQ crop classes and subclasses, and address any missing classes, as those classes are the default we're using and it would be inefficient to have to match these on the fly. Much simpler to just have the dataset pre-matched.

mdietze (changes requested):
Partially fulfills step 3 of issue #3762 (still requires matching against crop type and phenology to be fully integrated into the irrigation code). Tagging @ashiklom for reference.

modules/data.land/R/data-docs.R
dlebauer: added lookup table in [3c3c52d](https://github.com/PecanProject/pecan/pull/3765/commits/3c3c52db2b0f179a6cf594faa02378ba942c9593)

mdietze: Could you do the join with the look up table and use that to update the bism_crop_coefficients, as requested. There's no reason to ask users to do that join on-the-fly every time we use it.

dlebauer: joined

mdietze (approved):

PR #3770, opened 2026-01-11
made changes as mentioned to BADM_IC_process using PEcAn Settings

### LINKED ISSUE #3699: Make `BADM_IC_process()` use PEcAn Settings / MultiSettings consistently
### Description

`BADM_IC_process()` in `modules/data.land/R/IC_BADM_Utilities.R` currently implements its own ad‑hoc logic to distinguish single‑site vs multi‑site runs by checking for a top‑level "run" element and then wrapping `settings` in `list(settings)`

### Proposed Refactor
* Use `PEcAn.settings::papply()` instead of manual normalization
* `papply()` internally handles `is.Settings()` / `is.MultiSettings()` type checking
* Update _tests_ to ensure this consistent behavior across single-site and multi-site configurations.

shreyannandanwar:
## Description
fixes #3699 
tested with *test-IC_BADM_Utilities.R*
implemented PEcAn Settings successfully

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


shreyannandanwar:
@divine7022 should I also write special unit tests because I don't really see specific test for the implemented changes.

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -236,50 +236,38 @@ netcdf.writer.BADM <- function(lat, long, siteid, outdir, ens){
...
+    # Name outputs consistently
+    site.outputs <- stats::setNames(site.outputs, rep("path", length(site.outputs)))
     
-    out.ense <- c(out.ense, site.outputs)
-  }
+    return(site.outputs)
+  })
+  
+  # Flatten output into a single list
+  out.ense <- unlist(out.ense, recursive = FALSE)
```
infotroph: How sure are you this needs to be flattened? I haven't tried running it so I could be wrong, but the `out.ense` returned by papply looks to me like it'll arrive already equivalent to the one previously created by the for loop, with no unlisting needed.

infotroph (changes requested):

modules/data.land/R/IC_BADM_Utilities.R
shreyannandanwar: You were right about the `unlist` removed it. `papply` already returns the flat structure we need.

PR #3783, opened 2026-01-14
Empty ENS and SA plots in uncertainty notebook

AritraDey-Dev:
## Description
Due to random ID generation during the ENS and SA runs, the generated `.RData` file ends up with a non-deterministic (run-specific) name. As a result, the ENS and SA plots are empty in the rendered documentation [here](https://pecanproject.github.io/pecan-documentation/develop/rendered-demo-notebooks/uncertainty.html#sensitivity-and-variance-decomposition-visualization)

Now it should render properly as shown below  
 
image
image

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


documentation/tutorials/Demo_02_Uncertainty_Analysis/uncertainty.qmd
```diff
@@ -265,25 +266,22 @@ variable <- settings$ensemble$variable
...
-ens_file <- PEcAn.uncertainty::ensemble.filename(
-  settings,
-  prefix = "ensemble.output",
-  ensemble.id = ensemble.id,
-  variable = variable,
-  start.year = start.year,
-  end.year = end.year
-)
+# Find the ensemble output file directly by pattern
+ens_file <- list.files(settings$outdir, pattern = "ensemble\\.output\\..*\\.Rdata", full.names = TRUE)[1]
```
mdietze: I'm curious about the assumption that the ensemble you are looking for is the first one in any list of files. Is this going to generalize to multi-site runs or runs configured to generate SA or EA outputs for multiple variables? Also, if the non-BETY workflow is using a random hash for the ensemble id, why isn't that ending up in settings$ensemble$id?

AritraDey-Dev: >Is this going to generalize to multi-site runs or runs configured to generate SA or EA outputs for multiple variables? 

No, assuming the first file is correct will not generalize. In multi-variable runs, I believe simply looping through the variables and plotting for each variable should fix it.will do that.

AritraDey-Dev: >Also, if the non-BETY workflow is using a random hash for the ensemble id, why isn't that ending up in settings$ensemble$id?

`write.ensemble.configs` function generates and returns the ID in memory but does not save it to updated pecan.xml, so reloading settings from the file causes the ID to be lost.

documentation/tutorials/Demo_02_Uncertainty_Analysis/uncertainty.qmd
```diff
@@ -265,50 +266,75 @@ variable <- settings$ensemble$variable
...
-# Construct the path to the ensemble output file
-ens_file <- PEcAn.uncertainty::ensemble.filename(
-  settings,
-  prefix = "ensemble.output",
-  ensemble.id = ensemble.id,
-  variable = variable,
-  start.year = start.year,
-  end.year = end.year
-)
+requested_vars <- settings$ensemble$variable
```
mdietze: just checking that you've confirmed that if multiple variables are requested they show up as a vector, not a single comma-separated string that needs to be parsed into a vector?

documentation/tutorials/Demo_02_Uncertainty_Analysis/uncertainty.qmd
mdietze: Probably moot under the way you've reorganized the code, but I was confused by this response as 
```
settings <- PEcAn.workflow::runModule.run.write.configs(settings)
```
was called after the settings was loaded from file, and I didn't see any place you reloaded the setting from file after that, so I'm not sure what "reloading settings from the file causes the ID to be lost" was referring to. 

Also, it's good practice to save the settings to make sure this sort of information isn't lost (definitely at the end of the workflow, often at multiple places along the way if you think there's a risk of failure and you want to be able to restart from where you left off). See pecan/web/workflow.R for an example of how often we used to call write.settings.

AritraDey-Dev: Yeah, I debugged it and after running `runModule.run.write.configs(settings)` and then debugging into settings object,it correctly writes the ID(generated hash) to the settings object. 
The root cause was indeed a syntax error,not an ID issue and how we are considering destructuring the ID: it was trying to access `settings$ensemble$id` instead of the correct property `settings$ensemble$ensemble.id`. will push the commit.

documentation/tutorials/Demo_02_Uncertainty_Analysis/uncertainty.qmd
AritraDey-Dev: Yes, `read.settings` parses multiple <variable> tags into a character vector (e.g., c("NPP", "NEE")), so the loop works correctly without extra parsing.

mdietze (approved):

documentation/tutorials/Demo_02_Uncertainty_Analysis/uncertainty.qmd
mdietze: TBH, `settings$ensemble$id` is more intuitive, but let's leave that global find/replace for another Issue, as it will break lots of existing pecan.xml

infotroph:
* You can avoid nondeterministic names by setting ensemble IDs to arbitary fixed strings in the settings file. The gotcha here is that EA you need to set `settings$ensemble$id`  (which is then used to override any `settings$ensemble$ensemble.id` you pass, then `settings$ensemble$ensemble.id` is the one that gets consulted after that), whereas for SA you have to set  `settings$sensitivity.analysis$ensemble.id` directly.
* You can make these plots work without knowing their ensemble ids by calling `ens_file <- PEcAn.uncertainty::ensemble.filename(settings, prefix = "ensemble.output", all.var.yr = FALSE)` at line 276 and `sens_file <- PEcAn.uncertainty::sensitivity.filename(settings, prefix = "sensitivity.results", all.var.yr = FALSE)` at line 322, at which point the `ensemble.id` lookup on line 268 is unused.

AritraDey-Dev:
Oh, thanks @infotroph for the help. It will make things a lot simpler.

infotroph (approved):

Issue #2517, opened 2020-01-16
BioCro day of year conversion error


KristinaRiemer:
### Bug Description

When running `run.biocro` from `PEcAn.BIOCRO` for BioCro version 0.95, it calls `call_biocro` which converts the day of year to a different value. See [lines 34-45](https://github.com/PecanProject/pecan/blob/develop/models/biocro/R/call_biocro.R). This occurs only if planting date is set as after the first day of the year (i.e., `day1 > 1`). 

As an example, we're running the model for *Setaria*. If we set the planting date in `config.xml` to 2010/02/01, the above code converts day1 from 1 to 1.041667 and dayn from 364 to 10656. 

This dayn then causes the following error downstream in `BioCro::BioGro` [lines 283-284](https://github.com/ebimodeling/biocro/blob/master/R/BioCro.R) because it exceeds the upper limit of 365: 

```
Error in BioCro::BioGro(WetDat = WetDat, iRhizome = as.numeric(iplant$iRhizome),  : 
  day1 and dayn should be between 0 and 365
```

@infotroph better understands what this conversion is doing than I do, but it comes from BioCro wanting the met data to start on the first day of the year but still allowing us to start at a later date. A possible solution is to pad the beginning of the weather data with NA values instead of doing this conversion.

PR #3787, opened 2026-01-17
fix biocro planting date bug

AritraDey-Dev:
## Description

When running BioCro with a planting date after January 1st (e.g., 2010/02/01), the model fails with:
```
Error in BioCro::BioGro(WetDat = WetDat, iRhizome = as.numeric(iplant$iRhizome),  : 
  day1 and dayn should be between 0 and 365
Calls: call_biocro_0.9 -> <Anonymous>
Execution halted
```
In call_biocro.R the code converts DOY values to row indices for BioCro:
https://github.com/PecanProject/pecan/blob/e3d02d047d3655da556dd6d55ca1ab08b74e7213/models/biocro/R/call_biocro.R#L48-L50

The code was looking for the last day of weather data, but if you have multiple years of data, it would find day 349 at the very end of year 3 (row 22,896) instead of year 1. Then it calculates:
```
dayn = 22896 / 24 = 954      Too big! (should be ≤ 365)
```

The PR fixes it by instead of searching through all the data, just count how many unique days exist.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


infotroph:
It's been a long time since I touched BioCro, but I think I remember it only accepting one year of input at a time -- hence the year loops in met2model and run.biocro. Were you able to get a multi-year run after making this change?

AritraDey-Dev:
Yes,without this patch only the single-year case works,and for multi year I was getting the error. With this patch, both single-year and multi-year runs work (using the logic described above).
I have 2004–2006 AmeriFlux dataset (US-Bo1), and using that multi-year run works correctly.

models/biocro/R/call_biocro.R
```diff
@@ -38,15 +38,17 @@ call_biocro_0.9 <- function(WetDat, genus, year_in_run,
     }
   )
   if (!biocro_checks_doy && min(WetDat[, "doy"]) > 1) {
+    n_unique_doy <- length(unique(WetDat[, "doy"]))
+    hours_per_day <- nrow(WetDat) / n_unique_doy / length(unique(WetDat[, "year"]))
+    if (hours_per_day < 1 || !is.finite(hours_per_day)) hours_per_day <- 24
```
dlebauer: It doesn't look like `hours_per_day` is being used - is it needed?

AritraDey-Dev: yeah yeah not needed, just a leftover from an earlier iteration of the fix.

AritraDey-Dev:
This one might be related to https://github.com/PecanProject/pecan/issues/2517

models/biocro/DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PEcAn.BIOCRO
 Type: Package
 Title: PEcAn Package for Integration of the BioCro Model
-Version: 1.7.5
+Version: 1.7.5.9000
```
infotroph, line 4: 👍 Can you also add a bullet (in a new section for this new release) to NEWS.md summarizing the fix and referencing PR 3787?

models/biocro/R/call_biocro.R
```diff
@@ -38,15 +38,13 @@ call_biocro_0.9 <- function(WetDat, genus, year_in_run,
     }
   )
   if (!biocro_checks_doy && min(WetDat[, "doy"]) > 1) {
+    n_unique_doy <- length(unique(WetDat[, "doy"]))
+    
     if (!is.null(day1)) {
-      # Biocro calculates line number as `indes1 <- (day1 - 1) * 24`
-      indes1 <- Position(function(x) x == day1, WetDat[, "doy"])
-      day1 <- indes1 / 24 + 1
+      day1 <- 1
```
infotroph: Have you tested this with WetDat starting on DOY 1 and day1 > 1? Looks to me like this will make BioCro always start from the beginning of the file.

AritraDey-Dev: I didn't test this scenario separately,but looking at the code of `run_biocro.R` 

https://github.com/PecanProject/pecan/blob/4b7eb28d8860a5e5bccc6bb9d5b0d9162ab30774/models/biocro/R/run.biocro.R#L63C1-L66C1

The weather data is already filtered before reaching call_biocro_0.9.
So if day1 = 32 (Feb 1), WetDat will only contain DOY 32-365. The full-year data is already removed.
Then in call_biocro.R
https://github.com/AritraDey-Dev/pecan/blob/ee5d7dbe19cf501afd40454f9503b17ef1f57302/models/biocro/R/call_biocro.R#L40-L42

since min(WetDat$doy) = 32 > 1, the fix runs. setting day1 = 1 means "start from row 1 of this filtered file", which is actually DOY 32.Does that make sense?

infotroph: Ah OK, that makes sense, but then I have another question because that line is down inside a loop over years -- how do multi-year runs get through to the model for this problem to happen?

AritraDey-Dev: Yes right,the issue happens when the met file itself contains multiple years.

In run.biocro.R
[WetDat <- data.table::fread(metfile)](https://github.com/PecanProject/pecan/blob/4b7eb28d8860a5e5bccc6bb9d5b0d9162ab30774/models/biocro/R/run.biocro.R#L43)

So if metfile was generated with multiple years of data (by met2model or manual prep), it would contain all years.

infotroph: So both BioCro itself and all the public PEcAn.BIOCRO functions have a pretty deeply baked assumption that one call to BioCro will be one growing season. I think rather than try to support multi-year inputs here in an unexported wrappr, it may be better to check right away after reading that WetDat contains only one year and throw an error if not.

I no longer remember why I calculated day1 the way I did rather set it to 1 and rely on the upstream subsetting, but I suspect it was for confidence that the result would exactly align with the subsetting BioCro does internally.

AritraDey-Dev: Ok,that sounds good. Since run.biocro already handles the logic for multi-year data (by looping), I'm fine with adding an error for multi-year inputs in call_biocro.

Just curious though - is there a specific reason why biocro has this single-year constraint unlike other models, given it is designed to simulate specific growing seasons? (P.S. I couldn't find a clear answer of this in internet)

infotroph: One season per call is pretty common for crop growth models, which is the world BioCro came from. For someone primarily thinking about the plant itself (especially one that's planted and harvested annually) and not worrying about the rest of the ecosystem, it's an easy framework to work in and lets the model avoid a lot of complexity. 

Of course that doesn't mean the PEcAn wrapper _had_ to enforce yearly input files; that was a design decision driven partly by noting that both the model downstream and the met processing workflow upstream were defaulting to separate files per year, so it seemed easier to carry that assumption through the PEcAn.BIOCRO stages too.

FTR BioCro is now on version 3.3.0 and as far as I know nobody has tried using the PEcAn coupler with any version newer than 1.0, so I suspect the next person who wants to use BioCro in PEcAn will want to take on a substantial update -- that would be a good time to address other needed refactors in PEcAn.BIOCRO as well.

models/biocro/R/call_biocro.R
```diff
@@ -38,15 +38,16 @@ call_biocro_0.9 <- function(WetDat, genus, year_in_run,
     }
   )
   if (!biocro_checks_doy && min(WetDat[, "doy"]) > 1) {
+    if (length(unique(WetDat[, "year"])) > 1) {
+      PEcAn.logger::logger.severe("WetDat must contain only one year of data when using BioCro 0.9")
+    }
```
infotroph: Let's move this up to line 15 to enforce single-year files regardless of start date.

models/biocro/R/call_biocro.R
```diff
@@ -38,15 +38,16 @@ call_biocro_0.9 <- function(WetDat, genus, year_in_run,
     }
   )
   if (!biocro_checks_doy && min(WetDat[, "doy"]) > 1) {
```
infotroph: Given call_biocro_0.9 is used only for the long-since-completed Biocro 0.9x release series, honestly it seems pretty safe to delete the `biocro_checks_doy` behavior check and all the mocking contortions that were used to test it, leaving just a shortened version of the comment that explains why we rescale day1 and dayn.

models/biocro/tests/testthat/test-call_biocro.R
```diff
@@ -142,6 +142,38 @@ test_that("call_biocro_0.9 adjusts day1 and dayn when weather is not a whole yea
...
+  
+  WetDat_subset <- WetDat_multi[WetDat_multi$doy >= 2, ]
+  
+  biomock <- mockery::mock(fake_b0.9_result, cycle = TRUE)
+  mockery::stub(
+    call_biocro_0.9,
+    "BioCro::BioGro",
+    function(WetDat, ...) {
+      if (nrow(WetDat) == 1 && WetDat$doy > 1) {
+        stop("This error should be caught silently")
```
infotroph: I couldn't get these tests to fail when run against the develop version, and I think this line is why -- In the test above this I conditioned on the one-row WetDat as a way to test the `biocro_checks_doy` behavior check, not to match a real behavior of `BioGro()`. Seems like with multi-year inputs this is always returning success.

models/biocro/tests/testthat/test-call_biocro.R
```diff
@@ -142,6 +142,38 @@ test_that("call_biocro_0.9 adjusts day1 and dayn when weather is not a whole yea
...
+  expect_error(
+    call_biocro_0.9(
+      WetDat = WetDat_subset, genus = "Miscanthus", year_in_run = 1,
+      config = config, lat = 40, lon = -88,
+      tmp.result = list(), HarvestedYield = 0),
+    "WetDat must contain only one year of data"
+  )
+})
+
+
```
infotroph, line 158: To reason through this I put together a few test cases that use the real BioCro instead of relying on mocks of its (complicated) internal behavior. Though as I paste this I see that if the multi-year check becomes unconditional as I suggested above, then we probably don't need to check 3 different multi-year cases...

```suggestion
test_that("adjustments to day1 and dayn work right with live biocro calls", {
  skip_if_not_installed("BioCro")
  skip_if_not(packageVersion("BioCro") == "0.95")

  met_2004 <- read.csv("data/US-Bo1.2004.csv") |>
    dplyr::filter(doy <= 365) # final timepoint is labeled 366, biocro complains

  call_with <- function(met) {
    PEcAn.BIOCRO:::call_biocro_0.9(
      WetDat = met,
      config = config,
      genus = "Miscanthus",
      year_in_run = 1,
      HarvestedYield = 1
    )$tmp.result
  }

  # Whole year
  expect_equal(length(call_with(met_2004)$LAI), 365 * 24)

  # Q2 data, real DOY
  met_q2_2004 <- met_2004[met_2004$doy %in% 91:180, ]
  res_q2 <- call_with(met_q2_2004)
  expect_equal(length(res_q2$LAI), 90 * 24)

  # Q2 data, labeled as if starting DOY 1
  # If day1/dayn rescaling work right, should give identical growth
  res_q2_d1 <- call_with(met_q2_2004 |> dplyr::mutate(doy = doy - 90))
  expect_equal(
    res_q2 |> purrr::discard_at("doy"),
    res_q2_d1 |> purrr::discard_at("doy")
  )

  # two whole years
  expect_error(
    call_with(
      dplyr::bind_rows(
        met_2004,
        met_2004 |> dplyr::mutate(year = 2005, Temp = 1)
      )
    ),
    "must contain only one year"
  )

  # Multiple years starting with partial year
  expect_error(
    call_with(
      dplyr::bind_rows(
        met_2004 |> dplyr::filter(doy %in% 91:300),
        met_2004 |>
          dplyr::mutate(year = 2005)
      )
    ),
    "must contain only one year"
  )

  # Multiple years, all partial
  expect_error(
    call_with(
      dplyr::bind_rows(
        met_2004 |> dplyr::filter(doy %in% 91:200),
        met_2004 |>
          dplyr::filter(doy %in% 91:300) |>
          dplyr::mutate(year = 2005)
      )
    ),
    "must contain only one year"
  )
})
```

models/biocro/R/call_biocro.R
infotroph: So after poking more at this, I suspect this logic can be simplified WAY down -- day1 and dayn are never null by these lines, and once we enforce one year I _think_ it'll be enough to always tell BioCro that day1 is 1 and dayn is max(doy) - min(doy) -- at which point we can do that without even conditioning on  min(doy) > 1.

infotroph (changes requested):

models/biocro/R/call_biocro.R
infotroph: It may also be prudent to enforce hourly timestep near line 15 as well -- probably enough to just check that the data has `24*n_days` rows.

models/biocro/tests/testthat/test-call_biocro.R
AritraDey-Dev: Yes,  removed those lines in https://github.com/PecanProject/pecan/pull/3787/changes/91a01166ff25bc790c8adea492f56f1645695fde since they were specifically for testing `biocro_checks_doy` which we no longer use

models/biocro/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# PEcAn.BIOCRO 1.7.5.9000
+
+* Fixed `call_biocro_0.9` crash by enforcing single-year weather data input (#3787)
```
infotroph: nit: rephrasing because users ideally don't even know `call_biocro*` exists
```suggestion
* Fixed a crash when using BioCro 0.9x by enforcing single-year weather data input (#3787)
```

infotroph (approved):

infotroph:
Thanks! I agree this will close #2517 too.

PR #3788, opened 2026-01-18
Rothc: Soil file support + read options from settings file

infotroph:
## Description

* Reads texture, bulk density, soil C from a soil_physics file (currently with no fallback)
    * TODO let carbon values from IC override these
* Reads hydrology options and soil depth from settings$model, with fallback to RothC 2.1's out-of-the-box defaults.

Both were using constants hardcoded in write.config.RothC until now.

### New utils function

Also adds `PEcAn.utils::netcdf2df()`, intended to be a general-purpose "please just dump this netcdf into a dataframe" function. So far I've tested it on files with only one dimension that varies (e.g. site-level files that might have a constant lat/lon alongside the depth or time); I'd love if it worked for higher dimensions too but did not tackle that today.

I use it here to read soil files -- I know `PEcAn.data.land::pool_ic_netcdf2list()` would work for this, but I'd rather not pull all the data.land dependencies into a model package if I don't have to. 

Opinions requested on whether this is an improvement and where else it might come in handy.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


infotroph:
~Current test failure is for not declaring a data.land import (for `pool_ic_netcdf2list`), but on consideration that's avoidable. I'll refactor~ Now refactored.

models/rothc/R/read_soil_physics.R
```diff
@@ -0,0 +1,83 @@
...
+      depth_cm = .data$depth |>
+        PEcAn.utils::ud_convert(soil_units$depth, "cm")
+    ) |>
+    # TODO 1: Assumes depth is given to bottom of layer -- is that correct?
+    # TODO 2: this drops layers that extend past bottom
+    # (eg with depth=23 and 0-10/10-30 layering, would use only 0-10)
+    # Consider rescaling partial layers
+    # (Or throwing an error on mismatch and making everyone generate their soil
+    #  files with layers that match model depth?)
+    dplyr::filter(.data$depth_cm <= model_depth) |>
```
abbylewis: Might be helpful to have an explicit error here if there are no depths <= model depth

models/rothc/R/read_soil_physics.R
```diff
@@ -0,0 +1,83 @@
...
+
+  # Input is documented to be in meters,
+  # but providing cm instead seems to be a _very_ common error;
+  # might as well try to handle it here
+  if (tolower(soil_units$depth) %in% c("m", "meters")
+      && any(soil_vals$depth >= 10)) {
+    PEcAn.logger::logger.warn(
+      "Soil depths reported to be in meters, but found values >= 10",
+      "in file", path,
+      "Assuming these are mislabeled cm and treating them as such."
```
abbylewis: Maybe "Assuming the units have been mislabeled and treating all depths as cm" just to be clear that it is not only depths recorded as >= 10 that get set to cm.

Would it be better to have a warning and make people change this themselves?

abbylewis (approved):
@infotroph this all seems very reasonable to me. I added two very minor suggestions.

models/rothc/R/read_soil_physics.R
infotroph:
```suggestion
      "Assuming the units have been mislabeled and treating all depths as cm."
```

models/rothc/R/read_soil_physics.R
```diff
@@ -0,0 +1,83 @@
...
+      && any(soil_vals$depth >= 10)) {
+    PEcAn.logger::logger.warn(
+      "Soil depths reported to be in meters, but found values >= 10",
+      "in file", path,
+      "Assuming the units have been mislabeled and treating all depths as cm."
+    )
+    soil_units$depth <- "cm"
+  }
+
+  soil_vals |>
```
infotroph:
```suggestion
  soil_vals <- soil_vals |>
```

models/rothc/R/read_soil_physics.R
infotroph:
```suggestion
    dplyr::filter(.data$depth_cm <= model_depth)

  if (nrow(soil_vals) == 0) {
    PEcAn.logger::logger.error(
      "No soil layers <=", model_depth, "found in file", sQuote(path)
    )
  }
  
  soil_vals |>
```

PR #3790, opened 2026-01-21
Fix CI failure: pin xfun 0.39 only for R 4.2

AritraDey-Dev:
## Description

Fix ci failure for R 4.2
https://github.com/PecanProject/pecan/actions/runs/21185564344/job/60939202625
https://github.com/PecanProject/pecan/actions/runs/21185564344/job/60939202582

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,15 @@ include Makefile.depends
 	+ ./scripts/time.sh "devtools ${1}" Rscript -e ${SETROPTIONS} -e "if(!requireNamespace('devtools', quietly = TRUE)) install.packages('devtools')"
 	echo `date` > $@
 
-.install/roxygen2: | .install .install/devtools
+.install/xfun: | .install .install/devtools
+	+ ./scripts/time.sh "xfun ${1}" Rscript -e ${SETROPTIONS} \
+		-e "if (getRversion() < '4.4.0' && (!requireNamespace('xfun', quietly = TRUE) || packageVersion('xfun') >= '0.40')) {" \
+		-e "  cran <- c(getOption('repos'), 'cloud.r-project.org')" \
+		-e "  remotes::install_version('xfun', '0.39', repos = cran, upgrade = FALSE)" \
+		-e "}"
+	echo `date` > $@
+
+.install/roxygen2: | .install .install/devtools .install/xfun
 	+ ./scripts/time.sh "roxygen2 ${1}" Rscript -e ${SETROPTIONS} \
 		-e "if (!requireNamespace('roxygen2', quietly = TRUE)" \
 		-e "    || packageVersion('roxygen2') != '"${EXPECTED_ROXYGEN_VERSION}"') {" \


infotroph:
Thanks for working on this. Let's dig a little further before merging -- my bias is that adding more complexity to the Makefile should be a ~last resort, and this is tripping my "feels like the real problem might be elsewhere" spidey sense.

Specifically:
* Both those runs installed roxygen2 7.3.2 as a binary in the last step of the "install new dependencies" step, so I'm not sure why the Make roxygen step is firing at all
* Then once it decides to reinstall Roxygen, I'm not sure  how it's picking up a different xfun:  The R4.2 image installs packages from a static CRAN snapshot, so I don't know what would have changed. I don't see xfun listed in the updates earlier in the log.

AritraDey-Dev:
> Both those runs installed roxygen2 7.3.2 as a binary in the last step of the "install new dependencies" step, so I'm not sure why the Make roxygen step is firing at all

since xfun isn't in `pecan_package_dependencies.csv`, so it's not pinned,so when `docker/depends/pecan.depends.R` installs roxygen2 7.3.2 with `dependencies = TRUE`, it pulls in `xfun` as a transitive dependency, roxygen2 7.3.2 allows `xfun >= 0.30`, so newer versions (>= 0.40) get installed, which don't have `xfun::attr`

AritraDey-Dev:
i believe most proper solution in this case will be to install the xfun specifc version in `docker/depends/pecan.depends.R`

infotroph:
I dug in more here. tl;dr: let's specify `knitr (> 1.50)`

The dependency chain here is:
* roxygen2 imports knitr, with no version specified. The R4.2 image has roxygen 7.3.2 and knitr 1.42, both with no recent change.
* knitr 1.42 imports xfun (>= 0.34). The xfun installed on the R4.2 image has been >0.50 for a very long time was 0.55 in the last successful CI run on Jan 19, and since then has updated to 0.56.
* xfun 0.56 removes the long-deprecated `attr` function, which knitr used in the installed 1.42 and continued to use until 1.50. knitr 1.50 imports xfun (>= 0.51).
* The error from knitr not finding `xfun::attr` makes the roxygen2 namespace fail to load, which explains why the Makefile tries to reinstall it (because it tries to test if it's installed at all using `requireNamespace`).
* I've verified locally that installing knitr 1.50 is sufficient to make Roxygen work again inside the R4.2 image.

Now combining this with @AritraDey-Dev ideas above: holding xfun back would have the same effect, and it would be sufficient to pin it to 0.55 (and continue to ignore the deprecation warnings) rather than need to go all the way back to 0.39. But I don't recommend that because  (1) we don't have a good place to do it (`pecan_package_dependencies.csv` is intentionally limited to first-order dependencies) and would rather not be in the business of worrying about transitive dependency versions at all, and (2) My experience of trying to hold back specific packages is that the headaches compound fast. The entire R toolchain has a pretty strong "when you update one, you'll wind up updating everything and that's OK" assumption.

So after all that digging, I think the easiest fix will be to install the latest knitr by searching for "knitr" in package DESCRIPTIONS and changing all the existing Suggests entries from `knitr` or `knitr (0.42)` to `knitr (>= 0.50)`, then running `./scripts/generate_dependencies.R`. Technically bumping it for one package would be enough to make the build work (because `pecan.depends.R` takes the highest version requirement it finds in `pecan_package_dependencies.csv`), but might as well declare it everywhere while we're at it.

And if further complications arise, we can also consider whether it's time to stop testing on R 4.2: It's the oldest R version we test, so when we start to consider special hacks that aren't needed on newer releases it's a great moment to consider dropping support for it.

infotroph:
Replaced by #3792 -- thanks for reporting and debugging even though we didn't end up using this patch!

PR #3798, opened 2026-02-01
Fix and standardize `@examples` (`\dontrun{}`)

### LINKED ISSUE #3797: Rethink \dontrun in examples
Examples given inside a `\dontrun{}` block are not executed during normal package checking or when the user calls `example()`. But they are still required to be syntactically valid so that users can call them with `example(..., run.dontrun = TRUE)`, and  {pkgdown} attempts to parse them at documentation compilation time, leading to [warnings like](https://github.com/PecanProject/pecan/actions/runs/21419451205/job/61675507431?pr=3796#step:5:2376)

```
── Finished building pkgdown site for package PEcAn.data.remote ────────────────
Warning messages:
1: Failed to parse example for topic "construct_remotedata_filename" 
2: Failed to parse example for topic "get_site_info" 
3: Failed to parse example for topic "remotedata_db_check" 
4: Failed to parse example for topic "remotedata_db_insert" 
```

The first two of these were caused respectively by a missing comma and a stray paren. I imagine most others will be similarly easy, but there are enough of them across all packages that I'm opening this issue rather than try to check them all right now.

## Minimal fix:

Scan a pkgdown build log for "Failed to parse example", look at the Roxygen @examples for each corresponding function, and fix the parsing issue.

* Typos: Fix
* Anything that's not R code (e.g. places that use `\dontrun` to display a bash / python / other non-R command to be run before calling the explicated function): Convert to a comment

## Better fix:

Scan the Roxygen blocks of all package code for uses of `\dontrun`, review, and make changes as needed:

* Is it syntactically correct?
* Is it useful to the example?
* Is it clear why it's marked \dontrun? Does it need to be?
  - Examples of good reasons: Intentionally generates an error; performs a destructive action or has side-effects a user might not want; unavoidably long runtime
  - Examples of bad reasons: Requires inputs that we could provide in the example but currently don't; is not R code; long runtime that could be reduced without compromising example clarity
  - If the only reason for the `\dontrun` is long runtime, consider `\donttest{}` instead.
* Do your answers to the above give you ideas to improve the examples that _are_ run for this function? Implement them.

S1DDHEY:
## Description

fixes #3797 
This WIP PR starts cleaning up Roxygen `@examples` across PEcAn packages.

So far this includes:
- Fixing syntactic errors that cause pkgdown example parse failures
- Making examples self-contained (no undefined symbols)
- Converting non-R commands inside `@examples` to comments
- Standardizing use of `\dontrun{}` vs `\donttest{}` based on side effects

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


S1DDHEY:
I've added most of changes to @examples here, only the `make document` step remains to update the `.Rd` files which I'll  do after this is reviewed.

There are also a few functions whose examples require specific inputs or domain knowledge that I don’t currently have. For those, I plan to open a separate issue.

base/db/R/query.trait.data.R
```diff
@@ -17,8 +17,8 @@
 ##' @export query.trait.data
 ##' @examples
 ##' \dontrun{
-##' settings <- read.settings()
```
infotroph, line 20: Wow, a blast from the past! This is left over from a design that assumed a `settings` object in the user's global environment. We removed that years ago, but apparently missed updating this example. Yes, deleting it here is the right move 👍

base/settings/R/get_args.R
```diff
@@ -7,7 +7,10 @@
 #' @export
 #'
 #' @examples
-#' \dontrun{./web/workflow.R -h}
+#' \dontrun{
+#' ## bash:  
+#' ## ./web/workflow.R -s
+#' }
```
infotroph: This one doesn't need a \dontrun since it's just comments.
```suggestion
#' # Example using a script called from your bash terminal:
#' # If workflow.R contains `get_args()` and settings are in foo.xml, run as
#' # ./workflow.R -s settings.xml
```

base/settings/R/read.settings.R
```diff
@@ -27,7 +27,7 @@
 #' \dontrun{
 #' ## bash shell:
 #' ## example workflow.R and pecan.xml files in pecan/tests
```
infotroph: I think these bash lines can just be deleted here , as honestly can most of the R lines -- one shows a deprecated usage with no filepath, one is a good example of typical usage but not runnable as-is, and one tries to load a file from the package but the path it specifies doesn't exist.

If someone wanted to spend more time on this they could come up with a minimal settings that can be loaded for an example with no \dontrun needed, then store it in a path like `base/settings/inst/settings_example.xml` and call it in the example using system.file(). But meanwhile for thie PR let's just cut the whole section down to one example and move on.

(Continued in another comment below because GitHub doesn't support suggestions that apply to lines both above and below the change I'm commenting on)
```suggestion
```

base/utils/R/timezone_hour.R
```diff
@@ -5,7 +5,6 @@
 #' @param timezone to be converted
 #' @return hours offset of the timezone
 #' @examples
-#' \dontrun{
 #' timezone_hour('America/New_York')
 #' }
```
infotroph:
```suggestion
```

base/utils/R/utils.R
infotroph: Make sure to remove the closing brace too
```suggestion
```

models/fates/R/model2netcdf.FATES.R
```diff
@@ -12,8 +12,23 @@
...
+##' pfts <- c(needleleaf = 1, broadleaf = 2)
+##'
+##' model2netcdf.FATES(
+##'   outdir = outdir,
+##'   sitelat = sitelat,
+##'   sitelon = sitelon,
+##'   start_date = start_date,
+##'   end_date = end_date,
+##'   vars_names = vars_names,
+##'   pfts = pfts
```
infotroph:
```suggestion
##' model2netcdf.FATES(
##'   outdir = "~/",
##'   sitelat = 42.54,
##'   sitelon = -72.17,
##'   start_date = as.POSIXct("2001-01-01", tz = "UTC"),
##'   end_date = as.POSIXct("2005-12-31", tz = "UTC"),
##'   vars_names = c("AGB", "NPP"),
##'   pfts = c(needleleaf = 1, broadleaf = 2)
```

modules/data.land/R/IC_BADM_Utilities.R
```diff
@@ -13,9 +13,8 @@
 #'
 #' @export
 #' @examples
-#' \dontrun{
-#'   badm_test <- Read.IC.info.BADM(45.805925,-90.07961)
-#'}
+#' badm_test <- Read.IC.info.BADM(45.805925,-90.07961)
```
infotroph, line 16: 👍

modules/data.land/R/soil2netcdf.R
```diff
@@ -22,9 +22,12 @@
...
-#'  (0.3,0.4,0.5), fraction_of_clay_in_soil = c(0.3,0.3,0.3), soil_depth = c
-#'  (0.2,0.5,1.0))
+#' \dontrun{ 
+#' soil.data <- list(
+#'   fraction_of_sand_in_soil = c(0.3,0.4,0.5),
+#'   fraction_of_clay_in_soil = c(0.3,0.3,0.3),
+#'   soil_depth = c(0.2,0.5,1.0)
+#' )
 #'
 #' soil2netcdf(soil.data,"soil.nc") }
```
infotroph:
```suggestion
#' soil2netcdf(soil.data,"soil.nc")
#' }
```

infotroph (changes requested):
Thanks for working through this! Overall this looks great. I left some minor suggestions inline.

base/settings/R/read.settings.R
infotroph:
```suggestion
#' settings <- read.settings(file="willowcreek.xml")
```

infotroph: This comment was supposed to be suggesting that you delete lines 30-34 and 37, leaving the current line 36 in their place. No idea why it's showing as only changing one line 🤷

base/utils/R/utils.R
S1DDHEY: kept missing these closing braces a lot. Thank you for noticing

base/settings/R/read.settings.R
```diff
@@ -25,16 +25,8 @@
...
 #' \dontrun{
-#' ## bash shell:
-#' ## example workflow.R and pecan.xml files in pecan/tests
-#' R --vanilla -- --settings path/to/mypecan.xml < workflow.R
-#'
-#' ## R:
-#'
-#' settings <- read.settings()
 #' settings <- read.settings(file="willowcreek.xml")
 #' test.settings.file <- system.file("tests/test.xml", package = "PEcAn.settings")
```
infotroph:
```suggestion
```

infotroph (approved):

base/settings/man/read.settings.Rd
```diff
@@ -30,16 +30,8 @@ is passed through a higher priority method.
...
 \dontrun{
-## bash shell:
-## example workflow.R and pecan.xml files in pecan/tests
-R --vanilla -- --settings path/to/mypecan.xml < workflow.R
-
-## R:
-
-settings <- read.settings()
 settings <- read.settings(file="willowcreek.xml")
 test.settings.file <- system.file("tests/test.xml", package = "PEcAn.settings")
```
infotroph:
```suggestion
```

modules/data.remote/R/remote_process.R
```diff
@@ -293,17 +293,17 @@ remote_process <- function(settings) {
...
-##'   projection=NULL
+##'   scale=10.0,
+##'   projection=NULL,
 ##'   qc=1.0,
 ##'   algorithm="snap",
-##'   out_process_data="lai")
-##' }
+##'   out_process_data="lai"
+##' )
+##' remotedata_file_names
```
infotroph, line 307:
```suggestion
##' remotedata_file_names
##' }
```

modules/data.remote/R/remote_process.R
```diff
@@ -293,17 +293,17 @@ remote_process <- function(settings) {
 ##' @param out_process_data variable name requested for the processed file, NULL by default
 ##' @return remotedata_file_names
 ##' @examples
-##' \dontrun{
 ##' remotedata_file_names <- construct_remotedata_filename(
```
infotroph, line 297: The current CI failure is because this function is unexported and therefore not user-visible when executing examples. It's possible that means we should tell Roxygen to skip documenting it entirely (with `@noRd`), but for now this is the easy route.

```suggestion
##' \dontrun{
##' remotedata_file_names <- construct_remotedata_filename(
```

modules/data.remote/man/construct_remotedata_filename.Rd
```diff
@@ -39,17 +39,17 @@ remotedata_file_names
 construct remotedata module file names
 }
 \examples{
-\dontrun{
 remotedata_file_names <- construct_remotedata_filename(
```
infotroph, line 43:
```suggestion
\dontrun{
remotedata_file_names <- construct_remotedata_filename(
```

modules/data.remote/man/construct_remotedata_filename.Rd
```diff
@@ -39,17 +39,17 @@ remotedata_file_names
...
-  projection=NULL
+  scale=10.0,
+  projection=NULL,
   qc=1.0,
   algorithm="snap",
-  out_process_data="lai")
-}
+  out_process_data="lai"
+)
+remotedata_file_names
```
infotroph, line 53:
```suggestion
remotedata_file_names
}
```

S1DDHEY:
@infotroph Thank You for the review, as mentioned earlier there are a few places where specific inputs are needed for which I don't have that domain specific knowledge (for this one I'll open an issue mentioning all of them).

> If someone wanted to spend more time on this they could come up with a minimal settings that can be loaded for an example with no \dontrun needed, then store it in a path like base/settings/inst/settings_example.xml and call it in the example using system.file(). But meanwhile for thie PR let's just cut the whole section down to one example and move on.

Also I can look forward to implement something similar to this one as well.

PR #3802, opened 2026-02-03
Remove skip from read.settings test

AritraDey-Dev:
## Description

Removes a skip() call from the "read.settings returned correctly" test,After digging into git history , it was added as a temporary workaround in 2017 ([ac060fde9](https://github.com/PecanProject/pecan/commit/ac060fde9)) during the multisite/MultiSettings feature development (PR #1325).

The underlying issue has long been resolved and the test now passes. This unskips 1 previously disabled test.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/base/settings/tests/testthat/test.read.settings.R
+++ b/base/settings/tests/testthat/test.read.settings.R
@@ -74,8 +74,7 @@ skip_if_not(
 )
 
 test_that("read.settings returned correctly", {
-  s <- .get.test.settings()
-  skip("Tests failing due to multisite?")
+  s <- .get.test.settings(testdir)
   expect_true(file.exists(s$outdir))
   expect_true(file.info(s$outdir)$isdir)
 })


infotroph:
Looks like GHA found a way to make these fail 😅 -- I wonder if the pass/fail  result has more to do with tests exiting at the top-level `skip_if_not` on line 71 than with line 78.

But on inspection, that whole test block is only checking whether `read.settings()` creates an outdir. It does not and I think that's correct: Reading a settings file and validating its paths should be separate operations. What if we delete  all of lines 75-81?

AritraDey-Dev:
Yes, locally just removing the `skip() `worked for me. It was failing because it used the XML default `outdir: pecan`, which doesn’t exist in CI.
Now passing `testdir` uses the temp dir that gets created, so CI is satisfied.

infotroph:
Ah, OK, so this is actually a check that we've set up for the test blocks downstream rather than a check of read.settings behavior! 

This whole file could use a refactor to adopt better test isolation (ideally to the point of each test_that being independent of the others), but that's a different PR.

infotroph (approved):

Issue #3799, opened 2026-02-02
Ambiguous row filtering when all BADM carbon pool values are NA


shreyannandanwar:
### Bug Description

In `IC_BADM_Utilities.R`, rows where all carbon pool values are `NA` are filtered using `which()`. When all rows match the condition (all values are NA), `which(ind)` returns all row indices, and `-which(ind)` attempts to exclude all rows, which works but is implicit. When no rows match (or when the data frame is empty), `which(ind)` returns `integer(0)`, and `-integer(0)` becomes `integer(0)`, which keeps all rows due to R's subsetting behavior. This reliance on implicit behavior obscures intent and may be fragile.

## Location
`modules/data.land/R/IC_BADM_Utilities.R`

## Current Code
```r
ind <- apply(entries[, 5:8], 1, function(x) all(is.na(x)))
entries <- entries[-which(ind), ]
```

### Case 1: Mixed data 
```r
entries <- data.frame(
  A = 1:3, B = 1:3, C = 1:3, D = 1:3,
  pool1 = c(1, NA, NA),
  pool2 = c(2, NA, NA),
  pool3 = c(3, NA, NA),
  pool4 = c(4, NA, NA)
)

ind <- apply(entries[, 5:8], 1, function(x) all(is.na(x)))
entries[-which(ind), ]  

# Works: removes rows 2-3
```

### Case 2: All rows are NA (edge case)
```r
entries_all_na <- data.frame(
  A = 1:3, B = 1:3, C = 1:3, D = 1:3,
  pool1 = c(NA, NA, NA),
  pool2 = c(NA, NA, NA),
  pool3 = c(NA, NA, NA),
  pool4 = c(NA, NA, NA)
)

ind <- apply(entries_all_na[, 5:8], 1, function(x) all(is.na(x)))

# which(ind) returns c(1, 2, 3)
# -which(ind) returns integer(0) after attempting to exclude all rows

entries_all_na[-which(ind), ]  

# Returns 0-row data frame (works but implicit)
```

### Case 3: Empty data frame (edge case)
```r
entries_empty <- entries[0, ]
ind_empty <- apply(entries_empty[, 5:8], 1, function(x) all(is.na(x)))
entries_empty[-which(ind_empty), ]  

# Returns empty data frame
```

## What I think we should do:
Just use a boolean mask instead - it's clearer and doesn't rely on the which() trick:
```r
rind <- apply(entries[, 5:8], 1, function(x) all(is.na(x)))
[...]

PR #3801, opened 2026-02-03
Use explicit boolean mask to filter all-NA carbon pool rows

anshul23102:
## Description
This PR updates the row-filtering logic in `IC_BADM_Utilities.R` to remove rows where
all carbon pool values are `NA` using an explicit boolean mask instead of relying on
`-which(ind)`.

The filtering logic has been changed to:

```r
entries <- entries[!ind, , drop = FALSE]
```

## Motivation and Context
Previously, rows with all-NA carbon pool values were filtered using `-which(ind)`.
While this approach works, it relies on implicit behavior in R subsetting,
particularly in edge cases such as:

- when all rows match the condition (all values are `NA`),
- when no rows match the condition, or
- when the input data frame is empty.

Using a boolean mask (`!ind`) makes the logic explicit, improves readability, and
ensures robust handling of all edge cases without changing the resulting output.
This is particularly relevant when BADM returns all-NA values for a site.

## Related Issues
- [x] #739

---

## Review Time Estimate
- [x] Immediately
- [ ] Within one week
- [ ] When possible

---

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

---

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] My name is in the list of CITATION.cff
- [x] I agree that PEcAn Project may distribute my contribution under any or all of
  - the same license as the existing code,
  - and/or the BSD 3-clause license.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3804, opened 2026-02-06
Add regression test for all-NA BADM carbon pool filtering

shreyannandanwar:
### Summary

This PR adds a regression test covering the case where all BADM carbon pool values are `NA`.

### Changes

- Add a test to `test-IC_BADM_Utilities.R` verifying that all-NA carbon pool
  entries are handled cleanly
  
### Issue

#3799

### Tests

- Added test to `test-IC_BADM_Utilities.R`
- All existing tests pass

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


modules/data.land/tests/testthat/test-IC_BADM_Utilities.R
```diff
@@ -19,6 +19,30 @@ test_that("Read.IC.info.BADM falls back to L1 and ALL if no L2 data", {
...
+  entries <- data.frame(
+    GROUP_ID = c("A", "B", "C"),
+    LAT = c(42, 43, 44),
+    LON = c(-72, -73, -74),
+    UNIT = c("kg/m2", "kg/m2", "kg/m2"),
+    leafC = c(NA, NA, NA),
+    stemC = c(NA, NA, NA),
+    rootC = c(NA, NA, NA),
+    soilC = c(NA, NA, NA),
+    stringsAsFactors = FALSE
```
S1DDHEY: For these test entries, it would be better to use the documented `BADM` long-format (`VARIABLE` / `DATAVALUE`) structure rather than introducing ad-hoc columns.

Those can be found here https://pecanproject.github.io/package-documentation/develop/PEcAn.data.land/reference/BADM.html#format

This will be helpful while testing using the function itself.

modules/data.land/tests/testthat/test-IC_BADM_Utilities.R
```diff
@@ -19,6 +19,30 @@ test_that("Read.IC.info.BADM falls back to L1 and ALL if no L2 data", {
...
+  )
+  
+  # Run the same filtering logic used internally
+  ind <- apply(entries[, 5:8], 1, function(x) all(is.na(x)))
+  filtered <- entries[-which(ind), ]
```
S1DDHEY: Using the function directly is recommended instead of re-implementing ad-hoc logic, so the test validates the real behavior.

S1DDHEY (changes requested):

shreyannandanwar:
Thanks for the feedback!
You're right - I was testing the internal logic rather than the actual function behavior. I will be updating the test

S1DDHEY (approved):

modules/data.land/tests/testthat/test-IC_BADM_Utilities.R
```diff
@@ -19,6 +19,23 @@ test_that("Read.IC.info.BADM falls back to L1 and ALL if no L2 data", {
...
+  result <- Read.IC.info.BADM(lat, lon)
+  
+  expect_s3_class(result, "data.frame")
+  
+  if (nrow(result) > 0) {
+    carbon_cols <- c("AGB", "soil_organic_carbon_content", "litter_carbon_content", "root_carbon_content")
+    
+    has_data <- apply(result[, carbon_cols], 1, function(x) any(!is.na(x)))
+    expect_true(all(has_data), info = "All rows should have at least one non-NA carbon value")
+  }
```
infotroph: `if` is undesirable here:
* The result is deterministic, but this makes it look like there could be variation
* If a future regression did make this query return 0 rows, that should be an error not a a silent pass.

```suggestion
  carbon_cols <- c("AGB", "soil_organic_carbon_content", "litter_carbon_content", "root_carbon_content")  
  has_data <- apply(result[, carbon_cols], 1, function(x) any(!is.na(x)))
  expect_true(all(has_data), info = "All rows should have at least one non-NA carbon value")
```

infotroph: Note also that technically this isn't testing that NAs are _removed_, just that this specific query doesn't _return_ any. Knowing that 3 all-NA record did exist to be removed by the filter requires digging around in the source data.

infotroph (changes requested):

modules/data.land/tests/testthat/test-IC_BADM_Utilities.R
shreyannandanwar: Thanks @infotroph, 
my previous approach didn't actually prove all-NA rows were being filtered out. So, I made a new one with controlled data: 
I created a test dataset with a known all-NA row, applied the filtering logic. I also tested edge cases - like all rows NA and no NA at all - to ensure it works reliably. This way, I’m testing the fix directly.

infotroph: I appreciate you trying here, but this update is a big step backwards -- it goes back to not actually calling `IC_BADM_Utilities`.  

Remember that when you're testing the task is always to prove correct behavior of _the code that's actually implemented_. And since this is a regression test for an issue fixed in a specific known patch, the test-driven development mantra applies as well: If you haven't seen the test _fail_ against the buggy code (i.e. the version of `IC_BADM_Utilities` from before PR #3801) then seeing it _pass_ against the patched code doesn't prove the bug is fixed.

infotroph:
@shreyannandanwar Are you interested in working more on this? My comments in https://github.com/PecanProject/pecan/pull/3804#discussion_r2855021682 still apply: The output you test on needs to be generated by the function you're trying to test.

shreyannandanwar:
@infotroph 
Yes, I will work on it. And provide the solution for review.

shreyannandanwar:
I've updated the regression test so that it exercises `Read.IC.info.BADM()` directly rather than reimplementing the filtering logic inside the test. The assertions now validate the function output and verify that all-NA carbon pool rows are not returned.

PR #3806, opened 2026-02-09
Bug fix for the current ERA5 extract function.

DongchenZ:
Bug fixes for the current ERA5 extraction function.

## Description

Here are the summary of the bugs I found and where I fixed those:
1. Line 145: The time should be in the `dim` field instead of the `var` field in an ERA5 netCDF file.
2. Line 167: My ERA5 nc file contains the `short` data type, so I added it. Otherwise, all variables will be skipped because they don't align with the current data type restrictions.
3. Line 200: here I first simplified the logics here (`If` controls) so that we will be extracting ERA5 depending on the condition that an NC file is ensemblized or not (which is one of the previous features). After that, instead of storing large arrays as an intermediate variable in R (which consumes a lot of memory), directly extracting those through `raster::brick(ncfile, varname = vname, level = ens)` will be faster and save a lot of memory because it uses pointers instead of the full datasets (the previous ERA5 extraction function will not be working at the NEON scale due to the memory limitations).

Code has been tested from 2012 to 2024 at the CONUS NEON scale.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.atmosphere/R/extract_ERA5.R
+++ b/modules/data.atmosphere/R/extract_ERA5.R
@@ -142,7 +142,7 @@ extract.nc.ERA5 <-
       ncfile <- file.path(in.path, paste0(in.prefix, year, ".nc"))
       # open the file
       nc_data <- ncdf4::nc_open(ncfile)
-      time_var <- if ("time" %in% names(nc_data$var)) "time" else "valid_time"
+      time_var <- if ("time" %in% names(nc_data$dim)) "time" else "valid_time"
       t <- ncdf4::ncvar_get(nc_data, time_var)
       tunits <- ncdf4::ncatt_get(nc_data, time_var)
       tustr <- strsplit(tunits$units, " since ")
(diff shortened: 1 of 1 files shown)


infotroph (approved):
Approving because I bet I know your answer, but just to confirm: Have you verified this works on ERA5 files downloaded before their format changes in 2024-2025? I don't see any reason it wouldn't, but worth checking.

DongchenZ:
> Approving because I bet I know your answer, but just to confirm: Have you verified this works on ERA5 files downloaded before their format changes in 2024-2025? I don't see any reason it wouldn't, but worth checking.

Thanks for the note. I just reran the code, and it worked from 2012 to 2024. Because the ERA5 files I downloaded have two versions of the data: 2012 - 2021 in the old format and 2022 - 2024 in the new format, the new edits should apply in both cases.

PR #3807, opened 2026-02-10
Add the by-product debias feature.

DongchenZ:
## Description

The debias module was executed solely by state variables. However, if a variable has multiple assimilated data constraints (e.g., LandTrendr and GEDI AGB), the residual error predictions should also be split by the product used for assimilation. This PR added this feature to the general SDA debiasing workflow.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/assim.sequential/R/sda_bias_correction.R
```diff
@@ -56,7 +56,36 @@ sda_bias_correction <- function (site.locs,
...
+      v.ind.pre <- which(names(obs.mean[[t-1]][[s]]) == v)
+      v.ind.t <- which(names(obs.mean[[t]][[s]]) == v)
+      if (length(v.ind.pre) > 0 &
+          length(v.ind.t) > 0) {
+        # match data name.
+        a.pre <- attributes(obs.mean[[t-1]][[s]])
+        a.pre <- a.pre$Source[v.ind.pre]
+        a.t <- attributes(obs.mean[[t]][[s]])
+        a.t <- a.t$Source[v.ind.t]
+        if (a.pre == a.t) {
```
mdietze: The logic of what's happening in this bit of code isn't clear. The comment is about matching names, but it's not clear what you're actually doing (e.g., what is site.ind?) or why. If the goal is to have product-specific bias corrections, why does it matter that the attributes from the previous time step have to match the attributes of the current time step? Doesn't it just matter what the data constraints are at the current time point? Importantly, don't hard code assumptions based on current analyses -- in the general case you could have k constraints for any particular variable and you can't assume a one-transition among any particular product to another

modules/assim.sequential/R/sda_bias_correction.R
```diff
@@ -95,51 +119,70 @@ sda_bias_correction <- function (site.locs,
...
+        model <- randomForest::randomForest(formula,
+                                            data = s.df,
+                                            ntree = 1000,
+                                            na.action = stats::na.omit,
+                                            keep.forest = TRUE,
+                                            importance = TRUE)
+        var.imp <- randomForest::importance(model)
+        models[[v]][[s]] <- var.imp # store the variable importance.
+        # predict residuals for the current time point.
+        res.current <- stats::predict(model, s.cov.df)
```
mdietze: question: how does the code work if a state variable has multiple constraints. That would mean that there could be multiple predicted residuals for the same variable in the same time step and site. How are you then combining those to different residuals to determine what the final bias correction should be?

modules/assim.sequential/R/sda_bias_correction.R
DongchenZ: In short, this bit of code tries to split the before and after observations into three categories across locations: 1) if a certain location has observations from the same product, it will be `1`; If the location has no observations or observations from different products, it will be `0`; 3) The `data.source` will be `NA` for the current site if the flag is `0`.

Regarding your concerns above, it is important to match the product name, because residual errors between products in our SDA reanalysis should be internally distinct (e.g., AGB errors between SDA and LandTrendr should differ from those between SDA and GEDI). Therefore, we need to identify which site has continuous observations for the same state variable and product type. 

In general, if we have `k` products to constrain `m` state variables across `n` sites. We need to identify which product consistently falls within a single variable for a specific site.

I can definitely add more context in the code to clarify the intent here.

modules/assim.sequential/R/sda_bias_correction.R
DongchenZ: The short answer is: we train and predict them separately. In detail, for each variable, for each product, we will have an intermediate predicted residual error called `res.current`. After that, we will put the current predicted residual error `res.current` back to the overall residual error vector across site(`res.all`), which was created outside of the by-product `for-loop` (Line 125).

DongchenZ: You are right; I think there is a bug in the step where we insert the current residual error (`res.current`) into the overall residual error (`res.all`) vector. I think we need to index which residual error prediction belongs to which site, then combine them.

DongchenZ (commented):
Thanks for the comments. After reviewing your comments, I found a bug in the step where we insert the current residual error (res.current) into the overall residual error (res.all) vector. I will need to index which residual error prediction belongs to which site, then combine them.

mdietze (approved):

PR #3808, opened 2026-02-10
move {traits} to Suggests

infotroph:
Problem: data.land imports {traits}, which imports {taxize}, which was recently [re?]archived on CRAN because it uses the archived {wikidata}. Therefore our CI checks currently fail on R-devel when {traits} fails to install.

Solution: Noting that {traits} is used by only two functions, in both cases as a fallback when no Bety connection is provided, move {traits} to Suggests in data.land and add checks for availability to the code branches that use it.

### DIFF
--- a/modules/data.land/NEWS.md
+++ b/modules/data.land/NEWS.md
@@ -11,6 +11,11 @@
   *  `landiq_crop_mapping_codes` dataset mapping LandIQ crop classification codes to human-readable crop names.
   *  `bism_kc_by_crop` dataset containing BISm crop coefficient schedules and stage timing references for use in ET estimation, including columns that map to LandIQ class and subclass.
 
+## Changed
+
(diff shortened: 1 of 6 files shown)


modules/data.land/NEWS.md
```diff
@@ -11,6 +11,11 @@
   *  `landiq_crop_mapping_codes` dataset mapping LandIQ crop classification codes to human-readable crop names.
   *  `bism_kc_by_crop` dataset containing BISm crop coefficient schedules and stage timing references for use in ET estimation, including columns that map to LandIQ class and subclass.
 
+## Changed
+
+* Package `traits`, used by `match_pft()` and `match_species_id()` only when no local database connection is provided, is now suggested rather than required.
```
infotroph:
```suggestion
* Package `traits`, used by `match_pft()` and `match_species_id()` only when no database connection is provided, is now suggested rather than required.
```

dlebauer (approved):

Issue #3735, opened 2025-12-25
Use `saveRDS/readRDS` instead of `save/load`


ashiklom:
## Description

`save` saves objects according to their name in the current environment, and `load` reads them into the target environment under the name that they were saved. This creates some problems/annoyances:

For `save`, the following will throw a (confusing) error because the object literally called `'mylist[["thing1"]]'` does not exist in the environment.

```r
mylist <- list(thing1 = "something", thing2 = "something_else")
save(mylist[["thing1"]], file = "thing1.RData")
# Error in save(mylist[["thing1"]], file = "thing1.RData") : 
#   object ‘mylist[["thing1"]]’ not found
```

To accomplish this, you need to do annoying things like:

```r
thing1 <- mylist[["thing1"]]
save(thing1, file = "thing1.RData")
rm(thing1)
```

For `load`, to read specific objects into other names, we need to do some awkward gymnastics with environments, like:

```r
myenv <- new.env()
load("thing1.RData", envir = myenv)
thing1a <- myenv$thing1
```

If you _don't_ do that, `load` will silently override objects with the same name in the current environment:

```r
thing1 <- "something else"
load("thing1.RData")
print(thing1)
# [1] "something"   # NOTE: `thing1` has been overridden!
```

## Proposed Solution

All of these problems go away with `saveRDS` / `readRDS`.

```r
saveRDS(mylist[["thing1"]], file = "thing1.rds")
thing1a <- readRDS("thing1.rds")
```

ashiklom:
For a real-life example, see https://github.com/PecanProject/pecan/pull/3728/changes/f1f7146cc8135637467d1115c233c54159677205.

SouravVerma-art:
I'd like to work on this issue. I've reviewed the problem and have a good understanding of what needs to be done. Could you please assign this to me?

Thomas-Sedhom:
Hi @ashiklom ,
I’d like to work on this issue if it’s still open and no one else is working on it.

infotroph:
@Thomas-Sedhom Thanks for diving in on this! One design point to settle before we get too far into the implementation is that we should make a distinction between:
 
* writing and loading of temporary checkpoint objects, where format changes are minimally disruptive as long as we change the creating and consuming processes together;
* handling of user artifacts, where people may be counting on PEcAn to keep accepting an RData files they generated in the past. 

[...]

PR #3810, opened 2026-02-10
Replace save/load with saveRDS/readRDS to avoid environment side effects

Thomas-Sedhom:
Replace save/load with saveRDS/readRDS to avoid environment side effects

## Description

The current codebase uses save() and load() for intermediate objects. This approach has several drawbacks:
1- save() requires creating temporary variables to store list elements which is awkward and error prone.
2- load() injects objects by name into the environment and potentially overwriting existing variables silently.
3- Reading specific objects requires extra gymnastics with custom environments.

and the solution is: 
All problematic uses of save() and load() have been converted to saveRDS() and readRDS():
1- saveRDS() saves the object directly, without relying on its name.
2- readRDS() explicitly returns the object, which can be assigned to any variable.
This eliminates side effects and makes the code more readable and maintainable.

## Motivation and Context

This pattern makes the code harder to maintain and debug.

https://github.com/PecanProject/pecan/issues/3735

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


infotroph:
@Thomas-Sedhom Thanks for taking this on! You're taking the right direction for what was originally suggested in #3735, but seeing the diff makes me realize there's a backwards compatibility angle that needs some attention. I said more in the [issue thread](https://github.com/PecanProject/pecan/issues/3735#issuecomment-3879459854) -- Let's discuss there and then come back here to implement once the design is right.

infotroph:
Closing this because of substantial overlap with a bigger refactor @omkarrr2533 is leading. Thanks for the contribution and sorry it went unused!

PR #3813, opened 2026-02-14
Sipnet v2 support

infotroph:
## Description

Pushing some long-stashed changes so we can iterate on them. <del>NOT YET TESTED.</del>

Implemented so far:
* write clim files in the 12-column vs format by default
* switch on version when choosing a template for sipnet.in and sipnet.param

Before merging, need to at least:

* add CH4 and N2O to model2netcdf
* update and extend unit tests

And probably also:
* check for simplifications made possible by the param file being name/value pairs instead of multicolumn
* update write_restart/read_restart
* Implement changing command-line switches from settings (esp. turning off GDD)

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/sipnet/inst/template.param_v2
```diff
@@ -0,0 +1,70 @@
+plantWoodInit	30000
```
divine7022, line 1: The v2 template drops `totNitrogen` and `microbeNC`, but sipnet v2 still reads both via `initializeOneModelParam`. They default to 0 when absent, which silently disables N cycle dynamics. Is that intentional here? If want N-cycle output (n2o, N leaching), need these parameters ; might be worth including them with sensible defaults rather than omitting

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -38,16 +46,28 @@
...
+met2model.SIPNET <- function(in.path,
+                             in.prefix,
+                             outfolder,
+                             start_date,
+                             end_date,
+                             var.names = NULL,
+                             overwrite = FALSE,
+                             verbose = FALSE,
+                             year.fragment = FALSE,
+                             use_v1_format = FALSE,
```
divine7022: defaulting to FALSE is a breaking change for any workflow passing sipnet v1 binaries; the 12 col clim file will cause v1 to misparse columns. The NEWS.md documents this, but i am concerned about the caller side:
driver calls `met2model_method(in.path, in.prefix, outfolder, start_date, end_date)` it doesn't pass `use_v1_format` at all, so it will always get the new default

so options would be:
1) the driver should detect the sipnet version from `settings$model$revision` and pass `use_v1_format` accordingly (but met2model doesn't receive settings).
2 ) default to TRUE (backward compat) and let v2 users opt in, at least until the broader workflow plumbing is updated

what do you think here? i would lean toward 2nd until we can thread the version through the met processing pipeline

models/sipnet/inst/template.param_v2
infotroph: Good catch -- sounds like Sipnet's expectations have changed since I stashed this draft. Yes, including them here is the right move.

models/sipnet/R/met2model.SIPNET.R
infotroph: Short answer: Yeah, I know this is breaking for v1 but I think that's the simplest route to the right long-term result.

I did consider option 1 and am fairly sure it's not what we want. PEcAn's existing model is to provide one standard met format and send the same information to every `met2model` call, which the model coupler is then responsible for converting as needed. Given most models do not change met formats between versions (or even ever in their lives), passing version information through the PEcAn layer seems like it would add a lot of complexity for something that's both rare and conceptually the model's job.

I also lean away from defaulting to V1:
* I want to make sure PEcAn can take advantage of Sipnet v2's simplifications rather than get stuck using old formats for compatibility. I'd be OK with waiting for some specific future change before switching the default to v2, but if replumbing of the broader workflow isn't needed then I don't know what change we'd be waiting for.
* I'm assuming that once PEcAn gains full support for Sipnet v2, most users will want to switch to it. I admit this assumption is untested.
* Folks still using Sipnet v1 are likely to be folks who've been using it for a while and have a collection of already-processed v1 met files. So the fraction of met2model calls that want v2 will rise even faster than the fraction of write.configs calls that want v2.
* For as much work as we put into the met.convert pipeline that was supposed to fetch raw data and convert it on demand, it has never worked as smoothly as we envisioned and my understanding is that much (most? all?) met processing now gets done in distinct steps -- so users now tend to call met2model.SIPNET directly and can add version flags to those calls without passing the information further up the pipeline.

FWIW, it's also "easy" to convert between formats outside of PEcAn using e.g. `cut -f2-13 v1.clim > v2.clim` and `awk '{print 0 "\t" $0 "\t" 0.6}' v2.clim > v1.clim`, but the ease or lack thereof is about the same in both directions so I don't think that affects which default we should pick.

divine7022:
picking up from Chris's initial v2 scaffolding , the main gaps were that the v2 templates were incomplete and model2netcdf couldn't parse v2 output. Here's what this round adds:

`template.param_v2` -- stripped the 22 params that no longer exist in v2 (microbeInit, qualityLeaf, cn_ prefixed ones, etc.) and added the 16 N-cycle/anaerobic/CH4 params that `initializeOneModelParam()` now expects. All 73 params verified against `sipnet.c`

`sipnet.in_v2` -- added the six runtime flags (NITROGEN_CYCLE, LITTER_POOL, ANAEROBIC, GROWTH_RESP, WATER_HRESP, LEAF_WATER) that replaced the old compile time #ifdefs in v2

`write.config.SIPNET` -- flag injection block so users can toggle N-cycle/litter/anaerobic from the XML settings. Also added guards so v1 only params (litterWHC, litWaterDrainRate, litterWFracInit, microbeInit, m_ballBerry) don't get written into v2 configs where they would be unknown

`model2netcdf.SIPNET` -- v2 dropped the "Notes:" line before the header row, so the old skip=1 broke on v2 output. Now auto detects the format. Also handles missing litterWater column (LITTER_WATER flag removed in v2) and converts the 8 new N/CH4 output columns to NetCDF (mineral_N, soil_organic_N, litter_N, N2O, N_leaching, N_fixation, N_uptake, CH4)

added tests for v1 and v2 format coverage, including N2O/CH4 extraction and the missing litterWater edge case.

v1 configs still work exactly as before when `<revision>` is unset or 1.0; and set 2.0 for v2 support

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -18,9 +16,19 @@
...
+#'
+#' If use_v1_format is TRUE, the table has 14 columns:
+#'
+#'  * column 1 is a spatial grid index, always set to 0
+#'  * columns 2-13 are the 12 described above
+#'  * column 14 is soil wetness as a fraction of saturation, always set to 0.6
+#'
+#' These were needed by older versions of Sipnet, but never had any effect on
+#'  PEcAn runs: PEcAn never used gridded runs and always configured Sipnet to
+#'  calculate soil moistuer internally.
```
dlebauer:
```suggestion
#'  calculate soil moisture internally.
```

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -18,9 +16,19 @@
...
+#' Model support for both of these columns was removed from Sipnet >= v2.0.
```
dlebauer:
```suggestion
#' Model support for both the spatial index and soil moisture columns was removed from Sipnet >= v2.0.
```

models/sipnet/R/met2model.SIPNET.R
dlebauer: I favor making v2 the default, but it isn't clear to me what we should use a v1 Boolean rather than a version number or other identifier.  I'm not confident that the current schema will never need to be updated.

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -276,25 +296,29 @@ met2model.SIPNET <- function(in.path, in.prefix, outfolder, start_date, end_date
...
-                 e_a,
-                 ws, # wind
-                 rep(0.6, n)) # put soil water at a constant. Don't use, set SIPNET to MODEL_WATER = 1
-    
+    tmp <- cbind(
+      if (use_v1_format) rep(0, n),
+      yr[1:n],
+      doy[1:n],
+      hr[1:n],
+      rep(dt / 86400, n),
```
dlebauer: I suggest using ud_convert or a centralised dataset of constants. Though that can be part of the met workflow refactor.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -14,10 +14,50 @@
 ##' @author Michael Dietze
 write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs = NULL, IC = NULL,
                                 restart = NULL, spinup = NULL) {
+
+  sipnet_version <- package_version(settings$model$revision, strict = FALSE)
```
dlebauer: Isn't there an output file that already provides version information? Seems like that would be a more reliable source of version information - at least for >= v2

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -14,10 +14,50 @@
...
   writeLines(config.text, con = file.path(settings$rundir, run.id, "sipnet.in"))
+
+  # for v2, allow settings$model to override runtime flags in sipnet.in
+  if (rev_str == "v2") {
+    sipnet_in_path <- file.path(settings$rundir, run.id, "sipnet.in")
+    nc <- as.integer(settings$model$nitrogen_cycle %||% 1)
+    lp <- as.integer(settings$model$litter_pool %||% 1)
+    an <- as.integer(settings$model$anaerobic %||% 1)
+    # NITROGEN_CYCLE requires both LITTER_POOL and ANAEROBIC
+    if (nc == 1 && !(lp == 1 && an == 1)) {
```
dlebauer, line 49: This is enforced by sipnet. In one hand, catching it early makes sense. On the other, it simplifies the pecan code if we delegate consistency checks to sipnet

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -143,11 +183,16 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
 
 
   ### WRITE *.param-spatial
-  template.paramSpatial <- system.file("template.param-spatial", package = "PEcAn.SIPNET")
-  file.copy(template.paramSpatial, file.path(settings$rundir, run.id, "sipnet.param-spatial"))
+  if (rev_str == "v1") {
+    template.paramSpatial <- system.file("template.param-spatial", package = "PEcAn.SIPNET")
```
dlebauer, line 194: Has this type of parameter file ever been used with pecan?

dlebauer (approved):
Thanks for getting this started @infotroph and for bringing this home @divine7022  

I left a few questions & comments, none critical; most are thoughts to consider in future iterations.

models/sipnet/R/write.configs.SIPNET.R
divine7022: good idea in principle, sipnet v2 now has a version.h that has version "2.0.0" (and optionally a git hash) into the binary);
but major limitation would be write.config runs pre execution so there's no output to inspect yet, but that querying the binary's version string could be a future enhancement. Worth noting that sipnet v2 binary does support --version output, so this could be added as a validation step

models/sipnet/R/write.configs.SIPNET.R
divine7022: agreed on the tradeoff.; keeping it for now because it gives clearer error messages before job submission, but noting it's duplicated from sipnet's validateContext()

models/sipnet/R/write.configs.SIPNET.R
infotroph: There are no outputs at this stage -- we have a binary path but haven't tried calling it yet, and conceptually Sipnet might not even exist on the machine writing the configs. Seems simpler to me to have write.configs do what the settings file says rather than try to infer it from any other source.

models/sipnet/R/write.configs.SIPNET.R
infotroph: 👍 To delegating to Sipnet here.

infotroph: And while we're looking at this section:

* How do we feel about  the current named options (`<model><nitrogen_cycle>0</nitrogen_cycle>...`) vs the approach PEcAn.ed2 takes of passing a literal string of command flags (`<model><binary_args>--no-nitrogen-cycle</binary_args>...`)? I think I prefer the named options, but seemed worth asking.
* If using named options, would it make sense to accept them as a block and pass them all without needing to enumerate? something like

```
<model>
  <options>
    <NITROGEN_CYCLE>1</NITROGEN_CYCLE>
    <GDD>0</GDD>
    <NEW_FLAG_SIPNET_HAS_ADDED_SINCE_LAST_REVISION_OF_WRITE.CONFIGS_CODE>1</NEW_FLAG_...>
    ...
  </options>
  ...
```

...which could then be processed as `flag_lines <- paste(names(settings$model$options), "=", settings$model$options, collapse = "\n")`?

If this looks good I'll implement it.

models/sipnet/R/write.configs.SIPNET.R
infotroph: Depends what you mean by "used." We don't process it in any way, but my understanding is that sipnet V1 requires this 1-byte file to exist in the run directory.

models/sipnet/inst/template.param_v2
infotroph: @divine7022 Did your last push resolve this or is there more I need to do here?

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -432,7 +477,8 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
     # 10/31/2017 IF: these were the two assumptions used in the emulator paper in order to reduce dimensionality
     # These results in improved winter soil respiration values
     # they don't affect anything when the seasonal soil respiration functionality in SIPNET is turned-off
-    if(TRUE){
+    # 2025-07-22 CKB: soilRespQ10Cold and baseSoilRespCold were removed from Sipnet V2.0
+    if (rev_str == "v1") {
```
infotroph, line 488: For a future PR: Adding all these version switches into the middle of the existing spaghetti was the right move to keep the initial diff traactable, but we'll definitely pay later in readability. Maybe the next bite-sized step would be to reorder so that we handle all version-specific parameters in one place?

models/sipnet/inst/template.param_v2
divine7022: i am unsure which version i looked when i commented this, but looking at the params in sipnet.c, they are not read in v2 at all (totNitrogen nor microbeNC); however they are in template.param_v1; 
wrt to template.param_v2 no changes needed

models/sipnet/R/write.configs.SIPNET.R
divine7022: correct, as far as I can tell, it's always used this as a single location boilerplate. v2 dropped multi site support entirely so it's not needed there; for v1 needs to keep copying it since signet v1 expects the file to exist

models/sipnet/R/write.configs.SIPNET.R
divine7022: this is a good design decision, since new SIPNET flags don't require code changes to write.config;
and for me it seems like could be deferred to follow up PR but shouldn't be mixed into the current one (it changes the XML schema and requires documentation)
make sense ?

infotroph: > deferred to follow up PR

Fine with me

> changes the XML schema 

No more than this PR does already -- we're already going from not expecting any options in the XML to expecting them directly inside `model`. But it's probably worth a review of how other models do this so that we can keep some uniformity.

models/sipnet/R/met2model.SIPNET.R
divine7022: @dlebauer addressed -- https://github.com/PecanProject/pecan/pull/3813/commits/c1cc1c68ad05a2caea1b54c0a0e61f1c8b5bffa3

models/sipnet/R/met2model.SIPNET.R
divine7022: addressed -- https://github.com/PecanProject/pecan/pull/3813/commits/c1cc1c68ad05a2caea1b54c0a0e61f1c8b5bffa3

models/sipnet/NEWS.md
```diff
@@ -4,7 +4,9 @@
 * Initial support for SIPNET v2.0, whose features include simplified input files,
     simulation of management events, tracking of N cycle components, and anaerobic CH4 generation.
 * Breaking: `met2model.SIPNET` now writes 12-column clim files (as expected by Sipnet >= v2.0) by default.
-	To get the previously standard 14-column output, set `use_v1_format = TRUE`.
+	To get the previously standard 14-column output, set `sipnet_version = "1"`.
+* Replaced hardcoded `86400` (seconds per day) in `met2model.SIPNET` with
+	`PEcAn.utils::ud_convert(1, "day", "s")` for clarity.
```
infotroph: I tend to leave changes this small out of NEWS -- they're not going to have any user-visible effects

models/sipnet/R/write.configs.SIPNET.R
divine7022: haven’t pushed my local changes for this yet, I am commenting here since the context flows -- https://github.com/PecanProject/pecan/pull/3813/commits/2d7559c9382eba9036e6c3d1f5d5aafa1ed3158b

models/sipnet/NEWS.md
divine7022: good catch!

models/sipnet/R/write.configs.SIPNET.R
divine7022: heads up: 
https://github.com/PecanProject/pecan/actions/runs/22737237619/job/65941705561?pr=3813

fixed version parsing to handle non standard DB revisions like "136", "unk", "git", now only recognizes major versions 1 and 2 as valid sipnet versions; everything else defaults to v1

https://github.com/PecanProject/pecan/pull/3813/commits/f99032d65b1db33af2188cfba4e7cae3268f3b19

CHANGELOG.md
```diff
@@ -16,16 +16,22 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - Added datasets to `PEcAn.data.land` 
   *  `landiq_crop_mapping_codes` dataset mapping LandIQ crop classification codes to human-readable crop names.
   *  `bism_kc_by_crop` dataset containing BISm crop coefficient schedules and stage timing references for use in ET estimation, including columns that map to LandIQ class and subclass.
+- Add SIPNET v2 support in PEcAn.SIPNET; v2 parameter template (`template.param_v2`) with
+  73 parameters, v2 input template (`sipnet.in_v2`) with runtime flags, nitrogen cycle / methane
+  NetCDF output in `model2netcdf.SIPNET`, and runtime flag injection in `write.config.SIPNET`
 
 ### Fixed
+- Fixed `model2netcdf.SIPNET` to auto detect v2 output format (no "Notes:" header line),
+  handle missing `litterWater` column, and correctly convert N/CH4 outputs to NetCDF
```
```suggestion
- PEcAn.SIPNET gains support for SIPNET v2, whose features includes management events, nitrogen cycle tracking, explicit N2O and methane fluxes, runtime setting of feature flags, and changes to the parameter set (now 73 parameters). SIPNET v1 is still fully supported, but workarounds for bugs in the legacy `sipnet.unk` version have been removed.

### Fixed
```

CHANGELOG.md
```diff
@@ -16,16 +16,22 @@ For more information about this file see also [Keep a Changelog](http://keepacha
...
 
 ### Changed
 - `assign.treatments` has been renamed to `assign_treatments` and moved from `PEcAn.utils` to `PEcAn.MA` since that's the only place where it's used.
 - With new `PEcAn.MA::meta_analysis_standalone` function, `PEcAn.MA::run.meta.analysis.pft` now saves all files all at once _after_ the complete meta-analysis runs (and only if it is successful, including prior and posterior checks), rather than saving intermediate objects (like "JAGS-ified" data) as they are created.
 - Sensitivity analysis and ensemble runs now generate separate input design matrices with appropriate dimensions, fixing dimension mismatch errors in multisite workflows. (#3708)
 - Generated runs are now stored in a `runs_manifest.csv` file in the output directory instead of modifying `samples.Rdata` (#3708)
 - SDA workflows now maintain joint input sampling, via internal calls to `generate_joint_ensemble_design()` (#3634).
-
+- `write.config.SIPNET` now injects NITROGEN_CYCLE, LITTER_POOL, and ANAEROBIC runtime flags
+  for SIPNET v2 runs and guards v1 only parameters from being written to v2 configs.
```
infotroph: I'm inclined to leave this out of CHANGELOG on the grounds it's already part of the "support for V2" chunk above, but am not strongly convinced. Any objections?
```suggestion
```

models/sipnet/NEWS.md
```diff
@@ -1,7 +1,20 @@
 # PEcAn.SIPNET 1.10.0.9000
 
 * Removed `tests/Rcheck_reference.log`, which was used to ignore historic check messages that have now been fixed.
-
+* Initial support for SIPNET v2.0, whose features include simplified input files,
+    simulation of management events, tracking of N cycle components, and anaerobic CH4 generation.
+* Breaking: `met2model.SIPNET` now writes 12-column clim files (as expected by Sipnet >= v2.0) by default.
+	To get the previously standard 14-column output, set `sipnet_version = "1"`.
+* NITROGEN_CYCLE, LITTER_POOL, and ANAEROBIC enabled by default for SIPNET v2 runs.
+* Fixed `model2netcdf.SIPNET` for v2 output header format (no Notes line; SIPNET #267).
```
```suggestion
* `model2netcdf.SIPNET` now handles both V1 and v2 output header format (v2 has no Notes line; SIPNET #267).
```

models/sipnet/NEWS.md
```diff
@@ -1,7 +1,20 @@
...
+* Removed 13 obsolete v1 parameters from `template.param_v2` (microbeInit, qualityLeaf, etc.)
```
```suggestion
* Removed 13 obsolete v1 parameters from the v2 parameter template `template.param_v2` (microbeInit, qualityLeaf, etc.)
```

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -38,16 +46,37 @@
...
-#' @param var.names character: list of variable names to be extracted. Default is NULL.
+#' @param var.names character: list of variable names to be extracted.
+#'  Default NULL uses all variables from input file
 #' @param overwrite should existing files be overwritten
 #' @param verbose should the function be very verbose
 #' @param year.fragment the function should ignore whether or not the data is
 #'  stored as a set of complete years (such as for forecasts).
+#' @param sipnet_version Version string for SIPNET (e.g. "1", "2", "2.0").
+#'  Versions < 2 write the legacy 14-column format; >= 2 writes 12 columns.
+#'  Default "2"
```
infotroph:
```suggestion
#' @param sipnet_clim_version Version of SIPNET clim file format to generate.
#'  Default "v2" writes 12 columns, "v1" writes the legacy 14-column format.
```

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -38,16 +46,37 @@
...
+met2model.SIPNET <- function(in.path,
+                             in.prefix,
+                             outfolder,
+                             start_date,
+                             end_date,
+                             var.names = NULL,
+                             overwrite = FALSE,
+                             verbose = FALSE,
+                             year.fragment = FALSE,
+                             sipnet_version = "2",
```
infotroph:
```suggestion
                             sipnet_clim_version = c("v2", "v1"),
```

CHANGELOG.md
divine7022: make sense; point coved in NEWS.md

models/sipnet/R/met2model.SIPNET.R
```diff
@@ -38,16 +46,37 @@
...
+                             ...) {
+
+  # parse version to decide output format, same approach as write.config.SIPNET
+  sv <- sipnet_version
+  if (grepl("^[0-9]+$", sv)) sv <- paste0(sv, ".0")
+  sv <- package_version(sv, strict = FALSE)
+  use_legacy_format <- is.na(sv) || sv < "2.0"
```
infotroph:
```suggestion
   use_legacy_format <- match.arg(sipnet_clim_version) == "v1"
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -14,10 +14,59 @@
...
+  # determine which SIPNET major version we're configuring for.
+  # settings$model$revision may be a SIPNET version ("1", "2", "2.0")
+  # or a internal model identifier ("136", "unk", "git", NULL).
+  # Only "1" and "2" (and their dotted forms) are valid SIPNET versions;
+  # everything else defaults to v1 for backward compatibility
+  rev_str <- "v1"
+  if (!is.null(rev_raw) && nzchar(rev_raw)) {
+    if (grepl("^[0-9]+$", rev_raw)) {
+      rev_raw <- paste0(rev_raw, ".0")
+    }
```
infotroph: This feels like a lot of complication to me, especially since it lets us accept a bare integer but not `"v<integer>"`; IMO the latter is quite a bit more widely used.  

I wonder if we should instead require a numeric version for anything but an enumerated set of legacy versions (maybe just 136 and git, dropping unk?)

models/sipnet/NEWS.md
```diff
@@ -2,7 +2,20 @@
...
+  and added 16 new nitrogen cycle, anaerobic, and methane parameters.
+* Added nitrogen cycle (mineral_N, soil_organic_N, litter_N, N2O, N leaching, N fixation,
+  N uptake) and methane (CH4) output conversion to NetCDF in `model2netcdf.SIPNET`.
+* Fixed crash when `litterWater` column absent in v2 output (LITTER_WATER removed in v2).
+* `write.config.SIPNET` now validates runtime flag dependencies (NITROGEN_CYCLE requires
+  LITTER_POOL and ANAEROBIC) and guards v1 only parameters (litterWHC, litWaterDrainRate,
+  litterWFracInit, microbeInit, m_ballBerry) from being set when using v2 templates.
```
infotroph, line 18:
```suggestion
  litterWFracInit, microbeInit, m_ballBerry) from being set when using v2 templates.
* Removed workarounds for column naming bugs in output from long-outdated legacy
  Sipnet version `sipnet.unk`.
  
```

models/sipnet/R/write.configs.SIPNET.R
divine7022: I think the right approach would be to simplify to accept "v1", "v2", "1", "2" and default everything else to v1

divine7022: may be 
```
rev_raw <- settings$model$revision %||% "1"
# accept "v1", "v2", "1", "2", "1.0", "2.0"; legacy DB identifiers (136, git) default to v1
rev_str <- if (grepl("^v?2(\\.0)?$", rev_raw, ignore.case = TRUE)) "v2" else "v1"
```
make sense ?

models/sipnet/NEWS.md
divine7022: done

models/sipnet/NEWS.md
divine7022: done

models/sipnet/R/write.configs.SIPNET.R
infotroph: > right approach would be to simplify to accept "v1", "v2", "1", "2"

If you're saying accept those but reject x.y.z numeric versions, I disagree. `settings$model$revision` is declaring the version of Sipnet that will actually run when you call `settings$model$binary`, and needs to be interpretable by any function that needs it and not just run.write.configs. 

Say hypothetically we do a couple releases with restart support and then decide we want to change its API without breaking support for the previous ones; we'd want to be able to add conditions like `if (settings$model$revision >= "2.0.3") {...` into write.restart without breaking the write.configs v1/v2 logic.

models/sipnet/R/write.configs.SIPNET.R
infotroph: Filed as #3854

models/sipnet/R/write.configs.SIPNET.R
divine7022: enumerated known legacy identifiers ("136", "git") and map them to v1 explicitly;
sipnet_version variable is also available in function scope for any future >= comparisons (e.g.  >= "2.0.3" check for api changes)

https://github.com/PecanProject/pecan/pull/3813/commits/3f367149690c075151b835ca4abad92fd7520232

divine7022:
https://github.com/PecanProject/pecan/actions/runs/22837731305/job/66245148269?pr=3813

is this an upstream issue with how the sipnet v1.3.0 binary interacts with the current gha runner ?

infotroph:
@divine7022 Took me a while to track this down (had to run the Github action locally), but: `Read unexpected input item: GROWTH_RESP`, because the GHA Bety returns revision "102319", which compares >= "2.0", so PEcAn.SIPNET uses the v2 parameter template.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -14,10 +14,57 @@
 ##' @author Michael Dietze
 write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs = NULL, IC = NULL,
                                 restart = NULL, spinup = NULL) {
+
+  rev_raw <- settings$model$revision
+  legacy_v1 <- c("136", "git")
```
```suggestion
  legacy_v1 <- c("102319", "136", "r136", "ssr", "git")
```

PR #3816, opened 2026-02-16
write n2o and ch4 flux variables to SIPNET netCDF

divine7022:
## Description

- updates model2netcdf.SIPNET to include n2o and ch4 flux variables
- add test

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/sipnet/R/model2netcdf.SIPNET.R
```diff
@@ -178,6 +178,15 @@ model2netcdf.SIPNET <- function(outdir, sitelat, sitelon, start_date, end_date,
...
     output[["GWBI"]] <- (sub.sipnet.output$woodCreation * 0.001) / 86400 ## kgC/m2/s - this is daily in SIPNET
     output[["AGB"]] <- (sub.sipnet.output$plantWoodC + sub.sipnet.output$plantLeafC) * 0.001 # Total aboveground biomass kgC/m2
+    if ("n2oFlux" %in% names(sub.sipnet.output)) {
+      output[["N2O_flux"]] <- (sub.sipnet.output$n2oFlux * 0.001) / timestep.s
+      # convert g N m-2 per timestep -> kg N m-2 s-1
+    }
+    ch4_col <- intersect(c("ch4Flux", "ch4"), names(sub.sipnet.output))
+    if (length(ch4_col) > 0) {
+      output[["CH4_flux"]] <- (sub.sipnet.output[[ch4_col[1]]] * 0.001) / timestep.s
+      # convert g C m-2 per timestep -> kg C m-2 s-1
```
infotroph: Let's use one standard name per column here and let the development churn stay inside Sipnet.

Also while I'm nitpicking: Since these are the first optional columns in model2netcdf it feels worth putting them in their own block with an explanatory comment

```suggestion

    # Columns only present in Sipnet >= v2 with N and methane turned on
    if ("n2o" %in% names(sub.sipnet.output)) {
      output[["N2O_flux"]] <- (sub.sipnet.output$n2o * 0.001) / timestep.s
      # convert g N m-2 per timestep -> kg N m-2 s-1
    }
    if ("ch4" %in% names(sub.sipnet.output)) {
      output[["CH4_flux"]] <- (sub.sipnet.output$ch4 * 0.001) / timestep.s
      # convert g C m-2 per timestep -> kg C m-2 s-1

```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
+test_that("model2netcdf.SIPNET produces netCDF from v2 output with GHG fluxes", {
+  outdir <- withr::local_tempdir(pattern = "sipnet_out_")
+  rundir <- withr::local_tempdir(pattern = "sipnet_run_")
+
+  # minimal sipnet.param — model2netcdf only reads leafCSpWt for LAI
```
infotroph: Not for this PR, but: Given we go to the trouble of reading `leafCSpWt` from the param file, why does it get combined with a constant `leafC <- 0.48` instead of reading `cFracLeaf` from the very next param line?

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+  writeLines(
+    c("plantWoodInit\t30000\t0\t6600\t14000\t200",
+      "leafCSpWt\t32\t0\t13\t500\t0"),
```
infotroph: Sipnet v2 drops the extra numeric columns and I would hope it's safe to do that here too:
```suggestion
    c("plantWoodInit\t30000",
      "leafCSpWt\t32"),
```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+    n2oFlux = 0.002,
+    ch4Flux = 0.001
+  )
+
+  out_path <- file.path(outdir, "sipnet.out")
+  writeLines(
+    "Notes: units in g/m2 per timestep; water in cm",
+    out_path
+  )
+  suppressWarnings(
```
infotroph: Why suppress warnings here? If this step fails that seems like a problem that we should know about.

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+    write.table(sipnet_dat, file = out_path, append = TRUE,
+                row.names = FALSE, quote = FALSE, sep = "\t")
+ )
+
+  # outdir must contain "/out/" so the gsub to find "/run/" works
+  # we've set up rundir separately, so patch the path convention:
+  # model2netcdf does gsub("/out/", "/run/", outdir) to find sipnet.param
+  # use the structure: <base>/out/run1  and  <base>/run/run1
```
infotroph: Gross! This seems worth fixing in the function, but for testing purposes I recommend following the convention from the start of the test instead of patching around it -- I'll scroll back up and leave a suggestion at the top of the file; once that's applied all these files can just be written into `rundir` or `outdir` with no further fuss.

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
+test_that("model2netcdf.SIPNET produces netCDF from v2 output with GHG fluxes", {
+  outdir <- withr::local_tempdir(pattern = "sipnet_out_")
+  rundir <- withr::local_tempdir(pattern = "sipnet_run_")
```
infotroph:
```suggestion
  tmpdir <- withr::local_tempdir()
  # model2netcdf does gsub("/out/", "/run/", outdir) to find sipnet.param
  outdir <- file.path(tmpdir, "out", "run1")
  rundir <- file.path(tmpdir, "run", "run1")
  dir.create(outdir, recursive = TRUE)
  dir.create(rundir, recursive = TRUE)
```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+  # outdir must contain "/out/" so the gsub to find "/run/" works
+  # we've set up rundir separately, so patch the path convention:
+  # model2netcdf does gsub("/out/", "/run/", outdir) to find sipnet.param
+  # use the structure: <base>/out/run1  and  <base>/run/run1
+  base <- withr::local_tempdir(pattern = "sipnet_base_")
+  real_outdir <- file.path(base, "out", "run1")
+  real_rundir <- file.path(base, "run", "run1")
+  dir.create(real_outdir, recursive = TRUE)
+  dir.create(real_rundir, recursive = TRUE)
+  file.copy(out_path, file.path(real_outdir, "sipnet.out"))
```
infotroph: (Most of these lines just move to the top of the file rather than go away, but this way it's all set up once and no files need to be moved after creation)
```suggestion
```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+  writeLines(
+    c("plantWoodInit\t30000\t0\t6600\t14000\t200",
+      "leafCSpWt\t32\t0\t13\t500\t0"),
+    file.path(real_rundir, "sipnet.param")
+  )
```
infotroph: duplicates lines 6-9
```suggestion
```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+
+  suppressMessages(
```
infotroph: Same question as above -- what messages are being suppressed here?

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+    model2netcdf.SIPNET(
+      outdir     = real_outdir,
```
infotroph:
```suggestion
      outdir     = outdir,
```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+      revision   = "r136"
+    )
+  )
+
+  nc_file <- file.path(real_outdir, "2002.nc")
+  expect_true(file.exists(nc_file))
+
+  nc <- ncdf4::nc_open(nc_file)
+  on.exit(ncdf4::nc_close(nc), add = TRUE)
+  vars <- names(nc$var)
```
infotroph: Tests like this seem like a great use case for the `netcdf2df()` I proposed in #3788 (which isn't merged yet, so no foul for not using it here!) -- one line to slurp the test file into a df instead of needing to remember these incantations every time.

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+  expect_equal(ch4, rep(0.001 * 1e-3 / ts_s, n), tolerance = 1e-12)
+  expect_equal(gpp, rep(0.30  * 1e-3 / ts_s, n), tolerance = 1e-12)
+
+  expect_equal(nc$var$N2O_flux$units, "kg N m-2 s-1")
+  expect_equal(nc$var$CH4_flux$units, "kg C m-2 s-1")
+  expect_equal(nc$var$GPP$units,      "kg C m-2 s-1")
+
+  # -- dimensions --
+  expect_equal(nc$dim$time$len, n)
+  expect_true(grepl("days since 2002", nc$dim$time$units))
```
infotroph: nit: I think `expect_match` would be more idiomatic here
```suggestion
  expect_match(nc$dim$time$units, "days since 2002")
```

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,197 @@
...
+})
+
+
+test_that("model2netcdf.SIPNET omits N2O/CH4 when columns absent (backward compat)", {
```
infotroph: And now all my comments above repeat, suggesting it'd be worth pulling the "create tempdir, write files, read files" piece out into a test fixture.

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
divine7022: the warning is "appending column names to file"; R warns when we write column headers to an already existing file (cuz could end up with headers in the middle of data). In this test it's expected behavior (first writeLines the notes header, then append the data with col headers). Anyways make sense, removed

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
divine7022: model2netcdf.SIPNET calls print(paste("--- Processing year: ", y)) for each year

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
divine7022: thanks! opened PR https://github.com/PecanProject/pecan/pull/3820

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
infotroph: Also: is the plantWoodInit line needed for this test to work?

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
infotroph: I think Sipnet will shortly (or has already?) drop the Notes line, so that may be another change needed in model2netcdf but will let us simplify here.

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
```diff
@@ -0,0 +1,102 @@
...
+  on.exit(ncdf4::nc_close(nc), add = TRUE)
+  vars <- names(nc$var)
+
+  expect_true("N2O_flux" %in% vars)
+  expect_true("CH4_flux" %in% vars)
+  expect_true(all(c("GPP", "NEE", "TotalResp", "TotSoilCarb") %in% vars))
+
+  n2o <- as.numeric(ncdf4::ncvar_get(nc, "N2O_flux"))
+  ch4 <- as.numeric(ncdf4::ncvar_get(nc, "CH4_flux"))
+  gpp <- as.numeric(ncdf4::ncvar_get(nc, "GPP"))
```
infotroph: All of these should be stored as numeric already, and if they aren't then we want to know (as the expect_equal below will do if we skip coercion)

```suggestion
  n2o <- ncdf4::ncvar_get(nc, "N2O_flux")
  ch4 <- ncdf4::ncvar_get(nc, "CH4_flux")
  gpp <- ncdf4::ncvar_get(nc, "GPP")
```

infotroph (approved):

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
divine7022: right! that these should already be numeric , as.numeric() was masking the type check. But ncvar_get() returns a 1-D array (carries dim attribute), so bare comparison with rep() fails on attributes. Switched to as.vector(), which strips the array dimension without coercing type

models/sipnet/tests/testthat/test-model2netcdf.SIPNET.R
divine7022: no, `plantWoodInit` is not needed. model2netcdf.SIPNET only reads `leafCSpWt` from the param file. I will trim `sipnet.param` down to just that one line

Issue #3764, opened 2026-01-08
Irrigation management: flooded rice


mdietze:
## Description

Building on #3762, flood-irrigated rice represents a MVP-relevant edge case. Specifically, we need a more nuanced control over the runoff term in the irrigation bucket model (in general this means turning runoff off, but there are times when it would be allowed) and setting minWHC > WHC so that irrigation maintains a flooded state. To add to the complexity, sometimes rice fields are flooded during the fallow period for weed control or conservation purposes (e.g. habitat for migratory birds).

Yang's analyses of SIPNET for wetlands suggests the model consistently underestimates ET, which we believe is driven by runoff (which occurs by default in SIPNET). To get these flood irrigation values to "stick" in the model requires adding the ability to specify when and where runoff gets turned off in the model.

ashiklom:
A crude version of this is implemented in #3821. You can enforce a permanently wet field with `whc_min_frac = 1` (`w_min == whc`), and there's a seepage term (required if `is_rice = TRUE`) that slowly the soil at a constant rate. My sense is that this crudely approximates the management behavior of maintaining a flooded field in the face of ET + seepage losses. However, I think rice might need its own dedicated logic that I'm going to punt to a separate PR:

```r
# In calc_water_balance(...)
if (is_rice) {
  return(calc_water_balance_rice(...))
}
# resume normal water balance logic...
```

ashiklom:
A less crude version of this is implemented at the link below. In my tests, with `flood_target = 125, flood_min = 62.5, flood_max = 175, and seepage = 2.5` (all in mm or mm/day),  this leads to approximately weekly irrigation events for rice fields. Conversely, with the simple bucket model implementation, we got basically daily irrigation to compensate for constant ET + seepage losses. I asked Claude and it suggested that nearly constant rice irrigation is more correct in practice, but that the more important thing was to get the annual total correct. I also asked Claude what typical rice irrigation totals for CA look like, and its numbers (1000-1300 nm / season) seem to agree reasonably well with my results (400--1411 mm/year across all sampled parcels, 999--1135 mm/year for the middle 50%, 1070mm mean).

https://github.com/ashiklom/pecan/blob/cimis-et-complete/modules/data.land/R/water_balance.R

Issue #3763, opened 2026-01-08
Irrigation management inputs: soils and rooting


mdietze:
## Description

Current irrigation code uses a simple bucket model for hydrology but assumes the same size "bucket" everywhere and the same tolerance to water stress for every crop.

## Proposed Solution

1. Use site-specific soils information (e.g. from soils.nc files) to set soil saturation and maximum soil depth, initially using means but ideally ensembled (with ensemble IDs matching soil ensembles being assigned to model runs)

2. Use crop identity to set crop-specific rooting depth and minimum water tolerance. Crop specific irrigation recommendations (e.g.,  management allowable depletion, MAD) should be available from CIMIS (e.g., https://cimis.water.ca.gov/Content/pdf/Irrigation_Scheduling_Using_CIMIS.pdf). The current test value (allowing SM to drop to 15% of WHC) looks too low relative to what I'm reading in CIMIS, which suggests a number around 50-60% WHC (40-50% MAD) for sensitive crops and 20% minWHC (80% MAD) for deeply rooted perennials on ideal soils. As an initial starting point for defaults I'd recommend we set ~25% minWHC for woody PFTs and 50% minWHC for annuals. Maybe something in between for irrigated pasture and other non-woody perennials.

3. Use soil saturation and min(rootingDepth,max soil depth) to set overall water holding capacity (WHC)

4. Finally, the irrigation management allows the specification of whether irrigation is done above or below canopy. For California this info should be in the harmonized LandIQ. When applying irrigation above canopy we also need to add a correction factor for canopy interception and direct evaporation.

## Additional Context 

#3762

ashiklom:
Update as of today:
1. Not started.
2. Partially done. I have a list of MAD values for each crop, but not rooting depth. (#3821)
3. Supported in `calc_water_balance`, but not implemented in the example workflow yet. (#3821).
4. Not started.

PR #3821, opened 2026-02-18
CIMIS irrigation workflow, v1

ashiklom:
See #3762. This is a draft with significant stylistic and performance issues. See the `qmd` file for an end-to-end overview of the functionality.

I definitely want to refine this, but I wanted to open it as a progress marker and to make confirm I'm on the right track.


modules/data.land/NAMESPACE
```diff
@@ -33,8 +39,12 @@ export(gSSURGO.Query)
 export(generate_soilgrids_ensemble)
 export(get.attributes)
 export(get.soil)
+export(get_cimis_region)
+export(get_crop_for_points)
+export(get_landiq_parcel_ids)
 export(get_resource_map)
 export(get_veg_module)
+export(growing_season_dates)
```
infotroph: Posted before I read any actual function definitions: This seems like a lot of new exports. Are some of these actually helpers that should stay internal?

modules/data.land/data-raw/kc_coefficients.R
```diff
@@ -0,0 +1,35 @@
...
+  date_a="c",
+  date_b="c",
+  date_c="c",
+  date_e="c",
+  kc1="d",
+  kc2="d",
+  kc3="d",
+  code="i"
+)
+cimis_kc_crops <- readr::read_csv(
```
mdietze: How are these coefficients different from those added in #3765

modules/data.land/R/water_balance.R
```diff
@@ -0,0 +1,84 @@
+#' Calculate water balance
+#'
+#' This is the core water balance calculation that operates on primitive
+#' numeric vectors for easy testing and debugging.
+#'
+#' @param et Vector of evapotranspiration values (mm/day)
+#' @param precip Vector of precipitation values (mm/day)
+#' @param whc Water holding capacity (mm), default 500
+#' @param w_min_frac Fraction of WHC for minimum water level, default 0.15
```
mdietze: We learned latter that the typical values are actually a good bit higher. See Issue #3763

modules/data.land/R/water_balance.R
```diff
@@ -0,0 +1,84 @@
+#' Calculate water balance
+#'
+#' This is the core water balance calculation that operates on primitive
+#' numeric vectors for easy testing and debugging.
+#'
+#' @param et Vector of evapotranspiration values (mm/day)
```
mdietze: for the sake of documentation, is this a time vector at a single site or a vector of sites at a single time?

modules/data.land/R/water_balance.R
```diff
@@ -0,0 +1,84 @@
...
+#' @return List with vectors: W_t (water balance), irr (irrigation), runoff
+#' @export
+calc_water_balance <- function(et, precip, whc = 500, w_min_frac = 0.15) {
+  n <- length(et)
+  if (length(precip) != n) {
+    stop("et and precip must have the same length")
+  }
+
+  w_min <- w_min_frac * whc
+  field_capacity <- whc / 2
```
mdietze: a good future direction would be to make the initial soil moisture an argument so that, in the future, we could better initialize the algorithm (whether that be from the previous year's end point, a spin-up, or remote sensing)

modules/data.land/R/water_balance.R
```diff
@@ -0,0 +1,84 @@
...
+      W_prev <- field_capacity #nolint: object_name_linter
+    } else {
+      W_prev <- W_t[t - 1] #nolint: object_name_linter
+    }
+
+    W0_t[t] <- W_prev + precip[t] - et[t] #nolint: object_name_linter
+
+    irr[t] <- max(w_min - W0_t[t], 0)
+
+    runoff[t] <- max(W0_t[t] - whc, 0)
```
mdietze: note that another future direction would be to prevent runoff on flood irrigation sites. see #3764

mdietze (commented):
overall looks to be on the right track to knock of #3762. Before moving on the #3763 and #3764 I think the main thing that remains is to scale up the calculations to the full set of CA crop parcels using the HLS phenology data extracted by @sarahkanee. That said, if we can sneak any bits of #3763 into here I'd prioritize better defaults for minWHC

modules/data.land/data-raw/kc_coefficients.R
ashiklom: I missed that PR when I was working on this. Oops. Definitely a lot of duplicated logic here and in that PR that I'll need to remove from mine.

I was using coefficients linked directly from the CIMIS website. I haven't done a comprehensive cross reference, but from a quick skim, the ones in the other PR look like an updated set from the same general source. I'm guessing we'll probably want to use those instead of the ones I pulled.

ashiklom:
I made the mistake of initially developing this completely offline of PEcAn and, in doing so, completely missed #3765, which implements a sizeable chunk of this work (and does so with better code). Won't be that mistake again! So we should definitely hold on merging this until I refactor my work to use that code and eliminate my redundant implementations.

That said, the refactor is a good opportunity to bang out the two companion issues mentioned above, at least in first pass form, since they seem like only minor straightforward tweaks to the water balance logic.

Can you point me to where the HLS phenology data are? If so, I can integrate that at the same time as the other changes.

modules/data.land/R/water_balance.R
ashiklom: resolved in ad1b49b

ashiklom:
I refactored this to use the BISM coefficients from #3765 . A nice side-effect is that the vignette now runs much faster, and the logic is simpler. I also vectorized the CHIRPS download code so we only read one file per year (rather than re-reading the same year's file separately for every day, like we were doing before) and updated some of the water balance docs.

I still need to clean up the deprecated CIMIS data (I'll do that as a rebase to avoid inflating the repo size).

I also need to make some of the changes to the water balance parameterization as suggested by @mdietze.

Finally, and most importantly, I still need to integrate the HLS-based phenology information.

modules/data.land/R/water_balance.R
```diff
@@ -0,0 +1,183 @@
...
+    W0 <- W_prev + precip[t] - et[t] - seepage
+
+    # If W0 falls below w_min (e.g., high ET and seepage; low precip), irrigate
+    # to field capacity.
+    if (W0 < w_min) {
+      irr[t] <- field_capacity - W0
+      W0 <- field_capacity
+    } else {
+      irr[t] <- 0
+    }
```
ashiklom: @mdietze (or others) Flagging that this logic is different from what we had before.

The Python code I was using previously irrigated back up to the _irrigation trigger threshold_ (in this code, `w_min`). However, from a bit of reading, it seems the more common practice for most crops is to irrigate back up to _field capacity_. This results in a smaller number of large, discrete irrigation events.

Let me know what you think.

modules/data.land/R/water_balance.R
ashiklom: Current version has a more sensible default (0.375) and, more importantly, has the ability to pull crop-specific values from a table (based on FAO-56 Irrigation manual).

modules/data.land/R/water_balance.R
ashiklom: `calc_water_balance` now accepts initial soil moisture as an optional argument. If `NULL`, it defaults to field capacity.

modules/data.land/R/water_balance.R
ashiklom: This PR includes primitive support for quasi-flooded-rice-irrigation (rice irrigation threshold = field_capacity (`whc_min_frac = 1.0`); for rice fields specifically, implement gradual seepage loss on top of ET losses).

ashiklom:
OK, I think I'm ready to call this PR done. It seems like a lot, but I'm hoping the changes are relatively self-contained and straightforward to review.

My suggestion for reviewing this is to start with the three new scripts in `inst` that show the relevant functionality and how it all fits together. From there, you can inspect the implementation details of individual functions.

The demonstration scripts are as follows:

- `ssurgo-soil-inputs.R` --- a quick test to confirm grabbing arbitrary data from the SSURGO API. This includes the new `ssurgo_mukey_*` functionality.
- `compare-cimis-openet.R` --- a comparison of CIMIS ET and OpenET ET data, including a demonstration of how to use the functions.
- `CIMIS-event-files.R` --- a complete workflow that
    - starts from design points and dates
    - gets ET from CIMIS
    - gets precip from CHIRPS
    - merges design points with LandIQ to get crop types
    - merges with BISM KC data to get Kc coefficients (to get actual ET rates)
    - merges with the **new** `crop_whc` table, which contains rooting depth and MAD (`whc_min_frac`) values for various crops
    - extracts precalculated available water content (AWC) from SSURGO for the sites, and combines with rooting depth information to get WHC for each site
    - calculates actual ET from CIMIS ETref by using the crop-specific Kc coefficients
    - joins everything together
    - performs water balance calculations
    - writes SIPNET event files based on the irrigation information

The most important science thing to check here is the `calc_water_balance` function. Per my comment above, the logic here is **changed** from the original prototype Python implementation such that, when you dip below the irrigation threshold (MAD), you irrigate back up to _field capacity_ (_not_ to MAD). This leads to larger and more episodic irrigation. From my (limited) research on agriculture, this seems closer to what farmers actually do, but you all should weigh in here.

Note also that I was unaware of #3643 when I wrote the SSURGO mukey functionality, so I inadvertently recreated a (possibly worse) version of `soilDB::mukey.wcs`. I'll remove my code as soon as I confirm that it does what I need.

EDIT: I tried out `soilDB::mukey.wcs` and poked around the `soilDB` docs. From a cursory look, the `soilDB` functions are all structured around querying bounding boxes, but I don't see support for querying individual points. If you pass a sparse list of coordinates to `soilDB::mukey.wcs`, it takes the bounding box of the polygon and returns the result, which is much more data than you need and will fail if the total AOI is greater than 10,100 km2 (e.g., if I want to sample 10 sites across CONUS, or even 10 sites spread out over all of California). I also don't see a way to overcome the spatial extent limit except splitting up requests by hand. My `ssurgo_mukeys_bigbbox` handles this directly by splitting up a large bounding box into smaller cells guaranteed to be less than the spatial limit and then querying those in parallel.

mdietze (approved):
@ashiklom there's a LOT here so I had to skip parts. A few quick questions before we merge

1. Has this workflow been run to produce outputs (e.g. on the BU server) and if so is that just for the anchor sites or statewide? Please share file paths on Slack as applicable.
2. If this has only been run for anchor sites, is there anything preventing a scaling up to statewide or any improvements you'd want to make before doing so?
3. In writing events, are you still aggregating irrigation needs up to a weekly timescale or does the switch to irrigating to FC alleviate this need?

modules/data.land/R/water_balance.R
```diff
@@ -0,0 +1,209 @@
...
+#' This function operates in *relative* WHC space, where:
+#' - `w = 0` represents wilting point (no plant-available water)
+#' - `w = whc` represents field capacity (maximum plant-available water)
+#' - `whc = (field_capacity - wilting_point) * rooting_depth` (the plant-available range)
+#'
+#' The handling of rice here is crude and primitive: setting `whc_min_frac` =
+#' 1.0 (as set in `crop_whc` for rice) means a near-constant need for
+#' irrigation to balance ET + seepage, which roughly mimics the behavior of
+#' maintaining a standing flood. However, proper treatment of rice requires
+#' maintaining a field *above* field capacity, and has other complications.
```
mdietze: Similar to how non-rice is irrigated up to FC, it looks like the average water depth for rice irrigation is 10-15cm. Could we implement a similar assumption here (i.e., when water drops below saturation, irrigate up to 10cm flood)

mdietze: That said, this may be a place where CARB and others are interested in exploring alternative scenarios, such as Alternative Wetting and Drying (AWD) -- for example allowing the water to drop to 15 cm below the surface before re-flooding to 5 cm above the surface. It may require more thought to determine how to best implement irrigation scenario planning.

ashiklom:
1. No, because the current outputs are fake and potentially somewhat nonsensical in the absence of real phenology/planting inputs (per your requests to integrate Sarah's files). All development of this has been local. If the core logic of the workflow looks sound, and once I have a pointer to real files that can be used for phenology and/or planting dates, I can try to scale it up on the BU HPC.

2. Strictly speaking, no limitations. Any memory and compute bottlenecks could be alleviated by just naive parallelization.

3. Weekly aggregation is still there. I have no idea why it was there originally (some SIPNET limitation/expectation?), so I can't comment on whether it's still appropriate. My intuition is that, with more episodic irrigation events under the new logic, it's no longer necessary. If so, I'm happy to remove it.

modules/data.land/inst/CIMIS-event-files.R
```diff
@@ -0,0 +1,370 @@
...
+  tidyr::fill(
+    "CLASS",
+    "SUBCLASS",
+    "crop_name",
+    .direction = "downup",
+    .by = "parcel_id"
+  ) |>
+  dplyr::mutate(date = purrr::map2(year, season, fill_season)) |>
+  tidyr::unnest(date) |>
+  dplyr::filter(date %in% !!dates)
```
ashiklom: @mdietze Here's where I convert LandIQ "season" indicators to real dates as needed for downstream water balance calculations. The current implementation is obviously nonsense.

Presumably, this is where Sarah's phenology/planting HLS work would fit in? If I have some examples of those files, I can swap in logic that uses those and run some real tests.

ashiklom: Or I can leave this as is to get this merged and then make this change in a future PR (or in developing the full, scaled workflow in the CCMMF organization or elsewhere, if that's what we want to do).

mdietze: Let's get this PR in and then do a separate PR that used the HLS planting and harvesting dates.  @sarahkanee could you point Alexey to where the HLS date outputs live?

modules/data.land/R/water_balance.R
ashiklom: Agreed on the approach, but I'm going to suggest punting that to a future PR.

I think the correct meta-design is probably a dedicated function for rice management that this calls out to -- e.g.,

```r
if (is_rice) {
  return(rice_water_balance(...))
}
# Continue with non-rice logic
```

...or maybe even a mapping between crop types and their corresponding water balance functions that gets used by `apply_water_balance`. That would allow downstream expansion to other special kinds of water management.

infotroph (commented):
Apparently I started a review ages ago and never clicked submit 🤦. See below

infotroph:
@ashiklom OK to merge this when checks pass?

ashiklom:
Yes!

ashiklom:
Actually, on second thought, maybe hold on this. I have a lot more changes related to the actual statewide irrigation workflow, so it's probably better to just do it all in one big PR. I'll add those to this PR this week.

modules/data.land/R/water_balance.R
ashiklom: Flooded rice now has its own implementation in here (see below).

ashiklom:
OK, _now_ this is ready for merge (or at least review). This now accurately reflects the actual statewide irrigation workflow we are using for the current batch of event files for the CCMMF MVP.

A few major changes since last time:

1. Similar to #3919, this puts the code in the `workflows` folder rather than abusing `data.land/inst` as it did before.
2. The entire workflow is implemented using `targets`. I know we originally shied away from `targets`, but I've found this works _really_ well here, especially for incrementally updating the pipeline and for scaling dynamically across the BU HPC.
    - Some of these functions should eventually be migrated into `data.land`, but that feels like a premature generalization since this is only one very specific workflow. Suggest punting that to post-MVP.
3. The CIMIS and CHIRPS functions I had originally developed for `data.land` scaled terribly for the statewide analysis. I replaced these with offline workflows (in `workflows/irrigation-statewide/preprocessing`) that do the time- and compute-intensive spatial extractions as separate steps. These run much faster. At some point, we can think about whether/how to integrate these more tightly into `data.land`, `data.remote`, or somewhere else.
4. There is an explicit implementation of flooded rice.
5. Upland canopy irrigation can be capped per-even via a `irrigation_max` parameter (reasonable default is 150 mm ~= 0.6 in).

Note that the CI `check` will currently fail because `data.land` now imports >= 25 packages, which causes `devtools::check` to raise a `NOTE` like:

```
0 warnings found in modules/data.land.
3 notes found in modules/data.land.
R check of modules/data.land returned new problems:
checking package dependencies ... NOTE: Imports includes 25 non-default packages.
Execution halted
```

The 3 packages I added are (1) `arrow`, which we actually want for reading parquet files; (2) `httr2`, which has been the modern alternative to `httr` for a while now (and which is used here for SSURGO queries); and (3) `tibble`, which we already implicitly depend on through `dplyr`. It's possible to artificially silence this NOTE by replacing a bunch of calls like `tibble::tibble`, `tibble::as_tibble`, `tidyselect::one_of`, etc. with `dplyr::` (which re-rexports all of these functions)...but that seems disingenuous. The better solution is to add this note to our NOTE checker, but that may require intervention by @infotroph or other CI wizards.

modules/data.land/inst/CIMIS-event-files.R
ashiklom: Moot! The new workflow now uses real MSLSP inputs.

modules/data.land/NAMESPACE
ashiklom: A lot of these have been removed in the final iteration.

modules/data.land/tests/Rcheck_reference.log
```diff
@@ -13,7 +13,7 @@
 * package encoding: UTF-8
 * checking package namespace information ... OK
 * checking package dependencies ... NOTE
-Imports includes 22 non-default packages.
+Imports includes 25 non-default packages.
```
ashiklom, line 16: @infotroph On reflection, I went with this approach because it's less work and less intrusive (and also, I was only able to get down to 23 packages with `dplyr` re-exports, not 22, so the problem would still persist).

I'd prefer to defer a cleanup of `data.land` dependencies to a separate dedicated PR.

modules/data.land/DESCRIPTION
```diff
@@ -25,13 +25,15 @@ URL: https://pecanproject.github.io
...
 Imports:
+    arrow,
     coda,
     curl,
     dplyr,
     foreach,
     fs,
     future,
     furrr,
+    httr2,
```
infotroph: Current CI failure on 4.2 is because it's pulling in httr2 0.2.2 that lacks `req_perform_parallel()`. Feel free to adjust if you like a different minimum version -- my read of the NEWS is it was added in 1.0.0 but took a few versions to become stable.
```suggestion
    httr2 (>= 1.1.0),
```

infotroph: (After this patch you'll need to run `./scripts/generate_dependencies.R` to propagate it through the dependency lists)

ashiklom: Addressed in 68170ee6ce4a7b525a25d8808f6f6a1851e6410c

PR #3823, opened 2026-02-20
fix(soil2netcdf): delete partial file if ncvar_put fails

### LINKED ISSUE #3822: soil2netcdf() leaves partially written corrupt netCDF file when write fails
### Description

If `soil2netcdf()` encounters an error during `ncdf4::ncvar_put()`, the function stops with an error but leaves behind a partially written netCDF file on disk.

### To Reproduce

```R
library(PEcAn.data.land)
library(ncdf4)

soil.data <- list(
  soil_depth = c(0.1, 0.3, 0.6),              # length 3
  fraction_of_sand_in_soil = c(0.4, 0.5),     # length 2 (mismatch)
  fraction_of_clay_in_soil = c(0.2, 0.2, 0.2) # length 3
)

soil2netcdf(soil.data, "test_partial_write.nc")
```

### Error Produced
```
Error in ncdf4::ncvar_put(...) :
  ncvar_put: error: you asked to write 3 values, but the passed data array only has 2 entries!
```
#### File Still Exists After Failure
```
file.exists("test_partial_write.nc")
```
Output:
```
[1] TRUE
```
<b>Inspecting the File</b>
```
nc <- nc_open("test_partial_write.nc")
nc$dim$depth$len
```
Output:
[1] 3

```
ncvar_get(nc, "fraction_of_sand_in_soil")
```
Output:
```
[1] 9.96921e+36 9.96921e+36 9.96921e+36
9.96921e+36 is the default netCDF _FillValue, indicating that data were never successfully written.
```

## Screenshots

Image

shreyannandanwar:
## Description

fixes #3822 

Replaced `on.exit` with an explicit `tryCatch` block around the write loop:
- On success: file is closed normally
- On failure: error is captured → file is closed → file is deleted → error is re-thrown

either a complete valid file exists, or no file exists.

## Output
Screenshot 2026-02-20 at 8 06 29 AM

<b>All tests passed
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.land/R/soil2netcdf.R
+++ b/modules/data.land/R/soil2netcdf.R
@@ -24,9 +24,9 @@
 #' @examples
 #' \dontrun{ 
 #' soil.data <- list(
-#'   fraction_of_sand_in_soil = c(0.3,0.4,0.5),
-#'   fraction_of_clay_in_soil = c(0.3,0.3,0.3),
-#'   soil_depth = c(0.2,0.5,1.0)
+#'  fraction_of_sand_in_soil = c(0.3,0.4,0.5),
+#'  fraction_of_clay_in_soil = c(0.3,0.3,0.3),
+#'  soil_depth = c(0.2,0.5,1.0)
 #' )
 #'
 #' soil2netcdf(soil.data,"soil.nc")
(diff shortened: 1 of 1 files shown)


modules/data.land/R/soil2netcdf.R
```diff
@@ -86,10 +86,22 @@ soil2netcdf <- function(soil.data, new.file) {
...
+    write_error <<- e
+  })
+
+  ncdf4::nc_close(nc)
+
+  if (!is.null(write_error)) {
+    if (file.exists(new.file)) {
+      unlink(new.file)
+      }
+    stop(write_error)
```
S1DDHEY, line 105: I think using `logger.severe("Failed to write NetCDF file:", write_error$message)` could be a better way. This helps in debugging.

infotroph:
Thank you for reporting and proposing a patch! I'm closing this in favor of #3825. 

Our general approach to ensuring file quality is to validate the data structure before attempting to write it, rather than go ahead with the write and then delete the result on failure -- this is faster and when someone tries to overwrite an existing valid file, it reduces the chance of deleting their old file without replacing it. I recognize there are many ways for a file write to fail and we can't guard against all of them, but our experience is that checking for format issues has the highest payoff -- At least in PEcAn, they are the most common cause of write failures and the most directly under user control.

Issue #3738, opened 2025-12-30
ggplot2 ≥ 3.4.0 deprecation warnings due to `size` aesthetic in plotting code


S1DDHEY:
## Description
With **ggplot2 version 3.4.0 and later**, the `size` aesthetic for line geoms has been deprecated in favor of `linewidth`.

When running PEcAn workflows (e.g. the uncertainty analysis demo), I came across this deprecation warnings. After searching the codebase, there appear to be multiple ggplot2 calls across different files that use `size` with line geoms.

## Screenshot
<img width="1220" height="176" alt="Image" src="https://github.com/user-attachments/assets/c669ed27-ebf1-4aa4-911e-80725ac5f647" />

S1DDHEY:
I was wondering whether the changes should be done in the entire codebase or not as I found using `grep` that lots of file use this currently.

Roughly these
<img width="1706" height="929" alt="Image" src="https://github.com/user-attachments/assets/52ba0aee-e559-435a-a85a-5b5b5450e983" />

infotroph:
@S1DDHEY Thanks for reporting! This should be fixed "almost everywhere", with the exception being that anything inside an `inst/` folder does not need to be updated: `inst/` folder contents are mostly scripts that are not fully integrated into the production code and have been stored there for reference or future refactoring, so we usually do not update them during normal maintenance.

PR #3826, opened 2026-02-21
Remove unused `create_execute_test_xml()` and `model_specific_tags()`

### LINKED ISSUE #3410: Eliminate model/ID-specific hacks in create_execute_test_xml (well, and everywhere)
@mdietze in https://github.com/PecanProject/pecan/pull/3407#discussion_r1905592352 :

> Not sure what the aim or this overall function is, or where it's called from, but this sort of deeply-buried hard-coded hack is something we should really avoid. I know you didn't add this and are just reformatting it, but I wonder if we could take the opportunity to fix it? First, this bit won't work if we're moving away from BETY being required (and never would have worked on a machine lacking this input ID). Second, as we move towards more people passing in ensemble soil inputs, this is going to break things (especially if a soil input is already present, which the code never checks for)

The code in question, from PEcAn.workflow::create_execute_test_xml:
```
model_specific_tags <- function(settings, model.info) {
  # some extra settings for LPJ-GUESS
  if (model.info$model_name == "LPJ-GUESS") {
    settings$run$inputs <- c(
      settings$run$inputs,
      list(soil = list(id = 1000000903))
```

S1DDHEY:
## Description

fixes #3410 
As mentioned by @infotroph in https://github.com/PecanProject/pecan/issues/3410#issuecomment-3936040382 deleting the entire file seems reasonable.

The only place that I see this being used is in the `base/workflow/inst/batch_run.R` which as per this https://github.com/PecanProject/pecan/issues/3738#issuecomment-3708459433 isn't usually pushed to production.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


base/workflow/NAMESPACE
```diff
@@ -1,12 +1,9 @@
...
-export(create_execute_test_xml)
 export(do_conversions)
-export(model_specific_tags)
 export(run.write.configs)
 export(runModule.get.trait.data)
 export(runModule.run.write.configs)
 export(runModule_start_model_runs)
 export(start_model_runs)
 importFrom(dplyr,"%>%")
-importFrom(dplyr,.data)
```
infotroph, line 12: I would not have guessed that this file was the only place in the whole workflow package we used `.data`! But grep agrees, so removing this is correct.

infotroph (changes requested):
Thanks! This looks good. Can you please also add an entry to NEWS.md describing this change? Make sure to credit yourself and reference this PR number.

base/workflow/NAMESPACE
S1DDHEY: These changes were not made by hand, all this is done by `make document`

infotroph: Yep, you did the right thing here! I was thinking out loud while checking -- and finding it's not an issue here! -- for the potential gotcha that having `#' @importFrom <pkg> <fn>` in any one function's Roxygen block makes that imported function available to the entire package, so then it would be easy for other functions to be using `.data` and make checks fail when we delete this file. If that had happened, the fix would be to add the importFrom in another file we aren't deleting.

(Many folks now collect all their `@importFrom`s into [one Roxygen block for the whole package](https://roxygen2.r-lib.org/articles/namespace.html#functions-1) rather than next to individual functions. Long term it might be worth moving PEcAn towards that approach, but that would a very low priority change.)

infotroph (approved):

PR #3829, opened 2026-02-23
Fix/3797: Enable examples in sendmail

### LINKED ISSUE #3797: Rethink \dontrun in examples

man080107:
I reviewed the \dontrun examples in `base/utils/R/mail.R` for the `sendmail` function. I verified that the code runs correctly and updated the tag from `\dontrun` to `\donttest` with an `if (interactive())` guard to improve test coverage per issue #3797.Fixes #3797

### DIFF
--- a/base/utils/R/mail.R
+++ b/base/utils/R/mail.R
@@ -9,10 +9,14 @@
 #' @return nothing
 #' @export
 #' @examples
-#' \donttest{
-#' if (interactive()) {
-#'   sendmail('bob@@example.com', 'joe@@example.com', 'Hi', 'This is R.')
-#' }
+#' \dontrun{
+#'   # Check if 'sendmail' is installed before running.
+#'   # This prevents errors in environments like Docker where the utility is missing.
+#'   if (nzchar(Sys.which("sendmail"))) {
+#'     sendmail("bob@@example.com", "joe@@example.com", "Hi", "This is R.")
+#'   } else {
+#'     message("System 'sendmail' utility not found. Skipping example.")
+#'   }
 #' }
 sendmail <- function(from, to, subject, body) {
   if (is.null(to)) {
--- a/base/utils/R/mail.R
+++ b/base/utils/R/mail.R
@@ -9,8 +9,10 @@
 #' @return nothing
 #' @export
 #' @examples
-#' \dontrun{
-#' sendmail('bob@@example.com', 'joe@@example.com', 'Hi', 'This is R.')
+#' \donttest{
+#' if (interactive()) {
+#'   sendmail('bob@@example.com', 'joe@@example.com', 'Hi', 'This is R.')
+#' }
 #' }
 sendmail <- function(from, to, subject, body) {
   if (is.null(to)) {
--- a/base/utils/man/sendmail.Rd
+++ b/base/utils/man/sendmail.Rd
@@ -22,10 +22,14 @@ nothing
 Sends email. This assumes the program sendmail is installed.
 }
 \examples{
-\donttest{
-if (interactive()) {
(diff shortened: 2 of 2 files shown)


infotroph:
Thanks for looking at this! The challenge here is that not all systems will have the `sendmail` command installed, and even if installed it might not be configured to accept mail from nonroot users. That means a given user might or might not be able to run the example, independent of whether their session is interactive.

```
chrisb@ulmus pecan % docker run --rm -it pecan/base:develop \                                                                                             
>   Rscript -e 'PEcAn.utils::sendmail("bob@@example.com", "joe@@example.com", "Hi", "This is R.")'
sh: 1: : Permission denied
Warning message:
In system2(sendmail, c("-f", paste0("\"", from, "\""), paste0("\"",  :
  error in running command
```

man080107:
Thanks for catching that! I didn't realize the Docker container would throw a permission error there.

I've updated the example to strictly check for the existence of the sendmail utility using **if (nzchar(Sys.which("sendmail")))** before attempting to run. This should prevent the **sh: 1:: Permission denied**  in the CI environment while still providing a working example for users who have it configured

infotroph:
Honestly I think the existing \dontrun is correct for this example. Email is complicated these days and even where a sendmail binary is found, that's not a reliable indicator that the system has a mail transfer agent running at all, correctly configured, authorized to send messages on behalf of the specified `From:` address, etc.

Given how little this function is used in the PEcAn code and how few modern systems will be configured to be capable of using it, it may be time (in a different PR) to deprecate the PEcAn `sendmail()` function entirely.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
I'm going to close this; feel free to propose other fixes in issues where we can discuss before starting implementation.

PR #3835, opened 2026-02-25
add N application rate and compost amendment datasets to data.land

divine7022:
## Description

Add crop-specific nitrogen application rates and organic amendment (compost) properties as packaged datasets in data.land, with lookup functions for downstream use

**New datasets:**
- `n_application_rate_data` -- recommended N rates (min/max) for 33 CA crops in lbs/acre and kg/m2, from CDFA-FREP and UC ANR sources
- `compost_amendment_data` -- properties (C:N, %C, %N, PAN, application rates) for 32 organic amendment materials in lbs/acre and kg/m2

**New functions:**
- `look_up_n_application_rate(crop, pft_group, unit)` -- partial match lookup for N rates
- `look_up_compost_amendment(material, n_class)` -- partial match lookup for amendment properties

**New files:**
- `data-raw/n_application_rates.csv`
- `data-raw/compost_amendments.csv`
- `data-raw/create_n_rate_data.R`
- `data-raw/create_compost_data.R`
- `data/n_application_rate_data.rda`
- `data/compost_amendment_data.rda`
- `R/look_up_n_application_rate.R`
- `R/data.R` (updated with dataset documentation)

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


CHANGELOG.md
```diff
@@ -15,7 +15,9 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - Added datasets to `PEcAn.data.land` 
   *  `landiq_crop_mapping_codes` dataset mapping LandIQ crop classification codes to human-readable crop names.
   *  `bism_kc_by_crop` dataset containing BISm crop coefficient schedules and stage timing references for use in ET estimation, including columns that map to LandIQ class and subclass.
-
+  *  `n_application_rate_data` dataset with recommended N application rates for 33 California crops.
```
dlebauer: Add ca to dataset name and make sure to document that these were specifically developed for CA ag

modules/data.land/R/look_up_n_application_rate.R
```diff
@@ -0,0 +1,127 @@
...
+}
+
+
+#' Look up compost amendment properties
+#'
+#' Returns properties of organic amendment materials including carbon and
+#' nitrogen content, C:N ratio, and plant-available nitrogen (PAN).
+#'
+#' @param material Character string. Amendment material to look up.
+#'   Case-insensitive partial matching (e.g. "cow" matches "Cow manure").
```
dlebauer: Partial matching can cause errors, eg if there are multiple matches. 
Better to have a function to provide a list, and if there isn't an exact match, use partial matching to provide user with possible matches.

modules/data.land/data-raw/compost_amendments.csv
```diff
@@ -0,0 +1,33 @@
+material,cn_min,cn_max,cn_avg,c_pct,n_pct,pan_pct,n_class,app_rate_min,app_rate_max,total_c_min_lbs_acre,total_c_max_lbs_acre,total_n_min_lbs_acre,total_n_max_lbs_acre,total_c_min_kg_m2,total_c_max_kg_m2,total_n_min_kg_m2,total_n_max_kg_m2
```
dlebauer: Round all values to meaningful levels of precision. If a range is given with one significant figure (eg 40-100 or 100-200 lbs / acre) then the transformed value shouldn't have more than two, but one may suffice

dlebauer: Also, I suggest kg/ha instead of kg/m2 because it is more commonly used in agronomic contexts, and has values in a range that is easier to interpret.

modules/data.land/R/look_up_n_application_rate.R
```diff
@@ -0,0 +1,127 @@
...
+#'   `source`. The `min_n` and `max_n` columns are in the requested unit.
+#'
+#' @examples
+#' look_up_n_application_rate("corn")
+#' look_up_n_application_rate("tomato")
+#' look_up_n_application_rate("wheat", unit = "lbs_acre")
+#' look_up_n_application_rate("pistachio", pft_group = "woody")
+#'
+#' @export
+look_up_n_application_rate <- function(
```
dlebauer: how about 'get' to be consistent with the get_fertilizer..  function?

dlebauer (commented):
Nice work. I left a few suggestions/questions.

Big picture: now we have two ways to store and look up compost amendments. Did you consider either
1. Storing the information in similar data structures, have the datasets and functions include the source in their name, and functions with similar arguments, or 
2. Combining sources into a single dataset with a source column, and have a single function 
?

We want to avoid having unique approaches to handling the same type of information.

modules/data.land/R/look_up_n_application_rate.R
divine7022: renamed to get_ca_n_rate() and get_ca_compost_amendment(). Note: the existing fertilizer function is actually look_up_fertilizer_components() (not get_*), but get_ is shorter and cleaner

modules/data.land/data-raw/compost_amendments.csv
divine7022: good catch!

modules/data.land/R/look_up_n_application_rate.R
divine7022: updated to exact match first behavior: (1) try case insensitive exact match, (2) if no exact match found, use partial matching to suggest possible matches with a warning message and return empty, (3) user picks from the suggestions

divine7022:
>two ways to store and look up compost amendments

two datasets serve complementary purposes:
- `fertilizer_composition_data` (SWAT/DayCent) answers "what fraction of this fertilizer is N/C?" it gives nutrient fractions that you multiply by an application amount.
- compost_amendment_data (CDFA HSP) answers "how much total C and N per hectare does this amendment contribute at CDFA recommended application rates?" it provides the area scaled amounts directly, which is what the Sobol priors need.

They overlap on materials (cow manure, poultry, etc.) but the columns and use cases differ. The existing `look_up_fertilizer_components(type, amount)` computes `amount x fraction_c`, while the new function returns pre computed totals from CDFA application rate guidelines

Two options for reconciliation:
1. add a `source` column to both datasets and document the distinction clearly. Keep separate functions since the return schemas differ.
2. use existing`look_up_fertilizer_components` as the single entry point and add an `application_rate` source from CDFA to `fertilizer_composition_data`.

I lean toward (1) these answer different questions and their schemas reflect that. But happy to iterate. Will add a clear cross reference `@seealso` between the two

divine7022:
@dlebauer addressed:
- rename datasets `n_application_rate_data` -> `ca_n_application_rate`, `compost_amendment_data` -> `ca_compost_amendment` (ca specific prefix)
- rename functions `look_up_n_application_rate` -> `get_ca_n_rate`, `look_up_compost_amendment` -> `get_ca_compost_amendment`
- change units from kg/m2 to kg/ha
- round converted values to meaningful precision (1 decimal in kg/ha)
- use exact match first behavior; partial match now suggests options instead of returning multiple rows
- add `@seealso` cross references to existing `look_up_fertilizer_components`
- document datasets as ca specific (CDFA-FREP, UC ANR sources)

modules/data.land/R/look_up_n_application_rate.R
dlebauer: Sorry, I misremembered. I think we switched get for look_up in review. So let's stick with look_up

divine7022: reverted to look_up_ca_n_rate() and look_up_ca_compost_amendment()

divine7022:
@dlebauer standardize unit to g/m2

modules/data.land/R/look_up_ca_n_rate.R
```diff
@@ -0,0 +1,171 @@
...
+#'
+#' @seealso [look_up_fertilizer_components()] for fertilizer nutrient
+#'   composition (N/C fractions) from the SWAT/DayCent database.
+#'   [ca_n_application_rate] for the underlying dataset.
+#'
+#' @examples
+#' look_up_ca_n_rate("Tomatoes, Processing")
+#' look_up_ca_n_rate("corn")
+#' look_up_ca_n_rate("wheat", unit = "lbs_acre")
+#' look_up_ca_n_rate("pistachio", pft_group = "woody")
```
dlebauer: 'pistachio' returns nothing
```suggestion
#' look_up_ca_n_rate("pistachios", pft_group = "woody")
```

modules/data.land/R/look_up_ca_n_rate.R
```diff
@@ -0,0 +1,171 @@
...
+#'   `total_n_min_g_m2`, `total_n_max_g_m2`.
+#'   Returns an empty data frame (with a warning) if no match is found.
+#'
+#' @seealso [look_up_fertilizer_components()] for fertilizer nutrient
+#'   composition (N/C fractions) from the SWAT/DayCent database.
+#'   [ca_compost_amendment] for the underlying dataset.
+#'
+#' @examples
+#' look_up_ca_compost_amendment("Cow manure")
+#' look_up_ca_compost_amendment("Poultry litter", n_class = "HIGHER")
```
dlebauer: 'HIGHER' returns nothing
```suggestion
#' look_up_ca_compost_amendment("Poultry litter", n_class = "LOWER")
```

modules/data.land/data-raw/compost_amendments.csv
```diff
@@ -0,0 +1,33 @@
...
+"Newspaper, shredded",400,800,600,40,0.07,-29,LOWER,8000,10600,3200,4240,5.33,7.07,358.672,475.24,0.597,0.792
+Paper,150,200,175,40,0.23,-26.57,LOWER,8000,10600,3200,4240,18.29,24.23,358.672,475.24,2.05,2.716
+Pine needles,80,80,80,40,0.5,-22.5,LOWER,8000,10600,3200,4240,40,53,358.672,475.24,4.483,5.941
+Poultry litter,5,20,12.5,40,3.2,18,LOWER,8000,10600,3200,4240,256,339.2,358.672,475.24,28.694,38.019
+Sawdust and wood chips,100,500,300,40,0.13,-28,LOWER,8000,10600,3200,4240,10.67,14.13,358.672,475.24,1.196,1.584
+Straw,40,100,70,40,0.57,-21.43,LOWER,8000,10600,3200,4240,45.71,60.57,358.672,475.24,5.123,6.789
+Straw — oats and wheat,70,80,75,40,0.53,-22,LOWER,8000,10600,3200,4240,42.67,56.53,358.672,475.24,4.783,6.336
+Swine manure,8,20,14,40,2.86,12.86,LOWER,8000,10600,3200,4240,228.57,302.86,358.672,475.24,25.619,33.946
+Vegetable waste,10,20,15,40,2.67,10,LOWER,8000,10600,3200,4240,213.33,282.67,358.672,475.24,23.911,31.683
+Vegetable waste,12,20,16,40,2.5,7.5,LOWER,8000,10600,3200,4240,200,265,358.672,475.24,22.417,29.703
```
dlebauer: Docs say " A tibble with one row per amendment" 
Here there are two rows for 'vegetable waste'
and `look_up_ca_compost_amendment("vegetable waste")` returns two rows. 

Rows should provide some indication of how the two rows differ. Are they different types of waste, do the values come from different sources, or other?

dlebauer (changes requested):
1. Add citations to dataset and function documentation. 
    - if one dataset came from a single reference, each row does not need a citation. 
    - citation for each source should be in function and dataset docs. 
    - use "Author, year" style, e.g. use "Rosenstock et al., 2013" instead of "https://escholarship.org/uc/item/5mk2q1sm",
    - if "CDFA-FREP & UC Davis" is different from Rosenstock et al 2013, then there it needs to be clear where the numbers came from - are they in both? are they a combination of the two? 
    - The Rosenstock citation is: "Rosenstock, T. S, Liptzin, D., Six, J., & Tomich, T. P. (2013). Nitrogen fertilizer use in California: Assessing the data, trends and a way forward. California Agriculture, 67(1). Retrieved from https://escholarship.org/uc/item/5mk2q1sm"
2. Disambiguate pairs of rows with the same "source" (see comment)
3. Add tests (or create a new issue to create these tests and tag 'good first issue')

modules/data.land/data-raw/compost_amendments.csv
divine7022: cuz same materials have different C:N ratios reported in different literature sources (NC State Extension and UNL Extension). I think we should add source_short column to disambiguate, e.g. "Vegetable waste (NC State)" vs "Vegetable waste (UNL)".
 would that work, or would you prefer aggregating into a single row per material ?

dlebauer: Add a source column. If a table has data from multiple sources, this should always be the case, and then the dataset documentation can have the full citations.

dlebauer: To return one row per amendment, consider adding an additional argument like `aggregate = c('none', 'mean')` if the goal is to have the function return a single value, could set mean by default.

divine7022:
thanks for suggestions @dlebauer, quick note :

- replaced urls with author year citations (e.g. "Rosenstock et al., 2013") in both `n_application_rates.csv` and `compost_amendments.csv`; full references with urls are now in `@source` blocks for both datasets and both lookup functions

- changed `look_up_ca_n_rate("pistachio")` to `look_up_ca_n_rate("pistachios")` in example, it now matches the actual data

- changed to `n_class = "LOWER"`, "HIGHER" doesn't exist in data, only "LOWER" does

- changed from "one row per amendment material" to correctly describe all rows with duplicate materials disambiguated by source. Added `source` column to `ca_compost_amendment` (values: "Eghball, UNL Extension" or "Rynk, NC State Extension").

- added `aggregate = c("none", "mean")` parameter to `look_up_ca_compost_amendment()`; default to `"none"` (returns all matching rows). `"mean"` collapses duplicates by averaging cols per material

- added tests in `test.look_up_ca_n_rate.R`

dlebauer (approved):
Looks great, thanks!

PR #3836, opened 2026-02-25
Update events schema: require crop id when planting, remove pft from site

infotroph:
As discussed in #3828: Adds events schema v0.1.1, requiring planting events to specify a crop type and dropping the requirement for a site-level PFT declaration. Also adds a first pass at schema documentation to the book; edits welcome.

Here's the diff between schemas:

```
--- /Users/chrisb/clones/pecan/modules/data.land/inst/events_schema_v0.1.0.json
+++ /Users/chrisb/clones/pecan/modules/data.land/inst/events_schema_v0.1.1.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://pecanproject.org/schema/events-mvp-0-1-0.json",
+  "$id": "https://pecanproject.org/schema/events-mvp-0-1-1.json",
   "oneOf": [
     { "$ref": "#/$defs/site" },
     { "type": "array", "items": { "$ref": "#/$defs/site" } }
@@ -10,9 +10,8 @@
       "type": "object",
       "required": ["pecan_events_version", "site_id", "events"],
       "properties": {
-        "pecan_events_version": { "type": "string", "const": "0.1.0" },
+        "pecan_events_version": { "type": "string", "const": "0.1.1" },
         "site_id": { "type": "string", "minLength": 1 },
-        "pft": { "type": "string" },
         "ensemble_id": { "type": ["string", "null"], "minLength": 1 },
         "geometry_uri": { "type": ["string", "null"], "format": "uri" },
         "provenance": { "type": "object", "additionalProperties": true },
@@ -58,7 +57,7 @@
             },
             "allOf": [
               { "if": { "properties": { "event_type": { "const": "planting" } } },
-                "then": { "required": ["leaf_c_kg_m2"] } },
+                "then": { "required": ["crop_code", "leaf_c_kg_m2"] } },
               { "if": { "properties": { "event_type": { "const": "harvest" } } },
                 "then": { "required": ["frac_above_removed_0to1"] } },
               { "if": { "properties": { "event_type": { "const": "irrigation" } } },
```


book_source/03_topical_pages/02_pecan_standards.Rmd
```diff
@@ -94,6 +94,86 @@ See the [Soil Data] section on more into on creating a standard soil data file.
...
+
+Required properties:
+
+* `crop_code`: a machine-readable identifier that specifies what type of plant was added. Should preferably be short and use only characters that are safe in filenames.
+* `leaf_c_kg_m2`: Amount of leaf biomass added. 
+
+Optional properties:
+
+* `wood_c_kg_m2`: Biomass added as aboveground stems. Assumed zero if not specified.
+* `fine_root_c_kg_m2`: Biomass added to belowground roots. Assumed zero if not specified.
```
mdietze, line 121: is calling this fine_root but not having a coarse_root potentially confusing/misleading?

book_source/03_topical_pages/02_pecan_standards.Rmd
```diff
@@ -94,6 +94,86 @@ See the [Soil Data] section on more into on creating a standard soil data file.
...
+Addition of live biomass that is expected to grow according to the model's plant growth rules. The current spec expects a "planting" to be a transplantation of leafed-out seedlings; models which represent planting as addition of unsprouted seeds might choose to adjust the date to account for germination time.
+
+Required properties:
+
+* `crop_code`: a machine-readable identifier that specifies what type of plant was added. Should preferably be short and use only characters that are safe in filenames.
+* `leaf_c_kg_m2`: Amount of leaf biomass added. 
+
+Optional properties:
+
+* `wood_c_kg_m2`: Biomass added as aboveground stems. Assumed zero if not specified.
```
mdietze, line 120: maybe call this stem not wood if it can (and does) include allocation to non-woody stems

book_source/03_topical_pages/02_pecan_standards.Rmd
```diff
@@ -94,6 +94,86 @@ See the [Soil Data] section on more into on creating a standard soil data file.
...
+#### planting
+
+Addition of live biomass that is expected to grow according to the model's plant growth rules. The current spec expects a "planting" to be a transplantation of leafed-out seedlings; models which represent planting as addition of unsprouted seeds might choose to adjust the date to account for germination time.
+
+Required properties:
+
+* `crop_code`: a machine-readable identifier that specifies what type of plant was added. Should preferably be short and use only characters that are safe in filenames.
+* `leaf_c_kg_m2`: Amount of leaf biomass added. 
+
+Optional properties:
```
mdietze, line 118: This list is missing the leaf, stem, and root N pools

book_source/03_topical_pages/02_pecan_standards.Rmd
```diff
@@ -94,6 +94,86 @@ See the [Soil Data] section on more into on creating a standard soil data file.
...
+
+Removal of biomass from the living pool, either by transferring it to dead (litter) biomass or by removing it from the system. Specified as a fraction of live biomass.
+
+Required properties:
+* `frac_above_removed_0to1`: Fraction of existing aboveground biomass removed from the system.
+
+Optional properties:
+
+* `frac_below_removed_0to1`: Fraction of existing beloweground biomass removed from the system. Assumed zero if not specified.
+* `frac_above_to_litter_0to1`, `frac_below_to_litter_0to1`: fraction of existing above- or below-ground live biomass converted to dead litter. Assumed zero if not specified.
```
mdietze, line 135: not sure why litter would be considered optional, or why it would be assumed to be 0 not 1-removed.

book_source/03_topical_pages/02_pecan_standards.Rmd
```diff
@@ -94,6 +94,86 @@ See the [Soil Data] section on more into on creating a standard soil data file.
...
+
+* `crop_code`: a machine-readable identifier that specifies what type of plant was added. Should preferably be short and use only characters that are safe in filenames.
+* `leaf_c_kg_m2`: Amount of leaf biomass added. 
+
+Optional properties:
+
+* `wood_c_kg_m2`: Biomass added as aboveground stems. Assumed zero if not specified.
+* `fine_root_c_kg_m2`: Biomass added to belowground roots. Assumed zero if not specified.
+* `cultivar`: Identifier for finer-grainer plant type identifiers that will be meaningful to models with detailed growth parameterizations but likely ignored by PFT-based models.
+* `crop_display`: A human-readable name for the added plant type.
```
mdietze, line 123: The mapping between machine and human readable crop type doesn't seem like the sort of thing that needs to be repeated for every row of data. Seems like basic metadata that should be in a different part of the JSON file

book_source/03_topical_pages/02_pecan_standards.Rmd
dlebauer: SIPNET assumes that seedlings emerge without coarse root.

book_source/03_topical_pages/02_pecan_standards.Rmd
infotroph:
```suggestion
* `wood_c_kg_m2`: Biomass added as aboveground woody matter. Assumed zero if not specified.
```

book_source/03_topical_pages/02_pecan_standards.Rmd
mdietze: Could you point to the SIPNET line numbers that show that?

book_source/03_topical_pages/02_pecan_standards.Rmd
mdietze: Ok, now the name "wood" is more accurate, but you've eliminated the ability to have nonwoody stems. I really think calling this stem not wood is the preferred solution

infotroph: That's true, but are there any PEcAn models that make a distinction between nonwoody stems and leaves? Wood vs leaf seems to me aligned with how both Sipnet's and the PEcAn output standard do it.

(Happy to stipulate that "does wood include roots" is still messy to answer, but that's the same mess we already have at other stages)

book_source/03_topical_pages/02_pecan_standards.Rmd
infotroph: It's optional so that if you remove 50% of the live biomass without specifying anything else, 50% stays alive and none is left dead. Note that I don't think Sipnet's harvests *have* a way to remove litter that was dead before the event, but forget if that was an oversight or an intentional decision.

[I think a previous comment got eaten here; sorry if you see this as a duplicate]

book_source/03_topical_pages/02_pecan_standards.Rmd
mdietze: Yes, my reading of the SIPNET code is that it (and most models) make a very clear distinction between nonwoody stems (which are aboveground but don't contribute to photosynthetic LAI) and leaves (which photosynthesize). I'd also argue that far fewer models make a functional distinction between woody and non-woody stems, but that the difference is mostly a parametric one (woody stems have lower turnover rates, higher C:N, etc). Some models do make a distinction between sapwood and heartwood, but again the distinction for a nonwoody stem is mostly parametric (0 allocation to heartwood)

book_source/03_topical_pages/02_pecan_standards.Rmd
dlebauer: I was wrong, sorry no lines.

dlebauer (commented):
Looks good to me, pending resolution of @mdietze's comments

book_source/03_topical_pages/02_pecan_standards.Rmd
infotroph: I'll move this up to its own section of the schema and ping when I push it

ashiklom:
FWIW, the planting files recently generated by @sarahkanee look like this (output of `dplyr::glimpse`):

```
Rows: 7
Columns: 16
$ event_type   <chr> "planting", "planting", "planting", "planting", "planting~
$ site_id      <chr> "39011", "39011", "39011", "39011", "39011", "39011", "39~
$ year         <int> 2018, 2019, 2020, 2021, 2022, 2022, 2023
$ date         <date> 2018-06-13, 2019-06-08, 2020-05-30, 2021-06-02, 2022-01-2~
$ code         <chr> "F1", "R1", "R1", "R1", "G2", "F16", "R1"
$ PFT          <chr> "row", "rice", "rice", "rice", "row", "row", "rice"
$ LAI          <dbl> 1.0436273, 1.3505321, 1.1833609, 1.2425158, 0.2681212, 0.~
$ C_LEAF       <dbl> 0.017673607, 0.032418944, 0.028406071, 0.029826060, 0.00~
$ C_STEM       <dbl> 2.930107e-04, 5.374735e-04, 4.709441e-04, 4.944861e-04, 7~
$ C_FINEROOT   <dbl> 0.005289909, 0.009703353, 0.008502256, 0.008927274, 0.001~
$ C_COARSEROOT <dbl> 0.019404199, 0.035593393, 0.031187581, 0.032746615, 0.004~
$ N_LEAF       <dbl> 9.781366e-07, 2.108748e-06, 1.847724e-06, 1.940090e-06, 3~
$ N_STEM       <dbl> 5.764009e-06, 1.120185e-05, 9.815263e-06, 1.030592e-05, 1~
$ N_FINEROOT   <dbl> 3.414607e-04, 3.210046e-04, 2.812701e-04, 2.953304e-04, 8~
$ N_COARSEROOT <dbl> 0.0003709448, 0.0008286138, 0.0007260466, 0.0007623409, 0~
```

The units here are kg X per m2, consistent with your table below. A few things to note:

- `code` is the LandIQ code (LandIQ class + LandIQ subclass), which presumably is equivalent to `crop_code` (or that's what I'm going to assume, anyway).
- `PFT` is mapping is included. @infotroph @dlebauer and I had a lengthy discussion about the relationship between `crop_code` and `PFT` in the context of plant traits. From what I recall, we agreed that, for now, the mapping of `crop_code` to `PFT` (since PFT is what is used to retrieve trait values from meta-analysis posteriors) would live somewhere else. However, having a way to (optionally) override that default mapping by passing an (optional) PFT key directly through the events file may be useful in some circumstances (e.g., if you want to force a specific nonstandard parameter lookup).
- `LAI` is included here. Since we're effectively planting seedlings, not seeds, it may be worth adding this as an (optional) parameter.
- I can't see the resolution of the discussion re: fine roots vs. coarse roots, but coarse and fine root data are provided here. Under the current schema, there's nowhere to put the coarse root data.
- Note that Nitrogen mass fractions for leaves, stems, and roots are also included, so @mdietze's comment about this is not just a hypothetical.

dlebauer:
What are next steps here?

dlebauer:
@infotroph what is the status of this PR? It looks like it is ready to merge after resolving conflicts.

infotroph:
The substantive changes here were already incorporated into #3919; this merge will just pick up changelog entries.

@mdietze's point in https://github.com/PecanProject/pecan/pull/3836#discussion_r2866489658 stands and deserves a follow-up that moves crop names from every-event to once-per-file. I have a local draft of that and will push it separately.

PR #3841, opened 2026-02-28
Refactor posterior loading in `get.parameter.samples()` into testable helper fn

### LINKED ISSUE #3662: Avoid overloading `pft$outdir` as an input path

S1DDHEY:
## Description

fixes #3662 
Extracts posterior loading logic from `get.parameter.samples()` into a dedicated `load.posteriors()` helper function with content-based posterior detection, Monte Carlo precedence, and backward-compatible legacy fallback (with deprecation warning).

## Changes
- New file: `modules/uncertainty/R/load_posteriors.R` with 3 functions `load.posteriors()`, `load.posteriors.from.path()`, `load.posteriors.legacy()`
- Content-based detection via `exists("trait.mcmc", envir = env)` instead of filename pattern matching
- Legacy fallback emits deprecation warning
- made required changes in `get.parameter.samples.R` to implement the helper fn 
- added tests for the helper functions

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


S1DDHEY:
Hi @infotroph @mdietze, 
Let me know if anything needs to be changed here, I'll add the `.Rd` files after the review.

Note on PDA detection: There is no content-level difference between MA and PDA `trait.mcmc` objects, both are named lists of `coda::mcmc.list`. The only distinguishing signal I found is the filename (`trait.mcmc.pda.*.Rdata`).

Currently `is.pda` in the returned result is set via `grepl("pda", filename)`, which matches the existing behavior in the original `get.parameter.samples()`. Are there any others ways you suggest this can be improved ?

S1DDHEY:
Hi @mdietze @infotroph,  just a gentle ping on this PR in case it got buried. Whenever you have a chance, I’d really appreciate any feedback.

modules/uncertainty/R/load_posteriors.R
```diff
@@ -0,0 +1,251 @@
...
+#' @author Siddhey Patil
+load.posteriors <- function(posterior.file,
+                            outdir = NULL,
+                            posteriorid = NULL,
+                            con = NULL,
+                            hostname = NULL) {
+  result <- list(
+    prior.distns = NULL,
+    trait.mcmc = NULL,
+    is.pda = FALSE
```
mdietze: In terms of what's returned, I think the key bit of information is whether the samples are independent or joint, not whether they were generated by the pda. Yes, it the default workflow the meta.analysis samples are independent and the pda samples are joint, but @ashiklom has code for generating joint samples in a meta.analysis.

modules/uncertainty/R/load_posteriors.R
```diff
@@ -0,0 +1,251 @@
...
+  if (!is.na(posterior.file)) {
+    result <- load.posteriors.from.path(posterior.file)
+  } else {
+    PEcAn.logger::logger.warn(
+      "Reading posterior from pft$outdir is deprecated. ",
+      "Please specify posterior.files explicitly."
+    )
+    result <- load.posteriors.legacy(
+      outdir = outdir,
+      posteriorid = posteriorid,
```
mdietze, line 38: it appears that number of function arguments are just for backwards compatibility and are slated for deprecation. I think it would be good to make that more clear in the roxygen function documentation

modules/uncertainty/R/load_posteriors.R
```diff
@@ -0,0 +1,251 @@
...
+#' @param hostname host name for db file lookup.
+#' @return list with \code{prior.distns}, \code{trait.mcmc}, \code{is.pda}.
+#' @keywords internal
+#' @author Siddhey Patil
+load.posteriors <- function(posterior.file,
+                            outdir = NULL,
+                            posteriorid = NULL,
+                            con = NULL,
+                            hostname = NULL) {
+  result <- list(
```
mdietze: in the code below, result is completely overwritten, so it's not clear to me what is gained by defining this list

modules/uncertainty/R/load_posteriors.R
```diff
@@ -0,0 +1,251 @@
...
+      load(f, envir = env),
+      error = function(e) {
+        PEcAn.logger::logger.warn(
+          "Failed to load posterior file: ", f, " -- ", conditionMessage(e)
+        )
+      }
+    )
+
+    # Detect Monte Carlo samples
+    if (exists("trait.mcmc", envir = env, inherits = FALSE)) {
```
mdietze: I wonder if we can get away from the assumption that the object is named "trait.mcmc" and instead detect by object class or coda::is.mcmc and coda::is.mcmc.list

S1DDHEY: Switched load.posteriors.from.path to iterate over all objects in the loaded environment and check each with `coda::is.mcmc.list()`. It handles two cases:

- A named list where every element passes `coda::is.mcmc.list()` (the standard `trait.mcmc` structure)
- A bare `mcmc.list` object (wrapped into a single-element named list)

Distribution summaries (`post.distns` / `prior.distns`) are still detected by variable name since they're plain data frames with no distinguishing class. Is there any other way you suggesting detecting these ?

modules/uncertainty/R/load_posteriors.R
S1DDHEY: Thanks for pointing out, totally missed it. Removed in the newer commits

modules/uncertainty/R/load_posteriors.R
S1DDHEY: Thanks for the clarification, renamed `is.pda` to `is.joint` across all three helpers and in `get.parameter.samples`. The return list is now `list(prior.distns, trait.mcmc, is.joint)`.

For detection in `load.posteriors.from.path`, I check for a companion `mcmc.pda.*` file in the same directory (PDA shoudl create one alongside `trait.mcmc.pda.*`, pls correct me if I a wrong here). This is a step toward content-based detection, though a more robust long-term approach would be to tag the object with `attr(trait.mcmc, "joint") <- TRUE` at save time, that would also cover @ashiklom's joint MA case. Happy to add that as a follow-up TODO comment if you'd like.

mdietze: I like the elegance of using attributes, but I also know from experience that we can't assume anyone working on calibration will remember to add this attribute 100% of the time, and we can't update legacy data. In practice, the practical way to tell if two parameters are jointly distributed is whether they are in the same mcmc.list or not (i.e., things in the same mcmc.list can be assumed to be jointly distributed). The opposite case would be when parameters are fit separately, in which case we'd store a list of mcmc.list objects, each named by the parameter being sampled. It's not impossible that someone would take a joint posterior and separate it out into independent lists under the (mistaken) impression that this is what the downstream code needs (since that's how the meta.analysis posteriors are structured), but I see this as less likely than them forgetting to include attributes. One known exception to this is cases where parameters in different PFTs are separated because PEcAn doesn't currently know how to handle the joint distribution across PFTs, but in practice isn't currently happening operationally. For future-proofing we'll have to think carefully about how we store posteriors for hierarchical multi-site runs -- at the moment I think the different posteriors are in separate files.

modules/uncertainty/R/load_posteriors.R
mdietze: Nice.

I think detecting post.distns and prior.distns by name is probably the best bet for now.

modules/uncertainty/R/load_posteriors.R
S1DDHEY: The current challenge I see is that `pda.postprocess` splits the joint posterior into separate per-trait `mcmc.list` objects before saving, so by the time `load.posteriors` sees the file, the structural signal is already lost. For this PR I'll keep the companion-file check (`mcmc.pda.*` in the same directory) as the best available heuristic.

As a follow-up, will it be worth updating `pda.postprocess` to save the joint posterior as a single multi-parameter `mcmc.list` instead of splitting it per trait, that would make the structural heuristic work naturally? I can open a separate issue for that if you'd like.

S1DDHEY:
@mdietze Please let me know, if there are any further changes needed to this, I'll add .Rd files accordingly.

modules/uncertainty/R/load_posteriors.R
mdietze: Great catch! Ok, lets go with your proposed `mcmc.pda.*` check for now.  Yes, opening an issue about this would be great. I'm also tagging @arob5 so he knows to keep future pda posteriors as joint distributions.

mdietze:
.Rd files: make sure those are workflow generated, not manual

mdietze (approved):

S1DDHEY:
@mdietze forgot to update the DESCRIPTION file, added now and fixed other ci issues as well

PR #3824, opened 2026-02-20
 prevent db connection when `bety` details missing

AritraDey-Dev:
## Description
Issue reported in slack(https://pecanproject.slack.com/archives/C8RD9BRLK/p1771538693295529) 
In the given pecan_2.xml used 
```
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<pecan>
<info>
<notes/>
<userid>-1</userid>
<username/>
<date>2025-06-19-15-34-01</date>
</info>
<outdir>/Users/jamesholmquist/GitHub/pecan/documentation/tutorials/Demo_1_Basic_Run/demo_outdir</outdir>
<pfts>
<pft>
<name>temperate.coniferous</name>
<posterior.files>pft/temperate.coniferous/prior.distns.Rdata</posterior.files>
<outdir>pft/temperate.coniferous</outdir>
</pft>
</pfts>
<ensemble>
<variable>NPP</variable>
<size>1</size>
<start.year>2004</start.year>
<end.year>2004</end.year>
<samplingspace>
<parameters>
<method>uniform</method>
</parameters>
<met>
<method>sampling</method>
</met>
</samplingspace>
</ensemble>
<model>
<type>SIPNET</type>
<revision>git</revision>
<delete.raw>FALSE</delete.raw>
<binary>demo_outdir/sipnet</binary>
<id>-1</id>
</model>
<run>
<site>
<met.start>2004/01/01</met.start>
<met.end>2004/12/31</met.end>
<name>Niwot Ridge Forest/LTER NWT1 (US-NR1)</name>
<lat>40.0329</lat>
<lon>-105.546</lon>
<id>-1</id>
</site>
<inputs>
<met>
<source>AmerifluxLBL</source>
<output>SIPNET</output>
<username>Aritra_2004</username>
<path>
<path1>dbfiles/AMF_US-NR1_BASE_HH_23-5.2004-01-01.2004-12-31.clim</path1>
</path>
</met>
</inputs>
<start.date>2004/01/01</start.date>
<end.date>2004/12/31</end.date>
</run>
<host>
<name>localhost</name>
<rundir>/Users/jamesholmquist/GitHub/pecan/documentation/tutorials/Demo_1_Basic_Run/demo_outdir/run</rundir>
<outdir>/Users/jamesholmquist/GitHub/pecan/documentation/tutorials/Demo_1_Basic_Run/demo_outdir/out</outdir>
</host>
<settings.info>
[...]

PR #3845, opened 2026-03-03
get.parameter.samples: skip DB connection attempt if no params given

infotroph:
## Description

The existing `try` kept connection failures from killing the run, but let a scary-looking Postgres connection failure bubble up. This patch replaces that with a more reassuring "will not use Bety" when `settings$database$bety` doesn't exist.

I retained the `try` for runs that do have Bety params, partly to handle other kinds of error and partly out of laziness.

## Motivation and Context

Continues from #3824 in trying to eliminate unwanted DB connection attempts in workflows that didn't request them.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/uncertainty/R/get.parameter.samples.R
+++ b/modules/uncertainty/R/get.parameter.samples.R
@@ -25,8 +25,15 @@ get.parameter.samples <- function(settings,
   }
 
   ## Open database connection
-  con <- try(PEcAn.DB::db.open(settings$database$bety))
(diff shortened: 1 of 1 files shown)


mdietze (approved):

AritraDey-Dev (commented):
Thanks a lot for this since the log msg is very much confusing to user.Initially, while implementing the notebook, I thought of doing this, and finally it’s done.

Issue #3736, opened 2025-12-25
Revisit special handling of `seedling_mortality` in run.meta.analysis


infotroph:
Current lines 103-104 of `run.meta.analysis.R`:

```
prior_variances[names(trait_average), ] <- 0.001 * trait_average ^ 2
  prior_variances["seedling_mortality", 1] <- 1
```

@infotroph in a [comment](https://github.com/PecanProject/pecan/pull/3728#discussion_r2646513164) on #3278:

 > Would it be possible to add a few comments here explaining (1) where the constants 0.001 and 1 come from, and (2) why seedling mortality gets special handling and whether that's a hack or a principled design choice?
>
> On scrolling down further I see this is pre-existing code moved up from run.meta.analysis.pft, so maybe these explanations are lost to time...

@ashiklom [replied](https://github.com/PecanProject/pecan/pull/3728#discussion_r2647225110):

> Yeah, this is pre-existing code. The oldest git blame gets me is here:
>
> https://github.com/PecanProject/pecan/blame/bed27a2b37619c2297c3c66becf97b8c16233494/modules/meta.analysis/R/run.meta.analysis.R
>
> ...but I'm almost certain this logic predates that, which looks more like refactoring than development.
>
> The more correct thing to do here is probably (1) remove the `seedling_mortality` hard-coded value altogether and replace it with a global default prior of some kind in the database; (2) pass the `0.001` as an (optional) argument called something like `default_prior_variance_fraction` with 0.001 as the default value.

aditi25srivastava:
@infotroph Heyy!!!!! I would love to work on this issue. Could you please assign it to me ?

infotroph:
@aditi25srivastava You're very welcome to take a crack at it, but be aware that the task is incompletely specified at the moment, so the bulk of the work will be in deciding how exactly to implement it in the context of PEcAn's statistical framework.

aditi25srivastava:
Thank you @infotroph for the clarification, that makes sense 
i understand that the task is still loosely specified and depends on the final React/framework decisions. I won't start implementing anything until the approach is clearly defined.. I'm keeping an eye on the issue and would be happy to help once the requirements are finalised. Please feel free to tag or assign me when it's ready to be picked up

mdietze:
Just took a quick look and I'll be honest that I'm more concerned about the (undocumented) default logic around `prior_variances` than the seedling_mortality special case (which I think can probably be dropped). 

```
[...]

PR #3850, opened 2026-03-04
fix(meta.analysis): replace gamma_tau with prior_n/prior_cv, fix tauB…

shreyannandanwar:
issue #3736 
related #3728

This PR addresses the undocumented prior logic in `meta_analysis_standalone()` inside `run_meta_analysis.R`. It replaces magic constants with named, statistically meaningful parameters, removes the hardcoded seedling_mortality special case, and fixes an incorrect tauB calculation.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:


shreyannandanwar:
The new parameters are currently only exposed in `meta_analysis_standalone()`. The full workflow call chain (`runModule.run.meta.analysis → run.meta.analysis → run.meta.analysis.pft`) still uses defaults and doesn't pass `prior_n`, `prior_cv`, `tauA`, or `tauB` through. I plan to address this in a follow-up PR once this approach is reviewed and approved.
Thank you and happy to hear any feedback.

modules/meta.analysis/R/run.meta.analysis.R
```diff
@@ -195,7 +215,10 @@ check_consistent <- function(point, prior,
 #'  posteriors with new ones
 #'
 #' @inheritParams meta_analysis_standalone
-run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.2, dbfiles, dbcon, use_ghs = TRUE, update = FALSE) {
+run.meta.analysis.pft <- function(pft, iterations, random = TRUE, threshold = 1.2,
+                                  dbfiles, dbcon, use_ghs = TRUE, update = FALSE,
+                                  prior_n = 0.02, prior_cv = sqrt(0.001),
+                                  tauA = NULL, tauB = NULL) {
```
ashiklom: If these parameters will not be used by the function, they should be removed. Exposing arguments that don't get used is confusing and misleading.

ashiklom (changes requested):

ashiklom (changes requested):
Introducing arguments that silently don't get used is bad practice. If we haven't propagated the full logic through to `run.meta.analysis.pft`, remove the parameters until we have.

shreyannandanwar:
@ashiklom 
thank you for the feedback. I will start working on it right now.
just wanted to know - Is the approach to solving the issue adequately implemented in `meta_analysis_standalone()` ?

shreyannandanwar:
`tauA` and `tauB` are not yet wired into runModule since they are advanced power-user overrides unlikely to be set via XML. Happy to add them if you'd prefer consistency.

PR #3851, opened 2026-03-05
Gh 2666 fix gssurgo bety registration clean

### LINKED ISSUE #2666: Registration of gSSURGO soil.nc files

man080107:
Fix BETY registration of gSSURGO soil.nc files (closes #2666)

## Problem

Soil `.nc` files generated by the NASA CMS SDA workflow were never registered in BETY.
This was caused by two bugs in [soil_process.R](cci:7://file:///c:/Users/Man/OneDrive/Desktop/pecan2/pecan/modules/data.land/R/soil_process.R:0:0-0:0):

1. **gSSURGO path**: `dbfile.input.insert()` was placed *inside* the
   `if (length(newfile) == 0)` block, so it only ran when files were freshly
   extracted. Pre-existing `.nc` files (generated externally, e.g. by SDA runs)
   were found on disk and the registration step was silently skipped.

2. **PalEON_soil path**: `extract_soil_nc()` ran and produced files, but there
   was **no `dbfile.input.insert()` call at all** — files were never registered.

### [modules/data.land/R/soil_process.R](cci:7://file:///c:/Users/Man/OneDrive/Desktop/pecan2/pecan/modules/data.land/R/soil_process.R:0:0-0:0)
- Moved `dbfile.input.insert()` **outside** the `if (length(newfile) == 0)` block
  in the gSSURGO path. Now runs for **all files**, whether newly extracted or
  pre-existing on disk. `dbfile.input.insert()` is idempotent — it will not create
  duplicates if the file is already registered.
- Fixed path accessor: used `as.character(newfile[[i]])` instead of
  `newfile[[i]]$path` since `extract_soil_gssurgo()` returns named character
  strings, not sub-lists.
- Added a log message when pre-existing files are found and re-registered.
- Added BETY registration to the **PalEON_soil code path**, which was previously
  completely missing it.

### [scripts/retroactive_register_gSSURGO_soils.R](cci:7://file:///c:/Users/Man/OneDrive/Desktop/pecan2/pecan/scripts/retroactive_register_gSSURGO_soils.R:0:0-0:0) (new)
- One-time script to retroactively register the existing `gSSURGO_site_xxxx`
  soil.nc files on the BU server (`/fs/data1/pecan.data/dbfiles`) that were
  generated by @para2x but never registered.
- Has a `dry_run = TRUE` mode to safely preview what would be registered before
  actually writing to BETY.

## Testing

- [ ] Run [retroactive_register_gSSURGO_soils.R](cci:7://file:///c:/Users/Man/OneDrive/Desktop/pecan2/pecan/scripts/retroactive_register_gSSURGO_soils.R:0:0-0:0) with `dry_run = TRUE` on BU
      server to verify it finds the correct files
- [ ] Run with `dry_run = FALSE` to register existing files
- [ ] Verify new `soil_process()` calls register files correctly going forward

## Related
- Closes #2666
- Related to previous partial fix in #2668 (which only covered the case where
  files did NOT already exist on disk)


man080107:
Hi @mdietze,,

I've implemented the fix for the BETY registration issue for gSSURGO soil.nc files (closes #2666). The changes ensure registration works for both newly extracted and pre-existing files, and also add the missing registration for the PalEON_soil path. I also included a script to retroactively register existing files on the BU server.

Could you please review it when you have time and let me know if any changes are needed? Thanks!

modules/data.land/R/soil_process.R
```diff
@@ -48,86 +50,137 @@ soil_process <- function(settings, input, dbfiles, overwrite = FALSE,run.local=T
...
-                             enddate = NULL,
-                             mimetype =  "application/x-netcdf",
-                             formatname = "pecan_soil_standard",
-                             con = con,
-                             ens=TRUE)
-      }
-
-
-      }
+    # Register ALL files in BETY — whether newly extracted or pre-existing.
```
mdietze, line 87: This makes me nervous as it puts a lot of trust on dbfile.input.insert to detect that a file hasn't been registered in the DB already. We tried to implement a similar logic in the meta-analysis workflow and it resulted in hundreds (possibly thousands) of entries that a human would have identified as duplicates but which the code didn't catch because of small logic errors.

Also not that there's been a general trend towards making the code more BETY optional -- regardless of where this is put we need to make sure that the code skips these chunks once (not separately checking within the function called in the `for`) if there's no BETY connection.

modules/data.land/inst/retroactive_register_gSSURGO_soils.R
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env Rscript
+#
+# retroactive_register_gSSURGO_soils.R
```
mdietze, line 3: this script looks promising but (a) it should live in modules/data.land/inst and (b) could you provide evidence that it runs correctly?

man080107:
Thanks for the feedback @mdietze!

**(a)** I've used `git mv` to move [retroactive_register_gSSURGO_soils.R](cci:7://file:///c:/Users/Man/OneDrive/Desktop/pecan2/pecan/scripts/retroactive_register_gSSURGO_soils.R:0:0-0:0) into `modules/data.land/inst/`.

**(b)** I also refactored the script to accept Environment variables like `DBFILES_ROOT=...` and skip the `DB.open` check if `DRY_RUN=TRUE` so it can be safely and easily tested. Here is the output showing that it safely finds `.nc` files correctly, parses the `site_id=42`, matches the dry run expectation, and registers without crashing!

```text
INFO  [ 2026-03-10 12:05:00 ] Found 1 gSSURGO_site_* directories.
INFO  [ 2026-03-10 12:05:00 ] [DRY RUN] Would register: /tmp/pecan_test/gSSURGO_site_42/site_42_soil.nc for site_id=42

===== SUMMARY =====
  Directories found : 1
  Files registered  : 1 (dry run — not actually written)
  Directories skipped (no .nc): 0
  Errors            : 0
===================

File: modules/data.land/inst/retroactive_register_gSSURGO_soils.R [hunk 3]
```diff
@@ -27,25 +27,29 @@ library(PEcAn.logger)
...
-dbfiles_root <- "/fs/data1/pecan.data/dbfiles"   # root where gSSURGO folders live
+dbfiles_root <- Sys.getenv("DBFILES_ROOT", "/fs/data1/pecan.data/dbfiles")   # root where gSSURGO folders live
 bety_settings <- list(
-  driver   = "PostgreSQL",
-  user     = "bety",
-  password = "bety",
-  host     = "localhost",
-  dbname   = "bety",
-  port     = 5432
+  driver   = Sys.getenv("BETY_DRIVER", "PostgreSQL"),
```
mdietze: Did someone suggest these changes? Are these env variables defined by default in the PEcAn workflow, or would someone need to know that they need to define them? I'd probably revert this bit

modules/data.land/inst/retroactive_register_gSSURGO_soils.R
```diff
@@ -27,25 +27,29 @@ library(PEcAn.logger)
 library(stringr)
 
 #------------------------------------------------------------------------------
-# CONFIGURATION — edit these before running
+# CONFIGURATION — edit these before running or pass via environment variables
 #------------------------------------------------------------------------------
-dbfiles_root <- "/fs/data1/pecan.data/dbfiles"   # root where gSSURGO folders live
+dbfiles_root <- Sys.getenv("DBFILES_ROOT", "/fs/data1/pecan.data/dbfiles")   # root where gSSURGO folders live
```
mdietze: I thought the original was fine, but if you want greater flexibility wouldn't the command line be a more likely location to get the path than an env variable? e.g., `./retroactive_register_gSSURGO_soils.R <folder>`

mdietze:
resolved comments on the retroactive script, but still looking for a response on soil_process.R

modules/data.land/inst/retroactive_register_gSSURGO_soils.R
man080107: You're right, those env variables (BETY_DRIVER, BETY_USER, BETY_PASSWORD, BETY_HOST, BETY_DBNAME, BETY_PORT) are not defined anywhere in the default PEcAn workflow — I added them on my own thinking it would offer more flexibility, but it would just burden the user with having to know they need to define them. Reverted to hardcoded defaults that match the standard local BETY setup. Thanks for catching this!

man080107:
Thanks for the detailed feedback! @mdietze, Two responses:

On duplicates: I share the concern about dbfile.input.insert missing duplicates. To address this, I added an explicit db.query("SELECT id FROM dbfiles WHERE file_path=... AND file_name=...") check before each call to dbfile.input.insert, so we only insert if the file isn't already in the dbfiles table. This should be more reliable than relying on the function's internal logic alone.

On BETY-optional: The !is.null(con) check already wraps the entire for loop as a single guard (not per-call inside the loop), so if there's no BETY connection the whole registration block is skipped at once. Happy to restructure it further if you'd prefer a different placement.

mdietze (approved):

man080107:
Hi @mdietze, I already have my GSoC proposal mostly ready. Can I also include this PR (#3851) as one of my contributions in the proposal?

PR #3853, opened 2026-03-06
Remove conversion and checkTemp helpers per TODO

Thomas-Sedhom:
## Description

This PR removes the local ```conversion``` and ```checkTemp``` helper functions from ```models/ed/R/model2netcdf.ED2.R``` and replaces their usage with inline ```ifelse``` statements at the existing call sites.
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


models/ed/R/model2netcdf.ED2.R
```diff
@@ -752,19 +752,6 @@ put_T_values <-
   umol2kg_C <- Mc * PEcAn.utils::ud_convert(1, "umol", "mol") * PEcAn.utils::ud_convert(1, "g", "kg")
   yr2s      <- PEcAn.utils::ud_convert(1, "s", "yr")
   
-  # TODO - remove this function and replace with ifelse statements inline below (SPS)
```
infotroph, line 755: I didn't write this comment (that was probably @serbinsh years ago), but I recommend against it -- inline ifelse statements get messy fast and don't show intent as well as named helpers. 

But while I'm looking at it, a higher-level fix that would eliminate the need for these operations would be to set all the missing values in `out` to literal NAs instead of -999. Then numeric conversions can be done on whole columns without worrying about disturbing the missing value flags, and the ncdf4 library will handle them automatically. 

This suggestion does assume `out` is only written to netcdf and not passed to any other processes that need the missing values to be -999, so someone who knows the ED model should weigh in on whether that's true. Even if it does get passed elsewhere, it may still be cleaner overall to convert the NAs once at the end rather than work around them when converting each column.

infotroph (changes requested):
Thanks for opening this. See my inline comment for the reasons I don't think ifelse is the right direction.

As a more general point, it's worth opening an issue to discuss changes like this even if there's an existing TODO -- first because programmers (me included!) tend to write TODOs when they're in a hurry and chances are they intended to defer the "thinking about the design and considering other options" part as well as the implementation of the change, second because often the code changes around the comment and leaves it outdated, and third because we have so many TODOs in the codebase that it's worth checking on priorities before picking up any one of them. Personally I'd rate this one as low priority but worth finishing now that you've started on it.

models/ed/R/model2netcdf.ED2.R
Thomas-Sedhom: I took a closer look at the data flow in model2netcdf.ED2.R. From what I found, the out object built by read_T_files() / read_E_files() is mainly used as an intermediate step before writing to NetCDF, and I did not find any internal code path in this repo where it is passed to other ED processing that specifically depends on -999.
I also noticed that the E-file path already seems to rely on NA internally in some places and still writes NetCDF variables with missval = -999, so your suggestion looks reasonable to me. The main remaining concern seems to be whether changing read_T_files() output from -999 to NA would affect any external callers, since these helper functions are exported. If that concern is not important, I agree that using NA internally and only handling missing values at the NetCDF boundary would likely be cleaner.

Thomas-Sedhom:
> Thanks for opening this. See my inline comment for the reasons I don't think ifelse is the right direction.
> 
> As a more general point, it's worth opening an issue to discuss changes like this even if there's an existing TODO -- first because programmers (me included!) tend to write TODOs when they're in a hurry and chances are they intended to defer the "thinking about the design and considering other options" part as well as the implementation of the change, second because often the code changes around the comment and leaves it outdated, and third because we have so many TODOs in the codebase that it's worth checking on priorities before picking up any one of them. Personally I'd rate this one as low priority but worth finishing now that you've started on it.

Sorry about that, I wasn't aware of the convention around TODOs, I assumed the comment was enough to justify jumping in. Good to know for the future. Should I go ahead and open an issue for this so we can continue the discussion there?

models/ed/R/model2netcdf.ED2.R
Thomas-Sedhom: I took a closer look at the data flow in ```model2netcdf.ED2.R```. From what I found, the out object built by ```read_T_files() / read_E_files()``` is mainly used as an intermediate step before writing to ```NetCDF```, and I did not find any internal code path in this repo where it is passed to other ```ED``` processing that specifically depends on -999.
I also noticed that the ```E-file``` path already seems to rely on NA internally in some places and still writes ```NetCDF``` variables with ```missval = -999```, so your suggestion looks reasonable to me. The main remaining concern seems to be whether changing ```read_T_files()``` output from -999 to NA would affect any external callers, since these helper functions are exported. If that concern is not important, I agree that using NA internally and only handling missing values at the ```NetCDF``` boundary would likely be cleaner.

infotroph:
@Thomas-Sedhom Are you interested in working more on this? If not I'll close it just to keep the PR queue manageable but will keep being happy to review work related to it.

Issue #1958, opened 2018-06-03
Update bety.sql for Travis


infotroph:
**Description**

Travis tests currently install Bety from http://isda.ncsa.illinois.edu/~kooper/PEcAn/data/bety.sql, which best I can tell was last updated in spring 2017 and therefore does't contain newer entries, notably including the PFT definitions needed for full PEcAn.DB tests. 

If this is still the correct location to install from, bety.sql should be updated to at least include PFTs "Pavi_alamo" and "Pavi_all" from Bety-0. If there's a more up-to-date bety dump elsewhere, .travis.yml should be updated to point toward it.

(P.S. Would it make sense to gzip-compress bety.sql for faster copying?)

**Context** 

The above-mentioned test failures from PEcAn.DB were being masked by a change in how testthat's `reporter` argument works. I've fixed the test arguments in my [fix-tests](https://github.com/infotroph/pecan/tree/fix-tests) branch and temporarily skipped the PFT tests; I will re-enable them once the bety dump is updated.

robkooper:
I can recreate a dump. I want to make sure we use a dump that has only public data, so we can't dump it from one of the main servers. I might be able to do this on the pecan.ncsa.illinois.edu server.

infotroph:
@robkooper I'm testing this out in https://github.com/infotroph/pecan/tree/travis-new-betydump. Failures visible in the current-as-of-this-writing Travis build: https://travis-ci.org/infotroph/pecan/builds/394442188

infotroph:
https://github.com/PecanProject/bety/pull/695 provides a Docker image that covers this use case better than the static dump we were discussing here.

PR #3856, opened 2026-03-08
Add regression tests for get.trait.data.pft() before modularity refactor

omkarrr2533:
Regression tests for get.trait.data.pft()
Added test coverage for get.trait.data.pft() before starting the modularity refactor. No source changes — just tests.

Tests cover: input validation, error paths (bad PFT name, ambiguous name), file output and contents, return value shape, empty PFT handling (soil.ALL), and end-to-end consistency. Original error cases and the skipped cultivar test (#1958) are preserved.

All DB tests clean up via cleanup_posterior() — no orphaned rows.

Removed the old caching tests — traced through the source and the cache never actually hits with write=FALSE and posteriorid=NULL. The function deletes and recreates outdir contents on every call.

Without DB:
Screenshot 2026-03-27 224821

[ FAIL 0 | WARN 0 | SKIP 10 | PASS 2 ]

With DB (Codespace + Docker BETY):
image

[ FAIL 0 | WARN 2 | SKIP 1 | PASS 22 ]

Warnings are upstream (dbplyr interface, soil.ALL no priors). Skip is cultivar test #1958.

### DIFF
--- a/base/db/tests/testthat/test-get.trait.data.pft.R
+++ b/base/db/tests/testthat/test-get.trait.data.pft.R
@@ -7,6 +7,17 @@
 #
 # Tests requiring a live BETYdb connection skip cleanly when unavailable.
 # ── Fixtures ──────────────────────────────────────────────────────────────────
+# Helper needed for the restored cultivar test
+dbdir <- file.path(tempdir(), "dbfiles")
+
+get_pft <- function(pftname) {
+  get.trait.data.pft(
+    pft         = list(name = pftname, outdir = withr::local_tempdir()),
+    trait.names = "SLA",
+    dbfiles     = dbdir,
+    modeltype   = NULL,
+    dbcon       = test_dbcon)
+}
 
 make_test_pft <- function(outdir) {
   list(
@@ -58,19 +69,6 @@ test_dbcon <- tryCatch(
   error = function(e) NULL
 )
 
-# ── Block 0: Tests that work WITHOUT a database ──────────────────────────────
-
-test_that("make_test_pft returns correctly structured list", {
-  outdir <- withr::local_tempdir()
-  pft <- make_test_pft(outdir)
-  
-  expect_true(is.list(pft))
-  expect_equal(pft$name, "temperate.Early_Hardwood")
-  expect_equal(pft$outdir, outdir)
-  expect_null(pft$posteriorid)
-  expect_true(is.list(pft$constants))
-})
-
 test_that("get.trait.data.pft fails gracefully with NULL dbcon", {
   outdir <- withr::local_tempdir()
   
@@ -99,46 +97,10 @@ test_that("get.trait.data.pft fails with invalid modeltype and NULL dbcon", {
   )
 })
 
-test_that("get.trait.data.pft accepts trait.names parameter", {
-  expect_true(
-    "trait.names" %in% names(formals(get.trait.data.pft))
-  )
-})
-
-test_that("make_empty_pft returns correctly structured list", {
-  outdir <- withr::local_tempdir()
-  pft <- make_empty_pft(outdir)
-  
-  expect_true(is.list(pft))
-  expect_equal(pft$name, "soil.ALL")
-  expect_equal(pft$outdir, outdir)
-  expect_null(pft$posteriorid)
-  expect_true(is.list(pft$constants))
-})
-
-test_that("get.trait.data.pft function exists and is callable", {
-  expect_true(is.function(get.trait.data.pft))
-})
-
-test_that("get.trait.data.pft has expected parameters", {
-  params <- names(formals(get.trait.data.pft))
-  
-  expect_true("pft"         %in% params)
-  expect_true("modeltype"   %in% params)
-  expect_true("dbfiles"     %in% params)
-  expect_true("dbcon"       %in% params)
-  expect_true("trait.names" %in% params)
-})
-
 test_that("get.trait.data.pft errors with no arguments", {
   expect_error(get.trait.data.pft())
 })
 
-test_that("std_traits contains expected trait names", {
-  expect_true("SLA"    %in% std_traits)
-  expect_true("Vcmax"  %in% std_traits)
-  expect_equal(length(std_traits), 3)
-})
 
 # ── Block 1: Files are written to disk ────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


omkarrr2533:
Hi @mdietze,

Really excited that #3860 got merged — your feedback made it so much better!

I have #3856 ready too and I think you'll find it a straightforward read. It just locks down the current behavior of get.trait.data.pft() with regression tests before the refactor starts. No source changes at all, just a clean test file.

Looking forward to hearing your thoughts whenever you get a chance. Thanks so much for everything! 😊

mdietze:
A few quick questions
1. the way the PR diff is formatted it is unclear which previous tests were retained in modified form and which were completely deleted. Could you please comment on the status of the existing test
2. The screenshot you showed illustrates which tests pass when there is no BETY connection. Could you verify that the tests also pass when there is a BETY connection.
3. Given the style and volume of tests added, could you comment on the role that any AI / LLM tools may have been used in constructing the tests and, if they were, how you've verified that all code is correct and all tests are useful and necessary. I'll also note that its not 100% universally clear who the author of code written by a LLM actually is and whether such code falls within our open source requirements.

omkarrr2533:
The three original tests are preserved:

1. "error cases complain" — same two assertions (NOTAPFT, soil), 
   split into separate test_that blocks for individual skip guards.

2. "reference species and cultivar PFTs" — retained as-is, still 
   skipped (#1958). Only style changes (= → <-).

3. "PFT with no trait data (SIPNET soil)" — one intentional change: 
   the original called get.trait.data() (the wrapper), not 
   get.trait.data.pft(). I updated it to test the function this 
   file is named for. Cleanup logic preserved via withr::defer().

The get_pft() helper is retained. Infrastructure changes 
(context(), check_db_test(), teardown()) replaced with modern 
equivalents (skip_if_no_db(), withr::defer()).

omkarrr2533:
2. Tests with BETY connection

I wasn't able to verify this yet. CI's test job also skips the DB-dependent tests (SKIP 17 | PASS 3) — it doesn't appear to have a BETYdb connection configured. Locally I'm on an ARM-based machine where the pecan/bety and mdillon/postgis Docker images won't run (exec format error).

The 3 non-DB tests pass on both CI and locally. The 16 DB-dependent tests use skip_if_no_db() and are structured to run when a connection is available.

If you could point me to a machine with BETY access, or suggest the preferred way to run DB-connected tests in PEcAn's workflow, I'm happy to verify and post the results.

Screenshot 2026-03-21 113528
Screenshot 2026-03-21 113732

omkarrr2533:
I used an LLM to help scaffold test templates after I had already 
identified the behaviors to cover by reading the source. I then 
verified each test by tracing get.trait.data.pft(). For example, 
the "discards computed data" test exists because the function saves 
trait_data and prior_distns to disk but only returns the pft list 
— which is the core behavior the modularity refactor will change.

The initial output included around 6 extra tests that were only 
testing my own helpers rather than PEcAn code, so I removed those 
in this push.

On the licensing concern — happy to follow whatever policy PEcAn 
adopts, and I can rewrite any tests if needed. 

Thanks for the thoughtful review — really appreciate you taking the time. 😊

omkarrr2533:
Verified the tests with a live BETYdb connection using Docker in a GitHub Codespace:

Screenshot 2026-03-21 192359

35 tests pass with DB connected. The 5 warnings are all from upstream code (dbplyr interface, soil.ALL having no priors, file cleanup edge case). The 1 skip is the intentional cultivar test (#1958).

Without DB: 3 pass, 17 skip cleanly.

omkarrr2533:
Rebased against latest develop and cleaned up the test file a bit.
Just wanted to flag that the 5 CI failures aren't coming from my file — they're both pre-existing issues in test.symmetric-setdiff.R and test-query.traits.R. The symmetric-setdiff one expects an integer64 warning that's no longer being emitted, and the query.traits one expects median log messages but gets a transformstats error instead. test-get.trait.data.pft.R itself contributes 0 failures to that [ FAIL 5 | WARN 2 | SKIP 12 | PASS 193 ] result.

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
-context("get.trait.data.pft")
+# Tests for get.trait.data.pft()
+#
+# Regression safety net to lock down current behavior before the modularity
+# refactor. DB-dependent tests skip cleanly when no connection is available.
```
infotroph: Delete -- These comments are useful context for reviewing this PR, but will be less useful to future readers.

```suggestion
```

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
...
+    skip("No test database connection available")
+  }
+  has_schema <- tryCatch(
+    DBI::dbExistsTable(dbcon, "pfts"),
+    error = function(e) FALSE
+  )
+  if (!has_schema) {
+    skip("Test database has no pfts table")
+  }
+}
```
infotroph: The existing check_db_test() helper does effectively this already -- no need to reimplement

mdietze (commented):
Numerous tests failing

omkarrr2533:
Just wanted to clarify the CI failures — none of them are coming from my file. test-get.trait.data.pft.R contributes 0 failures, all my DB tests skip cleanly without a connection.

The 5 failures are pre-existing on develop — test.symmetric-setdiff.R expects an integer64 warning that's no longer emitted, and test-query.traits.R expects median log lines but gets a transformstats error instead. The Docker failures are from an invalid version specification 'r81283' in modules/data.atmosphere on develop.

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
...
-  sp_csv = file.path(dbdir, "posterior", pavi_sp$posteriorid, "species.csv")
-  sp_trt = file.path(dbdir, "posterior", pavi_sp$posteriorid, "trait.data.csv")
+  sp_csv <- file.path(dbdir, "posterior", pavi_sp$posteriorid, "species.csv")
+  sp_trt <- file.path(dbdir, "posterior", pavi_sp$posteriorid, "trait.data.csv")
   expect_true(file.exists(sp_csv))
   expect_true(file.exists(sp_trt))
-  expect_gt(file.info(sp_csv)$size, 40) # i.e. longer than the 40-char header
-  expect_gt(file.info(sp_trt)$size, 215) # ditto 215-char header
+  expect_gt(file.info(sp_csv)$size, 40)
+  expect_gt(file.info(sp_trt)$size, 215)
```
infotroph: These comments should be retained -- even though I don't know which header they're talking about, it's still much clearer than bare unexplained numbers
```suggestion
  expect_gt(file.info(sp_csv)$size, 40) # i.e. longer than the 40-char header
  expect_gt(file.info(sp_trt)$size, 215) # ditto 215-char header
```

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
...
   pavi_cv <- get_pft("Pavi_alamo")
   expect_equal(pavi_cv$name, "Pavi_alamo")
-  cv_csv = file.path(dbdir, "posterior", pavi_cv$posteriorid, "cultivars.csv")
-  cv_trt = file.path(dbdir, "posterior", pavi_cv$posteriorid, "trait.data.csv")
+  cv_csv <- file.path(dbdir, "posterior", pavi_cv$posteriorid, "cultivars.csv")
+  cv_trt <- file.path(dbdir, "posterior", pavi_cv$posteriorid, "trait.data.csv")
   expect_true(file.exists(cv_csv))
   expect_true(file.exists(cv_trt))
-  expect_gt(file.info(cv_csv)$size, 63) # cultivar.csv headers are longer
+  expect_gt(file.info(cv_csv)$size, 63)
```
infotroph:
```suggestion
  expect_gt(file.info(cv_csv)$size, 63) # cultivar.csv headers are longer
```

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
...
+      driver   = "PostgreSQL",
+      user     = Sys.getenv("PECAN_TEST_DB_USER", "bety"),
+      password = Sys.getenv("PECAN_TEST_DB_PASS", "bety"),
+      host     = Sys.getenv("PECAN_TEST_DB_HOST", "localhost"),
+      dbname   = Sys.getenv("PECAN_TEST_DB_NAME", "bety"),
+      port     = as.integer(Sys.getenv("PECAN_TEST_DB_PORT", "5432"))
+    )
+  ),
+  error = function(e) NULL
+)
```
infotroph: check_db_test does all of this for you too

infotroph:
>  test.symmetric-setdiff.R expects an integer64 warning that's no longer emitted, and test-query.traits.R expects median log lines but gets a transformstats error instead

Neither of those is failing in the most recent develop run (see [here](https://github.com/PecanProject/pecan/actions/runs/25385296326/job/74444587637#step:12:328) and [here](https://github.com/PecanProject/pecan/actions/runs/25385296326/job/74444587637#step:12:341)), so I fear you've likely uncovered some lack of isolation between test files that will need to be resolved.


> invalid version specification 'r81283' 

This was an upstream problem in r-universe that's reportedly now resolved.

You _will_ continue to see failures in the Docker GHA job and it's OK to ignore them -- we have it set up in a way that relies on access to repository secrets that are not accessible to runs started from forks.

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
...
+    error = function(e) FALSE
+  )
+  if (!has_schema) {
+    skip("Test database has no pfts table")
+  }
+}
+
+cleanup_posterior <- function(dbcon, posteriorid) {
+  if (!is.null(posteriorid)) {
+    try(DBI::dbExecute(dbcon,
```
infotroph, line 34: If you're trying to maintain equivalence with the existing tests, need to send `DELETE FROM dbfiles WHERE container_type = 'Posterior' AND container_id = $1` too

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,300 @@
...
+  error = function(e) NULL
+)
+
+old_log_level <- PEcAn.logger::logger.getLevel()
+PEcAn.logger::logger.setLevel("WARN")
+
+withr::defer({
+  unlink(dbdir, recursive = TRUE)
+  PEcAn.logger::logger.setLevel(old_log_level)
+}, envir = testthat::teardown_env())
```
infotroph: Check me on this, but I believe `testthat::teardown_env()` fires after all tests of the whole package while the `testthat::teardown` you're removing ran at the end of this file. This might explain the CI failures you're seeing in other DB tests.

infotroph (changes requested):

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: done removed

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: Done, removed it and switched to skip_if(is.null(test_dbcon), ...) throughout.

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: Restored, thanks!

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: Done, added the comment back.

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: Replaced with check_db_test(), much cleaner.

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: Done, added the DELETE FROM dbfiles before the posteriors delete. Thanks for catching that!

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: You were right — switched to teardown({}) so it's file-scoped. The logger level was leaking into other test files which was causing the failures in test-query.traits.R. 👍

mdietze (approved):

base/db/tests/testthat/test-get.trait.data.pft.R
```diff
@@ -1,81 +1,270 @@
...
+    dbfiles     = dbdir,
+    modeltype   = NULL,
+    dbcon       = test_dbcon
+  )
 }
 
-test_that("reference species and cultivar PFTs write traits properly",{
+# ── DB connection and teardown ───────────────────────────────────────────────
+
+test_dbcon <- check_db_test()
```
infotroph: The intended behavior of check_db_test() is that it issues a `skip()` when the connection fails, so if I'm reading right this will skip the whole file before any of the `skip_if(is.null(test_dbcon))` tests below take effect. I'd lean toward setting up a new connection inside each test_that block -- yes, it's good practice to reuse Postgres connections where we can, but here I think the enhanced block isolation is worth the minimal extra connection overhead.

infotroph (approved):

base/db/tests/testthat/test-get.trait.data.pft.R
omkarrr2533: Done moved check_db_test() inside each test_that block with a matching withr::defer(db.close(...)). No more file-level connection.

PR #3860, opened 2026-03-10
feat: extract pure parameter sampling logic + tests (Modularity Part 1)

omkarrr2533:
Hi team! 👋 

As part of my preparation for the GSoC "Increase PEcAn Modularity" project, I'm working on refactoring the `PEcAn.uncertainty` pipeline. I know reviewing massive refactor PRs is a pain, so I am breaking this work up into small, bite-sized PRs.

**This is Part 1.** It is strictly an *addition* and doesn't break any existing workflows.

### What's inside:
1. **New `get_parameter_samples_pure()` function:** I extracted the core mathematical/sampling logic out of the current disk-coupled function. This new version takes explicit R objects in memory and returns the sample lists directly. Zero database calls, zero disk I/O.
2. **Comprehensive Tests:** Added `test-get_parameter_samples_pure.R` with 28 tests covering MCMC routing, prior-only sampling, SA generation, and the various sampling methods. Because it doesn't need BETYdb, the whole suite runs locally in about ~2 seconds with 0 fails. 

### What's NOT inside (Next Steps):
I haven't touched the legacy `get.parameter.samples()` function in this PR. To keep things safe and easily reviewable, I will open a follow-up PR (Part 2) that turns the original function into a backward-compatible wrapper that calls this new pure function. 

Screenshot 2026-03-10 230952

### Checklist
- [x] Follows PEcAn Contribution Guidelines
- [x] New tests pass `devtools::test()` locally without errors/warnings

Would love any quick feedback on the architecture of this pure function before I wire it up to the legacy wrapper!


modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -1,90 +1,89 @@
...
 #' @param settings PEcAn settings object
 #' @param ensemble.size number of runs in model ensemble
 #' @param posterior.files list of filenames to read from
 #' @param ens.sample.method one of "halton", "sobol", "torus", "lhc", "uniform"
-#' @export
+#' @param save_to_disk logical. If TRUE (default), saves samples.Rdata.
+#'
+#' @return named list with: trait.samples, sa.samples, ensemble.samples, runs.samples, env.samples
 #'
-#' @author David LeBauer, Shawn Serbin, Istem Fer
```
mdietze, line 10: don't delete the author list, just add yourself

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -1,90 +1,89 @@
...
-  trait.samples <- sa.samples <- ensemble.samples <- env.samples <- list()
-  runs.samples <- param.names <- list()
-
-  # flag determining whether samples are independent
-  # (e.g. when params fitted individually)
-  independent <- TRUE
+  ## ---- Load priors and MCMC from disk ----
+  prior_distns_list <- vector("list", length(pft.names))
+  trait_mcmc_list   <- vector("list", length(pft.names))
+  independent       <- TRUE
```
mdietze: not sure why useful comments were removed

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -1,90 +1,89 @@
...
 
-  ## Load PFT priors and posteriors
```
mdietze, line 74: Here and next three edits are all examples of active effort being put into removing useful comments. Throughout the entire PR please revert deleted comments, except in cases where code changes require a change in the comment.

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -93,182 +92,52 @@ get.parameter.samples <- function(settings,
...
         load(trait.mcmc.file, envir = distns)
-
-
-        # PDA samples are fitted together, to preserve correlations downstream
-        # let workflow know they should go together
+        trait_mcmc_list[[i]] <- distns$trait.mcmc
         if (grepl("mcmc.pda", trait.mcmc.file)) independent <- FALSE
-        # NOTE: Global MA samples will also be together, right?
-      } else {
-        PEcAn.logger::logger.info(
```
mdietze: no clear why this `else` and associated error message were deleted

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -93,182 +92,52 @@ get.parameter.samples <- function(settings,
...
+    env.samples      <- result$env.samples
+    save(ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples,
+         file = file.path(settings$outdir, "samples.Rdata"))
   }
 
-  save(ensemble.samples, trait.samples, sa.samples, runs.samples, env.samples,
-    file = file.path(settings$outdir, "samples.Rdata")
-  )
-} # get.parameter.samples
+  return(result)
```
mdietze: given that the previous function didn't return a result you might want to use `invisible` instead of `return`

modules/uncertainty/R/get_parameter_samples_pure.R
```diff
@@ -0,0 +1,272 @@
...
+#' str(result$ensemble.samples)
+#' }
+#'
+#' @seealso \code{\link{get.parameter.samples}} for the backward-compatible
+#'   wrapper that handles file I/O and database lookups.
+#'
+#' @importFrom purrr map
+#' @importFrom magrittr %>%
+#' @export
+get_parameter_samples_pure <- function(pft_names,
```
mdietze: could you drop the `_pure` from the name? Doing so will help future-proof the code.

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -1,90 +1,89 @@
...
+#' @export
 #' @importFrom rlang %||%
 get.parameter.samples <- function(settings,
                                   ensemble.size = 1,
                                   posterior.files = rep(NA, length(settings$pfts)),
-                                  ens.sample.method = "uniform") {
-  ### Identify PFTs in the input settings.xml file
+                                  ens.sample.method = "uniform",
+                                  save_to_disk = TRUE) {
+
```
mdietze, line 28: I'd encourage adding a deprecation message using the standard deprecation functions, following R guidelines.

mdietze (changes requested):

modules/uncertainty/R/get.parameter.samples.R
omkarrr2533: I apologies! I will restore the original authors and append my name.

modules/uncertainty/R/get.parameter.samples.R
omkarrr2533: Understood. The restructuring accidentally stripped out the legacy inline comments. I will revert those deletions and restore all original comments.

modules/uncertainty/R/get.parameter.samples.R
omkarrr2533: i will restore this missing block and the logger message immediately.

modules/uncertainty/R/get.parameter.samples.R
omkarrr2533: Great point to prevent console spam. I'll change this to invisible(result). 👍

modules/uncertainty/R/get_parameter_samples_pure.R
omkarrr2533: Makes perfect sense for future-proofing. I will rename it to get_parameter_samples() and update the filenames/tests accordingly.

modules/uncertainty/R/get.parameter.samples.R
omkarrr2533: I will add .Deprecated("get_parameter_samples") at the top of the legacy function. Thank you for pointing out the standard R guideline here.  😊

modules/uncertainty/R/get.parameter.samples.R
```diff
@@ -1,81 +1,95 @@
...
   # flag determining whether samples are independent
   # (e.g. when params fitted individually)
-  independent <- TRUE
+  independent       <- TRUE
 
   ## Load PFT priors and posteriors
   for (i in seq_along(pft.names)) {
     distns <- new.env()
 
-    ## Load posteriors
```
mdietze: revert

docker/depends/pecan_package_dependencies.csv
```diff
@@ -188,6 +189,7 @@
 "magrittr","*","modules/benchmark","Imports",FALSE
 "magrittr","*","modules/data.land","Imports",FALSE
 "magrittr","*","modules/data.remote","Imports",FALSE
+"magrittr","*","modules/uncertainty","Imports",FALSE
```
mdietze: drop

modules/uncertainty/R/get_parameter_samples.R
```diff
@@ -0,0 +1,273 @@
...
+#' str(result$trait.samples)
+#' str(result$ensemble.samples)
+#' }
+#'
+#' @seealso \code{\link{get.parameter.samples}} for the backward-compatible
+#'   wrapper that handles file I/O and database lookups.
+#'
+#' @author David LeBauer, Shawn Serbin, Istem Fer, Om Kapale
+#' @importFrom purrr map
+#' @importFrom magrittr %>%
```
mdietze: replace with base R's native pipe |>

modules/uncertainty/DESCRIPTION
```diff
@@ -31,6 +31,7 @@ BugReports: https://github.com/PecanProject/pecan/issues
 Imports:
     dplyr,
     ggplot2,
+    magrittr,
```
mdietze: drop

modules/uncertainty/NAMESPACE
```diff
@@ -35,5 +36,7 @@ export(spline.truncate)
 export(write.ensemble.configs)
 export(write.sa.configs)
 importFrom(dplyr,"%>%")
+importFrom(magrittr,"%>%")
```
mdietze: Drop. Also, even if you *did* need %>% over |> you can see from the line above that we're already importing %>% from dplyr.

modules/uncertainty/R/get_parameter_samples.R
```diff
@@ -0,0 +1,273 @@
...
+#'
+#' str(result$trait.samples)
+#' str(result$ensemble.samples)
+#' }
+#'
+#' @seealso \code{\link{get.parameter.samples}} for the backward-compatible
+#'   wrapper that handles file I/O and database lookups.
+#'
+#' @author David LeBauer, Shawn Serbin, Istem Fer, Om Kapale
+#' @importFrom purrr map
```
mdietze: Why is this importFrom needed? The only use of map below includes the namespace

modules/uncertainty/NAMESPACE
mdietze: Note that there's a lot of legacy PEcAn code that predates |> but we should switch to the base R pipe when cleaning up code.

modules/uncertainty/R/get_parameter_samples.R
omkarrr2533: Totally understood! you're absolutely right, since I'm already using purrr::map() with the full namespace throughout, R can resolve it just fine without the @importFrom directive. I added it out of habit but it's completely unnecessary here. will remove it, thanks for pointing that out! 👍

modules/uncertainty/R/get.parameter.samples.R
omkarrr2533: done

docker/depends/pecan_package_dependencies.csv
omkarrr2533: done 👍

modules/uncertainty/R/get_parameter_samples.R
omkarrr2533: replaced with R's native pipe

modules/uncertainty/DESCRIPTION
omkarrr2533: done

mdietze (approved):

PR #3861, opened 2026-03-10
Allow SIPNET to take command line arguments

ashiklom:
This is necessary to allow SIPNET to work with restart files (https://github.com/PecanProject/sipnet/pull/276/changes#diff-3516f6e935bd13d372f21f3daf1fa563724eaad958591e7ed67919b5f3101a06R27-R33), but it also enables access to other SIPNET features accessible via command line.

The interface is inspired by, and extends, the ED2 approach: `settings$model$binary_args`.

XML:

```xml
<model>
  <binary>sipnet.dev</binary>
  <binary_args>
    <quiet></quiet>  # flag; no parameters
    <restart-in>path/to/restart-in</restart-in>
</model>
```

R:

```r
list(
  binary = "sipnet.dev",
  binary_args = list(quiet = NULL, "restart-in" = "path/to/restart-in")
)
```

Both of these result in:

```sh
sipnet.dev --quiet --restart-in=path/to/restart-in
```

You can also pass in literal strings (like ED2 does) if you'd like, though that triggers a warning:

```xml
<binary_args>--quiet --restart-in=path/to/restart-in</binary_args>
```

As an aside, I also took the opportunity to update the `write.config.SIPNET` documentation, and to turn on markdown documentation for `modules/SIPNET`.


dlebauer:
PEcAn is trying to move away from having information in the settings object passed opaquely to functions.

Would it make sense to add these as arguments to the function, for example:

```r
write.configs.SIPNET(...., sipnet_args = list(a = settings$x$a, b = settings$x$b, ...))

```

Or even 

```r
write.configs.SIPNET(...., a = settings$x$a, b = settings$x$b, ...)

```

?

infotroph:
Thanks for working on this! I actually have a competing proposal that I didn't quite get to pushing this morning -- per https://github.com/PecanProject/pecan/pull/3813/files#r2891149490 I think we should standardize across models on `settings$model$options` and for Sipnet I propose writing into sipnet.in instead of passing command-line flags. 

If that direction makes sense to you I can fold my local draft into this one.

ashiklom:
@dlebauer Conceptually, I agree with you. Practically, I'm a little bit concerned about leaving the API inconsistent. If we pretty much always default to `settings`, at least we know that's the first place to look. If we start mixing `settings` with direct arguments, then I at least would be constantly second guessing myself, "Is this controlled from `settings` or from an argument?" My recommendation would probably be to leave everything in `settings` _for now_ and then do a big (AI-assisted) refactor later (perhaps, as part of a major PEcAn release) where we deprecate the `settings` argument almost everywhere all at once in favor of more specific arguments.

@infotroph Agreed that your proposal is the better approach! It's almost identical to the approach we took with `ED2IN` (see `models/ed`).

But, por que no los dos? It doesn't strike me as a terrible idea to allow command line arguments in the general case, even if in practice we avoid them. Maybe not specifically for SIPNET (because we fully control the model), but I could imagine other models that _only_ get certain options through command line flags. It might not be the worst idea to start supporting this more generally. I could also simplify my implementation to _only_ accept a literal string, which would more clearly label this as a footgun to be used only for specific expert use cases.

mdietze:
A few cents of input...

I agree with @ashiklom that the movement away from monolithic settings has to be balanced against standardization (i.e., every write.configs.MODEL has to have the same API/arguments for PEcAn to work) and the need to maintain a reasonable number of inputs to a function. I've in the past used run.write.configs as an example of a function that will probably have to continue to take the whole settings as an input because there's SO MUCH information that needs to be passed in. If we do simplify things, we'll still probably be passing in large lists like settings$run and settings$model, so it doesn't make sense to pull things out of those blocks.

In terms of passing command line args to models, there's no PEcAn argument for or against it as we have to support whatever a model supports. Here we're in the unique case of also controlling SIPNET, in which case I have a question -- what else, other than the restart, gets both written by PEcAn and read by SIPNET when it is restarting? Basically I'm arguing that I/O is time consuming so let's not put the path to one file (restart) inside another file (sipnet.in), which doubles the number of reads and writes, if we can just tell the model directly where the restart file is, either via naming convention (which is how sipnet currently works) or command line args.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -1,17 +1,53 @@
...
-##' @param settings PEcAn settings object
-##' @param run.id run ID
-##' @param inputs list of model inputs
-##' @param IC initial condition
-##' @param restart In case this is a continuation of an old simulation. restart needs to be a list with name tags of runid, inputs, new.params (parameters), new.state (initial condition), ensemble.id (ensemble id), start.time and stop.time.See Details.
-##' @param spinup currently unused, included for compatibility with other models
-##' @export
-##' @importFrom rlang %||%
-##' @author Michael Dietze
+#' Writes a configuration files for SIPNET model
```
```suggestion
#' Writes configuration files for one invocation of the SIPNET model
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -1,17 +1,53 @@
...
+#'
+#' - `*.clim` --- SIPNET meteorology driver (from
+#' `settings$run$inputs$met$path`, overriden by `inputs$met$path`). Note that
+#' this determines the ' SIPNET start and end dates. 
+#' - `job.sh` --- Job submission script. Populated from `inst/template.job`
+#' - `events.in` --- Copied from `inputs$events$path` or
+#' `settings$run$inputs$events$path`. This needs to be in the SIPNET event
+#' format; see [write.events.SIPNET()] for generating these files
+#' - `*.param` --- SIPNET parameter file. Includes both traits and initial
+#' conditions.
```
infotroph: Maybe on the picky side, but the clim linking has confused me a few times:
```suggestion
#' - `job.sh` --- Job submission script. Populated from `inst/template.job`
#' - `events.in` --- Copied from `inputs$events$path` or
#' `settings$run$inputs$events$path`. This needs to be in the SIPNET event
#' format; see [write.events.SIPNET()] for generating these files
#' - `*.param` --- SIPNET parameter file. Includes both traits and initial
#' conditions.
#' - `*.clim` --- SIPNET meteorology driver (from
#' `settings$run$inputs$met$path`, overriden by `inputs$met$path`). Note that
#' this determines the ' SIPNET start and end dates. Technically write.configs does
#' not create this file, but rather records its current path in `job.sh` to be used
#' to create a symlink to it at runtime.
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -1,17 +1,53 @@
...
+#'
+#' @details
+#' 
+#' # Command line arguments
```
infotroph, line 34:
```suggestion
#' # Model version specification
#'
#' `write.config.SIPNET()` matches its output format to the version of Sipnet
#' listed in `settings$model$revision`. This should be a numeric version
#' (e.g. `2.0.1`) and needs to match the version of your Sipnet binary.
#' You can check your binary's version by running
#' `./path/to/your/sipnet --version`, which should report something similar
#' to `SIPNET version 2.0.0 (4baf19a66c)`; if it says "illegal option" then you
#' have Sipnet 1.x and can report the version as "v1".
#'
#' # Command line arguments
```

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -1,17 +1,53 @@
...
+#'   "restart-out" = "path/to/restart.out",
+#'   "quiet" = NULL
+#' )
+#' ```
+#'
+#' ...will be rendered the following string in `job.sh`:
+#'
+#' ```
+#' "path/to/sipnet.dev" --restart-in=path/to/restart.in --restart-out=path/to/restart.out --quiet
+#' ```
```
````suggestion
#' SIPNET run-time options can be passed through a named list via
#' `settings$model$options`. For example, this...
#'
#' ```
#' settings$model$binary <- "path/to/sipnet.dev"
#' settings$model$options <- list(
#'   RESTART_IN = "path/to/restart.in",
#'   RESTART_OUT = "path/to/restart.out",
#'   ANAEROBIC = 1,
#'   GDD = 0
#' )
#' ```
#'
#' ...will be rendered as the following in `sipnet.in...:
#'
#' ```
#' RESTART_IN = path/to/restart.in
#' RESTART_OUT = path/to/restart.out
#' GDD = 0
#' ANAEROBIC = 1
#' ```
#'
#' ...though not necessarily in this order. If the `sipnet.in` template already
#' defines an option specified in settings$model$options, its value will be
#' updated in place; options not already in the file will be added to the bottom.
````

File: models/sipnet/R/write.configs.SIPNET.R [hunk 5]
```diff
@@ -1,17 +1,53 @@
...
+#'
+#' @param defaults nested list of named constant parameter values. The
+#' structure is `list(list(constants = list(trait1 = <value>, trait2 = <value>, ...)))`.
+#' Only `defaults[[1]]$constants` is used; all other elements are silently ignored. 
+#' @param trait.values vector of samples for a given trait
+#' @param settings PEcAn settings object
+#' @param run.id run ID
+#' @param inputs list of model inputs
+#' @param IC initial condition
+#' @param restart In case this is a continuation of an old simulation. restart needs to be a list with name tags of runid, inputs, new.params (parameters), new.state (initial condition), ensemble.id (ensemble id), start.time and stop.time.See Details.
```
infotroph, line 73: I assume this is intentional, but flagging that the details section doesn't yet say anything about this; presumably that will be handled in the restarts PR.

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -147,6 +183,8 @@ write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs
   
   jobsh <- gsub("@BINARY@", settings$model$binary, jobsh)
   jobsh <- gsub("@REVISION@", settings$model$revision, jobsh)
+
+  jobsh <- gsub("@BINARY_ARGS@", binary_args_to_string(settings$model$binary_args), jobsh)
```
infotroph: Stopping this round of review here -- I think I've proposed doc changes above to match a model$options design, and in the morning I'll push my local code changes to match.

infotroph (changes requested):

infotroph:
> It doesn't strike me as a terrible idea to allow command line arguments in the general case

> we have to support whatever a model supports. Here we're in the unique case of also controlling SIPNET

I agree the output format will be determined by the model. Sipnet's interesting here because we treat sipnet.in and command line as coequal and accept any flag via either interface. Standardizing on sipnet.in feels to me like it gives the easier provenance tracking on PEcAn's side, but passing them as command-line flags would be "as easy as" pasting names and values together in a different format.

Whatever form the model takes them in, I think I hear agreement that in the settings file, runtime options can be  key-value pairs in `settings$model$options`:
* The key is an option name _recognized by the model_, so if the model expects "long.WEIRD.option.NaMe", the settings contains that and not an abbreviation or standardization. Rationale: Enforces that this section is for _model_  options; we have the whole rest of the settings file for values PEcAn will slice and dice before emitting a run configuration.
* The value can be arbitrarily complex but write.config.MODEL needs to be able to format it to pass on to the model _knowing only the option name and value_. So write.config.MODEL knows whether to convert `<some_opt>TRUE</some_opt>' into `--some_opt=yes` or '{"some_opt": "True"}` or whatever, but it doesn't need to know anything about the effect of setting `some_opt` or even whether it's accepted as valid by this model version. Rationale: Avoids cluttering write.config with explicit handling of specific options / options combinations, and means adding new options to the model doesn't automatically require editing write.config.MODEL to support them.

Of course some options might change input or output requirements or mean parameters need to change, in which case write.config.MODEL would need to process them individually to change the PEcAn run setup accordingly.

infotroph:
> let's not put the path to one file (restart) inside another file (sipnet.in), which doubles the number of reads and writes, if we can just tell the model directly where the restart file is

@mdietze that's probably a point for the discussion in https://github.com/PecanProject/sipnet/pull/276 that can be separate from the design of how we pass options, but briefly I believe the current proposal is for Sipnet's restart/checkpoint files to contain state variables but _not_ the full parameter structure, so drivers and parameters would continue to be set by the normal model startup process and then have initial state values overridden(?) by loading the restart file. I'll defer to the other thread on what tradeoffs are in play for that decision and how firmly settled that is.

ashiklom:
@infotroph Agreed on all points above. One possible minor hitch is data type -- especially the difference between numbers and strings.

The most coherent interface would probably treat everything as strings and paste them into config files exactly as they are provided.  That would require options expected to be quoted strings to be double quoted (e.g., if a model expects an option to have the value `"string"` then R would have to store it as `'"string"'`). This is straightforward in XML, where I think `<tag>value</tag>` will be read as `"tag"` and `<tag>"value"</tag>` will be read as `"\"value\""`.

For SIPNET, it seems that we pass strings unquoted, so this is not an issue.

To make this extremely clear, I might go so far as to throw a loud early error for any model options that are not passed as strings, to prevent bugs related to R's implicit rules about string formatting of numbers.

ashiklom:
Actually, thinking through this a bit more, I don't think we can make the `options` interface fully generic to different models. For example, ED2 accepts some options through its namelist (ED2IN) and a *different* set of options through its command line arguments. Other models may have more complicated nested configuration structures, possibly across multiple files, possibly mixing command line arguments and config files.

The meaning of `options` therefore becomes ambiguous.

My revised suggestion would therefore be to keep this interface **explicitly model-specific**. In ED2, we have an `ed2in_tags` field that takes key-value pairs. For SIPNET, we should call this `sipnet_in_tags` or something to clearly telegraph the destination of those options.

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -323,8 +324,14 @@ Some important tags are as follows:
...
 - `type` -- The model "type", matching the PEcAn database `modeltypes` table `name` column. This also refers to which PEcAn model-specific package will be used. In PEcAn, a "model" refers to a specific version (e.g. release, git commit) of a specific model, and "model type" is used to link different releases of the same model. Model "types" also have specific PFT definitions and other requirements associated with them (e.g. the ED2 model "type" requires a global land cover database).
-- `binary` -- The file path to the model executable. If omitted, PEcAn will use whatever path is registered in the PEcAn database for the current machine. 
+- `binary` -- The file path to the model executable. If omitted, PEcAn will use whatever path is registered in the PEcAn database for the current machine.
+- `revision` -- A unique identifier for the model version of `binary`. Format is model-specific, but many models expect a numeric version (eg "2.12.3") and rely on numeric comparisons to match PEcAn configuration behavior to model capabilities.
```
infotroph:
```suggestion
- `revision` -- A unique identifier for the model version of `binary`. Format is model-specific, but many models expect a semantic version (eg "2.12.3") and rely on numeric comparisons to match PEcAn configuration behavior to model capabilities.
```

book_source/03_topical_pages/03_pecan_xml.Rmd
```diff
@@ -323,8 +324,14 @@ Some important tags are as follows:
...
 - `type` -- The model "type", matching the PEcAn database `modeltypes` table `name` column. This also refers to which PEcAn model-specific package will be used. In PEcAn, a "model" refers to a specific version (e.g. release, git commit) of a specific model, and "model type" is used to link different releases of the same model. Model "types" also have specific PFT definitions and other requirements associated with them (e.g. the ED2 model "type" requires a global land cover database).
-- `binary` -- The file path to the model executable. If omitted, PEcAn will use whatever path is registered in the PEcAn database for the current machine. 
+- `binary` -- The file path to the model executable. If omitted, PEcAn will use whatever path is registered in the PEcAn database for the current machine.
+- `revision` -- A unique identifier for the model version of `binary`. Format is model-specific, but many models expect a semantic version (eg "2.12.3") and rely on numeric comparisons to match PEcAn configuration behavior to model capabilities.
```
infotroph:
```suggestion
- `revision` -- A unique identifier for the model version of `binary`. Format is model-specific, but many models expect a semantic version (eg "2.12.3") and rely on version comparisons (e.g. `revision >= "2.0"`) to match PEcAn configuration behavior to model capabilities.
```

models/sipnet/R/write.configs.SIPNET.R
ashiklom: This was unchanged from earlier code (including the docs -- the line diff is just because I changed the `##` to `#`). So not my doing. Suggest revisiting this specific option in a different PR.

ashiklom (commented):
@infotroph GitHub won't let me officially approve my own PR, but I think this looks good! Thanks for the thorough docs and tests.

I'm not 100% sold on the model-agnostic `model$options` design, but I'm not firmly against it either, and the blast radius of this change is currently small and easy to pivot if we change our minds later. In any case, this meets my immediate need to configure SIPNET restarts, so I'm satisfied!

infotroph (approved):

PR #3865, opened 2026-03-11
fix: Rmd links in demo table open in top window to escape iframe

### LINKED ISSUE #60: Broken links on tutorials page
From the develop documentation page: https://pecanproject.github.io/documentation/develop/
- Click "PEcAn Demos" 
  - note that all of the links to "Rmd" are broken
  - The Rmd files on the underlying page (inside the frame) work: https://pecanproject.github.io/pecan-documentation/develop/demo-table.html

What works:
- although only two html links are present, they work
- if you right click the 'Rmd' hyperlink and 'copy link' you get the correct link, e.g. https://github.com/PecanProject/pecan/tree/develop/documentation/tutorials/StateAssimilation, they just don't show up inside the website frame.

Possible solutions:
1. Fix the redirect
2. Figure out if the issue exists on other pages & fix
3. Another workaround would be to read the raw file content and display inside the docs thus:
  - currently the link directs to the tutorials folder https://github.com/PecanProject/pecan/tree/develop/documentation/tutorials/StateAssimilation
  - it might be better to open to the file with formatting, but I don't know why this would work if the above doesn't https://github.com/PecanProject/pecan/blob/develop/documentation/tutorials/StateAssimilation/TreeRingSDA.Rmd
  - if it works to render the raw file contents, ideally with appropriate formatting, replace github.com with raw.githubusercontent.com in the file URL, thus: https://raw.githubusercontent.com/PecanProject/pecan/develop/documentation/tutorials/StateAssimilation/TreeRingSDA.Rmd

man080107:
GitHub pages send X-Frame-Options: deny which prevents them from loading inside the bookdown iframe on the documentation site. Clicking an Rmd link would show a blank frame even though the URL was correct.

Fix: change target='_blank' to target='_top' on all Rmd/qmd source links in the tutorials table. This navigates the top-level browser window to the GitHub page, breaking out of the iframe entirely.

Also fix Meteorological Data vignette link which pointed to a directory listing (/tree/) instead of a specific file (/blob/).

Fixes PecanProject/PecanProject.github.io#60

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


man080107:
@dlebauer @infotroph @mdietze ,  Could you please review this PR when you get a chance?

This fixes PecanProject/PecanProject.github.io#60 — the broken Rmd links in
the demo table. The root cause was `target="_blank"` being blocked inside a
Bookdown iframe due to GitHub's `X-Frame-Options: deny` header. Fixed by
changing to `target="_top"` so links break out of the iframe properly.

Also fixed one link that pointed to a folder instead of a specific file.

Happy to make any changes based on your feedback. Thanks!

infotroph:
> Clicking an Rmd link would show a blank frame even though the URL was correct.

I can't reproduce this on my machine (MacOS with Safari 26.1 and Firefox 148.0.2) with the current `target = "_blank"` configuration -- both browsers open it in a new tab (containing the expected content, not blank and not inside a frame) on regular click and in my choice of a new tab or new window on right/cmd click. 

In quick testing on a local version I can't find a case where `target="_top"` differs from not specifying a target at all; those both load the link in the current tab (at the top level, not framed) on regular click and in my choice of a new tab or new window on right/cmd click -- if that's the behavior we want, are there reasons to do it via _top rather than as a plain old targetless anchor? I'd prefer the ability to open in the current tab instead of always forcing a new one, but am willing to defer if other people strongly favor the `target = "_blank" behavior.

infotroph (approved):

PR #3884, opened 2026-03-19
sipnet: support v2 clim in split_inputs

infotroph:
Was using hard-coded column numbers. Now using... hard-coded column numbers chosen conditional on the total number of columns. Hey, it works.🙈


ashiklom (approved):
Looks good!

models/sipnet/tests/testthat/test-split_inputs.SIPNET.R
```diff
@@ -0,0 +1,69 @@
...
+  # Lines are numerically identical
+  # NB raw text does differ (split_inputs changes some `0.000` to `0`, etc),
+  # but should parse equal when read as numeric.
+  expect_equal(
+    read.table(climfile, nrows = 5),
+    read.table(clim_split[1], nrows = 5)
+  )
+  expect_equal(
+    read.table(text = system2("tail", c("-n5", climfile), stdout = TRUE)),
+    read.table(text = system2("tail", c("-n5", clim_split[4]), stdout = TRUE))
```
ashiklom: This will fail on Windows in the absence of a `tail` command, right? Here and below, I would just use R's string manipulation functions -- e.g., `tail(read lines(...), 5)`. They are clearer to R users and will work anywhere.

models/sipnet/tests/testthat/test-split_inputs.SIPNET.R
```diff
@@ -0,0 +1,69 @@
...
+  )
+
+})
+
+test_that("v2 clim format", {
+  outdir <- withr::local_tempdir()
+  climfile <- file.path(outdir, "niwot_v2.clim")
+  v1_clim <- system.file("niwot.clim", package = "PEcAn.SIPNET")
+  system2("cut", c("-d' '", "-f2-13", v1_clim, ">", climfile))
```
ashiklom: I'm actually mildlysurprised this works, since `>` is a shell-level redirect rather than an argument to `cut`. I would have expected this to require something like `shell = TRUE` argument to `system2`.

As above, I would gently suggest replacing this with built in R functionality, for portability.

models/sipnet/tests/testthat/test-split_inputs.SIPNET.R
```diff
@@ -0,0 +1,67 @@
+
+count_file_lines <- function(path) {
+  system2("wc", c("-l", path), stdout = TRUE) |>
```
ashiklom: Why not just `length(read lines(path))`.

I know you're technically reading the whole file that way, which is less efficient, but these are tiny test files where it hardly matters.

models/sipnet/tests/testthat/test-split_inputs.SIPNET.R
infotroph: It will, though my understanding is that Sipnet isn't known to work on Windows anyway. But the potentially nastier failure case is that I noted four lines up how the number of decimal digits in the head of the file could differ, then promptly assumed here that the tails will be text-identical 😒.

models/sipnet/tests/testthat/test-split_inputs.SIPNET.R
infotroph: The redirection part works, the cut command doesn't because this file contains a mix of space and tab separators. Replaced this by committing the 12-column `tests/testthat/data/niwot_1999_v2.clim`

models/sipnet/tests/testthat/test-split_inputs.SIPNET.R
infotroph: You know I had to do it 😉 

Screenshot 2026-03-19 at 4 53 37 PM


But you're right. I switched to length(readLines) and simplified the tests by just reading all the files once and comparing them as a dataframe rather than mess with separate head/tail/count calls.

ashiklom (approved):
👍

PR #3888, opened 2026-03-19
Add unit tests for benchmarking metrics and document metric_R2 fallback

tanmaydimriGSOC:
## Summary

While exploring the benchmarking module for my GSoC 2026 proposal (validation framework project), I noticed that the core metric functions (metric_RMSE, metric_MAE, metric_cor, metric_R2) had no unit test coverage. This PR adds tests for all four functions and documents a previously undocumented behavior discovered while writing them.

## Changes

**tests/testthat/test-metrics.R** (new file)
- 9 unit tests covering metric_RMSE, metric_MAE, metric_cor, and metric_R2
- Tests cover: perfect predictions, known values, NA handling, and edge cases
- Includes a test that explicitly captures the warning triggered by metric_R2's lm() fallback path

**R/metric_R2.R** (modified)
- Added @details section documenting the silent fallback behavior:
  when model output is constant, the correlation formula returns 
  NA and the function switches to lm()-based R-squared without notifying the user
- Added inline comment on the fallback block for future developers
- This behavior was discovered through the tests and is relevant 
  for multi-site validation use cases where silent inconsistency across sites could produce misleading results

## Notes

The lm() fallback in metric_R2 triggers an "essentially perfect fit: summary may be unreliable" warning from stats::summary.lm 
in edge cases. This is now documented but not changed — a follow-up could add an explicit warning or input check for 
constant model output.

Tests were verified locally by sourcing metric functions directly due to dependency installation constraints in local environment.

### DIFF
--- a/modules/benchmark/R/metric_R2.R
+++ b/modules/benchmark/R/metric_R2.R
@@ -1,11 +1,23 @@
 ##' @name metric_R2
 ##' @title Coefficient of Determination (R2)
 ##' @export
-##' @param metric_dat dataframe
+##' @param metric_dat dataframe with columns \code{model} and \code{obvs}
 ##' @param ... ignored
-##' 
+##'
+##' @details
+##' Computes R-squared using the correlation-based formula:
+##' \eqn{R^2 = \left(\frac{\sum(obs - \bar{obs})(mod - \bar{mod})}
+##' {\sqrt{\sum(obs - \bar{obs})^2} \cdot \sqrt{\sum(mod - \bar{mod})^2}}\right)^2}
+##'
+##' If this formula returns \code{NA} (e.g. when model output is constant
+##' across all observations), the function silently falls back to an
+##' \code{lm()}-based R-squared via \code{summary(lm())$r.squared}.
+##' This fallback may produce unreliable results and triggers a warning
+##' from \code{stats::summary.lm}: "essentially perfect fit: summary may
+##' be unreliable". Consider checking for constant model output before
+##' calling this function.
+##'
 ##' @author Betsy Cowdery
-
 metric_R2 <- function(metric_dat, ...) {
   PEcAn.logger::logger.info("Metric: Coefficient of Determination (R2)")
   numer <- sum((metric_dat$obvs - mean(metric_dat$obvs)) * (metric_dat$model - mean(metric_dat$model)))
@@ -13,6 +25,10 @@ metric_R2 <- function(metric_dat, ...) {
   
   out <- (numer / denom) ^ 2
   
+  # If correlation formula returns NA (e.g. constant model output),
+  # fall back to lm()-based R-squared. Note: this fallback may trigger
+  # "essentially perfect fit" warning from stats::summary.lm and
+  # produce unreliable results in edge cases.
   if(is.na(out)){
     fit <- stats::lm(metric_dat$model ~ metric_dat$obvs)
     out <- summary(fit)$r.squared
--- a/modules/benchmark/tests/testthat/test-metrics.R
+++ b/modules/benchmark/tests/testthat/test-metrics.R
@@ -1,0 +1,49 @@
+# at the top of test-metrics.R, add this:
+if (!requireNamespace("PEcAn.logger", quietly = TRUE)) {
+  PEcAn.logger <- new.env()
+  PEcAn.logger$logger.info <- function(...) invisible(NULL)
+}
+
+test_that("metric_RMSE returns 0 for perfect predictions", {
+  dat <- data.frame(model = c(1, 2, 3), obvs = c(1, 2, 3))
+  expect_equal(metric_RMSE(dat), 0)
+})
+
+test_that("metric_RMSE handles NA values", {
+  dat <- data.frame(model = c(1, NA, 3), obvs = c(1, 2, 3))
+  expect_true(is.numeric(metric_RMSE(dat)))
+})
+
+test_that("metric_RMSE returns numeric", {
+  dat <- data.frame(model = c(2, 4), obvs = c(1, 3))
+  expect_equal(metric_RMSE(dat), 1)
+})
+
+test_that("metric_MAE returns 0 for perfect predictions", {
+  dat <- data.frame(model = c(1, 2, 3), obvs = c(1, 2, 3))
+  expect_equal(metric_MAE(dat), 0)
+})
+
+test_that("metric_MAE returns correct value", {
+  dat <- data.frame(model = c(3, 3), obvs = c(1, 1))
+  expect_equal(metric_MAE(dat), 2)
+})
+
+test_that("metric_cor returns 1 for perfect linear relationship", {
+  dat <- data.frame(model = c(1, 2, 3), obvs = c(1, 2, 3))
+  expect_equal(metric_cor(dat), 1)
+})
+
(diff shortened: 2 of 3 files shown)


ayushman1210:
@dlebauer  I've reviewed this PR in the context of my GSoC workplan. This decoupled, functional approach to calculating metrics (specifically metric_R2.R) fits perfectly with the architecture we are building for Phase 2
I suggest we merge this PR as-is.
Once  it merged I will use this exact pattern as the template to generalize the remaining metrics (RMSE, MAE, correlation) so they can operate independently of the legacy BETYdb infrastructure. Let's get this merged!

modules/benchmark/tests/testthat/test-metrics.R
```diff
@@ -0,0 +1,49 @@
+# at the top of test-metrics.R, add this:
+if (!requireNamespace("PEcAn.logger", quietly = TRUE)) {
+  PEcAn.logger <- new.env()
+  PEcAn.logger$logger.info <- function(...) invisible(NULL)
```
infotroph, line 4: Can you explain what's happening here and what it's trying to accomplish? It looks like you may be trying to mask the logger functions, which I do not think will work as expected (they live in their own namespace) and also isn't generally a good idea.

ayushman1210: @infotroph  I focused my review on the metric functions and the R-squared fallback documentation (which align perfectly with the Phase 2 goals of my GSoC workplan) and I admittedly glossed over that mocking workaround at the top of the test file.
Looking at the PR description, it seems the author added that block to bypass a local environment issue where they couldn't install PEcAn.logger. You are absolutely right that we shouldn't be trying to mask the namespace like this, especially since the package will be available in the standard CI/testing environment anyway.

ayushman1210: @tanmaydimriGSOC could you go ahead and remove lines 1-5 in test-metrics.R? The GitHub  CI will have the necessary dependencies installed  so the tests will run fine without that workaround. Once that is removed the core logic of this PR still looks good to me!!

modules/benchmark/R/metric_R2.R
```diff
@@ -1,23 +1,39 @@
...
+##'
+##' @details
+##' Computes R-squared using the correlation-based formula:
+##' \eqn{R^2 = \left(\frac{\sum(obs - \bar{obs})(mod - \bar{mod})}
+##' {\sqrt{\sum(obs - \bar{obs})^2} \cdot \sqrt{\sum(mod - \bar{mod})^2}}\right)^2}
+##'
+##' If this formula returns \code{NA} (e.g. when model output is constant
+##' across all observations), the function silently falls back to an
+##' \code{lm()}-based R-squared via \code{summary(lm())$r.squared}.
+##' This fallback may produce unreliable results and triggers a warning
```
divine7022, line 15: here "silently falls back" and next sentence says it triggers warning. can't both be true
drop "silently"; rest of the paragraph correctly describes the warning behavior

modules/benchmark/R/metric_R2.R
```diff
@@ -1,23 +1,39 @@
 ##' @name metric_R2
 ##' @title Coefficient of Determination (R2)
```
divine7022, line 2: title still says "Coefficient of Determination"  but formula in @details squared pearson correlation; formula in @details is correct.
change title

modules/benchmark/tests/testthat/test-metrics.R
```diff
@@ -0,0 +1,49 @@
...
+})
+
+test_that("metric_R2 silent NA fallback produces valid output", {
+  dat <- data.frame(model = c(2, 2, 2), obvs = c(1, 2, 3))
+  expect_warning(
+    result <- metric_R2(dat),
+    "essentially perfect fit"
+  )
+  expect_true(is.numeric(result))
+})
```
divine7022, line 49: test name says "silent NA fallback" but the input has no NAs, constant model triggers fallback via 0/0 returning NaN. assertion `is.numeric(result)` also passes on `NaN`, which is what the function actually returns here. worth tightening both

modules/benchmark/R/metric_R2.R
```diff
@@ -1,23 +1,39 @@
...
+##' \code{lm()}-based R-squared via \code{summary(lm())$r.squared}.
+##' This fallback may produce unreliable results and triggers a warning
+##' from \code{stats::summary.lm}: "essentially perfect fit: summary may
+##' be unreliable". Consider checking for constant model output before
+##' calling this function.
+##'
 ##' @author Betsy Cowdery
-
 metric_R2 <- function(metric_dat, ...) {
   PEcAn.logger::logger.info("Metric: Coefficient of Determination (R2)")
```
divine7022, line 22: also comment here -- https://github.com/PecanProject/pecan/pull/3888/changes#r3397678145
```suggestion
  PEcAn.logger::logger.info("Metric: Squared Pearson Correlation (R2)")
```

modules/benchmark/tests/testthat/test-metrics.R
```diff
@@ -0,0 +1,49 @@
...
+
+test_that("metric_RMSE returns 0 for perfect predictions", {
+  dat <- data.frame(model = c(1, 2, 3), obvs = c(1, 2, 3))
+  expect_equal(metric_RMSE(dat), 0)
+})
+
+test_that("metric_RMSE handles NA values", {
+  dat <- data.frame(model = c(1, NA, 3), obvs = c(1, 2, 3))
+  expect_true(is.numeric(metric_RMSE(dat)))
+})
```
divine7022, line 15: `is.numeric()` passes on `NA`, `NaN`, `Inf`, and `0` indifferently, this test doesn't actually constrain the return value. since metric_RMSE uses na.rm = TRUE, expected value here is 0. worth checking that directly


```suggestion
test_that("metric_RMSE handles NA values", {
  dat <- data.frame(model = c(1, NA, 3), obvs = c(1, 2, 3))
  expect_equal(metric_RMSE(dat), 0)
})
```

modules/benchmark/tests/testthat/test-metrics.R
divine7022, line 1: NA in model triggers lm() fallback, which by default just drops NA row via `na.action = na.omit` and fits perfect line through whatever's left. so single NA in model series silently returns R2 = 1. no test pins this. wondering if the ryt move here is to handle NAs same way metric_cor and metric_PPMC do, squaring that gives same r2 and lm fallback can go away entirely, since cor() already returns NA for degenerate constant input case; so constant input case returns NA, instead of NaN

something like:

```
metric_R2 <- function(metric_dat, ...) {
  PEcAn.logger::logger.info("Metric: Squared Pearson Correlation (R²)")
  stats::cor(metric_dat$model, metric_dat$obvs, use = "pairwise.complete.obs")^2
}
```
heads up: if this lands, the last test needs to update too , constant model case would return NA (with "standard deviation is zero" warning) instead of NaN (with "essentially perfect fit" warning)

divine7022 (changes requested):
solid first pass on tests; just needs few cleanups

divine7022:
filed follow up to track this - https://github.com/PecanProject/pecan/issues/4027

dlebauer:
Thank you for your contributions @tanmaydimriGSOC 

I'm closing this one and your commits here have been integrated into https://github.com/PecanProject/pecan/pull/4041

PR #3892, opened 2026-03-24
feat(benchmark): add run_benchmark() with dataframe-first pipeline, metrics, and tests

anshul23102:
## Description
Adds a new `run_benchmark()` function as a database-free entry point for the 
benchmark module. Unlike `calc_benchmark()`, this works without a BETYdb 
connection, taking validated dataframes directly as input.

Files added:
- `R/run_benchmark.R` — dataframe-first pipeline with `bm_validate()`, 
  `align_by_time()`, `compute_metrics()`, and `plot_time_series()`
- `inst/testdata/sample_model.csv` — sample model output
- `inst/testdata/sample_obs.csv` — sample observations
- `tests/testthat/test-run_benchmark.R` — unit tests for all four pipeline stages
- `README.md` — updated with quickstart example

## Motivation and Context
The existing `calc_benchmark()` requires a full database connection which 
makes it hard to test and use standalone. This adds a lightweight, 
dataframe-first entry point — `run_benchmark(model_df, obs_df)` — for users 
who want to quickly benchmark model output against observations without any 
database setup. The pipeline follows a four-stage design: validate → align → 
compute metrics → plot.

## Review Time Estimate
- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:
- [x] I have read the CONTRIBUTING document.


anshul23102:
Hi @dlebauer, just wanted to give you a quick update on PR #3892. All CI checks are now passing and I've refactored the API to be dataframe-first (`run_benchmark(model_df, obs_df)`) which directly aligns with the GSoC proposal I submitted. 
The pipeline now has four clean stages: validate, align, compute metrics, and plot, each usable independently or through the orchestrator. Would love your feedback whenever you get a chance!

modules/benchmark/R/run_benchmark.R
```diff
@@ -0,0 +1,99 @@
...
+
+##' Align model and observation data frames by nearest time
+##'
+##' @param model_df data.frame with columns: time (POSIXct), value
+##' @param obs_df   data.frame with columns: time (POSIXct), value
+##' @param tolerance_secs max allowed time difference in seconds
+##'
+##' @return data.frame with columns: time, model, obs
+align_by_time <- function(model_df, obs_df, tolerance_secs = 3600) {
+  aligned <- do.call(rbind, lapply(seq_len(nrow(model_df)), function(i) {
```
ayushman1210: this lapply loop computes difftime against the entire dataframe for every row, creating a slow $O(N \times M)$ bottleneck. It should be replaced with a data.table rolling join (obs_dt[model_dt, roll = "nearest"])

modules/benchmark/R/run_benchmark.R
```diff
@@ -0,0 +1,99 @@
...
+##' @return list with: metrics (data.frame), aligned (data.frame), plot (ggplot)
+##' @export
+##' @author Anshul Jain
+run_benchmark <- function(model_df, obs_df,
+                          metrics = c("RMSE", "MAE"),
+                          tolerance_secs = 3600,
+                          method = "nearest") {
+
+  # Stage 1: Validate schema
+  bm_validate(model_df, obs_df)
```
ayushman1210, line 21: The function signature accepts method  but  when calling align_by_time(), the method argument is completely ignored and not passed down

modules/benchmark/R/run_benchmark.R
```diff
@@ -0,0 +1,99 @@
...
+  data.frame(metric = toupper(metrics), value = unlist(results, use.names = FALSE))
+}
+
+##' Plot model vs observations time series
+##'
+##' @param aligned data.frame with columns: time, model, obs
+##' @return ggplot object
+plot_time_series <- function(aligned) {
+  ggplot2::ggplot(aligned, ggplot2::aes(x = .data$time)) +
+    ggplot2::geom_line(ggplot2::aes(y = .data$model, color = "Model")) +
```
ayushman1210, line 119: Using .data here will trigger an R CMD check warning. They need to add the rlang import tag to the roxygen block right above this function (e.g., adding ##' @importFrom rlang .data )

modules/benchmark/NAMESPACE
```diff
@@ -35,6 +35,7 @@ export(metric_run)
 export(metric_scatter_plot)
 export(metric_timeseries_plot)
 export(read_settings_BRR)
+export(run_benchmark)
```
ayushman1210, line 38: NAMESPACE is auto-generated by roxygen2 and shouldn't be edited directly. They should just revert the NAMESPACE changes and let roxygen handle it.

ayushman1210 (commented):
@anshul23102  Great work on this PR! The database-free pipeline structure is extremely clean and sets up exactly what we need for the new validation toolkit.
I've left a few inline code suggestions—the main one is regarding the align_by_time function. The current loop creates an O(N*M) bottleneck, so I've provided a drop-in data.table suggestion that will speed this up significantly to O(N log N). I also left a couple of minor notes regarding an unused argument and some package checks.  

Let me know once you've had a chance to update those and I'd be happy to take another look!

divine7022:
@ayushman1210 since all commits here are incorporated in #4017
and seems like I head you'r saying, #4017 PR suppressed #3892 

what's the plan here ? can we consider this safe to close ?

anshul23102:
Thanks for the detailed review @ayushman1210. Applied all four points in the latest push:

**1. align_by_time performance (O(N log N)):** Replaced the `lapply`/`difftime` loop with a `findInterval`-based binary-search join. For each model timestep, `findInterval` locates the bracketing observation in O(log M) time, then a single comparison picks the nearer neighbour. Total cost is O((N+M) log M) with no new package dependency.

**2. method argument not forwarded:** `run_benchmark()` now passes `method` down to `align_by_time()`. The argument is accepted for forward compatibility; "nearest" is the only implemented strategy.

**3. .data import tag:** Added `##' @importFrom rlang .data` to the `plot_time_series()` Roxygen block so the import is declared properly and R CMD check does not flag it.

**4. NAMESPACE:** The file was previously missing the Roxygen-generated `export(run_benchmark)` line, which was then added by hand. The `@export` tag was already present in the source, so `devtools::document()` would produce the same result. The NAMESPACE now reflects what Roxygen would generate from the current source declarations.

anshul23102:
@divine7022 Good question. This PR and #4017 serve different roles:

- **#3892** is a self-contained, reviewable addition of `run_benchmark()` with a dataframe-first API, `align_by_time()` (O(N log N) via `findInterval`), `compute_metrics()`, and `plot_time_series()`. All feedback from @ayushman1210 has been applied. It can stand alone as a merged base.
- **#4017** is @ayushman1210's GSoC follow-up that modularises the pipeline further (4-stage Validate/Align/Compute/Plot) and extends the work. Its PR description explicitly calls #3892 out as the foundation it builds on.

The standard path in open source is: merge the smaller, already-reviewed base first, then #4017 layers on top with a clean history. That avoids #4017 carrying unreviewed code from #3892 inside it.

If the maintainers prefer to close #3892 and let #4017 absorb everything, I am fine with that too. Happy to follow whatever direction @infotroph or @mdietze prefer.

divine7022:
thanks! for writeup @anshul23102, and nice work on putting this together
per our discussion on last team call, wanted to push back gently on "merge base first" framing though. that pattern works when the base is genuinely independent of the follow up, but here #4017 already contains every commit from this PR plus the phase 2 additions on top. finished full pass on #4017 and the code you wrote here is exactly what got reviewed there. since #4017 is GSoC umbrella PR, my read is that consolidating there is cleaner. your commits stay attributed in the #4017 history, you're still the author of those changes, nothing gets lost

I am closing this here and follow up in #4017

PR #3893, opened 2026-03-24
update sobol GSA to use sensobol package and validate ensemble design handling

divine7022:
## Description

This PR replaces the existing Sobol workflow with a sensobol-based implementation and hardens the ensemble design path used to run global sensitivity analyses

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -38,85 +104,129 @@
...
-    input_tag <- names(samp.ordered)[i]
-    parent_name <- samp.ordered[[i]]$parent
-
-    parent_ids <- if (!is.null(parent_name)) {
-      sampled_inputs[[parent_name]]
-    } else {
-      NULL
+
+  if (sobol) {
+    # every input in samplingspace (except parameters) is an independent factor
```
mdietze: you need to check the parent/child relationships to determine if this is true or not. Code related to parent, parent_id, parent_name, etc should not have been deleted

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -38,85 +104,129 @@
...
+    input_names <- setdiff(names(samp), "parameters")
+    sobol_factors <- c("param", input_names)
+
+    total_runs <- as.integer(ensemble_size) * (length(sobol_factors) + 2L)
+    samples_file <- file.path(settings$outdir, "samples.Rdata")
+    if (.sobol_parameter_bank_size(samples_file) < total_runs) {
```
mdietze: 1. I'm not following why the constraint here is based on total_runs not 2*ensemble_size. You need A and B parameter samples, you don't need additional independent parameter samples for the k permutations
2. This `if` doesn't have an `else`, but it's not clear to me that the code will keep running if the `if` is skipped and no parameters are loaded
3. Pretty sure the existing code supported the option to re-use a previous set of parameter samples, which also seems to have gone away

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -1,4 +1,64 @@
...
+  load(samples_file, envir = samples)
+
+  if (is.null(samples$trait.samples) || length(samples$trait.samples) == 0) {
+    return(0L)
+  }
+
+  trait_sample_bank_size(samples$trait.samples)
+}
+
+.map_sobol_to_indices <- function(x, size) {
```
mdietze, line 55: I'm not sure why this hack in nessisary, and what implications it has during the calculations of the sobol indices, would be nice to have a demonstration that this logic generates the right frequency distribution and doesn't over or under sample the two end members

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -38,85 +104,129 @@
...
+    if (.sobol_parameter_bank_size(samples_file) < total_runs) {
+      PEcAn.uncertainty::get.parameter.samples(
+        settings = settings,
+        ensemble.size = total_runs,
+        posterior.files = posterior.files,
+        ens.sample.method = ens.sample.method
+      )
+    }
+
+    sobol_design <- sensobol::sobol_matrices(
```
mdietze, line 169: Its unclear to me what specifically is "better" about the proposed approach over the existing code, which was a LOT shorter and simpler to understand/maintain. The proposed approach also results in qualitatively different sampling approaches between the sobol and non-sobol designs, and results in all the specifications/logic in the `<ensemble>` block being ignored and the basic sampling code being unnecessarily duplicated.

mdietze (changes requested):

modules/uncertainty/R/generate_joint_ensemble_design.R
divine7022: fixed in https://github.com/PecanProject/pecan/pull/3893/commits/849160cb2f7c01f31f961e0df5efd51f5619fc3e
current logic now distinguishes independent sobol factors from child inputs; only parentless inputs become independent sobol factors, they get quasi-random columns from sobol_matrices(). Inputs that do declare a parent are routed through input.ens.gen() with parent_ids, exactly as non sobol path does

modules/uncertainty/R/generate_joint_ensemble_design.R
divine7022: right on number of MCMC parameter draw we need here; and we don't have to re compute number of total runs, it could be detected by number of rows in sobol_desing.
sensobol::sobol_matrices() constructs quasi-random base matrix Q of dimension (N, 2k), then splits it into A (left k cols) and B (right k cols). cross-matrices AB_i copy all columns from A except column i, which comes from B. So any given parameter column only ever takes values from its A or B column,  which means at most 2N unique [0,1) values per factor, never N(k+2) (Puy et al. 2022, JSS 102(5), Eq. 10; Saltelli et al. 2010, Table 2)
https://cran.r-project.org/web/packages/sensobol/vignettes/sensobol.pdf --  (page no.6 and 9)

on the reuse question, yes existing parameter samples are reused. `.sobol_parameter_bank_size()` loads samples.Rdata, reads trait.samples, and returns the minimum bank size across all pfts/traits.
and it is also do more then just pre existed `if (!file.exists(...))` guard, it also checks whether the existing bank is large enough rather than just whether it exists

https://github.com/PecanProject/pecan/pull/3893/commits/37898592481de7ec1281f3b437f247a9742a8253


> 2. This if doesn't have an else, but it's not clear to me that the code will keep running if the if is skipped and no parameters are loaded

function doesn't need one; generate_joint_ensemble_design() only produces integer indices into the parameter bank it never loads or returns actual parameter values. if block is a guard that ensures samples.Rdata exists with at least 2N draws before we generate those indices. When condition is FALSE, file already has enough draws from a previous run, and the function proceeds to build the index design matrix via .map_sobol_to_indices(). actual MCMC draws are loaded later by run.write.configs, which reads samples.Rdata and selects draws using the indices we produce here.
This is the same pattern as the non sobol path at the bottom of the function,neither path has `else` both are "ensure file exists" guards before function returns its design matrix

modules/uncertainty/R/generate_joint_ensemble_design.R
divine7022: this is not a hack it's adapter required by sensobol's API design, because sensobol::sobol_matrices() returns a matrix of quasi-random numbers in [0, 1). But we need integer indices into a finite bank of MCMC parameter draws or discrete set of input file paths. We need mapping from continuous [0,1) to discrete {1, 2, ..., bank_size} that preserves uniformity; that said just by using sample() would destroy the quasi-random correlation structure that Saltelli design relies on for efficient variance decomposition. Using ceiling(x * size) would fail at x=0 (maps to 0, out of range). ryt now function uses floor(qunif(x, 1, size+1)) which is standard inverse CDF approach for discretizing uniform. The pmin(..., size) is a guard handles theoretical edge case where x = 1.0 exactly (shouldn't happen with QRN, but just for defensive puspose).

This does not affect sobol index computation. sensobol::sobol_indices() function takes the original [0,1) design matrix and the model output vector Y. It never sees the integer indices, those are only used to select which parameter draw or input file to use for each model run; sobol indices are computed from the outputs, not indices

modules/uncertainty/R/generate_joint_ensemble_design.R
divine7022: fair critique,
saltelli design requires specific cross-matrix structure (A, B, and k AB_i permutations) to produce valid first-order and total-order variance decomposition estimates. sensobol package implements this precisely per Saltelli et al. (2010) and uses sobol quasi-random sequences (type = "QRN") which give better space-filling properties than pseudo-random sampling for the same N, meaning more precise sensitivity indices per model run. Building this correctly from scratch is non-trivial; my inclination is using a published, peer-reviewed implementation (Puy et al. 2022, JSS 102(5)) is a right call here.

<ensemble> block sampling specifications (method = "sampling", "looping", etc.) are not ignored, in current code sobol path only handles its own independent factors via sobol_matrices(). All child inputs and non sobol inputs still go through input.ens.gen(), which respects the method from the <ensemble> block ( `method = samp.ordered[[i]]$method,  <-- from <ensemble> block`). only inputs that bypass input.ens.gen() are independent sobol factors and that's by design. For these sensobol determines their sampling structure because cross-matrix construction is what makes sobol index estimation work. we can't use sample() or rep_len() for independent Sobol factors, it would break the variance decomposition

PR #3896, opened 2026-03-25
Fix #3895: Add set.seed to tutorials for reproducible rendering

### LINKED ISSUE #3895: Use set.seed in demos
PEcAn Tutorials 1 and 2 both produce very slightly different plots each time they run, The user won't notice these differences, but Git does and it stores the resulting binary diff of each image every time the build job runs. These are rapidly increasing the size of the `pecan-documentation` repo (my local copy is now over 750 MB). The problem can likely be mitigated by setting a random seed at the top of each notebook.

We should:
* Add `set.seed` calls to the tutorials in `documentation/tutorials` and verify that this causes images to reproduce hash-identically.
* review tutorials and package vignettes for other sources of irreproducibility. Common culprits include calls to Sys.Date in places that don't strictly need it or  hashing of objects that are not expected to stay identical.

man080107:
Fixes #3895

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/documentation/tutorials/AnalyzeOutput/modelVSdata.Rmd
+++ b/documentation/tutorials/AnalyzeOutput/modelVSdata.Rmd
@@ -1,3 +1,7 @@
+```{r setup, include=FALSE}
+set.seed(20260325)
+```
+
 ### Simple Model-Data Comparisons
 
 #### Author: Istem Fer, Tess McCabe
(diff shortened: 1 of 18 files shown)


infotroph:
Duplicate of #3897

PR #3900, opened 2026-03-26
get.parameter.samples: avoid dropping NULL priors

infotroph:
One real bug, one unneeded intermediate object, and a bunch of formatting fixed while trying to find the problem:

* The real bug: In get.parameter.samples, `prior_distns_list[[i]] <- distns$prior.distns` was dropping list entries when `distns$prior.distns` was NULL, and therefore causing get_parameter_samples to (correctly!) fail its assertion that `length(prior_distns_list) == length(pfts)`. This is because R interprets `somelist[i] <- NULL` as "remove entry i from somelist entirely, shortening the list as needed" rather than "make the value at position i be null".
* The uneeded object: The temporary `distns` environment is no longer needed now that `load_posteriors` encapsulates all the file reading for us. Changed to assign directly from `posteriors` into the distn lists.
* The formatting: Before I tracked the issue into get.parameter.samples I spent some time being confused by missing braces and inconsistent indentation in get.ensemble.samples, so I committed some formatting fixes as I went. I recommend reviewing with "hide whitespace" turned on. If I did this right, the changes in ensemble.R will have zero functional effect.

### DIFF
--- a/modules/uncertainty/R/ensemble.R
+++ b/modules/uncertainty/R/ensemble.R
@@ -1,34 +1,35 @@
-##' Reads output from model ensemble
-##'
-##' Reads output for an ensemble of length specified by \code{ensemble.size} and bounded by \code{start.year} 
-##' and \code{end.year}
-##' @title Read ensemble output
-##' @return a list of ensemble model output 
-##' @param ensemble.size the number of ensemble members run
-##' @param pecandir specifies where pecan writes its configuration files
-##' @param outdir directory with model output to use in ensemble analysis
-##' @param start.year first year to include in ensemble analysis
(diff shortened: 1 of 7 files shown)


ashiklom (approved):
Nice catch!

ashiklom:
For what it's worth, if we're already depending on `purrr`, `purrr::pluck` (if missing, use `NULL` or a custom value) and `purrr::chuck` (if missing, throw an error) may be useful to adapt here and elsewhere to avoid depending on R's somewhat [confusing subsetting defaults](https://adv-r.hadley.nz/subsetting.html#subsetting-oob). But that's for a future PR.

PR #3901, opened 2026-03-27
Transition Matrix, Prediction, and Scenarios Scripts 

ananyak534:
Pull request for my script that generates a transition matrix on crop classes for the 2018-2023 data. The second file, 'predict_and_store.R', uses the matrix to predict crop classes up until a given year for one of the four seasons. 

What this does
- Combines crop data from 2018–2023
- Cleans up crop sequences per parcel
- Builds a transition matrix of crop changes
- Uses a Markov model to predict future crops (2024–2030)
- Plots actual vs predicted for a sample parcel

Key parts
- Transitions are counted and turned into a probability matrix (with smoothing)
- Predictions are generated by applying the matrix and sampling next states

Outputs
- seq_long.csv: long-format sequences
- full_transition_matrix.csv: transition probabilities

Notes
- Transition step is slow, so the results and files are saved and reused
- this uses simplified Markov behavior where next state depends only on current

- How to run
- Run 'transition_matrix.R' once to generate CSVs (but it will take a bit of time ~20 minutes)
- run 'predict_and_store.R' using saved files --> can update season_idx and end_year values to modify predictions


modules/data.remote/inst/predictions_with_tmatrix.R
```diff
@@ -0,0 +1,93 @@
...
+  top_p = numeric(length(years))
+  probs_list = if (return_probs) vector("list", length(years)) else NULL
+  
+  for (i in seq_along(years)) {
+    k = k_vec[i]
+    
+    if (k == 0) {
+      p = rep(0, nrow(tmat)); names(p) = rownames(tmat); p[cur] = 1
+    } else {
+      p = predict_k_steps(tmat, cur, k)
```
mdietze: I'd recommend that you rethink what you're doing here to make the predictions iteratively (use the last prediction as the input to the next prediction) rather than repeating the prediction from 0 to k each time k increases by 1

modules/data.remote/inst/predict_and_store.R
```diff
@@ -0,0 +1,120 @@
...
+  season == season_idx,
+  .SD[which.max(year)],
+  by = parcel_id
+]
+
+#clean classes and set in same order as transition matrix 
+start_info[, CLASS := trimws(as.character(CLASS))]
+states = rownames(tmat_final)
+
+##all predictions
```
mdietze: What's the relationship between "predictions_with_tmatrix.R" and the code in this file? It doesn't look like you're using those functions here. I'd recommend either using those functions or converting the core of this code into a function, and then have the script here focus on reading specific data, calling the prediction function, and then performing the validation. Indeed, such a script might be better reformatted as a Rmd vignette.

mdietze (approved):

PR #3905, opened 2026-03-28
Tweaks for non-SDA Sipnet restart

infotroph:
* Quick hack: When reading restarts, set litter and litter water NA if not present in the output file
* Overwrite an existing events file (I sure do wish file.copy's default `overwrite = FALSE` would default to reporting the file already exists instead of silently skipping the copy...)


ashiklom (approved):

mdietze:
@DongchenZ could you take a quick look at this to make sure it's not going to break the SDA read restart functionality

DongchenZ (approved):
Sorry for the late response. The current code will not affect the existing NA SDA runs, and it provides robust control over the value returned by the `read_restart` feature. Thanks!

PR #3906, opened 2026-03-29
fix(uncertainty): resolve SA median manifest lookup (#3882)

### LINKED ISSUE #3882: Manifest-based OAT sensitivity drops median row for shared SA-median run
### Bug Description

This is based on the `modules/uncertainty/vignettes/uncertainty.qmd` and I am unsure how general the issue is.

This was observed in the manifest-based OAT sensitivity. The older sa.run.ids-based path on main should not have this specific lookup problem.

The OAT sensitivity workflow appears to drop the median (`50`) row from sensitivity outputs.

In the full workflow, the median run exists (`out/SA-median--1/2004.nc`), but `read.sa.output()` warns:

> Run ID invalid or missing for `<trait>` 50

and the saved `sensitivity.output.*.Rdata` contains `NA` for the `50` row.

The issue seems to be a mismatch between manifest writing and lookup:

- `write.sa.configs()` writes one shared median SA run with:
  - `run_id = SA-median--1`
  - `pft_name = "NA"`
  - `trait = "NA"`
  - `quantile = 50`
- `read.sa.output()` looks up runs by matching:
  - `type`
  - `pft_name`
  - `trait`
  - `quantile`

So the shared median run can never match a trait- and PFT-specific lookup.

In the saved downstream analysis objects, the median sample value is still present in sa.samples, but the corresponding median output is NA, so the downstream sensitivity fit proceeds without a valid median response.

### To Reproduce

Minimal reprex showing the current `read.sa.output()` behavior:

``` r
withr::with_tempdir({
  manifest <- data.frame(
    type = "Sensitivity",
    pft_name = "temperate.coniferous",
    trait = "growth_resp_factor",
    quantile = "50",
    run_id = NA_character_
  )
  write.csv(manifest, "runs_manifest.csv", row.names = FALSE)

  out <- PEcAn.uncertainty::read.sa.output(
    traits = "growth_resp_factor",
    quantiles = "50",
    pecandir = getwd(),
    outdir = getwd(),
    pft.name = "temperate.coniferous",
    start.year = 2004,
    end.year = 2004,
    variable = PEcAn.utils::convert.expr("NPP")
  )

  sensitivity.output <- list("temperate.coniferous" = out)
  save(sensitivity.output, file = "sensitivity.output.test.NPP.2004.2004.Rdata")
  rm(sensitivity.output)
  load("sensitivity.output.test.NPP.2004.2004.Rdata")

  sensitivity.output[["temperate.coniferous"]]
})
#>    growth_resp_factor
#> 50                 NA
```

<sup>Created on 2026-03-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Standard output and standard error

``` sh
2026-03-18 18:29:52.480915 WARN   [PEcAn.uncertainty::read.sa.output] : 
   Run ID invalid or missing for growth_resp_factor 50 
2026-03-18 18:29:52.506574 INFO   [PEcAn.uncertainty::read.sa.output] : 
   reading sensitivity analysis output for model run at 50 quantiles of 
   trait growth_resp_factor 
```

## Expected behavior

The median row should be populated, since it is part of the intended OAT sensitivity workflow.

## Machine (please complete the following information):

OS: macOS
Version: PEcAn worktree based on develop as of 2026-03-18

## Additional context

This appears to be a mismatch between manifest writing and lookup:

`write.sa.configs()` writes one shared median SA run with run_id = SA-median--1`, pft_name = "NA", trait = "NA", quantile = 50
`read.sa.output()` looks up runs by matching type, pft_name, trait, and quantile
So the shared median run cannot match a trait- and PFT-specific lookup.

This appears to be specific to the newer manifest-based SA lookup path. In the older sa.run.ids-based path, the median run is stored in the per-trait/PFT run-id table, so this specific warning should not occur under normal use.

man080107:
## Summary

Fixes #3882 - 
read.sa.output() was silently dropping the median (q50) run when using the manifest-based OAT sensitivity path.

## Root Cause

write.sa.configs() writes a single **shared median run** with pft_name = "NA" and 	rait = "NA" in the manifest. However, 
ead.sa.output() performed an exact per-trait/per-PFT lookup, so no row ever matched for the median quantile - the result was always NA, and the median was excluded from downstream sensitivity analyses.

## Fix

Added a fallback in 
ead.sa.output(): when quantile == "50" and no exact match is found, the function now searches for the shared NA/NA median entry before emitting a warning.

## Changes

- modules/uncertainty/R/sensitivity.R - fallback median lookup logic
- modules/uncertainty/tests/testthat/test-read_sa_output_median.R - two regression tests:
  - positive: median row is correctly resolved via the fallback
  - negative: warns when neither exact match nor fallback entry exists
- modules/uncertainty/NEWS.md - changelog entry

## Testing

New tests added in 	est-read_sa_output_median.R covering both the happy path (fallback resolves correctly) and the error path (warning emitted when no match at all).


man080107:
@dlebauer!!! please do quick review on it....

modules/uncertainty/tests/testthat/test-read_sa_output_median.R
```diff
@@ -0,0 +1,99 @@
...
+#
+# Background: write.sa.configs() writes the median (q50) run as a single shared
+# entry with pft_name = "NA" and trait = "NA".  read.sa.output() previously did
+# a strict per-trait/per-PFT lookup, so the median row always failed to match
+# and the 50th-quantile cell was returned as NA.
+#
+# The fix adds a fallback: when quantile == "50" and no exact match is found,
+# look for the shared median entry before giving up.
+#
+# Reference: https://github.com/pecanproject/pecan/issues/<issue-number>
```
dlebauer:
```suggestion
# Reference: https://github.com/pecanproject/pecan/issues/3882
```

dlebauer (approved):
Logic looks right to me. The fallback matches how write.sa.configs() stores the shared median run, and the regression tests test the failure mode.

Thank you for fixing this!

dlebauer:
@man080107 Please check test failures

man080107:
@dlebauer !! now i think it is ready to merge

dlebauer:
@man080107 it is still failing one required test https://github.com/PecanProject/pecan/actions/runs/24274463741/job/70885636636?pr=3906

infotroph:
@man080107 is this PR still on your radar? The remaining test failures look real and will need to be addressed before we can merge this.

Issue #3595, opened 2025-08-13
create functions to translate NDTI --> tillage modifier


dlebauer:
Goal --> translate $\delta$NDTI to SIPNET decomposition rate modifiers

Approach: 
1. [Model agnostic function] use published mappings of NDTI --> tillage intensity categories, with linear or spline interpolation through category midpoints. [This could be model-independent]
  - inputs: date, delta NDTI
  - outputs: date, tillage intensity
2. [SIPNET specific function] that interpolates from tillage intensity --> decomposition modifier
  - inputs: date, tillage intensity
  - outputs: date, decomposition rate modifier

omkarrr2533:
Hi @dlebauer, is anyone working on this? I'd like to take it on

I see it's two functions one model-agnostic (δNDTI → tillage intensity) and one SIPNET-specific (tillage intensity → decomposition modifier). Could you point me to which published NDTI-tillage mappings you're referring to? Want to make sure I use the right references before I start

dlebauer:
Right now, @mdietze and @sarahkanee are working on converting NDTI to categorical zero, low, and conventional till. But we still don't have a way to convert to SIPNET specific values.

I'll let them chime in on the work that they are doing, their thoughts on conversion to SIPNET tillage factors, and what validation might look like.

mdietze:
We're producing a product that is the % change in NDTI between the fallow period baseline and the observed fallow minimum. Based on the literature we'd recommend <30% be considered no till and  >70% be considered conventional till. We have no thoughts on how to convert that to SIPNET tillage factors or how to validate, other than to note that we're providing % delta NDTI by fallow season for all(*) cropland parcels for 2016 and 2018-2024. I have no idea what data might be available to validate against in the field (SOC? soil respiration?), but in theory if you had such data you could compare the model outputs to the data under different scaling factors and then try to empirically optimize that relationship. I'll also note that there's no a priori reason to assume that the relationship is linear -- might be a parsimonious place to start modeling, but it'd be an assumption.

* all = almost all, with a small % of NAs do to missing data that we haven't gapfilled yet

omkarrr2533:
@mdietze @dlebauer — I've been going through the existing code to understand exactly how everything fits together before writing anything. A few things I want to confirm before I open a PR:
[...]

PR #3917, opened 2026-04-04
Add ndti_to_sipnet_tillage() to convert NDTI drop to SIPNET tillage effectiveness (#3595)

omkarrr2533:
Addresses part of #3595

Adds `ndti_to_sipnet_tillage()` in `modules/data.land/R/` to convert
the fractional NDTI drop from the CCMMF pipeline to `tillage_eff_0to1`
for the PEcAn events JSON.

Thresholds (<0.30 no-till, >0.70 conventional) are from @mdietze and
@sarahkanee. Linear mapping is the default; saturating curve is available
via `method = "saturating"`. All thresholds are exposed as parameters
for future calibration.

7 tests covering structure, boundaries, NA propagation, and error conditions.
No existing functions changed.

Screenshot 2026-04-04 180753


modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
+#' Convert fractional NDTI drop to SIPNET tillage effectiveness
+#'
+#' Maps the fractional NDTI drop produced by the CCMMF NDTI pipeline
+#' to the \code{tillage_eff_0to1} value used in the PEcAn events JSON
+#' schema and written to SIPNET \code{events.in} by
+#' \code{\link[PEcAn.SIPNET]{write.events.SIPNET}}.
+#'
+#' @param dates Date vector. Observation dates corresponding to each
```
mdietze: Drop. Not part of the calculation

modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
...
+#'   tillage event. Same length as \code{delta_ndti}.
+#' @param delta_ndti numeric vector. Fractional NDTI drop in \[0, 1\],
+#'   computed as \code{(max_before_min - min_val) / max_before_min}
+#'   from a smoothed NDTI time series over the fallow season.
+#' @param no_till_threshold numeric scalar. \code{delta_ndti} values at
+#'   or below this threshold are mapped to zero effectiveness (no-till).
+#'   Default 0.30, based on Dietze et al. (pers. comm.).
+#' @param conv_till_threshold numeric scalar. \code{delta_ndti} values
```
mdietze: Looking through what's actually being proposed, I think this parameter is going to end up being unneeded. drop

modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
...
+#'   at or above this threshold are mapped to \code{max_modifier}
+#'   (conventional tillage). Default 0.70.
+#' @param max_modifier numeric scalar. Maximum \code{tillage_eff_0to1}
```
mdietze: In what I'm going to propose below this should default to 2.5 to achieve a linear ramp from 0.3 to 0.7 (i.e. dy/dx = 1 / (0.7-0.3) )

modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
...
+#'   value, corresponding to conventional tillage. Default 1.0. Can be
+#'   lowered to constrain the modeled tillage response without changing
+#'   the shape of the mapping. Useful for sensitivity analysis or when
+#'   field calibration data suggest a lower ceiling.
+#' @param method character. Functional form of the NDTI-to-effectiveness
```
mdietze: Drop. Implementing an alternative is premature at this point and the proposed alternative isn't realistic

modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
...
+#'     \item{\code{"saturating"}}{Effectiveness follows
+#'       \code{1 - exp(-3 * x)} between the thresholds, reaching
+#'       approximately 95\% of \code{max_modifier} at
+#'       \code{conv_till_threshold}. Appropriate if moderate tillage
+#'       already produces a near-maximum decomposition response.}
+#'   }
+#'
+#' @return A data.frame with columns:
+#'   \describe{
+#'     \item{date}{Date. Same as input \code{dates}.}
```
mdietze: drop date per earlier comment

modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
...
+  if (any(delta_ndti < 0, na.rm = TRUE)) {
+    PEcAn.logger::logger.warn(
+      "Negative delta_ndti values detected; these will be treated as no-till (clamped to 0)."
+    )
+  }
+
+  ## --- mapping ---
+  # Scale delta_ndti to [0, 1] within the threshold window.
+  # Values below no_till_threshold → 0; above conv_till_threshold → 1.
+  scaled <- (delta_ndti - no_till_threshold) /
```
mdietze: what you've proposed isn't going to work as, once max_modifier is included, it's not going to respect the 0,1 range (i.e., if you put a slope of 2 in then you get a max of 2 which isn't acceptable. I'd instead recommend a linear response that has a x-intercept at no_till_threshold and then a slope specified by a parameter, which you'd then clamp to 0,1 using the pmin/pmax in the next line. The defaults I proposed would lead to that saturation at the old conv_till_threshold.

modules/data.land/R/ndti_to_sipnet_tillage.R
```diff
@@ -0,0 +1,171 @@
...
+    (conv_till_threshold - no_till_threshold)
+  scaled <- pmin(pmax(scaled, 0), 1)
+
+  if (method == "linear") {
```
mdietze: drop all of this if and else

mdietze (changes requested):

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: done

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: Makes sense. I've removed conv_till_threshold completely.

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: Added the slope parameter and set the default to 2.5 as suggested.

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: Agreed, dropped the method parameter to keep it simple.

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: done

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: Removed

modules/data.land/R/ndti_to_sipnet_tillage.R
omkarrr2533: good catch on the clamping issue. I dropped max_modifier and updated the calculation to exactly what you suggested: pmin(pmax((delta_ndti - no_till_threshold) * slope, 0), 1). It now safely clamps to [0, 1] regardless of the slope.

omkarrr2533:
@mdietze Sir, Thanks for taking the time to review this and for outlining the exact math needed! I dropped the extra parameters, dates, and if/else logic, and updated the core calculation to use the slope/intercept logic with your recommended defaults. I also updated the tests to match the new simplified vector output. 👍

Yeligay8:
What do think of adding these?
1)Review the updated vectorized calculation: pmin(pmax((delta_ndti - no_till_threshold) * slope, 0), 1).
2)Confirm tests pass and correctly cover NA and boundary cases.
3) Since all extra parameters and logic have been removed, this PR is now ready for merge after final approval. @mdietze

mdietze (approved):

Yeligay8:
New function: ndti_to_sipnet_tillage() in modules/data.land/R/.
Simplified calculation:
Uses linear mapping with a slope/intercept logic.
Clamped to [0, 1] via pmin(pmax(..., 0), 1) to ensure valid output.
No extra parameters, no dates column, and no if/else logic—fully vectorized.
Thresholds:
Default no-till threshold = 0.30 (values below → 0 effectiveness).
Default slope = 2.5, achieving linear ramp to the old conv-till threshold (~0.70).
Other thresholds and alternative methods dropped per reviewer guidance (simpler, realistic).
Tests added:
7 unit tests covering structure, boundaries, NA propagation, and error conditions.
Ensures output is always [0,1] and handles edge cases correctly.
No existing functions modified, ensuring backward compatibility.

PR #3919, opened 2026-04-05
SIPNET workflow for restarting with events

ashiklom:
Prototype of running SIPNET with event files that include changes in crops. A few implementation notes:

- This supports PEcAn ensemble inputs (including for events)

- `event.json` files for multiple ensemble members are stored in `run$inputs$events$source`. SIPNET-specific `event.in` files are stored in `run$inputs$events$path` (like other inputs). This is because the current functionality for finding segments to split the runs uses the JSON files. @infotroph 's `subset_paths` function from an earlier draft has been adapted and modified to subset both `path` and `source` paths (as long as they have the same lengths).

- This circumvents `runModule.start.model.runs` and uses a direct execution loop instead. However, the output is PEcAn standard and follows PEcAn configuration conventions, so downstream analyses/workflows should work out of the box.

It's a bit hacky, but I think it does what it's supposed to and should be enough to unblock other CCMMF modeling tasks (@dlebauer).


pixi.lock
infotroph, line 1: remove

pixi.toml
infotroph, line 1: remove

modules/data.land/tests/testthat/test-events_to_crop_cycle_starts.R
```diff
@@ -0,0 +1,49 @@
+test_that("non-planting events are ignored", {
+  dat <- dplyr::tribble(
+    ~site_id, ~date, ~event_type, ~crop,
```
divine7022:
```suggestion
    ~site_id, ~date, ~event_type, ~crop_code,
```

modules/data.land/tests/testthat/test-events_to_crop_cycle_starts.R
```diff
@@ -0,0 +1,49 @@
...
+  )
+  res <- find_crop_changes(dat)
+  expect_equal(nrow(res), 1)
+  expect_equal(res$date, "2016-01-01")
+  expect_equal(res, find_crop_changes(dat[-c(2, 4), ]))
+})
+
+test_that("nonconsecutive runs of the same crop counted separately", {
+  dat <- dplyr::tribble(
+    ~site_id, ~date, ~event_type, ~crop,
```
divine7022: same as commented above, use consistent column name; and else where used

workflows/sipnet-restart-workflow/02-prepare-settings.R
```diff
@@ -0,0 +1,141 @@
...
+dp_data <- read.csv(config[["dp_path"]]) |>
+  dplyr::filter(.data$id == .env$site_id)
+site_lat <- dp_data[["lat"]]
+site_lon <- dp_data[["lon"]]
+
+events_dir <- file.path(outdir_root, "events")
+events_files <- list.files(events_dir, recursive = TRUE, full.names = TRUE)
+events_json_files <- as.list(grepv(".*\\.json$", events_files))
+names(events_json_files) <- paste0("path", seq_along(events_json_files))
+sipnet_eventfiles <- as.list(grepv(".*\\.sipnet/events-.*\\.in", events_files))
```
divine7022: grepv() is not base R, utils, or any package loaded in this script; guessing this was a personal helper that didn't make it into the PR ?

```suggestion
events_json_files <- as.list(grep(".*\\.json$", events_files, value = TRUE))
names(events_json_files) <- paste0("path", seq_along(events_json_files))
sipnet_eventfiles <- as.list(grep(".*\\.sipnet/events-.*\\.in", events_files, value = TRUE))
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+    run_settings$run$inputs$events$source
+  ) |>
+    dplyr::ungroup()
+
+  # Get segments
+  segment_rootdir <- file.path(file.path(run_dir, "segments"))
+  segments <- crop_cycles |>
+    dplyr::rename(start_date = "date") |>
+    dplyr::mutate(
+      end_date = dplyr::lead(.data$start_date - 1, default = as.Date(run_settings$run$end.date)),
```
divine7022: ok so the -1 binds to start_date before lead() shifts it; what you actually get is lead(start_date - 1), not lead(start_date) - 1. So for non final segments end date is next segment's start minus TWO days instead of one, that's a 1 day gap at every crop transition where sipnet isn't running and produces no output.
five crop changes in a rotation = five missing days. Not gonna blow up results, but output timestamps won't line up with event file and someone's gonna spend a day debugging that 😬

```suggestion
      end_date = dplyr::lead(.data$start_date, default = as.Date(run_settings$run$end.date) + 1) - 1,
```

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
+    )
+  }
+  result
+}
+
+#' Split sipnet `events.in` files according to start and stop date
+#'
+#' @param eventfile Path to `events.in` file.
+#' @inheritParams split_inputs.SIPNET
+split_sipnet_events <- function(start.time, stop.time, eventfile, overwrite = FALSE, outpath = FALSE) {
```
divine7022:
```suggestion
split_sipnet_events <- function(start.time, stop.time, eventfile, overwrite = FALSE, outpath = NULL) {
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+    PEcAn.logger::logger.severe("Could not find manifest file: ", manifest_file)
+  }
+  inputs_runs <- read.csv(manifest_file)
+  if (!is.null(input_design)) {
+    inputs_runs <- cbind.data.frame(inputs_runs, input_design)
+  }
+
+  new_jobfiles <- character()
+
+  for (i in seq_along(nrow(inputs_runs))) {
```
divine7022: nrow(inputs_runs) returns a single integer (let say, 100). seq_along(100) returns c(1), effectively writes configs for first ens member; rest get original unsegmented job.sh
```suggestion
  for (i in seq_len(nrow(inputs_runs))) {
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+
+  crop_cycles <- PEcAn.data.land::events_to_crop_cycle_starts(
+    run_settings$run$inputs$events$source
+  ) |>
+    dplyr::ungroup()
+
+  # Get segments
+  segment_rootdir <- file.path(file.path(run_dir, "segments"))
+  segments <- crop_cycles |>
+    dplyr::rename(start_date = "date") |>
```
divine7022: still starts segments from first crop cycle, not run$start.date. If met starts 2016 but first planting is spring 2017, no spinup for soil. worth at least documenting the decision and putting a TODO for an optional fallow pre-segment

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+  dplyr::case_when(
+    crop_code == "P1" ~ "annual_crop",
+    crop_code == "G2" ~ "annual_crop",
+    cls == "D" ~ "temperate.deciduous",
+    cls == "F" ~ "annual_crop",
+    cls == "G" ~ "grass",
+    cls == "P" ~ "grass",
+    cls == "R" ~ "grass",
+    is.na(crop_code) ~ "soil",
+    TRUE ~ "UNKNOWN_PFT"
```
divine7022, line 57: "UNKNOWN_PFT" potentially gonna tradeoff downstream without any error. 
seems like deserves logger.severe

I know it's named _example but this function is what actually runs. If it's destined to be replaced, put a TODO

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+
+  jobsh_files <- character()
+
+  for (isegment in seq_len(nrow(segments))) {
+    segment <- segments[isegment, ]
+    dstart <- segment[["start_date"]]
+    dend <- segment[["end_date"]]
+    segment_dir <- segment[["segment_dir"]]
+
+    unlink(segment_dir, recursive = TRUE)
```
divine7022: which does before each segment, assume segment 3 of 5 fails, everything reruns from scratch (know it's existing limitation. does it make sense to check segment already completed (output files exist) and skip ?
but not a blocker

workflows/sipnet-restart-workflow/config.yml
```diff
@@ -0,0 +1,25 @@
...
+  outdir_root: "workflows/sipnet-restart-workflow/_test"
+  irrigation_path: "/projectnb/dietzelab/ccmmf/usr/ashiklom/event-outputs/irrigation_10000.parquet"
+  parcel_path: "/projectnb/dietzelab/ccmmf/LandIQ-harmonized-v4.1/parcels.gpkg"
+  dp_path: "/projectnb/dietzelab/ccmmf/management/irrigation/design_points.csv"
+  sipnet_binary: "/projectnb/dietzelab/ccmmf/usr/ashiklom/pecan/sipnet/sipnet"
+  met_dir: "/projectnb/dietzelab/ccmmf/ensemble/ERA5_SIPNET"
+  ic_dir: "/projectnb/dietzelab/ccmmf/ensemble/IC_files"
+  planting_events_dir: "/projectnb/dietzelab/ccmmf/management/event_files"
+  mslsp_path: "/projectnb/dietzelab/ccmmf/management/phenology/matched_landiq_mslsp_v4.1"
+  pft_dir: "..."
```
divine7022: will crash, either fill in SCC path or make error message helpful

workflows/sipnet-restart-workflow/02-prepare-settings.R
```diff
@@ -0,0 +1,141 @@
...
+names(pfts) <- rep("pft", length(pfts))
+
+ensemble_settings <- list(
+  size = n_ensemble,
+  variable = "LAI",
+  variable = "SoilMoist",
+  variable = "GPP",
+  variable = "SoilResp",
+  variable = "AGB",
+  variable = "NEE",
```
divine7022, line 87: as separate list keys ? but only last one is accessible by name
```suggestion
  variable = c("LAI", "SoilMoist", "GPP", "SoilResp", "AGB", "NEE")
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+  i_param <- run_row[["param"]] %||% 1
+  run_traits <- lapply(ensemble_samples, \(dat) dat[i_param, ])
+
+  crop_cycles <- PEcAn.data.land::events_to_crop_cycle_starts(
+    run_settings$run$inputs$events$source
+  ) |>
+    dplyr::ungroup()
+
+  # Get segments
+  segment_rootdir <- file.path(file.path(run_dir, "segments"))
```
divine7022:
```suggestion
  segment_rootdir <- file.path(run_dir, "segments")
```

pixi.lock
divine7022, line 1: is lock file needed ? cuz pecan doesn't use pixi

models/sipnet/R/write.configs.SIPNET.R
divine7022:
```suggestion
#' @param trait.values list of samples for a given trait
```

models/sipnet/R/combine_sipnet_out.R
```diff
@@ -0,0 +1,30 @@
...
+#' readable with [read_sipnet_out()].
+#'
+#' @return `outfile` (path to output file), invisibly
+#' @export
+combine_sipnet_out <- function(directory, outfile, files = NULL) {
+  if (missing(directory) && is.null(files)) {
+    PEcAn.logger::logger.severe("Must provide either `directory` or `files`")
+  }
+  if (is.null(files)) {
+    files <- sort(list.files(directory, "sipnet\\.out", full.names = TRUE, recursive = TRUE))
```
divine7022, line 25: since segments are named segment_001, segment_002, etc. worth comment noting the assumption

modules/data.land/R/events_to_crop_cycle_starts.R
```diff
@@ -0,0 +1,53 @@
...
+    tidyr::unnest(.data$events) |>
+    dplyr::mutate(date = as.Date(.data$date)) |>
+    find_crop_changes()
+}
+
+# helper for events_to_crop_cyle_starts,
+# mostly to ease unit testing
+find_crop_changes <- function(event_df) {
+  event_df |>
+    dplyr::filter(.data$event_type == "planting") |>
```
divine7022, line 48: restart carries forward all state, planting events are additive; if there is no harvest between segments, you get accumulation. then crop cycle ends up only looks at planting events not harvest

Worth either validating harvest precedes later plantings or log warning when it doesn't

divine7022 (approved):
LGTM, shaping up nicely, shift to job_segmented.sh with symlink integration to standard pipeline is right direction, killing cdo dependency with combine_sipnet_out is solid cleanup.
none of the stuff I flagged are blockers, it's all fixable; 
except few inline suggested.

resolve conflicts and drop 91-debug.R and 92-find-winter-crop.R

workflows/sipnet-restart-workflow/02-prepare-settings.R
infotroph: It's in base R since 4.5, but I agree grep(..., value=TRUE) is the compatible fallback.

modules/data.land/R/events_to_crop_cycle_starts.R
infotroph: That's a reasonable check to do somewhere to make sure management is reasonable, but this function needs to faithfully create unreasonable schedules if you give them to it.  For restart purposes we really do only care if a new crop has arrived (because that's when parameters need to change); you can plant the same crop many times, and harvest it as many or few as you like in between them, and not change anything for (crop-triggered) restart purposes.

workflows/sipnet-restart-workflow/02-prepare-settings.R
ashiklom: https://stat.ethz.ch/R-manual/R-devel/library/base/html/grep.html

Base R, but added in a very recent R version (4.5.0). I'll change this to `grep(..., values = TRUE)` for compatibility.

workflows/sipnet-restart-workflow/utils.R
ashiklom: This might need a bit of more (slightly) complex logic. I previously had the settings start date to the first planting as the first segment, but didn't correctly handle the edge case of start date == first planting date.

workflows/sipnet-restart-workflow/02-prepare-settings.R
ashiklom: PEcAn settings XML expects this (awkward) format of flat repeated tags:

```xml
<variable>NPP</variable>
<variable>AGB</variable>
<variable>LAI</variable>
```

This is downstream of XML's inability to represent arrays/vectors.

dlebauer (changes requested):
Great work! Excited to have this working! 🥳 

Before merging, please create a ticket to cover the refactoring plan we discussed last week.

From what I recall, the plan was approximately:

Goal: refactor segmented runs so that they integrate more cleanly with standard start_model_runs workflows. 

Specifically:
- write.configs generates config files
- sites and ensembles run in parallel via start_model_runs 
- This will require modifying or replacing the top‑level job.sh so that:
    - Segments run sequentially within a site, and
    - Sites still run in parallel via start_model_runs.

models/sipnet/R/combine_sipnet_out.R
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env Rscript
+
+#' Combine a bunch of `sipnet.out` files into a single file
+#'
+#' @param directory Parent directory to search for `sipnet.out` files. You must
+#' provide either this or an explicit vector of `files`.
+#' @param outfile File to which combined `sipnet.out` will be written
+#' @param files (optional) Explicit vector of paths to `sipnet.out` files to
+#' combine. Note that files need not be named `sipnet.out`, but they must be
+#' readable with [read_sipnet_out()].
```
infotroph:
```suggestion
#' @param files Optional vector of paths to files to combine.
#' All must be readable with [read_sipnet_out()].
#' If NULL, looks for files named `sipnet.out` in `directory` or its subdirs.
```

models/sipnet/R/combine_sipnet_out.R
```diff
@@ -0,0 +1,30 @@
...
+#' @return `outfile` (path to output file), invisibly
+#' @export
+combine_sipnet_out <- function(directory, outfile, files = NULL) {
+  if (missing(directory) && is.null(files)) {
+    PEcAn.logger::logger.severe("Must provide either `directory` or `files`")
+  }
+  if (is.null(files)) {
+    files <- sort(list.files(directory, "sipnet\\.out", full.names = TRUE, recursive = TRUE))
+  }
+  if (!(length(files) > 0)) {
```
infotroph:
```suggestion
  if (length(files) == 0) {
```

infotroph: (I'm half expecting @ashiklom to school me on the horrifying edge case that makes ! > 0 the more reliable choice here 😉)

models/sipnet/R/combine_sipnet_out.R
```diff
@@ -0,0 +1,30 @@
...
+    PEcAn.logger::logger.severe("No files provided; nothing to combine.")
+  }
+  flist <- lapply(files, read_sipnet_out)
+  combined <- do.call(rbind.data.frame, flist)
```
infotroph, line 31: Potential FUTURE enhancement: Provide an option to add a column with a filename or other source identifier, so that we could pipe output from directories full of sipnet.out straight into a multi-site/run netCDF without needing to write the plaintext back out in between.

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
-##' @title split_inputs.SIPNET
-##' @name  split_inputs.SIPNET
-##' @author Mike Dietze and Ann Raiho
-##' 
+#!/usr/bin/env Rscript
+
+#' Split SIPNET inputs into multiple files based on start and end time
+#'
+#' Subset each SIPNET input file and write a new file containing values `>=
+#' start.time` and `<= end.time` (note: `end.time` is inclusive!)
```
infotroph: This is a change from the existing split_inputs, correct?

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
+      overwrite = overwrite,
+      outpath = outpath
+    )
+  }
+
+  if ("events" %in% names(inputs)) {
+    result[["events"]][["path"]] <- split_sipnet_events(
+      start.time,
+      stop.time,
+      inputs$events$path,
```
infotroph: Copying `inputs` but then continuing to refer back to the original has potential for confusion -- I tend to just mutate `inputs` and let copy-on-write do its thing, but since you're already setting up `result` here I suggest using it everywhere below:

```suggestion
  result <- inputs
  if ("met" %in% names(result)) {
    result[["met"]][["path"]] <- split_sipnet_met(
      start.time,
      stop.time,
      result$met$path,
      overwrite = overwrite,
      outpath = outpath
    )
  }

  if ("events" %in% names(result)) {
    result[["events"]][["path"]] <- split_sipnet_events(
      start.time,
      stop.time,
      result$events$path,
```

models/sipnet/R/split_inputs.SIPNET.R
infotroph: +1

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
+  # Read events.in
+  prefix <- sub(".in", "", basename(eventfile), fixed = TRUE)
+  outpath <- outpath %||% dirname(eventfile)
+  dir.create(outpath, recursive = TRUE, showWarnings = FALSE)
+
+  #Changing the name of the files, so it would contain the name of the hour as well.
+  formatted_start <- strftime(start.time)
+  formatted_stop <- strftime(stop.time)
+  outfile <- file.path(outpath, paste0(prefix, ".", formatted_start, "-", formatted_stop, ".in"))
+  names(outfile) <- paste(start.time, "-", stop.time)
```
infotroph, line 62: Seems odd to put times into the filename when rows have to be days. Is this needed for matching names elsewhere in the process?

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
+  # different numbers of columns depending on event type. Instead, we just
+  # filter based on the event date, which is always the first two columns (year
+  # and doy).
+  events_in_raw <- readLines(eventfile)
+  events_in_list <- strsplit(events_in_raw, "[[:space:]]+")
+  years_in <- vapply(events_in_list, \(x) as.integer(x[[1]]), integer(1))
+  doys_in <- vapply(events_in_list, \(x) as.integer(x[[2]]), integer(1))
+  # Not using sipnet2datetime here because it returns times with time zones,
+  # which could cause subtle timezone-related bugs
+  dates_in <- as.Date(sprintf("%d-01-01", years_in)) + (doys_in - 1)
```
infotroph, line 82: This is clever, elegant, and I had to read it a few times to understand it 😅

models/sipnet/R/write.configs.SIPNET.R
```diff
@@ -78,6 +78,13 @@
 write.config.SIPNET <- function(defaults, trait.values, settings, run.id, inputs = NULL, IC = NULL,
                                 restart = NULL, spinup = NULL) {
 
+  if (!is.list(trait.values)) {
+    PEcAn.logger::logger.severe(paste0(
+      "Even though SIPNET only takes one PFT at a time, ",
```
infotroph: 👍 to the message -- I always forget this detail when I try to construct a call by hand.
That said the "only one PFT at a time" part isn't quite true; we can pass multiple PFTs and this function lumps them together into one actually-executed parameter set.
```suggestion
```

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
+  writeLines(events_out_str, outfile)
+  invisible(outfile)
+}
+
+##' split sipnet clim file based on start and end time
+##'
+##' @author Mike Dietze, Ann Raiho, Alexey Shiklomanov
+##'
 ##' @param start.time start date and time for each SDA ensemble
 ##' @param stop.time stop date and time for each SDA ensemble
```
infotroph:
```suggestion
##' @param start.time start date and time at which to start extraction
##' @param stop.time stop date and time at which to stop extraction
```

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -1,22 +1,108 @@
...
+  idx_keep <- (dates_in >= start.time) & (dates_in <= stop.time)
+  if (length(idx_keep) == 0) {
+    PEcAn.logger::logger.warn("No events to keep, so `events.in` will be empty")
+  }
+  events_out_str <- events_in_raw[idx_keep]
+  writeLines(events_out_str, outfile)
+  invisible(outfile)
+}
+
+##' split sipnet clim file based on start and end time
```
infotroph: I know this is a pre-existing function, but it keeps surprising me that `split_*` only extracts a single time range rather than creating as many files as you give it start and end dates. Maybe this title tweak can help without signing us up for deeper redesign?

```suggestion
##' Extract subset of a sipnet clim file based on start and end time
```

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -60,3 +150,21 @@ split_inputs.SIPNET <- function(start.time, stop.time, inputs, overwrite = FALSE
...
+    return(x)
+  }
+  xname <- deparse(substitute(x))
+  if (!inherits(x, "Date")) {
+    PEcAn.logger::logger.severe(
+      "Invalid ", xname, " : ", x,
+      " (class: ", class(x), ")"
+    )
+  }
+  PEcAn.logger::logger.warn(paste0(
```
infotroph: How sure are you this needs a warning? Seems like wanting to pass only dates would be a common(?) usage pattern and I'm not immediately thinking of how the coercion would hurt anything.

models/sipnet/R/split_inputs.SIPNET.R
```diff
@@ -60,3 +150,21 @@ split_inputs.SIPNET <- function(start.time, stop.time, inputs, overwrite = FALSE
...
+    xname, " is a date, but this function expects a datetime. ",
+    "Coercing to datetime by setting to midnight UTC."
+  ))
+  as.POSIXct(x)
```
infotroph: In particular since we know this is arriving as a date and we want the added time to be midnight, what are the cases that wouldn't be handled by 
```suggestion
  as.POSIXct(x, tz = "UTC")
```

modules/uncertainty/R/generate_joint_ensemble_design.R
```diff
@@ -58,7 +58,7 @@ generate_joint_ensemble_design <- function(settings,
   if (sobol) {
     ensemble_size <- as.numeric(ensemble_size) * 2
   }
-  ens.sample.method <- settings$ensemble$samplingspace$parameters$method
+  ens.sample.method <- settings$ensemble$samplingspace$parameters$method %||% "uniform"
```
infotroph: Is this change intentional or coming in from a merge? I have a fuzzy memory that maybe this was intentionally _not_ given a default, but the blame on this line isn't telling me much.

workflows/sipnet-restart-workflow/01-prepare-events.R
```diff
@@ -0,0 +1,172 @@
...
+    "wood_c_kg_m2" = "C_STEM",
+    "fine_root_c_kg_m2" = "C_FINEROOT",
+    "coarse_root_c_kg_m2" = "C_COARSEROOT",
+    "leaf_n_kg_m2" = "N_LEAF",
+    "wood_n_kg_m2" = "N_STEM",
+    "fine_root_n_kg_m2" = "N_FINEROOT",
+    "coarse_root_n_kg_m2" = "N_COARSEROOT"
+  )
+
+# Phenology is not currently used...
```
infotroph, line 56: My workflows pull it in, but as one CSV that gets used by every site without any subsetting needed -> restart workflow shouldn't need to worry about it.

workflows/sipnet-restart-workflow/01-prepare-events.R
```diff
@@ -0,0 +1,172 @@
...
+    .data$date >= .env$start_date
+  ) |>
+  dplyr::collect() |>
+  tibble::as_tibble() |>
+  dplyr::select(
+    "site_id", "event_type", "date", dplyr::starts_with("frac_")
+  )
+
+# End with the final harvest
+end_date <- max(harvest_events$date)
```
infotroph, line 83: Why not continue through the end of the year to capture post-harvest biogeochemistry?

workflows/sipnet-restart-workflow/01-prepare-events.R
```diff
@@ -0,0 +1,172 @@
...
+  arrow::open_dataset() |>
+  dplyr::collect() |>
+  dplyr::filter(.data$site_id == as.character(.env$pid)) |>
+  dplyr::mutate(date = as.Date(.data$date)) |>
+  # Start no earlier than 2016 because our met doesn't go back before 2015
+  dplyr::filter(date >= as.Date("2016-01-01")) |>
+  tibble::as_tibble()
+
+# Start date is the first planting (after 2016)
+start_date <- min(planting$date)
```
infotroph, line 40: Is this a run-wide start date? Recall that there might be perennials in the run too...

infotroph: Ah, ok, I see this is run on one parcel at a time. Probably still worth extracting events in whole-year chunks rather than clip to plant and harvest, though.

workflows/sipnet-restart-workflow/01-prepare-events.R
infotroph, line 1: Lots of good stuff in this file that can eventually get combined with the approach in https://github.com/ccmmf/workflows/pull/19/files#diff-005e9efe810f71308274e412c56e3aaa2b21e5fb22ecdfa18fee44c8832e354c .  For this PR I'll assume the restart workflow only cares that its artifacts exist in the right outfolders.

workflows/sipnet-restart-workflow/02-prepare-settings.R
```diff
@@ -0,0 +1,141 @@
...
+  purrr::keep(\(x) x$event_type == "harvest") |>
+  purrr::map_chr("date") |>
+  as.Date()
+end_date <- max(harvest_dates)
+
+met_dir <- sprintf(
+  "%sN_%sW",
+  format(round(abs(site_lat) * 2) / 2, nsmall = 0, drop0trailing = TRUE),
+  format(round(abs(site_lon) * 2) / 2, nsmall = 0, drop0trailing = TRUE)
+)
```
infotroph, line 56: Huh! I've been using 

```
paste0(
      ((lat + 0.25) %/% 0.5) * 0.5, "N_",
      ((abs(lon) + 0.25) %/% 0.5) * 0.5, "W"
    )
```

... and had a moment of "oh no, I'm offsetting by a quarter-degree and @ashiklom isn't, one of us must be wrong", but these seem to produce the same result (at least for points in the CA area) 

```
> lats <- runif(1000, 30, 40)
> lons <- runif(1000, -122, -119)
> lats <- runif(1000, 30, 40)
> as_cell <- function(site_lat, site_lon) sprintf(
+   "%sN_%sW",
+   format(round(abs(site_lat) * 2) / 2, nsmall = 0, drop0trailing = TRUE),
+   format(round(abs(site_lon) * 2) / 2, nsmall = 0, drop0trailing = TRUE)
+ )
> ckb_cell <- function(lat, lon) paste0(
+       ((lat + 0.25) %/% 0.5) * 0.5, "N_",
+       ((abs(lon) + 0.25) %/% 0.5) * 0.5, "W"
+     )
> all.equal(as_cell(lats, lons), ckb_cell(lats, lons))
[1] TRUE
```

workflows/sipnet-restart-workflow/02-prepare-settings.R
infotroph: Yeah, this is a necessary kludge for now. Post-May I'd love to replumb so this could be `<variables><v1>LAI</v1><v2>SoilMoist</v2>...</variables>` -- I think this block is one of the few remaining places a repeated name is required.

workflows/sipnet-restart-workflow/03-run-sipnet.R
```diff
@@ -0,0 +1,26 @@
...
+)
+write.csv(sens_design$X, file.path(settings_raw$outdir, "input_design.csv"))
+
+settings <- PEcAn.workflow::runModule.run.write.configs(
+  settings_raw,
+  input_design = sens_design$X
+)
+
+source("workflows/sipnet-restart-workflow/utils.R")
+jobfiles <- write_segmented_configs.SIPNET(settings, sens_design$X)
```
infotroph: 🎉

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,229 @@
...
+    path_idx <- path_nums[[input]]
+    all_paths <- settings$run$inputs[[input]]$path
+    if (path_idx > length(all_paths)) {
+      PEcAn.logger::logger.severe("No path at input ", sQuote(input), " index ", path_idx)
+    }
+    settings$run$inputs[[input]]$path <- all_paths[[path_idx]]
+    # If we define a list of `source`s, also try to subset that (if the lengths
+    # match). This is especially useful for processing events (because we store
+    # the original JSON path in the `source`).
+    if (!is.list(settings$run$inputs[[input]]$source)) {
```
infotroph: I don't think I've seen input lists that define a `source` key -- is that new here?

workflows/sipnet-restart-workflow/utils.R
infotroph: Eh, this seems like a place to fall through to something that runs instead of complaining. In big runs there _will_ be data issues and we usually don't want to kill the run over anything that only affects one site.

(As a separate issue, I think there at least 4 of these lookups floating around the various different project repos, but unifying those can happen after everything runs)

workflows/sipnet-restart-workflow/utils.R
ashiklom: I think `lead(start_date - 1)` is correct and what we want here.

```r
run_end <- as.Date("2023-09-18")
crop_cycles <- tibble::tibble(date = as.Date(c("2021-01-01", "2021-02-01", "2023-01-01")))
crop_cycles |>
  dplyr::rename(start_date = "date") |>
  dplyr::mutate(
    end_date = dplyr::lead(.data$start_date - 1, default = run_end),
  )
# # A tibble: 3 × 2
#   start_date end_date  
#   <date>     <date>    
# 1 2021-01-01 2021-01-31
# 2 2021-02-01 2022-12-31
# 3 2023-01-01 2023-09-18
```

modules/data.land/tests/testthat/test-events_to_crop_cycle_starts.R
ashiklom: Addressed in 8ab19bdcbe7a3da0ec949f6755abea2b2c860f78.

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Addressed in 7f03d0464d79b27548431c1b3b1257c5d6156090

models/sipnet/R/write.configs.SIPNET.R
ashiklom: Addressed in 10eec01d741b07d7cca92a868c15d04f26e47639

models/sipnet/R/combine_sipnet_out.R
ashiklom: Addressed in c31c63bc1806203ae17316d5076d0d325f19aab1

modules/data.land/R/events_to_crop_cycle_starts.R
ashiklom: Addressed in d6555db706cc7eac35dd613982e2e4f06ca209b0.

infotroph (commented):
Leaving comments from my first in-browser readthrough; I'll switch next to trying to run it locally and will report back.

pixi.lock
ashiklom: [I think it should](https://github.com/orgs/ccmmf/discussions/157) :wink: 

But agreed that this isn't needed. Removed in 2173db96bcadc2a8d223a492f24db131fb0efcad.

models/sipnet/R/combine_sipnet_out.R
ashiklom: Addressed in 11fa8634a2f10c3eaef994de7f251c8dadd2cb59.

modules/data.land/R/events_to_crop_cycle_starts.R
ashiklom: @infotroph My addition doesn't change the function's behavior, but it does add a warning if it detects sequential harvests or plantings. But I can remove it if you think that check is misplaced.

modules/data.land/tests/testthat/test-events_to_crop_cycle_starts.R
ashiklom: Addressed in 8ab19bdcbe7a3da0ec949f6755abea2b2c860f78

workflows/sipnet-restart-workflow/utils.R
ashiklom: Better logic implemented here dc4e8bbbefe15dba9495c31b10c4c60382b48c75. The new implementation supports spin-up before first planting (add another segment before the first planting) and start date _after_ first planting (in that case, clip the segments to start on the start_date).

workflows/sipnet-restart-workflow/utils.R
ashiklom: Added a TODO comment in 2ce339ec8ff96a55d740b02625c0900405461108.

workflows/sipnet-restart-workflow/utils.R
ashiklom: Added boolean `force_rerun` argument in 3843f1220eaaadbf857b6f0d972a8445fefe22ff.

workflows/sipnet-restart-workflow/02-prepare-settings.R
ashiklom: Resolving as "won't fix here". Something to address in the future.

workflows/sipnet-restart-workflow/config.yml
ashiklom: Addressed in 275678608ecf60122baa5dd8b64e906c33b9e3dd.

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: This is unchanged from the previous code. I don't use the dates in the filenames, but I'm not sure if other code (SDA?) does. Suggest leaving as is since this doesn't break anything in our pipeline while changing it _could_ break someone else's pipeline.

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Addressed here: 53f251626029634d6a20f887fc2cf82218152347

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Addressed here: 53f251626029634d6a20f887fc2cf82218152347

models/sipnet/R/write.configs.SIPNET.R
ashiklom: Addressed here 3ea3b2bb3e001ef2d76746b3cd3507ffd8479726

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Per the team meeting today, keeping this as is.

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Downgraded to debug here: 3ab8def831f640b1c23ec838b0c157682a68c190

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Downgraded to debug and added `tz = "UTC"` here: 3ab8def831f640b1c23ec838b0c157682a68c190

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Addressed in 691e7d5c3636cb55fb01d102e3e0eee8eca43f43

modules/uncertainty/R/generate_joint_ensemble_design.R
ashiklom: Reverted in 7456de0594e22708a19cec6382d0477496d8cd30

modules/data.land/R/events_to_crop_cycle_starts.R
infotroph: I really do think it's misplaced. In addition to my argument above that this is the wrong place to enforce it, my local test runs are also complaining about annual orchard pruning and multiple hay harvests, both of which are real and common management patterns that create multiple harvests per planting.

Also the current implementation lumps all sites together, but I realize that'd be fixable with `.by = "site_id"`

workflows/sipnet-restart-workflow/01-prepare-events.R
ashiklom: Extended to 12/31 in f061e0923c9385f02283371c00e7f9cd3eecfc50.

models/sipnet/R/split_inputs.SIPNET.R
ashiklom: Yes. But the behavior was internally inconsistent within my split_inputs (:grimacing:), so I reverted back to the current behavior of `>= start.date` and `< end.date`, and made sure this was the behavior for both events and met.

0b3ac45e17d0d1943505f6065dda716c1bfad20a

models/sipnet/R/combine_sipnet_out.R
ashiklom: image

Since you apparently like living on the edge :wink:, changed this to your version in 6554a6b06340e3081d1f1053226b02ff4f00f5f6

workflows/sipnet-restart-workflow/01-prepare-events.R
ashiklom: Note that these scripts were more intended to be helpers/test functions for setting up the inputs for the workflow. The expectation is that folks will provide their own `settings.xml` files, etc., and then just use the workhorse functions from this PR.

That said, I agree that it's useful to test full year chunks, so I extended this to start Jan 1 of the year of first planting and end Dec 31 of the year of last planting in d5c359e264fa4aa84accb8269b45bfd3d25cbaaa.

workflows/sipnet-restart-workflow/01-prepare-events.R
ashiklom: > For this PR I'll assume the restart workflow only cares that its artifacts exist in the right outfolders.

Correct. Per my other comment, the prepare events and prepare settings scripts in this PR are more for testing/demonstration purposes for a single-run simulation. I fully expect folks to write their own `settings.xml` files, etc. and only use the capabilities as demonstrated in `03-run-sipnet.R`.

ashiklom:
OK, I think I've resolved all of @divine7022 and @infotroph comments. I also resolved a significant bug with `write_events.SIPNET` (namely, the events were always being written out in the same order as `events.json`, _not_ sorted by date as intended), and did a bit of other miscellaneous cleanup.

Note that this PR sits on top of https://github.com/PecanProject/pecan/pull/3836 and https://github.com/PecanProject/pecan/pull/3828 --- if we merge this, we can probably just close those.

@dlebauer :

> Great work! Excited to have this working! 🥳
> 
> Before merging, please create a ticket to cover the refactoring plan we discussed last week.
> 
> From what I recall, the plan was approximately:
> 
> Goal: refactor segmented runs so that they integrate more cleanly with standard start_model_runs workflows.
> 
> Specifically:
> 
>     * write.configs generates config files
> 
>     * sites and ensembles run in parallel via start_model_runs
> 
>     * This will require modifying or replacing the top‑level job.sh so that:
>       
>       * Segments run sequentially within a site, and
>       * Sites still run in parallel via start_model_runs.

All of this already works in this PR. The new `write_segmented_configs.SIPNET` does a bunch of config preparation behind the scenes and then modifies the `job.sh` files associated with each ensemble member such that, instead of running SIPNET directly, they run the segments in sequence and then post-process the results. This means that the actual interface here is very "PEcAn-ic":

```r
settings <- PEcAn.workflow::runModule.run.write.configs(
  settings_raw,
  input_design = sens_design$X
)

source("workflows/sipnet-restart-workflow/utils.R")
jobfiles <- write_segmented_configs.SIPNET(settings, sens_design$X, force_rerun = TRUE)

PEcAn.workflow::runModule_start_model_runs(settings)
```

Whatever method the user has specified for running `job.sh` files in `settings` will continue to work here without modification.

modules/data.land/R/events_to_crop_cycle_starts.R
ashiklom: OK --- I removed the check in 77022811302c14f63557e90f61634f344bddd0af.

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,279 @@
...
+crop2pft_example <- function(crop_code) {
+  cls <- substr(crop_code, 1, 1)
+  dplyr::case_when(
+    crop_code == "P1" ~ "annual_crop",
+    crop_code == "G2" ~ "annual_crop",
+    cls == "D" ~ "temperate.deciduous",
+    cls == "F" ~ "annual_crop",
+    cls == "G" ~ "grass",
+    cls == "P" ~ "grass",
+    cls == "R" ~ "grass",
```
infotroph: Yes, this table is still temporary, but needed this one for the sites I grabbed
```suggestion
    cls == "R" ~ "grass",
    cls == "T" ~ "annual_crop",
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,279 @@
...
+    TRUE ~ "UNKNOWN_PFT"
+  )
+}
+
+write_segmented_configs.SIPNET <- function(settings, input_design = NULL, ...) {
+  manifest_file <- file.path(settings$outdir, "runs_manifest.csv")
+  if (!file.exists(manifest_file)) {
+    PEcAn.logger::logger.severe("Could not find manifest file: ", manifest_file)
+  }
+  inputs_runs <- read.csv(manifest_file)
```
infotroph:
```suggestion
  inputs_runs <- read.csv(manifest_file) |>
    dplyr::filter(.data$site_id == settings$run$site$id) |>
    # TODO the manifest should probably report these already...
    dplyr::mutate(
      ens_num = .data$run_id |>
        stringr::str_extract("ENS-(\\d+)", group = 1) |>
        as.integer()
    )
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,279 @@
...
+  if (!is.null(input_design)) {
+    inputs_runs <- cbind.data.frame(inputs_runs, input_design)
```
infotroph: In a multi-site run the cbind winds up becoming a cross join. Need to align explicitly by ensemble number
```suggestion
    inputs_runs <- inputs_runs |>
      dplyr::left_join(
        input_design |> tibble::rowid_to_column("ens_num"),
        by = "ens_num",
        relationship = "many-to-one")
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,279 @@
...
+  }
+  invisible(new_jobfiles)
+}
+
+segment_dataframe <- function(run_settings) {
+  events_json <- run_settings$run$inputs$events$source
+  stopifnot(file.exists(events_json))
+
+  crop_cycles <- PEcAn.data.land::events_to_crop_cycle_starts(events_json) |>
+    dplyr::ungroup()
```
infotroph:
```suggestion
    dplyr::filter(.data$site_id == run_settings$run$site$id) |>
    dplyr::ungroup()
```

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,279 @@
...
+  ensemble_samples <- PEcAn.utils::load_local(ens_samples_file)[["ens.samples"]]
+  i_param <- run_row[["param"]] %||% 1
+  run_traits <- lapply(ensemble_samples, \(dat) dat[i_param, ])
+
+  segments <- segment_dataframe(run_settings) |>
+    dplyr::mutate(
+      pft = crop2pft(.data$crop_code),
+      segment_dir = file.path(segment_rootdir, sprintf("segment_%s", .data$segment_id))
+    )
+
```
infotroph: Seems useful to retain this information for diagnostics. I didn't think carefully about format or location, though -- counterproposals welcome
```suggestion
  write.csv(segments, file = file.path(run_dir, "segments.csv"), row.names = FALSE)

```

workflows/sipnet-restart-workflow/03-run-sipnet.R
infotroph: Just leaving a breadcrumb hint for the next person
```suggestion
jobfiles <- write_segmented_configs.SIPNET(settings, sens_design$X)
# Note: If running a multi-site workflow, use:
# jobfiles <- papply(settings, \(s) write_segmented_configs.SIPNET(s, sens_design$X))
```

infotroph (approved):
This looks great! I was able to run through 3 ensembled sites in a lightly modified MAGiC ensemble workflow using just utils.R with the changes noted below. I think this is ready to merge. Next iteration can be to decide which package(s) to drop the functions into.

workflows/sipnet-restart-workflow/utils.R
```diff
@@ -0,0 +1,279 @@
...
+        input, n_path, n_source
+      ))
+      next
+    }
+    settings$run$inputs[[input]]$source <- all_source_paths[[path_idx]]
+  }
+  settings
+}
+
+# TODO: We need a better, consistent implementation of this. However, this is
```
dlebauer, line 42: I thought this table was with the landiq code. If not, I think @infotroph , @sarahkanee , and I have each implemented one or more versions of this mapping. And there is one in https://github.com/ccmmf/cadwr-landuse/blob/eacec3ba471b80ed3fc9404d300767f5b12a3654/data/CARB_PFTs_table.csv#L66 ... that doesn't meet the need here, but we can add columns (probably can have multiple pft cols representing different levels of splits, or could have a many-many table)

dlebauer (approved):

PR #3920, opened 2026-04-07
support tillage incorporation fraction in events pipeline

divine7022:
## Description
ryt now sipnet's tillage only touches decomposition rates, but actual tillage also physically mixes litter into soil; so added incorporationFraction on sipnet side (see https://github.com/PecanProject/sipnet/pull/312), so this wires it up through pecan events pipeline. 
writer picks up incorporation_frac_0to1 from json when it's there and tacks it on as second param in events.in. If it's not there, nothing changes, old json files still produce the same 1 param till lines they always did

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.land/inst/events_schema_v0.1.0.json
+++ b/modules/data.land/inst/events_schema_v0.1.0.json
@@ -53,6 +53,7 @@
               "no3_n_kg_m2": { "type": "number", "minimum": 0 },
 
               "tillage_eff_0to1": { "type": "number", "minimum": 0 },
+              "incorporation_frac_0to1": { "type": "number", "minimum": 0, "maximum": 1 },
               "intensity_category": { "type": "string" },
               "depth_m": { "type": "number", "minimum": 0 }
             },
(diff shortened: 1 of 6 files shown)


mdietze (approved):

modules/data.land/inst/events_schema_v0.1.0.json
```diff
@@ -53,6 +53,7 @@
               "no3_n_kg_m2": { "type": "number", "minimum": 0 },
 
               "tillage_eff_0to1": { "type": "number", "minimum": 0 },
+              "incorporation_frac_0to1": { "type": "number", "minimum": 0, "maximum": 1 },
```
mdietze, line 56: incorporation fraction should be a parameter not an event

mdietze (changes requested):

dlebauer (commented):
@divine7022 I believe there was a misunderstanding when we spoke yesterday. 

I'll close this for now, and we can get back to it if indicated by calibration and validation.

When you identified an issue with tillage pumping too much C into SOC, thus having the unrealistic effect of protecting it, I didn't mean that we should add a new parameter.

What I meant by "reduce the flux" was to change the fracLitterRespired which partitions between respiration and transfer to SOC.

Considering that we haven't started formal calibration, we are mostly trying to get sensible answers at this point, that we can later formally calibrate.

This does highlight the urgency of formal calibration. But for now let's manually tune a few knobs.

Perhaps first, tillage intensity. Then, fracLitterRespired.

_Originally posted by @dlebauer in https://github.com/PecanProject/sipnet/pull/312#issuecomment-4200289547_

PR #3924, opened 2026-04-10
write.events: Avoid hanging on event files with large number of validation errors

infotroph:
## Description

Noticed on a file that returned more than 150k(!) errors, for which `json_validate` does finish after about 25 seconds but then `validate_events_json` hangs forever trying to paste all the errors together into a single error message.

Attacked this in two places:

(1) inside write.events.SIPNET, call `validate_events_json` with verbose = FALSE for a quick result and advise users to call `validate_events_json` directly for more details.
(2) inside validate_events_json, only print the first n (default 50) errors and return the full error object as an attribute (same as json_validate does) for anyone who wants the thousands of gory details.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/data.land/R/validate_events.R
+++ b/modules/data.land/R/validate_events.R
@@ -11,8 +11,11 @@
 #'
 #' @param events_json character. Path to the JSON file to validate.
 #' @param verbose logical. When `TRUE`, include detailed AJV messages on error.
+#' @param max_errs integer. Print only this many validation errors.
+#'  To see the rest, use the `errors` attribute of the return value.
 #'
-#' @return Logical TRUE if valid, FALSE if invalid.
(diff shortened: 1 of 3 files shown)


dlebauer (approved):

modules/data.land/R/validate_events.R
```diff
@@ -44,11 +47,13 @@ validate_events_json <- function(events_json, verbose = TRUE) {
   detail <- if (is.null(errs)) {
     "<no details>"
   } else {
+    errs <- head(errs, max_errs)
```
infotroph:
```suggestion
    errs <- utils::head(errs, max_errs)
```

PR #3928, opened 2026-04-11
fix(utils): use units::ud_convert directly for non-difftime conversions

### LINKED ISSUE #3927: use units::ud_convert directly in PEcAn.utils::ud_convert?
Noted by @dlebauer in https://github.com/PecanProject/pecan/pull/3925#pullrequestreview-4092876746:

>  there are a lot of calls in the PEcAn version of `ud_convert` (`ud_are_convertible`, `as_units`, `set_units`, `drop_units`) while the units package directly calls a C function. So I checked out the penalty, and found that the `units` implementation is about 20x faster and uses infinitely less memory.

```r
bench::mark(
  pecan = PEcAn.utils::ud_convert(42, "m", "km"),
  units = units::ud_convert(42, "m", "km"),
  iterations = 500
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 pecan      142.35µs 151.56µs     6445.    6.44MB     52.0
#> 2 units        6.56µs   7.58µs   114353.        0B    229.
```

Consider replacing the `as_units`/`set_units` dance inside PEcAn.utils::ud_convert with a single call to `units::ud_convert`?  This would (probably) capture the speed gains without requiring all other packages to add a direct {units} dependency. 

One piece to consider: PEcAn.utils::ud_convert has a special case for `difftime` object that I don't _think_ udunits::ud_convert knows how to handle. My impression is that's a little-used feature that could be retained without losing the everyday speedup, but I haven't checked in detail.

anshul23102:
## Description

For non-difftime inputs, replace the slow `set_units/set_units/drop_units` chain with a direct call to `units::ud_convert`. The difftime path is unchanged.

Closes #3927

## Motivation and Context

`units::ud_convert` is ~10x faster than the original `set_units` chain for scalar inputs. An affine factor/offset approach was attempted for large-vector speedup but caused floating-point precision errors in datetime conversions and was reverted.

## Benchmark (scalar, real usage pattern)

| approach | median |
|---|---|
| original (`set_units` chain) | ~183µs |
| this PR (`units::ud_convert`) | ~18µs |

## Types of changes
- [x] Bug fix / performance improvement (non-breaking)

## Checklist
- [x] All new and existing tests passed
- [x] Added tests for vector inputs, NA passthrough, and integer input
- [x] I agree that PEcAn Project may distribute my contribution under BSD 3-clause license
- [x] I have read the CONTRIBUTING document


infotroph:
Thanks for the PR! Can you please tell us how you tested this and whether the speed difference in a real workflow was comparable to @dlebauer's benchmark result?

anshul23102:
Tested locally with `bench::mark()` (500 iterations):

| expression | min | median | itr/sec | mem_alloc |
|---|---|---|---|---|
| old | 162.89µs | 168.31µs | 5780 | 6.37MB |
| new | 8.73µs | 9.27µs | 93252 | 0B |

The new approach is ~18x faster with zero memory allocation, consistent with @dlebauer's benchmark in #3927. The `difftime` path is unchanged and all existing tests pass.

infotroph:
Well yes, repeating the same test case is going to give the same results. That's why I said "real workflow": Inside PEcAn `ud_convert` gets called in many different ways that differ greatly in datatype, units, vector length, and execution context; one test of a scalar input is promising but not indicative of the PEcAn-wide result. I'm particularly interested in whether this produces detectable speedups in the various model-specific `met2model`, `write.config`, and `model2netcdf` functions -- most models use ud_convert extensively for those.

Even before we get to testing on a PEcAn workflow, we can observe in benchmark results that the memory allocation difference is only seen the first time you call it in a session, and the headline "20x" number only holds for very simple and _small_ inputs. Converting a vector of 10 numbers instead of a scalar cuts `units`'s advantage from ~18x to ~6x, and with a vector of 1000 I'd argue the difference isn't even consistently detectable.

<details>

```
> bench::mark(
+  pecan = PEcAn.utils::ud_convert(1, "g", "kg"),
+  units = units::ud_convert(1, "g", "kg"),
+  iterations=1000)
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    memory                 time               gc                  
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    <list>                 <list>             <list>              
1 pecan      175.73µs  179.8µs     5461.    6.82MB     38.5   993     7   181.85ms <dbl [1]> <Rprofmem [3,940 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
2 units        8.61µs    9.1µs   107270.        0B      0    1000     0     9.32ms <dbl [1]> <Rprofmem [0 × 3]>     <bench_tm [1,000]> <tibble [1,000 × 3]>
> bench::mark(
+  pecan = PEcAn.utils::ud_convert(1, "g", "kg"),
+  units = units::ud_convert(1, "g", "kg"),
+  iterations=1000)
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result    memory             time               gc                  
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>    <list>             <list>             <list>              
1 pecan      175.73µs 194.71µs     4872.        0B     39.3   992     8    203.6ms <dbl [1]> <Rprofmem [0 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
2 units        8.98µs   9.59µs   101483.        0B      0    1000     0     9.85ms <dbl [1]> <Rprofmem [0 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
> x <- rnorm(10)
> bench::mark(
+  pecan = PEcAn.utils::ud_convert(x, "g", "kg"),
+  units = units::ud_convert(x, "g", "kg"),
+  iterations=1000)
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result     memory             time               gc                  
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>     <list>             <list>             <list>              
1 pecan         193µs  201.5µs     4385.        0B     48.8   989    11    225.5ms <dbl [10]> <Rprofmem [0 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
2 units          31µs   33.3µs    25921.        0B     25.9   999     1     38.5ms <dbl [10]> <Rprofmem [0 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
> bench::mark(
+  pecan = PEcAn.utils::ud_convert(x, "g", "kg"),
+  units = units::ud_convert(x, "g", "kg"),
+  iterations=1000)
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result     memory             time               gc                  
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>     <list>             <list>             <list>              
1 pecan       196.3µs    213µs     3867.        0B     43.0   989    11      256ms <dbl [10]> <Rprofmem [0 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
2 units        30.8µs   33.4µs    25604.        0B     51.3   998     2       39ms <dbl [10]> <Rprofmem [0 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
> x <- rnorm(1e3)
> bench::mark(
+  pecan = PEcAn.utils::ud_convert(x, "g", "kg"),
+  units = units::ud_convert(x, "g", "kg"),
+  iterations=1000)
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result        memory              time               gc                  
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>        <list>              <list>             <list>              
1 pecan        2.67ms   3.45ms      278.    31.7KB     42.0   869   131      3.12s <dbl [1,000]> <Rprofmem [36 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
2 units        2.52ms   3.44ms      296.    31.7KB     40.8   879   121      2.96s <dbl [1,000]> <Rprofmem [13 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
> bench::mark(pecan = PEcAn.utils::ud_convert(x, "g", "kg"), units = units::ud_convert(x, "g", "kg"), iterations=1000)
# A tibble: 2 × 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result        memory             time               gc                  
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>        <list>             <list>             <list>              
1 pecan        2.69ms   3.19ms      279.    31.7KB     41.8   870   130      3.11s <dbl [1,000]> <Rprofmem [5 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
2 units        2.49ms   3.36ms      296.    31.7KB     41.2   878   122      2.96s <dbl [1,000]> <Rprofmem [5 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
```

</details>

That said, we might still want to make this change! We use ud_convert so many places in PEcAn that even a context-dependent speedup could still pay off if it eases a bottleneck  in an important place. 

But the flip side is that because we rely so heavily on ud_convert in so many places, the bar is high for changes: We want to be VERY sure they won't break anything. I know we have some existing test coverage of ud_convert, but it's worth considering if there are any additional test cases needed to ensure this change didn't miss any corner cases.

anshul23102:
I investigated the real call sites in `met2model.SIPNET` and `met2model.ED2` and found `ud_convert` is called on full NetCDF arrays (e.g. hourly Tair, n=8760). I tried an affine factor/offset approach (compute scale+offset once,
apply with base R arithmetic) which benchmarked ~750x faster for large vectors - but it caused floating-point precision errors in `datetime2cf` due to catastrophic cancellation with large time offsets like `"seconds since 1970-01-01"`. Reverted to `units::ud_convert` directly.

So the honest summary is: this PR gives a real speedup (~10x) for scalar calls but no meaningful gain for large vectors. The CI failure on R 4.2 (`test.cf2date`) appears pre-existing - the same test passes locally on R 4.5 with both the original and new code, and the failure is unrelated to our changes.

Happy to close this if the scalar-only speedup isn't worth it, or keep it open if you think it's still a net improvement.

Yeligay8 (commented):
ud_convert() now correctly processes vector inputs instead of assuming scalar values.
Added test:
expect_equal(ud_convert(c(0, 10, 100), "degC", "K"), c(273.15, 283.15, 373.15))

Recursive Conversion Logic:
return(units::ud_convert(as.numeric(x), u1, u2))
This ensures consistent handling when units do not match u1.

Function now preserves NA values during conversion.
Added test:
expect_equal(result, c(100, NA, 300))

Support for Integer Inputs: Ensures integer vectors are handled correctly.
Added test:
expect_equal(ud_convert(1L, "m", "cm"), 100)

Explicit test ensures output length matches input:
expect_length(result, 1000)

Removed code from original code:
Direct Unit Assignment to x1
x1 <- units::set_units(as.numeric(x), value = u1, mode = "standard")

Replaced with recursive call to ud_convert() for cleaner handling.
Intermediate Variable Handling
Old approach:
x2 <- units::set_units(x1, value = u2, mode = "standard")
units::drop_units(x2)
Refactored into a more streamlined flow while keeping:
x2 <- units::set_units(x1, value = u2, mode = "standard")
units::drop_units(x2)
Removes redundant steps and improves readability.
put units

anshul23102:
Thanks for the review @Yeligay8!

Just a small clarification: the call `return(units::ud_convert(as.numeric(x), u1, u2))` is not recursive - it calls `units::ud_convert` from the `units` package directly, not `PEcAn.utils::ud_convert` itself.

The `set_units/drop_units` code is still present but only for the **difftime path** (unchanged). The non-difftime path now bypasses it entirely with the direct `units::ud_convert` call.

infotroph:
Thanks for working on this. It's still on my list to look at in more detail, but short answer is 

- Yes, I think it's worth getting merged even if it's not the slam-dunk speedup we hoped
- On looking more, I think the difftime branch _may_ be able to be reduced to something like `if (inherits(x, "difftime")) { difftime_units <- base::units(x); if(u1 != difftime_units) { warn(...); u1 <- difftime_units }}`, but do not believe me until we've tested more extensively.

base/utils/tests/testthat/test-ud_convert.R
```diff
@@ -53,4 +53,21 @@ test_that("model-specific flux conversions", {
...
+})
+
+test_that("ud_convert handles NA in vectors", {
+  result <- ud_convert(c(1, NA, 3), "m", "cm")
+  expect_equal(result, c(100, NA, 300))
+})
+
+test_that("ud_convert handles integer input", {
+  expect_equal(ud_convert(1L, "m", "cm"), 100)
 })
```
infotroph:
```suggestion
  expect_equal(ud_convert(1L, "m", "cm"), 100)
})

test_that("ud_convert difftime", {
  x <- as.difftime(c(1, 7, 14), units = "days")
  expect_equal(ud_convert(x, "days", "weeks"), 1:3)
  expect_warning(wk_res <- ud_convert(x, "weeks", "days"), "using days instead")
  expect_equal(wk_res, x)  
})

```

anshul23102:
Great to hear, thanks! 

Understood on the difftime simplification - using `base::units(x)` to extract  the units and falling through to `units::ud_convert` for both paths would make  the function much cleaner. Should I implement that in this PR, or would you prefer to merge what's here and handle the difftime path in a follow-up once it's been tested more thoroughly?

dlebauer (approved):
Looks good to me. The change is small and targeted, preserves the existing difftime path, and adds useful tests for vectors, NA, and integer inputs. Thank you! Even though the performance gain seem most meaningful for scalar conversions, this PR clearly improves the codebase. Thanks for the tests!

PR #3930, opened 2026-04-14
Sipnet: Add waterDrainFrac to template

infotroph:
To support Sipnet's freshly-merged soil flooding capabilities


divine7022 (approved):
thanks, LGTM!

PR #3934, opened 2026-04-17
data prep and post bias correction for NEE and LE

Amedieus:
bias correct the results from SDA output of NEE and LE

### DIFF
--- a/modules/assim.sequential/inst/bias_correction.R
+++ b/modules/assim.sequential/inst/bias_correction.R
@@ -1,0 +1,747 @@
+## Load packages
+library(data.table)
+library(dplyr)
+library(xgboost)
+library(parallel)
+library(doParallel)
+library(caret)
+
+##' @title parse_bcm_args
+##' @name parse_bcm_args
+##' @author Yang Gu
+##'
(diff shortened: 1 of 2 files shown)


modules/assim.sequential/inst/bias_correction.R
```diff
@@ -0,0 +1,747 @@
+## Load packages
```
mdietze, line 1: This file just contains functions, which I think would be better migrated to the R folder instead of the inst folder, in which case you'll need to drop the library calls, make namespaces explicit, and compile + commit the Rd man files for each function

modules/assim.sequential/inst/data_prep.R
```diff
@@ -0,0 +1,633 @@
+## Load packages
```
mdietze, line 1: same points about migrating to R folder. For both files, I'd `@export` the minimal set of functions (I'm assuming many of these functions are exclusively being called by other functions internally). Since there's no "master script" here that runs your actual debiasing, I'd either add that (either as an inst script or uncompiled vignette) or provide an unrun example for the few high-level functions that are being exported.

PR #3936, opened 2026-04-20
Fix posterior pagination URL and workflow zip download logic

omkarrr2533:
Found a couple of functional bugs while going through the API code:

> error in posteriors.R: In searchPosteriors, the prev_page URL was pointing to the /api/workflows endpoint instead of /api/posteriors. I've updated this to use the correct path.

> Broken zip download in workflows.R: The getWorkflowFilesAsZip function was returning the filename string "output.zip" rather than the actual binary content. I updated it to use readBin() so the API correctly serves the file, and added file.remove() to clean up the temporary zip after it's sent.

### DIFF
--- a/apps/api/R/posteriors.R
+++ b/apps/api/R/posteriors.R
@@ -104,7 +104,7 @@ searchPosteriors <- function(req, pft_id = NA, host_id = NA,
       result$prev_page <- paste0(
         req$rook.url_scheme, "://",
         req$HTTP_HOST,
-        "/api/workflows",
+        "/api/posteriors",
         req$PATH_INFO,
         substr(req$QUERY_STRING,
                0,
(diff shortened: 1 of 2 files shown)


mdietze:
Have you tested that the API fix works correctly and now returns the expected zip file? Is there anything you can show (e.g., a screenshot) to document this?

omkarrr2533:
@mdietze yes, tested both!

Pagination fix = prev_page now correctly points to /api/posteriors instead of /api/workflows 

Screenshot 2026-04-22 073722


Zip fix — was returning the string "output.zip" before, now readBin() sends the actual binary. Hit the endpoint with curl, got back application/octet-stream, and unzip -l confirms pecan.xml is inside 

Screenshot 2026-04-22 073528

omkarrr2533:
also spotted something unrelated the pagination URLs have limit duplicated twice (?limit=10&offset=20&limit=10). looks like an existing bug in the shared pagination helper, not introduced by this PR. happy to open a separate issue for it if useful. 👍

mdietze (approved):
Yes, please submit a PR on the multiple definition issue

Issue #3937, opened 2026-04-20
Fix logical operator, type check, and qqline misuse across db and benchmark modules


omkarrr2533:
Hey, was going through some files and spotted a few bugs worth fixing.

1st one is in dbfile.move() around line 330 of base/db/R/dbfiles.R — the file type check uses | instead of &&, which makes the condition always TRUE. So both "clim" and "nc" immediately hit the error block, basically making the function unusable for its intended inputs.
r# current

> if (file.type != "clim" | file.type != "nc") {

2nd is in insert.format.vars() around line 40 of base/db/R/insert.format.vars.R — skip defaults to 0 (numeric) but the guard checks is.character(skip), so every default call logs an error. The check should be is.numeric instead.

> rif(!is.character(skip)){
  logger.error("Skip must be of type character")
}

3rd is in metric_lmDiag_plot() around line 30 of modules/benchmark/R/metric_lmDiag_plot.R — stats::qqline() is being passed inside aes() but it's a base R side-effect function that returns NULL, so the Q-Q line just never shows up silently. Fix would be to manually compute the slope and intercept from quantiles and pass them straight to geom_abline().

> rp2 <- p2 + ggplot2::geom_abline(ggplot2::aes(stats::qqline(.data$.stdresid)))
Happy to open a PR with all three fixes if that works!

Planning to open a PR tomorrow with fixes for all three, will reference this issue there.

infotroph:
These seem independent enough that they should be three separate PRs.

omkarrr2533:
Sure, I will create three separate PRs for these issues.

PR #3939, opened 2026-04-21
fix logical operator in dbfile.move type check

omkarrr2533:
dbfile.move was using | instead of && in the file type check, which made the condition always true and caused the error block to fire for both valid inputs. Changed to &&.

Part of #3937


infotroph (approved):

infotroph:
Thanks! Easy fix, though the fact it's stayed unreported this long suggests we should question whether anyone's using dbfile.move at all...

omkarrr2533:
Good point — might be worth a follow-up to check if dbfile.move is actually called anywhere in the codebase or if it needs deprecation.

PR #3940, opened 2026-04-21
fix skip type check in insert.format.vars

omkarrr2533:
skip defaults to 0 (numeric) but the guard was calling is.character(skip), so every default call logged an error. Changed the check to is.numeric and updated the message to match.

Part of #3937


base/db/R/insert.format.vars.R
```diff
@@ -57,9 +57,9 @@ insert.format.vars <- function(con, format_name, mimetype_id, notes = NULL, head
   }
 
    #Test if skip is an integer
-  if(!is.character(skip)){
+  if(!is.numeric(skip)){
   PEcAn.logger::logger.error(
-    "Skip must be of type character"
+    "Skip must be of type numeric"
```
infotroph: Confusingly, I think this line was correct before -- `skip` ultimately gets passed to the BetyDB `formats` table, which declares the column as `character varying(255) DEFAULT ''`. I don't know why the schema didn't use int -- that decision predates my involvement. @dlebauer?

Maybe It'd be simplest to coerce the argument to character instead of complaining?

infotroph (changes requested):

base/db/R/insert.format.vars.R
omkarrr2533: Ya, I missed that the column is typed as character varying in BetyDB. Replaced the type guard with a silent as.character(skip) coercion instead, so numeric input just works without erroring. Thanks for the context.

omkarrr2533:
I've replaced the type guard with a silent as.character(skip) 
coercion as you suggested, and rebased on latest develop. Could you take 
another look when you get a chance? Thanks!

infotroph:
Sorry for the wait on this one -- the diff I'm seeing still shows a test and warning rather than silent conversion. Can you check that you've pushed all changes?

omkarrr2533:
You're right, sorry — looks like the coercion change didn't make it into the last push. Just pushed the actual fix now, the type guard is gone and skip is coerced silently. Let me know if anything else needs tweaking. 👍

infotroph (approved):

PR #3941, opened 2026-04-21
fix qqline misuse in metric_lmDiag_plot

### LINKED ISSUE #3937: Fix logical operator, type check, and qqline misuse across db and benchmark modules

omkarrr2533:
stats::qqline() is a base R side-effect function that returns NULL — wrapping it in aes() means the Q-Q line never shows up. Replaced with an explicit slope/intercept calculation passed directly to geom_abline().

Fix : #3937

### DIFF
--- a/modules/benchmark/R/metric_lmDiag_plot.R
+++ b/modules/benchmark/R/metric_lmDiag_plot.R
@@ -21,10 +21,13 @@ metric_lmDiag_plot <- function(metric_dat, var, filename = NA, draw.plot = FALSE
   p1 <- p1 + ggplot2::ggtitle("Residual vs Fitted Plot") 
   p1 <- p1 + ggplot2::theme_bw()
   
-  p2 <- ggplot2::ggplot(fit, ggplot2::aes(stats::qqnorm(.data$.stdresid)[[1]], .data$.stdresid)) 
-  p2 <- p2 + ggplot2::geom_point(na.rm = TRUE)
-  p2 <- p2 + ggplot2::geom_abline(ggplot2::aes(stats::qqline(.data$.stdresid))) 
-  p2 <- p2 + ggplot2::xlab("Theoretical Quantiles") 
+  p2 <- ggplot2::ggplot(
+  data = data.frame(sample = stats::na.omit(ggplot2::fortify(fit)$.stdresid)),
+  ggplot2::aes(sample = sample)
+  )
+  p2 <- p2 + ggplot2::stat_qq()
+  p2 <- p2 + ggplot2::stat_qq_line()
+  p2 <- p2 + ggplot2::xlab("Theoretical Quantiles")
   p2 <- p2 + ggplot2::ylab("Standardized Residuals")
   p2 <- p2 + ggplot2::ggtitle("Normal Q-Q") 
   p2 <- p2 + ggplot2::theme_bw()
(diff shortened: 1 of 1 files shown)


modules/benchmark/R/metric_lmDiag_plot.R
```diff
@@ -23,7 +23,12 @@ metric_lmDiag_plot <- function(metric_dat, var, filename = NA, draw.plot = FALSE
   
   p2 <- ggplot2::ggplot(fit, ggplot2::aes(stats::qqnorm(.data$.stdresid)[[1]], .data$.stdresid)) 
   p2 <- p2 + ggplot2::geom_point(na.rm = TRUE)
```
infotroph: Calculation from these lines is now discarded -- is that intentional?

omkarrr2533: No, that wasn't intentional those two lines are now dead code since p2 gets fully reassigned right after. Removed them in the latest push. Thanks for catching it! 👍

modules/benchmark/R/metric_lmDiag_plot.R
```diff
@@ -21,10 +21,13 @@ metric_lmDiag_plot <- function(metric_dat, var, filename = NA, draw.plot = FALSE
   p1 <- p1 + ggplot2::ggtitle("Residual vs Fitted Plot") 
   p1 <- p1 + ggplot2::theme_bw()
   
-  p2 <- ggplot2::ggplot(fit, ggplot2::aes(stats::qqnorm(.data$.stdresid)[[1]], .data$.stdresid)) 
-  p2 <- p2 + ggplot2::geom_point(na.rm = TRUE)
-  p2 <- p2 + ggplot2::geom_abline(ggplot2::aes(stats::qqline(.data$.stdresid))) 
-  p2 <- p2 + ggplot2::xlab("Theoretical Quantiles") 
+  p2 <- ggplot2::ggplot(
+  data = data.frame(sample = stats::na.omit(ggplot2::fortify(fit)$.stdresid)),
+  ggplot2::aes(sample = sample)
```
infotroph:
```suggestion
    data = data.frame(sample = stats::na.omit(ggplot2::fortify(fit)$.stdresid)),
    ggplot2::aes(sample = sample)
```

infotroph (approved):
Thanks! This looks good

For the record when I test this locally with ggplot2 4.0.2 I get a deprecation warning from ggplot2::fortify, but I think that was present before this patch as well (because it called fortify on the qqnorm outputs under the hood). We can discuss a fix for that separately.

PR #3945, opened 2026-04-22
landiq crop to pft lookup

divine7022:
## Description
adds a canonical LandIQ  crop to pft  lookup so we stop rolling our own versions of this mapping across downstream.
pitched after @dlebauer comment on https://github.com/PecanProject/pecan/pull/3919/changes/BASE..81e49dc5a0af79c72c1d3d5111cb4ac5ff81722e#r3105651074 asking for one single canonical reference

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


mdietze (commented):
looks good to me. Just want to make sure @sarahkanee, @ashiklom, Ananya, Matt Kim, and Xinyuan see this and can chime in and/or adopt this into their codes

ashiklom (approved):
Looks good to me!

infotroph (changes requested):
As I said in Slack, this is the wrong approach. PFT mappings are not a single constant  and treating them that way goes against the point of PFTs.

I really think what we need here is just a csv in the S3 bucket, not an interlocking set of functions with unit tests designed to prevent edits.

dlebauer:
@infotroph I do not think that objection applies here.

This PR does not claim that PFT mappings are universal or immutable. It provides a canonical default so we can stop maintaining multiple ad hoc versions downstream.

The package already includes datasets such as bsim and cadwr that use different, opinionated crop groupings.

This may be a better fit for a ccmmf repository such as [cadwr-landuse](https://github.com/ccmmf/cadwr-landuse) or workflows instead. But if we expect multiple evolving mappings, I think they should still live under version control so it is clear which version is being used and there is a reliable process for proposing changes.

divine7022:
the dataset here, function plus tests setup was heavier than this problem needs. Closing this PR if no objections and migrating to cadwr-landuse per Slack thread with @dlebauer and @infotroph

for the record:

* mapping lives as CSV in cadwr-landuse, version controlled
* dropping the value pinning tests in migration; at most schema level validation so edits to the mapping do not break anything mechanical
* regarding API that current form look_up_crop_pft("D", "12") is opaque for humans and that passing two parallel vectors (class + subclass) is unsafe under reordering. Migrating to a name based primary form; look_up_pft(crop = c("rice", "cotton", "corn")) keyed on the crop description field, case insensitive and multi row return for names that repeat across class/subclass pairs
* field renames for provenance also land in cadwr-landuse PR, crop_type to class, crop_code to subclass, crop_desc to subclass_name, landiq_cat to class_name, and filename CARB_PFTs_table.csv to cadwr_pfts.csv

PR #3948, opened 2026-04-22
Fix typo in landiq_crop_mapping_codes.tsv

dlebauer:
Trivial fix "subtropiscal" --> "subtropical"


infotroph (approved):
Nice catch! Should there be a change in data/landiq_crop_mapping_codes.rda to accompany this?

dlebauer:
@infotroph good point, done!

PR #3952, opened 2026-04-23
fix: replace 1:length() anti-pattern with seq_along() across core modules

### LINKED ISSUE #3951: bug: replace 1:length() anti-pattern with seq_along() across core modules
## Summary

The `1:length(x)` anti-pattern was found throughout the `PEcAn.assim.sequential` and `PEcAn.data.atmosphere` modules. This construct evaluates to `c(1, 0)` when `length(x) == 0`.

---

### Bug Description

In R, when a vector/list `x` is empty, `length(x)` is `0`. Using `1:length(x)` results in `1:0`, which creates the sequence `c(1, 0)`.

This introduces silent bugs:
1. **Loops:** `for (i in 1:length(x))` will execute twice (for `i=1` and `i=0`) instead of zero times, leading to `NULL` references, out-of-bounds errors, or silent logical corruption.
2. **Matrix Indexing:** In `sda.enkf.R`, expressions like `x[, 1:length(names.y)]` when `names.y` is empty extract the **first column** of the matrix (because R ignores `0` indices and `1` extracts the first column) instead of returning an empty matrix. This leads to `rowSums` returning the sum of the first column instead of 0.

### The Fix

Replace `1:length(x)` with:
- `seq_along(x)` for loops and vector subsetting.
- `seq_len(length(x))` when generating indices (e.g. matrices).

For the matrix subsetting, I also added `drop = FALSE` so that if `names.y` is length 1, the matrix doesn't accidentally get cast to a vector before `rowSums` is called.

Files affected:
- `modules/assim.sequential/R/sda.enkf.R`
- `modules/assim.sequential/R/Analysis_sda.R`
- `modules/data.atmosphere/R/metgapfill.R`

man080107:
## Summary

Fixes the `1:length()` bugs found in `PEcAn.assim.sequential` and `PEcAn.data.atmosphere`. Closes #3951.

---

### Description
The `1:length(x)` construct evaluates to `c(1, 0)` when `length(x) == 0`.
This creates silent bugs:
1. Loops execute twice (for `i=1` and `i=0`) instead of zero times.
2. `x[, 1:length(names.y)]` extracts the first column (`x[, c(1, 0)]`) instead of an empty matrix, silently corrupting calculations like `rowSums()`.

Fixes applied to:
- `modules/assim.sequential/R/sda.enkf.R` (also added `drop = FALSE` to array subsetting)
- `modules/assim.sequential/R/Analysis_sda.R`
- `modules/data.atmosphere/R/metgapfill.R`


mdietze (approved):

mdietze:
Thanks!  tagging @DongchenZ @yinghaoSunn @cami-webb so they know about this SDA change

PR #3956, opened 2026-04-24
Rewrite plot_variance_decomposition as a single facetted plot

infotroph:
##Description

The existing variance decomposition plots are secretly a list of four independent panels (three sets of lollipops and one stack of labels) all `grid.arrange()`'d together, and its row alignment is pretty sensitive to font size and  figure dimensions. I reworked the function using ggplot facets, for simpler calls and hopefully more reliable alignment.

Also changed the default ordering to put the highest-variance parameters at the top of the plot, with option to change sorting if needed.

Bonus: Fixed Roxygen complaint "`@md` must not be followed by any text" and moved some tags to keep them from making `@description` blocks get parsed as `@details`.

### Before
Quick, which variable is that showing a large elasticity near the middle of the plot? I had to count up from the bottom to figure out whether it was daily Amax fraction or optimum temperature for photosynthesis. The row order also isn't meaningful (it's just whichever order the results arrived in), making it harder to either find specific variables or pick out which ones have high or low contributions.

Screenshot 2026-04-24 at 1 44 41 PM

### After
Rows are reliably aligned, so it's now clear optimum temperature is the elastic one. Rows are also now sorted by fraction of variance, but can also be set to CV, elasticity, parameter name, or "none" to keep input order as before.

Screenshot 2026-04-24 at 1 49 29 PM

##


modules/uncertainty/R/plots.R
```diff
@@ -1,63 +1,88 @@
...
+    ggplot2::theme_classic() +
+    ggplot2::theme(
+      axis.line = ggplot2::element_blank(),
+      axis.text.x = ggplot2::element_text(size = fontsize$axis, vjust = -1),
+      axis.text.y = ggplot2::element_text(size = fontsize$axis, hjust = 1),
+      axis.ticks = ggplot2::element_blank(),
+      axis.title.x = ggplot2::element_blank(),
+      axis.title.y = ggplot2::element_blank(),
+      geom = ggplot2::element_geom(pointsize = ggplot2::rel(3),
+                                   linewidth = ggplot2::rel(1.25)),
```
infotroph: CI failure informs me that `element_geom()` was new in ggplot 4.0.0. 
Options I see: Easiest is to require ggplot2 >= 4.0.0, medium is to apply this scaling only if ggplot version is sufficient, harder is find a scaling (or set of them) that works across versions.

 @dlebauer do you have opinions here?

infotroph: Tried an approach that branches on ggplot version in 59ec0a621 -- gives results that look OK to me with both ggplot2 4.0.2 and 3.5.2

dlebauer (approved):
Looks good. I like the new sort by argument!

modules/uncertainty/R/plots.R
dlebauer: I'm in favor of requiring>4.0.

Then you can simplify the lollipop function.

dlebauer: But since you already solved it no need to revert. (When I read this question I didn't realize that lollipop was the fix)

PR #3958, opened 2026-04-25
fix(shiny/Data-Ingest): remove duplicate package entries in app.R and DESCRIPTION

### LINKED ISSUE #3957: fix(shiny/Data-Ingest) :  remove duplicate package entries in app.R and DESCRIPTION
While going through the `shiny/Data-Ingest` module, I noticed two cases where 
the same package is listed twice — once in `app.R` and once in `DESCRIPTION`. 
These are straightforward copy-paste leftovers, no logic change needed, just 
cleaning up the redundancy.

omkarrr2533:
Fixes #3957 

Two duplicate package entries removed:
- shiny was listed twice in the lapply call in app.R
- shinyjs was listed twice under Imports in DESCRIPTION

No logic changes, just cleanup.


infotroph (approved):
Thanks! This is another reason it's good practice to keep import lists alphabetized (but clearly none if us have come back to fix these existing ones...)

shiny/Data-Ingest/DESCRIPTION
```diff
@@ -17,6 +17,5 @@ Imports:
 	shinyTime,
```
infotroph, line 18: ☝️ Can you add a new line 3 (between Title: and License:) that reads `Version: 0.1.0.9000`, please? GitHub won't let me post it as a suggestion that far from the edited lines.

omkarrr2533: Done, added Version: 0.1.0.9000 between Title and License.

PR #3962, opened 2026-04-27
Add UAN-32 and CAN-17 to fertilizer_composition_data

AritraDey-Dev:
## Description
This PR adds `uan_32` and `can_17` to `fertilizer_composition_data` per discussion with @dlebauer in slack, so the Nichols et al. 2024 California almond orchard cal/val dataset (doi:10.1002/saj2.20615) can split single `kg N ha⁻¹` fertigation values into the schema's `nh4_n_kg_m2` / `no3_n_kg_m2` / `org_n_kg_m2` fields. Mass fractions are taken directly from the paper: UAN-32 (7.75% NH₄-N + 7.75% NO₃-N + 16.5% urea-N) and CAN-17 (5.4% NH₄-N + 11.6% NO₃-N).

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


dlebauer (approved):
LGTM, thanks! 

After thinking about it, a pre-existing issue is that urea-n should be a separate column in this table. Many models treat it as nh4-n rather than organic N. Could you either update this PR to add a new column or open a new issue?

AritraDey-Dev:
I'll keep this PR scoped to the Nichols additions and open a separate issue for fraction_urea_n. note that the existing data is already inconsistent, urea encodes its 0.46 N as nh3_n, but 46_00_00 (also urea) encodes it as no3_n. So the new column would also unify those rows. I'll mention that in the issue.

AritraDey-Dev:
created here https://github.com/PecanProject/pecan/issues/3963

PR #3965, opened 2026-04-29
Fix inefficient spatial covariance calculation and parameter mapping in uncertainty SA

### LINKED ISSUE #3964: Fix inefficient spatial covariance calculation and parameter mapping in uncertainty SA
There are two bugs in the codebase:
1. **modules/emulator/R/calcSpatialCov.matrix.R**: The `calcSpatialCov.matrix` function contains a redundant inner loop `for (j in seq_len(nl))` which recalculates and assigns the same vector to a matrix row `nl` times, turning a vectorized element-wise operation into an $O(n^3)$ bottleneck.
2. 2. **modules/uncertainty/R/sensitivity.R**: In `write.sa.configs` (around line 470), `do.call` is invoked with an unnamed argument for `run.id` inside the `args` list. This positional mapping is brittle and can lead to bugs if the target function signature changes. It should explicitly use `run.id = run.id`.

man080107:
Resolves #3964


mdietze (approved):
In the future, I'd recommend that separate bugs be separate issues and PRs. These two fixes have absolutely no relationship with one another.

Also, can you confirm that the cov calculation has been tested and does indeed return an object of the correct format, dimensions, and values?

modules/emulator/R/calcSpatialCov.matrix.R
```diff
@@ -1,22 +1,3 @@
...
-##' Can make gaussian by passing squared distance matrix
-##'
-##' @param d spatial distance matrix
-##' @param psi spatial corr
-##' @param tau spatial var
-##' @param ... additional arguments (currently unused)
-##' @return spatial covariance matrix
-##' @author Michael Dietze
-##' @export
 calcSpatialCov.matrix <- function(d, psi, tau, ...) {
```
mdietze, line 12: Revised version should retain Roxygen documentation

modules/emulator/R/calcSpatialCov.matrix.R
mdietze, line 8: Unclear why this was added to the function since it's unused

infotroph:
Changes already merged in #3969

PR #3967, opened 2026-04-29
add CalAdapt WRF download function for CMIP6 hourly met data

divine7022:
## Description
adds `download.CalAdaptWRF()` to `PEcAn.data.atmosphere`  new met driver that pulls hourly WRF dynamically downscaled CMIP6 projections from the Cal-Adapt Analytics Engine (WUS-D3 dataset, Rahimi et al. 2024). Data sits on public AWS S3, no auth needed

this is implemented as a part of CCMMF where we need future climate forcing at ~200 California sites for SIPNET runs under multiple GCMs and SSPs.

- looked at how pecan handles met downloads and this follows the same pattern as CRUNCEP/GFDL, the download function does everything (fetch, extract, convert, write CF) in one shot, so we skip the met2CF and extract.nc stages. Main reason: WRF uses a Lambert Conformal grid and `extract.nc` / `closest_xy` assume lat-lon grids with NARR-style bounds, so they can't handle this projection. Adding CalAdaptWRF to skip from list in met.process.R was the ryt direction
- the tricky part is that pecan's `papply` calls `download.CalAdaptWRF` once per site, it never sees the full site list. Naively that means re-reading the same WRF grid from S3 for every site. The 45 km grid is small (~20-30 MB per var per year), so we cache the full grid as rds in `tempdir()` on the first site. Sites 2 through N just do
`readRDS()` and extract their grid cell locally. For 200 sites x 8 vars x 20 years, that cuts S3 round trips from 32,000 to 160. The cache auto cleans when R exits
- added `caladapt_wrf` column to `pecan_standard_met_table`
- 9 output variables total:
   - direct (no conversion): air_temperature, air_pressure, shortwave, longwave, eastward_wind, northward_wind
   - converted: precipitation_flux (mm/hr -> kg/m2/s), specific_humidity (mixing ratio -> q/(1+q))
   - derived: wind_speed (sqrt(u10^2 + v10^2))

- data coverage:
  - 8 GCMs under SSP3-7.0: CESM2, CNRM-ESM2-1, EC-Earth3, EC-Earth3-Veg, FGOALS-g3, MPI-ESM1-2-HR, MIROC6, TaiESM1
  - CESM2 also has SSP2-4.5 and SSP5-8.5
  - 1980-2100, hourly, 45 km (d01)
  - 3 models store precip as rainc + rainnc components instead of a single prec field; but caladaptR handles that transparently

pipeline that orchestrates R (caladaptaer) and .sh scripts, along with the data, is here at /projectnb/dietzelab/ccmmf/ensemble/CalAdapt_runs/
for 198 design points × 3 GCM/SSP scenarios (CESM2.ssp245, CESM2.ssp370, MPI-ESM1-2-HR.ssp370) × 2025–2045 hourly; if anyone wants to poke at the outputs

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:


book_source/03_topical_pages/06_data/01_meteorology.Rmd
```diff
@@ -175,3 +176,23 @@ Resolution: 30 min
...
+
+## Cal-Adapt WRF {#caladaptwrf}
+
+Scale: Regional (Western US)
+
+Resolution: 1 hr, 45 km (d01), 9 km (d02), 3 km (d03)
+
+Availability: 1980--2100
+
+Notes: CMIP6 dynamically downscaled projections from the Cal-Adapt Analytics Engine (WUS-D3 dataset, Rahimi et al. 2024). Eight GCMs available under SSP3-7.0; CESM2 also has SSP2-4.5 and SSP5-8.5. Data is publicly available on AWS S3 (no authentication required). Requires the `caladaptaer` package from GitHub. To use this option, set `source` as `CalAdaptWRF` and specify `model` and `scenario` in the `met` section of `pecan.xml`:
```
dlebauer, line 188: add historical reanalysis

book_source/03_topical_pages/06_data/01_meteorology.Rmd
```diff
@@ -175,3 +176,23 @@ Resolution: 30 min
...
+
+```xml
+<met>
+  <source>CalAdaptWRF</source>
+  <model>CESM2</model>
+  <scenario>ssp370</scenario>
+</met>
+```
+
+Available GCMs: CESM2, CNRM-ESM2-1, EC-Earth3, EC-Earth3-Veg, FGOALS-g3, MPI-ESM1-2-HR, MIROC6, TaiESM1. See `caladaptaer::cae_models(activity = "WRF")` for the current list.
```
dlebauer, line 198: 1. Be more specific, e.g.
```suggestion
Available GCMs: CESM2, CNRM-ESM2-1, EC-Earth3, EC-Earth3-Veg, FGOALS-g3, MPI-ESM1-2-HR, MIROC6, TaiESM1. See `caladaptaer::cae_models(activity = "WRF")` for the current list of available climate models.
```
2. What does 'activity' do?
3. add reference to availability of reanalysis data

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
+#' Download Cal-Adapt WRF CMIP6 outputs for a single site and convert to CF
+#'
+#' Fetches hourly WRF dynamically downscaled data from the Cal-Adapt Analytics
+#' Engine (CADCAT S3 bucket) via caladaptaer, extracts the nearest grid cell to
+#' the site, converts units to CF-1.8, and writes one NetCDF per year.
+#'
+#' WRF grids are cached in tempdir() so that when met.process calls this for
+#' multiple sites in the same R session, each grid is only fetched from S3 once.
+#' For 200 sites x 8 vars x 20 years that cuts S3 round trips from 32,000 to 160.
+#'
```
dlebauer, line 10: add references, including links to docs and Rahimi paper

document availablity of downscaled ERA5 reanalysis

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+#' @param scenario SSP experiment id, default "ssp370"
+#' @param resolution WRF nested-domain id: "d01" (45 km, default and only
+#'   one with full coverage), "d02" (9 km), or "d03" (3 km). Higher
+#'   resolutions are only partially released; check
+#'   \code{caladaptaer::cae_check_variables()} for availability.
+#' @param overwrite Overwrite existing files? Default FALSE
+#' @param verbose Extra debug output? Default FALSE
+#' @param ... further arguments, currently ignored
+#'
+#' @return invisible data.frame with file info for BETY registration
```
dlebauer, line 34:
```suggestion
#' @return invisible data.frame with file information.
```

I think it can return the standardized table without implying that the only intent is for BETY registration.

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+#' EC-Earth3-Veg, FGOALS-g3, MPI-ESM1-2-HR, MIROC6, TaiESM1.
+#' Only CESM2 has ssp245 and ssp585 in addition to ssp370.
+#'
+#' Precipitation for MPI-ESM1-2-HR, MIROC6, and TaiESM1 is derived from
+#' rainc + rainnc components; caladaptaer handles this transparently.
+#'
+#' @param outfolder Directory for storing output
+#' @param start_date Start date for met data
+#' @param end_date End date for met data
+#' @param site_id BETY site id
```
dlebauer, line 21: I think this param can be defined as `unique identifier for site` without reference to BETYdb.
Is it still necessary / useful to explicitly support BETYdb with this?

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+#' @param lat.in Latitude of site (decimal degrees, WGS84)
+#' @param lon.in Longitude of site (decimal degrees, WGS84)
+#' @param model WRF GCM name, default "CESM2"
+#' @param scenario SSP experiment id, default "ssp370"
+#' @param resolution WRF nested-domain id: "d01" (45 km, default and only
+#'   one with full coverage), "d02" (9 km), or "d03" (3 km). Higher
+#'   resolutions are only partially released; check
```
dlebauer, line 28: more specific details here please, what is the scope of unreleased data - presumably the constraint is on variables rather than spatial or temporal domain coverage. Seems something useful to document in caladaptaer package

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+      "Install with: remotes::install_github('lebauerapproach/caladaptaer')")
+  }
+  if (!requireNamespace("sf", quietly = TRUE)) {
+    PEcAn.logger::logger.severe("sf package required for CRS transform")
+  }
+  if (!requireNamespace("stars", quietly = TRUE)) {
+    PEcAn.logger::logger.severe("stars package required for grid extraction")
+  }
+
+  # null guard, convert_input sometimes passes NULL for optional args
```
dlebauer, line 57: Seems there should be a bug report + test created to prevent this convert_input behavior

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+  if (is.null(resolution)) resolution <- "d01"
+
+  start_year <- lubridate::year(start_date)
+  end_year   <- lubridate::year(end_date)
+
+  # BETY site id formatting
+  site_id <- tryCatch(as.numeric(site_id),
+                       warning = function(w) as.character(site_id))
+  if (is.numeric(site_id) && site_id > 1e9) {
+    siteid_str <- paste0(site_id %/% 1e9, "-", site_id %% 1e9)
```
dlebauer, line 69: this seems dangerous and non-unique

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+
+      vals <- extracted[[1]]
+      if (inherits(vals, "units")) vals <- units::drop_units(vals)
+      dat.list[[wrf_var]] <- as.numeric(vals)
+    }
+
+    ##unit conversions
+    # precip: WRF hourly accumulation (mm) -> CF flux (kg/m2/s)
+    # 1 mm water = 1 kg/m2, divide by 3600s for hourly timestep
+    if ("prec" %in% names(dat.list)) {
```
dlebauer, line 187: for unit conversions 
- use PEcAn.utils::ud_convert, even if for speed it is done once at the head of the file to create a conversion factor.
- use (or create) functions from modules/data.atmosphere metutils.R so that all assumptions about transformations are in one place rather than the more error-prone approach of implementing the same transforms for each data source.

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+        units   = fetch_tbl$units[j],
+        dim     = dim,
+        missval = -9999.0,
+        verbose = verbose
+      )
+      var.names[j] <- fetch_tbl$caladapt_wrf[j]
+    }
+
+    # wind speed as derived variable
+    if (!is.null(wind_speed) && nrow(derived_tbl) > 0) {
```
dlebauer, line 239: why add as missing value rather than omitting?

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+    time_secs <- as.numeric(difftime(
+      time_vals,
+      as.POSIXct(paste0(year, "-01-01 00:00:00"), tz = "UTC"),
+      units = "secs"
+    ))
+
+    lat_dim  <- ncdf4::ncdim_def("latitude", "degree_north",
+                                  lat.in, create_dimvar = TRUE)
+    lon_dim  <- ncdf4::ncdim_def("longitude", "degree_east",
+                                  lon.in, create_dimvar = TRUE)
```
dlebauer, line 216: these write the site lat/lon rather than the lat lon from the source data. if we want to support potentially multiple sites mapping to the same met input, should this be handled differently here?

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+                       warning = function(w) as.character(site_id))
+  if (is.numeric(site_id) && site_id > 1e9) {
+    siteid_str <- paste0(site_id %/% 1e9, "-", site_id %% 1e9)
+  } else {
+    siteid_str <- as.character(site_id)
+  }
+  outfolder <- paste0(outfolder, "_site_", siteid_str)
+
+  lat.in <- as.numeric(lat.in)
+  lon.in <- as.numeric(lon.in)
```
dlebauer, line 76: is it worthwhile to add a check that these are inside domain of dataset? I'm not sure if this is worth the computational cost of a point in polygon check, but if there is a bounding box it should be efficient to do `lat < maxlat & lat > minlat` type checks. If there isn't a bounding box, perhaps caladaptaer should create one. Either way, out of domain lat/lon should be handled gracefully.

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+          scenario   = scenario,
+          timescale  = "1hr",
+          resolution = resolution,
+          start_time = year_start,
+          end_time   = year_end
+        )
+        saveRDS(grid, cache_file)
+      }
+
+      # grab time dimension and build the projected point once
```
dlebauer, line 166: Does CalAdapt gaurantee that, for a given model/scenario/resolution combination, time_vals and grid will be consistent? is the grid the same across all data at the same resolution?

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+    PEcAn.logger::logger.severe("sf package required for CRS transform")
+  }
+  if (!requireNamespace("stars", quietly = TRUE)) {
+    PEcAn.logger::logger.severe("stars package required for grid extraction")
+  }
+
+  # null guard, convert_input sometimes passes NULL for optional args
+  if (is.null(model))      model <- "CESM2"
+  if (is.null(scenario))   scenario <- "ssp370"
+  if (is.null(resolution)) resolution <- "d01"
```
dlebauer, line 60:
```suggestion
model      <- model %||% "CESM2"
scenario   <- scenario %||% "ssp370"
resolution <- resolution %||% "d01"
```

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
+#' Download Cal-Adapt WRF CMIP6 outputs for a single site and convert to CF
+#'
+#' Fetches hourly WRF dynamically downscaled data from the Cal-Adapt Analytics
+#' Engine (CADCAT S3 bucket) via caladaptaer, extracts the nearest grid cell to
+#' the site, converts units to CF-1.8, and writes one NetCDF per year.
```
dlebauer, line 5: to be strictly CF-1.8 compliant, files need a global metadata attribute `Conventions = CF-1.8`.

dlebauer (changes requested):
This is a great first pass. I've made a few comments; a few key points:

It seems `download.CalAdaptWRF()` is doing both the download and conversion standardization (q2 transformation, derived wind_speed).

It isn't clear if it is an intentional design choice, but the standard way to handle met ingest is to first download and then standardize. i.e  `download.CalAdaptWRF()` responsible for downloading and `met2CF.CalAdaptWRF()` responsible for conversion to PEcAn standard met. That seems more consistent with PEcAn’s pattern of deriving standardized variables in the met conversion layer rather than in the data interface itself.

Two other points - 1) it seems like it would be useful to put conversion functions for `q = f(q2)`, `wind = f(uwind, vwind)`, and `precip = f(rainc, rainnc)` into metutils.R and call those. and 2) I think that the conversion `precip = f(rainc, rainnc)` can be removed from caladaptaer in order to keep that package's scope to data access (with apologies for previously suggesting otherwise, that was before we were developing a dedicated package intended for the caladaptae ecosystem).

modules/data.atmosphere/R/download.CalAdaptWRF.R
```diff
@@ -0,0 +1,266 @@
...
+        grid <- caladaptaer::cae_fetch(
+          variable   = wrf_var,
+          model      = model,
+          scenario   = scenario,
+          timescale  = "1hr",
+          resolution = resolution,
+          start_time = year_start,
+          end_time   = year_end
+        )
+        saveRDS(grid, cache_file)
```
dlebauer, line 163: is this safe if run in parallel?

dlebauer (commented):
Need to update to use caladaptaer functionality, download and met2cf should be minimal wrappers of that package

PR #3969, opened 2026-05-01
Fix inefficient spatial covariance calculation in emulator and add tests

man080107:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

### DIFF
--- a/modules/emulator/tests/testthat/test.calcSpatialCov.matrix.R
+++ b/modules/emulator/tests/testthat/test.calcSpatialCov.matrix.R
@@ -9,6 +9,11 @@ test_that("calcSpatialCov.matrix returns correct format, dimensions, and values"
   expect_true(is.matrix(res))
   expect_equal(dim(res), c(2, 2))
   
-  expected <- tau * exp(-psi * d)
+expected <- matrix(
+  c(2.00000000000000, 1.21306131942527,
+    1.21306131942527, 2.00000000000000),
+  nrow = 2,
+  byrow = TRUE
+)
   expect_equal(res, expected)
 })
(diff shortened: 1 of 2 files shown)


modules/emulator/tests/testthat/test.calcSpatialCov.matrix.R
```diff
@@ -0,0 +1,14 @@
...
+test_that("calcSpatialCov.matrix returns correct format, dimensions, and values", {
+  d <- matrix(c(0, 1, 1, 0), 2, 2)
+  psi <- 0.5
+  tau <- 2.0
+  res <- calcSpatialCov.matrix(d, psi, tau)
+  
+  expect_true(is.matrix(res))
+  expect_equal(dim(res), c(2, 2))
+  
+  expected <- tau * exp(-psi * d)
```
mdietze: Having the expected value come from the exact same calculation isn't a strong test that the calculation is correct. Would be better to store what the expected cov matrix should be.

mdietze (approved):

modules/emulator/tests/testthat/test.calcSpatialCov.matrix.R
infotroph: Seconding @mdietze point -- I recommend updating before we merge this.

dlebauer: like this?

```suggestion
expected <- matrix(
  c(2.00000000000000, 1.21306131942527,
    1.21306131942527, 2.00000000000000),
  nrow = 2,
  byrow = TRUE
)
```

PR #3970, opened 2026-05-03
Downgrade warning to info in read.results

infotroph:
What it says on the tin. No need to warn people when we're doing what they asked us to do.

### DIFF
--- a/base/utils/R/read.output.R
+++ b/base/utils/R/read.output.R
@@ -63,11 +63,6 @@ read.output <- function(runid, outdir,
   ## cflux = c('GPP', 'NPP', 'NEE', 'TotalResp', 'AutoResp', 'HeteroResp', 'DOC_flux', 'Fire_flux') # kgC m-2 s-1
   ## wflux = c('Evap', 'TVeg', 'Qs', 'Qsb', 'Rainf') # kgH20 m-2 d-1
 
-  if (!isTRUE(verbose)) {
-    loglevel <- PEcAn.logger::logger.setLevel("WARN")
-    on.exit(PEcAn.logger::logger.setLevel(loglevel), add = TRUE)
-  }
-
   if ((missing(runid) || missing(outdir)) && is.null(ncfiles)) {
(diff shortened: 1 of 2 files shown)


divine7022 (approved):
could you tack on sentence to `@param verbose` spelling that out ? 
something quick like - "when `FALSE`, also bumps the logger to `WARN` for the call, so INFO/DEBUG from callees gets muted too." 
that way nobody's left wondering where their debug logs went

infotroph:
On consideration, the read.output change needs more thought: (1) most of the debugs are already wrapped in `if (verbose)`, so maybe it's more consistent / obvious to extend that to the rest of them; (2) the tests check heavily for exact message wording and since they need updating anyway, we should update them to check that each relevant _output_ is as expected (i.e. `expect_match(out_log, "Reading output for all variables", all = FALSE)` -> check that all variables were in fact read)

infotroph:
@divine7022 Thanks for trying to be proactive, but please don't push test changes to other people's PRs without checking first. I'm going to force-push this back to 8d28b46.

base/utils/R/read.output.R
```diff
@@ -63,6 +63,11 @@ read.output <- function(runid, outdir,
   ## cflux = c('GPP', 'NPP', 'NEE', 'TotalResp', 'AutoResp', 'HeteroResp', 'DOC_flux', 'Fire_flux') # kgC m-2 s-1
   ## wflux = c('Evap', 'TVeg', 'Qs', 'Qsb', 'Rainf') # kgH20 m-2 d-1
 
+  if (!isTRUE(verbose)) {
+    loglevel <- PEcAn.logger::logger.setLevel("WARN")
+    on.exit(PEcAn.logger::logger.setLevel(loglevel), add = TRUE)
+  }
+
```
infotroph: Deferring this as #3996 so we can merge the get.results change. 
```suggestion
 ```

infotroph:
Deferring the read.output changes as #3996, merging the read.results change

PR #3977, opened 2026-05-08
Events parquet to JSON pipeline

ashiklom:
## Description

- Adds code for converting specifically formatted event parquet files into events.json.
- Defines a (simple) schema / standard for events data in parquet format.
- Defines and implements a mechanism for relating PEcAn ensemble IDs to heterogeneous ensembles of management events

@infotroph and @dlebauer have discussed this approach already. Also tagging @divine7022 and @sarahkanee as other people that generate management events and @mdietze as the person in charge of the management events pipeline.

## Motivation and Context

https://github.com/ccmmf/organization/issues/229

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X ] When possible

## Types of changes

## Checklist:

- [X] My name is in the list of CITATION.cff
- [X] I agree that PEcAn Project may distribute my contribution under any or all of


book_source/03_topical_pages/02_pecan_standards.Rmd
```diff
@@ -103,6 +103,8 @@ Externally imposed changes in system state, such as planting and harvest or till
...
 
+Although JSON is the official PEcAn format for management events, PEcAn supports more efficient storage of large collections of event data in parquet format. Such data can be converted to the JSON format with `data.land::event_parquet_to_json`. Data for each type of managmeent event must be stored in its own parquet file (e.g., you _cannot_ combine data for leaf on and leaf off events in a `phenology.parquet` file). Each parquet file must have columns `date`, `site_id`, and additional columns for event-specific data (e.g., `method` and `amount_mm` for irrigation). For ensemble uncertainty analysis, events may also (optionally, but highly recommended) include an `event_member_id` column (as a character) that uniquely identifies each ensemble member for downstream uncertainty analysis. (This is distinct from PEcAn's ensemble ID because PEcAn's ensemble code treats all management events as a single input type. The `event_member_id` for each event type is therefore used to extend PEcAn's ability to attribute sensitivity or uncertainty to specific event types. See `data.land::event_parquet_to_json` docs for additional details).
```
mdietze, line 106: are we locked into `event_member_id` or could we look for something more in line with other conventions? For example, the function`get_event_ensemble_ids` grabs the `event_member_id` not the `event_ensemble_id` so why not adopt `event_ensemble_id` as the name?

infotroph: Without any claim to speak for @ashiklom, I like `member_id` to emphasize it applies to one element. I've gotten used to the whole-workflow usage of `ensemble_id` as the shared identifier for the whole set and would find it confusing for `event_ensemble_id`s 2, 8, and 43 to comprise `ensemble_id` 2cfc98dd.

ashiklom: Two hard things in computer science: Cache invalidation, naming things, and off-by-one errors!

This is trivial to change --- no lock in --- so I'm happy to call it whatever you all prefer. I went with event_member_id because I think @infotroph suggested that when we discussed it a few weeks ago. I have no strong opinions, as long as the word "event" is somewhere in there to disambiguate it from the generic ensemble ID.

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+#' @param event_paths (character) Named character vector of paths to PEcAn
+#' standard event parquet files. Names correspond to event types; e.g.,
+#' `c(leafon = "/path/to/leaf_on.parquet")`. If not names are provided, they
+#' are deduced from the file name (stripping the `.parquet` extension). Folders
+#' containing parquet files are also supported.
+#' @param events_ensemble_manifest `data.frame` that maps PEcAn ensemble IDs
+#' and events.json paths onto ensembles of each event type. (See Details).
+#' @param parquet_dir Alternative way to pass `event_paths` by just specifying
+#' a directory of PEcAn standard event files. Each file or folder must be named
+#' `<event_type>.parquet` (e.g., `irrigation.parquet/`, `leafon.parquet`).
```
infotroph, line 86: Does this allow event files that are organized by year, e.g. the existing `planting/v1.0/planting_statewide_2016.parquet`?

book_source/03_topical_pages/02_pecan_standards.Rmd
infotroph: I agree "event" does help disambiguate, and also I take @mdietze point that the function name and the variable name should match.
* Obvious option: `get_event_member_ids()`?
* Maybe clearer, but possibly too cute:  `get_event_ids_for_ensemble`?

infotroph (approved):
updated paths for my machine, ran the examples, liked the result ✅ ✅ ✅. Thanks!

book_source/03_topical_pages/02_pecan_standards.Rmd
mdietze: If we're referencing the PEcAn ensemble_id, the elements in an ensemble are run_ids not member_ids, so would event_run_id and get_event_run_ids work or will that generate a different source of confusion.

modules/data.land/R/event_parquet_to_json.R
ashiklom: Haven't tested, but I _believe_ `arrow::open_dataset("path/to/dir", format="parquet")` will recursively search `dir` for all `.parquet` files, regardless of file name, and load them together. So yes, this should work.

That said, if we're adapting this kind of organization, I highly recommend adopting Parquet's ["hive partitioning" style](https://arrow.apache.org/docs/r/reference/open_dataset.html#partitioning) (e.g., `path/to/dir/year=2010/file.parquet`). This allows arrow to get the year directly out of the file name, which makes querying on `year` _much_ faster.

mdietze: IIRC, won't the arrow library automatically write this partitioning correctly if you explicitly have a `group_by` on the data frame?

book_source/03_topical_pages/02_pecan_standards.Rmd
infotroph: I think that'd be a different (probably smaller) confusion because a "run" is so tied to the moment of model execution that it seems confusing to apply it upstream of design matrix generation for identifiers that won't change between workflows and could repeat within workflows (eg event_member_id 3 gets selected for run_ids 5 and 18).

I'm advocating for "member" because (to me) it's a decent generic term for "individual in the grouping at hand" and I'm comfortable using it across multiple levels of ensemble assembly -- and if we were starting from scratch I'd happily apply the same convention to say "ensemble member id" where we use "run id" now, but I feel no need to change that one retroactively.

But at this point I'm definitely overthinking it and I know we're all aiming for a consistent convention however we interpret the details. @mdietze I'll let you make the call here.

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+      if (any(is_no_ens)) {
+        stop(
+          "In ", shQuote(event_type),
+          ": Cannot mix ", NO_EVENT_ENSEMBLE,
+          " with actual ensemble members."
+        )
+      }
+      dat <- dat |>
+        dplyr::filter(.data$event_member_id %in% .env$pick_ensembles)
+    }
```
divine7022, line 149: If I'm reading this right, there's an inverse case check doesn't catch; parquet has event_member_id but manifest column for this event type is all _NO_EVENT_ENSEMBLE
i wondering if it's worth tightening check ?

workflows/preprocess-event-parquet/02-events-to-json.R
```diff
@@ -0,0 +1,42 @@
...
+  jsonvalidate::json_validate(
+    path,
+    schema_file,
+    engine = "ajv",
+    verbose = TRUE,
+    error = TRUE
+  )
+  utils::setTxtProgressBar(pb, i)
+}
+close(pb)
```
divine7022: just calling validate_events_json would be even cleaner than rolling validate loop by hand here

```suggestion
message("Validating event files")
pb <- utils::txtProgressBar(0, nrow(events_files))
for (i in seq_len(nrow(events_files))) {
  path <- events_files[["json_path"]][[i]]
  PEcAn.data.land::validate_events_json(path, schema_version = "0.1.2", verbose = TRUE)
  utils::setTxtProgressBar(pb, i)
}
close(pb)
```

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+  message("Writing event files")
+  nevent <- nrow(combined_events)
+  pb <- utils::txtProgressBar(0, nevent)
+  for (irow in seq_len(nevent)) {
+    drow <- combined_events[irow, ]
+    event_data <- drow[["event_data"]][[1]]
+    outfile <- drow[["json_path"]]
+    jsonlite::write_json(event_data, outfile, pretty = TRUE, auto_unbox = TRUE)
+    utils::setTxtProgressBar(pb, irow)
+  }
```
divine7022, line 206: wondering if this is right place to optionally call validate_events_json; ryt now example workflow re implements validation outside the function, catching schema violation at write site should give better error than letting write.events.SIPNET choke on bad file later
could be opt-in so it stays cheap by default
```
event_parquet_to_json <- function(
  event_paths,
  events_ensemble_manifest,
  site_ids = NULL,
  start_date = NULL,
  end_date = NULL,
  parquet_dir = NULL,
  pecan_events_version = "0.1.2",
  validate = TRUE
) {
```
...then inside the write loop
```
  jsonlite::write_json(event_data, outfile, pretty = TRUE, auto_unbox = TRUE)
  if (isTRUE(validate)) {
     validate_events_json(outfile, schema_version = pecan_events_version)
  }
```

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+  combined_events <- all_events_list |>
+    dplyr::bind_rows() |>
+    dplyr::summarize(
+      events = list(purrr::list_c(.data$events)),
+      .by = c("pecan_events_version", "site_id", "ensemble_id", "json_path")
+    ) |>
+    tidyr::nest(
+      .key = "event_data",
+      .by = c("ensemble_id", "json_path")
+    )
```
divine7022, line 196: events_schema_v0.1.2.json allows ensemble_id. ryt now ensemble identity only helps in the filename
seems like would it be worth stamping ensemble_id into each site object ?
I think this is just a matter of carrying ensemble_id through the inner nest


```suggestion
  combined_events <- all_events_list |>
    dplyr::bind_rows() |>
    dplyr::summarize(
      events = list(purrr::list_c(.data$events)),
      .by = c("pecan_events_version", "site_id", "ensemble_id", "json_path")
    ) |>
    tidyr::nest(
      .key = "event_data",
      .by = c("ensemble_id", "json_path")
      .keep = TRUE  # keep ensemble_id in event_data too
    )
```
not blocking, but it's kinda metadata that becomes painful to backfill later

workflows/preprocess-event-parquet/01b-clean-other-events.R
```diff
@@ -0,0 +1,94 @@
...
+  arrow::write_parquet(
+    file.path(outdir, "planting.parquet"),
+    compression = "ZSTD"
+  )
+
+pct_to_tillage <- function(x) {
+  xfrac <- x / 100
+  x1 <- pmax(xfrac - 0.3, 0.0)
+  pmin(x1 / 0.7, 1.0)
+}
```
divine7022, line 50: I worked math on this, it stays at 0 until 30% NDTI change, ramps linearly, saturates at 100% 
wondering if you have got citation for 30% threshold ?
100% saturation cap feels especially arbitrary, under heavy till ndti can change by way more than 100%
two thoughts (1) quick ref to whoever derived this mapping would help, and (2) could 0.3 and 0.7 be function args with defaults ? 

```suggestion
#' @param threshold_pct NDTI percent change below which tillage_eff is 0 (default 30%, from <citation>)
#' @param saturation_pct NDTI percent change at which tillage_eff saturates at 1 (default 100%)
pct_to_tillage <- function(x, threshold_pct = 30, saturation_pct = 100) {
  xfrac <- x / 100
  thresh_frac <- threshold_pct / 100
  sat_frac <- saturation_pct / 100
  x1 <- pmax(xfrac - thresh_frac, 0.0)
  pmin(x1 / (sat_frac - thresh_frac), 1.0)
}
```

workflows/preprocess-event-parquet/01a-clean-irrigation.R
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env Rscript
+
+irr_path <- "/projectnb/dietzelab/ccmmf/usr/ashiklom/event-outputs/irrigation_all"
+
+outdir <- "_output"
+dir.create(outdir, showWarnings = FALSE, recursive = TRUE)
+
+dbdir <- file.path(Sys.getenv("TMPDIR", "/tmp"), "temp.duckdb")
+conn <- DBI::dbConnect(duckdb::duckdb(dbdir = dbdir))
```
divine7022, line 9:
```suggestion
conn <- DBI::dbConnect(duckdb::duckdb(dbdir = dbdir))
on.exit({
  DBI::dbDisconnect(conn, shutdown = TRUE)
  unlink(dbdir)
}, add = TRUE)
```

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+    }
+    event_paths <- list.files(
+      parquet_dir,
+      "*.parquet",
+      full.names = TRUE,
+      include.dirs = TRUE
+    )
+  }
+  if (is.null(names(event_paths))) {
+    names(event_paths) <- sub("\\.parquet", "", basename(event_paths))
```
divine7022: couple of regex things
"*.parquet" literally means any char followed by .parquet ( `.` matches any char) should be "\\.parquet$"
and anchoring with $ should be safer here
```suggestion
      parquet_dir,
      "\\.parquet$",
      full.names = TRUE,
      include.dirs = TRUE
    )
  }
  if (is.null(names(event_paths))) {
    names(event_paths) <- sub("\\.parquet$", "", basename(event_paths))
```

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+#' for each ensemble member as a side effect.
+#' @export
+event_parquet_to_json <- function(
+  event_paths,
+  events_ensemble_manifest,
+  site_ids = NULL,
+  start_date = NULL,
+  end_date = NULL,
+  parquet_dir = NULL,
+  pecan_events_version = "0.1.2"
```
divine7022, line 106: just to make sure i understand the intent here,is there a use case for setting pecan_events_version to anything other than "0.1.2" ? column logic doesn't branch on version, so passing "0.1.1" produces a file labeled with wrong version but laid out per v0.1.2. v0.1.2 adds leafon/leafoff and makes crop_code required, etc so this isn't just a label.
feels like it should fail loudly at function entry rather than at validate time

either drop arg (always write v0.1.2 since that's what code produces) or check it against schemas at function start. I would lean toward dropping it for now

modules/data.land/R/event_parquet_to_json.R
```diff
@@ -0,0 +1,209 @@
...
+    if (!("event_member_id" %in% names(ads))) {
+      results[[event_type]] <- NO_EVENT_ENSEMBLE
+      next
+    }
+    results[[event_type]] <- glue::glue(
+      "SELECT DISTINCT event_member_id FROM read_parquet('{fname}')"
+    ) |>
+      dplyr::sql() |>
+      dplyr::tbl(src = conn) |>
+      dplyr::pull("event_member_id")
```
divine7022, line 50: just curious why duckdb sql for this one rather than staying in arrow ? same operation in arrow is one chain and avoids the glue/SQL round
```suggestion
    results[[event_type]] <- arrow::open_dataset(fname) |>
      dplyr::select("event_member_id") |>
      dplyr::distinct() |>
      dplyr::collect() |>
      dplyr::pull("event_member_id")
```
I get why duckdb is right for 600M row sort in 01a, but get_event_ensemble_ids is doing a small DISTINCT scan and function already imports arrow elsewhere, dropping DBI/duckdb dependency at this one spot would avoid pulling in extra depends here

not blocker mostly asking to make sure I am not missing a perf reason

workflows/preprocess-event-parquet/README.md
```diff
@@ -0,0 +1,7 @@
+# Generating `events.json` ensembles from management pipeline
```
divine7022, line 1: so I went through this complete on fresh checkout and workflow runs clean, but I had to read all three scripts to figure out which one to run first, where the outputs land. 
things we could potentially add to README while we're touching it
first something like "These scripts take the monitoring pipeline parquet outputs (irrigation from CIMIS style data, planting/harvest/tillage/phenology from statewide monitoring stack) and reshape them into the PEcAn standard events parquet format defined in 02_pecan_standards.Rmd, then convert that into events.json ensembles"

second, run order and where the outputs land;easy to miss that 01a and 01b are independent (can run in parallel) but 02 needs both _output/irrigation.parquet/ and others before it'll work. and _output/ vs _events/ split is useful split to spell out

workflows/preprocess-event-parquet/02-events-to-json.R
```diff
@@ -0,0 +1,42 @@
...
+event_dir <- "_events"
+dir.create(event_dir, showWarnings = FALSE, recursive = TRUE)
+parquet_dir <- "_output"
+site_ids <- c(0, 1, 10, 100, 1000, 100005)
+# site_ids <- seq(1, 100)
+# site_ids <- seq(1, 500)
+
+ens_ids <- PEcAn.data.land::get_event_ensemble_ids(parquet_dir = parquet_dir)
+
+events_ensemble_manifest <- dplyr::as_tibble(ens_ids) |>
```
divine7022, line 12: not for this PR but at some point second example (maybe 03-coordinated-manifest.R or just a section in the README) showing one or two non trivial manifest patterns (correlated event types, hand crafted scenarios, joint prior sampling) would make the design self explanatory. ryt now anyone wanting to do anything beyond independent recycling would have to read thread to know that capability exists

divine7022 (changes requested):
LGTM! and ran it against current monitoring data.
few thing I would really want a look at before merge
and tagging @infotroph have been in thread curious whether the parcel_id vs site_id decision is something we want to lock in via sentence in 02_pecan_standards.Rmd or leave for a follow up ?

modules/data.land/R/event_parquet_to_json.R
divine7022: went and checked the arrow docs and confirmed, partitioning arg of write_dataset() defaults to dplyr::group_vars(dataset) and `hive_style` defaults to TRUE, so group_by() carries through automatically
just verified
```
> td <- tempfile()
> tibble(year = c(2016, 2017, 2016), x = 1:3) |>
+   group_by(year) |>
+   write_dataset(td)
> list.files(td, recursive = TRUE)
[1] "year=2016/part-0.parquet" "year=2017/part-0.parquet"
```
one thing worth noting, 01b-clean-other-events.R is currently calling arrow::write_parquet() (singular file writer), not write_dataset(). write_parquet just dumps one file regardless of grouping, so adding group_by(year) upstream wouldn't do anything unless we switch the writer too. If we end up wanting year partitioned monitoring outputs change would be small
```
arrow::open_dataset(harvest_files, format = "parquet") |>
  dplyr::mutate(site_id = as.integer(site_id), date = as.Date(date)) |>
  dplyr::group_by(year) |>
  arrow::write_dataset(file.path(outdir, "harvest.parquet"))
  ```
 for current data sizes single file output is fine, but once we want partition on the read side (`open_dataset -> filter(year >= 2020`) becoming a directory listing instead of full scan), so switching to write_dataset would pay off
@infotroph's earlier question whether year organized files like planting_statewide_2016.parquet work as input; yeah they do but writing outputs in hive form (year=2016/...) makes year virtual column on read which is a nicer downstream interface
probably worth a follow up issue rather than dragging this PR's scope further

File: book_source/03_topical_pages/02_pecan_standards.Rmd (hunk 1 above)
Comment (divine7022) [reply book_source/03_topical_pages/02_pecan_standards.Rmd:106]: vote here @infotroph, existing input_design is already uses member index semantics in its cols (param, met, events, soil, ...) holds integer indices into that input's pool, and uses those to pick a specific sample. integer 12 in input_design$irrigation would naturally point at event_member_id = "irr_ens_12" in parquet so parquet column and design matrix column sit at adjacent semantic levels (string label and integer index) but both are "member" semantics never "run"
going with `member` keeps parquet column compatible with how those upstream downstream relationships propagate through input.ens.gen

File: modules/data.land/R/event_parquet_to_json.R (hunk 11 above)
Comment (ashiklom) [reply modules/data.land/R/event_parquet_to_json.R:50]: Performance. I originally had something like this pattern, but it was roughly an order of magnitude slower. I think DuckDB doesn't actually use Arrow --- it has its own implementations of everything --- and I think DuckDB has some internal optimizations that this allows it to leverage.

File: modules/data.land/R/event_parquet_to_json.R (hunk 5 above)
Comment (ashiklom) [reply modules/data.land/R/event_parquet_to_json.R:206]: Personally, I'd rather not. My preference is generally to have functions do as few things as possible, and to have as few arguments as possible. But it's not a strong preference — if other folks think it makes sense here, I'm happy to add it.

File: modules/data.land/R/event_parquet_to_json.R (hunk 10 above)
Comment (ashiklom) [reply modules/data.land/R/event_parquet_to_json.R:106]: This function makes no assumptions about the types of events in the parquet data it reads except that they are acceptable for the target schema type. For example, if you don't include leaf on / leaf off event parquet files, it will work just fine with version 0.1.1. If in schema version 0.1.3, we add a new event type (and corresponding parquet files), or add new columns to irrigation events or something, all of those changes can be implemented upstream and this code will still work; you'll just need to specify the new event version.

File: modules/data.land/R/event_parquet_to_json.R (hunk 11 above)
Comment (ashiklom) [reply modules/data.land/R/event_parquet_to_json.R:50]: Out of curiosity, I dug into the reasons for the performance difference. Basically, Arrow has a lot of nice primitives for zero-copy access, but no built in analytics engine that can optimize actual operations. You need external libraries that can start from the Arrow memory model and then implement efficient algorithms on top of that.

DuckDB does exactly this -- it has efficient C implementations of things like `distinct` that use hashing and other tricks to make this go much faster.

File: modules/data.land/R/event_parquet_to_json.R (hunk 5 above)
Comment (infotroph) [reply modules/data.land/R/event_parquet_to_json.R:206]: I'd rather not as well -- validation is time-consuming and I'd rather do it intentionally at known handoff points rather than insert it into at every intermediate conversion step.

File: modules/data.land/R/event_parquet_to_json.R (hunk 6 above)
Comment (infotroph) [reply modules/data.land/R/event_parquet_to_json.R:196]: What's the use case for this? These files are already big enough to slow things down and will get bigger, so if it fits in the filename then we should read it from there rather than repeat it for every site object.

File: modules/data.land/R/event_parquet_to_json.R (hunk 10 above)
Comment (infotroph) [reply modules/data.land/R/event_parquet_to_json.R:106]: Building on @ashiklom's point: Even if we didn't expect anyone to change it, passing constants as default argument values is a decent strategy for making functions more interpretable. For internal uses I do a lot of 

```
#' @param n how many things to do
#' @param df data source. Only works with "foo" for now
do_things <- function(n, df = foo) {
  [code here that assumes df is structured like `foo`]
}
```

That way if I or someone else comes back later with a `foo2`, it's fast and easy to try dropping it in and either see that it works or find out what changes are needed.

File: modules/data.land/R/event_parquet_to_json.R (hunk 6 above)
Comment (divine7022) [reply modules/data.land/R/event_parquet_to_json.R:196]: use case I had in mind was self describing files for traceability (e.g grabbing single events_ens_001.json out of an archive without surrounding filename context), but right that if filename has it, that's canonical source
agree on size cost; happy to drop this one

File: modules/data.land/R/event_parquet_to_json.R (hunk 10 above)
Comment (divine7022) [reply modules/data.land/R/event_parquet_to_json.R:106]: ah that makes sense, it is cleaner than what I had in my head
thanks both !

File: modules/data.land/R/event_parquet_to_json.R (hunk 11 above)
Comment (divine7022) [reply modules/data.land/R/event_parquet_to_json.R:50]: thanks for explanation

File: workflows/preprocess-event-parquet/01b-clean-other-events.R (hunk 7 above)
Comment (infotroph) [reply workflows/preprocess-event-parquet/01b-clean-other-events.R:50]: Good point, @divine7022 -- consider using `PEcAn.data.land::ndti_to_sipnet_tillage()` here

Comment (divine7022) [reply workflows/preprocess-event-parquet/01b-clean-other-events.R:50]: one thing worth flagging before swap, PEcAn.data.land::ndti_to_sipnet_tillage default (slope = 2.5) saturates at delta_ndti at 0.70, while inline here saturates at delta_ndti at 1.0
```
> pct_to_tillage <- function(x) {
+    xfrac <- x / 100; x1 <- pmax(xfrac - 0.3, 0.0); pmin(x1 / 0.7, 1.0)
+  }
>  ndti_to_sipnet_tillage <- function(d, no_till_threshold = 0.30, slope = 2.5) {
+    pmin(pmax((d - no_till_threshold) * slope, 0), 1)
+  }
>  x_pct <- c(0, 30, 50, 70, 100)
>  data.frame(
+    ndti_pct_change = x_pct,
+    inline_01b      = pct_to_tillage(x_pct),
+    pkg_default     = ndti_to_sipnet_tillage(x_pct/100),
+    pkg_slope_1.43  = ndti_to_sipnet_tillage(x_pct/100, slope = 1/0.7)
+  )
  ndti_pct_change inline_01b pkg_default pkg_slope_1.43
1               0  0.0000000         0.0      0.0000000
2              30  0.0000000         0.0      0.0000000
3              50  0.2857143         0.5      0.2857143
4              70  0.5714286         1.0      0.5714286
5             100  1.0000000         1.0      1.0000000
> 
```
If ndti_to_sipnet_tillag default is intended mapping going forward then this is fine. but if we want to preserve the current inline behavior, we can call with slope = 1/0.7
open question: do we want to post swap default (saturation at delta_ndti = 0.70), or pin to slope = 1/0.7 to preserve old tillage_eff numbers ?

```suggestion
```

workflows/preprocess-event-parquet/01b-clean-other-events.R
```diff
@@ -0,0 +1,94 @@
...
+tillage <- arrow::open_dataset(tillage_files, format = "parquet") |>
+  dplyr::filter(
+    is.finite(.data$ndti_pct_change),
+    .data$ndti_pct_change >= 0
+  ) |>
+  dplyr::mutate(
+    site_id = as.integer(site_id),
+    tillage_eff_0to1 = pct_to_tillage(ndti_pct_change),
+    date = as.Date(.data$OGMn_date)
+  ) |>
```
divine7022, line 62: followup https://github.com/PecanProject/pecan/pull/3977#discussion_r3248107943
```suggestion
  dplyr::mutate(
    site_id = as.integer(site_id),
    tillage_eff_0to1 = PEcAn.data.land::ndti_to_sipnet_tillage(
      .data$ndti_pct_change / 100
    ),
    date = as.Date(.data$OGMn_date)
  ) |>
```

workflows/preprocess-event-parquet/01b-clean-other-events.R
infotroph: OK, I figured out why this is here: Arrow knows how to convert the locally defined function into C++ and execute it inside the Parquet query without needing an expensive `collect()` to pull the data back into R, but it doesn't know how to convert the type and range checks in `ndti_to_sipnet_tillage()`.

Unless @ashiklom says otherwise I vote keep the function here but use a slope of 2.5 with a comment citing `PEcAn.data.land::ndti_to_sipnet_tillage`

infotroph:
Heads up that I plan to pull a copy of the workflow scripts into the CCMMF `workflows` repo to add some additional project-specific data cleaning and path conventions. IF everyone agrees, let's continue to merge the version here into the PEcAn `workflows/` directory -- they seem likely to be a better generic starting point for the next person.

workflows/preprocess-event-parquet/01b-clean-other-events.R
dlebauer: Or can the `PEcAn.data.land::ndti_to_sipnet_tillage` function be refactored so that parquet can use it? I think this is the use case that the function was designed for.

dlebauer:
@infotroph if this workflow is designed for CCMMF, I'm not sure I understand the rationale for having a separately maintained CCMMF specific copy. Could you help me understand the rationale?

Isn't this exactly where it would make more sense to have parameterized functions and/or scripts that are called by the CCMMF workflows?

workflows/preprocess-event-parquet/01b-clean-other-events.R
mdietze: Ok, I'm lost, why are we implementing two functions, `pct_to_tillage` and `ndti_to_sipnet_tillage` that are designed to give the same answer? If we need to change where `ndti_to_sipnet_tillage` is defined that's fine, but don't create a duplicate function.

FWIW the function is designed to go from (0.3, 0) to (0.7, 1). @divine7022 the saturation is for SIPNET, which can't have a tillage modifier > 1.

ashiklom: I didn't mean for this to be a big deal.

I was originally developing this outside of PEcAn, so I just created a temporary, simplified copy of the logic as part of this demonstration pipeline. The performance benefits from Arrow hadn't even occurred to me, though they're good to know about.

In general, the point of these workflow scripts is just to demonstrate some working logic and have some simple working examples that demonstrate how the new code works. I didn't intend for these to be "official" or to be picked over so carefully (unlike the actual internal PEcAn code I added, which should get the more thorough review!).

The existing PEcAn ndti_to_tillage implementation is fine and I'll change it here to alleviate all of these concerns.

